PR #143, opened 2014-12-03
Support Atmos command line

greg-king5:
This patch makes cc65-built Atmos files look like BASIC programs (with machine language) instead of pure ML programs.  And, the tape files don't have the autostart flag turned on.  That means:

1. We finally are allowed to type command-line arguments for Atmos programs.
2. We don't need to remember an address number, in order to start an Atmos program; we just type "RUN".


oliverschmidt:
Hi Greg,

First of all sorry for the lack of feedback, but today is really the first day I able to actually put some thoughts into cc65. However I'm pretty sure you know how much I appreciate your contributions to cc65 ;-)

Oliver

cfg/atmos.cfg
```diff
@@ -1,22 +1,30 @@
...
+    BASHDR:   load = BASHEAD, type = ro,  define = yes, optional = yes;
     STARTUP:  load = RAM,     type = ro;
     LOWCODE:  load = RAM,     type = ro,                optional = yes;
-    INIT:     load = RAM,     type = ro,  define = yes, optional = yes;
     CODE:     load = RAM,     type = ro;
     RODATA:   load = RAM,     type = ro;
+    INIT:     load = RAM,     type = ro,  define = yes, optional = yes;
     DATA:     load = RAM,     type = rw;
-    ZPSAVE:   load = RAM,     type = bss, define = yes;
+    ZPSAVE:   load = RAM,     type = rw,  define = yes;
```
oliverschmidt: In case I'm not missing the point... Until some time ago the zpsave area was just `DATA` (and thus `rw`) on all targets. Then someone pointed out that this added ~26 zeros to all binaries. Therefore the `ZPSAVE` segment was introduced. So I don't consider it a nice solution to add those zeros again to the binary. My suggestion is to add some dummy/trailer/<...> `rw` segment after the data segment and place a single byte into it.

cfg/atmos.cfg
```diff
@@ -1,22 +1,30 @@
...
+    __RAMEND__:    type = weak, value = $9800; # graphics RAM not grabbed
+#   __RAMEND__:    type = weak, value = $B400; # graphics RAM grabbed
 }
 MEMORY {
     ZP:      file = "", define = yes, start = $00E2, size = $001A;
-    TAPEHDR: file = %O, type   = ro,  start = $0000, size = $0011;
-    RAM:     file = %O, define = yes, start = $0500, size = $9300 - __STACKSIZE__;
+    TAPEHDR: file = %O, type   = ro,  start = $0000, size = $001F;
+    BASHEAD: file = %O, define = yes, start = $0501, size = $000D;
+    RAM:     file = %O, define = yes, start = __BASHEAD_LAST__, size = __RAMEND__ - __RAM_START__ - __STACKSIZE__;
```
oliverschmidt: > I think that we should go in the opposite direction; that is, we should use more symbols and expressions.

This is once again a question of "quality" vs. "consistency" ;-) I tend to favor consistency. Could I probably persuade you to change the other configuration files in the same way - in another upcoming pull request - so that we improve quality while keeping consistency?

cfg/atmos.cfg
greg-king5: Originally, the zpsave buffer was put somewhere in the middle of the `DATA:` segment.  ld65 _filled_ it with the current `fillval`, so that the rest of `DATA:` would be loaded at its proper place.  If that buffer had been put at the end of `DATA:`, then it wouldn't have been filled.  But, we couldn't control where it was put.  That's why it was moved into its own segment.  Now, we _can_ make sure that it is put after the last segment that has stuff written into the program file.  Therefore, ld65 won't fill it; its space isn't in the program binary.  ("type=bss" stops object files from writing into a segment; it doesn't stop ld65 from _padding_ that segment.)

The new Atmos format does write one byte into `ZPSAVE:`; but, nothing is written after that byte.  Therefore, ld65 doesn't fill the rest of `ZPSAVE:`; the other 25 bytes aren't in the binary.  (I double-checked it by compiling a program, loading it into an emulator, and looking at zpsave's RAM in the emulator's monitor.)

cfg/atmos.cfg
greg-king5: > Could I probably persuade you to change the other configuration files in the same way -- in another upcoming pull request -- so that we improve quality while keeping consistency?

Yes.  (I had planned to mention, in my previous note, that I'm willing to do it; but, I forgot to type it.)

cfg/atmos.cfg
oliverschmidt: Hm, my understanding of segments of type `rw` or `ro` is that they are always written to the outputfile (of course only if the memory area in question is written to the output file). I'd consider it a bug if that wouldn't be true.

Maybe I still don't understand the scenario at hand correctly but here's my test: Have a single asm file containing
```
rts
.data
.res 10
```
and create a binary from it using the config `c64-asm.cfg`. If I understand you correctly you say that the binary would contain only 1 byte (from the `rts`) and the `.res 10` wouldn't go into the output file as it is the only content (and therefore for sure the last content) of the `data` segment. However my finding is that the binary contains 11 bytes.

cfg/atmos.cfg
oliverschmidt: Cool :-)

cfg/atmos.cfg
greg-king5: OK.  I looked at my tests in a different way; and, I saw that you're correct.  So, I cut the `ZPSAVE:` segment into two pieces -- one byte is `rw` and the rest is `bss`.

cfg/atmos.cfg
```diff
@@ -1,22 +1,31 @@
...
     STARTUP:  load = RAM,     type = ro;
     LOWCODE:  load = RAM,     type = ro,                optional = yes;
-    INIT:     load = RAM,     type = ro,  define = yes, optional = yes;
     CODE:     load = RAM,     type = ro;
     RODATA:   load = RAM,     type = ro;
+    INIT:     load = RAM,     type = ro,  define = yes, optional = yes;
     DATA:     load = RAM,     type = rw;
-    ZPSAVE:   load = RAM,     type = bss, define = yes;
+    ZPSAVE1:  load = RAM,     type = rw,  define = yes;
+    ZPSAVE2:  load = RAM,     type = bss;
```
oliverschmidt: Great that we sorted the ".res in rw segment" question out :-) Given that you are at least as much into formal aspects as I am I'm wondering about the naming of `ZPSAVE1` and `ZPSAVE2`. As far as I understand the byte in `ZPSAVE1` has nothing to do at all with saving ZP values. May I ask you to rename to something like DUMMY / BASTRAIL or <...> that makes sense (as suggested before)?

greg-king5: `ZPSAVE1:` and `ZPSAVE2:` are the components of the original `ZPSAVE:` segment.  Together, those two are the same size, sit in the same place, and do the same job, as the original one.

I need to add an extra byte to the end of the program file; but, I don't want to add an _extra_ byte to the program's memory map.  So, I overlay it on the buffer that will save some original content from the zero page.  Therefore, I needed to define two types in the `ZPSAVE:` segment.  But, ld65 won't allow it.  So, I fake it by defining two segments, then using them together as though they're one segment.
[See the "crt0.s" diff.](https://github.com/cc65/cc65/pull/143/files#diff-1f6c7aeb02f4f3f0722799ba433435a8R81)

While the program isn't running, that extra byte protects the program from the ROM bug.  While the program is running, that same space is used by the `zpsave` buffer.

oliverschmidt: Sorry, I didn't notice that you want to reuse that byte. Thanks for your patience with me :-)

PR #145, opened 2015-01-11
Move SBC to correct location in 6280 instruction table

wlabelle:
I found the SBC instruction for the HuC6280 was not in the proper sorted order and
was being missed by the binary search of FindInstruction.


oliverschmidt:
First of all thanks for contributing to cc65@GitHub :-)

Apart from that: Nice find!

mrdudz:
btw, just in case you feel like it... i have put my pcengine library work here: https://github.com/mrdudz/cc65/tree/pcenginetarget - it doesnt work yet with the current compiler/toolchain (i used it to port contiki ages ago), but shouldnt be terribly hard to fix.

PR #146, opened 2015-02-08
Add target "osic1p" for Ohio Scientific Challenger 1P

smuehlst:
Support for Ohio Scientific Challenger 1P:

* New target called "osic1p"
* conio implementation complete except for kbhit()
* No support for stdio


libsrc/osic1p/osic1p.inc
```diff
@@ -0,0 +1,5 @@
+
+SCRNBASE        := $D000    ; Base of video RAM
+VIDEORAMSIZE    := $0400    ; Size of C1P video RAM (1 kB)
```
greg-king5: The ca65 assembler has two types of assignment operators: label (":=") and equate ("=").  Ca65 doesn't care which one we use; but, they do affect [the information](https://github.com/cc65/wiki/wiki/Debug-info-data#symbols) in [a debug file](https://github.com/cc65/wiki/wiki#debug-information).

`VIDEORAMSIZE` isn't an address; therefore, you should use the equate operator to create it.

cfg/osic1p-asm.cfg
```diff
@@ -0,0 +1,20 @@
...
+    RAM:      file = %O, define = yes, start = %S, size = __HIMEM__ - __STACKSIZE__ - %S;
+}
+SEGMENTS {
+    INIT:     load = RAM, type = ro,  define = yes, optional = yes;
+    CODE:     load = RAM, type = rw;
+    RODATA:   load = RAM, type = rw;
+    DATA:     load = RAM, type = rw;
+    BSS:      load = RAM, type = bss, define = yes;
+    ZEROPAGE: load = ZP,  type = zp;
+}
```
oliverschmidt: Please add a final newline here.

doc/osi.sgml
```diff
@@ -0,0 +1,181 @@
...
+<sect>Loadable drivers<p>
+
+There are no loadable drivers available.
+
+<sect>Limitations<p>
+
+<sect1>conio implementation<p>
+
+The conio implementation is complete except for the kbhit() function. A
+call to cgetc() always blocks until a character is entered.
```
oliverschmidt: Isn't the statement

> A call to cgetc() always blocks until a character is entered.

generally true for all implementations? Or do you want to state that a key typed before entering cgetc() can't be fetched. In the latter case you should probably state that (even more) explicitly.

libsrc/osic1p/cgetc.s
```diff
@@ -0,0 +1,29 @@
...
+        .include        "osic1p.inc"
+        .include        "extzp.inc"
+
+; Input routine from 65V PROM MONITOR, show cursor if enabled
+_cgetc:
+        lda     cursor          ; show cursor?
+        beq     nocursor
+        ldy     CURS_X
+        lda     (SCREEN_PTR),y  ; fetch current character
+        sta     CURS_SAV        ; save it
```
oliverschmidt: As far as I can see there's no need for a special zero page variable here. Instead you can re-use a standard variable (like `tmp1`). There are quite some you can use in every leaf function. See https://github.com/cc65/wiki/wiki/Using-runtime-zeropage-locations-in-assembly-language

libsrc/osic1p/cgetc.s
```diff
@@ -0,0 +1,29 @@
...
+        jsr     INPUTC
+        pha                     ; save retrieved character
+        lda     cursor          ; was cursor on?
+        beq     nocursor2
+        lda     CURS_SAV        ; fetch saved character
+        ldy     CURS_X
+        sta     (SCREEN_PTR),y  ; store at cursor position
+nocursor2:
+        pla                     ; restore retrieved character
+        rts
```
oliverschmidt: - Setting the Z flag for the `beq nocursor2`can be done without destroying A. This way the cycles to restore it need only be spent when the cursor actually needs to be removed.
- Functions declared to return an (unsigned) char still need to return a valid (unsigned) int !!! Therefore you'll see quite some `ldx #$00`in function epilogs.
- If you use X to check for `cursor` then you can avoid the cycles for the `ldx #$00` in case `cursor`is false.
```
nocursor:
  jsr INPUTC
  ldx cursor
  beq done
  tax
  lda tmp1
  ldy CURS_X
  sta (SCREEN_PTR),y
  txa
  ldx #$00
done:
  rts
```

You may think this is "over optimizing" but that's in general the amount of optimization applied to cc65 C library code.

libsrc/osic1p/crt0.s
```diff
@@ -0,0 +1,63 @@
...
+_init:    ldx     #$FF                 ; Initialize stack pointer to $01FF
+          txs
+          cld                          ; Clear decimal mode
+
+; ---------------------------------------------------------------------------
+; Initialize screen width
+; TODO: Can initialization be done in a more idiomatic way?
+; TODO: Create function for changing screen width
+          lda     #$18
+          sta     SCR_LINELEN
```
oliverschmidt: It seems to me as if `SCR_LINELEN` would be a constant. In that case we surely don't want to spend a byte in RAM to load it from but rather use immediate opcodes. If you want to avoid a magic constant in your code you can define a symbol in "your" .inc file - using `=` instead of `:=` to denote that it isn't a (zeropage) address - see Gregs comment.

Note 1: If `SCR_LINELEN` would actually be a variable necessary for conio I'd suggest to create a constructor for it. We don't want to "pollute" crt0.s with initializations for specific C library modules like conio. See https://github.com/cc65/cc65/blob/master/libsrc/apple2/cputc.s for an example of a constructor for conio.

libsrc/osic1p/crt0.s
```diff
@@ -0,0 +1,63 @@
...
+          sta     sp
+          lda     #>(__RAM_START__ + __RAM_SIZE__)
+          sta     sp+1
+
+; ---------------------------------------------------------------------------
+; Initialize memory storage
+; copydata seems to be only necessary for special systems
+
+          jsr     zerobss              ; Clear BSS segment
+          ; jsr     copydata           ; Initialize DATA segment
```
oliverschmidt: I have a strong dislike for commented out code without a clear statment why it is commented out.

libsrc/osic1p/extzp.s
```diff
@@ -0,0 +1,21 @@
...
+; ------------------------------------------------------------------------
+
+        .include        "extzp.inc"
+
+.segment        "EXTZP" : zeropage
+
+CURS_X:         .byte   0
+CURS_Y:         .byte   0
+CURS_SAV:       .byte   0
+SCR_LINELEN:    .byte   24
```
oliverschmidt: The `.byte <value>`directive is only useful in segments that get written (like `TEXT`, `DATA`, `RODATA`). But in bss-type segments (like `BSS`, `ZP`, `EXTZP`) the value is lost anyway. So I suggest to use only the `.res` directive for definitions in bss-type segments.

libsrc/osic1p/gotox.s
```diff
@@ -0,0 +1,13 @@
+;
+; copied from CBM implementation
+; originally by:
+; Ullrich von Bassewitz, 07.08.1998
+;
+; void gotox (unsigned char x);
+;
+        .export         _gotox
+        .import         plot           
+        .importzp       CURS_X
```
oliverschmidt: Include your "extzp.inc" instead.

libsrc/osic1p/gotoxy.s
```diff
@@ -0,0 +1,17 @@
...
+; copied from CBM implementation
+;
+; originally by:
+; Ullrich von Bassewitz, 06.08.1998
+;
+; void gotoxy (unsigned char x, unsigned char y);
+;
+        .export         _gotoxy
+        .import         popa, plot
+        .importzp       CURS_X, CURS_Y
```
oliverschmidt: Include your "extzp.inc" instead.

libsrc/osic1p/gotoy.s
```diff
@@ -0,0 +1,13 @@
+;
+; copied from CBM implementation
+; originally by:
+; Ullrich von Bassewitz, 0.08.1998
+;
+; void gotoy (unsigned char y);
+;
+        .export         _gotoy
+        .import         plot
+        .importzp       CURS_Y
```
oliverschmidt: Include your "extzp.inc" instead.

libsrc/osic1p/wherex.s
```diff
@@ -0,0 +1,14 @@
+;
+; Ullrich von Bassewitz, 06.08.1998
+; Copied from cbm/wherex.s
+
+; unsigned char wherex (void);
+;
+        .export         _wherex
+        .import         CURS_X: zp
```
oliverschmidt: Include your "extzp.inc" instead.

libsrc/osic1p/wherey.s
```diff
@@ -0,0 +1,14 @@
+;
+; Ullrich von Bassewitz, 06.08.1998
+; Copied from cbm/wherey.s
+;
+; unsigned char wherey (void);
+;
+        .export         _wherey
+        .import         CURS_Y: zp
```
oliverschmidt: Include your "extzp.inc" instead.

libsrc/osic1p/clrscr.s
```diff
@@ -0,0 +1,31 @@
+;
+; void clrscr (void);
+;
+        .export         _clrscr
+        .import         plot
+        .importzp       CURS_X, CURS_Y
```
oliverschmidt: Include your "extzp.inc" instead.

oliverschmidt:
Hi Stephan,

Thanks for providing this pull request. It's pretty clear that you invested time to adhere to the coding style and especially to provide useful documentation :-)

Please note the GitHub won't automatically create a notification for me if you add commits to this pull request so please be sure to write a short comment here when you're done with your adjustments.

Regards,
Oliver

doc/osi.sgml
smuehlst: Right, that statement is redundant, I'll remove it. I had implied somehow that on systems where kbhit() is working and returns false, a call to cgetc() would also return immediately and tell the caller that nothing is available, but that's obviously not the case.

smuehlst:
Greg and Oliver,

thanks for the detailed comments. It will take a while until I'm through with these, and I'll update the pull request once I'm finished.

Best regards
Stephan

smuehlst:
Greg and Oliver,

I implemented all the suggested changes plus the following:

* Added a dummy kbhit() function that always returns true
* Added Ohio Scientific Challenger 1P to the list of platforms in the top-level README.md

Best regards
Stephan

libsrc/osic1p/chline.s
```diff
@@ -0,0 +1,29 @@
...
+        pha                     ; Save the length
+        jsr     popa            ; Get y
+        jsr     _gotoxy         ; Call this one, will pop params
+        pla                     ; Restore the length
+
+_chline:
+        cmp     #0              ; Is the length zero?
+        beq     L9              ; Jump if done
+        sta     tmp1
+L1:     lda     #$94             ; Horizontal line, screen code
```
oliverschmidt: Sorry for the nitpicking but please align the comment.

libsrc/osic1p/clrscr.s
```diff
@@ -0,0 +1,31 @@
...
+        ldy       #BANKS
+        ldx       #$00
+staloc:
+        sta       SCRNBASE,X
+        inx
+        bne       staloc
+        inc       staloc+2
+        dey
+        bne       staloc
+        lda       #>(SCRNBASE); load high byte
```
oliverschmidt: Please put the comments in this file in the same column as in the other files.

libsrc/osic1p/cvline.s
```diff
@@ -0,0 +1,29 @@
...
+        pha                     ; Save the length
+        jsr     popa            ; Get y
+        jsr     _gotoxy         ; Call this one, will pop params
+        pla                     ; Restore the length and run into _cvline
+
+_cvline:
+        cmp     #0              ; Is the length zero?
+        beq     L9              ; Jump if done
+        sta     tmp1
+L1:     lda     #$95             ; Vertical bar
```
oliverschmidt: Please align the comment.

libsrc/osic1p/osic1p.inc
```diff
@@ -0,0 +1,9 @@
+; Addresses
+SCRNBASE        := $D000    ; Base of video RAM
```
oliverschmidt: Please put the comments in this file in the same column as in the other files.

libsrc/osic1p/crt0.s
```diff
@@ -0,0 +1,54 @@
...
+
+; ---------------------------------------------------------------------------
+; Place the startup code in a special segment
+
+.segment  "STARTUP"
+
+; ---------------------------------------------------------------------------
+; A little light 6502 housekeeping
+
+_init:    ldx     #$FF                 ; Initialize stack pointer to $01FF
```
oliverschmidt: Please put the comments in this file in the same column as in the other files.

oliverschmidt:
Hi Stephan,

First of all thanks for the update pull request!

> Added a dummy kbhit() function that always returns true

I must admit that I have an issue with this.

In general it's not usual to add dummy functions to a target "just to" make the samples/demos link. The Apple II targets i.e. don't have a clock() function. Instead of hiding that fact with some dummy it is explicitly taken care of i.e. in https://github.com/cc65/cc65/blob/master/samples/sieve.c 

Specifically kbhit() is even a good example for not trying to provide "clever" dummy functions. It might be used to i.e. just flash a cursor while waiting for input. In that scenario returning true might work as the cursor just doesn't flash because the program runs right into the cgetc() and blocks there. But kbhit() might as well be used to run some demo until the user has seen enough. Then it would be better to return false to see the demo at all (and end it via reset or power cycle).

And finally I'm not sure if you are aware that the cc65 targets tend to not supply a firmware function suitable to build kbhit() upon. Usually it is necessary to rather bypass the firmware and check some memory-mapped hardware register. Maybe that's true for this target too.

Anyway if you can't come up with a working kbhit() please omit it. If you want to you can modify samples/demos/... coming with cc65 or in http://sourceforge.net/projects/cc65/files/contrib/ which work otherwise on this target to not depend on kbhit() in a subsequent pull request.

smuehlst:
Oliver,

thanks for the additional review and feedback, all done.

> Specifically kbhit() is even a good example for not trying to provide "clever" dummy functions.

I was also thinking back and forth whether the dummy kbhit() is really a good idea. Removed again. A real implementation is planned for a later stage.

Best regards
Stephan

libsrc/osic1p/clrscr.s
```diff
@@ -0,0 +1,31 @@
...
+        .include        "extzp.inc"
+        .include        "osic1p.inc"
+
+; Adapted from the Challenger Character Graphics
+; Reference Manual, "2.3.3 MACHINE LANGUAGE SCREEN CLEAR"
+; This is self-modifying code!
+BANKS = VIDEORAMSIZE / $100
+
+_clrscr:
+        lda       #$20          ; ' '
```
oliverschmidt: I'm not going to keep this pull request further from being merged but please be aware that the assembler allows to directly define ASCII constants...

oliverschmidt:
@Stephan: I suggest that you post an announcement for the new target on the mailing list.

libsrc/osic1p/clrscr.s
smuehlst: > please be aware that the assembler allows to directly define ASCII constants...

Understood. I'm getting slowly more proficient in 6502 assembler...

Thanks for merging the pull request!

oliverschmidt: > Thanks for merging the pull request!

Thanks for the substantial contribution!

smuehlst:
> I suggest that you post an announcement for the new target on the mailing list.

Will do.

libsrc/osic1p/crt0.s
```diff
@@ -0,0 +1,54 @@
...
+_init:    ldx     #$FF          ; Initialize stack pointer to $01FF
+          txs
+          cld                   ; Clear decimal mode
+
+; ---------------------------------------------------------------------------
+; Set cc65 argument stack pointer
+
+          lda     #<(__RAM_START__ + __RAM_SIZE__)
+          sta     sp
+          lda     #>(__RAM_START__ + __RAM_SIZE__)
```
greg-king5: I'm surprised that we didn't see this bug before now!  That expression points to the bottom of the C data stack.  It needs to point to the top, that is: "(__RAM_START__ + __RAM_SIZE__ + __STACKSIZE__)".

oliverschmidt: Yep - maybe as additional explanation: Thats because of the `- __STACKSIZE__` in the linker config file...

smuehlst: Greg, thanks for spotting this. I will include the fix in the next pull request.

PR #148, opened 2015-02-28
kbhit() implementation, fixed C data stack pointer initialization

smuehlst:
Significant changes since last pull request:

* kbhit() implementation with one-character buffer for cgetc()
* fixed incorrect initialization of C data stack pointer


cfg/osic1p.cfg
```diff
@@ -7,7 +7,7 @@ SYMBOLS {
 }
 MEMORY {
     # for size of ZP see runtime/zeropage.s and c1p/extzp.s
-    ZP:       file = "", define = yes, start = $0002, size = $001A + $0004;
+    ZP:       file = "", define = yes, start = $0002, size = $001A + $0005;
```
greg-king5: A memory area's `size` attribute gives a maximum limit.  The area's actual size can be smaller.  Therefore, if you think that you might add more variables to `extzp`, at some future date, then you can set `size` to some bigger number.  Then, you wouldn't need to edit the configure file _every_ time that you did create another variable.

libsrc/osic1p/cgetc.s
```diff
@@ -8,8 +10,21 @@
...
+        rts
+
 ; Input routine from 65V PROM MONITOR, show cursor if enabled
 _cgetc:
+        lda     CHARBUF         ; character in buffer available?
+        beq     nobuffer
+        tax                     ; save character in X
+        lda     #$00
+        sta     CHARBUF         ; empty buffer
+        jmp     restorex        ; restore X and return
```
greg-king5: That instruction can be `beq restorex` (it always will branch).

cfg/osic1p.cfg
smuehlst: I had thought that it would be a good thing not reserve unused memory here. I will add some extra space the next time I add a variable.

libsrc/osic1p/cgetc.s
smuehlst: Ok, thanks for the suggestion. Branch updated.

libsrc/osic1p/cclear.s
```diff
@@ -22,7 +22,7 @@ _cclear:
         cmp     #0              ; Is the length zero?
         beq     L9              ; Jump if done
         sta     tmp1                                 
-L1:     lda     #$20            ; Blank - screen code
+L1:     lda     #' '
```
oliverschmidt: Nice that you remembered this...

libsrc/osic1p/cgetc.s
```diff
@@ -8,8 +10,21 @@
         .include        "extzp.inc"
         .include        "zeropage.inc"
 
+; Initialize one-character buffer that is filled by kbhit()
+initcgetc:
+        lda     #$00
+        sta     CHARBUF         ; No character in buffer initially
+        rts
```
oliverschmidt: I don't want to say that it would be preferable beause every access would be both longer and slower but nevertheless I'd like to point out that moving `CHARBUF` from zeropage to the BSS segment would make this constructor obsolete as the BSS segment is completely initialized to zero by the startup code in ctr0.s.

BTW: A variable that needs an initial value different from zero can be placed in the DATA segment. Then again there's no need for explicit initialization code as the value is read from the binary (just like code).

libsrc/osic1p/cputc.s
```diff
@@ -48,8 +48,23 @@ newline:
...
-        lda     #0              ; wrap around to line 0
-        sta     CURS_Y
+        dec     CURS_Y          ; bottom of screen reached, scroll
+        ldx     #0
+scroll: lda     SCRNBASE+$00A5,x
+        sta     SCRNBASE+$0085,x
+        lda     SCRNBASE+$01A5,x
+        sta     SCRNBASE+$0185,x
+        lda     SCRNBASE+$02A5,x
+        sta     SCRNBASE+$0285,x
```
oliverschmidt: I don't know the target well but in general I'd presume that these "magic constants" could be replaced by some simple expressions (beside 'plus' you have i.e. 'multiply' and 'shift') and potentially using binary constants instead of hex constants that makes to more obvious what they mean.

libsrc/osic1p/cclear.s
smuehlst: I hoped it would earn me some bonus points for the rest of the review :jack_o_lantern:

libsrc/osic1p/cputc.s
smuehlst: I tried to improve the code by using symbolic constants plus a .repeat macro. Change commited and pull request updated.

PR #149, opened 2015-03-05
osic1p character translations

greg-king5:
In the Ohio Scientific computers, the character font ROM has three characters (right brace, vertical bar, and tilde) in positions that are different from where ASCII puts them.  Therefore, I added a character translation table to the tools, so that those character literals will be printed properly.


oliverschmidt:
Thanks for your improvements of the OSI C1P target !

PR #150, opened 2015-03-07
A second OSI C1P program file format.

greg-king5:
I created a new file format.  It's loadable; it doesn't need to be converted.

But, one of the emulators can't handle this format.  Therefore, it can't replace the original format.  I chose to make the new one the alternate format because it's easier to select it, than it would be to select the original format (if it were the alternate).


libsrc/osic1p/bootstrap.s
```diff
@@ -0,0 +1,138 @@
...
+
+; ------------------------------------------------------------------------
+
+load_addr       :=      __RAM_START__
+load_size       =       __BSS_RUN__ - __RAM_START__
+ram_top         :=      __RAM_START__ + __RAM_SIZE__
+
+        .segment        "BOOT"
+
+.ifdef ASM
```
oliverschmidt: I seems to me that the purpose and usage of `ASM` deserves some explanation.

doc/osi.sgml
```diff
@@ -34,42 +35,63 @@ information.
...
+  You can make the alternate format by adding the option <tt/-u __BOOT__/ to
+  <tt/cl65/'s or <tt/ld65/'s command lines.
 
-<tscreen>
-srec_cat hello -binary -offset 0x200 -o hello.c1p -Ohio_Scientific -execution-start-address=0x200
-</tscreen>
+  This format doesn't need to be converted. It is smaller than the text-only
+  format.  But, it cannot be loaded by <url
+  url="http://www.pcjs.org/docs/c1pjs/" name="C1Pjs">; you must use the
+  SRecord-produced text-only format with that emulator.
```
oliverschmidt: Just to make sure I got this right: A physical machine can make use of the new "self booting" format but this emulator can't, correct?

cfg/osic1p.cfg
```diff
@@ -2,24 +2,28 @@ FEATURES {
     STARTADDRESS: default = $0200;
 }
 SYMBOLS {
+# Add "-u __BOOT__" to a command line if you want a file that can be loaded directly.
+#    __BOOT__:      type = import;
```
oliverschmidt: I don't get the point of this line when - as then line above states - the user is supposed to "undefine" the symbol on the cmdline. I'm in general unsure if this type of inline documentation belongs here. After all the symbol isn't referenced anywhere in this file. After all I'd rather omit both line - and probably add to the documentation that an alternative to the `-u __BOOT__` may be to add `__BOOT__: type = import;` to the `SYMBOLS` section of the cfg file.

oliverschmidt:
Cool stuff - you seem to be pretty interested in the target - do you happen to own a physical machine?

libsrc/osic1p/bootstrap.s
greg-king5: OK, done.

doc/osi.sgml
greg-king5: In the intro document, I describe how to load a program into a real, physical machine and two emulators.  The real box and one of the emulators can handle the new format.  The Javascript-based emulator can't handle it because that program doesn't emulate the cassette player device.  Instead, it auto-types text-only files through the keyboard emulation.

cfg/osic1p.cfg
greg-king5: The upper-case `-U` option undefines names.  The lower-case `-u` option "uses" names (i.e., it makes the linker import them).

I included those commented lines as a convenience to programmers.  If they never use **C1Pjs**, then they _simply_ could uncomment the lines in the two configure files; they wouldn't need to type a full line into those files.
I added a mention about it in the documentation.

P.S.  I also like the symmetry of having it there; it's a companion of the `HEAD:` and `BOOT:` lines.

greg-king5:
> Cool stuff -- you seem to be pretty interested in the target -- do you happen to own a physical machine?

No, I own only a Commodore 64.  But, if a target has an emulator and enough documentation online, then I enjoy making the cc65 project more useful for that target.

cfg/osic1p.cfg
oliverschmidt: > If they never use C1Pjs, then they simply could uncomment the lines in the two configure files; they wouldn't need to type a full line into those files.

Sorry for being ambiguous. The above was in fact clear to me. But I think it should be mentioned in a comment. Maybe it would make much more sense to tell that in a comment than the '-u ...' hint.

Or to put it in other words: As always I don't like code commented out without explanation why it is commented out. And your comment above the line in question does _not_ actually explain why the line in question is commented out.

libsrc/osic1p/bootstrap.s
```diff
@@ -0,0 +1,151 @@
...
+; If you want to change how this bootstrap loader works, then:
+; 1. edit this assembly source code,
+; 2. define the constant ASM (uncomment the line below),
+; 3. assemble this file (and, make a listing of that assembly),
+; 4. copy the listing's hex codes into the .byte lines below (notice that most
+;    of the strings are followed by CR; it's required by the OS65V monitor)
+;    (be sure to match the listing's lines against the .byte lines),
+; 5. undefine ASM (recomment the line),
+; 6. assemble this file, again,
+; 7. and, add the object file to "osic1p.lib".
```
oliverschmidt: Reading this comment I'm happy to have asked for it :-)

cfg/osic1p.cfg
greg-king5: >> If they never use C1Pjs, then they simply could uncomment the lines in the two configure files; they wouldn't need to type a full line into those files.
>
> The above was, in fact, clear to me. But, I think it should be mentioned in a comment. Maybe it would make much more sense to tell that in a comment than the '-u ...' hint.

Good point.
Is the new comment good enough?

oliverschmidt: > Is the new comment good enough?

Not only "good enough" but just fine :-)

PR #152, opened 2015-03-14
Fix for use of uninitialized memory in osic1p bootstrap.s

smuehlst:
The low byte of address "load" was never initialized in the new bootloader code for the osic1p target. While this accidentally might work with an emulator that implicitly initializes all memory to zero, on a real machine this causes a random displacement where the program will be loaded. Therefore the boot loader fails to load a program on a real machine.

I discussed my original fix with @greg-king5, and the pull request I'm submitting now is improved according to his suggestions. See also the discussion in my [original pull request to Greg](https://github.com/greg-king5/cc65/pull/2).


oliverschmidt:
> I discussed my original fix with @greg-king5, and the pull request I'm submitting now is improved according to his suggestions. See also the discussion in my original pull request to Greg.

I'm happy to see that you both seem to collaborate well on the topic :-)

PR #153, opened 2015-03-17
Support for different screen layouts via configuration modules

smuehlst:
Added support for different screen layouts. All screen-related parameters were moved to the screen.s module, which contains the default parameters for the standard 24 columns by 24 lines mode.

The default layout can be overridden by specifying a module from the libsrc/osic1p/extra directory on the linker command line.

The new code was tested successfully with a real Challenger 1P (default mode 24x24) and a [Briel Superboard /// replica](http://www.brielcomputers.com/wordpress/?cat=39) in 32x32 mode.


doc/osi.sgml
```diff
@@ -161,6 +161,42 @@ There is no specific support for direct hardware access.
...
+for the Challenger 1P, like under BASIC. In addition to that there is support
+for other screen layouts with extra configuration modules.
+
+There is a default module <tt/screen.s/ in the OSI-specific
+cc65 runtime library. This module contains all parameters
+related to the screen layout. No further configuration is needed for using the
+default screen layout of the Challenger 1P.
+
+For other screen layouts additional copies of the screen.s module must
+be created in the <tt>libsrc/osic1p/extra</tt> directory. The cc65 build
```
oliverschmidt: The documentation at hand is targeted towared the cc65 user (aka the C or ASM programmer using it). However what is described here is relevant for the cc65 developer only. So it doesn't belong here.

libsrc/osic1p/cputc.s
```diff
@@ -36,10 +41,13 @@ L1:     cmp     #$0D            ; LF?
 cputdirect:
```
oliverschmidt: Just to make sure there's no misunderstanding: When we discussed the usage of 'extra' modules to support different screen sizes I presumed that you'd create modules which export `cputdirect` in a different way. If you can however come up with a reasonable way to have only one implementeation of `cputdirect` just with different parameters coming form 'extra' modules than this is in general fine with me.

But please keep in mind that cc65 C library programming is mostly about size minimizing. If having several `cputdirect` implementations for diffrent screen sizes would result in (significant) smaller code size than the current approach then I'd argue to modify it. If you don't want to repeat similiar code paths then maybe phrasing the actual code as macro (which is called with different parameters) in the different modules might be a way to go. The various Atari TGI drivers follow a similiar pattern: https://github.com/cc65/cc65/tree/master/libsrc/atari/tgi

And finally I'd like to point out that CONIO is in general not required to support scrolling. From the API perspective the behaviour of writing beyond the lower right corner of the screen is just undefined. Programs doing it might simply crash or end up to continue writing in the upper left corner or <...>. The reason for that design decision is that the CONIO implementation is supposed to be small (and fast).

I don't know of the target(s) at hand but usually cc65 targets with text i/o capabilities come with some firmware allowing for (sometimes rather slow) character output - and that firmware code usually supports scrolling. So the usual approach is to map the firmware functionality (with scrolling) to POSIX I/O on stdout (and stderr) aka file number 1 (and 2)  and implement CONIO with as-fast-as-possible direct video RAM access (without scrolling).

In case you want to argue that the POSIX I/O approach doesn't apply to the target(s) at hand because of the missing mass storage support you should know that it is perfectly acceptable to only support stdin, stdout and stderr and have no open() / close() at all.

doc/osi.sgml
smuehlst: Ok, I will shorten that paragraph accordingly.

libsrc/osic1p/cputc.s
smuehlst: I hadn't though about using macros to accomplish the same goal, and that's certainly an alternative that will result in slightly smaller code. I will try to change the implementation to macros.

Regarding the scrolling in conio I would prefer to leave that in place. This was actually the first contribution that I received from another user who needed it.

At the end of line the conio functions are supposed to wrap around to the next line, aren't they? So a comparison for end-of-line is necessary in any case. Scrolling at the end of the screen makes that behavior consistent. If the programmer does not rely on scrolling and takes care himself not to run over the end of the screen, the only speed impact will be one comparison for end-of-screen that never evaluates to true. I hope that this minimal speed impact is acceptable.

POSIX I/O is on the radar, and in my plan it will be based on the conio implementation for stdout/stderr/stdin..

oliverschmidt: > [...] I hope that this minimal speed impact is acceptable.

Surely! But the whole discussion isn't about speed but about size. Other targets intentionally omit scrolling to get a smaller `cputdirect`...

smuehlst: Ok, at the moment 100% of the known users of the osic1p target are willing to accept slightly larger cputdirect code for getting support for scrolling :grimacing:. Let's talk about removing scrolling support from conio once we have POSIX I/O.

oliverschmidt: If I wanted to have you remove the scrolling support from your CONIO implementation I would have already stated so before merging your recent pull request ;-)

I just wanted to provide you with an overall overview how things are usually setup on other (comparable) targets. This pull request seems to invest quite some effort in scrolling for different screen sizes so I wanted to make sure you know that scrolling isn't required in the first place - and explain why this was decided this way.

Regarding the single user: Is he using CONIO features like cursor positioning and/or text attributes? If not then he might actually rather be a POSIX I/O user only driven to your CONIO because of POSIX I/O missing ;-)

Regarding POSIX I/O based on CONIO: Other targets with well known firmware entry points for character output tend to have a redirection mechanism of one kind or another allowing BASIC programs to print another terminal, a printer, a different video adapter, <...>. The idea of mapping POSIX I/O to those firmware entry points for character output is to allow those redirection mechanisms to work like with BASIC programs.

smuehlst: > Regarding the single user: Is he using CONIO features like cursor positioning and/or text attributes? If not then he might actually rather be a POSIX I/O user only driven to your CONIO because of POSIX I/O missing ;-)

That is actually the case, but conio is better than nothing.

> Regarding POSIX I/O based on CONIO: Other targets with well known firmware entry points for character output tend to have a redirection mechanism of one kind or another allowing BASIC programs to print to another terminal, a printer, a different video adapter, <...>. The idea of mapping POSIX I/O to those firmware entry points for character output is to allow those redirection mechanisms to work like with BASIC programs.

I will do some reseach, but as far as I know there are no such features in the OSI BASIC.

smuehlst:
Closing this pull request. I will come back when I have the macro-based implementation ready.

PR #154, opened 2015-03-24
Support for different screen layouts via macro-based modules

smuehlst:
* Reimplementation of the [previous attempt](https://github.com/cc65/cc65/pull/153) to provide support for different screen layouts. The reimplementation is now based on macros, which saves 48 bytes of code and read-only data compared to the old implementation. Tested successfully on real machines: Challenger 1P and Briel Superboard ///.
* Documentation for the \_\_OSIC1P\_\_ predefined macro/symbol for cc65/ca65.
* Reserved by default more room for ZP segment to avoid constant recalculation of the size when new variables are added ([suggestion](https://github.com/cc65/cc65/pull/148#discussion_r25559563) by @greg-king5).


oliverschmidt:
I must admit that I didn't check every instruction (again) but the general layout of th macro-based approach looks fine :-)

PR #155, opened 2015-04-20
Added function .ADDRSIZE to ca65

Movax12:
Hi!
I have a few ideas I want to add to ca65. This one was borne from the need to be able to check the address size of a symbol when creating custom instructions (that have different opcodes - like zeropage or absolute) with macros. This information is known by ca65 if the symbol is constant or imported. This function can give that information to a macro. 

See here for a practical example: http://forums.nesdev.com/viewtopic.php?p=121457#p121457

More to come, but one at a time I think is best.

### DIFF
--- a/doc/ca65.sgml
+++ b/doc/ca65.sgml
@@ -1278,6 +1278,35 @@ writable.
 Pseudo functions expect their arguments in parenthesis, and they have a result,
 either a string or an expression.
 
+<sect1><tt>.ADDRSIZE</tt><label id=".ADDRSIZE"><p>
+
+  The <tt/.ADDRSIZE/ function is used to return the interal address size 
+  associated with a symbol. This can be helpful in macros when knowing the address 
+  size of symbol can allow for custom instructions.
+
+  Example:
+
+  <tscreen><verb>
+        .macro myLDA foo
+
+  	        .if .ADDRSIZE(foo) = 1
+  	          	        ;do custom command based on zeropage addressing:
+  	          	        .byte .A5h, foo
+  	        .elseif .ADDRSIZE(foo) = 2
+  	          	        ;do custom command based on absolute addressing:
+  	          	        .byte .ADh, foo
+  	        .elseif .ADDRSIZE(foo) = 0
+  	          	        ; no address size define for this symbol:
+  	          	        .out .sprinft("Error, address size unknown for symbol %s", .string(foo))
+  	        .endif
+        .endmacro
+  </verb></tscreen>
+
+  This command is new and must be enabled with the <tt/.FEATURE addrsize/ command.
+
+  See: <tt><ref id=".FEATURE" name=".FEATURE"></tt>
+
+
 
 <sect1><tt>.BANK</tt><label id=".BANK"><p>
 
@@ -2596,6 +2625,12 @@ Here's a list of all control commands and a description, what they do:
 
   <descrip>
 
+  <tag><tt>addrsize</tt><label id="addrsize"></tag>
+
+    Enables the .ADDRSIZE pseudo function. This function is experimental and not enabled by default.
+
+    See also: <tt><ref id=".ADDRSIZE" name=".ADDRSIZE"></tt>
+
   <tag><tt>at_in_identifiers</tt><label id="at_in_identifiers"></tag>
 
     Accept the at character (`@') as a valid character in identifiers. The
@@ -2647,11 +2682,6 @@ Here's a list of all control commands and a description, what they do:
     reserved keywords built into the assembler, that starts with a dot, may be
     overridden. When using this feature, you may also get into trouble if
     later versions of the assembler define new keywords starting with a dot.
-
-  <tag><tt>loose_char_term</tt><label id="loose_char_term"></tag>
-
-    Accept single quotes as well as double quotes as terminators for char
-    constants.
 
   <tag><tt>loose_string_term</tt><label id="loose_string_term"></tag>
 
--- a/doc/ca65.sgml
+++ b/doc/ca65.sgml
@@ -1297,7 +1297,7 @@ either a string or an expression.
                         .byte 0ADh
                         .word foo
                 .elseif .ADDRSIZE(foo) = 0
-                        ; no address size define for this symbol:
+                        ; no address size defined for this symbol:
                         .out .sprintf("Error, address size unknown for symbol %s", .string(foo))
                 .endif
         .endmacro
--- a/doc/ca65.sgml
+++ b/doc/ca65.sgml
@@ -1278,34 +1278,34 @@ writable.
 Pseudo functions expect their arguments in parenthesis, and they have a result,
 either a string or an expression.
 
+
 <sect1><tt>.ADDRSIZE</tt><label id=".ADDRSIZE"><p>
 
   The <tt/.ADDRSIZE/ function is used to return the interal address size 
   associated with a symbol. This can be helpful in macros when knowing the address 
-  size of symbol can allow for custom instructions.
+  size of symbol can help with custom instructions.
 
   Example:
 
   <tscreen><verb>
         .macro myLDA foo
-
-  	        .if .ADDRSIZE(foo) = 1
-  	          	        ;do custom command based on zeropage addressing:
-  	          	        .byte .A5h, foo
-  	        .elseif .ADDRSIZE(foo) = 2
-  	          	        ;do custom command based on absolute addressing:
-  	          	        .byte .ADh, foo
-  	        .elseif .ADDRSIZE(foo) = 0
-  	          	        ; no address size define for this symbol:
-  	          	        .out .sprinft("Error, address size unknown for symbol %s", .string(foo))
-  	        .endif
+                .if .ADDRSIZE(foo) = 1
+                        ;do custom command based on zeropage addressing:
+                        .byte 0A5h, foo
+                .elseif .ADDRSIZE(foo) = 2
+                        ;do custom command based on absolute addressing:
+                        .byte 0ADh
+                        .word foo
+                .elseif .ADDRSIZE(foo) = 0
+                        ; no address size define for this symbol:
+                        .out .sprintf("Error, address size unknown for symbol %s", .string(foo))
+                .endif
         .endmacro
   </verb></tscreen>
 
   This command is new and must be enabled with the <tt/.FEATURE addrsize/ command.
 
   See: <tt><ref id=".FEATURE" name=".FEATURE"></tt>
-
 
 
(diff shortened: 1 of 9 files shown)


Movax12:
I should add: As it is now, this function will not work until the user enables it with the feature command:
.feature addrsize

src/ca65/expr.c
```diff
@@ -629,6 +629,92 @@ static ExprNode* FuncReferenced (void)
 
 
 
+static ExprNode* FuncAddrSize(void)
```
oliverschmidt: Please adhere to the exsisting coding style.

src/ca65/expr.c
```diff
@@ -629,6 +629,92 @@ static ExprNode* FuncReferenced (void)
...
+
+
+    /* Assume we don't know the size */
+    AddrSize = 0;
+
+    /* Check for a cheap local which needs special handling */
+    if (CurTok.Tok == TOK_LOCAL_IDENT) {
+
+        /* Cheap local symbol */
+        Sym = SymFindLocal(SymLast, &CurTok.SVal, SYM_FIND_EXISTING);
```
oliverschmidt: Please adhere to the exsisting coding style. The same is true for the other function definitions and call sites too.

src/ca65/expr.c
```diff
@@ -629,6 +629,92 @@ static ExprNode* FuncReferenced (void)
...
+        if (ParentScope == 0) {
+            /* No such scope */
+            SB_Done(&ScopeName);
+            SB_Done(&Name);
+            return GenLiteral0();
+        }
+
+        /* If ScopeName is empty, no explicit scope was specified. We have to
+        * search upper scope levels in this case.
+        */
```
oliverschmidt: Please adhere to the exsisting coding style, here multiline comment formatting.

src/ca65/expr.c
```diff
@@ -629,6 +629,92 @@ static ExprNode* FuncReferenced (void)
...
+        }
+
+    }
+
+    /* Check if we have a size */
+    /* if we don't know, return it anyway, zero can mean unknown, or uncomment this code for an error
+    if (AddrSize == 0 ) {
+    Error ("Address size of `%m%p%m%p' is unknown", &ScopeName, &Name);
+    }
+    */
```
oliverschmidt: I have a strong opinion against commented-out code - especially where there's no comment explaining why the code is commented out.

oliverschmidt:
Hi,

First of all thanks for your interest in contributing to cc65@GitHub :-)

I'm in general not exactly keen in adding specific features to the cc65 tools. If however a) the new code doesn't add significant complexity and b) the new feature doesn't (seem to) change the exsisting behaviour when not explicitly using the new feature I'm open to new features.

Anyway I consider it obligatory that the documentation is extended accordingly - which the same pull request. Here that would mean (as far as I understand) a new `.addrsize`section and an addition to the `.feature`section.

Regards,
Oliver

src/ca65/expr.c
greg-king5: Perhaps, it can be a warning (instead of error) message.  Then, that code might not need to be commented out.

mrdudz: On Monday 20 April 2015, 05:31:11 greg-king5 <notifications@github.com> wrote:
> > +        if (Sym) {
> > +            AddrSize = Sym->AddrSize;
> > +        }
> > +        else {
> > +            Error("Unknown symbol or scope: `%m%p%m%p'",
> > +                &ScopeName, &Name);
> > +        }
> > +
> > +    }
> > +
> > +    /* Check if we have a size */
> > +    /* if we don't know, return it anyway, zero can mean unknown, or
> > uncomment this code for an error +    if (AddrSize == 0 ) {
> > +    Error ("Address size of `%m%p%m%p' is unknown", &ScopeName, &Name);
> > +    }
> > +    */
> 
> Perhaps, it can be a warning (instead of error) message.  Then, that code
> might not need to be commented out.

shouldnt it really be an error? what about taking the address from a non-
constant expression?

-- 

Sch�n w�re es, wenn wir wirklich eine Frei-zeitgesellschaft h�tten, dann w�ren 
echtes Bem�hen und Anstrenungen f�r Dinge m�glich, die �ber dem Arbeiten f�rs 
Fressen liegen, insbesondere der Kunst. Mir scheint es eher, als h�tten wir 
eine Arbeits- bzw -losigkeitsgesellschaft; alles, was an Freizeit entsteht 
wird umgemodelt zur Arbeitslosigkeit, und das, was wir an "Freizeitangeboten" 
haben, erscheint mir eher als Symptomkuriererei f�r die, die Arbeit, aber 
keine Zeit haben, und so "das meisste" aus ihrer Zeit machen m�ssen.

Movax12: After some testing and thinking about it I am sure that if a symbol is valid it should have an internal address size which would be defined in Sym->AddrSize

The only exception is when a symbol is used before it is defined:

   .scope 
          lda foo
          .out .sprintf(" address size of foo: %d", .addrsize(foo))         
    .endscope

In this case it would return 0 and I think it should emit a warning.

Movax12:
I have updated the code to match the coding style and have updated .addrsize to warn if size is not known. Is this a good solution? Documentation updates to follow.

doc/ca65.sgml
```diff
@@ -1278,6 +1278,35 @@ writable.
 Pseudo functions expect their arguments in parenthesis, and they have a result,
 either a string or an expression.
```
oliverschmidt: One LF missing.

doc/ca65.sgml
```diff
@@ -1278,6 +1278,35 @@ writable.
...
+  	          	        .out .sprinft("Error, address size unknown for symbol %s", .string(foo))
+  	        .endif
+        .endmacro
+  </verb></tscreen>
+
+  This command is new and must be enabled with the <tt/.FEATURE addrsize/ command.
+
+  See: <tt><ref id=".FEATURE" name=".FEATURE"></tt>
+
+
```
oliverschmidt: One LF too much.

doc/ca65.sgml
```diff
@@ -2648,11 +2683,6 @@ Here's a list of all control commands and a description, what they do:
     overridden. When using this feature, you may also get into trouble if
     later versions of the assembler define new keywords starting with a dot.
 
-  <tag><tt>loose_char_term</tt><label id="loose_char_term"></tag>
-
-    Accept single quotes as well as double quotes as terminators for char
-    constants.
-
```
oliverschmidt: Why did you remove this section? Looking at feature.c it seems that `loose_char_term`is present.

src/ca65/expr.c
```diff
@@ -629,6 +629,87 @@ static ExprNode* FuncReferenced (void)
...
+            AddrSize = Sym->AddrSize;
+        } else {
+            Error ("Unknown symbol or scope: `%m%p%m%p'",
+                &ScopeName, &Name);
+        }
+
+    }
+
+    if (AddrSize == 0) {
+        Warning(1, "Unknown address size: `%m%p%m%p'",
```
oliverschmidt: Space missing.

src/ca65/expr.c
```diff
@@ -629,6 +629,87 @@ static ExprNode* FuncReferenced (void)
...
+            Sym = SymFindAny (ParentScope, &Name);
+        } else {
+            Sym = SymFind (ParentScope, &Name, SYM_FIND_EXISTING);
+        }
+        /* If we found the symbol retrieve the size, otherwise complain */
+        if (Sym) {
+            AddrSize = Sym->AddrSize;
+        } else {
+            Error ("Unknown symbol or scope: `%m%p%m%p'",
+                &ScopeName, &Name);
```
oliverschmidt: Rather no LF here - but if you want then please align `&` with `"`.

src/ca65/expr.c
```diff
@@ -629,6 +629,87 @@ static ExprNode* FuncReferenced (void)
...
+        }
+
+    }
+
+    if (AddrSize == 0) {
+        Warning(1, "Unknown address size: `%m%p%m%p'",
+            &ScopeName, &Name);
```
oliverschmidt: Rather no LF here - but if you want then please align `&` with `1`.

oliverschmidt:
Thanks for acting on all aspects of the feedback - incl. the demand for doc :-)

Apart from my usual style nitpicking (and my question above) this pull request is ready for merge - if no one objects.

doc/ca65.sgml
Movax12: This was accidental.

Movax12:
I'll fix the requested later today.

doc/ca65.sgml
```diff
@@ -1278,6 +1278,35 @@ writable.
...
+
+  <tscreen><verb>
+        .macro myLDA foo
+
+  	        .if .ADDRSIZE(foo) = 1
+  	          	        ;do custom command based on zeropage addressing:
+  	          	        .byte .A5h, foo
+  	        .elseif .ADDRSIZE(foo) = 2
+  	          	        ;do custom command based on absolute addressing:
+  	          	        .byte .ADh, foo
```
greg-king5: `.byte` rejects word-sized operands.

doc/ca65.sgml
```diff
@@ -1278,6 +1278,35 @@ writable.
...
+  	        .elseif .ADDRSIZE(foo) = 0
+  	          	        ; no address size define for this symbol:
+  	          	        .out .sprinft("Error, address size unknown for symbol %s", .string(foo))
```
greg-king5: `.sprinft` is a typo.

doc/ca65.sgml
```diff
@@ -1278,6 +1278,35 @@ writable.
...
+  size of symbol can allow for custom instructions.
+
+  Example:
+
+  <tscreen><verb>
+        .macro myLDA foo
+
+  	        .if .ADDRSIZE(foo) = 1
+  	          	        ;do custom command based on zeropage addressing:
+  	          	        .byte .A5h, foo
```
greg-king5: Now, it's my turn "to pick a nit". ;-)

That looks, to me, like too much indentation.

doc/ca65.sgml
greg-king5: What are `.A5h` and `.ADh` supposed to be -- ca65 rejects them.  (If you want them to be hex numbers, then the period must be replaced with a zero.)

Movax12: Thanks for the notes.

doc/ca65.sgml
Movax12: I matched the other entries, though that blank line after the macro declaration should not be there.

Movax12:
Hopefully I have addressed all concerns. The example code in the documentation was tested and works correctly, so all is good (I hope).

PR #156, opened 2015-04-29
New functionality for ca65: Support for raw string constants.

Movax12:
Hi. I mentioned there would be more. I've tested this quite a bit and it seems to work well. This is  helpful for large assemble time messages, but especially for writing inline scripts. A good example is thefox's NintendulatorDX. This is an NES emulator that can run LUA script for debugging puposes. I can write the LUA script right in the assembly source code, but ca65 requires me to do so line by line. This makes things much nicer.

### DIFF
--- a/libsrc/apple2/crt0.s
+++ b/libsrc/apple2/crt0.s
@@ -46,9 +46,9 @@
         sta     $94
         sty     $95
 
-        ; Call into the Applesoft Block Transfer Utility -- which handles zero-
+        ; Call into Applesoft Block Transfer Up -- which handles zero-
         ; sized blocks well -- to move the content of the LC memory area.
-        jsr     $D396           ; BLTU + 3
+        jsr     $D39A           ; BLTU2
 
         ; Set the source start address.
         lda     #<__ZPSAVE_RUN__
@@ -68,9 +68,9 @@
         sta     $94
         sty     $95
 
-        ; Call into the Applesoft Block Transfer Utility -- which handles moving
+        ; Call into Applesoft Block Transfer Up -- which handles moving
         ; overlapping blocks upwards well -- to move the INIT segment.
-        jsr     $D396           ; BLTU + 3
+        jsr     $D39A           ; BLTU2
 
         ; Delegate all further processing, to keep the STARTUP segment small.
         jsr     init


src/ca65/scanner.c
```diff
@@ -769,6 +769,101 @@ static void ReadIdent (void)
     }
 }
 
+static void ReadRawStringConst (void)
```
greg-king5: Style point:
There should be three blank lines between functions.

src/ca65/scanner.c
```diff
@@ -769,6 +769,101 @@ static void ReadIdent (void)
...
+
+        /* Maybe we are done ? */
+        if (C == ')') {
+            NextChar ();
+
+            /* If we had a opening delimiter, try a match to previous delimiter */
+            if (D->Len)
+            {
+                /* Does the first char match the first char of the stored delimiter ? */
+                if (C == D->Buf[0]) {
```
greg-king5: StrBufs are supposed to be "encapsulated".  We shouldn't access their struct members directly; we always should use the StrBuf functions.

oliverschmidt:
Reading http://en.wikipedia.org/wiki/String_literal two things come into my mind:
- Is "raw" really the right / the best term here?
- What about this section from the Wikipedia article?
"Two problems with multiline string literals are leading and trailing newlines, and indentation. If the initial or final delimiters are on separate lines, there are extra newlines, while if they are not, the delimiter makes the string harder to read, particularly for the first line, which is often indented differently from the rest. Further, the literal must be unindented, as leading whitespace is preserved – this breaks the flow of the code if the literal occurs within indented code."

mrdudz:
I really don't like this at all - isn't .incbin the solution for this problem? (what exactly IS the problem this is trying to solve anyway?)

Movax12:
mrdudz, This is nothing like .incbin. I think I explained it well in the documentation changes, what part doesn't make sense?
olivershmidt, they call it raw here: https://github.com/oliverschmidt (see raw/cooked) It doesn't have to be called raw here, string literal is fine. The problems mentioned are unavoidable I think.

oliverschmidt:
> they call it raw here: http://en.wikipedia.org/wiki/C%2B%2B11 (see raw/cooked) It doesn't have to be called raw here, string literal is fine.

I see.

> The problems mentioned are unavoidable I think.

Okay. Wouldn't it then make sense to document the behaviour with regards to the two mentioned aspects?

Movax12:
I'll work on all suggested changes later tonight. I see strbuf.h I can use is SB_AtUnchecked() and SB_NotEmpty()

src/ca65/scanner.c
greg-king5: (I hope that you see this message _before_ you start changing the code!)

I was only partly correct about StrBuf encapsulation.  They should be "opaque" when we use pointers to them.  But, when we allocate them ourselves, we must access their members directly (there are no functions for those operations).

Pointers to them make sense when we want to pass them between functions; but, you use `D` as a temporary variable, completely internal to `ReadRawStringConst()`.  Therefore, I now think that you should define `D` as a StrBuf (not a pointer to it); then, use `SB_Init()` and `SB_Done()` to start and stop it.  Your code would touch `Len` and `Buf` very directly (`D.Len` and `D.Buf`).

greg-king5: (I hope that you see this message _before_ you start changing the code!)

I was only partly correct about StrBuf encapsulation.  They should be "opaque" when we use pointers to them.  But, when we allocate them ourselves, we must access their members directly (there are no functions for those operations).

Pointers to them make sense when we want to pass them between functions; but, you use `D` as a temporary variable, completely internal to `ReadRawStringConst()`.  Therefore, I now think that you should define `D` as a StrBuf (not a pointer to it); then, use `SB_Init()` and `SB_Done()` to start and stop it.  Your code would touch `Len` and `Buf` very directly (`D.Len` and `D.Buf`).  [And, of course, `SB_AppendChar (&D, C);`.]

Movax12: Okay that makes sense, I'll go that route. If I work on this today it will be late this evening (EST).

mrdudz:
Movax12: i am still not sure why this is needed - to me it seems that the same can be achieved by using the existing string literals and escaping where necessary. the documentation should perhaps mention an example that is not possible otherwise. it should probably also define exactly how (or if) character translation comes into play, and how exactly this deals with host specific issues such as unix vs windows linebreaks. i referred to .incbin, because that is infact "the" way to include "raw" data (whether it is a string or not doesnt matter at all) - and adding a second way to do that seems odd, unless there is a very good reason to do so (which the documentation should mention).

greg-king5:
The example in the ca65 document has "`.out ImportantNotice`"; `.incbin` can't do that.
Does ca65 support character escapes?  (If not, then it would be a very welcome upgrade!)

The syntax is adopted from C++11; I think that its name of the object (raw string literal) should be used (after all, that prefix `R"` comes from the name "Raw string").

Perhaps, you can add line-continuation.  Then, there could be only a slight indentation of the first line of the literal:
```
.DEFINE ImportantNotice R"STRING_END\
(I can put anything here!
It all will become part of the string, until ca65 finds the
keyword "STRING_END".  But, it has to find a ")" first, and a
single '"' at the very end!)STRING_END"
```

Movax12:
I could add line continuation if turned on I suppose. Not sure if like that.

One reason this is helpful: Currently there is no way to escape quotes (") in ca65. This isn't a big deal, but it does solve that small issue. I don't think there is anything else that needs escaping, and if you use the string constant with sprintf(), it is still going to recognise format specifiers (like %d).

I touched on one reason why I want this. This is mostly for what I use ca65 for: for NES development. An individual called thefox has made modifications to a well known NES emulator, Nintendulator, and called it NintendulatorDX. Using some macro code and some changes to the linker config, you can create data that tells the emulator where to call Lua code in your ca65 source. The Lua interface thefox has created allows for detailed debugging and even prototyping code in Lua.

For quick LUA code:

    LDA foo
    STA bar
    ; print A to Lua console:
    ndxLuaExecStr "print( 'In Lua, and A is ' .. NDX.getRegister( 'A' ) )"
    CLC
    ...

His solution for large Lua functions was to use this macro:

    ndxLuaExecFile "foo.lua"

I wanted inline code always, so I was awkwardly building a string from multiple strings:

    STARTLUA
    LUA " (some lua code) "
    LUA " More LUA code.."
    ENDLUA

So you can see how this would make this much nicer to work with, especially since LUA does not care about whitespace and can use both types of quotes.

As for the CR+LF issues, if built on windows the windows standard library will convert CR+LF pairs to '\n'. In Linux, from my understanding, it will read both \r\n if reading a Windows file. To solve that I can match '\r' to space. I want to mess with the literal as little as possible so maybe only check for that character, unless there is a reason to convert something else to space? Currently the assembler returns IsSpace() true when  (C == ' ' || C == '\n' || C == '\r' || C == '\t' || C == '\v' || C == '\f')

Movax12:
Looking closer at the source, I've found that ca65 removes all trailing whitespace on a line by line basis, so I don't have to worry about line endings. BUT this means the raw string functionality, as it is, will not allow trailing whitespace per line. I don't think this is a problem, though I don't really like it, it will still do what I want. Thoughts? 
BTW I also found a bug in the implementation of .linecont, which I can add as a separate pull request or just remove this line: https://github.com/cc65/cc65/blob/master/src/ca65/scanner.c#L1420
Currently, something like this fails:

    .linecont +
    .out \
    "Hello"

While this works:

    .linecont +
    .out \
    X"Hello"

Due to the extra NextChar(); call skipping the first character of the next line.

Anyway, thoughts on no trailing whitespace in a raw string? Maybe raw string is the wrong name if it is left this way.

greg-king5:
On 2015-05-01 7:11 PM, Movax12 wrote:
 >
> BTW, I also found a bug in the implementation of .linecont, which I can
> add as a separate pull request; or, just remove this line:
> https://github.com/cc65/cc65/blob/master/src/ca65/scanner.c#L1420
> Currently, something like this fails:
>
> .linecont +
> .out \
> "Hello"
>
> While this works:
>
> .linecont +
> .out \
> X"Hello"
>
> Due to the extra NextChar(); call skipping the first character of the
> next line.

Several things are wrong with that entire "case" block:
1. The backslash should be skipped only if the next character is a newline.
2. Are MS-DOS and Mac line-endings converted at an earlier stage?  If 
not, then they need to be checked.
3. The line-ending should be skipped only; it shouldn't be replaced with 
a space.

Movax12:
1./2. As I mentioned, all trialing whitespace will be removed, so the next char will be read as '\n' unless something other than whitespace was typed there, and then you will get an error, though a new error could be added there ("expected: newline" ) or something. Or I guess you could NOT honor the backslash if the next character is not a '\n'.
3. Yes, one NextChar(); would also work, and remove the C = ' ';

greg-king5:
On 2015-05-03 4:34 AM, Movax12 wrote:
> 1./2. As I mentioned, all trailing whitespace will be removed; so, the next char. will be read as '\n' unless something other than whitespace was typed there; and then, you will get an error, though a new error could be added there ("expected: newline" ), or something. Or, I guess you could NOT honor the backslash if the next character is not a '\n'.

NOT honoring the backslash is what should be done.  In other words, if the next character isn't a newline, then it shouldn't matter whether the linecont variable is true or false.  The error message should print the code of that backslash.  Something like this:
~~~C
case '\\':
    /* Line continuation? */
    if (LineCont) {
        NextChar ();
        if (C == '\n') {
            NextChar ();
            goto Again;
        }
        else {
            /* Invalid backslash; next time, reread current character */
            char* str = xmalloc (2);

            str[0] = C;
            str[1] = '\0';
            NewInputData (str, 1);
            C = '\\';
        }
    }
    break;
~~~

Movax12:
I don't think there is any need to re-read anything. If Linecont is off, the '\\' will be skipped. If Lincecont is on '\\' is valid, it's just not valid to have anything after it. This does the job fine:

    case '\\':
            /* Line continuation? */
            if (LineCont) {
                NextChar ();
                /* Next char should be a LF, if not, will result in an error later */
                if (C == '\n') {
                    /* Ignore the '\n' */
                    NextChar ();
                    goto Again;
                }
            }
            break;

Or, to make it clear:

    case '\\':
            /* Line continuation? */
            if (LineCont) {
                NextChar ();
                /* Next char should be a LF, if not, will result in an error later */
                if (C == '\n') {
                    /* Ignore the '\n' */
                    NextChar ();
                    goto Again;
                }
                else {
                    /* Make it clear what the problem is: */
                    Error ("Newline expected after line continuation.");
                }
            }
            break;

I think the second one is a bit better, though the first will still Error, it may not be immediately clear what the problem is.

greg-king5:
> I don't think there is any need to re-read anything. If Linecont is off, the '\' will be skipped.

Before the function skips it, that _specific_ character (the backslash) is described as invalid.

> If Linecont is on, '\' is valid; it's just not valid to have anything after it.

I disagree.  A backslash isn't a line-continuation command.  It starts a character-escape; for example: `\a`, `\b`, `\f`, `\n`, `\r`, `\t`, and `\v`.  `\<newline>` is actually a line-continuation escape.

In the same way that a backslash without a `t` has nothing to do with a TAB character,  a backslash without a `<newline>` has nothing to do with a line-continuation.  It's some other character-escape.  If that other escape isn't inside of a character literal (or a comment), then it's wrong.  It's backslash is wrong (it doesn't belong there); and, ca65 should say that it's wrong.

Try this example:

```gas
    .linecont off
    stx    label\+1
```

We want the assembler to tell us that there is a typo.  If `.linecont` is `off`, then the current code will say, correctly, that the `\` is the typo.  But, if `.linecont` is `on`, then the current code will claim that the `+` is the typo.  My suggested patch makes sure that the error message always gives the code number of the actual typo (and, doesn't mention trailing garbage when, obviously, the rest of the line is correct).

Movax12:
I understand your point, but not sure if it is valid to call it an escape sequence, since ca65 will accept (escaped strings):
"stx label \                                                                                                                              \n"
"+1"

Movax12:
A lot of changes, raw strings are truly raw strings. I also modified that case block for linecont, but we can do that separately.

Movax12:
Cleaned up code. Before I was double buffering each line, now only one line buffer. Parsing text is actually a touch faster than before, yet supports raw strings fully. 
I record the index of the start of trailing whitespace per line in `S->V.File.EndLinePos`
Maybe there is a better place to put this? Pending any suggestions, I feel this functionality is done. The only thing to do is add some better documentation.

oliverschmidt:
Just to make sure there's no misunderstanding - from my perspective this pull request waits for
- feedback from Greg
- better documentation

src/ca65/scanner.c
```diff
@@ -112,7 +114,7 @@ struct CharSource {
     CharSource*                 Next;   /* Linked list of char sources */
     token_t                     Tok;    /* Last token */
     int                         C;      /* Last character */
-    const CharSourceFunctions*  Func;   /* Pointer to function table */
+    CharSourceFunctions*  Func;   /* Pointer to function table */
```
oliverschmidt: Please realign comment.

src/ca65/scanner.c
```diff
@@ -426,6 +432,88 @@ static void IFNextChar (CharSource* S)
...
     S->V.File.Pos.Col = SB_GetIndex (&S->V.File.Line);
 
+    /* Return the next character:
+    ** Check for special case - we are at the begining of whitespace
+    ** if so, return EOL
+    */
+    if (SB_GetIndex (&S->V.File.Line) == S->V.File.EndLinePos) {
+        C = '\n';
+        /* Increment the index */
+        SB_SetIndex(&S->V.File.Line, SB_GetIndex (&S->V.File.Line) + 1);
```
oliverschmidt: Missing space.

src/ca65/scanner.c
```diff
@@ -674,11 +769,61 @@ static unsigned DigitVal (unsigned char C)
 static void NextChar (void)
 /* Read the next character from the input file */
 {
+    /* When in raw character mode, call IFNextRawChar
+    ** Otherwise, call IFNextChar
```
oliverschmidt: Missing period.

src/ca65/scanner.c
```diff
@@ -674,11 +769,61 @@ static unsigned DigitVal (unsigned char C)
...
+    */
+     
     Source->Func->NextChar (Source);
 }
 
 
 
+static void EnterRawTextMode (void)
+/* Switch to raw character mode at the current position in the input file.
+**  Calls to EnterRawTextMode and LeaveRawTextMode may be nested.
```
oliverschmidt: Please remove excess space in front of `Calls`.

src/ca65/scanner.c
```diff
@@ -771,6 +916,109 @@ static void ReadIdent (void)
...
+                    /* Scan through the keyword if the char matches */
+                    unsigned K = 1;
+                    NextChar ();
+                    while ((K < D.Len) && (C == D.Buf[K]))  {
+                        NextChar ();
+                        K++;
+                    }
+
+                    /* If was not a match, append anything we skipped to the string literal.
+                    ** If Len is less then the length of the delimiter OR the next char is NOT a quote it wasn't a match
```
oliverschmidt: Missing period.

src/ca65/scanner.c
```diff
@@ -771,6 +916,109 @@ static void ReadIdent (void)
...
+        }
+
+        if (C == EOF) {
+            Error ("End of file in raw string constant");
+            break;
+        }
+        
+        /* Ignore Carriage Return, Vertical Tab and Formfeed */
+        if (C != '\r' && C != '\v' && C != '\f')
+            SB_AppendChar (&CurTok.SVal, C);
```
greg-king5: Coding style point:  the braces are missing.

greg-king5:
I don't like the double line-spacing that is in the listing output now.

doc/ca65.sgml
```diff
@@ -858,6 +858,55 @@ symbol file via the linker. Unnamed labels are not written to the object file,
...
+newlines are not added, this can make the string harder to read, and it may not 
+be possible to match the indentation of the first line with the rest of a multiline 
+string. 
+As well, any leading whitespace to match the formatting of the surrounding 
+code will be preserved in the string.
+
+This command is new and must be enabled with the <tt/.FEATURE raw_strings/ command.
+
+  See: <tt><ref id=".FEATURE" name=".FEATURE"></tt>
+
```
greg-king5: The patch lines that are above here describe string literals and raw string literals; they don't describe (named) string constants.  Those lines should be moved to the "`Input format`" section of the document.

The sentence about `.DEFINE` should be removed because subsection "`Using macros to define labels and constants`" explains that `.DEFINE` is the only way to create string constants.  (In that subsection, you should expand the name "string constants" to "string constants and raw-string constants".)

Movax12:
I'll look at the listing problem today.. might have to revert to double buffering the strings - it's not really much slower.

Movax12:
greg-king5 - Not sure if this is the documentation change you think is needed. I see I should change 'constant' to 'literal' in some cases, but I am going to wait for your comments before I start adding more changes.

doc/ca65.sgml
```diff
@@ -655,7 +655,54 @@ problem in most cases.
 
 To force a specific order of evaluation, parentheses may be used, as usual.
 
+<sect>String Constants and Raw-String Constants<label id="String_Constants"><p>
```
greg-king5: It should be a `<sect1>`.  It should be placed between
`<sect1>Number format`
and
`<sect1>Conditional assembly`.

In other words, it's a companion of the "Number format" subsection.

oliverschmidt:
Hi,

Maybe I'm wrong but the state of this pull request seems to be that @greg-king5 made a suggestion regarding the documentation "In other words, it's a companion of the "Number format" subsection." and that @Movax12 hasn't reacted so far on that.

In case my assessment is correct it seems to me that this pull request is pretty close to be merged so I encourage @Movax12 to take care of this final (?) "issue". My personal experience is that pull requests staying open too long tend to be closed unmerged as the participants just loose interest in it....

Regards,
Oliver

oliverschmidt:
Seems that my assessment above was correct :-(

Movax12:
I didn't close this ?

oliverschmidt:
GitHub says you did so 7 hours ago...

Movax12:
That's really odd. I pushed to the definedmacro branch, other than that, I didn't use github. I'll rebase and reopen in any case.

Movax12:
Seems a misunderstanding (I'm still new to git) I see when I pushed definedmacro, it pushed to this branch as well, which somehow closed it.

greg-king5:
I suspect that GitHub closed this Pull Request because that mistaken push removed _all_ of this PR's **diff** commits.

PR #157, opened 2015-05-06
Rockwell C39/L39/R6750/R6760 and .ASCIIZX

coyxc:
Added support for Rockwell C39/L39/R6750/R6760 processor and escapes in
string literals.


asminc/cpu.mac
```diff
@@ -7,6 +7,9 @@ CPU_ISET_65C02     = $0010
 CPU_ISET_65816     = $0020
 CPU_ISET_SWEET16   = $0040
 CPU_ISET_HUC6280   = $0080
+CPU_ISET_M740      = $0100
+CPU_ISET_C39_EMUL  = $0200
+CPU_ISET_C39_NATIVE = $0400
 
 ; CPU capabilities
 CPU_NONE           = CPU_ISET_NONE
```
greg-king5: You should add a `CPU_C39` name that combines the EMUL and NATIVE bits.  And maybe, it should be used in many of the library source files, instead of `CPU_ISET_C39_EMUL`.

greg-king5:
The cc65 project has a set of coding-style rules.  They haven't been typed and published in the Wiki, yet.  So, we hope that coders will try to make their new code look like the style that they see in the current source code files.  Some of the rules are:
(_Note:_ The word "must" indicates a requirement.  The word "should" indicates a recomendation.)

**In all files:**

* TAB characters must be expanded to spaces.
* All text files must end with new-line characters.  Don't leave the last line "dangling".
* You should avoid typing non-ASCII characters.

**In C source files:**

* You must use ANSI C comments (`/* */`); you must not use C++ comments (`//`).
* The normal indentation width should be four spaces.
* When a function's argument list wraps around to a next line, you should indent that next line by either the normal width or enough spaces to align it with the arguments on the previous line.
* When you add functions to an existing file, you should separate them by the same number of blank lines that separate the functions that already are in that file.

**In Assembly source files:**

* Op-code mnemonics must have lower-case letters.  The names of instruction macroes may have upper-case letters.
* Hexadecimal number constants should be used except where decimal or binary numbers make much more sense in that constant's context.
* Hexadecimal letters should be upper-case.
* When you set two registers or two memory locations to a 16-bit zero, you should use the expressions `#<$0000` and `#>$0000` (they make it obvious where you are putting the lower and upper bytes).
* If a function is declared to return a char-sized value, it actually must return an integer-sized value.  (When cc65 promotes a returned value, it sometimes assumes that the value already is an integer.)
* Functions, that are intended for a platform's system library, should be optimized as much as possible.
* Sometimes, there must be a trade-off between size and speed.  If you think that a library function won't be used often, then you should make it small.  Otherwise, you should make it fast.
* Comments that are put on the right side of instructions must be aligned (start in the same character columns).
* Assembly source fields (label, operation, operand, comment) should start after character columns that are multiples of eight (such as 1, 9, 17, 33, and 41).

oliverschmidt:
Hi Greg,

> The cc65 project has a set of coding-style rules. They haven't been typed
> and published in the Wiki, yet. So, we hope that coders will try to make
> their new code look like the style that they see in the current source code
> files. Some of the rules are:
> (*Note:* The word "must" indicates a requirement. The word "should"
> indicates a recomendation.)
>
> *In all files:*
>
>    - TAB characters must be expanded to spaces.
>    - All text files must end with new-line characters. Don't leave the
>    last line "dangling".
>    - You should avoid typing non-ASCII characters.
>
> *In C source files:*
>
>    - You must use ANSI C comments (/* */); you must not use C++ comments (
>    //).
>    - The normal indentation width should be four spaces.
>    - When a function's argument list wraps around to a next line, you
>    should indent that next line by either the normal width or enough spaces to
>    align it with the arguments on the previous line.
>    - When you add functions to an existing file, you should separate them
>    by the same number of blank lines that separate the functions that already
>    are in that file.
>
> *In Assembly source files:*
>
>    - Op-code mnemonics must have lower-case letters. The names of
>    instruction macroes may have upper-case letters.
>    - Hexadecimal number constants should be used except where decimal or
>    binary numbers make much more sense in that constant's context.
>    - Hexadecimal letters should be upper-case.
>    - When you set two registers or two memory locations to a 16-bit zero,
>    you should use the expressions #<$0000 and #>$0000 (they make it
>    obvious where you are putting the lower and upper bytes).
>    - If a function is declared to return a char-sized value, it actually
>    must return an integer-sized value. (When cc65 promotes a returned value,
>    it sometimes assumes that the value already is an integer.)
>    - Functions, that are intended for a platform's system library, should
>    be optimized as much as possible.
>    - Sometimes, there must be a trade-off between size and speed. If you
>    think that a library function won't be used often, then you should make it
>    small. Otherwise, you should make it fast.
>    - Comments that are put on the right side of instructions must be
>    aligned (start in the same character columns).
>    - Assembly source fields (label, operation, operand, comment) should
>    start after character columns that are multiples of eight (such as 1, 9,
>    17, 33, and 41).
>
> Wow. Nice writeup. It would really be a pity to have this buried on "some"
pull request dicussion. Can I persuade you to put it in the Wiki ?

Thanks in advance, Oliver

greg-king5:
> Nice writeup. It would really be a pity to have this buried on "some" pull request dicussion. Can I persuade you to put it in the Wiki?

That's my plan (see the last word on the first line of that message).  For a while, I have been adding rules and guidelines, bit by bit, to a couple of my files (one about code sources and one about documentation).  I haven't published them -- yet -- because I haven't settled my mind about their format.

src/ca65/ea65.c
```diff
@@ -106,7 +106,10 @@ void GetEA (EffAddr* A)
         if (CurTok.Tok == TOK_COMMA) {
             /* [dir],y */
             NextTok ();
-            Consume (TOK_Y, "`Y' expected");
+            if (GetCPU() == CPU_C39_NATIVE)
+              Consume (TOK_X, "`X' expected");
+            else
+              Consume (TOK_Y, "`Y' expected");
             A->AddrModeSet = AM65_DIR_IND_LONG_Y;
```
greg-king5: The bracket addressing modes (`AM65_*LONG*`) are used in 65816 code.  They aren't used in the code for any of the eight-bit 65xx processors.  So, you don't need to patch that section of GetEA().

src/ca65/instr.c
```diff
@@ -373,6 +381,153 @@ static const struct {
...
+static const struct {
+  unsigned Count;
+  InsDesc  Ins[135]; // FIX WHEN ADDING INSN!
+} InsTabC39 = {
+    sizeof (InsTabC39.Ins) / sizeof (InsTabC39.Ins[0]),
+    {
+      // C39: (ind)   = AM65_DIR_IND   = 0x0400 allowed
+      // C39: (ind,x) = AM65_DIR_X_IND = 0x8000 not allowed
+      { "ADC",  0x080266C, 0x60, 0, C39PutArith }, 
+      { "ADX",  0x0800024, 0x80, 0, C39PutAdd }, // C39: renamed ADD to ADX because it conflicts with add macro
```
greg-king5: I think that renaming the instruction `ADD` is a good idea; but, I'm not happy with the spelling `ADX`.  It looks like an instruction to add to the X index register.  I suggest the spelling `ADA` (ADd to Accumulator).

groessler: I think renaming ADD is not a good idea if it's called ADD in the official C39 documentation. The ADD macro should be renamed.

coyxc: The Rockwell ADD insn has only 3 AMs: IMD, ZP and ZP,X so renaming the
add-macro will not work.

ADD Add Memory to Accumulator without Carry

Operation: A + M → A, C N V – B D I Z C

C C – – – – C C

ADD adds the value in memory to the value in the accumulator and
stores the result in the accumulator.  The C flag is set when the sum
exceeds hexadecimal 255 (binary mode) or decimal 99 (decimal mode),
otherwise C is reset.  In the binary mode, the V flag is set when the
sign (bit 7) changes due to the result exceeding +127 or –128,
otherwise V is reset. In the decimal mode, V is reset. The N flag is
set if bit 7 of the result is a 1, otherwise N is reset. The Z flag is
set if the result is zero, otherwise Z is reset.

Addressing Assembly Language Op No. No.
Mode Form Code Bytes Cycles Ref.

Immediate ADD #Oper 89 2 2c C2.1
Zero Page ADD Oper 64 2 3c C3.1
Zero Page,X ADD Oper,X 74 2 4c C4.1

c. Add 1 if in decimal mode.



On 08.05.2015 05:25, Christian Groessler wrote:
>
> In src/ca65/instr.c 
> <https://github.com/cc65/cc65/pull/157#discussion_r29914926>:
>
> > @@ -373,6 +381,153 @@ static const struct {
> >      }
> >  };
> >
> > +/* Instruction table for the Rockwell C39 */
> > +static const struct {
> > +  unsigned Count;
> > +  InsDesc  Ins[135]; // FIX WHEN ADDING INSN!
> > +} InsTabC39 = {
> > +    sizeof (InsTabC39.Ins) / sizeof (InsTabC39.Ins[0]),
> > +    {
> > +      // C39: (ind)   = AM65_DIR_IND   = 0x0400 allowed
> > +      // C39: (ind,x) = AM65_DIR_X_IND = 0x8000 not allowed
> > +      { "ADC",  0x080266C, 0x60, 0, C39PutArith },
> > +      { "ADX",  0x0800024, 0x80, 0, C39PutAdd }, // C39: renamed ADD to ADX because it conflicts with add macro
>
> I think renaming ADD is not a good idea if it's called ADD in the 
> official C39 documentation. The ADD macro should be renamed.
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/cc65/cc65/pull/157/files#r29914926>.
>

-- 
Mit freundlichen Grüssen

Rainer Keuchel

groessler: I thought about renaming the 'add' macro and replacing all occurrences of it in the runtime lib. But it*s probably used outside the runtime lib in user programs, so it cannot be renamed.

So using a different name for the "ADD" instruction is probably the only way to go.

src/ca65/main.c
```diff
@@ -307,6 +307,14 @@ static void SetSys (const char* Sys)
             NewSymbol ("__OSIC1P__", 1);
             break;
 
+        case TGT_CREATIX_EMUL:
+            NewSymbol ("__CREATIX_EMUL__", 1);
+            break;
+            
+        case TGT_CREATIX_NATIVE:
```
groessler: I don't know if the emulation setting (EMUL vs. NATIVE) justifies a different target. After all, the target machine is always "creatix". EMUL or NATIVE just changes the behavior of the assembler.

coyxc: I think it's difficult to build native and emul libs in one subdir using the current makefile in libsrc. Using two targets is the easiest way. What is your suggestion to solve this problem?

src/ca65/instr.c
```diff
@@ -373,6 +381,153 @@ static const struct {
...
+      { "DEX",  0x0000001, 0xca, 0, PutAll },
+      { "DEY",  0x0000001, 0x88, 0, PutAll },
+      { "EOR",  0x080266C, 0x40, 0, C39PutArith }, // C39: (ind,x) not allowed
+      { "EXC",  0x0000020, 0xd3, 0, C39PutExchange }, // C39 
+      { "INC",  0x000006f, 0x00, 4, PutAll },
+      { "INI",  0x0000001, 0xbb, 0, PutAll }, // C39
+      { "INX",  0x0000001, 0xe8, 0, PutAll },
+      { "INY",  0x0000001, 0xc8, 0, PutAll },
+      { "JMP",  0x0010808, 0x4c, 6, PutAll }, // C39: (abs,x) allowed..
+      { "JPI",  0x0000001, 0x0c, 0, PutAll }, // C39
```
greg-king5: The description of `JPI` is wrong in the "MCU Technical Reference Manual" that you cited in your original mailing-list post.  `JPI` is a three-byte instruction; therefore, it must have an operand.  Its addressing mode is (Indirect); therefore, its correct syntax must be:
```gas
    jpi (indirect)
```

So, that entry in the instruction table should be:
```c
    { "JPI",  0x0000800, 0x0c, 7, PutAll },       /* C39 */
```

coyxc: In the L28 datasheet the JPI instruction is a one byte insn without 
operand, so I used

jpi
.word addr

until now..




On 08.05.2015 10:27, greg-king5 wrote:
>
> In src/ca65/instr.c 
> <https://github.com/cc65/cc65/pull/157#discussion_r29923890>:
>
> > +      { "CLV",  0x0000001, 0xb8, 0, PutAll },
> > +      { "CLW",  0x0000001, 0x52, 0, PutAll }, // C39
> > +      { "CMP",  0x080266C, 0xc0, 0, C39PutArith }, // C39: (ind,x) not allowed
> > +      { "CPX",  0x080000C, 0xe0, 1, PutAll },
> > +      { "CPY",  0x080000C, 0xc0, 1, PutAll },
> > +      { "DEC",  0x000006F, 0x00, 3, PutAll },
> > +      { "DEX",  0x0000001, 0xca, 0, PutAll },
> > +      { "DEY",  0x0000001, 0x88, 0, PutAll },
> > +      { "EOR",  0x080266C, 0x40, 0, C39PutArith }, // C39: (ind,x) not allowed
> > +      { "EXC",  0x0000020, 0xd3, 0, C39PutExchange }, // C39
> > +      { "INC",  0x000006f, 0x00, 4, PutAll },
> > +      { "INI",  0x0000001, 0xbb, 0, PutAll }, // C39
> > +      { "INX",  0x0000001, 0xe8, 0, PutAll },
> > +      { "INY",  0x0000001, 0xc8, 0, PutAll },
> > +      { "JMP",  0x0010808, 0x4c, 6, PutAll }, // C39: (abs,x) allowed..
> > +      { "JPI",  0x0000001, 0x0c, 0, PutAll }, // C39
>
> The description of |JPI| is wrong in the "MCU Technical Reference 
> Manual" that you cited in your original mailing-list post. |JPI| is a 
> three-byte instruction; therefore, it must have an operand. Its 
> addressing mode is (Indirect); therefore, its correct syntax must be:
>
>      jpi (indirect)
>
> So, that entry in the instruction table should be:
>
>      {"JPI",  0x0000800, 0x0c,7, PutAll },/* C39 */
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/cc65/cc65/pull/157/files#r29923890>.
>

-- 
Mit freundlichen Grüssen

Rainer Keuchel

src/ca65/instr.c
```diff
@@ -1538,3 +1707,270 @@ void HandleInstruction (unsigned Index)
...
+    if (CurTok.Tok != TOK_HASH)
+      {
+        ErrorSkip ("BARBAS: Invalid immed, no # found");
+        return;
+      }
+    
+    NextTok ();
+    
+    /* MUST USE .DEFINE, NOT = FOR CONSTANTS! */
+    mask = ConstExpression();
```
greg-king5: I definitely don't like that!  We must be able to use names that were assigned by `=`.  But even more important, we must be able to use names that were defined by the `.enum` directive.

mrdudz: On Friday 08 May 2015, 20:13:55 greg-king5 <notifications@github.com> wrote:
> > +    ExprNode* addr;
> > +    long mask;
> > +
> > +    addr = Expression();
> > +    ConsumeComma ();
> > +
> > +    if (CurTok.Tok != TOK_HASH)
> > +      {
> > +        ErrorSkip ("BARBAS: Invalid immed, no # found");
> > +        return;
> > +      }
> > +
> > +    NextTok ();
> > +
> > +    /* MUST USE .DEFINE, NOT = FOR CONSTANTS! */
> > +    mask = ConstExpression();
> 
> I definitely don't like that!  We must be able to use names that were
> assigned by `=`.  But even more important, we must be able to use names
> that were defined by the `.enum` directive.

indeed :)

src/ca65/instr.c
```diff
@@ -1538,3 +1707,270 @@ void HandleInstruction (unsigned Index)
...
+              Emit1(0x5B, A.Expr); /* ORA (zp),y -> JSB5 .db expr */
+              break;
+            case 0x40: 
+              Emit1(0x6B, A.Expr); /* EOR (zp),y -> JSB6 .db expr */
+              break;
+            case 0xC0: 
+              Emit1(0x7B, A.Expr); /* CMP (zp),y -> JSB7 .db expr */
+              break;
+            default:
+              ErrorSkip ("ARITH: Invalid arith base code 0x%x", Ins->BaseCode);
```
greg-king5: That seems like an internal ca65 error message.  And, there are other error messages like that one; I think that they would be seen only if there is a bug in ca65.  Therefore, they should be handled by calls to `Internal()`.

src/ca65/main.c
groessler: Why do you need emul and native libs? You want to have a runtime lib for "creatix", correct? If it uses emul features (and I guess it does, since that was your reason to add it), build it with "emul".

greg-king5: The emulation layer is needed by programs that are written in C (the compiler doesn't generate the C39 addressing modes).  Assembly-only programs might not need that layer; therefore, it wouldn't need to be installed in the PROM.

But, you do have a valid point.  Maybe, that issue could be handled by an ld65 command-line option.  The `-u` option could be used to force the layer's code to be linked into a ROM program.  (Or, two configuration files could be created; that choice was done for some other targets, such as the C64.)  Only one library would need to be built.

src/ca65/main.c
```diff
@@ -1040,7 +1048,14 @@ int main (int argc, char* argv [])
...
         if (Target != TGT_UNKNOWN) {
-            SetCPU (GetTargetProperties (Target)->DefaultCPU);
+            int cpu;
+
+            cpu = GetTargetProperties (Target)->DefaultCPU;
+
+            if (cpu == CPU_UNKNOWN) {
+              Error ("CPU of target is unknown");
+            }
+            SetCPU (cpu);
```
greg-king5: That test isn't needed.  None of the `DefaultCPU` entries is allowed to have `CPU_UNKNOWN` -- even the `none` target has a valid CPU.

src/ca65/instr.c
coyxc: Greg, thanks for your hint, my IDA plugin shows it this way also.
It seems I have not checked this insn correctly..

ROM:89BD                 JPI     (0x89CA)
ROM:89C0                 JPI     (0x89C8)
ROM:89C3                 JPI     (0x89C6)

src/ca65/main.c
coyxc: I had an error in this place at some time and had to check the code what 
exactly went wrong.
Some existing error messages give no idea what the problem could be..

On 09.05.2015 21:05, greg-king5 wrote:
>
> In src/ca65/main.c 
> <https://github.com/cc65/cc65/pull/157#discussion_r29995407>:
>
> > @@ -1040,7 +1048,14 @@ int main (int argc, char* argv [])
> >      /* If no CPU given, use the default CPU for the target */
> >      if (GetCPU () == CPU_UNKNOWN) {
> >          if (Target != TGT_UNKNOWN) {
> > -            SetCPU (GetTargetProperties (Target)->DefaultCPU);
> > +            int cpu;
> > +
> > +            cpu = GetTargetProperties (Target)->DefaultCPU;
> > +
> > +            if (cpu == CPU_UNKNOWN) {
> > +              Error ("CPU of target is unknown");
> > +            }
> > +            SetCPU (cpu);
>
> That test isn't needed. None of the |DefaultCPU| entries is allowed to 
> have |CPU_UNKNOWN| -- even the |none| target has a valid CPU.
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/cc65/cc65/pull/157/files#r29995407>.
>

-- 
Mit freundlichen Grüssen

Rainer Keuchel

src/ca65/instr.c
coyxc: How would you handle this case?


On 09.05.2015 05:13, greg-king5 wrote:
>
> In src/ca65/instr.c 
> <https://github.com/cc65/cc65/pull/157#discussion_r29989025>:
>
> > +    ExprNode* addr;
> > +    long mask;
> > +
> > +    addr = Expression();
> > +    ConsumeComma ();
> > +
> > +    if (CurTok.Tok != TOK_HASH)
> > +      {
> > +        ErrorSkip ("BARBAS: Invalid immed, no # found");
> > +        return;
> > +      }
> > +
> > +    NextTok ();
> > +
> > +    /* MUST USE .DEFINE, NOT = FOR CONSTANTS! */
> > +    mask = ConstExpression();
>
> I definitely don't like that! We must be able to use names that were 
> assigned by |=|. But even more important, we must be able to use names 
> that were defined by the |.enum| directive.
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/cc65/cc65/pull/157/files#r29989025>.
>

-- 
Mit freundlichen Grüssen

Rainer Keuchel

src/ca65/main.c
coyxc: Building with emul does not support (ind),x, which is needed for native 
code.

The emul option is needed for code the compiler generates and the existing
libsrc asm files that use (ind),y.

The emul build must waste lots of cylces with emulating every (ind),y
with code that uses (ind),x.

I think the best option would be to switch between emul and native
with an assembler directive.


On 09.05.2015 10:53, Christian Groessler wrote:
>
> In src/ca65/main.c 
> <https://github.com/cc65/cc65/pull/157#discussion_r29991175>:
>
> > @@ -307,6 +307,14 @@ static void SetSys (const char* Sys)
> >              NewSymbol ("__OSIC1P__", 1);
> >              break;
> >
> > +        case TGT_CREATIX_EMUL:
> > +            NewSymbol ("__CREATIX_EMUL__", 1);
> > +            break;
> > +
> > +        case TGT_CREATIX_NATIVE:
>
> Why do you need emul and native libs? You want to have a runtime lib 
> for "creatix", correct? If it uses emul features (and I guess it does, 
> since that was your reason to add it), build it with "emul".
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/cc65/cc65/pull/157/files#r29991175>.
>

-- 
Mit freundlichen Grüssen

Rainer Keuchel

groessler: On 05/09/15 20:38, greg-king5 wrote:
>
> The emulation layer is needed by programs that are written in C (the 
> compiler doesn't generate the C39 addressing modes). Assembly-only 
> programs might not need that layer; therefore, it wouldn't need to be 
> installed in the PROM.
>


Yes, but do assembly programs need the C runtime lib?

groessler: On 05/09/15 22:56, coyxc wrote:
>
> I think the best option would be to switch between emul and native 
> with an assembler directive.
>

Agreed.

src/ca65/pseudo.c
```diff
@@ -421,7 +421,133 @@ static void DoAlign (void)
...
+  size_t byte_len;
+  char *s;
+
+  len = SB_GetLen(Data);
+  
+  if((bufin = (char *) malloc(len+1)) == NULL)
+    {
+      perror("malloc");
+      exit(1);
+    }
```
greg-king5: Use `xmalloc ()`; it does error-checking.

src/ca65/pseudo.c
```diff
@@ -421,7 +421,133 @@ static void DoAlign (void)
...
+          bufout[byte_len++] = *s++;
+        }
+    }
+
+  bufout[byte_len] = 0;
+
+  Data->Len = byte_len;
+
+ done:
+  free(bufin);
```
greg-king5: Use `xfree ()`.

src/ca65/pseudo.c
```diff
@@ -421,7 +421,133 @@ static void DoAlign (void)
...
+
+  len = SB_GetLen(Data);
+  
+  if((bufin = (char *) malloc(len+1)) == NULL)
+    {
+      perror("malloc");
+      exit(1);
+    }
+  
+  strcpy(bufin, SB_GetBuf(Data));
```
greg-king5: Changed advice:  Use `xstrdup (SB_GetConstBuf (Data))`.  It will do the `malloc()` and the `strcpy()`.

oliverschmidt:
>> Nice writeup. It would really be a pity to have this buried on "some" pull request dicussion. Can I persuade you to put it in the Wiki?

>That's my plan (see the last word on the first line of that message). For a while, I have been adding rules and guidelines, bit by bit, to a couple of my files (one about code sources and one about documentation). I haven't published them -- yet -- because I haven't settled my mind about their format.

Great, looking forward to it :-)

oliverschmidt:
I had a hard time finding authoritative sources regarding the genealogy of those Rockwell CPUs/MCUs. But there are several hints that the R65C19 may be considered as the one introducing the instruction set discussed here. An interesting one: https://github.com/agatti/hopper-plugins#6502-cpu-plugin

From that perspective I strongly suggest to rename the stuff discussed here to `R65C19`.

oliverschmidt:
It's generally no good idea to mix several topics into one pull request. If the pull request is a very small one it might be acceptable but given the pure amount of changes we see here I want to have the .ASCIIZX stuff separated out.

libsrc/Makefile
```diff
@@ -14,14 +14,16 @@ CBMS = c128   \
...
-TARGETS = apple2    \
+TARGETS = \
+          $(CBMS)   \
+          $(GEOS)   \
+          apple2    \
           apple2enh \
           atari     \
-          atarixl   \
           atari5200 \
+          atarixl   \
```
oliverschmidt: This pull request isn't related to Atari stuff so please leave Atari stuff alone. I strongly disgree to mix "general cleanup" operations with significant changes. Apart from that the order of Atari items was carefully chosen.

libsrc/.gitignore
```diff
@@ -0,0 +1 @@
+*~
```
oliverschmidt: I don't see ignoring Git all kinds of temporary files of all possible editors like '.bak' etc. If you don't like to see them make your editor not create them.

libsrc/Makefile
coyxc: Greg told me that this list must be in ascending order..


On 10.05.2015 13:51, Oliver Schmidt wrote:
>
> In libsrc/Makefile 
> <https://github.com/cc65/cc65/pull/157#discussion_r30001365>:
>
> >            atari5200 \
> > +          atarixl   \
>
> This pull request isn't related to Atari stuff so please leave Atari 
> stuff alone. I strongly disgree to mix "general cleanup" operations 
> with significant changes. Apart from that the order of Atari items was 
> carefully chosen.
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/cc65/cc65/pull/157/files#r30001365>.
>

-- 
Mit freundlichen Grüssen

Rainer Keuchel

libsrc/creatix-emul/c39.h
```diff
@@ -0,0 +1,84 @@
+#ifndef C39_H
```
oliverschmidt: Comment regarding the `c39...` files in general: I don't see why a target library should contain files named like the CPU it uses. Please elaborate what it means.

libsrc/creatix-emul/zeropage.s
```diff
@@ -0,0 +1,41 @@
...
+tmp1:           .res    1
+tmp2:           .res    1
+tmp3:           .res    1
+tmp4:           .res    1
+regbank:        .res    6       ; 6 byte register bank
+; C39   
+tx:             .res    1
+ty:             .res    1
+indptr1:        .res    2
+indptr2:        .res    2
```
oliverschmidt: Please explain why the four vars above can't go into the EXTZP segment.

oliverschmidt:
Sorry if I'm totally missing the point but...

I don't understand why we need two CPU types. I understand that the CPUs in question have additional  instructions. And I understand that addressing mode `(zp),y` doesn't exsist. Given that the latter is used both in the runtime library and generated by the compiler it needs to be emulated. But how does that lead to two CPU types? Please elaborate.

oliverschmidt:
Closed as requested by the author.

PR #158, opened 2015-05-11
Added .DEFINEDMACRO psuedo function

Movax12:
This is a separate branch from the other work I was doing and can be evaluated on it's own.

This allows ca65 to check if an identifier is a macro. This is actually needed by me to be able to call other macros cleanly from within macros. (I need to evaluate if a parameter is a macro, instruction or variable.)

PR #160, opened 2015-05-19
.DEFINEDINSTR

Movax12:
This is very close to the .DEFINEDMACRO function, but for instructions. It does need to check UbiquitousIdents however. I removed the .FEATURE requirement since it was suggested we don't need to for .DEFINEDMACRO.


src/ca65/expr.c
```diff
@@ -417,6 +418,34 @@ static ExprNode* FuncDefined (void)
 
 
 
+static ExprNode* FuncDefinedInstr (void)
+/* Handle the .DEFINEDINSTR builtin function */
+{
+    int Instr = 0;
```
greg-king5: `FindInstruction()` returns `-1` if its argument isn't an instruction.

src/ca65/expr.c
```diff
@@ -417,6 +418,34 @@ static ExprNode* FuncDefined (void)
...
+            /* Macros CAN be instructions, so check for them first */
+            if (FindMacro (&CurTok.SVal) == 0) {
+                Instr = FindInstruction (&CurTok.SVal);
+            }
+        } else {
+            /* Macros and symbols may NOT use the names of instructions, so just check for the instruction */
+            Instr = FindInstruction (&CurTok.SVal);
+        }
+    } else {
+        Error ("Idenitifier expected.");
```
greg-king5: The usual typo. :wink:

greg-king5:
We don't define instructions; they're built in.  I think that this function should be named something like `.isinst()` (for "is instruction").

I noticed that the description of `.DEF, .DEFINED` was put in the wrong section.  It isn't a control command; it's a function.  It should be moved to the `Pseudo functions` section (and, that section should be renamed just `Functions`) -- but, do it in a separate Pull Request.

I mentioned `.defined()` because `.ISINST` (and, `.DEFINEDMACRO`) also need to be moved.

Movax12:
How about `.isinstr()` ? I'd like the 'r' to be included.

greg-king5:
"instr" makes me think of "in string" (some programming languages have a string function with that name).

src/ca65/expr.c
MonteCarlos: Is

.definedopstr()

a possible choice? For "operation code"

MC

Am 19.05.2015 um 11:49 schrieb greg-king5:

greg-king5:
`.isopstr()` looks like a viable suggestion to me.

Movax12:
I can't help but see string when I see `str` now. I think I like `.ismnem()` and `.ismnemonic().` better, adn I think we should include both.

greg-king5:
Implementing both spellings is a good idea.

Movax12:
Changes to documentation to come.

oliverschmidt:
Hi,

Maybe the comment I just made in https://github.com/cc65/cc65/pull/158 is true here too...

Regards,
Oliver

Movax12:
Thanks for checking on these requests. I am still working on them - I have just moved and have been busy, but I plan to get back to work on this soon. Incidentally, I think this specific request is complete from my perspective.

oliverschmidt:
Hi,

> Thanks for checking on these requests. I am still working on them - I have
> just moved and have been busy, but I plan to get back to work on this soon.
>
> Good to know. I just wanted to make sure that not "everybody" is waiting
for me without me noticing it...

> Incidentally, I think this specific request is complete from my
> perspective.
>
I see. Let's wait if Greg has something to add.

Regards,
Oliver

>

greg-king5:
I agree with "Movax12".  In fact, as far as I can tell, _both_ PR #158 and this one are finished.  Both of them are ready for Oliver's action.

oliverschmidt:
Hi Greg,

> In fact, as far as I can tell, both PR #158 and this one are finished.

Thanks for your fedback :-)

> Both of them are ready for Oliver's action.

I will act accordingly.

Regards,
Oliver

PR #162, opened 2015-06-25
Test-suite work-arounds for the new compiler

greg-king5:
I forgot to run the test-suite after I added some more error traps to cc65.  So, I didn't notice that a few of its programs couldn't be compiled.  This PR works around the "Kernighan & Ritchie" incompatibility, in the usual way.


oliverschmidt:
As I'm right now trying to finally activate the test-suite this pull request is pretty welcome :-)

PR #163, opened 2015-06-29
added 3 more tests

mrdudz:
this adds 3 more tests to the regression testsuite, including one for the unfixed "equality problem"

### DIFF
--- a/test/ref/Makefile
+++ b/test/ref/Makefile
@@ -45,10 +45,6 @@ $(WORKDIR)/%.ref: %.c
 #
 $(WORKDIR)/init%prg: CC65FLAGS += -Wc --all-cdecl
 $(WORKDIR)/switch.%rg: CC65FLAGS += -Wc --all-cdecl
-
-# Also, yacc.c does some things that fail when stack operations are optimized.
-# Therefore, don't optimize them.
-#
 $(WORKDIR)/yacc.%rg: CC65FLAGS += -Wc --all-cdecl
 $(WORKDIR)/yaccdbg%prg: CC65FLAGS += -Wc --all-cdecl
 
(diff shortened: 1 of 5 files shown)


test/ref/Makefile
```diff
@@ -49,8 +49,8 @@ $(WORKDIR)/switch.%rg: CC65FLAGS += -Wc --all-cdecl
 # Also, yacc.c does some things that fail when stack operations are optimized.
 # Therefore, don't optimize them.
 #
```
greg-king5: The comment, too, should disappear.

mrdudz:
indeed :)

oliverschmidt:
Thanks :-)

MonteCarlos:
Are these functions

void __fastcall__ _heapadd (void* mem, size_t size);
/* Add a block to the heap */

size_t __fastcall__ _heapblocksize (const void* block);
/* Return the size of an allocated block */

size_t _heapmemavail (void);
/* Return the total free heap space */

size_t _heapmaxavail (void);
/* Return the size of the largest free block on the heap */

CC65 inventions or do they have another origin like a specific known
c library? If i try a google search using f.e. _heapmaxavail, i get only
CC65 related links which supports the first theory in my opinion.


Yours Monte Carlos

polluks:
In Turbo Pascal "MemAvail" and "MaxAvail" are known
http://putka.upm.si/langref/turboPascal/049E.html
http://putka.upm.si/langref/turboPascal/049B.html

MonteCarlos:
The idea of those functions is quite on the hand. Maybe Uz borrowed
the part of the functions names from Pascal?
However, using these functions should break any compatibility with other
c compilers.

Am 02.07.2015 um 13:33 schrieb polluks:

PR #165, opened 2015-07-01
Fix pointer-to-array indirection bug.

greg-king5:
This fixes a bug that was [reported by Monte Carlos](https://sourceforge.net/p/cc65/mailman/message/34227065/) on the mailing list; then, analyzed by Daniel Serpell.

The fix is very simple -- which makes me a little nervous. ;-)
I hope that there are no regressions.  It passes the test-suite.  The suite already has [a simple test for array-pointer indirection](https://github.com/cc65/cc65/blob/master/test/ref/cc65070303.c); but, it didn't catch this bug (because it uses a constant subscript, which takes a different path through the compiler).  I added a bigger test program that does catch it.

### DIFF
--- a/src/cc65/expr.c
+++ b/src/cc65/expr.c
@@ -1,7 +1,7 @@
 /* expr.c
 **
 ** 1998-06-21, Ullrich von Bassewitz
-** 2015-04-19, Greg King
+** 2015-06-26, Greg King
 */
 
 
@@ -1713,8 +1713,13 @@ void hie10 (ExprDesc* Expr)
                 } else {
                     Error ("Illegal indirection");
                 }
-                /* The * operator yields an lvalue */
-                ED_MakeLVal (Expr);
+                /* If the expression points to an array, then don't convert the
+                ** address -- it already is the location of the first element.
+                */
+                if (!IsTypeArray (Expr->Type)) {
+                    /* The * operator yields an lvalue */
+                    ED_MakeLVal (Expr);
+                }
             }
             break;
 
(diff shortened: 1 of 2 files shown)


oliverschmidt:
Hi Greg,

> This fixes a bug that was reported by Monte Carlos on the mailing list; then, analyzed by Daniel Serpell.

Thanks once more for your continuous contribution to cc65 - especially in not-so-sexy areas like bug fixing :-))

> The fix is very simple -- which makes me a little nervous. ;-)

I fully understand you !

> I hope that there are no regressions. It passes the test-suite

From my perspective that's a valid working hypothesis.

> I added a bigger test program that does catch it.

Thanks - I guess that especially @mrdudz will be happy to see that the test suite is alive and kicking.

Regards,
Oliver

mrdudz:
On Wednesday 01 July 2015, 11:54:11 Oliver Schmidt <notifications@github.com> 
wrote:
> > I added a bigger test program that does catch it.
> 
> Thanks - I guess that especially @mrdudz will be happy to see that the test
> suite is alive and kicking.

yup :) personally i would make adding a testcase a hard requirement for making 
changes to the compiler (which shouldnt be a big problem... as code that fails 
must exist to even notice the problem in the first case)

-- 

�hnliche Synergieeffekte gibt es zwischen Breitband-Internet und 
Festplattenbedarf, zwischen Computerspielen und Grafikkartenbedarf und 
zwischen Webforen und Tischkantenbedarf. 
<Fefe>

PR #173, opened 2015-07-09
two more tests

mrdudz:
two more tests that i found on my hd... add3a is related to this: https://sourceforge.net/p/cc65/mailman/cc65-devel/thread/546287AB.80100%40users.sf.net/#msg33028349 - and i dont recall where the other relates to :)


oliverschmidt:
Thanks :-)

PR #174, opened 2015-07-10
use rm -r instead of rmdir, and use -f for rm

mrdudz:
this makes subsequent runs of "make clean" not fail on *nix

### DIFF
--- a/test/Makefile
+++ b/test/Makefile
@@ -16,9 +16,9 @@ ifdef CMD_EXE
   RMDIR = rmdir /s /q $(subst /,\,$1)
 else
   EXE :=
-  DEL = $(RM) $1
+  DEL = $(RM) -f $1
   MKDIR = mkdir $1
-  RMDIR = rmdir $1
+  RMDIR = $(RM) -rf $1
 endif
 
 WORKDIR := ../testwrk
--- a/test/Makefile
+++ b/test/Makefile
@@ -13,7 +13,8 @@ ifdef CMD_EXE
   EXE := .exe
   DEL = del /f $(subst /,\,$1)
   MKDIR = mkdir $(subst /,\,$1)
-  RMDIR = rmdir /s /q $(subst /,\,$1)
+  DIRLIST = $(strip $(foreach dir,$1,$(wildcard $(dir))))
+  RMDIR = $(if $(DIRLIST),rmdir /s /q $(subst /,\,$(DIRLIST)))
 else
   EXE :=
   DEL = $(RM) $1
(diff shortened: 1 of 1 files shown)


test/Makefile
```diff
@@ -16,9 +16,9 @@ ifdef CMD_EXE
   RMDIR = rmdir /s /q $(subst /,\,$1)
 else
   EXE :=
-  DEL = $(RM) $1
+  DEL = $(RM) -f $1
```
oliverschmidt: According to the GNU make doc the default value for `RM` is `rm -f`. That's the reason why I introduced `$(RM)` in the first place. Are you sure that the `-f`is actually necessary?

mrdudz: On Friday 10 July 2015, 02:04:02 Oliver Schmidt <notifications@github.com> 
wrote:
> > @@ -16,9 +16,9 @@ ifdef CMD_EXE
> > 
> >    RMDIR = rmdir /s /q $(subst /,\,$1)
> >  
> >  else
> >  
> >    EXE :=
> > 
> > -  DEL = $(RM) $1
> > +  DEL = $(RM) -f $1
> 
> According to the GNU make doc the default value for `RM` is `rm -f`. That's
> the reason why I introduced `$(RM)` in the first place. Are you sure that
> the `-f`is actually necessary?

oh! you are correct, it is not... i also removed the (hopefully now 
unnecessary) "-" prefix in the clean rule ... someone might want to test if it 
still works on windows now :)

-- 

Some people say that I must be a horrible person, but that's not true. I have 
the heart of a small boy --- in a jar on my desk 
<Stephen King>

mrdudz:
"I thought you'd consider it too "lengthy"..."
ugh, now thats ugly =) i cant really test it on windows atm, no idea how it would have to be for DEL exactly

oliverschmidt:
So I was right not to bring that approach to the test Makefile ;-)

With https://github.com/cc65/cc65/commit/992f0f03c5706fb91e60a8ebbcf5de2c79af68e9 I implemented something more lightweight...

greg-king5:
> DIRLIST = $(strip $(foreach dir,$1,$(wildcard $(dir))))

$(wildcard) has a built-in $(foreach).  So, $(foreach) can be removed from those "macroes".

mrdudz:
ok, merged your changes, and replaced rmdir by rm -r for *nix, then it will work fine on both, i guess :)

oliverschmidt:
I think so too :-)

greg-king5:
Strictly speaking, `DEL` isn't needed anymore.  Both `RMDIR` are recursive, now.  Both of them can erase bdiff.

oliverschmidt:
@greg-king5: Thanks for the hint - see https://github.com/cc65/cc65/commit/79e1b25c6c10f5a12b607d201bcddc3b1d98d999

PR #176, opened 2015-07-10
patch from Uz that makes some illegal operations on pointers error out

mrdudz:
see: http://www.cc65.org/mailarchive/2015-03/11726.html
and: http://www.cc65.org/mailarchive/2015-03/11734.html

this is NOT ready to be merged yet - at least a few more tests should created. also perhaps someone else should test the patch with some bigger program first :) 

that said i have run the testsuite on it and that works - so chances are the patch actually works as intended.

### DIFF
--- a/src/cc65/expr.c
+++ b/src/cc65/expr.c
@@ -50,6 +50,7 @@
 /* Generator attributes */
 #define GEN_NOPUSH      0x01            /* Don't push lhs */
 #define GEN_COMM        0x02            /* Operator is commutative */
+#define GEN_NOFUNC      0x04            /* Not allowed for function pointers */
 
 /* Map a generator function and its attributes to a token */
 typedef struct {
@@ -2042,6 +2043,11 @@ static void hie_compare (const GenDesc* Ops,    /* List of generators */
         Tok = CurTok.Tok;
         NextToken ();
 
+        /* If lhs is a function, convert it to pointer to function */
+        if (IsTypeFunc (Expr->Type)) {
+            Expr->Type = PointerTo (Expr->Type);
+        }
+
         /* Get the lhs on stack */
         GetCodePos (&Mark1);
         ltype = TypeOf (Expr->Type);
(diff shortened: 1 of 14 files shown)


mrdudz:
ok here are the missing tests.... in theory it can be merged, i guess. perhaps you want to compile contiki with it first, or something :)

src/cc65/expr.c
```diff
@@ -2775,10 +2816,10 @@ static void hie6 (ExprDesc* Expr)
...
     static const GenDesc hie6_ops [] = {
-        { TOK_LT,       GEN_NOPUSH,     g_lt    },
-        { TOK_LE,       GEN_NOPUSH,     g_le    },
-        { TOK_GE,       GEN_NOPUSH,     g_ge    },
-        { TOK_GT,       GEN_NOPUSH,     g_gt    },
+        { TOK_LT,       GEN_NOPUSH | GEN_NOFUNC,     g_lt    },
+        { TOK_LE,       GEN_NOPUSH | GEN_NOFUNC,     g_le    },
+        { TOK_GE,       GEN_NOPUSH | GEN_NOFUNC,     g_ge    },
+        { TOK_GT,       GEN_NOPUSH | GEN_NOFUNC,     g_gt    },
         { TOK_INVALID,  0,              0       }
```
oliverschmidt: Uz' original patch contains alignment adjustment of this line ;-)

oliverschmidt:
Contiki compiles still fine - I'd opt to merge if no one opposes.

src/cc65/expr.c
mrdudz: perhaps... i had to apply most of it manually :)

mrdudz:
same for the chameleon menu ... go for it :)

src/cc65/expr.c
oliverschmidt: That's why. I fixed it for you - just for once ;-)

...And thanks for preparing the PR incl. additional tests !

PR #177, opened 2015-07-12
new target: PC Engine (TurboGrafx16)

mrdudz:
this adds initial and rudimentary support for the PC Engine family ... it's a bit rough still, so i am putting it up here for discussion - it has been rotting on my HD long enough now (this is basically what i used for my PCE contiki port ages ago)

libsrc/pce/readme.txt has some more info


asminc/pce.inc
```diff
@@ -0,0 +1,105 @@
...
+.macro staio arg1
+        .byte $8d
+        .word arg1
+.endmacro
+
+;; stz abs
+.macro stzio arg1
+        .byte $9c
+        .word arg1
+.endmacro
```
greg-king5: Those macroes aren't needed.  In ca65, we can use a prefix to force a desired addressing mode:
```
lda  a:VDC_CTRL
stz  a:VDC_DATA_HI
```

include/pce.h
```diff
@@ -0,0 +1,95 @@
...
+#define COLOR_BROWN             0x09
+#define COLOR_LIGHTRED          0x0A
+#define COLOR_GRAY1             0x0B
+#define COLOR_GRAY2             0x0C
+#define COLOR_LIGHTGREEN        0x0D
+#define COLOR_LIGHTBLUE         0x0E
+#define COLOR_GRAY3             0x0F
+
+#define CLOCKS_PER_SEC          50      // FIXME: is this correct?
+#define CLK_TCK                 50      // FIXME: is this correct?
```
greg-king5: Those two macroes belong in `<time.h>`.

libsrc/pce/conio.s
```diff
@@ -0,0 +1,102 @@
+                .include "pce.inc"
+
+                .import vce_init
+                .import psg_init
+                .import vdc_init
+
+                .export initconio
```
greg-king5: `.constructor` does a `.export`.

libsrc/pce/conio.s
```diff
@@ -0,0 +1,102 @@
...
+                stz     VCE_ADDR_HI
+
+                ldx #0
+@lp:
+                .repeat 16
+                lda colors,x
+                sta     VCE_DATA_LO
+                lda colors+1,x
+                sta     VCE_DATA_HI
+                .endrepeat
```
greg-king5: That code is done only once.  It doesn't need to be fast.  So, it should be shortenned by rolling it into an inner loop.

libsrc/pce/cputc.s
```diff
@@ -0,0 +1,99 @@
...
+                staio   VDC_DATA_LO     ; character
+
+                lda   CHARCOLOR
+
+                asl a
+                asl a
+                asl a
+                asl a
+
+                and #$f0
```
greg-king5: Not needed -- the left-shifts did that already.

libsrc/pce/crt0.s
```diff
@@ -0,0 +1,213 @@
...
+
+        .include "pce.inc"
+
+        .importzp sp
+        .importzp ptr1,ptr2
+
+; ------------------------------------------------------------------------
+; Create an empty LOWCODE segment to avoid linker warnings
+
+                .segment "LOWCODE"
```
greg-king5: Those lines are obsolete.  We use "optional=yes" in the .cfg file, now.

libsrc/pce/crt0.s
```diff
@@ -0,0 +1,213 @@
...
+                bne @l3
+
+                ; setup the stack
+                lda #<(__RAM_START__+__RAM_SIZE__)
+                sta        sp
+                lda #>(__RAM_START__+__RAM_SIZE__)
+                sta        sp+1
+
+                ; Init the Heap
+                jsr initheap
```
greg-king5: Not needed, it's a constructor.

asminc/pce.inc
mrdudz: see how old this stuff is :)

mrdudz:
fixed those, thanks for the review!

libsrc/pce/crt0.s
mrdudz: or not at all in this case, as its pointless for PCE :)

mrdudz:
ok well, thats it for a while from my side - its on a similar state as the NES target now (except missing TGI driver) and can be merged for all i care :)

mrdudz:
mmmh i considered these being in the common portion of conio - wouldnt it make a lot more sense to introduce a common conio.inc and then make it so? i can't see a good reason to not do it at least (and i really don't like duplicating code in archdep code when it is not absolutely necessary) - or is there one?

oliverschmidt:
@mrdudz 
Mybe I'm missing the point but when I search for `_chline` in the cc65 codebase I find several implementations - and they are not equal. How they be in the common portion of conio?

From my understanding conio is supposed to be fast and therefore supposed to make use of target-specific options a common code wouldn't have.

Having that said I know that there's quite some duplicate code in the cc65 C libraries. That's because with Uz's build system a single target needing a specific implementation of a certain function meant to duplicate the "nearly common" implementation into all other targets. The build system I created on the other hand allows to start with a common function implementation usable for many targets and only "override" the implementation for targets requiring (or benefiting form) a specific one.

However I diid not go over the exsisting code and de-duplicate exsisting implementations. I know i.e. quite some code duplicate in almost all CBM targets. That could go into the `cbm` directory with one or two overrides. However wants to optimize...

Maybe it makes sense to see the `chline.s` / `cvline.s` as new common implementations overridden by most exsisting targets.

mrdudz:
oh well.. i didnt look at all targets before, but a lot of the implementations are near identical and the only difference are the character codes used (and slightly different order of the instructions). these seems to be very good candidates for what you say anyhow =)

for the time being, i added them anyway (not how they are the same as NES and CBM ones, and perhaps others too)

oliverschmidt:
>  these seems to be very good candidates for what you say anyhow =)

Yep. I'm pretty strict on DRY - even if it adds complexity to the code structure. Uz was pretty much the opposite. I.e. have the `apple2` and the `apple2enh` shared one source directory was something I more or less pushed through as prerequisite for maintaining the targets. For Uz it would have been totally obvious to duplicate the code...

> for the time being, i added them anyway (note how they are the same as NES and CBM ones, and perhaps others too :))

You can't take care of everything at once - that code style topic is going to be enough additional burden ;-))

polluks:
Well, cvline and chline are no problem but hello.c also requires cgetc. No common code anymore...

mrdudz:
mmmh cgetc doesnt make a lot of sense if there is no keyboard though.... should i add a dummy function that never returns? :)

polluks:
You need something because of flickering, better check the joystick IMHO. Imagine you would connect a CX21 :-)

oliverschmidt:
I pretty strongly disagree here: It's not considered reasonable to implement some function as dummy and/or emulation "just" to make some demo/test program link.

To me consistency is very important - so in almost all cases it makes sense to look what's already there and how similar issues have already been solved. Like with the hline/vline question: The current style is obviously to rather provide a very similar function for every target.

So we already have targets without a keyboard. They do _not_ implement `_cgetc` at all. That's not because people are necessarily stupid or lazy ;-) Rather it's because that's how it is wanted: No keyboard -> no keyboard functions.

Regarding the `hello world` I'd suggest to use #ifdef to switch between cget() and something else. A timer? A joystick button? Just a loop?

greg-king5:
The Atari Lynx has no keyboard; but, Karri implemented a `kbhit()` and a `cgetc()` that return a few characters that are appropriate for the three control buttons on the box.  You could do the same thing for the PCE's Select and Run buttons.

mrdudz:
i agree with oliver there, the examples should be fixed accordingly instead. infact i had such a "mapping" function before and deleted it while polishing the lib :)

oliverschmidt:
You can easily have different, still valid perspectives on theses questions. You could i.e. even argue that a target without hardware chargen shouldn't implement conio at all (but fully rely on TGI)...

However here's my personal take: If a machine has a remote device attached to it with a small number of buttons then this device should likely be classified as a joypad. And joypads should generally be handled by joystick drivers - not console input handlers.

The PC Engine joypad seems pretty similiar to the NES joypad. The cc65 NES target does _not_ implement a console input handler but rather a joystick driver - a good decision from my pov.

The Lynx is from my pov different as it combines "the machine" and "the joypad" in one device. Looking at pics it seems reasonable to see the buttons near the screen as part of the machine rather than the joypad. And as such I can follow the decision to create a console input handler for them.

Anyway - with https://github.com/cc65/cc65/commit/1072f7d6f7f343b4dcc48739d482f01a2abc24fb I modified `hello world` to check for any button pressed on the first joystick when running on the NES. You may just add you new target to the `#if`there...

mrdudz:
i have (hopefully) fixed all style issues (by hand, the plan with modifying c65html didnt work out :/) - please review and let me know if i missed something.
i also need help with adding support for the interrupt constructor/hook stuff, i dont even know how to use it (never did, never will =P), let alone implement it :)
other than that it could be merged now... i guess.

doc/pce.sgml
```diff
@@ -0,0 +1,219 @@
...
+<sect>Platform specific header files<p>
+
+Programs containing PCE specific code may use the <tt/pce.h/ header file.
+
+
+<sect1>PCE specific functions<p>
+
+<itemize>
+<item>waitvblank</item>
+<item>get_tv</item> (since all PCE systems are NTSC, this always returns TV_NTSC)
```
greg-king5: The `</item>` tag needs to be at the end of that line.

doc/pce.sgml
```diff
@@ -0,0 +1,219 @@
...
+The names in the parentheses denote the symbols to be used for static linking of the drivers.
+
+
+<sect1>Graphics drivers<p>
+
+<descrip>
+
+No TGI graphics drivers are currently available for the PCE.
+
+</descrip><p>
```
greg-king5: The `<descrip>` and `</descrip><p>` lines should be removed (there currently is nothing to describe there).

doc/pce.sgml
```diff
@@ -0,0 +1,219 @@
...
+<item>fputc
+<item>fscanf
+<item>fwrite
+<item>...
+</itemize>
+
+<sect>Other hints<p>
+
+<itemize>
+<item>a good emulator to use for PC-Engine is "mednafen" (<url url="http://mednafen.sourceforge.net">)
```
greg-king5: Mednafen moved to http://mednafen.fobby.net/.

mrdudz:
fixed, thanks for the review!

testcode/lib/pce/Makefile
```diff
@@ -0,0 +1,12 @@
+
+all: conio.pce
+
+conio.pce: conio.c
+	../../../bin/cl65 -t pce conio.c ../../../joy/pce-stdjoy.joy --mapfile conio.map -o conio.pce
```
greg-king5: The static drivers are in the libraries.  We don't need to link to the dynamic ones (remove that
 `../../../joy/pce-stdjoy.joy`).

testcode/lib/pce/conio.c
```diff
@@ -0,0 +1,121 @@
...
+        cprintf("stackvar: %02x\n\r", stackvar);
+
+        j = joy_count();
+        gotoxy(0,10);
+        cprintf("Found %d Joysticks.", j);
+
+        for (i = 0; i < 4; ++i) {
+                gotoxy(0, 17 + i);
+                p = malloc(16);
+                memcpy(p, "01234567890abcdef", 16);
```
greg-king5: There is an extra zero in the string (therefore, the '`f`' isn't copied).

README.md
```diff
@@ -4,6 +4,8 @@
 
 [![build status](http://travis-ci.org/cc65/cc65.png)](http://travis-ci.org/cc65/cc65/builds)
```
greg-king5: That line is strange.  It is different in cc65:master (see dd75d3f8841b5828e173de5fe7f4a7fc8dfe7c47); but, it isn't flagged as a change here.  It might cause a merge conflict.

README.md
```diff
@@ -23,6 +25,7 @@ including
 - the Atari 8 bit machines.
 - the Atari 5200 console.
 - GEOS for the C64, C128 and Apple //e.
+- the NEC PC-Engine
```
greg-king5: Missing period (or, full-stop).

asminc/pce.inc
```diff
@@ -0,0 +1,89 @@
+;
+; PCE definitions. By Groepaz/Hitmem.
+;
+
+; FIXME: optimize zeropage usage
+CURS_X          = $30
+CURS_Y          = $31
+SCREEN_PTR      = $32           ;2
+CRAM_PTR        = $34           ;2
```
greg-king5: That pointer variable isn't needed (because there is no separate color RAM).

libsrc/pce/kplot.s
```diff
@@ -0,0 +1,36 @@
...
+
+_plotlo:
+        .repeat screenrows,line
+        .byte   <($0000+(line*$80))
+        .endrepeat
+
+_plothi:
+        .repeat screenrows,line
+        .byte   >($0000+(line*$80))
+        .endrepeat
```
greg-king5: Those tables belong in `.rodata`.

README.md
polluks: maybe add (aka TurboGrafx-16).

libsrc/pce/psg.s
```diff
@@ -0,0 +1,29 @@
+
+        .include        "pce.inc"
+
+        .export         psg_init
+
+psg_init:
+        clx
+        stx     $0800           ; Select channel
```
greg-king5: That line belongs inside the loop (otherwise, the loop just repeats on the first channel).

libsrc/pce/psg.s
```diff
@@ -0,0 +1,29 @@
...
+psg_clear_loop:
+        stz     $0801           ; Clear global balance
```
greg-king5: That line belongs outside the loop (it controls the sound _after all six_ channels have been mixed together).

mrdudz:
ok, fixed... thanks again!

asminc/pce.inc
```diff
@@ -0,0 +1,99 @@
...
+VDC_HSR         = 9             ; Horizontal Sync Register (?)
+VDC_HDR         = 10            ; Horizontal Display Register (?)
+VDC_VPR         = 11            ; (unknown)
+VDC_VDW         = 12            ; (unknown use)
+VDC_VCR         = 13            ; (unknown use)
+VDC_DCR         = 14            ; (DMA) Control Register
+VDC_SOUR        = 15            ; (DMA) Source Register
+VDC_DESR        = 16            ; (DMA) Destination Register
+VDC_LENR        = 17            ; (DMA) Length Register
+VDC_SATB        = 18            ; Sprite Attribute Table
```
greg-king5: Those VDC register assignments are wrong.  VWR and VRR are the same register.  And, registers 3 and 4 are not used.

After you fix that list, you must change `VDC_RCR` to `VDC_CR`, in "conio.s".

mrdudz:
wow! those must have been wrong in the old vdc doc i was using back then.... its correct in the one i have now, of course. thanks!

libsrc/pce/color.s
```diff
@@ -0,0 +1,64 @@
...
+        sta     VCE_DATA_LO
+        lda     colors+1,y
+        sta     VCE_DATA_HI
+
+        txa
+        rts
+
+_bordercolor:
+        lda     #0
+        txa
```
greg-king5: That line makes the function return garbage.

libsrc/pce/color.s
```diff
@@ -0,0 +1,64 @@
...
+        rts
+
+        .export colors
+
+colors:
```
greg-king5: This table should be in RODATA.

asminc/pce.inc
```diff
@@ -0,0 +1,101 @@
...
+;
+
+; FIXME: optimize zeropage usage
+CURS_X          = $30
+CURS_Y          = $31
+SCREEN_PTR      = $32           ;2
+CHARCOLOR       = $34
+RVS             = $35
+BGCOLOR         = $36
+tickcount       = $37           ;4
```
greg-king5: I recommend that you put those variables into an "`extzp.s`" file.  If you define them in an `EXTZP` segment, then you won't need to manage their addresses by hand.

(See the `libsrc/*/extzp.*` files.)

libsrc/pce/crt0.s
```diff
@@ -0,0 +1,207 @@
...
+        ;tam     #%01000000      ; C000-DFFF  Page 5
+        ;lda     #0
+        ;tam     #%10000000      ; e000-fFFF  hucard/syscard bank 0
+
+        ; Clear work RAM (2000-3FFF)
+        stz     <$00
+        tii     $2000, $2001, $1FFF
+
+        ; Initialize hardware
+        stz     TIMER_COUNT     ; Timer off
```
greg-king5: That I/O register must be `TIMER_CTRL`.

asminc/pce.inc
oliverschmidt: Full ACK.

greg-king5:
> Perhaps, you could give some hints where to steal that interruptor stuff?  And more importantly, how to test it? :-)

Try this:
In "irq.s":

~~~gas
;
; IRQ handling (PCE version)
;

        .export         initirq, doneirq, IRQStub

        .import         callirq

        .include        "pce.inc"

.segment        "INIT"

; a constructor
;
initirq:
        rts

.code

; a destructor
;
doneirq:
        rts

IRQStub:
	pha

; Save the display-source flags (and, release the interrupt).
;
        lda     VDC::CTRL
        sta     vdc_flags

        phx
        phy

        jsr     callirq

        ply
        plx
        pla
        rti
~~~

In "ticktock.s":

~~~gas
        .interruptor    ticktock, 24

        .include        "pce.inc"

ticktock:
        bbr5    vdc_flags,@s1   ; not vertical-blank interrupt

        ; Increment the system tick counter.
        inc     tickcount
        bne     @s1
        inc     tickcount+1
        bne     @s1
        inc     tickcount+2
        bne     @s1
        inc     tickcount+3

@s1:    clc                     ; interrupt not handled
        rts
~~~

`vdc_flags` is a byte variable in Zero-page.  Both "clock.s" and "waitvblank.s" should
`.forceimport ticktock`.  Remove the interrupt handler in "crt0.s"; and, point all of the interrupt vectors to `IRQStub`.  Uncomment the interruptor stuff in "pce.cfg".

Your "conio.c" test program will prove that it works (it's broken if the screen doesn't change).

libsrc/pce/clrscr.s
```diff
@@ -0,0 +1,32 @@
...
+        sta     a:VDC_DATA_LO
+        lda     #$02
+        sta     a:VDC_DATA_HI
+
+        dex
+        bne     colloop
+        dey
+        bne     rowloop
+
+        rts
```
greg-king5: This function must put the cursor, also.  Therefore, that `rts` must be replaced with

~~~gas
; Go to the home position.

        stz     CURS_X
        stz     CURS_Y
        jmp     plot
~~~

mrdudz:
ok yay.... so this is it? :)

libsrc/pce/clrscr.s
```diff
@@ -0,0 +1,38 @@
...
+        dex
+        bne     colloop
+        dey
+        bne     rowloop
+
+; Go to the home position.
+
+        stz     CURS_X
+        stz     CURS_Y
+        jmp     PLOT
```
greg-king5: It must be the lower-case spelling of `plot` (i.e., the one in "cputc.s").

greg-king5:
OK Oliver, I think that this pce target is ready to be shown to the world.

oliverschmidt:
@greg-king5: A big 'thank you' for all the effort you put into reviewing this pull request !

@mrdudz: Thanks for the substantial contribution - and your cooperation during the review process :-)

libsrc/pce/crt0.s
```diff
@@ -0,0 +1,178 @@
...
+        jsr     vdc_init
+
+        ; Turn on background and VD interrupt/IRQ1
+        lda     #$05
+        sta     IRQ_MASK        ; IRQ1=on
+
+        ; Clear the BSS data
+        jsr     zerobss
+
+        ; Copy the .data segment to RAM
```
oliverschmidt: Is there a reason not to use `copydata` (see common/copydata.s) ? If no, please create a PR that replaces the following inline code with a call to it. It's btw. even documented: http://cc65.github.io/doc/ld65.html#ss5.4

greg-king5: The HuC6280 CPU lets us do even better -- a single instruction:
`tii __DATA_LOAD__, __DATA_RUN__, __DATA_SIZE__`.

polluks: A typo in http://cc65.github.io/doc/ld65.html#ss5.4 it's --> its

oliverschmidt: > The HuC6280 CPU lets us do even better -- a single block-copy instruction:
tii __DATA_LOAD__, __DATA_RUN__, __DATA_SIZE__.

At that point it may be worth mentioning that the cc65 Build system intentionally allows to "override" source files:

If a target i.e. has a specific way to implement memcpy() and/or memmove() it can do so be placing files with the same name the files in /common have in its own source directory. Then those files will be used.

That behavior is hierarchical. I.e. /c64 files may override /cbm files and /cbm files may override /common files. This allows btw. for quite some de-duplication of code. If. i.e. all CBM targets but one or two can share code then that can be placed in /cbm and the one or two have specific code in their directories.

polluks: Indeed, but copydata is only used in crt0, so you don't want jsr/rts overhead.

oliverschmidt: > Indeed, but copydata is only used in crt0, so you don't want jsr/rts overhead.

* The same holds true for i.e. zreobss and the constructor/destructor stuff. Not inlining them has to do with code maintainability and code reuse.
* The PCE target uses several init_* functions that could be inlined without reducing code reuse.

Maybe the drawbacks could be avoided be replacing the functions in question with macros. Maybe you want to create a pull request re-engineering this area.

Anyhow I'd still like to the see the inline code below replaced with a function call (or the CPU intrinsic pointed out by Greg).

Issue #172, opened 2015-07-08
Two bugs in the string output functions


greg-king5:
1. `sprintf()`, `snprintf()`, `vsprintf()`, and `vsnprintf()` might crash if `NULL` (0) is used as the first (buffer pointer) argument.
2. `snprintf()` and `vsnprintf()` always return zero when the second (buffer-length) argument is zero.  They should return the same value that they do return when the buffer length isn't zero (i.e., the length of the string that the format produces).

P.S. Assign me to this one.

oliverschmidt:
@greg-king5 Thanks for the fix :-)

PR #180, opened 2015-07-18
Make the sprintf() family of functions handle special argument conditions.

greg-king5:
The functions properly will handle NULL buffer pointers and the zero buffer-length condition.

Fixes issue #172.


oliverschmidt:
Thanks for the fix :-)

Issue #175, opened 2015-07-10
Bug in decimal (base 10) number evaluation


Movax12:
ca65 allows 'a' or 'A' to be included after a leading decimal digit. It is evaluated as if it was the digit 10.

example:
```
; foo == 20 (decimal)
foo = 1a
```
This line needs to be changed: 
https://github.com/cc65/cc65/blob/master/src/ca65/scanner.c#L1014
to:

```C
    if (DVal >= Base) {
                Error ("Invalid digits in number");
                CurTok.IVal = 0;
                break;
     }
```

Movax12:
https://github.com/cc65/cc65/pull/181

oliverschmidt:
@Movax12 Thanks for the fix :-)

PR #181, opened 2015-07-19
Fix base 10 bug (ca65 allows 'a' or 'A' in base10 value)

Movax12:
Fix for bug as mentioned: https://github.com/cc65/cc65/issues/175


oliverschmidt:
Thanks for the fix :-)

Issue #182, opened 2015-07-20
ca65 line continuation feature bug


Movax12:
Using the line continuation feature results in ca65 skipping the leftmost column when a line is continued. Example:
````
.linecont +

foo = \
22

; outputs "foo:2"
.out .sprintf("foo:%d", foo)

; this does not cause an error!
bar = \
"1

; outputs "bar:1"
.out .sprintf("bar:%d", bar)
```

There was some discussion on how to fix this earlier. I'll submit a patch of how I prefer it, opinions welcomed.

edit:update example.

patch: https://github.com/cc65/cc65/pull/183

PR #183, opened 2015-07-20
Fix bug #182

Movax12:


oliverschmidt:
Hi Greg,

I seem to remember that you looked a lot into these "ca65 syntax extension pull requests". Could you please provide some feedback here?

Thanks in advance,
Oliver

greg-king5:
This patch is good; it's ready to be merged.
It removes the one statement (`C = ' ';`) that caused the trouble, adds a simple error trap, and doesn't do anything that could cause a regression.

oliverschmidt:
@Movax12: Thanks for the fix.
@greg-king5: Thanks for your feedback.

PR #185, opened 2015-07-23
Add support for weak symbols in assembly source files

groessler:
These changes add support for weak symbols in assembly source files.

'.weak' behaves like '.import', except that if the symbol isn't defined at the linking stage, it's resolved to zero and not error is thrown.
A slightly different behaviour is implemented if the weak symbol was the argument of a JSR: In this case the JSR instruction is replaced by NOPs.

I had to change the object file format to implement this and increased the version field in the object file header.

### DIFF
--- a/src/ca65/instr.c
+++ b/src/ca65/instr.c
@@ -110,6 +110,11 @@ static void PutTMAn (const InsDesc* Ins);
 static void PutTST (const InsDesc* Ins);
 /* Emit a TST instruction (HuC6280). */
 
+static void PutJSR (const InsDesc* Ins);
+/* Emit a JSR instruction. It will indicate that it's "NOPable" for weak
+** symbols.
+*/
+
 static void PutJMP (const InsDesc* Ins);
 /* Handle the jump instruction for the 6502. Problem is that these chips have
 ** a bug: If the address crosses a page, the upper byte gets not corrected and
@@ -181,7 +186,7 @@ static const struct {
         { "INX",  0x0000001, 0xe8, 0, PutAll },
         { "INY",  0x0000001, 0xc8, 0, PutAll },
         { "JMP",  0x0000808, 0x4c, 6, PutJMP },
-        { "JSR",  0x0000008, 0x20, 7, PutAll },
+        { "JSR",  0x0000008, 0x20, 7, PutJSR },
         { "LDA",  0x080A26C, 0xa0, 0, PutAll },
         { "LDX",  0x080030C, 0xa2, 1, PutAll },
         { "LDY",  0x080006C, 0xa0, 1, PutAll },
@@ -255,7 +260,7 @@ static const struct {
         { "ISC",  0x000A26C, 0xE3, 0, PutAll },         /* X */
         { "JAM",  0x0000001, 0x02, 0, PutAll },         /* X */
         { "JMP",  0x0000808, 0x4c, 6, PutJMP },
-        { "JSR",  0x0000008, 0x20, 7, PutAll },
+        { "JSR",  0x0000008, 0x20, 7, PutJSR },
         { "LAS",  0x0000200, 0xBB, 0, PutAll },         /* X */
         { "LAX",  0x080A30C, 0xA3, 11, PutAll },        /* X */
         { "LDA",  0x080A26C, 0xa0, 0, PutAll },
@@ -335,7 +340,7 @@ static const struct {
         { "INX",  0x0000001, 0xe8, 0, PutAll },
         { "INY",  0x0000001, 0xc8, 0, PutAll },
         { "JMP",  0x0010808, 0x4c, 6, PutAll },
-        { "JSR",  0x0000008, 0x20, 7, PutAll },
+        { "JSR",  0x0000008, 0x20, 7, PutJSR },
         { "LDA",  0x080A66C, 0xa0, 0, PutAll },
         { "LDX",  0x080030C, 0xa2, 1, PutAll },
         { "LDY",  0x080006C, 0xa0, 1, PutAll },
@@ -427,7 +432,7 @@ static const struct {
         { "INX",  0x0000001, 0xe8, 0, PutAll },
         { "INY",  0x0000001, 0xc8, 0, PutAll },
         { "JMP",  0x0010808, 0x4c, 6, PutAll },
-        { "JSR",  0x0000008, 0x20, 7, PutAll },
+        { "JSR",  0x0000008, 0x20, 7, PutJSR },
         { "LDA",  0x080A66C, 0xa0, 0, PutAll },
         { "LDX",  0x080030C, 0xa2, 1, PutAll },
         { "LDY",  0x080006C, 0xa0, 1, PutAll },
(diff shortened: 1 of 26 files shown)


greg-king5:
I doubt that this is what Uz was planning for "weak assembler symbols".  I suspect that he wanted `.weak` to export symbols; and, we would be allowed to change their values, on ld65's command line.  In other words, it would be the same as ld65's weak symbols; but, ca65 would be the source of the name.

Perhaps, "`.loose`", or something similar to that, would be a better name for this feature.

> A slightly different behaviour is implemented if the weak symbol was the argument of a JSR: In this case, the JSR instruction is replaced by NOPs.

Maybe, that behaviour should be done for all op-codes that have an address operand.  After all, we don't want something like "`sta There`" to store at location zero if `There` is missing.

src/ca65/instr.c
```diff
@@ -783,7 +788,7 @@ static const InsTable* InsTabs[CPU_COUNT] = {
     (const InsTable*) &InsTab65816,
     (const InsTable*) &InsTabSweet16,
     (const InsTable*) &InsTabHuC6280,
-    0,                                  /* Mitsubishi 740 */
+    0,
```
greg-king5: I think that the comment should stay there.  Otherwise, a reader might wonder, "Huh, why is that NULL there?"  Or, he might think that the list needs to be terminated by a NULL pointer.

src/ca65/instr.c
```diff
@@ -1043,7 +1048,7 @@ static int EvalEA (const InsDesc* Ins, EffAddr* A)
 
 
 
-static void EmitCode (EffAddr* A)
+static void EmitCode (EffAddr* A, int Nopable)
```
greg-king5: I would like that second parameter to be spelled `NOPable`.

oliverschmidt:
I agree with Greg that `.weak` is (supposed to be) rather a "special" `.export` than a "special" `.import`.

In GCC `__attribute__((weak))` has sort of both meanings depending on the context. But if we don't use the term as an attribute but as an "replacement" for `.import` or `.export` then it has to be rather a replacement for `.export`.

Regarding a nice name for the feature discussed here it seems a bit like the opposite of `.forceimport`. Maybe that notion can help to come up wich a good idea (like `.dontforce`).

Another approach to the name could simply say when we already have `.forceimport` as a special `.import` we can as well have `.weakimport` as another special `.import` - not that bad after all...

oliverschmidt:
Regarding the replacement of NULL by "something special" I must admit that I'm not convinced that this is a good idea. How does one know that the correct reaction to the symbol missing is "to do nothing". Maybe a special action needs to be performed to compensate for the missing symbol. There seems just to much semantics implied.

Wouldn't it make much more sense to actually have `.weak` beside the `.weakimport` discussed here. Then one could define a fallback for the missing symbol - in case of a function maybe/likely simply consisting of an `rts`.

src/ca65/instr.c
groessler: But ca65 doesn't seem to support a "Misubishi 740" cpu...

spiro-trikaliotis: * On Fri, Jul 24, 2015 at 12:33:41PM -0700 Christian Groessler wrote:
> In src/ca65/instr.c:
> 
> > @@ -783,7 +788,7 @@ static const InsTable* InsTabs[CPU_COUNT] = {
> >      (const InsTable*) &InsTab65816,
> >      (const InsTable*) &InsTabSweet16,
> >      (const InsTable*) &InsTabHuC6280,
> > -    0,                                  /* Mitsubishi 740 */
> > +    0,
> 
> But ca65 doesn't seem to support a "Misubishi 740" cpu...

The Mitsubishi 740 seems to be very similar to the 6502:

http://www.hifi-remote.com/files/chip-manuals/Mits-740-sum.pdf

I believe that Uz once said that he made a special version of ca65
(under NDA) for a special processor. Perhaps, it was the Mitsubishi 740,
and the comment above is a leftover of that?

Regards,
Spiro.

-- 
Spiro R. Trikaliotis
http://www.trikaliotis.net/

spiro-trikaliotis: I stand correct, Uz looked into adding the Mitsubishi/Renesa 740 in 2003: http://www.cc65.org/mailarchive/2003-08/3202.html

groessler:
> Wouldn't it make much more sense to actually have |.weak| beside the 
> |.weakimport| discussed here. Then one could define a fallback for the 
> missing symbol - in case of a function maybe/likely simply consisting 
> of an |rts|. 

I'll look into it...

greg-king5:
> Another approach to the name could simply say, when we already have `.forceimport` as a special `.import`, we can, as well, have `.weakimport` as another special `.import` -- not that bad, after all.

Now that you mention it, I think that Uz would have chosen names like `.exportweak`, `.importweak`, and maybe even `.globalweak`.  Or maybe, he would have used a suffix qualifier:

```GAS
.import Name:absolute:weak
.export Name:abs:wk
```

oliverschmidt:
Hi,

I don't wanna be the spoilsport but thinking about this some more...

I guess we all agree that it makes sense to allow the assembler to define/export weak symbols behaving just like the weak symbols defined/exported by the linker.

But when I presume the existence of "weak exporting" then I don't see the use case for "weak importing" as it is proposed in this pull request: If I want to refer to a symbol in case it was defined/exported but don't want to fail to link in case it wasn't then I can import it in the usual fashion and add a weak export used by the linker in case it wasn't defined/exported elsewhere. This approach has the benefit over "weak importing" that I have exact control over the symbol value in case it wasn't defined/exported elsewhere.

Or do I miss the point here ?!?

Regards,
Oliver

greg-king5:
You are not a spoilsport; you are a sage.  ;-)

(I would not have found out about this if the pull request had not started this conversation.)
We can do what you described -- but, we must do it indirectly, by importing and weak-exporting through a configure file.  It makes sense to assume that Uz had planned that we could do it without needing to use custom configure files.

src/ca65/instr.c
groessler: Responding to the original comment, I think the comment should stay there, unless we remove the incomplete support for the Misubishi/Renesas chip altogether.

groessler:
On 07/27/15 03:21, greg-king5 wrote:
>  It makes sense to assume that Uz had planned that we could do it 
> without needing to use custom configure files.


I don't know what Uz had planned, but I don't think weak symbols for the 
assembler were on his radar at all.

groessler:
I'm closing this pull request. I'll try to implement the missing semantics (see https://en.wikipedia.org/wiki/Weak_symbol) and come back with a new one. Since I'm busy with other things this might take a while.
(That's what I get for not checking the requirements... :-/)

greg-king5:
I had seen that the wiki "Reminders" page mentions assembler weak symbols; and, I had assumed that Uz had written that reminder long ago.  But now, I see that you added it to the wiki.  I apologize for "robbing" you of the credit for the idea.

oliverschmidt:
Hi Chris,

I'm glad to see that you are able find a such a constructive perspective - and I'm of course nevertheless sorry for your effort invested in vain!

Regards,
Oliver

groessler:
> I'm glad to see that you are able find a such a constructive 
> perspective - and I'm of course nevertheless sorry for your effort 
> invested in vain!

I don't think the effort was invested in vain. Even if my next pull 
request is rejected I've learned quite a bit of the ca65 and ld65 
internals... :-)

But maybe I wasn't clear. I intend to implement "weak" as kind of an 
attribute, both for imports and for exports. Weak imports behave like in 
this pull request.
Weak exports can be overridden by a non-weak symbol, like it is 
described in the Wikipedia page.

I like Greg's idea to have a ":weak" qualifier for .import and .export. 
In a simple solution I'd got for .importweak and .exportweak.

oliverschmidt:
Hi Chris,

> I intend to implement "weak" as kind of an
attribute, both for imports and for exports. Weak imports behave like in
this pull request.
Weak exports can be overridden by a non-weak symbol, like it is
described in the Wikipedia page.

I argued that when having weak exports there would be no use case for weak imports. I'm open to being convinced that I'm wrong. However without being convinced I'd have a hard time to merge a pull request bringing in new code which is "unnecessary".

Please make sure to have this clarified before investing further effort!

Regards,
Oliver

Issue #187, opened 2015-07-29
No BLT mnemonic when using 65816 mode


furrykef:
[This page](http://cc65.github.io/doc/ca65.html#ss4.2) claims that BLT is accepted as a synonym for BCC when in 65816 mode. This doesn't appear to be the case, at least of the latest Windows snapshot; I get "Error: ':' expected" when I try to use it, even though I know for a fact that this code is assembling in 65816 mode.

I assume the same is true of the other instructions in the list, but BLT was the only one I've tested so far.

greg-king5:
The first two lines in that list are obsolete.  Those mnemonics are useful for all of the CPUs.  I would have added them, to ca65, for all of the CPUs; but, Uz chose to move them into a macro package.
That package is [documented here](http://cc65.github.io/doc/ca65.html#ss13.1).

furrykef:
That's even better, because I certainly like having these mnemonics available on the 6502. Thanks for pointing out the extended macpack. The old version had only "add" and "sub".

PR #188, opened 2015-07-29
65816 assembler document fix.

greg-king5:
This fixes the confusion that was described by bug report #187.


oliverschmidt:
Thanks :-)

PR #196, opened 2015-08-13
Remove pointless rules from a regression test makefile.

greg-king5:
The files in `test/err/` cannot be compiled.  It doesn't make sense to attempt to assemble and link them.  It doesn't make sense to try different amounts of optimization.

This patch also reduces the number of lines that are written into a Travis-CI build's listing.

### DIFF
--- a/test/err/Makefile
+++ b/test/err/Makefile
@@ -1,4 +1,3 @@
-
 # makefile for the tests that MUST NOT compile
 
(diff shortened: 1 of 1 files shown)


oliverschmidt:
Thanks for the cleanup - incl. the detailed description (which is especially nice when removing stuff) !

Issue #191, opened 2015-08-05
extern/static conflict


polluks:
<pre>extern int xx;
static int xx;

int main()
{
return xx;
}
</pre>

gcc4 says:
a.c:2: error: static declaration of 'xx' follows non-static declaration
a.c:1: note: previous declaration of 'xx' was here

oliverschmidt:
Please elaborate on what you want to express here. In its current state I can't even classify the issue :-(

polluks:
Well, gcc gives an error but cc65 simply ignores it. If you swap the first two lines it's ok.

oliverschmidt:
Thanks for the explanation.

greg-king5:
The C Standard tells what a compiler must do when it sees
```C
static int xx;
extern int xx;
```

But, the Standard doesn't tell what a compiler must do when it sees
```C
extern int xx;
static int xx;
```

cc65 does the same thing for both pairs:  It merges them into the single statement:
```C
int xx;
```

That's a bug for the first pair -- the `extern` shouldn't change the `static` (I'll fix it).

It's OK for the second pair because the Standard says that the behavior is "undefined".
So ... what do people want cc65 to do?  Should it

1. continue to say nothing;
2. give a warning message;
3. or, give an error message?

(I vote for number 2.)

mrdudz:
On Friday 07 August 2015, 14:32:06 greg-king5 <notifications@github.com> 
wrote:
> The C Standard tells what a compiler must do when it sees
> ```C
> static int xx;
> extern int xx;
> ```
> 
> But, the Standard doesn't tell what a compiler must do when it sees
> ```C
> extern int xx;
> static int xx;
> ```
> 
> cc65 does the same thing for both pairs:  It merges them into the single
> statement: ```C
> int xx;
> ```
> 
> That's a bug for the first pair -- the `extern` shouldn't change the
> `static` (I'll fix it).
> 
> It's OK for the second pair because the Standard says that the behavior is
> "undefined". So ... what do people want cc65 to do?  Should it
> 
> 1. continue to say nothing;
> 2. give a warning message;
> 3. or, give an error message?
> 
> (I vote for number 2.)

definately 2 :)

-- 

http://www.hitmen-console.org    http://magicdisk.untergrund.net
http://www.pokefinder.org        http://ar.pokefinder.org

Wer gackert, muss auch Eier legen.

oliverschmidt:
@greg-king5: First of all thanks for looking into this :-)

[...]

PR #197, opened 2015-08-14
Handle almost-duplicate C declarations that have different linkages.

greg-king5:
("Linkages" come from the `extern` and `static` qualifiers that can be put on global declarations/definitions.)

This Pull Request fixes the bug that was described by me in issue #191.  It adds some warning messages, as requested in that issue.  And, it adds some compiler regression tests for that issue.


oliverschmidt:
Thanks! Now I know why you looked at the error test Makefile ;-)

PR #199, opened 2015-08-27
Fix source-code bugs that cause linking problems when set_irq() is used in a program.

greg-king5:


oliverschmidt:
Hi,

You wrote on Uz' mailing list:

=====

The two bugs _are inherited_ from Uz's version 2.13.9:

1. A ".code" line was left out of the "libsrc/common/interrupt.s" source file.  Therefore, ca65 puts the functions into the .bss segment that is defined just above that code.

2. That file uses ".importzp __ZP_START__".  But, of course, ld65 generates the self-defined labels for memory areas and segments as absolute addresses, not zero-page addresses -- even when the addresses are less than $0100.  The fix is to use ".import"; and then, use "<__ZP_START__,x" as the operand in the two instructions that need it.

=====

Regarding 1.) I don't see a fix for this in this pull request. So maybe there wasn't an issue at all.

Regarding 2.) The original code was contributed by me. I'd guess it worked at some point. Strange...

Thanks for the fix :-)

greg-king5:
> Regarding 2.) The original code was contributed by me. I'd guess it worked at some point. Strange.

It did work; but, it was broken by Uz, near the end of 2012: dc4e898abd273970b52c9fbfda6d721c6ee77a09.

oliverschmidt:
Thanks - I appreciate the explanation :-)

It explains the missing `.code` too as Uz reordered stuff. He never liked that C level interrupt thingy - so he maybe didn't care too much about it ;-)

Issue #201, opened 2015-08-27
cl65 crashes with a certain command line option


fo-fo:
cl65 crashes when invoked like this (foo.c can be an empty file):

    cl65 foo.c -l

Should give an error about the missing argument of `-l` instead, I guess.

fo-fo:
I didn't check whether this is a bug in cl65, or the C compiler. Also didn't check whether it occurs in the assembler as well.

greg-king5:
The bug is in the common helper-function library.  argv[], the array of pointers to command-line arguments, is replaced by a second array.  That act allows command files (@filename) to be read and expanded into the command lines that the tools see.  But, `InitCmdLine()` failed to copy the NULL at the end of the original array.

That bug has existed for a long time; but, we didn't see it because most function libraries' `malloc()` happenned to put zeroes where that NULL should be copied.  But, MinGW's `malloc()` doesn't do it; a command like
```sh
cl65 -l
```
will crash with a "`Segmentation fault`".

oliverschmidt:
@greg-king5 Thanks for the analysis and detailed explanation :-)

PR #202, opened 2015-08-30
Fix the preparation of tool command lines.

greg-king5:
This Pull Request fixes bug #201.

It's interesting that no one had noticed that bug in all of the many years that it has existed.


oliverschmidt:
Thanks :-)

Issue #206, opened 2015-09-08
C64 randomize function using wrong timer byte


emuslor:
Current function looks like this:

_randomize:
       ldx     VIC_HLINE       ; Use VIC rasterline as high byte
       lda     TIME            ; Use 60HZ clock as low byte
       jmp     _srand          ; Initialize generator

Second line should be using TIME+2, similar to the way the Atari version uses RTCLOK+2, since the first byte is the longest time to change at 18+ minutes.  +2 gets you the jiffy counter.

polluks:
Indeed TIME is not 60 Hz, also fix
* c128/randomize.s
* c16/randomize.s
* plus4/randomize.s
* cbm510/randomize.s
* cbm610/randomize.s
* vic20/randomize.s

MonteCarlos:
Isn't it possible that combining 50/60Hz timers generates
better random numbers because of the interference?
Using first: period is 20ms,
using second: period is 17ms
using both: period is 300ms (greates common mutiple is 300 pulses).
Maybe this was the original intention of this?

Monte Carlos


Am 08.09.2015 um 17:51 schrieb polluks:

mrdudz:
no, this is only setting the seed anyway. the point is that TIME probably will be always 0, unless you wait a very long time before running your program :) (and the comment clearly shows the intention)

polluks:
Right now pet/randomize.s uses 1 Hz...

mrdudz:
..which would still give you always 0 in the common case of "autostart in emulator" :)

polluks:
Anyway  "autostart in emulator" is an artificial environment, if you need real randomness use a real PET :-p

PR #207, opened 2015-09-11
Fix the CBM random-seed generators.

greg-king5:
Those functions had assumed that TIME is stored in little-endian order.  But, it's stored "backwards".

This fixes bug report #206.


spiro-trikaliotis:
+1 for this patch.
I am sorry, at first, I commented #208 with this. This was a mistake, I meant this patch here. ;)

oliverschmidt:
Thanks :-)

PR #210, opened 2015-09-25
Make an ar65 warning show both object files' names.

greg-king5:
When an object file is added to an archive, ar65 adds the symbols that are exported.  If any one of those symbols already exists in that archive, then ar65 shows a warning.  That message gives the name of the newly added object file.  But, instead of giving the name of the already existing file, it shows the symbol twice.

This fix makes the message say, "`External symbol '<symbol-name>' in module '<existing file.o>', library '<archive-name>', is duplicated in module '<new file.o>'`".


oliverschmidt:
I was always somehow irritated by the warning but never realized it was actually wrong ;-)

PR #213, opened 2015-09-27
soft80 implementation for C-64

mrdudz:
This is the soft80 stuff we talked about on the ml - again i am putting it up here for review and discussion. Its mostly working, the two major issues left are a) generate the characterset(s) at init time rather than including it as binary (which wastes a lot of space) and b) perhaps hack the putchar function to accept regular cbm character codes (as in cbm.h) for graphical characters.

and i am sure that since this is code from 10 years ago, there is likely room for optimization in both speed and size :)


mrdudz:
ok mostly fixed... by now i am tempted to say that including the charset as is might be a better idea than generating it from the ROM char... since the included one is probably a lot more readable than what some algo would spit out (i vaguely remember creating this one manually back then...)

oliverschmidt:
I'm unfortunately really very tight on time I can spend on cc65 these days :-( However I felt obligated to at least quickly check out the soft80 contribution so here are my findings:

My ultimate goal is - of course ;-) an 80 column Contiki web browser for the C64. So I quickly hacked the Contiki build config files to make use of soft80...

- The build complains with
ld65: Warning: Duplicate external identifier: `SCREEN'

- The soft80 stuff is (much) larger than I had expected - which is however easily because of my wrong expectations. Anyway I had to remove quite some stuff (aka features) from the web browser to get it successfully loaded. I'd really like to not do that (at least to this extent) with the final "product"...
```
c64.lib(soft80_cgetc.o):
    CODE              Offs=008C4B  Size=0000AF  Align=00001  Fill=0000
c64.lib(soft80_chline.o):
    CODE              Offs=008CFA  Size=000018  Align=00001  Fill=0000
c64.lib(soft80_color.o):
    CODE              Offs=008D12  Size=00006E  Align=00001  Fill=0000
    BSS               Offs=0020DE  Size=000002  Align=00001  Fill=0000
c64.lib(soft80_conio.o):
    CODE              Offs=008D80  Size=000073  Align=00001  Fill=0000
    RODATA            Offs=0004D7  Size=000400  Align=00001  Fill=0000
c64.lib(soft80_cputc.o):
    CODE              Offs=008DF3  Size=0003FD  Align=00001  Fill=0000
c64.lib(soft80_cvline.o):
    CODE              Offs=0091F0  Size=00001B  Align=00001  Fill=0000
c64.lib(soft80_kclrscr.o):
    CODE              Offs=00920B  Size=0000A4  Align=00001  Fill=0000
c64.lib(soft80_kplot.o):
    CODE              Offs=0092AF  Size=00002B  Align=00001  Fill=0000
    RODATA            Offs=0008D7  Size=000104  Align=00001  Fill=0000
c64.lib(soft80_kscreen.o):
    CODE              Offs=0092DA  Size=000005  Align=00001  Fill=0000
```

- After getting the web browser loaded it worked out of the box :-)
webbrowser-soft80

- There are some glitches in the display not visible on the screenshot.
  Instead of `>` it displays `}`.
  Sometimes the wrong color is displayed.

From my perspective the results are pretty encouraging, aren't they :-)

mrdudz:
> The build complains with
> ld65: Warning: Duplicate external identifier: `SCREEN'

i noticed that too... but no idea where it comes from

> The soft80 stuff is (much) larger than I had expected - which is however easily because of my wrong expectations.

as always, its a compromise between speed and size (and to some extend, features). when i wrote this thing speed was one somewhat important, since i wanted to use it for both a terminal program (which i never wrote...) and a tetris clone (which i wrote and posted back then....) - it can be made shorter by doing the following:

- generate the charset from the ROM charset at runtime. that will however result in somewhat uglier characters and still require a fair bit of code
- remove the loop unrolling that is present at some places (mostly putchar)... this will obviously make everything noticeable slower and probably save less than 100 bytes or so
- remove the "color voodoo" (mentioned below). that might save a bit of code, and even add some speed - but in return result in uglier output in some cases
- remove special case optimizations (putchar) for space and reverse mode.
- remove color support alltogether (that could shave off a lot of code)

> There are some glitches in the display not visible on the screenshot.
> Instead of > it displays }

ok...mmh. the charset must be changed then, i guess (curly braces do not even exist in regular c64 charset).... i dont remember how i created it back then, so this probably means hacking the binary data :)
> Sometimes the wrong color is displayed.

that is "by design" - or rather, i can not be avoided. in C64 hires mode, each 8x8 cell can use two colors only. one is used for the background color, so there is only one more left for the characters. so if two characters in the same cell contain different colors, one of them will be wrong. that said, the implementation does do some voodoo and remembers what color should actually be there, and then fixes the color when the other character is removed/replaced by a space.

my proposal would be having a second "extra" which has a size optimized version of the implementation that has no color support (or only for one common background/text color). i have no idea how one would do this with the cc65 build system though :)

greg-king5:
>> The build complains with
>> ld65: Warning: Duplicate external identifier: `SCREEN' 
> 
> I noticed that too; but, no idea where it comes from.

It comes from `kernal.s` and `soft80.s`.  `soft80.s` cannot substitute its own `SCREEN` because the original one is exported by a file that exports other names that are needed by I/O code.  You must substitute `screensize`, instead.

> 
>> The soft80 stuff is (much) larger than I had expected -- which is, however, easily because of my wrong expectations.
> 
> 
> As always, it's a compromise between speed and size (and to some extent, features). When I wrote this thing, speed was one somewhat important, since I wanted to use it for both a terminal program (which I never wrote) and a Tetris clone (which I wrote and posted back then) -- it can be made shorter by doing the following:
> 
> * Generate the charset from the ROM charset at runtime. That will, however, result in somewhat uglier characters; and, still require a fair bit of code.
> * . . .

Create a separate "start-up" program that puts that font into RAM, before the main Contiki program is run.  As long as nothing else in Contiki needs to use that "hidden" RAM, the font will stay safe and available for Soft80.

> 
> My proposal would be having a second "extra object file" which has a size-optimized version of the implementation that has no color support (or, only for one common background/text color). I have no idea how one would do that with the cc65 build system though. :-)

Use another prefix:  `soft80_` would be for the faster and feature-filled version; `s80_` would be for the smaller versions of some files.

oliverschmidt:
> Create a separate "start-up" program that puts that font into RAM, before the main Contiki program is run. As long as nothing else in Contiki needs to use that "hidden" RAM, the font will stay safe and available for Soft80.

While that would technically work it would mean that we turn a nicely encapsulated appraoch into sort of a hack ;-)

Maybe you both haven't looked so closely at the Apple II stuff... There I introduced a from my perspective nice way to deal with stuff needed only at initialization time: The INIT segment. The default linker config file(s) place the INIT segment at the very end of the binary created. The address range the INIT segment is loaded to is shared with the BSS segment. The startup code copies the INIT segment right after the BSS segment just before calling zerobss(). This way the content of the INIT segment is technically available until the heap is used and/or the stack grows large. Conceptually the content of the INIT segment is solely used by constructors.

I suggest that this is the right time to introduce just the same on the C64 (and any other machine someone feels the need to do so). All common constructors already live in the INIT segment so one has a benefit right from the start. The only thing to be aware of is that the routine to copy the INIT segment to its final destination needs to be able to cope with overlapping source and destination areas as the INIT segment can easily be larger than the BSS segment.

With that infrastructure in place the soft80 font (and the actual constructor code) are simply to be moved to the INIT segment and everything is fine :-)

mrdudz:
I'd also much prefer this init segment stuff, actually thats the first idea that came to my mind but i wasnt sure if thats a valid scenario in cc65 world :)

> Use another prefix: soft80_ would be for the faster and feature-filled version; s80_ would be for the smaller versions of some files.

as for that, i dont like the idea - as that would result in a fair bit of code duplication... isnt there a way to do this with conditional assembly instead?

oliverschmidt:
>> There are some glitches in the display not visible on the screenshot.
Instead of > it displays }

> ok...mmh. the charset must be changed then, i guess (curly braces do not even exist in regular c64 charset).... i dont remember how i created it back then, so this probably means hacking the binary data :)

Rethinking about it I likely just read the 3 pixel wide `>` as `}` ;-))

>> Sometimes the wrong color is displayed.

> that is "by design" - or rather, i can not be avoided. [...]

Thanks for the explanation which makes sense to me. However I felt that I saw colors not intentionally used by the program at all. I'll check and provide further feedback...

> as always, its a compromise between speed and size (and to some extend, features). when i wrote this thing speed was one somewhat important, since i wanted to use it for both a terminal program (which i never wrote...) and a tetris clone (which i wrote and posted back then....)

I REALLY like the speed. And the web browser does a lot refreshing of the whole client area as anything else would make the code way to complicated aka large. So it really benefits from good conio performance!

Sorry if I turn out to be naive but... could it be that right now not all RAM "behind something" is used by the video screen and the font? In that case I guess one could move code to the RAM left, couldn't one? Maybe even so much that there would be no need to remove features/speed and/or create jet another variant?

oliverschmidt:
> I'd also much prefer this init segment stuff, actually thats the first idea that came to my mind but i wasnt sure if thats a valid scenario in cc65 world :)

:-) Yes, the Apple II show it's pretty valid for cc65. In fact it's a nice usecase for the ld65 linker feature of separate LOAD and RUN addresses.

Talking about code duplication: I haven't looked really closely at the code but but some parts seem to be identical to the stock conio code. I.e. the cvline/chline code...

mrdudz:
> Sorry if I turn out to be naive but... could it be that right now not all RAM "behind something" is used by the video screen and the font? In that case I guess one could move code to the RAM left, couldn't one? Maybe even so much that there would be no need to remove features/speed and/or create yet another variant?

mmmh perhaps - there is 1k left from dc00-dfff under i/o - it'd be tricky to put code there however, since the code has to bank in and out i/o too. we could remove this requirement by moving the temporary color buffer to dc00, but then we can not put code into the color ram that was freed (since its only 4bit RAM). iirc i have opted to use the color ram so that these 1k remain free for eventual sprite data (like a mouse pointer).

i'll first have a look at saving as much as possible using that init segment.... it should remove more than 1k (1k for the charset, and perhaps another page for the constructor code) :)

> Talking about code duplication: I haven't looked really closely at the code but but some parts seem to be identical to the stock conio code. I.e. the cvline/chline code...

correct, like i mentioned in the pce discussion as well - same code *except* for the character code used (regular conio uses screencodes for cputdirect, this soft80 uses petscii since the characterset is organized differently)

oliverschmidt:
> .... same code except for the character code used  ...

If I get this right than this could be solved like we did with the TGI black and TGI white values: The 8-bit values are put in by the linker as "pseudo zeropage locations". See i.e.: https://github.com/cc65/cc65/blob/master/libsrc/c64/tgi_colors.s

oliverschmidt:
> ... eventual sprite data (like a mouse pointer)

The Contiki web browser usually comes with mouse support (it was one of the things I turned of for the current hack to gain memory). Please check out https://github.com/contiki-os/contiki/blob/master/platform/c64/contiki-main.c to see how/where the sprite is placed (the code was c&p'ed from some former cc65 mouse demo program).

mrdudz:
On Thursday 01 October 2015, 06:53:57 Oliver Schmidt 
<notifications@github.com> wrote:
> > .... same code except for the character code used  ...
> 
> If I get this right than this could be solved like we did with the TGI black
> and TGI white values: The 8-bit values are put in by the linker as "pseudo
> zeropage locations". See i.e.:
> https://github.com/cc65/cc65/blob/master/libsrc/c64/tgi_colors.s

seems doable.... lets put that aside and work on general code deduplication 
later in another pull request, there are still a few more things i have laying 
around =)

-- 

My aim is to produce games that have graphics and audio that will rape your 
eyeballs and take your ears from behind with a strap-on 
<Jeff Minter>

mrdudz:
> The Contiki web browser usually comes with mouse support (it was one of the things I turned of for the current hack to gain memory). Please check out https://github.com/contiki-os/contiki/blob/master/platform/c64/contiki-main.c to see how/where the sprite is placed (the code was c&p'ed from some former cc65 mouse demo program).

the few lines in the #if (WITH_GUI && WITH_MOUSE) look like a good candidate for a constructor that can be put into the init segment too =) besides that, to work with the soft80 implementation it should be updated like this:

```C
    asm("sei\nlda #$34\nsta $01"); // disable I/O
    memcpy((void*)0xdc00, mouse_sprite, sizeof(mouse_sprite)); // copy sprite data
    *(uint8_t*)0xdbf8 = (0xdc00 / 64) & 0xff; // setup sprite pointer
    asm("lda #$36\nsta $01\cli"); // enable I/O
```

mrdudz:
another thought: if we make the (c64) conio export a symbol "mousesprite" or something similar that defines where in memory the mousesprite can be located, the mouse-driver code could use that symbol and copy the sprite data to wherever its needed and write the sprite pointers accordingly (since both depends on where the screen memory is located, it makes sense to make it dependent on the conio implementation)

oliverschmidt:
I personally consider it very ugly that a user program needs to fiddle with sprites in order to "just" use a mouse - especially given the general cross target capabilities of cc65. However back when Uz decided he wanted it that way - I don't remember his reasons anymore.

Both the Apple II and the Atari mouse drivers come "of course" with their mouse cursors - and I'd like to have the CBM mouse drivers behave the same.

> the few lines in the #if (WITH_GUI && WITH_MOUSE) look like a good candidate for a constructor that can be put into the init segment too =)

Hm, in my opinion constructors are rather something used in the cc65 code than in user code...

> if we make the (c64) conio export a symbol "mousesprite" or something similar that defines where in memory the mousesprite can be located, the mouse-driver code could use that symbol and copy the sprite data to wherever its needed and write the sprite pointers accordingly

However that export would be only half of the story. The value of that exported symbol would need to be made part of the library reference of CBM mouse drivers in order to allow them to access it.

mrdudz:
> Both the Apple II and the Atari mouse drivers come "of course" with their mouse cursors - and I'd like to have the CBM mouse drivers behave the same.

i agree

> Hm, in my opinion constructors are rather something used in the cc65 code than in user code...

well, ignoring the fact that this code should be part of the library/mouse driver anyway... i can still see the use for constructors in user code. why not? init code that is "fire and forget" isnt that uncommon afterall :)

> However that export would be only half of the story. The value of that exported symbol would need to be made part of the library reference of CBM mouse drivers in order to allow them to access it.

easy enough - the mouse driver would need a constructor (ideally in common code) that imports that symbol, and then copies the sprite data, sets up the sprite pointers, and enables the sprite.

it comes with a price though (which i guess is why Uz didnt do it like that) - when using the mouse driver you'll have to reserve that memory for the pointer, and you'll have to make sure to not change the videobank and screen base address after initializing it (in that case you'll lose the pointer sprite). not sure if it'd be worth the trouble to somehow make the pointer sprite optional....

oliverschmidt:
> i can still see the use for constructors in user code. why not? init code that is "fire and forget" isnt that uncommon afterall :)

Surely, but
* That priority thingy isn't exactly obvious
* On the Apple II the memory mapping while constructors run is different from the memory mapping while main() runs
* People tend to not read documentation ;-)

> easy enough - the mouse driver would need a constructor (ideally in common code) that imports that symbol, and then copies the sprite data, sets up the sprite pointers, and enables the sprite.

Do we have a misunderstanding here? The non-dynamically-loadable part of the mouse driver code (aka the mouse driver kernel) is all common code - and the other targets don't need such a constructor at all.

Apart from that. For sure right now the sprite is initialized on program startup but wouldn't it be more "correct" to initialize it on initialization of the mouse driver? Maybe a program uses the mouse only optionally and maybe it uses the "resource" used by the sprite otherwise when not using the mouse? Maybe sort of hypothetical but nevertheless...

mrdudz:
On Friday 02 October 2015, 04:41:08 Oliver Schmidt <notifications@github.com> 
wrote:
> Surely, but
> * That priority thingy isn't exactly obvious
> * On the Apple II the memory mapping while constructors run is different
> from the memory mapping while main() runs * People tend to not read
> documentation ;-)

none of which is a reason for not using constructors in user code - by people 
who know what they are doing - IMHO ;)

> Do we have a misunderstanding here? The non-dynamically-loadable part of the
> mouse driver code (aka the mouse driver kernel) is all common code - and
> the other targets don't need such a constructor at all.

ok, no, no misunderstanding, just stating the obvious :)

> Apart from that. For sure right now the sprite is initialized on program
> startup but wouldn't it be more "correct" to initialize it on
> initialization of the mouse driver? Maybe a program uses the mouse only
> optionally and maybe it uses the "resource" used by the sprite otherwise
> when not using the mouse? Maybe sort of hypothetical but nevertheless...

thats the "price" i was talking about... if its not in a constructor that runs 
at startup, then we cant take advantage of the "init" segment stuff - and then 
you can just aswell set up the sprite in user code if needed.

-- 

Es ist nie ganz richtig irgend etwas f�r ganz falsch zu halten. 
<Heinz Pol>

oliverschmidt:
> none of which is a reason for not using constructors in user code - by people
who know what they are doing - IMHO ;)

Certainly ;-)

> thats the "price" i was talking about... if its not in a constructor that runs
at startup, then we cant take advantage of the "init" segment stuff - and then
you can just aswell set up the sprite in user code if needed.

While from the memory usage perspective that's true I still think it doesn't belong into user code but that isn't really relevant because...

Rethinking it seems obvious that the mouse cursor isn't an attribute of the mouse driver. It's an attribute of the target (and it's video mode). The exsistence of the driver independent mouse callbacks is a clear indication that Uz shared that perspective. Further one can say that the mouse cursor is an attribute of the mouse callbacks. And from that perspective the puzzle pieces fall into place: The CBM default mouse callbacks (i.e. https://github.com/cc65/cc65/blob/master/libsrc/c64/mcbdefault.s) should have a constructor taking care of the mouse sprite. As a side effect that would allow for custom mouse sprites by "just" not using the default mouse callbacks when initializing the mouse driver...

On a conceptual level the soft80 code could come with it's own default mouse callbacks. But in order to reduce code duplication we should go into the direction already pointed meaning that the default conio and the soft80 conio provide parameter(s) to the (shared) mouse default callbacks.

Looks pretty good - doesn't it?

mrdudz:
On Friday 02 October 2015, 07:44:28 Oliver Schmidt <notifications@github.com> 
wrote:
> On a conceptual level the soft80 code could come with it's own default mouse
> callbacks. But in order to reduce code duplication we should go into the
> direction already pointed meaning that the default conio and the soft80
> conio provide parameter(s) to the (shared) mouse default callbacks.
> 
> Looks pretty good - doesn't it?

yes. so what needs to be done is:
- define a symbol that points to a 64byte block useable for a single sprite
- put the sprite init code into the "show" callback

? :)

-- 

Wenn in Deutschland der Karneval den Karnevalisten gehoert, wer betreibt in 
Oesterreich dann den Fasching?

greg-king5:
Mice must be usable on stdout screens, too.  Therefore, the default symbol pointer should be defined outside of conio.  But then, how do we make sure that it always is available to the `show()` callback -- while still allowing it to be substituted?

Hmm, it is a good use-case for weak exports.

mrdudz:
On Friday 02 October 2015, 09:50:57 greg-king5 <notifications@github.com> 
wrote:
> Mice must be usable on stdout screens, too.  Therefore, the default symbol
> pointer should be defined outside of conio.  But then, how do we make sure
> that it always is available to the `show()` callback -- while still
> allowing it to be substituted?
> 
> Hmm, it is a good use-case for weak exports.

well, there could be a standard symbol defined outside conio (conveniently in 
this case: there is no difference between kernal screen and conio) which is 
then overridden by either conio or soft80

-- 

Due to the current business conditions, the light at the end of the tunnel 
will be switched off until further notice.

oliverschmidt:
> Mice must be usable on stdout screens, too.

Surely.

>Therefore, the default symbol pointer should be defined outside of conio. But then, how do we make sure that it always is available to the show() callback -- while still allowing it to be substituted?

At least from my perspective it was never discussed to tie the pointer to conio. The options from my perspective were:
* Inside the mouse driver(s) (but can't benefit from INIT segment, doesn't really belong there)
* As constructor in the mouse kernel (but the kernel is per se not target specific)
* As constructor in the default mouse callbacks (no downsides)

So far this is about removing the sprite stuff from user code and saving runtime memory by using the INIT segment.

The actual sprite data lives in a separate file exporting just one symbol. This allows the user to export the same symbol thus allowing for pointer substitution. The symbol has (at least) one underscore allowing to define the substitution in C.

Now for the soft80 aspect - I see two viable approaches:
* The constructor doesn't actually live in the callbacks file. Rather that file force-imports the constructor symbol. The actual constructor comes (again) in two variants: The usual one for conio and the soft80 one which is aliased via the extra file. This way the soft80 constructor can take care of the right addresses and memory mapping. The downside is that this again pulls stuff into the binary not necessarily used. And in this case even likely not used. But it's in the INIT segment so it doesn't cost memory at runtime.
* There's only one constructor that lives in the callbacks file. It get's the sprite "target" addresses from imported symbols. It always switches memory banks. But the actual values it puts in $01 are as well imported (zeropage) symbols. There are two files exporting those symbols. The usual one for conio and and the soft80 one which is aliased via the extra file. This way there's nothing unused added to the binary as the exported symbols are just not used when the callbacks file isn't pulled in.

oliverschmidt:
>>> Sometimes the wrong color is displayed.

>> that is "by design" - or rather, i can not be avoided. [...]

> Thanks for the explanation which makes sense to me. However I felt that I saw colors not intentionally used by the program at all. I'll check and provide further feedback...

See the screenshot below. I'm pretty sure the program never puts blue chars on the screen so there's no reason for something to be blue (see the color definitions in https://github.com/contiki-os/contiki/blob/master/platform/c64/contiki-conf.h). I'd say that the soft80 code does not behave as you explained.

colorissue

mrdudz:
urks *shrug* ... a small testcase for this would be great :)

mrdudz:
try again, this might fix it eventually :)

oliverschmidt:
> a small testcase for this would be great :)

Certainly! However I'm afraid I won't be able to come up with one - at least anytime soon :-((

> try again, this might fix it eventually :)

Just did so. But as far as I see it doesn't make a difference. However for sure it doesn't solve the issue. However playing with the web browser it seems that the issue has to do with what is present on the screen when conio overwrites it. As already mentioned the web browser does (too) many redraws. I'd say when the web browser writes something twice then the result is always correct. I don't know if this helps but it's what I can offer now...beside of course providing you with the web browser if that makes sense. It's not that hard to get it working with VICE's RR-Net emulation...

mrdudz:
damn... i could reproduce it here by changing bgcolor() and then printing inverted spaces.... its somehow related to the voodoo that happens when spaces are printed at aleast, i am pretty sure.
if you could zip the source that you used to compile the browser so i can play around with it, perhaps that would help.... (send via mail)

oliverschmidt:
> if you could zip the source that you used to compile the browser so i can play around with it, perhaps that would help.... (send via mail)

Done - good luck on hunting down the bug!

mrdudz:
ok, try another time :) it seems to work ok in the browser now as far as i can see ...

oliverschmidt:
> ok, try another time :)

Thanks for contiguous effort.

> it seems to work ok in the browser now as far as i can see ...

Indeed MUCH better but still not there yet (the blue is gone but the cyan 'n' in 'Main' is still there):
issueii

libsrc/c64/soft80_color.s
```diff
@@ -0,0 +1,109 @@
...
+; unsigned char __fastcall__ bgcolor (unsigned char color);
+; unsigned char __fastcall__ bordercolor (unsigned char color);
+;
+
+        .export         soft80_textcolor, soft80_bgcolor, soft80_bordercolor
+        .export         __textcolor, __bgcolor
+
+        .importzp       tmp1, tmp2
+
+        .import         soft80_checkchar
```
oliverschmidt: As far as I see this line is unnecesarry - and otherwise wouldn't work anyway as there's no export.

libsrc/c64/soft80_cputc.s
```diff
@@ -0,0 +1,463 @@
...
+        ; colram!=bgcolor => alread 2 chars in cell
+@l2a:
+
+        ; Set color
+        lda     CHARCOLOR
+        sta     (CRAM_PTR),y    ; vram
+
+        rts
+
+        ;.export soft80_checkchar
```
oliverschmidt: I dislike commented-out code without a comment providing the reason for the comment-out. As far as I see this line should just be removed.

libsrc/c64/extra/soft80.s
```diff
@@ -0,0 +1,58 @@
...
+
+        ; soft80_kscreen.s
+        .import soft80_screensize
+        .export screensize := soft80_screensize
+
+;-------------------------------------------------------------------------------
+; force the init constructor to be imported
+
+        .import soft80_init
+conio_init      = soft80_init
```
oliverschmidt: I'd rather like to have this file only contain the discussed aliasing.

As far as i see all code directly or indirectly depends on soft80_kplot.s so that would be good place to pull in the constructor from (maybe with a copy in soft80_cgetc.s).

libsrc/c64/soft80_conio.s
```diff
@@ -0,0 +1,230 @@
+;
+; Low level stuff for screen output/console input
+;
+
+        .constructor    soft80_init, 24
```
oliverschmidt: Is there a specific reason for the prio 24 ?

oliverschmidt: Please add this constructor to https://github.com/cc65/wiki/wiki/Priority-of-constructors-and-destructors when this pr is merged.

mrdudz: > Is there a specific reason for the prio 24 ?

no, just copy paste :)

oliverschmidt: I see. Then I suggest to just drop the `, 24` meaning the prio defaults to 7. This is the prio used be CONIO constructors on the Apple II and NES - which seems to be fine.

Does the PCE CONIO contructor prio 24 happen to be also just copy past?

libsrc/c64/extra/soft80.s
mrdudz: ok, just to make sure its always pulled in i put it into _color.s and _scrsize.s as well (not that its terribly useful to call these and not use any of the others, but still...)

libsrc/c64/soft80_conio.s
mrdudz: of course... no idea where that comes from :)

oliverschmidt: Okay, I removed the explicit prios from both PCE constructors (and added them to the Wiki page mantioned).

libsrc/c64/extra/soft80.s
greg-king5: That constructor tells the video chip where to find data, and how to interpret that data.  The constructor is tied to that data; it's tied to the buffers that hold that data.  Therefore, only files that store into those buffers need that constructor.  That means that "soft80_kplot.s" and "soft80_scrsize.s" don't need to force-import it.

In other words, if a function -- directly -- changes what is on the display, then it needs the display to be activated.  If it doesn't -- directly -- then it doesn't need.

mrdudz: but if you call any of those functions you really want the display to be enabled, no? i dont see the point in not calling the constructor as soon as possible if any of the soft80 function is used... it only makes things more complicated. (and as for kplot, i am planning to move the plot tables to ram under i/o as well, then it needs the constructor too)

oliverschmidt:
> the few lines in the #if (WITH_GUI && WITH_MOUSE) look like a good candidate for a constructor that can be put into the init segment too =)

I've taken care of this with https://github.com/cc65/cc65/commit/575f859a030262b60fce9a0a136c54f4b198ca89

> besides that, to work with the soft80 implementation it should be updated like this: [...]

As a next step I'll allow to do that without code duplication...

libsrc/c64/extra/soft80.s
greg-king5: > But, if you call any of those functions, you really want the display to be enabled, no?

No.  Consider this very contrived example:
My program calls `screensize()`, `gotoxy()`, `revers()`, and `textcolor()`; but, it never calls `cgetc()`, `cputc()`, `cputs()`, `cprintf()`, or even `clrscr()`.  It is pointless to activate the display -- we wouldn't see anything!

> And, as for kplot, I am planning to move the plot tables to RAM-under-I/O, as well; then, it needs the constructor, too.

Strictly speaking, that should have its own constructor, inside "soft80_kplot.s".  And, the font should be set by a constructor that is inside "soft80_cputc.s" (the font data should be in a separate file, so that we can substitute it.).  But, I suppose that you can save a few bytes by combining the constructors.

mrdudz: > > But, if you call any of those functions, you really want the display to be
> > enabled, no?
> No.  Consider this very contrived example:
> My program calls `screensize()`, `gotoxy()`, `revers()`, and `textcolor()`;
> but, it never calls `cgetc()`, `cputc()`, `cputs()`, `cprintf()`, or even
> `clrscr()`.  It is pointless to activate the display -- we wouldn't see
> anything!

doh. sure. the whole program would be pointless in the first place :)

-- 

Those who would sacrifice liberty for security deserve neither. 
<Ben Franklin>

oliverschmidt:
One can for sure question if it makes sense to optimize for something as special as Contiki. However when it comes to cc65 programs running out of RAM usually Contiki comes into - at least my - mind.

The question is: How large can the INIT segment become before it actually doesn't save any runtime RAM anymore for a Contiki GUI program like the web browser. The INIT segment overlays the heap and the software stack so let's look at those...

Stack:
Pretty small with Contiki: $200 bytes

Heap:
The heap is not used at all by the Contiki core code. But the cc65 specific code uses the heap for loadable drivers. It loads one mouse driver and one Ethernet driver.
The largest mouse driver: $1D1 bytes
The largest Ethernet driver: $1B5 bytes

So taking some heap management overhead into account the INIT segment shouldn't become larger than $590 bytes.

How much of the INIT segment is already in use without soft80? Building the web browser _without_ soft80 but with my change that puts the mouse sprite stuff into the INIT segment and with Gregs PR which moves some crt0 code into the INIT segment shows: $EC bytes. (Nothing of those has to do with the stock CONIO.)

So rounded down the soft80 library can use up to $4A0 bytes in the INIT segment before increasing the necessary runtime RAM in just the same way it would by using an "ordinary" segment.

As far as I understand the font already uses $400 bytes. So there are only $A0 bytes left for code. Maybe worth to have in mind when thinking about moving more stuff "behind something" - which as far as I understand increases the overall size because of the need to copy it there and some additional banking.

Again: Contiki is just one usecase - but sort of a relevant one ;-)

mrdudz:
try again, its slightly bigger than before now, but it should work properly :)

mrdudz:
On Friday 09 October 2015, 14:29:20 Oliver Schmidt <notifications@github.com> 
wrote:
> So rounded down the soft80 library can use up to $4A0 bytes in the INIT
> segment before increasing the necessary runtime RAM in just the same way it
> would by using an "ordinary" segment.

i dont understand... as i take it, the init segment is completely discarded 
after initial startup, so moving stuff there will always leave more memory 
behind for the rest of the program. whats the catch?

-- 

It doesn't matter who votes. What matters is who counts the votes. 
<Josef Stalin>

oliverschmidt:
> i dont understand... as i take it, the init segment is completely discarded
after initial startup, so moving stuff there will always leave more memory
behind for the rest of the program. whats the catch?

Maybe there are misunderstandings: The memory used by the INIT segment is later used by the BSS. But (at least currently) the idea is to allow the code in the INIT segment to access the BSS. Therefore the content of the INIT segment is copied to the memory "above" the BSS before calling the contructors.

That memory "above" the BSS is used by the heap and the stack. But if heap+stack are smaller than the INIT segment it doesn't fit there. So one has to make it larger just to have the INIT segment fit into it. And by doing so one would loose exactly the amount of memory gained by using the INIT segment.

This would still be okay if it would be 1:1 because other programs with larger heap+stack would still benefit. But if making INIT larger comes with a price (for copy routines and banking routines) than the program with rahter small heap+stack actually suffers from a larger INIT segment. and that is what I see happening with the Contiki programs.

However I'm currently thinking about options to circumvent the isssue. So this whole heads-up posting was maybe to soon...

oliverschmidt:
> try again, its slightly bigger than before now, but it should work properly :)

Done. I can't see an issue anymore. :-))

Just to make sure there's no misunderstanding: It's not that you'd that the web browser uses CONIO in a (too) strange way or alike? And that you had to add "workarounds" for that? Just plain bugs in the soft80 code which are fixed now - correct?

mrdudz:
> Just to make sure there's no misunderstanding: It's not that you'd that the web browser uses CONIO in a (too) strange way or alike? And that you had to add "workarounds" for that? Just plain bugs in the soft80 code which are fixed now - correct?

yes. previously the "color voodoo" was really just that... the code didnt really care of working correct in all cases... now it does :)

oliverschmidt:
> the code didnt really care of working correct in all cases... now it does :)

Cool !

mrdudz:
On Saturday 10 October 2015, 04:02:37 Oliver Schmidt 
<notifications@github.com> wrote:
> > i dont understand... as i take it, the init segment is completely
> > discarded
> 
> after initial startup, so moving stuff there will always leave more memory
> behind for the rest of the program. whats the catch?
> 
> Maybe there are misunderstandings: The memory used by the INIT segment is
> later used by the BSS. But (at least currently) the idea is to allow the
> code in the INIT segment to access the BSS. Therefore the content of the
> INIT segment is copied to the memory "above" the BSS before calling the
> contructors.
> 
> That memory "above" the BSS is used by the heap and the stack. But if
> heap+stack are smaller than the INIT segment it doesn't fit there. So one
> has to make it larger just to have the INIT segment fit into it. And by
> doing so one would loose exactly the amount of memory gained by using the
> INIT segment.
> 
> This would still be okay if it would be 1:1 because other programs with
> larger heap+stack would still benefit. But if making INIT larger comes with
> a price (for copy routines and banking routines) than the program with
> rahter small heap+stack actually suffers from a larger INIT segment. and
> that is what I see happening with the Contiki programs.
> 
> However I'm currently thinking about options to circumvent the isssue. So
> this whole heads-up posting was maybe to soon...

thats indeed not what i thought it would work like - and it doesnt seem like a 
good idea to me either (it is wasting a lot of potential and limits the 
usefulness of the init segment quite a bit).

what i would do is:
- as a convention, define that any code in the .init segment can not use the 
.bss section, except for temporary storage that is discarded after that code 
has run.
- put any variables that the init code initializes that were in bss before 
into another segment (which could be just "data" or perhaps a seperate 
"initbss")

that shouldnt have any drawbacks, and since most stuff in bss in a larger 
program is probably not initialized by the init code, allow to use most of the 
bss for init code, you'd even save the copy code for it :)

oliverschmidt:
So far the INIT segment was only used by pretty small code fragments initializing global variables - usually in BSS. A usual INIT segment size was $80 - $180 bytes. From that perspective the current design was valid - at least from my point of view.

> what i would do is: [...]

What you describe are exactly two of the options I have in mind :-)
- no BSS access at all
- something like "INITBSS" (to be merged with the already exsisting ZPSAVE)

Just give me a little time to evaluate the pro and cons of the different options...

oliverschmidt:
>> besides that, to work with the soft80 implementation it should be updated like this: [...]

> As a next step I'll allow to do that without code duplication...

Done. See https://github.com/cc65/cc65/commit/08efc299ffacac0e13296a33e50ef64be99358e3

oliverschmidt:
In order to have VIC initialized properly before the mouse sprite initialization uses it, the soft80 conio constructor needs to have a higher priority than the default of 7. Please adjust it to 8.

mrdudz:
ok so ... cant think of what else to do now except doing the init segment stuff :)

oliverschmidt:
>> If I get this right than this could be solved like we did with the TGI black
 and TGI white values: The 8-bit values are put in by the linker as "pseudo
 zeropage locations". See i.e.:
 https://github.com/cc65/cc65/blob/master/libsrc/c64/tgi_colors.s

> seems doable.... lets put that aside and work on general code deduplication
later in another pull request, there are still a few more things i have laying
around =)

I analyzed the soft80 files and as far as I can see only the chline.s and cvline.s actually fall into the category. I already prepared everything with https://github.com/cc65/cc65/commit/5e960d106e050abe01616618907a95dbe8bf7dd0 so it should be very little effort to make use of it. So please remove the duplicated chline and cvline implementations.

oliverschmidt:
> ok so ... cant think of what else to do now except doing the init segment stuff :)

* It would be nice if you could check if the mouse stuff actually works as designed with samples/mousetest.c

* I'm sure by now that the way forward is - as discussed - to allow the INIT segment to grow in one way or another. So it's save if you invest effort in moving "as much as possible" into INIT.

mrdudz:
> It would be nice if you could check if the mouse stuff actually works as designed with samples/mousetest.c

works for me :)

> I'm sure by now that the way forward is - as discussed - to allow the INIT segment to grow in one way or another. So it's save if you invest effort in moving "as much as possible" into INIT.

OK

mrdudz:
ok, there... waiting for the cfg magic ... :)

oliverschmidt:
>> It would be nice if you could check if the mouse stuff actually works as designed with samples/mousetest.c

> works for me :)

Cool. So as far as I can see this PR is "feature complete" :-)

I'm pretty curious to see where we'll end up with respect to memory consumption...

mrdudz:
yes, curious myself now :) played with the source you sent the other night... at least i can build a mouse enabled CTK which even works .... but need to disable DNS because of memory - lets see what the init stuff brings.

oliverschmidt:
> at least i can build a mouse enabled CTK which even works ....

:-)

libsrc/c64/soft80_conio.s
```diff
@@ -0,0 +1,166 @@
...
+soft80_init:
+        lda     soft80_first_init
+        bne     @skp
+        jsr     firstinit
+@skp:
+        ; the "color voodoo" in other parts of the code relies on the vram and
+        ; colorram being set up as expected, which is why we cant use the
+        ; _bgcolor and _textcolor functions here.
+
+        lda     646                             ; use current textcolor
```
greg-king5: "c64.inc" defines a label for that location.

libsrc/c64/soft80_bordercolor.s
```diff
@@ -0,0 +1,17 @@
...
+; unsigned char __fastcall__ bordercolor (unsigned char color);
+;
+
+        .export         soft80_bordercolor
+
+        .include        "c64.inc"
+
+        ; FIXME: if we'd move this function into a seperate file in the regular
+        ;        conio lib, then we dont need this override at all.
+soft80_bordercolor:
```
greg-king5: Soft80 is the inspiration for separating `bordercolor()` from the other color functions.  Therefore, I believe that you should do it in this Pull Request.

mrdudz:
ok so some more refinement :) tempted to do an optimized no-color version now... :)

oliverschmidt:
> tempted to do an optimized no-color version now... :)

Maybe it makes sense to first check how things look like with respect to memory consumption.

I decided to introduce an INITBSS segment to be used by the constructors. As you can see from https://github.com/cc65/cc65/commit/0ee9b2e446198746c3a05b142ecd00784becf727 and https://github.com/cc65/cc65/commit/76a5a72403266b923b9bdc9ccd5f4de2a352459a this turns out to be a pretty substantial effort. While usage of INITBSS is generally still in a rather inconsistent state I think that the c64 target is complete.

Note: As far as I understand the current Contiki code doesn't work with my changes as it contains a constructor that uses the BSS segment - so far for user coding with constructors ;-)

I'll try to move theses things forward asap but I'm already way over my time budget...

mrdudz:
I quickly hacked contikis pfs.S to use .data instead, and then it works just fine.... so it shouldnt take much of an effort :) (i could even set MTU=1000 now...)
BTW - when the soft80 conio is used, the regular screen memory is now free for whatever else (1k afterall, $0400-$07ff) ... perhaps the soft80 init should add it to the heap? *shrug*

oliverschmidt:
> I quickly hacked contikis pfs.S to use .data instead, and then it works just fine....

Good to know that it works :-))

> i could even set MTU=1000 now...

With or without DNS? With or without mouse?

> when the soft80 conio is used, the regular screen memory is now free for whatever else (1k afterall, $0400-$07ff)

That would have been one of my next questions.

> perhaps the soft80 init should add it to the heap?

I would rather not do so. Apps not using the heap at all shouldn't have the heap logic pulled in imho. Could soft80 make use of it for moving "stuff" there - i.e. code?

Otherwise I'd leave it up the the application to add it to the heap. It would however be nice to have that documented. See i.e.: http://cc65.github.io/doc/apple2.html#ss7.1

With respect to Contiki adding the screen to the heap should work out just fine. The two drivers should fit there and use the area pretty well:
The largest mouse driver: $1D1 bytes
The largest Ethernet driver: $1B5 bytes

As you play with Contiki you might want to check it out...

mrdudz:
> With or without DNS? With or without mouse?

including both :)

> Could soft80 make use of it for moving "stuff" there - i.e. code?

mmh perhaps... will check.

however, a more "c64ish" solution would be to just let .code start at $0400 ... the resulting program would typically be packed by some cruncher then so you can run it from basic. oh well, and then we'd finally be at the point that we create a new target =)

and yes, regarding contiki i was looking at the drivers as well

oliverschmidt:
> including both :)

Ah, then loading the drivers into the screen should allow to revert everything back to the original state and remove the MTU limitation altogether (the default is 1500). That would be pretty cool :-)

> let .code start at $0400

As you surely presume by now I have no idea what it takes to make that happen. First stupid question: Wouldn't it "hurt" if the "de-cruncher" writes the program to the screen before the soft80 initializer reprogrammed the VIC. Or would that "just" mean some short flicker?

mrdudz:
On Thursday 15 October 2015, 07:11:44 Oliver Schmidt 
<notifications@github.com> wrote:
> > including both :)
> 
> Ah, then loading the drivers into the screen should allow to revert
> everything back to the original state and remove the MTU limitation
> altogether (the default is 1500). That would be pretty cool :-)
> > let .code start at $0400
> 
> As you surely presume by now I have no idea what it takes to make that
> happen. First stupid question: Wouldn't it "hurt" if the "de-cruncher"
> writes the program to the screen before the soft80 initializer reprogrammed
> the VIC. Or would that "just" mean some short flicker?

you'd get some garbage on screen for a short time - as you can see with many 
cracked games, for example :)

-- 

There are 10**11 stars in the galaxy. That used to be a huge number. But it's 
only a hundred billion. It's less than the national deficit! We used to call 
them astronomical numbers. Now we should call them economical numbers. 
<Richard Feynman>

oliverschmidt:
> we'd finally be at the point that we create a new target =)

As said I'm not strictly against that. However maybe a more modular approach would be superior. Like having support for a cruncher that - as side effect - allows to set the start address freely. And let the user combine that if desired with soft80...

mrdudz:
> Ah, then loading the drivers into the screen should allow to revert everything back to the original state and remove the MTU limitation altogether (the default is 1500). That would be pretty cool :-)

indeed, just adding that area to the heap at start of contiki-main.c does the trick :)

oliverschmidt:
> indeed, just adding that area to the heap at start of contiki-main.c does the trick :)

Nice :-))

I'll try to have that "productized" asap. Obviously there should be one PRG with and one PRG without soft80. And the same for the IRC client too...

mrdudz:
eh, so here is it anyway... saves roughly 1k over the full featured version. just incase... :)

greg-king5:
Files that import `soft80*conio.s` variables don't need _to force_ an import of the constructors.

In my opinion, the files that don't import any of those variables really don't need the constructors.  Therefore, I think that the "forcers" should be removed from the Soft80 files.

oliverschmidt:
> eh, so here is it anyway... saves roughly 1k over the full featured version. just incase... :)

I see. Do you have a feeling for the performance side of things. Would it make sense for a program not suffering from RAM shortage to go for the monochrome variant to get noticable faster conio?

oliverschmidt:
I certianly don't have the necessary overview but I'd like to point out just in case that if there were color and monochrome files with signficant and similiar content then (following DRY) it might make sense to have two wrapper .s files just defining a macro in one or the other way and then including a shared .inc file with the actual code using `.if` or `.ifdef`to do the needful.

mrdudz:
> > eh, so here is it anyway... saves roughly 1k over the full featured
> > version. just incase... :)
> I see. Do you have a feeling for the performance side of things. Would it
> make sense for a program not suffering from RAM shortage to go for the
> monochrome variant to get noticable faster conio?

i have not tried it, my guess would be that its about the same speed as the 
other.

doc/c64.sgml
```diff
@@ -116,6 +116,40 @@ cl65 -o file.prg -u __EXEHDR__ -t c64 -C c64-asm.cfg source.s
...
+
+The C64 package comes with an alternative software driven 80 columns
+module <tt/c64-soft80.o/ which uses the memory under I/O between &dollar;d000
+and &dollar;ffff.
+
+In memory constrained situations the memory from &dollar;400 to &dollar;7FF
+can be made available to a program by calling <tt/_heapadd ((void *) 0x400, 0x400);/
+at the beginning of <tt/main()/. Doing so is beneficial even if the program
+doesn't use the the heap explicitly because loading a driver (and in fact
+already opening a driver file) uses the heap implicitly.
```
oliverschmidt: The `(and in fact already opening a driver file)`is specific to the Apple II file I/O and not true here.

doc/c64.sgml
```diff
@@ -216,6 +250,9 @@ configuration.
   palette of the 16 C64 colors).
 </descrip><p>
 
+Note that the graphics drivers are incompatible with the
+<tt/c64-ram.emd (c64_ram_emd)/ extended memory driver and the
+ <tt/c64-soft80.o/ software 80 columns conio driver.
```
oliverschmidt: Just for curiosity: Is the way the TGI driver programs the VIC incompatible with the way soft80 does? Or would it - at least conceptually - be possible to use soft80 and TGI together?

mrdudz: > > @@ -216,6 +250,9 @@ configuration.
> > 
> >    palette of the 16 C64 colors).
> >  
> >  </descrip><p>
> > 
> > +Note that the graphics drivers are incompatible with the
> > +<tt/c64-ram.emd (c64_ram_emd)/ extended memory driver and the
> > + <tt/c64-soft80.o/ software 80 columns conio driver.
> 
> Just for curiosity: Is the way the TGI driver programs the VIC incompatible
> with the way soft80 does? Or would it - at least conceptually - be possible
> to use soft80 and TGI together?

it shouldnt be terribly hard to make both work together... later, perhaps :=)

oliverschmidt: >> [...] use soft80 and TGI together?

> it shouldnt be terribly hard to make both work together... later, perhaps :=)

I see. I wasn't suggesting this. I just wanted to find out if I understood it right that from the VIC perspective both things are the same...

oliverschmidt:
@mrdudz: Just to avoid a communication deadlock...
* Are you considering to share code between the color and monochrome variant with the include-approach I pointed out in a recent comment?
* Are you willing to adjust the documentation as suggested by me in a recent comment?

mrdudz:
> Are you considering to share code between the color and monochrome variant with the include-approach I pointed out in a recent comment?

in this case i think it might be a better idea to have seperate files... the code is often very different and using ifdefs wont save enough to justify the added confusion and less readability, IMHO :)

> Are you willing to adjust the documentation as suggested by me in a recent comment?
done

oliverschmidt:
> in this case i think it might be a better idea to have seperate files... the code is often very different and using ifdefs wont save enough to justify the added confusion and less readability, IMHO :)

I see.

So I think we're ready to have this merged :-)

oliverschmidt:
> I'll try to have that "productized" asap. Obviously there should be one PRG with and one PRG without soft80. And the same for the IRC client too...

Done. See https://github.com/contiki-os/contiki/pull/1343

oliverschmidt:
@mrdudz: With https://github.com/mrdudz/cc65/commit/5ea842b88b7e799826f33ec9b8c80d4a0091bebe you removed the duplicate export of `SCREEN` as suggested by Greg. However with https://github.com/mrdudz/cc65/commit/e6f5c7d8da65d1d249fc6a636ed668a4c229eac0 you brough it back (and with it the original issue). Please elaborate on why you did so.

oliverschmidt:
@mrdudz: Given that clrscr.s contains `_clrscr = CLRSCR` I don't understand why you do both `.export _clrscr := soft80_kclrscr` and `.export CLRSCR := soft80_kclrscr`.

mrdudz:
> @mrdudz: With mrdudz@5ea842b you removed the duplicate export of SCREEN as suggested by Greg. However with mrdudz@e6f5c7d you brough it back (and with it the original issue). Please elaborate on why you did so.

i don't remember. perhaps because once you want to replace soft80_scrsize by the regular one (ie remove it) as discussed earlier, it has to be done that way.

i don't see any "issue" right now though, at least no warning is shown that indicates that there is one (when compiling the conio test anyways)

i commented it out now for the time being...

> @mrdudz: Given that clrscr.s contains _clrscr = CLRSCR I don't understand why you do both .export _clrscr := soft80_kclrscr and .export CLRSCR := soft80_kclrscr.

that one just slipped through... removed

new pull request coming up... :)

oliverschmidt:
> i don't see any "issue" right now though, at least no warning is shown that indicates that there is one (when compiling the conio test anyways)

Then there's something wrong with your setup. I see
```
ld65: Warning: Duplicate external identifier: `SCREEN'
```
both when building testcode/conio.c and building the Contiki programs.

PR #214, opened 2015-09-28
Make the CBM exec() work in gigantic programs.

greg-king5:
Contiki's web browser should be able to run the wget program.


oliverschmidt:
Great - works nicely !

Thanks :-)

PR #215, opened 2015-10-05
Save space by using the apple2 target's INIT overlay scheme on the c64 target.

greg-king5:
This Pull Request implements the INIT segment scheme that Oliver urged us (in PR #213 comments) to use.

The c64 configuration files have other changes that come from my project to make the configure files more versatile and more consistent with each other.  I put them in this PR because they work well with the INIT changes.

I plan to do the INIT scheme on most of the other targets.  I publish these c64 changes so that Soft80 and Contiki can take advantage of them now.

### DIFF
--- a/cfg/c64-overlay.cfg
+++ b/cfg/c64-overlay.cfg
@@ -1,64 +1,70 @@
+FEATURES {
+    STARTADDRESS:    default = $0801;
+}
 SYMBOLS {
     __LOADADDR__:    type = import;
     __EXEHDR__:      type = import;
     __OVERLAYADDR__: type = import;
-    __STACKSIZE__:   type = weak, value = $0800; # 2k stack
-    __OVERLAYSIZE__: type = weak, value = $1000; # 4k overlay
+    __STACKSIZE__:   type = weak,   value = $0800; # 2k stack
+    __OVERLAYSIZE__: type = weak,   value = $1000; # 4k overlay
+    __HIMEM__:       type = weak,   value = $D000;
+    __HIMEM2__:      type = export, value = __HIMEM__ - 2;
 }
 MEMORY {
-    ZP:       file = "", define = yes, start = $0002,                   size = $001A;
-    LOADADDR: file = %O,               start = $07FF,                   size = $0002;
-    HEADER:   file = %O,               start = $0801,                   size = $000C;
-    RAM:      file = %O, define = yes, start = $080D,                   size = $C7F3 - __OVERLAYSIZE__ - __STACKSIZE__;
-    OVL1ADDR: file = "%O.1",           start = $CFFE - __OVERLAYSIZE__, size = $0002;
-    OVL1:     file = "%O.1",           start = $D000 - __OVERLAYSIZE__, size = __OVERLAYSIZE__;
-    OVL2ADDR: file = "%O.2",           start = $CFFE - __OVERLAYSIZE__, size = $0002;
-    OVL2:     file = "%O.2",           start = $D000 - __OVERLAYSIZE__, size = __OVERLAYSIZE__;
-    OVL3ADDR: file = "%O.3",           start = $CFFE - __OVERLAYSIZE__, size = $0002;
-    OVL3:     file = "%O.3",           start = $D000 - __OVERLAYSIZE__, size = __OVERLAYSIZE__;
-    OVL4ADDR: file = "%O.4",           start = $CFFE - __OVERLAYSIZE__, size = $0002;
-    OVL4:     file = "%O.4",           start = $D000 - __OVERLAYSIZE__, size = __OVERLAYSIZE__;
-    OVL5ADDR: file = "%O.5",           start = $CFFE - __OVERLAYSIZE__, size = $0002;
-    OVL5:     file = "%O.5",           start = $D000 - __OVERLAYSIZE__, size = __OVERLAYSIZE__;
-    OVL6ADDR: file = "%O.6",           start = $CFFE - __OVERLAYSIZE__, size = $0002;
-    OVL6:     file = "%O.6",           start = $D000 - __OVERLAYSIZE__, size = __OVERLAYSIZE__;
-    OVL7ADDR: file = "%O.7",           start = $CFFE - __OVERLAYSIZE__, size = $0002;
-    OVL7:     file = "%O.7",           start = $D000 - __OVERLAYSIZE__, size = __OVERLAYSIZE__;
-    OVL8ADDR: file = "%O.8",           start = $CFFE - __OVERLAYSIZE__, size = $0002;
-    OVL8:     file = "%O.8",           start = $D000 - __OVERLAYSIZE__, size = __OVERLAYSIZE__;
-    OVL9ADDR: file = "%O.9",           start = $CFFE - __OVERLAYSIZE__, size = $0002;
-    OVL9:     file = "%O.9",           start = $D000 - __OVERLAYSIZE__, size = __OVERLAYSIZE__;
+    ZP:       file = "", define = yes, start = $0002,                        size = $001A;
+    LOADADDR: file = %O,               start = %S - 2,                       size = $0002;
+    HEADER:   file = %O, define = yes, start = %S,                           size = $000D;
+    RAM:      file = %O,               start = __HEADER_LAST__,              size = __HIMEM__ - __OVERLAYSIZE__ - __STACKSIZE__ - __HEADER_LAST__;
+    MOVE:     file = %O,               start = __ZPSAVE_LOAD__,              size = __HIMEM__ - __INIT_RUN__;
+    OVL1ADDR: file = "%O.1",           start = __HIMEM2__ - __OVERLAYSIZE__, size = $0002;
+    OVL1:     file = "%O.1",           start = __HIMEM__  - __OVERLAYSIZE__, size = __OVERLAYSIZE__;
+    OVL2ADDR: file = "%O.2",           start = __HIMEM2__ - __OVERLAYSIZE__, size = $0002;
+    OVL2:     file = "%O.2",           start = __HIMEM__  - __OVERLAYSIZE__, size = __OVERLAYSIZE__;
+    OVL3ADDR: file = "%O.3",           start = __HIMEM2__ - __OVERLAYSIZE__, size = $0002;
+    OVL3:     file = "%O.3",           start = __HIMEM__  - __OVERLAYSIZE__, size = __OVERLAYSIZE__;
+    OVL4ADDR: file = "%O.4",           start = __HIMEM2__ - __OVERLAYSIZE__, size = $0002;
+    OVL4:     file = "%O.4",           start = __HIMEM__  - __OVERLAYSIZE__, size = __OVERLAYSIZE__;
+    OVL5ADDR: file = "%O.5",           start = __HIMEM2__ - __OVERLAYSIZE__, size = $0002;
+    OVL5:     file = "%O.5",           start = __HIMEM__  - __OVERLAYSIZE__, size = __OVERLAYSIZE__;
+    OVL6ADDR: file = "%O.6",           start = __HIMEM2__ - __OVERLAYSIZE__, size = $0002;
+    OVL6:     file = "%O.6",           start = __HIMEM__  - __OVERLAYSIZE__, size = __OVERLAYSIZE__;
+    OVL7ADDR: file = "%O.7",           start = __HIMEM2__ - __OVERLAYSIZE__, size = $0002;
+    OVL7:     file = "%O.7",           start = __HIMEM__  - __OVERLAYSIZE__, size = __OVERLAYSIZE__;
+    OVL8ADDR: file = "%O.8",           start = __HIMEM2__ - __OVERLAYSIZE__, size = $0002;
+    OVL8:     file = "%O.8",           start = __HIMEM__  - __OVERLAYSIZE__, size = __OVERLAYSIZE__;
+    OVL9ADDR: file = "%O.9",           start = __HIMEM2__ - __OVERLAYSIZE__, size = $0002;
+    OVL9:     file = "%O.9",           start = __HIMEM__  - __OVERLAYSIZE__, size = __OVERLAYSIZE__;
 }
 SEGMENTS {
-    LOADADDR: load = LOADADDR, type = ro;
-    EXEHDR:   load = HEADER,   type = ro;
-    STARTUP:  load = RAM,      type = ro;
-    LOWCODE:  load = RAM,      type = ro,                optional = yes;
-    INIT:     load = RAM,      type = ro,  define = yes, optional = yes;
-    CODE:     load = RAM,      type = ro;
-    RODATA:   load = RAM,      type = ro;
-    DATA:     load = RAM,      type = rw;
-    ZPSAVE:   load = RAM,      type = bss;
-    BSS:      load = RAM,      type = bss, define = yes;
-    ZEROPAGE: load = ZP,       type = zp;
-    OVL1ADDR: load = OVL1ADDR, type = ro;
-    OVERLAY1: load = OVL1,     type = ro,  define = yes, optional = yes;
-    OVL2ADDR: load = OVL2ADDR, type = ro;
-    OVERLAY2: load = OVL2,     type = ro,  define = yes, optional = yes;
-    OVL3ADDR: load = OVL3ADDR, type = ro;
-    OVERLAY3: load = OVL3,     type = ro,  define = yes, optional = yes;
-    OVL4ADDR: load = OVL4ADDR, type = ro;
-    OVERLAY4: load = OVL4,     type = ro,  define = yes, optional = yes;
-    OVL5ADDR: load = OVL5ADDR, type = ro;
-    OVERLAY5: load = OVL5,     type = ro,  define = yes, optional = yes;
-    OVL6ADDR: load = OVL6ADDR, type = ro;
-    OVERLAY6: load = OVL6,     type = ro,  define = yes, optional = yes;
-    OVL7ADDR: load = OVL7ADDR, type = ro;
-    OVERLAY7: load = OVL7,     type = ro,  define = yes, optional = yes;
-    OVL8ADDR: load = OVL8ADDR, type = ro;
-    OVERLAY8: load = OVL8,     type = ro,  define = yes, optional = yes;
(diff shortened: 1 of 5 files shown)


oliverschmidt:
@greg-king5: Thanks a lot for taking care of this :-) As I felt that I couldn't let you do all the work after all I started to implement this myself but - as always recently - i didn't find enough time to finish it...

It would be - of course - great if you could actually do the INIT thingy for the other targets too as you indicated.

I have a few line comments...

oliverschmidt:
> The c64 configuration files have other changes that come from my project to make the configure files more versatile and more consistent with each other.

Yes, I remember that we discussed that you wanted to replace more "magic constants" with computed values...

cfg/c64-overlay.cfg
```diff
@@ -1,64 +1,70 @@
...
+    LOADADDR: load = LOADADDR,        type = ro;
+    EXEHDR:   load = HEADER,          type = ro;
+    STARTUP:  load = RAM,             type = ro;
+    LOWCODE:  load = RAM,             type = ro,                optional = yes;
+    CODE:     load = RAM,             type = ro;
+    RODATA:   load = RAM,             type = ro;
+    DATA:     load = RAM,             type = rw;
+    ZPSAVE:   load = RAM,             type = bss, define = yes;
+    BSS:      load = RAM,             type = bss, define = yes;
+    INIT:     load = MOVE, run = RAM, type = ro,  define = yes, optional = yes;
```
oliverschmidt: Given that the startup code itself uses the INIT segment it doesn't have to be `optional`. (Yes: This is "wrong" in the Apple config files too). Another side effect is that the routine to copy the INIT segment doesn't need to take care of an empty segment.

cfg/c64-overlay.cfg
```diff
@@ -1,64 +1,70 @@
...
-    OVL7:     file = "%O.7",           start = $D000 - __OVERLAYSIZE__, size = __OVERLAYSIZE__;
-    OVL8ADDR: file = "%O.8",           start = $CFFE - __OVERLAYSIZE__, size = $0002;
-    OVL8:     file = "%O.8",           start = $D000 - __OVERLAYSIZE__, size = __OVERLAYSIZE__;
-    OVL9ADDR: file = "%O.9",           start = $CFFE - __OVERLAYSIZE__, size = $0002;
-    OVL9:     file = "%O.9",           start = $D000 - __OVERLAYSIZE__, size = __OVERLAYSIZE__;
+    ZP:       file = "", define = yes, start = $0002,                        size = $001A;
+    LOADADDR: file = %O,               start = %S - 2,                       size = $0002;
+    HEADER:   file = %O, define = yes, start = %S,                           size = $000D;
+    RAM:      file = %O,               start = __HEADER_LAST__,              size = __HIMEM__ - __OVERLAYSIZE__ - __STACKSIZE__ - __HEADER_LAST__;
+    MOVE:     file = %O,               start = __ZPSAVE_LOAD__,              size = __HIMEM__ - __INIT_RUN__;
```
oliverschmidt: Technically the `start` of the MOVE memory area doesn't matter. When thinking what to put in on the Apple I considered it hard to decide between the "load start" and the "run start". So finally I just set it to $0. I'm however fine with your approach too. Maybe it would be nice to adjust the Apple config files when bringing the INIT handling to the other targets too for consistency reasons...

cfg/c64.cfg
```diff
@@ -1,26 +1,31 @@
...
 MEMORY {
-    ZP:       file = "", define = yes, start = $0002, size = $001A;
-    LOADADDR: file = %O,               start = $07FF, size = $0002;
-    HEADER:   file = %O,               start = $0801, size = $000C;
-    RAM:      file = %O, define = yes, start = $080D, size = $C7F3 - __STACKSIZE__;
+    ZP:       file = "", define = yes, start = $0002,           size = $001A;
+    LOADADDR: file = %O,               start = %S - 2,          size = $0002;
+    HEADER:   file = %O, define = yes, start = %S,              size = $000D;
+    RAM:      file = %O,               start = __HEADER_LAST__, size = __HIMEM__ - __STACKSIZE__ - __HEADER_LAST__;
+    MOVE:     file = %O,               start = __ZPSAVE_LOAD__, size = __HIMEM__ - __INIT_RUN__;
```
oliverschmidt: See above.

cfg/c64.cfg
```diff
@@ -1,26 +1,31 @@
...
+    LOADADDR: load = LOADADDR,        type = ro;
+    EXEHDR:   load = HEADER,          type = ro;
+    STARTUP:  load = RAM,             type = ro;
+    LOWCODE:  load = RAM,             type = ro,                optional = yes;
+    CODE:     load = RAM,             type = ro;
+    RODATA:   load = RAM,             type = ro;
+    DATA:     load = RAM,             type = rw;
+    ZPSAVE:   load = RAM,             type = bss, define = yes;
+    BSS:      load = RAM,             type = bss, define = yes;
+    INIT:     load = MOVE, run = RAM, type = ro,  define = yes, optional = yes;
```
oliverschmidt: See above.

libsrc/c64/crt0.s
```diff
@@ -3,13 +3,13 @@
...
 
         .export         _exit
+        .exportzp       init_load_, init_run_
         .export         __STARTUP__ : absolute = 1      ; Mark as startup
+
         .import         initlib, donelib
-        .import         zerobss
-        .import         callmain
+        .import         move_init, zerobss, callmain
         .import         RESTOR, BSOUT, CLRCH
```
oliverschmidt: As you seem to do a general overhauling of this file I suggest to remove the unrefenced `RESTOR` and `CLRCH`.

libsrc/c64/crt0.s
```diff
@@ -19,17 +19,16 @@
 ; ------------------------------------------------------------------------
 ; Startup code
 
-.segment        "STARTUP"
```
oliverschmidt: Yet another leftover that I noticed too ;-)

libsrc/c64/crt0.s
```diff
@@ -19,17 +19,16 @@
...
+; Two zero-page pointers are needed before any zero-page stuff is saved.
+; Choose locations that are not used by anything.
 
-Start:
```
oliverschmidt: Yet another leftover I noticed too ;-)

libsrc/c64/crt0.s
```diff
@@ -39,35 +38,30 @@ L1:     lda     sp,x
...
-
         tsx
         stx     spsave          ; Save the system stack ptr
 
-        lda     #<(__RAM_START__ + __RAM_SIZE__ + __STACKSIZE__)
-        sta     sp
-        lda     #>(__RAM_START__ + __RAM_SIZE__ + __STACKSIZE__)
-        sta     sp+1            ; Set argument stack ptr
+; Allow some re-entrancy by skipping the next task if it already was done.
+; This often can let us rerun the program without reloading it.
```
oliverschmidt: I don't know if you remember that there was some time ago a discussion about reRUNing CBM programs. I was pretty surprised back then that even Uz considered this desirable - while I personally don't do at all for the well known reason of wrong content in the DATA segment.
I would have easily dropped reRUNabilty when introducing the INIT handling - and several users would have been pretty unhappy. So your approach seems way more diplomatic ;-))

libsrc/c64/crt0.s
```diff
@@ -39,35 +38,30 @@ L1:     lda     sp,x
...
-        jsr     initlib
+; Move the INIT segment from where it was loaded (over ZPSAVE and BSS)
+; into where it must be run (in the heap).
 
-; Push the command-line arguments; and, call main().
+        jsr     move_init
+        dec     moveinit        ; set to false
+
+; Save space by putting the rest of the start-up code in the INIT segment,
+; which can be re-used by the heap.
```
oliverschmidt: ...re-used by the heap (or the software stack).

libsrc/common/moveinit.s
```diff
@@ -0,0 +1,54 @@
...
+
+; Move the INIT segment from where it was loaded (over the bss segments)
+; into where it must be run (in the heap).  The two areas might overlap; and,
+; the segment is moved upwards.  Therefore, this code starts at the highest
+; address, and decrements to the lowest address.  The low bytes of the starting
+; pointers are not sums.  The high bytes are sums; but, they do not include the
+; carry.  Both the low-byte sums and the carries will be done when the pointers
+; are indexed by the .Y register.
+
+move_init:
```
oliverschmidt: Just a matter of taste: `moveinit` would fit better to `callinit`, `callmain`, `zerobss` and `copydata`.

libsrc/common/moveinit.s
```diff
@@ -0,0 +1,54 @@
...
+        lda     #<__INIT_LOAD__
+        ldx     #>__INIT_LOAD__ + >__INIT_SIZE__
+        sta     init_load_
+        stx     init_load_+1
+        lda     #<__INIT_RUN__
+        ldx     #>__INIT_RUN__ + >__INIT_SIZE__
+        sta     init_run_
+        stx     init_run_+1
```
oliverschmidt: My partial implementation of the move routine looks (of course) pretty much like yours. Howeve I thought that it is used exactly once and therefore it might be appropriate to put it in the DATA segment and use selfmodifying code instead of indirect adressing in order to save space. This approach would basically remove the code block above...

libsrc/common/moveinit.s
```diff
@@ -0,0 +1,54 @@
...
+        jmp     L3
+.endif
+
+L1:     dec     init_load_+1
+        dec     init_run_+1
+
+L2:     dey
+        lda     (init_load_),y
+        sta     (init_run_),y
+        tya
```
oliverschmidt: Yep, I couldn't find a way circumvent this "dummy transfer just to load the Z flag" either ;-)

libsrc/common/moveinit.s
```diff
@@ -0,0 +1,54 @@
...
+; First, move the last, partial page.
+; Then, move all of the full pages.
+
+        ldx     #>__INIT_SIZE__ + 1     ; number of pages, including partial
+        ldy     #<__INIT_SIZE__         ; size of partial page
+.if     .cpu & CPU_ISET_65SC02
+        bra     L3
+.else
+        jmp     L3
+.endif
```
oliverschmidt: If you load `#>__INIT_SIZE__ + 1` last than you can simply use a `bne` on all CPUs.

libsrc/c64/crt0.s
mrdudz: srsly, just drop it :)

libsrc/common/moveinit.s
```diff
@@ -0,0 +1,54 @@
...
+        dec     init_run_+1
+
+L2:     dey
+        lda     (init_load_),y
+        sta     (init_run_),y
+        tya
+L3:     bnz     L2                      ; page not finished
+
+        dex
+        bnz     L1                      ; move next page
```
mrdudz: should we really use these unusual/non standard mnemonics in library code? IMHO they make it harder to read and are confusing at best.

oliverschmidt: I'm personally fully with you. I never ever use "mnemonic type macros". No combos like `add` or `ldax`, no renames like `bnz`.

But if you check the cc65 code base (i.e. with https://github.com/cc65/cc65/search?utf8=%E2%9C%93&q=bnz) you'll notice that this pull request doesn't exactly introduce `bnz`.

mrdudz: maybe - but then these (few!) occurrences should get fixed, IMHO =)

cfg/c64-overlay.cfg
greg-king5: >  Maybe, it would be nice to adjust the Apple config files when bringing the INIT handling to the other targets too, for consistency reasons.

OK.

libsrc/c64/crt0.s
greg-king5: I very often rerun programs.  I will load a program, run it on a directory/disk, then change to another directory/disk, rerun, then change yet again to another directory/disk, and rerun.  (The DATA segment issue seldom causes trouble for me.)  The ability to rerun programs is so important to me that I want to add more code to the libraries, that will make it easier (the stdio destructors do cause trouble; they change the default device number).

libsrc/common/moveinit.s
greg-king5: I added `bze` and `bnz` because I was sick and tired of my thoughts being jerked to a halt by the misuse of `beq` and `bne`.  When I saw code like this:

```gas
   dex
   bne label
```

I constantly had to think, "Huh?  There's no comparison here.  What's happenning?  Oh, now I remember:  `bne` is used also to test for zero conditions.  Bah!  I wish that there is a mnemonic that just said what that code really is doing."
Now, there are those mnemonics.

By the way, I am not the only one who thought it.  Western Design Center's programmers put `blt` and `bge` into their assembler.  I just finished the job.

I believe that everyone should learn them, get used to them, and make them standard and usual.  They make code _less_ confusing -- in my opinion.

libsrc/c64/crt0.s
oliverschmidt: You see: Uz did a lot with cc65 and considered it important. You do a lot with cc65 and consider it important. I already moved away from my original perspective to actively prohibit it (by patching in an RTS as first byte or alike). But I don't see me merging pull requests which spend more than a handful of bytes for it.

greg-king5:
OK, I have done those suggestions.

I shrank "moveinit.s" a little more than I had expected that I could do it.  I changed some run-time code into link-time expressions.

libsrc/c64/crt0.s
```diff
@@ -98,14 +85,51 @@ L2:     lda     zpsave,x
...
+
+
+; ------------------------------------------------------------------------
+; Data
+
+.data
 
-spsave: .res    1
 mmusave:.res    1
+spsave: .res    1
```
oliverschmidt: Sorry for not commenting on this in the first place but... I guess we agree that we don't like `.res` in data segments don't we? I always "feel" that `.res` should only be used in bss segments. Anyway just for consistency reasons the three variables should be defined in the same way - or am I missing a point here?

oliverschmidt:
> OK, I have done those suggestions.

Thanks, the config files look nice now :-)

libsrc/common/moveinit.s
oliverschmidt: I've been thinking about this once more. I must admit that I don't follow your argument here. I don't see the point in "inventing a new assembly language" for programming the 6502 after so many years. As you know I appreciate your contributions very much and therefore I won't try to keep you from using those macros in code you contribute. But I'm not convinced that they are desirable.

libsrc/c64/crt0.s
greg-king5: > I guess we agree that we don't like `.res` in data segments.

We don't agree about that.  Look at the end of https://github.com/cc65/cc65/blob/master/libsrc/apple2/mainargs.s
`argv` is defined as

```gas
.data
argv:   .addr   FNAM
        .res    MAXARGS * 2
```

The `.res` in the third line is a sign to readers that the initial values in those locations are irrelevant.  Those variables become important only after some piece of code has put something into them.

The value in `move_init` is important even before the program runs.
But, `mmusave` and `spsave` are significant only after registers are saved into them.  I think that we should use `.res` to show that.

mrdudz: i agree with oliver there - if the initial value is irrelevant, why not put it into .bss in the first place? (i'd even go that far as emitting a warning in such case...)

greg-king5: Those variables have had a roundabout journey.  They started in DATA; then, they were moved to BSS; and now, I have moved them back to DATA.  I did it because the code that stores into them can be smaller if it is done early, before INIT is moved off of BSS.

oliverschmidt: Okay, trying to sort this out...

> The .res in the third line is a sign to readers that the initial values in those locations are irrelevant.

Sorry, but I just can't follow you here!

The semantics of variables not relying on implicit initialization either by being loaded from disk or via zerobss() just isn't something usually expressed in neither assembly nor C. While I can see that it might make sense to have ways to express that...
* I don't see us changing the language we're working with. After all this is a community project and others should be at least be able to read the source code.
* According to http://cc65.github.io/doc/ca65.html#ss11.91 `.res` by no means results in the variable having an undefined initial value. Rather it's initialized with zero. So even someone would try to understand what you want to express he'd have no chance doing so when reading the `.res` docs.

Result: The desired cc65 style is to use `.res` in the BSS and `.byte` (and friends) for DATA and RODATA. I'd appreciate if you would adjust your code.

> `argv` is defined as [...]

What is the reason to not place it in the BSS - with the obvious benefit of reducing the file size?

In contrast `mmusave` and `spsave` are from a semantic perspective BSS type variables but have to live somewhere else for technical reasons - they are used before BSS is cleared. With that perspective in mind I can see the point of `.res` for them but I'd vote for a comment stating why they can't live in BSS. After all it's obviously not that obvious ;-)

oliverschmidt:
Thanks.

libsrc/c64/crt0.s
oliverschmidt: Just for clarification:

If we have a data structure that is partially initialized with something non-zero - and which therefore lives in the DATA segment - then it is of course an option to use `.res` to initialize the rest of the data structure with zeros. Otherwise you might have to write lots of `0` literals into the source code. An example:

```
.data
argv:   .addr   name
        .res    MAXARGS * 2
```

In the example below however we wouldn't have multiple `0` literals:

```
.data
ctrl:   .addr   _read
        .res    2                       ; CALLERDATA
        .res    2                       ; MODULE
        .res    2                       ; MODULE_SIZE
        .res    2                       ; MODULE_ID
```

I guess Uz and Greg see a parallel to the unnecessary explicit initialization of global (or static) C variables with zero (although that technically happens in the BSS segment). You see:

```
int my_global = 0;
```
vs.
```
int my_global;
```

In the same way the code above doesn't emphasize that `my_global` is initialized (although it is anyway) a `.res 2 my_global` doesn't emphasize that `my_global` is initialized (although it is anyway).

While I believe I understand the argument I don't follow it for two reasons:

* I always just write `int my_global;` even _IF_ the code depends on the implicit initialization - after all that's what it's for.

* The `.res` approach looses the benefits of 16-bit types. I consider the code below more readable:

```
.data
ctrl:   .addr   _read
        .addr   0                       ; CALLERDATA
        .addr   0                       ; MODULE
        .word   0                       ; MODULE_SIZE
        .word   0                       ; MODULE_ID
```

PR #222, opened 2015-10-22
use tii instead of copy loop to copy data section to ram

mrdudz:
... as requested :)


oliverschmidt:
Nice :-)

As already discussed (?) it might make sense to provide target-specific implementations for `memcpy` (and `memmove` ?) based on `tii`...

PR #224, opened 2015-10-23
Change the way that ld65 handles bad offset/start segment-attributes.

greg-king5:
This Pull Request fixes a bug that was reported by Brad Smith on [the old cc65 mailing list](http://www.cc65.org/mailarchive/2015-10/11761.html).


oliverschmidt:
Thanks :-)

PR #225, opened 2015-11-02
Handle the special case of a zero-length move on the PC Engine.

greg-king5:
This fixes the code that was added by Pull Request #223.

I had forgotten that the PC Engine CPU's transfer instructions cannot copy a zero amount of bytes.  A length operand of $0000 actually means $10000 (64K).


oliverschmidt:
I see :-)

PR #226, opened 2015-11-07
Add a PC-Engine version of memset().

greg-king5:
This version of the library function `memset()` is like `memcpy()` and `memmove()`:  it takes advantage of enhanced instructions in the PC-Engine CPU.

### DIFF
--- a/libsrc/pce/memcpy.s
+++ b/libsrc/pce/memcpy.s
@@ -35,25 +35,25 @@ _memcpy:
         jsr     memcpy_getparams
 
 memcpy_increment:
-        ldy     #$73                    ; TII
+        ldy     #$73                    ; TII opcode
 
 memcpy_transfer:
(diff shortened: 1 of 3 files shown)


libsrc/pce/memcpy.s
```diff
@@ -34,25 +35,25 @@ _memcpy:
...
 memcpy_transfer:
-        sty     transfer+opcode
+        sty     transfer + opcode
 
         lda     ptr1
         ldx     ptr1+1
-        sta     transfer+source
-        stx     transfer+source+1
+        sta     transfer + source
+        stx     transfer + source+1
```
oliverschmidt: Yep, that's exactly the style I personally like: In general blanks around the operators with the exception of 16-bit hibyte access "+1"s.

oliverschmidt:
Cool !

Issue #218, opened 2015-10-05
Possible compiler bug with pointer variable vs direct pointer


clbr:
This file:
```
#include <nes.h>

int main() {

        unsigned char *ppuaddr = (unsigned char *) 0x2006;
        unsigned char *ppudata = (unsigned char *) 0x2007;
        unsigned char *ppuctrl = (unsigned char *) 0x2000;
        unsigned char *ppumask = (unsigned char *) 0x2001;
        unsigned char *ppuscroll = (unsigned char *) 0x2005;

        waitvblank();

        *((unsigned char*)0x2006) = 0x3F;
        *((unsigned char*)0x2006) = 0x00;
//      *((unsigned char*)0x2007) = 1;
        *ppudata = 1; // uncomment the above and comment this -> it works

        *ppuscroll = 0;
        *ppuscroll = 0;

        *ppumask = 8;

        while (1);
        return 0;
}
```

fails when run - the background color is grey. If I comment the *ppudata assignment and uncomment the direct pointer assignment, it works - the bg color is blue.

I don't speak enough asm to say why the pointer-var fails and direct pointer works. Here is the diff of the assemblies:
```
--- wrong.s     2015-10-05 22:33:49.000000000 +0300
+++ right.s     2015-10-05 22:33:43.000000000 +0300
@@ -46,13 +46,9 @@
        ldx     #$00
        lda     #$00
        sta     $2006
-       ldy     #$07
-       jsr     ldaxysp
-       jsr     pushax
        ldx     #$00
        lda     #$01
-       ldy     #$00
-       jsr     staspidx
+       sta     $2007
        ldy     #$01
        jsr     ldaxysp
        jsr     pushax

```

clbr:
Enabling optimizations, either at -O or -Oisr level, does not change anything. Right still works, wrong still fails even on full optimization, even if pointers are declared as "unsigned char * const", at which level the compiler should be able to optimize them to direct calls.

Still, a missing optimization doesn't explain the failure.

oliverschmidt:
Rather independent from the question of a potential compiler bug you're much better of with the direct assignment. `ppudata` and friends are pointer variables while ` *((unsigned char*)0x2007)`and friends are pointer constants. Using variables is in general slower/larger than using constants. Think of
```
int x = a;
```
versus
```
int x = 1;
```
[...]

PR #229, opened 2015-11-12
Update the <nes.h> include file.

greg-king5:
As I mentioned in issue #218, `<nes.h>` now has the chip declarations.  Therefore, people like @clbr should have much less trouble using direct NES I/O in a C program.


oliverschmidt:
Nice :-)

Issue #228, opened 2015-11-10
Recursive macro issue in ca65


furrykef:
I tried modifying one of the push macros in the examples by adding a fourth argument:

    .macro  push    r1, r2, r3, r4
            lda     r1
            pha
    .if     .paramcount > 1
            push    r2, r3, r4
    .endif
    .endmacro

    push 1, 2, 3, 4

While the original three-argument version worked fine, this version fails spectacularly with a ton of error messages about an illegal addressing mode. I have no idea what the problem is.

The .ifblank/.exitmacro equivalent of this macro works fine, by the way; it's only the .paramcount version that doesn't work.

greg-king5:
`.paramcount` cannot be used in recursive macroes because it counts empty (blank) arguments, too.  That example, in the ca65 document, works only because of a bug in .paramcount's implementation (the last argument isn't counted if it's only a comma [`,{}` is counted]).

I will fix that bug.  And, I will fix that document; it won't say that `.paramcount` can be used in recursive macroes.

PR #232, opened 2015-11-17
Fix ca65's ".paramcount" Assembly-code read-only variable.

greg-king5:
This PR resolves issue #228.

* `.paramcount` will count all of a macro invocation's arguments -- including an empty one at the end of the line.
* The ca65 document doesn't mislead us about `.paramcount`; we won't think that it can be used in recursive macroes.


oliverschmidt:
Thanks :-)

PR #233, opened 2015-11-22
reset screen editor at shutdown

mrdudz:
there.... this restores default colors at shutdown

### DIFF
--- a/libsrc/c64/soft80_conio.s
+++ b/libsrc/c64/soft80_conio.s
@@ -51,13 +51,9 @@ soft80_init:
         jmp     soft80_kclrscr
 
 soft80_shutdown:
-        lda     #$1b
-        sta     VIC_CTRL1
-        lda     #$03
-        sta     CIA2_PRA
-        lda     #$15
-        sta     VIC_VIDEO_ADR
-        rts
+
+        jsr     $fda3   ; Initialise I/O
+        jmp     $ff5b   ; Initialize screen editor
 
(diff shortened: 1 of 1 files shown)


oliverschmidt:
Thanks - and it's even shorter :-)

libsrc/c64/soft80_conio.s
```diff
@@ -51,13 +51,9 @@ soft80_init:
...
-        lda     #$1b
-        sta     VIC_CTRL1
-        lda     #$03
-        sta     CIA2_PRA
-        lda     #$15
-        sta     VIC_VIDEO_ADR
-        rts
+
+        jsr     $fda3   ; Initialise I/O
+        jmp     $ff5b   ; Initialize screen editor
```
greg-king5: That first subroutine call cannot be used because it re-enables the BASIC ROM.  If a large program (such as Contiki's `webbrowser80`) has destructor code or data "behind" that ROM, then the program might crash when it tries to quit gracefully.  Changing that code to the following works well enough:

~~~gas
        lda     #$07
        sta     CIA2_PRA
        jmp     $FF5B           ; Initialize video I/O
~~~

oliverschmidt: I followed your suggestion with https://github.com/cc65/cc65/commit/377f31d08527060d9e2ef6efff6e4c7a058f5dd4 and it works like a charm :-)

Thanks for looking into this issue !

Issue #230, opened 2015-11-12
ca65 dependency file generation doesn't escape spaces in paths


fo-fo:
Ca65 dependency file generation doesn't escape spaces in paths, causing ambiguity:

    ; test2.s ("test test.h" is an empty file)
    .include "test test.h"

Command:

    cl65 --create-dep test2-dep.txt test2.s

Result (test2-dep.txt):

    test2.o:	test2.s test test.h

    test2.s test test.h:

The C compiler does do it properly (by escaping with "\", similarly to how gcc does it):

    // test1.c ("test test.h" is an empty file)
    #include "test test.h"

Command:

    cl65 --create-dep test1-dep.txt test1.c

Result (test1-dep.txt):

    test1.o:	test1.c test\ test.h

    test1.c test\ test.h:

oliverschmidt:
* I agree (of course) that both the compiler and the assembler should behave the same.
* However I personally don't consider this issue as important as the dependency files created are supposed to be consumed by make. And make just doesn't work with files having spaces - although some people tend to think otherwise.

fo-fo:
Some considerations:

- GCC escapes the spaces, so that feels like an acceptable default behavior.
- The dependency files can also be consumed by other tools, like Ninja. (I'm pretty sure I've had success with spaces in paths with Make as well, but maybe there are some hidden problems?)
- I never personally use spaces in source code filenames, but sometimes the files come from a 3rd party source and renaming them would be a hassle.
- Current behavior is "generate a broken file and don't warn about it"

greg-king5:
In my experience, make doesn't support escapes.  In most places, apostrophes, backslashes, and quotation marks are read and used verbatim.  So, your example has three dependencies:

1. `test1.c`
2. `test\`
3. `test.h`

Sometimes, we can replace the space with the question mark wildcard character; but, it isn't a generic solution.

fo-fo:
Would you accept a patch that implements the escaping?

PR #236, opened 2015-11-26
Escaping of spaces in ca65 dependency files

fo-fo:
WRT #230

WriteEscaped() was copied from cc65's input.c, I didn't attempt to DRY because there seemed to be a fair amount of copy&paste going on between the two files (cc65 input.c and ca65 filetab.c) as it is.


oliverschmidt:

PR #238, opened 2015-11-29
Bit Corporation 'Gamate' support

mrdudz:
what the topic says... yes its obscure :)


doc/gamate.sgml
```diff
@@ -0,0 +1,159 @@
...
+<item>fscanf
+<item>fwrite
+<item>...
+</itemize>
+
+<sect>Other hints<p>
+
+<itemize>
+<item>The Gamate is emulated by MESS (<url url="http://www.mess.org/">),
+run like this: <tt>mess gamate -debug -window -skip_gameinfo -cart test.bin</tt>
```
oliverschmidt: I was about to ask for such a hint - and there it is already :-)

include/gamate.h
```diff
@@ -0,0 +1,200 @@
...
+#define ZP_IRQ_CNT3     0x10
+#define ZP_IRQ_CNT4     0x11
+
+#define ZP_NMI_FLAG     0xe8    /* set to 0xff each NMI */
+
+/* constants for the conio implementation */
+#define COLOR_BLACK     0x03
+#define COLOR_WHITE     0x00
+
+#define CH_HLINE                1
```
oliverschmidt: Please unify the indentation of the macro value with the rest of the file (same for the following lines).

libsrc/gamate/cputc.s
```diff
@@ -0,0 +1,141 @@
...
+
+putchar:
+        sta     ptr3
+
+        txa
+        pha
+
+        lda     #0
+        sta     ptr3+1
+        ; * 8
```
oliverschmidt: Please have a black line before comment-only lines like this.

testcode/lib/conio.c
```diff
@@ -14,6 +14,14 @@
 #include <string.h>
 #include <stdlib.h>
 
+#if defined(__GAMATE__)
+#define NUMCHARS        128
```
oliverschmidt: Please elaborate why this is necessary.

testcode/lib/conio.c
```diff
@@ -120,6 +131,10 @@ void main(void)
                     cputc(i);
                     inpos = (inpos + 1) & 7;
                 }
+#endif
+#if defined(__NES__) || defined(__PCE__) || defined(__GAMATE__)
+                waitvblank();
```
oliverschmidt: Sorry for my ignorance but what happens if this is missing?

oliverschmidt:
I guess samples/hello.c line 71 needs adjustment too.

oliverschmidt:
Such a substantial pull request deserves more review than I'm able to provide right now. Please be prepared to wait in case the usual suspect doesn't come to the rescue.

libsrc/gamate/cputc.s
mrdudz: a WHAT?! *shrug*

oliverschmidt: ;-))

"black" was of course meant to read "blank"...

testcode/lib/conio.c
mrdudz: "something" must limit the speed the inner loop is running at, else the revers printed char(s) do not blink but just flicker irritatingly fast :)

doc/gamate.sgml
mrdudz: what the hint doesnt say: gamate support only was added very recently to MESS - you might night a more recent build than the last stable release. you also need a proper gamate bios (which doesnt come with MESS).

include/gamate.h
```diff
@@ -0,0 +1,200 @@
...
+/* locations 0x0a-0x0c, 0x0e-0x11 and 0xe8 are in use by the BIOS IRQ/NMI handlers */
+#define ZP_NMI_4800     0x0a    /* content of I/O reg 4800 gets copied here each NMI */
+
+#define ZP_IRQ_COUNT    0x0b    /* increments once per IRQ, used elsewhere in the
+                                   BIOS for synchronisation purposes */
+#define ZP_IRQ_CTRL     0x0c    /* if 0 then cartridge irq stubs will not get called */
+
+/* each of the following 4 increments by 1 per IRQ - it is _not_ a 32bit
+   counter (see code at $ffa6 in BIOS)
+   these are not used elsewhere in the bios and can be (re)setted as needed by
```
silverdr: "set" falls into "irregular" category. Meaning "setted" -> "set"
set 1 |sɛt|
verb (sets, setting; past and past participle: set)

doc/gamate.sgml
oliverschmidt: Okay, I start to understand what you mean with "obscure".

One of the non-written rules for new target in cc65@GitHub is that there's a reasonable free emulator supporting the target allowing "everybody" to verify it's functional correctness.

With (a non-release) MESS supporting Gamate I see that fullfilled - and as far as I understand it's not unusual for MESS to not include necessary ROMs (usually of IP reasons).

testcode/lib/gamate/nachtm.c
```diff
@@ -0,0 +1,1147 @@
...
+        cputsxy (X, T->Y, T->Msg);
+        ++T;
+    }
+}
+
+void UpdateScreen(void)
+{
+    register VoiceCtrl* VC;
+    /* Play all three voices */
+    for (I
```
oliverschmidt: I'm certainly willing to be less style nit-picking for test code but I still think one source file should have at least a consistent style - and this isn't the case in this file regarding blanks before actual parameter lists.

testcode/lib/conio.c
```diff
@@ -14,6 +14,15 @@
 #include <string.h>
 #include <stdlib.h>
 
+#if defined(__GAMATE__)
+/* there is not enough screen space to show all 256 characters at the bottom */
```
oliverschmidt: I see - so this isn't a limitation of the Gamate CONIO implementation (supporting only chars  < 128) but rather specific to this test program.

testcode/lib/conio.c
```diff
@@ -120,6 +132,11 @@ void main(void)
                     cputc(i);
                     inpos = (inpos + 1) & 7;
                 }
+#endif
+/* not all targets have waitvblank() */
```
oliverschmidt: Maybe I'm missing the point but the porgram seems to be intended to run through the loop only once after user interaction. And in that scenario there's just no need for waitvblank() - idenpendetly from the fact if it's available or not.

Wouldn't it be preferable to replace the CONIO input with some joystick input on the targets in question - simliar to hello.c ?

libsrc/gamate/vga.inc
```diff
@@ -0,0 +1,228 @@
+
+; VGA charset for the PC-Engine conio implementation
```
oliverschmidt: Seems like a C&P leftover.

BTW: Was the term 'VGA' correct for the PC-Engine in the first place?

libsrc/gamate/vga.inc
```diff
@@ -0,0 +1,228 @@
...
+
+        .byte   $00, $00, $00, $00, $00, $00, $00, $00
+        ; 1
```
oliverschmidt: Please add black (aka blank) lines before the comments.

BTW: I guess libsrc/pce/vga.inc would benefit from the comments too.

testcode/lib/gamate/nachtm.c
mrdudz: i agree - and you should fix the code style guide to whatever is preferred :o)

libsrc/gamate/vga.inc
```diff
@@ -0,0 +1,228 @@
...
+        .byte   %00010000
+        .byte   %00010000
+        .byte   %00010000
+        .byte   %00011111
+        .byte   %00010000
+        .byte   %00010000
+        .byte   %00010000
+        .byte   %00010000
+
+        .byte   $3F, $33, $3F, $30, $30, $70, $F0, $E0
```
oliverschmidt: Sorry for no asking this for the PC-Engine in the first place, but wouldn't it be nice to have all chars defined using binary constants (%) like done above? Maybe it would be actually worth the effort to hack a simple converter and use it to process the rest of this file - what do you think?

I don't know if there are other scenarios (maybe C64 soft80?) where large piles of constants could be rendered actually comprehensible by using (nicely formatted) binary constants?

testcode/lib/conio.c
mrdudz: yes, IF you add joystick input (which doesnt make a lot of sense here =P) AND dont check with kbhit() inside the loop (as it should have done from the start in the first place), then what you say is true :)

libsrc/gamate/vga.inc
mrdudz: to be honest, i dont quite remember where the "VGA" part comes from... iirc its from some old VGA Bios, but that could also be wrong :)

libsrc/gamate/vga.inc
mrdudz: nice... sure. perhaps. but it wont be me who is putting a lot of effort into it when i dont need it :) (IMHO the proper solution to all the charsets that are in various parts of the lib is having them in eg .png format plus some kind of converter program - which then could be used by users to replace the default chars too)

libsrc/gamate/vga.inc
oliverschmidt: Makes sense...

libsrc/gamate/vga.inc
oliverschmidt: I see...

testcode/lib/conio.c
oliverschmidt: Okay, I added joystick input (just incrementing n). I don't understand your remark regarding kbhit() as it wasn't present AND I don't see a reason to add it. It should now be possible to operate the (from my perspective not that useful) program as intended without a keyboard.

mrdudz:
since this is a rather "huge" PR i request additional review - in particular a good reason for why it should'nt be merged right now (i don't see any).

oliverschmidt:
As you noticed ;-) I've been looking through the open pull requests the other day. I don't see myself capable of investing mor effort into an in-deepth review anytime soon. On the other hand I see no known issue with this pull request and - not to forget - I presume that you are willing and able to provide additional fixes / improvements if they should turn out to be necessary. As a result I agree with you that it is appropriate to merge this pull request now.

mrdudz:
ok, so i will merge it - for things like that (which dont have the potential to break either the compiler or other core components like the runtime) i much prefer early merging and then fixing in the main tree when necessary.

oliverschmidt:
I see your point. However I suggest even for those "non-breaking" things to:

- still use the pull request mechanism as it fosters discussion - which I consider good.
- allow others to comment for a certain amount of time (I suggest a week) and only merge if there's no (reasonable negative) feedback at the end of the time.

Does that make sense to you?

oliverschmidt:
Regarding the Gamate:

I see that you already took care of https://github.com/cc65/cc65/blob/gh-pages/index.html :-)

https://github.com/cc65/cc65/blob/master/README.md is replicated to

- https://github.com/cc65/wiki/blob/master/README.md
- https://github.com/cc65/mailarchive/blob/master/README.md
- https://github.com/cc65/doc/blob/master/README.md

Please update the one in the Wiki repo. I'll take care of the other too...

mrdudz:
[quote]Does that make sense to you?[/quote]
sure, i wont push anything directly unless its a trivial change (codestyle or such)
[quote]I see that you already took care of https://github.com/cc65/cc65/blob/gh-pages/index.html[/quote]
that actually wasnt supposed to go to the main repo directly either - but git decided to use it as default for the gh-pages branch anyway for some reason (trying to fix that right now.... with little success so far =P)
i'll have a look at the wiki...

PR #240, opened 2015-12-01
Fixed the c128 bank 1 emd.

blackystardust:
These fixes make the c128 bank 1 emd work again.


oliverschmidt:
I'd like to mention that https://github.com/cc65/wiki/wiki/Reminders#C128_library contains
```
Test EM drivers with the multidemo sample; and, fix the broken ones.
```
so fixes in this area are appreciated.

blackystardust:
Just for the people who do not know, I'm the 'leader' of the VICE project, and I'm in the process of adding test programs for the various memory expansions, joysticks, mice and lightpen that are supported in VICE.

Groepaz suggested I do it in cc65, that way it benefits both projects, so I'm testing them, fixing where needed and adding new drivers where needed.

oliverschmidt:
> Just for the people who do not know, I'm the 'leader' of the VICE project

Thanks for the background - i.e. I didn't know ;-)

> [...] adding test programs for the various memory expansions, joysticks, mice and lightpen that are supported in VICE [...] Groepaz suggested I do it in cc65 [...]

So you want those test programs to become part of cc65? Then I suggest to place them in `testcode/lib`. If they are specific to CBM and/or C64 please create corresponding subdirectories `cbm` and/or `c64`.

> so I'm testing them, fixing where needed and adding new drivers where needed.

Great :-)

Just a short note on the pull request review process: You may be wondering why there's - maybe even lenghty - discussion on small aspects. Especially for regular contributors (and what you write makes me think that you see yourself as becoming one) the idea is that pull requests can be merged as-is without any post processing on style and other aspects. In fact I usually merge from the GitHub web UI.

mrdudz:
[quote]So you want those test programs to become part of cc65?[/quote]
the testprogrogs itself live in the VICE testprog repo (-> https://sourceforge.net/p/vice-emu/code/HEAD/tree/testprogs/ )... but the (new) drivers would have to go into cc65 :)

oliverschmidt:
> the testprogrogs itself live in the VICE testprog repo

I see - and it makes perfect sense to me. I was just wondering because of the recent changes to `testcode/lib/em-test.c`.

> but the (new) drivers would have to go into cc65

:-)

blackystardust:
The changes in em-test.c are so you can easily choose which emd you want to test, makes more sense to me than having to change the source every time you want to change a different emd meant for the same 'platform',

So in this case a generic change that benefits both projects.

oliverschmidt:
Maybe a misunderstanding: I'm totally fine with your changes to em-test.c! I was just wondering why you took interest in changing it when you don't use it as part of your VICE tests.

blackystardust:
Oh, I/we do, I recently added a .d64 with the test program and all the emd's, that way we can test for any regressions in the ram expansions emulation code.

It's just that VICE seems to be 'ahead' of driver implementations in cc65, from my/our point of view there are a lot of 'missing' drivers in cc65, not just emd, but also joystick, mice and light pen wise.

So, I'll wind up adding them ;)

greg-king5:
> It's just that VICE seems to be "ahead" of driver implementations in cc65; from my/our point of view, there are a lot of "missing" drivers in cc65 -- not just emd, but also joystick-, mice-, and lightpen-wise.
>
> So, I'll wind up adding them. :wink:

Keep in mind that the mouse API needs to be expanded with the button-mapping feature from the joystick API (so that drivers can describe middle and "scroll" buttons).

PR #241, opened 2015-12-01
C128 function ram emd

blackystardust:
Added c128 internal/external function ram emd's and added items to the em-test.c struct.

### DIFF
--- a/asminc/c128.inc
+++ b/asminc/c128.inc
@@ -204,8 +204,8 @@ MMU_CFG_RAM0     := %00111111    ; Bank 0 full RAM
 MMU_CFG_RAM1     := %01111111    ; Bank 1 full RAM
 MMU_CFG_RAM2     := %10111111    ; Bank 2 full RAM
 MMU_CFG_RAM3     := %11111111    ; Bank 3 full RAM
-MMU_CFG_INT_FROM := %00010111    ; Bank 0 with Internal Function RAM/ROM
-MMU_CFG_EXT_FROM := %00101011    ; Bank 0 with External Function RAM/ROM
+MMU_CFG_INT_FROM := %01010111    ; Bank 1 with Internal Function RAM/ROM
+MMU_CFG_EXT_FROM := %01101011    ; Bank 1 with External Function RAM/ROM
 
 ; ---------------------------------------------------------------------------
 ; Super CPU
--- a/asminc/c128.inc
+++ b/asminc/c128.inc
@@ -198,14 +198,14 @@ CIA2_CRB        := $DD0F
 ; ---------------------------------------------------------------------------
 ; I/O: MMU
 
-MMU_CR           := $FF00
-MMU_CFG_CC65     := %00001110    ; Bank 0 with kernal ROM
-MMU_CFG_RAM0     := %00111111    ; Bank 0 full RAM
-MMU_CFG_RAM1     := %01111111    ; Bank 1 full RAM
-MMU_CFG_RAM2     := %10111111    ; Bank 2 full RAM
-MMU_CFG_RAM3     := %11111111    ; Bank 3 full RAM
-MMU_CFG_INT_FROM := %01010111    ; Bank 1 with Internal Function RAM/ROM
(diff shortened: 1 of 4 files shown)


libsrc/c128/emd/c128-efnram.s
```diff
@@ -0,0 +1,326 @@
...
+
+; ------------------------------------------------------------------------
+; INSTALL routine. Is called after the driver is loaded into memory. If
+; possible, check if the hardware is present and determine the amount of
+; memory available.
+; Must return an EM_ERR_xx code in a/x.
+;
+
+INSTALL:
+        sei
```
greg-king5: Interrupts should be disabled as late as possible; and, re-enabled as early as possible (i.e., they should be blocked for as short a time as possible).

libsrc/c128/emd/c128-efnram.s
```diff
@@ -0,0 +1,326 @@
...
+        ldx     #0
+        stx     ptr1
+        ldx     #$80
+        stx     ptr1+1
+        ldx     #<ptr1
+        stx     FETVEC
+        stx     STAVEC
+        ldy     #0
+        ldx     #MMU_CFG_EXT_FROM
+        jsr     FETCH
```
greg-king5: Just before that `jsr FETCH`
seems like a good place to put the `sei`.

libsrc/c128/emd/c128-efnram.s
```diff
@@ -0,0 +1,326 @@
...
+        tax
+        inx
+        txa
+        sta     tmp1
+        ldx     #MMU_CFG_EXT_FROM
+        jsr     STASH
+        ldx     #MMU_CFG_EXT_FROM
+        jsr     FETCH
```
greg-king5: Just after that `jsr FETCH`
seems like a good place to put the `cli`.

libsrc/c128/emd/c128-efnram.s
blackystardust: In that case a lot of other emd code will need to be 'fixed' as well.
Is the emphasis of an emd on code size, execution time or interrupt handling availability ??

asminc/c128.inc
```diff
@@ -198,12 +198,14 @@ CIA2_CRB        := $DD0F
...
-MMU_CFG_RAM1    := %01111111    ; Bank 1 full RAM
-MMU_CFG_RAM2    := %10111111    ; Bank 2 full RAM
-MMU_CFG_RAM3    := %11111111    ; Bank 3 full RAM
+MMU_CR           := $FF00
+MMU_CFG_CC65     := %00001110    ; Bank 0 with kernal ROM
+MMU_CFG_RAM0     := %00111111    ; Bank 0 full RAM
+MMU_CFG_RAM1     := %01111111    ; Bank 1 full RAM
+MMU_CFG_RAM2     := %10111111    ; Bank 2 full RAM
+MMU_CFG_RAM3     := %11111111    ; Bank 3 full RAM
+MMU_CFG_INT_FROM := %01010111    ; Bank 1 with Internal Function RAM/ROM
```
greg-king5: MMU_CFG_IFROM   := %01010111    ; Bank 1 with Internal Function RAM/ROM

asminc/c128.inc
```diff
@@ -198,12 +198,14 @@ CIA2_CRB        := $DD0F
...
+MMU_CFG_EXT_FROM := %01101011    ; Bank 1 with External Function RAM/ROM
```
greg-king5: We prefer that the Assembly fields are aligned on eight-character boundaries.  Therefore, I think that those last two names should be shortenned.  I suggest:

+ `MMU_CFG_IFROM`
+ `MMU_CFG_EFROM`

libsrc/c128/emd/c128-efnram.s
oliverschmidt: Maybe I'm missing the point but at least in my cc65 user expereince I run much more often in memory shortage issues than performance issues. From that perspective I'd say that in general code size has the highest prio. However here are of course hot spots rather optimized for performance. An example coming into my mind are unrolled loops in CONIO `clrscr()` implementations.

Issue #244, opened 2015-12-11
Bug shifting right


mojontwins:
Hello.

I've been a cc65 user for quite a long time. I've been using 2.13.3 but today I decided to, finally, upgrade to the latest version, but I found something: This code:

```c
unsigned char uc;
signed int si;

...

void main (void) {
    [...]
    si = 160 << 6;
    uc = si >> 6;
    [...]
}
```
"pry" should equal 160, but the result is 224 instead with the latest "win32 snapshot". I was afraid this was an error on my side, but I double checked with some other values and asked other users in nesdev in this thread.

http://forums.nesdev.com/viewtopic.php?f=2&t=13616&p=160508#p160508

It seems that I found a bug, as the user **lidnariq** mentions:

> In the older release, the >>6 generates calls to asrax4 and asrax2. In the newer one, it instead generates a call to just asrax4 and then tries to inline the remaining >>2, but it does it wrong (it forgets about X and sign-extends A instead of copying the two remaining bits from X).

Thank you very much.

drexegar:
I have this bug on the latest cc65 last week on the main website I downloaded it from https://cc65.github.io/ can someone update it?

Also i have no clue how to download cc65 from github.

mrdudz:
The snapshot linked from the website is the latest one. If you think there is a bug, then open another issue.

However, what you describe looks like expected behaviour. When A is 1, and carry is set, then rotating right two times will result in $c0

PR #246, opened 2015-12-13
Fix a signed char right-shift compiler optimization.

greg-king5:
This PR fixes bug report #244.

A compiler shift optimization that is intended for signed character expressions and the high byte of signed integer expressions (when the right-shift is greater than eight bits) wasn't limited enough.  Sometimes, it was used on only the low byte of integers (when a high byte wasn't going to be used after the shift); but, that would fail to shift some bits from the high byte into the low byte.

This patch makes sure that cc65 won't optimize the wrong integer shifts.

### DIFF
--- a/src/cc65/coptshift.c
+++ b/src/cc65/coptshift.c
@@ -320,6 +320,11 @@ unsigned OptShift2 (CodeSeg* S)
 **      ror     a
 **
 ** if X is not used later (X is assumed to be zero on entry).
+** If the sequence is followed immediately by another
+**
+**      jsr     asraxN
+**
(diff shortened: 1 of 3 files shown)


mrdudz:
it'd be really really REALLY nice if you could add a testprogram for this as well :)

greg-king5:
OK, I added a regression test program.  The old compiler fails it; the new compiler passes it.

mrdudz:
great, thanks!

oliverschmidt:
I'd like to express (yet again) that I REALLY appreciate your work for cc65 - thanks !

PR #247, opened 2015-12-14
nes: Fix reading the second controller

clbr:
See
http://wiki.nesdev.com/w/index.php/Controller_port_registers


oliverschmidt:
Thanks for the fix :-)

PR #248, opened 2015-12-15
Remove a const qualifier from <nes.h>.

greg-king5:
I found a bug in my Pull Request #229 while I was looking at PR #247.


oliverschmidt:
I see.

PR #249, opened 2015-12-27
My first PR

polluks:


oliverschmidt:
Hi,

Thanks for actively contributing to cc65@GitHub :-)

Regards,
Oliver

polluks:
Nearly three years, I'm not the fastest.
What about chrcvt at all, would you like to include it in the makefile? It should be part of the toolchain like sp65 IMHO.

oliverschmidt:
> What about chrcvt at all, would you like to include it in the makefile? It should be part of the toolchain like sp65 IMHO.

The from my perspective only reason to not have it in the Makefile is that it doesn't follow the <...>65 naming scheme. I'd like to see cc65 stick to it and in fact it's necessary for having https://github.com/cc65/cc65/blob/master/Makefile lines 19/20 work. Maybe you want to adjust the name via a pull request...

Issue #251, opened 2015-12-27
Comparing const and non-const pointers


clbr:
According to C89, comparing a non-const pointer to a const pointer is valid.

The following fails with cc65 and succeeds with gcc -ansi -pedantic.
```
unsigned char foo(const unsigned char *cp, unsigned char *p) {
        return cp == p;
}
```

oliverschmidt:
Thanks for the report :-)

PR #253, opened 2015-12-29
Make cc65 accept comparisons between pointers with different qualifiers.

greg-king5:
This fixes issue #251.


oliverschmidt:
Thanks :-)

Issue #252, opened 2015-12-27
Internal compiler error: Code generation messed up: StackPtr is -4, should be -2


clbr:
This file causes cc65 to ICE.

Code generation messed up: StackPtr is -4, should be -2

Input: if (tableType == byU32) { U32* hashTable = (U32*) tableBase; return hashTable[h] + srcBase; }
cl65: Subprocess `cc65' aborted by signal 6

```
# include <stdint.h>
typedef  uint8_t BYTE;
typedef uint16_t U16;
typedef uint32_t U32;
typedef  int32_t S32;

typedef enum { byPtr, byU32, byU16 } tableType_t;

static const BYTE* LZ4_getPositionOnHash(U32 h, void* tableBase, tableType_t tableType, const BYTE* srcBase)
{
    if (tableType == byPtr) { const BYTE** hashTable = (const BYTE**) tableBase; return hashTable[h]; }
    if (tableType == byU32) { U32* hashTable = (U32*) tableBase; return hashTable[h] + srcBase; }
    { U16* hashTable = (U16*) tableBase; return hashTable[h] + srcBase; }   /* default, to ensure a return */
}
```

greg-king5:
Those `return` expressions add an offset and a pointer.  The offset must be scaled (multiplied) by the size of the pointer's target, before being added to the pointer.  Usually, we put the pointer first and the offset second; cc65 can handle that sequence easily.  But, when the offset term is first, cc65 cannot know that it needs to be scaled, until cc65 has compiled the pointer term.

At that point, the offset is on the stack; and, the pointer is in the accumulator.  cc65 must use a runtime subroutine to swap those two numbers.  The problem is that the library does not have a 32-bit swapper.  But, cc65 keeps going, anyway.  It creates code that swaps 16 bits; and, adds 16 bits.  The result is that the high half of a 32-bit offset is left orphaned on the stack.

greg-king5:
After more analysis, I discovered that the bug is less severe than I though it is.  cc65 thinks that the offset is an `int` because the function that should demote it is told about the pointer instead of the offset.  Therefore, it thinks that it does not need to do anything.  (The bug is a single-character typo.)

oliverschmidt:
@clbr: Thanks for the report :-)

PR #254, opened 2016-01-02
Fix an unbalanced C stack that happens when a pointer is added to a 32-bit offset.

greg-king5:
This fixes bug report #252.

In certain conditions, a number on the left side of an addition operation should be recast to a smaller data type.  But, cc65's code looks at the right side instead of the left side.  Therefore, it doesn't realize that the recast needs to be done.  Later actions assume that the recast was done.  cc65 creates program code that would leave garbage on the C stack.  When cc65 notices that the stack would be bad, it shows a message, and aborts.

This PR fixes a typo that names the wrong side of the addition.
And, I included a simple regression test for that problem.


oliverschmidt:
Such a small fix ;-))

Thanks !

PR #257, opened 2016-01-15
Make ca65's CONDES-type pseudo-instructions save line numbers correctly.

### LINKED ISSUE #212: ca65 exits with internal error instead of giving a proper error message when a non existing label is used as .constructor
what the topic says, test.s:
```assembler
        .constructor foo
bar: rts
```

 $ ./bin/cl65 -o test.prg test.s
ca65: Precondition violated: Index < C->Count, file `common/coll.h', line 155

greg-king5:
Fixes #212.

ca65 memorizes the line numbers where symbol names are defined, and where they are used -- in two separate collections for each name.  Those line numbers are displayed in diagnostic messages.

The `.constructor`, `.destructor`, `.interruptor`, and `.condes` instructions saved those numbers in the "defined here" collection, instead of the "used here" collection (i.e., they acted like `.import`, instead of like `.export`).  Therefore, if ca65 needed to show error messages about those instructions, it did not have any "used here" line numbers to put in the messages.  ca65 failed a PRECONDITION (which tests the integrity of collections of objects); and, stopped.

### DIFF
--- a/src/ca65/symentry.c
+++ b/src/ca65/symentry.c
@@ -570,8 +570,8 @@ void SymConDes (SymEntry* S, unsigned char AddrSize, unsigned Type, unsigned Pri
         }
     }
 
-    /* If the symbol was already declared as a condes, check if the new
-    ** priority value is the same as the old one.
+    /* If the symbol already was declared as a condes of this type,
+    ** check if the new priority value is the same as the old one.
     */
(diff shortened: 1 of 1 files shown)


oliverschmidt:
Thanks for the fix - and the comprehensive explanation :-)

Issue #260, opened 2016-01-17
cc65 is missing useful Apple 2 Text macros


Michaelangel007:
On the Apple ][ //e //c programmers need a variety of way of controlling ASCII (high-bit off), APPLE (high-bit on), and Control characters ($0..$1F).

Since out-the-box cc65 doesn't have anything useful in `asminc/`, here is an `apple2text.inc` file that provides this functionality.

```assembly

; NOTE: There is no "standard" directive to set the high bit on ASCII text.
;
; * The assembler used in the redbook uses `msb on` and `msb off`
; * S-C Macro Assembler used
;     .AS for normal ASCII, and
;     .AT for normal ASCII but the last char has the high bit ON
; * Merlin uses yet another variation:
;     ASC 'Hello'   ; high bit is off
;     ASC "Hello"   ; high bit is ON
;   See the new Merlin-32 project
;   http://brutaldeluxe.fr/products/crossdevtools/merlin/
; 
; "The 'nice' thing about standards, is that there are so many to pick from!"

; Force APPLE 'text' to have high bit on
.macro AS text
    .repeat .strlen(text), I
        .byte   .strat(text, I) | $80
    .endrep
.endmacro

; Force ASCII 'text' but last character has high bit on
; Will display as FLASHING characters
; NOTE: There is no lowercase flashinng characters - they will appar as gibbersh!
.macro AT text
    .repeat .strlen(text)-1, I
        .byte   .strat(text, I) & $7F
    .endrep
    .byte   .strat(text, .strlen(text)-1) | $80
.endmacro

; Force APPLE 'text' with high bit on but last character has high bit off
.macro AR text
    .repeat .strlen(text)-1, I
        .byte   .strat(text, I) | $80
    .endrep
    .byte   .strat(text, .strlen(text)-1) & $7F
.endmacro

; Force ASCII 'text' to be control chars: $00..$1F
; Will display as INVERSE characters
.macro AC text
    .repeat .strlen(text), I
        .byte   .strat(text, I) & $1F
    .endrep
.endmacro
```

Example usage:

```assembly
AppleTextDemo:
    AR "ABC"            ; NORMAL : APPLE high-bit text, last char is ASCII (FLASHING)
    AS "Hello world"    ; NORMAL : APPLE high-bit text
    AT "XYZ"            ; FLASH  : ASCII text, last char is APPLE high-bit (NORMAL)
    AC "DEFM"           ; INVERSE: Force to be control chars0 $0..$1F (INVERSE)
[...]

PR #262, opened 2016-01-25
Cc65upstream

Michaelangel007:
Cleanup done for #260, ready for integration.

`Note:` I'm not sure if these short 2 character macro names will clash with an user's existing names since apple2.mac is automatically included?

I also don't like the non-descript `AT`. May want to rename these macros with the `A2` prefix to try to manage the global "name pollution".

Example:

* AS -> `A2S`
* AT -> `A2T`
* AR -> `A2R`
* AC -> `A2C`
* AI -> `A2I`

Since source code written for other assemblers will have to be cleaned up anyways, the final names should be chosen as not to create too much of a minor burden.


oliverschmidt:
Unfortunately the discussion that should happened here was placed in https://github.com/cc65/cc65/issues/260 so far.

The last topic there is:

>> What would a user type to get the various normal, flashing, or inverse strings/characters?

> I'm think along the lines of scrcode, scrcode_fla and scrcode_inv. But maybe there are better ideas...

Given that (one of) the strengths of the cc65 tool chain is it's cross target approach I'd really like to see the same macro name being used for the same purpose as on the other targets. Therefore I suggest to go for ```scrcode``` and add variants of that for text different text attributes as I proposed above.

Apart from that I second that ```Apple 2``` should rather read ```Apple II```.

Given the amount of work invested so far - including the nice documentation - I suggest to have the minor issues left sorted out and get this merged.

mrdudz:
i am not into apple2 much so i can only comment generally....

i don't like those short "AT" etc style macro names, like Oliver already said, i'd propose .scrcode for normal characters, and .scrcode_xy for the others. IF these are actual screencodes (and not something that resembles "ascii") that is :)

besides this, seems ok to merge

oliverschmidt:
@mrdudz: Full ACK. Beside the actual macro names the pull request is fine from my pov.

doc/ca65.sgml
```diff
@@ -4835,6 +4869,76 @@ are used for the cbm load address.
...
+</verb></tscreen>
+
+NOTE: Even Apple's assembler was slightly inconsistent:
+
+<itemize>
+<item>The DOS 3.3 version of EDASM defaulted to MSB ON, and
+<item>The ProDOS  version of EDASM defaulted to MSB OFF.
+</itemize>
+
+More more info see the book "ProDOS Assembler Tools."
```
greg-king5: Typo: `More more` --> `For more`.

Michaelangel007:
Thanks for the catch Greg!

I've been way for work last week and sadly too much Real-Life (TM) crap this week to deal with but I'll make another pass this coming up weekend to sync up the screen code names.

oliverschmidt:
@Michaelangel007: Thanks for the intermediate feedback.

oliverschmidt:
This pull request didn't show much momentum recently. I went ahead and added a basic scrcode macro designed like the existing Atari and CBM ones (i.e. allowing several arguments) with https://github.com/cc65/cc65/commit/e2419ece0b596d137b6a9b98f4eb2844c8c29578.

I didn't see much use for the other macros, especially as they depend on the currently active charset. And without those macros the comparison with other assemblers didn't seem to make too much sense to me either.

However https://github.com/cc65/cc65/commit/e2419ece0b596d137b6a9b98f4eb2844c8c29578 shows how useful the scrcode macro in general is - so thanks for bringing up the topic in general :-)

PR #269, opened 2016-02-25
New ".feature" to use brackets instead of parens for 6502 indirect addressing.

jbrandwood:
Things to note ...

1) It's actually a pretty small change, but it seems bigger because I'm moving the code that checks for the 65816's "far" addressing mode to below the check for the 6502 indirect addressing mode. I've also added a comment to the "far" check to indicate that it isn't executed when the BracketAsIndirect option is activated.

2) It could all be done a tiny bit more efficiently with static variables, but that didn't seem to be as easy-to-read.

### DIFF
--- a/doc/ca65.sgml
+++ b/doc/ca65.sgml
@@ -2701,13 +2701,15 @@ Here's a list of all control commands and a description, what they do:
 
   <tag><tt>bracket_as_indirect</tt><label id="bracket_as_indirect"></tag>
 
-    Use <tt>[]</tt> intead of <tt>()</tt> for the indirect addressing mode.
+    Use <tt>[]</tt> instead of <tt>()</tt> for the indirect addressing modes.
     Example:
 
     <tscreen><verb>
         lda     [$82]
         lda     [$82,x]
         lda     [$82],y
+        jmp     [$fffe]
+        jmp     [table,x]
     </verb></tscreen>
     <em/Note:/ This should not be used in 65186 mode because it conflicts with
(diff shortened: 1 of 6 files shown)


doc/ca65.sgml
```diff
@@ -2699,6 +2699,20 @@ Here's a list of all control commands and a description, what they do:
     at character is not allowed to start an identifier, even with this
     feature enabled.
 
+  <tag><tt>bracket_as_indirect</tt><label id="bracket_as_indirect"></tag>
+
+    Use <tt>[]</tt> intead of <tt>()</tt> for the indirect addressing mode.
```
oliverschmidt: ```intead``` should read ```instead```.

I think we're talkong about multiple addressing ```modes``` here.

doc/ca65.sgml
```diff
@@ -2699,6 +2699,20 @@ Here's a list of all control commands and a description, what they do:
...
+    Example:
+
+    <tscreen><verb>
+        lda     [$82]
+        lda     [$82,x]
+        lda     [$82],y
```
oliverschmidt: Is this supposed to be applicable to JMP (...) too? If yes:
- Did you tested it? 
- It should be added here.

jbrandwood:
Thanks for the feedback, and "sorry" about the typo.
I've made the changes that were requested.
And "yes", I've tested that the JMP [] and JMP [,x] instructions work.

oliverschmidt:
@John: Thanks for contributing to cc65@GitHub :-)

There seem to be no additional comments on this pull request, so...

PR #279, opened 2016-03-26
Convert the Atmos configuration to the new constructor segment model.

greg-king5:
This PR brings the Atmos library into sync with the C64 and Apple II libraries.  (`INIT` and `ONCE` are at the end of the program file; and, `BSS` is in a separate memory area.)

A nice consequence is that the work-around, for the firmware bug, is simpler, now.


oliverschmidt:
Thanks :-)

PR #283, opened 2016-04-13
Fixed CPU definition of Lynx.

polluks:
Greg wrote: The target platforms automatically choose a default CPU instruction-set. Therefore, CPU directives shouldn't be in the include files. Commit 55d6556 must be changed. It should remove the directive in "supervision.inc".

Currently, the Lynx target chooses the 65C02. If it should choose the 65SC02, then "src/common/target.c" should be patched.

### DIFF
--- a/src/common/target.c
+++ b/src/common/target.c
@@ -139,7 +139,7 @@ struct TargetEntry {
 };
 
 /* Table that maps target names to ids. Sorted alphabetically for bsearch.
-** Allows mupltiple entries for one target id (target name aliases).
+** Allows multiple entries for one target id (target name aliases).
 */
 static const TargetEntry TargetMap[] = {
     {   "apple2",       TGT_APPLE2      },
@@ -168,7 +168,6 @@ static const TargetEntry TargetMap[] = {
     {   "sim6502",      TGT_SIM6502     },
     {   "sim65c02",     TGT_SIM65C02    },
     {   "supervision",  TGT_SUPERVISION },
-    {   "vc20",         TGT_VIC20       },
     {   "vic20",        TGT_VIC20       },
 };
 #define MAP_ENTRY_COUNT         (sizeof (TargetMap) / sizeof (TargetMap[0]))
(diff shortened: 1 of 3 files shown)


src/common/target.c
```diff
@@ -168,7 +168,6 @@ static const TargetEntry TargetMap[] = {
     {   "sim6502",      TGT_SIM6502     },
     {   "sim65c02",     TGT_SIM65C02    },
     {   "supervision",  TGT_SUPERVISION },
-    {   "vc20",         TGT_VIC20       },
```
spiro-trikaliotis: Why remove "VC20"? That was the name of the VIC20 in Germany.
While I would not have introduced that alias at the beginning, I do not think it is useful to remove it now and possibly break some build environments. What do we gain from this removal?

polluks: Well, it's undocumented, so I don't see any compatibility breaks.
What about vic1001 or apple2euro? Simply don't support any explicit aliases.

spiro-trikaliotis: Just because it is undocumented does not mean noone is using it.

Yes, do not introduce explicit aliases. But, if they are already there, do not remove them without the need. Do not break the work of others just for the sake of breaking it.

polluks: If you are using undocumented features you are on your own. C'est la vie.
We already have many targets, we will get even more and I want clean source.
At least I mentioned the VC20 in vic20.sgml.

mrdudz: > If you are using undocumented features you are on your own. C'est la vie.
i agree - even when i was using it, its good to remove redundant stuff.

groessler: The question is, should the "vc20" entry be removed, or the "vic20".

mrdudz: its only called vc20 in germany, and vic20 everywhere else... so why would it be vc20? :)

groessler: Because I'm living in Germany? :-)
I don't see the point in removing it....

mrdudz: i also live in germany :) however, i don't see a reason for having a single localised commandline option when the whole rest isnt localized at all (and god forbid, don't =P)

oliverschmidt:
It's nice to see the GitHub pull request conversion feature used for such a lively discussion.

I personally think that removal of 'vc20' is appropriate. So I see more momentum to actually remove it.

@polluks: The fact that we discuss an aspect of this pull request which doesn't relate to its goal as given in its title nor in its description is a pretty strong indication that it should have gone in a separate pull request.

As there seem to be no other issues I'm going to merge now...

Issue #275, opened 2016-03-02
__asm__ volatile


dhustkoder:
Hello, I know that cc65 doesn't support volatile variables.
But It can be very helpful to write NES games using cc65.
I'm trying to write a little api for NES with assembly and C.

Is there a way I can get something like __asm__ volatile ?
I have strong background in C and C++. and learning 6502 assembly
if someone guide me in the right direction I could try to implement that in the compiler.
But I've no experience in compiler source code yet. 
Thanks.

mrdudz:
On Wednesday 02 March 2016, 12:53:54 Rafael Moura <notifications@github.com> 
wrote:
> Hello, I know that cc65 doesn't support volatile variables.
> But It can be very helpful to write NES games using cc65.
> I'm trying to write a little api for NES with assembly and C.
> 
> Is there a way I can get something like __asm__ volatile ?
> I have strong background in C and C++. and learning 6502 assembly
> if someone guide me in the right direction I could try to implement that in
> the compiler. But I've no experience in compiler source code yet.
> Thanks.

you can just use it as you are used to do - the compiler will ignore the 
"volatile" keyword, and it will not attempt to optimize your assembly if its 
in one multiline asm() statement.

-- 

http://www.hitmen-console.org    http://magicdisk.untergrund.net
http://www.pokefinder.org        http://ar.pokefinder.org

A scientist can discover a new star, but he cannot make one. He would have to 
ask an engineer to do that. 
<Gordon L. Glegg>

dhustkoder:
well, for some reason my cc65 is not parsing and ignoring the volatile keyword.
version: cc65 V2.15 - Git 60958f0

source:
`
unsigned char x;

void main(void)
{
	__asm__ volatile ("LDA $4016");
	x = 10;	
	cputc(x);
}
`



error: 
t.c(7): Error: `(' expected
t.c(7): Error: `;' expected
t.c(7): Error: Expression expected
t.c(7): Error: Illegal function call

Ok, but even if cc65 parses and ignore the volatile, It would be very beneficial if
the optimizer do not erase the __asm__ instruction in there, as for NES the 
instruction LDA addr have important side effects

of course I can do something like 
`void main(void)
{
	__asm__ ("LDA $4016");
        __asm__ ("STA %v ; REMOVE ME ; ", x);

	x = 10;	
	cputc(x);
}
`
but these kind of tricks are not so reliable, and I would be doing this a lot.
[...]

PR #285, opened 2016-04-21
Add the C keyword "volatile" to the __asm__ statement grammar.

### LINKED ISSUE #276: volatile keyword is not ignored when used with __asm__
```C
volatile int bla = 0; // works
{
    __asm__ volatile ("nop"); // gives an error
}
```
'volatile' should be ignored also in the second case.

greg-king5:
This PR fixes #276.  See also issue #275.

If we insert some Assembly code into a C function, then we might want that code to stay exactly as it was typed.  We might want to block the optimizer from changing our code.  We can use `#pragma optimize()`; or, we can use `volatile` in the `__asm__` statement:
```c
asm volatile ("<assembly code>");
```
`volatile` is more direct (it shows why we want to stop the optimizer); and, if we remove that statement at a latter time, the C function automatically will be optimized normally again.

### DIFF
--- a/doc/cc65.sgml
+++ b/doc/cc65.sgml
@@ -2,8 +2,9 @@
 
 <article>
 <title>cc65 Users Guide
-<author><url url="mailto:uz@cc65.org" name="Ullrich von Bassewitz">
-<date>2015-05-26
+<author><url url="mailto:uz@cc65.org" name="Ullrich von Bassewitz">,<newline>
+<url url="mailto:gregdk@users.sf.net" name="Greg King">
+<date>2016-04-22
 
 <abstract>
 cc65 is a C compiler for 6502 targets. It supports several 6502 based home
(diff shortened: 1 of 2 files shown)


oliverschmidt:
First of all thanks for the pull request.

I'm not sure if I remember the discussion on that topic correctly - but it seems to me it would be desirable if the `volatile` wouldn't be just a NOP but rather have the same effect on the function it is placed in that the `#pragma optimize (off)` has.

greg-king5:
OK.  Now, `volatile` actually does something in cc65!  ;-)

I updated this PR's description.  It mentions the original discussion.

oliverschmidt:
Very nice !

I understand wall that from a conceptual perspective it's not "correct" to fully disable optimization because of the `volatile` but it fully conforms to the expectation of any programmer knowing `volatile asm` from i.e. gcc.

I sort of hoped that it would be possible to do this with a rather small code change "just turning of the optimizer" just in time. I very happy to see that this was indeed the case :-)

PR #287, opened 2016-05-13
moved output of target utils and drivers to separate directory

silverdr:
Drivers moved to goodies/drivers, targetutil moved to goodies/targetutil, zip target now zips samples in addition to previously supported directories and doesn't break when html directory is missing (like not created due to lack of linuxdoc-tools)

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -41,7 +41,9 @@ ifeq ($(wildcard ../info),../info)
 endif
 
 zip:
+ifneq "$(wildcard ../html)" ""
 	@cd .. && zip cc65 html/*.*
+endif
 
 doc: html info
 
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -33,11 +33,11 @@ install:
 	$(if $(prefix),,$(error variable `prefix' must be set))
 ifeq ($(wildcard ../html),../html)
 	$(INSTALL) -d $(DESTDIR)$(htmldir)
-	$(INSTALL) -m644 ../html/*.* $(DESTDIR)$(htmldir)
+	$(INSTALL) -m0644 ../html/*.* $(DESTDIR)$(htmldir)
 endif
 ifeq ($(wildcard ../info),../info)
 	$(INSTALL) -d $(DESTDIR)$(infodir)
-	$(INSTALL) -m644 ../info/*.* $(DESTDIR)$(infodir)
+	$(INSTALL) -m0644 ../info/*.* $(DESTDIR)$(infodir)
 endif
 
 zip:
--- a/libsrc/Makefile
+++ b/libsrc/Makefile
@@ -39,12 +39,13 @@ DRVTYPES = emd \
            tgi
 
 OUTPUTDIRS := lib                 \
-              target              \
               asminc              \
               cfg                 \
               include             \
               samples             \
-              $(subst ../,,$(filter-out $(wildcard ../include/*.*),$(wildcard ../include/*)))
+              $(subst ../,,$(filter-out $(wildcard ../include/*.*),$(wildcard ../include/*)))\
+              $(subst ../,,$(wildcard ../target/*/drv/*))\
+              $(subst ../,,$(wildcard ../target/*/util))\
 
 .PHONY: all mostlyclean clean install zip lib $(TARGETS)
 
@@ -80,8 +81,8 @@ mostlyclean:
 # Transitional line active. Final line commented out below in order to
 # allow some time for transition between the directory structures
 clean:
-	$(call RMDIR,../libwrk ../lib ../targetutil ../target $(addprefix ../,$(DRVTYPES)))
-# 	$(call RMDIR,../libwrk ../lib ../target)
+	$(call RMDIR,../libwrk ../lib ../targetutil ../$(TARGETDIR) $(addprefix ../,$(DRVTYPES)))
+# 	$(call RMDIR,../libwrk ../lib ../$(TARGETDIR))
 
 ifdef CMD_EXE
 
@@ -95,12 +96,13 @@ define INSTALL_recipe
 
 $(if $(prefix),,$(error variable `prefix' must be set))
 $(INSTALL) -d $(DESTDIR)$(datadir)/$(dir)
-$(INSTALL) -m644 ../$(dir)/*.* $(DESTDIR)$(datadir)/$(dir)
+$(INSTALL) -m0644 ../$(dir)/*.* $(DESTDIR)$(datadir)/$(dir)
 
 endef # INSTALL_recipe
 
 install:
 	$(foreach dir,$(OUTPUTDIRS),$(INSTALL_recipe))
+
 
 endif # CMD_EXE
 
--- a/libsrc/Makefile
+++ b/libsrc/Makefile
@@ -37,12 +37,15 @@ DRVTYPES = emd \
            ser \
            tgi
 
-OUTPUTDIRS := lib         \
-              $(DRVTYPES) \
-              targetutil  \
-              asminc      \
-              cfg         \
-              include     \
+DRVOUTPUTDIRS := $(foreach drvtype,$(DRVTYPES),goodies/drivers/$(drvtype))
+
+OUTPUTDIRS := lib                 \
+              $(DRVOUTPUTDIRS)    \
+              goodies/targetutil  \
+              asminc              \
+              cfg                 \
+              include             \
+              samples             \
               $(subst ../,,$(filter-out $(wildcard ../include/*.*),$(wildcard ../include/*)))
 
 .PHONY: all mostlyclean clean install zip lib $(TARGETS)
@@ -76,8 +79,11 @@ all lib: $(TARGETS)
 mostlyclean:
 	$(call RMDIR,../libwrk)
 
+# Transitional line active. Final line commented out below in order to
+# allow some time for transition between the directory structures
 clean:
-	$(call RMDIR,../libwrk ../lib ../targetutil $(addprefix ../,$(DRVTYPES)))
+	$(call RMDIR,../libwrk ../lib ../targetutil ../goodies $(addprefix ../,$(DRVTYPES)))
+# 	$(call RMDIR,../libwrk ../lib ../goodies)
 
 ifdef CMD_EXE
 
@@ -212,7 +218,7 @@ define DRVTYPE_template
 $1_SRCDIR = $$(SRCDIR)/$1
 $1_STCDIR = ../libwrk/$$(TARGET)
 $1_DYNDIR = ../libwrk/$$(TARGET)/$1
-$1_DRVDIR = ../$1
+$1_DRVDIR = ../goodies/drivers/$1
 
 $1_SRCPAT = $$($1_SRCDIR)/$$(OBJPFX)%.s
 $1_STCPAT = $$($1_STCDIR)/$$(OBJPFX)%-$1.o
@@ -283,7 +289,7 @@ $(EXTRA_OBJPAT): $(EXTRA_SRCPAT) | ../lib
 ../lib/$(TARGET).lib: $(OBJS) | ../lib
 	$(AR65) a $@ $?
 
-../libwrk/$(TARGET) ../lib ../targetutil:
+../libwrk/$(TARGET) ../lib ../goodies/targetutil:
 	@$(call MKDIR,$@)
 
 $(TARGET): $(EXTRA_OBJS) ../lib/$(TARGET).lib
--- a/libsrc/Makefile
+++ b/libsrc/Makefile
@@ -78,11 +78,8 @@ all lib: $(TARGETS)
 mostlyclean:
 	$(call RMDIR,../libwrk)
 
-# Transitional line active. Final line commented out below in order to
-# allow some time for transition between the directory structures
 clean:
-	$(call RMDIR,../libwrk ../lib ../targetutil ../target $(addprefix ../,$(DRVTYPES)))
-# 	$(call RMDIR,../libwrk ../lib ../target)
+	$(call RMDIR,../libwrk ../lib ../target)
 
 ifdef CMD_EXE
 
@@ -102,7 +99,6 @@ endef # INSTALL_recipe
 
 install:
 	$(foreach dir,$(OUTPUTDIRS),$(INSTALL_recipe))
-
 
 endif # CMD_EXE
 
--- a/libsrc/Makefile
+++ b/libsrc/Makefile
@@ -1,3 +1,4 @@
+
 ifneq ($(shell echo),)
   CMD_EXE = 1
 endif
@@ -37,11 +38,8 @@ DRVTYPES = emd \
            ser \
            tgi
 
-DRVOUTPUTDIRS := $(foreach drvtype,$(DRVTYPES),goodies/drivers/$(drvtype))
-
 OUTPUTDIRS := lib                 \
-              $(DRVOUTPUTDIRS)    \
-              goodies/targetutil  \
+              target              \
               asminc              \
               cfg                 \
               include             \
@@ -82,8 +80,8 @@ mostlyclean:
 # Transitional line active. Final line commented out below in order to
 # allow some time for transition between the directory structures
 clean:
-	$(call RMDIR,../libwrk ../lib ../targetutil ../goodies $(addprefix ../,$(DRVTYPES)))
-# 	$(call RMDIR,../libwrk ../lib ../goodies)
+	$(call RMDIR,../libwrk ../lib ../targetutil ../target $(addprefix ../,$(DRVTYPES)))
+# 	$(call RMDIR,../libwrk ../lib ../target)
 
 ifdef CMD_EXE
 
@@ -218,7 +216,7 @@ define DRVTYPE_template
 $1_SRCDIR = $$(SRCDIR)/$1
 $1_STCDIR = ../libwrk/$$(TARGET)
 $1_DYNDIR = ../libwrk/$$(TARGET)/$1
-$1_DRVDIR = ../goodies/drivers/$1
+$1_DRVDIR = ../target/$$(TARGET)/drv/$1
 
 $1_SRCPAT = $$($1_SRCDIR)/$$(OBJPFX)%.s
 $1_STCPAT = $$($1_STCDIR)/$$(OBJPFX)%-$1.o
@@ -289,7 +287,7 @@ $(EXTRA_OBJPAT): $(EXTRA_SRCPAT) | ../lib
 ../lib/$(TARGET).lib: $(OBJS) | ../lib
 	$(AR65) a $@ $?
 
-../libwrk/$(TARGET) ../lib ../goodies/targetutil:
+../libwrk/$(TARGET) ../lib ../target/$(TARGET)/util:
 	@$(call MKDIR,$@)
 
 $(TARGET): $(EXTRA_OBJS) ../lib/$(TARGET).lib
--- a/libsrc/Makefile
+++ b/libsrc/Makefile
@@ -42,7 +42,6 @@ OUTPUTDIRS := lib                 \
               asminc              \
               cfg                 \
               include             \
-              samples             \
               $(subst ../,,$(filter-out $(wildcard ../include/*.*),$(wildcard ../include/*)))\
               $(subst ../,,$(wildcard ../target/*/drv/*))\
               $(subst ../,,$(wildcard ../target/*/util))\
--- a/libsrc/Makefile
+++ b/libsrc/Makefile
@@ -81,8 +81,8 @@ mostlyclean:
 # Transitional line active. Final line commented out below in order to
 # allow some time for transition between the directory structures
 clean:
-	$(call RMDIR,../libwrk ../lib ../targetutil ../$(TARGETDIR) $(addprefix ../,$(DRVTYPES)))
-# 	$(call RMDIR,../libwrk ../lib ../$(TARGETDIR))
+	$(call RMDIR,../libwrk ../lib ../targetutil ../target $(addprefix ../,$(DRVTYPES)))
+# 	$(call RMDIR,../libwrk ../lib ../target)
 
 ifdef CMD_EXE
 
--- a/libsrc/nes/Makefile.inc
+++ b/libsrc/nes/Makefile.inc
@@ -1,8 +1,8 @@
-../goodies/drivers/tgi/nes-64-56-2.tgi: ../libwrk/nes/clrscr.o \
-                                               ../libwrk/nes/cputc.o  \
-                                               ../libwrk/nes/get_tv.o \
-                                               ../libwrk/nes/gotoxy.o \
-                                               ../libwrk/nes/popa.o   \
-                                               ../libwrk/nes/ppu.o    \
-                                               ../libwrk/nes/ppubuf.o \
-                                               ../libwrk/nes/setcursor.o
+../target/nes/drv/tgi/nes-64-56-2.tgi: ../libwrk/nes/clrscr.o \
+                                   ../libwrk/nes/cputc.o  \
+                                   ../libwrk/nes/get_tv.o \
+                                   ../libwrk/nes/gotoxy.o \
+                                   ../libwrk/nes/popa.o   \
+                                   ../libwrk/nes/ppu.o    \
+                                   ../libwrk/nes/ppubuf.o \
+                                   ../libwrk/nes/setcursor.o
--- a/libsrc/nes/Makefile.inc
+++ b/libsrc/nes/Makefile.inc
@@ -1,8 +1,9 @@
-../target/nes/drv/tgi/nes-64-56-2.tgi: ../libwrk/nes/clrscr.o \
-                                       ../libwrk/nes/cputc.o  \
-                                       ../libwrk/nes/get_tv.o \
-                                       ../libwrk/nes/gotoxy.o \
-                                       ../libwrk/nes/popa.o   \
-                                       ../libwrk/nes/ppu.o    \
-                                       ../libwrk/nes/ppubuf.o \
-                                       ../libwrk/nes/setcursor.o
+../target/nes/drv/tgi/nes-64-56-2.tgi:  \
+                ../libwrk/nes/clrscr.o  \
+                ../libwrk/nes/cputc.o   \
+                ../libwrk/nes/get_tv.o  \
+                ../libwrk/nes/gotoxy.o  \
+                ../libwrk/nes/popa.o    \
+                ../libwrk/nes/ppu.o     \
+                ../libwrk/nes/ppubuf.o  \
+                ../libwrk/nes/setcursor.o
--- a/libsrc/nes/Makefile.inc
+++ b/libsrc/nes/Makefile.inc
@@ -1,8 +1,8 @@
 ../target/nes/drv/tgi/nes-64-56-2.tgi: ../libwrk/nes/clrscr.o \
-                                   ../libwrk/nes/cputc.o  \
-                                   ../libwrk/nes/get_tv.o \
-                                   ../libwrk/nes/gotoxy.o \
-                                   ../libwrk/nes/popa.o   \
-                                   ../libwrk/nes/ppu.o    \
-                                   ../libwrk/nes/ppubuf.o \
-                                   ../libwrk/nes/setcursor.o
+                                       ../libwrk/nes/cputc.o  \
+                                       ../libwrk/nes/get_tv.o \
+                                       ../libwrk/nes/gotoxy.o \
+                                       ../libwrk/nes/popa.o   \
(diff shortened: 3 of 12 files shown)


doc/Makefile
```diff
@@ -41,7 +41,9 @@ ifeq ($(wildcard ../info),../info)
 endif
 
 zip:
+ifneq "$(wildcard ../html)" ""
 	@cd .. && zip cc65 html/*.*
+endif
```
oliverschmidt: From your pr comment I understand that the above is rather a fix and not actually related to the change in directory layout - correct?

silverdr: Yes. Without this, if the html directory was not created, the zip target would fail. I felt it'd be consistent that if we allow the build to pass without building docs then we shouldn't demand them to be always present here.

oliverschmidt:
In general I like the idea behind this pull request - moving everything that is intended to be put into a taget filesystem into one directory.

However it isn't fun to change this type of things too often. Therefore I suggest to do somewhat more cleanup: It's just tautologous to have all files of i.e. type 'mou' in a directory 'mou'. I see two viable options:
1.) Put the different drivers for one target into one directory named according to the target.
2.) Put just all drivers into one single directory. After all their names are already prefixed with the (shorted) target names.

Apart from that more general aspect I suggest to replace the (very generic) 'goodies' with 'target'. As a result 'targetutil' can become just 'util'.

Result:

- target/util

and

- target/drv

or

- target/apple2
- target/c64
- ...

Finally we need to keep in mind that this type of change almost certainly breaks http://software.opensuse.org/download.html?project=home%3Astrik&package=cc65 so I'm wondering if you could imagine to get in touch with Spiro and ask him if he'll be able to adjust his code.

doc/Makefile
oliverschmidt: Surely. I justed wanted to make sure I got this right.

silverdr:
I agree with all the comments above. F. e. I didn't feel too good with the naming either. Therefore - once done (huh - hopefully not too far away in time) I shall also contact Spiro on the matter. It will be easier to basically change the naming. Changing structure to switch from driver_type-centered to target-centered would be more difficult. Shall have a look but no promises ;-)

oliverschmidt:
>  Changing structure to switch from driver_type-centered to target-centered 

Well said. It just makes more sense that way because an individual user is interested in the files for "his" target !

greg-king5:
You will need to adjust `samples/Makefile` and `.gitignore`, too.

spiro-trikaliotis:
Hello,

* On Fri, May 13, 2016 at 01:54:37PM -0700 Oliver Schmidt wrote:
 
> Finally we need to keep in mind that this type of change almost certainly
> breaks http://software.opensuse.org/download.html?project=home%3Astrik&package=
> cc65 so I'm wondering if you could imagine to get in touch with Spiro and ask
> him if he'll be able to adjust his code.

Although I do not answer regularly, I am reading here. Thus, no need to
contact me separately.

Of course, a detailed list of what has changed before the pull request
is merged might help minimising the downtime of the automatic build.

Regards,
Spiro.

-- 
Spiro R. Trikaliotis
http://www.trikaliotis.net/

oliverschmidt:
> Although I do not answer regularly, I am reading here. Thus, no need to
contact me separately.

- I presumed that you read the mailing list but not necessarily that you check the pull request discussions.

- My proposal to silverdr wasn't primarily about making you aware of this pull request but about asking you of are willing to make the necessary adjustments soon. If you would i.e. answer that in 6 weeks is much better than now I would have proposed to postpone merging this pull request.

spiro-trikaliotis:
Hello Oliver,

* On Sun, May 15, 2016 at 02:02:30PM -0700 Oliver Schmidt wrote:
 
>   • I presumed that you read the mailing list but not necessarily that you
>     check the pull request discussions.

I do not even see a difference in the first place. Everything is sorted
as if it was sent via the mailing list.

>   • My proposal to silverdr wasn't primarily about making you aware of this
>     pull request but about asking you of are willing to make the necessary
>     adjustments soon. If you would i.e. answer that in 6 weeks is much better
>     than now I would have proposed to postpone merging this pull request.

Now is as good as in 6 weeks. As soon as I know that something will
change, I can make the changes.

Tell me when you consider the pull request as "complete", and I can
prepare the adaptions if you want them before the merge. That would even
give me some time for questions if I find out that something is not the
way I would have expected it.

Regards,
Spiro.

-- 
Spiro R. Trikaliotis
http://www.trikaliotis.net/

oliverschmidt:
@spiro-trikaliotis: Thanks for the positive feedback :-)

silverdr:
OK - so we move all things target into "target" directory. And now - should there be two subdirs like `drv` and `util`, where `drv` contains all target-named dirs and `util` remains standalone under `target` or should the target-named directories be placed immediately under `target` and then appropriate `drv` and `util` (if available for a given target). The latter seems to make more sense in the light of the comment 
> an individual user is interested in the files for "his" target

its advantages are a) it covers the quoted case fully, b) less questions of the type "what are those files in `targetutil` and what target(s) are they for?". Disadvantage is a more complex directory structure where in most targets one has to dive into a hierarchy like `cc65/target/c64/drv/joy/c64-stdjoy.joy` or at least `cc65/target/c64/drv/c64-stdjoy.joy` even if there is nothing else than drivers for a given target (like no "util").

silverdr:
OK - done with the second option, which IMHO makes more sense and gives more consistency.

greg-king5:
> ...
> most targets, one has to dive into a hierarchy like `cc65/target/c64/drv/joy/c64-stdjoy.joy` or, at least, `cc65/target/c64/drv/c64-stdjoy.joy` even if there is nothing else than drivers for a given target
> ...

I prefer the second pattern for the driver pathnames.  There is less redundancy.  The driver's category is named by the suffix; therefore, we don't need category subdirectories (`emd/`, `joy/`, `mou/`, `ser/`, and `tgi/`).

silverdr:
I initially thought the same. I even made the first version delivering the second structure as above (although not cleanly from the `make` point of view as it seem to  require deeper changes to do it cleanly - current Makefile doesn't adapt easily) but when I looked at the output, especially for feature/driver-rich platforms like C64, I didn't like it. Since the suffix normally doesn't sort, the drivers weren't grouped by function/type and the C64 drivers directory looked messy. When I did it with the first variant, I eventually liked it better with e.g. various joystick drivers grouped in one dir. If the driver base is supposed to grow, even some time in the future, to support more and more modes/devices/peripherals/add-ons/etc. it might be better in the long run to have it grouped anyway.

samples/Makefile
```diff
@@ -12,15 +12,19 @@ SYS	= c64
 # source tree; otherwise, use the "install" directories.
 ifeq "$(wildcard ../src)" ""
 # No source tree
-MOUS = /usr/lib/cc65/mou/$(SYS)*.mou
-TGI  = /usr/lib/cc65/tgi/$(SYS)*.tgi
+MOUS = /usr/lib/cc65/target/$(SYS)/drv/mou/$(SYS)*.mou
```
greg-king5: There are two redundant pieces in that pathname.  I made a comment about the `mou/` repetition, elsewhere.  `$(SYS)`, also, is repeated, needlessly.  A shorter path filter is
```
MOUS = /usr/lib/cc65/target/$(SYS)/drv/*.mou
```

silverdr: The answer here is similar to the other one - In principle I fully agree - that's why I did it the way you suggest first and wasn't satisfied with the outcome.

greg-king5: OK, I accept the idea that directory listings look better with the names grouped as you described.  So, the path filter should look like
```
MOUS = /usr/lib/cc65/target/$(SYS)/drv/mou/*
```

oliverschmidt:
Hi,

The reason why I didn't see / don't see the 'util' directory "under" a target directory is, that target utils may not be actually specific to one target. Think of Apple ][ and Apple //e or ATARI and ATARI XL. There might be an util that works on quite some CBM machines. That is - at least from my perspective - the reason why the target utils don't have a target name prefix like the drivers have.

I understand however that one might argue that for consistency reasons target utils which are applicable to several targets might simply copied in several directories - this would strengthen the idea of a user likely being interested in just one target finding everything "under" the target directory. I'm open for discussion on this question.

Regarding placing drivers for one target into several subdirectories according to the driver type I must admit that I'm pretty strongly against it. I understand that it may look more "organized" but that's in my personal opinion not a strong-enough argument for deep directory structures. We're talking about programmers here. They should be able to sort and/or filter directory listings according to filename extensions. Apart from that most targets have only very vew drivers and I don't see the number of drivers increase significantly.

I'd even go one step further and question the (shortened) target prefix in the driver names. If all C64 drivers are in some 'c64' directory there's no use for the 'c64-' prefix. In contrast: The ATARI filesystem has 8.3 filenames and the prefixes don't look nice. And the Apple II filesystem doesn't allow for '-' so it has to be replaced with '.'. So ommiting the prefix altogether would make things easier. The only reason keeping me from strongly arguing for removing the prefix is backward compatibility. However I usually don't care too much about that anyway ;-)

silverdr:
Greg, Oliver - thanks for the comments. First of all I am not particularly precious about the current setup here. I actually recognised some of those matters you refer to as disadvantages in one of my previous posts. However I briefly describe why I eventually did it the way I did, even if originally I actually did **both** of the things Greg suggested later.

1. Directory structure
 * targets with sizeable number of drivers have a "mess" inside the `drv` directory if everything is inside one dir. Yes, one can sort things out by looking at the suffixes and filtering them on dir listings, etc. Surely programmers can do this but I felt like this is making it more difficult than needed when I had side-by-side look of the two outputs. Moreover it's quite good to see immediately "ah, this target has so many joystick drivers and so many mouse drivers while that one has only one here and two there. True, one probably can fetch this information from the documentation (?) but hey - it's always faster and more convenient to get an overview and rough ideas what are the capabilities by just looking at the directory structure and what files are there. I bet all of us do that before digging the docs, even if they *are* up to date.

2. Driver prefixes
 * when moving around the structure with prefixes removed I found myself a few times asking "okay.. which target am I now looking at" and needed to resort to path fetching in order to make sure
 * One of the greatest things about cc65 is that it enables cross-target development, people may want (and probably do) develop for several targets in parallel. Now, since the drivers are files that are to be copied here and there rather than used in-place, I see bigger chances of situation where one looks at the copied files and wonders "is that the right one for the target in question" when the prefixes are removed. Again - they will eventually sort it out but it seems like a potential waste to have to sort it out when it happened.

Now - having said all that - if even after considering the above you still feel that it's better to do it differently then I'll be fine to make the changes.

silverdr:
> I understand however that one might argue that for consistency reasons target utils which are applicable to several targets might simply copied in several directories

I somehow don't like the idea of copying the same file into several places ;-) OTOH once those are in one basket - it's like "well, ftw is this good for?" rather than "aha, so I can do something for Apple ][ with this thingy" - no strong opinion from my side here.

oliverschmidt:
@silverdr: First all thanks for the effort you put both into the this pull request so far - and the detailed explanation of your motivation!

> 1. Directory structure [....]

I'm still not "in love" with the additional directories. But you are the one who saw it without them and considered it bad enough invest additional effort. Apart from that: If we keep the prefix (see 2.) below) because we want to have the meta data as part of the filename we can as well consider the filename extension as just the same "additional" thing - and define the driver type directory as the "original" thing.

> 2. Driver prefixes [...]

I follow your point that the road target files have to take from the cc65 directory structure to the target filesystem is sometimes winded and that it therefore might be beneficial if the filename is "self describing" without knowing its origin. Apart from that changing this would mean to change the driver name references in the docs. All this together with the backward compatibility aspect makes me agree with keeping the prefixes as-is.

silverdr:
> First all thanks for the effort you put both into this pull request

It's just that I happen to have a little time before the next commercial project and I want to finalise that long overdue OSX binary packaging, which in turn depends on this one ;-)

> the road target files have to take from the cc65 directory structure to the target filesystem is sometimes winded

Exactly what I had in mind.

> we can as well consider the filename extension as just the same "additional" thing - and define the driver type directory as the "original" thing

I am not sure I understand… If you mean that we might for example remove the suffix (aka "extension") and make it part of the filename like `c64-joy-numpad` instead of `c64-numpad.joy` then it would partially solve the problem of driver type directories, yet it wouldn't work well on systems where the 8.3 limit applies.

oliverschmidt:
> I am not sure I understand… If you mean that we might for example remove the suffix (aka "extension") and make it part of the filename like c64-joy-numpad instead of c64-numpad.joy then it would partially solve the problem of driver type directories, yet it wouldn't work well on systems where the 8.3 limit applies.

Sorry for not being unclear. I just wanted to say that one can consider it consistent that i.e. for the C64 Numpad driver named 'c64-numpad.joy' both 'c64' as well 'joy' appear two times: One time as part of the directory path in order to organize things nicely and one time as part of the filename in order to make the file self-describing when moved away from the directory path.

silverdr:
> I just wanted to say that one can consider it consistent [...]

Right. I see it the same. Now we "only" need to decide what to do with the `util` directory. I find the point about the utils being useful for multiple targets very much valid but ATM I don't have any brilliant idea how to deal with it in an elegant way.

oliverschmidt:
>  Now we "only" need to decide what to do with the util directory.

Yep. I read your other comment - I just needed some time to think ;-)

I think we should go for target-specific 'util' directories.

- 'w2cas.com' belongs into 'target/atari/util' and according to http://cc65.github.io/doc/atari.html#ss4.1 it belongs _only_ there. Please note that this doc should be adjusted as it talks about the 'targetutil' directory.

- 'loader.system' belongs into 'target/apple2/util' _and_ 'target/apple2enh/util'. http://cc65.github.io/doc/apple2.html#ss5.1 and http://cc65.github.io/doc/apple2enh.html#ss5.1 would benefit from a minor adjustment: 'targetutil' -> 'target util'.

- 'convert.system' belongs into 'target/geos-apple/util'.

BTW: All existing GEOS drivers belong into 'target/geos-cbm/<drvtype>'.

silverdr:
> 'w2cas.com' belongs into 'target/atari/util' and [...] only there.

```
silverdr$ tree target/atari*
target/atari
├── drv
│   ├── emd
│   [...]
│   └── tgi
│       ├── atr10.tgi
│       [...]
│       └── atr9p2.tgi
└── util
    └── w2cas.com
target/atari5200
└── drv
    └── joy
        └── atr5200std.joy
target/atarixl
└── drv
    ├── emd
    [...]
    └── tgi
        ├── atrx10.tgi
        [...]
        └── atrx9p2.tgi
```

> 'loader.system' belongs into 'target/apple2/util' and 'target/apple2enh/util'.

```
silverdr$ tree target/apple2*
target/apple2
[...]
└── util
    └── loader.system
target/apple2enh
[...]
└── util
    └── loader.system
```

File in question (`loader.system`) is written into both locations.

> 'convert.system' belongs into 'target/geos-apple/util'. [...] All existing GEOS drivers belong into 'target/geos-cbm/[drvtype]'.

```
silverdr$ tree target/geos-*
target/geos-apple
└── util
    └── convert.system
target/geos-cbm
└── drv
    ├── emd
    │   └── geos-vdc.emd
    ├── joy
    │   └── geos-stdjoy.joy
    └── tgi
        └── geos-tgi.tgi
```

Docs adjusted.

silverdr:
When looking at the geos output - it still seems inconsistent with the prefixing scheme… Not that this inconsistency has been introduced with this PR so probably it doesn't need urgent fixing unless you guys think otherwise.

oliverschmidt:
> When looking at the geos output - it still seems inconsistent with the prefixing scheme…

I noticed this myself during putting together my last comment. Let's leave this aspect alone for now.

silverdr:
Wondering if there is something still to be done on this PR?

oliverschmidt:
@silverdr: Thanks for the ping. I was in fact totally unaware that you already took care of all the recent comments !

libsrc/Makefile
```diff
@@ -37,13 +38,14 @@ DRVTYPES = emd \
...
-              targetutil  \
-              asminc      \
-              cfg         \
-              include     \
-              $(subst ../,,$(filter-out $(wildcard ../include/*.*),$(wildcard ../include/*)))
+OUTPUTDIRS := lib                 \
+              asminc              \
+              cfg                 \
+              include             \
+              samples             \
```
oliverschmidt: I don't see why the libsrc Makefile should even know that there's something like 'samples'.

libsrc/Makefile
```diff
@@ -76,8 +78,11 @@ all lib: $(TARGETS)
 mostlyclean:
 	$(call RMDIR,../libwrk)
 
+# Transitional line active. Final line commented out below in order to
+# allow some time for transition between the directory structures
 clean:
-	$(call RMDIR,../libwrk ../lib ../targetutil $(addprefix ../,$(DRVTYPES)))
+	$(call RMDIR,../libwrk ../lib ../targetutil ../target $(addprefix ../,$(DRVTYPES)))
+# 	$(call RMDIR,../libwrk ../lib ../target)
```
oliverschmidt: I don't think that such a transition is necessary / desirable. Please remove it.

libsrc/Makefile
silverdr: True. There should be a better place for this. Shall have a look.

libsrc/Makefile
```diff
@@ -91,13 +96,14 @@ define INSTALL_recipe
 
 $(if $(prefix),,$(error variable `prefix' must be set))
 $(INSTALL) -d $(DESTDIR)$(datadir)/$(dir)
-$(INSTALL) -m644 ../$(dir)/*.* $(DESTDIR)$(datadir)/$(dir)
+$(INSTALL) -m0644 ../$(dir)/*.* $(DESTDIR)$(datadir)/$(dir)
```
oliverschmidt: I must admit that I'm no friend of those "while I'm at it I'll just do this minor thing as well" changes. And here we have already a reason why I think so. You didn't do the same change to the doc/Makefile. Please add it there too or remove it here.

libsrc/Makefile
```diff
@@ -91,13 +96,14 @@ define INSTALL_recipe
...
 
 endef # INSTALL_recipe
 
 install:
 	$(foreach dir,$(OUTPUTDIRS),$(INSTALL_recipe))
 
+
```
oliverschmidt: Is this blank line intentional? If yes, why?

libsrc/nes/Makefile.inc
```diff
@@ -1,8 +1,8 @@
...
-                        ../libwrk/nes/ppubuf.o \
-                        ../libwrk/nes/setcursor.o
+../target/nes/drv/tgi/nes-64-56-2.tgi: ../libwrk/nes/clrscr.o \
+                                   ../libwrk/nes/cputc.o  \
+                                   ../libwrk/nes/get_tv.o \
+                                   ../libwrk/nes/gotoxy.o \
+                                   ../libwrk/nes/popa.o   \
+                                   ../libwrk/nes/ppu.o    \
+                                   ../libwrk/nes/ppubuf.o \
+                                   ../libwrk/nes/setcursor.o
```
oliverschmidt: Before the '..' were all aligned. Please insert enough spaces to make them aligned again.

samples/Makefile
```diff
@@ -12,15 +12,19 @@ SYS	= c64
...
+TGI  = /usr/lib/cc65/target/$(SYS)/drv/tgi/$(SYS)*.tgi
 ifneq "$(wildcard /usr/local/lib/cc65)" ""
-MOUS = /usr/local/lib/cc65/mou/$(SYS)*.mou
-TGI  = /usr/local/lib/cc65/tgi/$(SYS)*.tgi
+MOUS = /usr/local/lib/cc65/target/$(SYS)/drv/mou/$(SYS)*.mou
+TGI  = /usr/local/lib/cc65/target/$(SYS)/drv/tgi/$(SYS)*.tgi
+endif
+ifneq "$(wildcard /opt/local/share/cc65)" ""
```
oliverschmidt: Sorry, but where does '/opt/local/share/...' come from?

libsrc/nes/Makefile.inc
greg-king5: Would it look better if the first "fold" was done at the colon (i.e., the "clrscr.o" path was on the second line)?  Then, the indent wouldn't need to be so wide.

libsrc/Makefile
silverdr: Removed.

libsrc/Makefile
silverdr: Added in doc/Makefile.

libsrc/Makefile
silverdr: Accidental. Removed.

libsrc/nes/Makefile.inc
oliverschmidt: I'm fine with that alternative as well.

silverdr: A regression after some later changes - corrected. If we want to "fold" it differently - let me know.

samples/Makefile
silverdr: OSX installation path. The binary package installer installs it there.

mrdudz: what happened to $PREFIX?

libsrc/nes/Makefile.inc
silverdr: Something like:
```
../target/nes/drv/tgi/nes-64-56-2.tgi:  \
                ../libwrk/nes/clrscr.o  \
                ../libwrk/nes/cputc.o   \
                ../libwrk/nes/get_tv.o  \
                ../libwrk/nes/gotoxy.o  \
                ../libwrk/nes/popa.o    \
                ../libwrk/nes/ppu.o     \
                ../libwrk/nes/ppubuf.o  \
                ../libwrk/nes/setcursor.o
```
?

silverdr:
In regards to the `samples` directory: what we try to achieve is that the directory in question ends up being archived in the "zip" target as well as installed in the "install" target. Main Makefile says though:

```
all mostlyclean clean install zip:
	@$(MAKE) -C src    --no-print-directory $@
	@$(MAKE) -C libsrc --no-print-directory $@
	@$(MAKE) -C doc    --no-print-directory $@
```
The cleanest way seem to be simply adding "zip" target to the `samples` dir Makefile. Yet we don't want targets other than "zip" for example enter the `samples` directory, do we? In this case the next that comes to mind is to separate the install and zip targets from the `all mostlyclean clean install zip` set. I somehow don't like this but unless you object/have better ideas, I can proceed this way.

samples/Makefile
silverdr: It doesn't seem to be used in this Makefile. The reason for that seems to be that the section in question actually checks how and where the cc65 executables and libs were installed, if at all, presumably at some point in the past - in order to build the samples. Therefore it is not connected to the $(prefix) used when compiling the cc65 sources.

oliverschmidt: I must admit that I'm not happy with paths like '/usr/local/lib' or '/opt/local/share' in this Makefile. However I obviously didn't care enough so far to improve it so it seems appropriate to keep it that way - at least for now...

oliverschmidt:
If I don't misunderstand things then there's already a pattern for handling that:

Check out https://github.com/cc65/cc65/blob/master/doc/Makefile#L12 and you see that it has an empty 'all' and 'mostlyclean' goal.

So from my perspective the way to go is to go ahead and add the samples Makefile to the list of three Makefiles in the main Makefile - and have empty goals in the samples Makefile.

libsrc/nes/Makefile.inc
greg-king5: My original thought was just a small (four column) indent; but, your right-edge alignment looks nicer.

samples/Makefile
silverdr: OK. FYI - I looked at it from the same angle and eventually used the same approach that's already utilised in this Makefile.

libsrc/nes/Makefile.inc
silverdr: OK - done the way above.

silverdr:
Yes, but the problem is that "all" is already present and non-empty in the `samples` Makefile. And since it's not part of the regular `make all` - we don't want to invoke it. But yes, we can tweak the targets naming in the `samples` Makefile and rename current "all" target to e. g. "samples" and add empty "all" target there.

samples/Makefile
oliverschmidt: > [...] and eventually used the same approach that's already utilised in this Makefile

... and that's the very reason why I don't oppose to it in the context of this pull request ;-)

oliverschmidt:
> But yes, we can tweak the targets naming in the samples Makefile and rename current "all" target to e. g. "samples" and add empty "all" target there.

That's the approach I'd suggest.

libsrc/Makefile
```diff
@@ -91,7 +93,7 @@ define INSTALL_recipe
 
 $(if $(prefix),,$(error variable `prefix' must be set))
```
silverdr: BTW - didn' t notice it earlier - but isn't that kind of broken here? I mean it requires $prefix but then uses $DESTDIR, which it doesn't require?

libsrc/Makefile
```diff
@@ -91,7 +93,7 @@ define INSTALL_recipe
...
 $(INSTALL) -d $(DESTDIR)$(datadir)/$(dir)
```
silverdr: Moreover some lines earlier $datadir is set as `datadir = $(prefix)/share/cc65`, which means that if somebody supplies $DESTDIR the output will go lalaland direction as $prefix can't be empty at the same time

libsrc/Makefile
oliverschmidt: > [...] isn't that kind of broken here?

No, it's as it is intended to be.

Following standards (https://www.gnu.org/prep/standards/html_node/DESTDIR.html) $DESTDIR is optional.

However in constrast to standards (https://www.gnu.org/prep/standards/html_node/Directory-Variables.html) $prefix isn't optional, therefore the explicit check.

libsrc/Makefile
oliverschmidt: I can't follow you here even after reading your comment several times :-(

libsrc/Makefile
silverdr: But the problem is that putting aside the check, it breaks when one supplies the DESTDIR.

libsrc/Makefile
silverdr: I am sorry for not being clear enough. Let's say that we have $prefix set to - say - "/usr/local". This means $datadir will become `/usr/local/share/cc65`. That's fine. Now we supply `DESTDIR=/my/dest/dir` as specified in https://www.gnu.org/prep/standards/html_node/DESTDIR.html
We end up with `install` lines trying to install things in `/my/dest/dir/usr/local/share/cc65`, which I thought was wrong but apparently that's how it should work as this is similar to the example given in the above URL.

spiro-trikaliotis: The idea behind $(destdir) is to "prefix" the directory with a path that is not part of the installation itself. Any absolute path in the compiled executables, documentation or anything like that is *not* allowed to use $(destdir).

Look at it as follows: If you provide a $(destdir), and you chroot $(destdir) after the build, the installation should work. (Of course, you would need the required environment, too).

I need to use it in order to build the DEBian packages: The debian build environment sets $(destdir) to a temporary directory, installs the package there, and packs that directory afterwards. This way, you have a clean environment.

If $(destdir) does not work, you have to do much more steps. I know because I had to do it for some years for VICE, which did not support $(destdir) at that time. I believe I had fixed it before I left the team, though.

silverdr:
Targets in `samples/Makefile` adjusted.

@spiro-trikaliotis : thank you for additional explanations.

oliverschmidt:
I don't see open issues anymore with this pull request. If nobody comes up with new aspects I'll merge it tomorrow.

@silverdr: Thanks for you patience with all the complains ;-)

silverdr:
@oliverschmidt: NP at all - I happen to know that this is the way things should work with reviews ;-)

One thing that comes to my mind - since we dropped the transitional clean line, some people might be left out with dangling dirs. It might possibly be good to somehow announce the change and that they need to clean their checkouts or re-clone.

oliverschmidt:
> It might possibly be good to somehow announce the change and that they need to clean their checkouts or re-clone.

Thanks for the hint - will do.

samples/Makefile
```diff
@@ -105,8 +108,11 @@ EXELIST	=	ascii		\
 # --------------------------------------------------------------------------
 # Rules to make the binaries
 
-.PHONY:	all
-all:	$(EXELIST)
+.PHONY: all samples
+all:
+
+samples:
+	$(EXELIST)
```
greg-king5: We missed a couple of things:

- `$(EXELIST)` must be prerequisites, not a recipe.
- The prerequisite for `samples.d64` must be changed to `samples`.

silverdr:
On it.

silverdr:
Done. Seems I "corrected" with $(EXELIST) before commiting.

PR #289, opened 2016-05-15
Update the missing entries in the kernel jump table for the Vic20 wit…

OzHawk:
…h the actual function addresses.

The Vic20 does not have kernal table entries for the following functions.

;-----------------------------------------------------------------------------
; Functions which are not in the kernal jump table for VIC-20 but are for C64

CINT        := $E518
IOINIT      := $FDF9
RAMTAS      := $FD8D

All other kernal entries are the same as the C64, however, without this change, the startup code fails.

Without this change the vic20.lib builds incorrectly.

### DIFF
--- a/libsrc/vic20/kernal.s
+++ b/libsrc/vic20/kernal.s
@@ -47,9 +47,9 @@
 ;-----------------------------------------------------------------------------
 ; All functions are available in the kernal jump table
 
-CINT            = $E518         ; No entries are in the kernal jump table for these functions.
-IOINIT          = $FDF9         ; The entries point directly to the function.
-RAMTAS          = $FD8D         ;
+CINT            = $E518         ; No entries are in the kernal jump table of the Vic20 for these three (3) functions.
+IOINIT          = $FDF9         ; The entries for these functions have been set to point directly to the functions 
+RAMTAS          = $FD8D         ; in the kernal to maintain compatibility with the other Commodore platforms.
 RESTOR          = $FF8A
 VECTOR          = $FF8D
 SETMSG          = $FF90
--- a/libsrc/vic20/kernal.s
+++ b/libsrc/vic20/kernal.s
@@ -47,9 +47,9 @@
 ;-----------------------------------------------------------------------------
 ; All functions are available in the kernal jump table
 
-CINT            = $FF81
-IOINIT          = $FF84
-RAMTAS          = $FF87
+CINT            = $E518         ; No entries are in the kernal jump table for these functions.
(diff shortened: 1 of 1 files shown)


libsrc/vic20/kernal.s
```diff
@@ -47,9 +47,9 @@
 ;-----------------------------------------------------------------------------
 ; All functions are available in the kernal jump table
 
-CINT            = $FF81
-IOINIT          = $FF84
-RAMTAS          = $FF87
+CINT            = $E518         ; No entries are in the kernal jump table for these functions.
+IOINIT          = $FDF9         ; The entries point directly to the function.
+RAMTAS          = $FD8D         ;
```
greg-king5: I think that the second comment should be moved down to the third line.  Otherwise, someone might think that the third semicolon is a stray (it might be removed).  Also, I think that the word "function" in that line should be plural ("functions").

Other than those two little things, this PR is a "no-brainer"; it can be merged, as soon as those things are changed.

oliverschmidt:
@OzHawk: Thanks for contributing to cc65@GitHub :-)

libsrc/vic20/kernal.s
```diff
@@ -47,9 +47,9 @@
 ;-----------------------------------------------------------------------------
 ; All functions are available in the kernal jump table
 
-CINT            = $FF81
-IOINIT          = $FF84
-RAMTAS          = $FF87
+CINT            = $E518         ; No entries are in the kernal jump table of the Vic20 for these three (3) functions.
+IOINIT          = $FDF9         ; The entries for these functions have been set to point directly to the functions 
+RAMTAS          = $FD8D         ; in the kernal to maintain compatibility with the other Commodore platforms.
```
OzHawk: Thanks for the prompt response Greg.  I've updated the comment with a bit more detail explaining the difference to the other Commodore platforms.

silverdr:
A minor thing: I believe a good practice is to write KERNAL in all caps. Like:
```
; No entries are in the KERNAL jump table of the Vic20 for these three (3) functions
```
That way it is clear that we talk about the Commodore OS and not a misspelled "kernel". It also helps to avoid mistakes and inconsistencies like the one between the title of this particular Pull Request and its contents.

OzHawk:
Silverdr- thanks for the comment. I hadn't even realised that I had used the incorrect term in the title. The patch is already closed so I won't change it now but will keep it in mind for any further updates.

silverdr:
@OzHawk it's easy and common not to notice those - been there, seen that myself ;-) The reason why I also eventually started to use all caps as advised by someone.

Issue #288, opened 2016-05-15
const pointer using unsigned 8 bit offset is very slow.


janrinze:
Using a 8bit offset to a const pointer is not optimized.

#define Source ((char *)#0x6000;
#define Dest ((char *)#0x6100;

char index=0;
do {
   Dest[index]=Source[index];
   index++;
} while (index);

Expected code:

ldy#0
.loop
 lda 0x6000,y
sta 0x6100,y
iny;
bne loop

result from compiler:

L0004:	ldx     #$61
	lda     #$00
	clc
	adc     L0002
	bcc     L000A
	inx
L000A:	jsr     pushax
	ldx     #$60
	lda     #$00
	ldy     L0002
	sta     ptr1
	stx     ptr1+1
	lda     (ptr1),y
	ldy     #$00
	jsr     staspidx
	lda     L0002
	clc
	adc     #$01
	sta     L0002
	lda     L0002
	bne     L0004

Now this is seriously slow in comparison.

Am I doing something wrong or is this a bug?

Best regards,
Jan Rinze.

MonteCarlos:
you forget that index can be negative in principle and then a simple lda 
xxxx,y doesnt work. i think the compiler code was designed to work for 
several types of the index variable and a specialized version for "char" 
is not available currently.
also it's really difficult for the compiler to detect that the index for 
source and destination are the same. therefore it is understandable that 
the code is not such trivial as expected from the first glance.

Am 15.05.2016 um 11:47 schrieb janrinze:

janrinze:
The index is a unsigned char and thus cannot be negative. The compiler seems to understand that very well. (see: lda (ptr1),y ) It seems not to notice that it need not setup the ptr1 in the first place but instead could have used the constant directly.  

If we help the compiler by adding our own temporary char t it seems to smarten up a little:

int main()
{
#define Source ((char *) 0x6000)
#define Dest ((char *) 0x6100)

    char index=0;

    do {
        char t=Source[index];
        Dest[index]=t;
        index++;
    } while (index);
    return 0;
}

compiled with optimization on will translate to:
.proc   _main: near

.segment        "CODE"

        lda     #$00
        jsr     pusha
L0003:  ldx     #$60
        lda     #$00
        tay
        clc
        adc     (sp),y
        bcc     L0009
        inx
L0009:  sta     ptr1
        stx     ptr1+1
        lda     (ptr1),y
        jsr     pusha
        ldy     #$01
        lda     (sp),y
        tax
        dey
[...]

PR #294, opened 2016-05-23
Make a small optimization to some library functions' inlined versions.

greg-king5:
Calls to memcpy() and memset() will compile to optimum code when the length argument is constant, and 256 or less.

(This patch was inspired by issue #288.)

### DIFF
--- a/src/cc65/codegen.c
+++ b/src/cc65/codegen.c
@@ -55,6 +55,7 @@
 #include "global.h"
 #include "segments.h"
 #include "stackptr.h"
+#include "stdfunc.h"
 #include "textseg.h"
 #include "util.h"
 #include "codegen.h"
@@ -4241,7 +4242,7 @@ void g_initauto (unsigned Label, unsigned Size)
         AddCodeLine ("lda %s,y", GetLabelName (CF_STATIC, Label, 0));
         AddCodeLine ("sta (sp),y");
         AddCodeLine ("iny");
-        AddCodeLine ("cpy #$%02X", (unsigned char) Size);
+	AddCmpCodeIfSizeNot256 ("cpy #$%02X", Size);
         AddCodeLine ("bne %s", LocalLabelName (CodeLabel));
     }
 }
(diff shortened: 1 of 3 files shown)


src/cc65/codegen.c
```diff
@@ -4241,7 +4241,9 @@ void g_initauto (unsigned Label, unsigned Size)
         AddCodeLine ("lda %s,y", GetLabelName (CF_STATIC, Label, 0));
         AddCodeLine ("sta (sp),y");
         AddCodeLine ("iny");
-        AddCodeLine ("cpy #$%02X", (unsigned char) Size);
+        if (Size != 256) {
+            AddCodeLine ("cpy #$%02X", Size);
+        }
```
spiro-trikaliotis: Wouldn't MaybeCodeLine() work here, too?

src/cc65/codegen.c
```diff
@@ -4266,10 +4268,12 @@ void g_initstatic (unsigned InitLabel, unsigned VarLabel, unsigned Size)
         AddCodeLine ("lda %s,y", GetLabelName (CF_STATIC, InitLabel, 0));
         AddCodeLine ("sta %s,y", GetLabelName (CF_STATIC, VarLabel, 0));
         AddCodeLine ("iny");
-        AddCodeLine ("cpy #$%02X", (unsigned char) Size);
+        if (Size != 256) {
+            AddCodeLine ("cpy #$%02X", Size);
+        }
```
spiro-trikaliotis: and, again, MaybeCodeLine()?

src/cc65/stdfunc.c
```diff
@@ -185,6 +185,15 @@ static void ParseArg (ArgDesc* Arg, Type* Type)
 
 
 
+static void MaybeCodeLine (const char* Code, long IVal)
```
spiro-trikaliotis: Of course, "static" must be removed for MaybeCodeLine() to be usable from the other functions above, too.

For me, the name MaybeCodeLine() is not good, because it does not describe what it actually does. Either use it generically ( void MaybeCodeLine(int GenerateLine, const char* Code, ...) and use it like "if (GenerateLine) { /* printf()-style handling of the arguments)", or rename it to something like "ConditionallyCompareToValueIfNotZero()".

greg-king5:
OK.  I renamed the function to `AddCmpCodeIfSizeNot256()`; and, used it in `codegen.c` too.

PR #295, opened 2016-05-24
Atari: add command line support for XDOS

groessler:
With this change the runtime library parses the command line arguments correctly on XDOS. Tested with XDOS 2.4.

### DIFF
--- a/asminc/atari.inc
+++ b/asminc/atari.inc
@@ -997,12 +997,15 @@ diopp_size   = 5        ; size of structure
 ; VALUES for dos_type
 ;-------------------------------------------------------------------------
 
-ATARIDOS     = 0
-SPARTADOS    = 1
-OSADOS       = 2        ; OS/A+
-MYDOS        = 3
-XDOS         = 4
+SPARTADOS    = 0
+OSADOS       = 1        ; OS/A+
+XDOS         = 2
+ATARIDOS     = 3
+MYDOS        = 4
 NODOS        = 255
+; The DOSes with dos_type below or equal MAX_DOS_WITH_CMDLINE do support
+; command line arguments.
+MAX_DOS_WITH_CMDLINE = XDOS
 
 ;-------------------------------------------------------------------------
 ; XDOS defines (version 2.4, taken from xdos24.pdf)
--- a/asminc/atari.inc
+++ b/asminc/atari.inc
@@ -1011,16 +1011,16 @@ MAX_DOS_WITH_CMDLINE = XDOS
 ; XDOS defines (version 2.4, taken from xdos24.pdf)
 ;-------------------------------------------------------------------------
 
-XOPT         = $70B     ; XDOS options
-XCAR         = $70C     ; XDOS cartridge address (+ $70D)
-XPAT         = $86F     ; XDOS bugfix and patch number
-XVER         = $870     ; XDOS version number
-XFILE        = $87D     ; XDOS filename buffer
-XLINE        = $880     ; XDOS DUP input line
-XGLIN        = $871     ; get line
-XSKIP        = $874     ; skip parameter
-XMOVE        = $877     ; move filename
-XGNUM        = $87A     ; get number
+XOPT         = $070B    ; XDOS options
+XCAR         = $070C    ; XDOS cartridge address (+ $70D)
+XPAT         = $086F    ; XDOS bugfix and patch number
+XVER         = $0870    ; XDOS version number
+XFILE        = $087D    ; XDOS filename buffer
+XLINE        = $0880    ; XDOS DUP input line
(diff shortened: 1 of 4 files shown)


asminc/atari.inc
```diff
@@ -997,12 +997,30 @@ diopp_size   = 5        ; size of structure
...
-ATARIDOS     = 0
-SPARTADOS    = 1
-OSADOS       = 2        ; OS/A+
-MYDOS        = 3
-XDOS         = 4
+SPARTADOS    = 0
+OSADOS       = 1        ; OS/A+
+XDOS         = 2
+ATARIDOS     = 3
+MYDOS        = 4
```
spiro-trikaliotis: Is there a reason why the enum type defines had to be rearranged? And again in the .h file below.

include/atari.h
```diff
@@ -261,11 +261,11 @@ extern void atrx15p2_tgi[];
...
-#define ATARIDOS    0
-#define SPARTADOS   1
-#define OSADOS      2
-#define MYDOS       3
-#define XDOS        4
+#define SPARTADOS   0
+#define OSADOS      1
+#define XDOS        2
+#define ATARIDOS    3
+#define MYDOS       4
```
spiro-trikaliotis: Again, here, the rearrangement looks arbitrary. Is there a technical reason, or does it just look better?

(Note that I know exactly nothing about the Atari platform, so, it may be obvious to you, but it is not for me)

silverdr: A valid question IMHO. I *think* the original arrangement reflected more or less the historical development (although I am not sure), which makes it "better" for me than an arbitrary arrangement. Besides, are we sure that this change doesn't break existing code? To my Atari-untrained eye it looks like it might, therefore justification for this change is even more important.

oliverschmidt: @spiro-trikaliotis & @silverdr: Doesn't the fragment

`; The DOSes with dos_type below or equal MAX_DOS_WITH_CMDLINE do support`
`; command line arguments.`
`MAX_DOS_WITH_CMDLINE = XDOS`

answer that question?

groessler:
The reason for the change was to simplify the test if it's a "command line DOS" in initmainargs, therefore saving some bytes of code.
If everybody is using the symbolic names, it should not be incompatible. But you're right, it might not be a good idea. I can revert that part.

oliverschmidt:
From my perspective the other reviewers (thanks !) "only" wanted to say that without any reason the symbols shouldn't be changed - and right they are. However now we all see the reason. I personally don't see why we should spend any bytes making the decision logic larger. Any cc65 program is supposed to make use of the given symbols, doesn't it?

groessler:
Ok, then I'll leave it as it is.

silverdr:
True.

> On 2016-05-24, at 11:51, Oliver Schmidt <notifications@github.com> wrote:
> 
> From my perspective the other reviewers (thanks !) "only" wanted to say that without any reason the symbols shouldn't be changed - and right they are. However now we all see the reason. I personally don't see why we should spend any bytes making the decision logic larger. Any cc65 program is supposed to make use of the given symbols, doesn't it?
> 
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> 

-- 
SD!

asminc/atari.inc
```diff
@@ -997,12 +997,30 @@ diopp_size   = 5        ; size of structure
...
+XOPT         = $70B     ; XDOS options
+XCAR         = $70C     ; XDOS cartridge address (+ $70D)
+XPAT         = $86F     ; XDOS bugfix and patch number
+XVER         = $870     ; XDOS version number
+XFILE        = $87D     ; XDOS filename buffer
+XLINE        = $880     ; XDOS DUP input line
+XGLIN        = $871     ; get line
+XSKIP        = $874     ; skip parameter
+XMOVE        = $877     ; move filename
+XGNUM        = $87A     ; get number
```
greg-king5: Style issue: non-zeropage addresses should be written with four hex characters.

groessler: Changed. Where is the official "style guide"?

groessler: I've noticed that you added to the comment in the meanwhile. I've added the XDOS symbols in the same order as they appear in the documentation (xdos24.pdf).

oliverschmidt: > Where is the official "style guide"?

- I presume you asked this question presuming there is none ;-) However there actually is one: https://github.com/cc65/wiki/wiki/C-and-Assembly-sources-style-guide
- The style guide is far from complete and doesn't cover what Greg wrote.

include/atari.h
spiro-trikaliotis: Good point, I totally overlooked that one. Yes, there is the reason I asked for.

PR #298, opened 2016-05-24
make BSS segment optional in atari-cassette.cfg

groessler:
Don't know why it wasn't optional.

Documentation says that this config file can be used for both C and Assembler programs. Assembler programs typically don't have  a BSS segment.


greg-king5:
It wasn't optional because no one noticed.  We didn't notice because ca65 always makes it.

`.code`, `.rodata`, `.data`, and `.bss` are the original segments.  They existed before Uz thought of optional segments.  Therefore, if you don't mention one of those four, then ca65 creates an empty segment for it.  (Strictly speaking, `CODE`, `RODATA`, `DATA`, and `BSS` don't need to be marked as optional.)

PR #303, opened 2016-05-31
Atari: get current drive on XDOS

groessler:


libsrc/atari/getdefdev.s
```diff
@@ -68,16 +69,23 @@ crvec:  jsr     $FFFF           ; will be set to crunch vector
...
         rts
 
+; XDOS version
+
+xdos:   lda     XDEFDEV
+        bne     done
+
         .data
 
+crvec:	.byte	$4C,$FF,$FF     ; will be set to crunch vector
```
oliverschmidt: What is the reason to not write 'jmp $FFFF' here?

The decision of mnemonics or .byte is totally independent from the segment. The cc65 code base already contains lots of places where mnemonics are used in the .data segment (and places where .byte is used in the .code segment).

groessler: > What is the reason to not write 'jmp $FFFF' here?

No real reason.

> The decision of mnemonics or .byte is totally independent from the segment. The cc65 code base already contains lots of places where mnemonics are used in the .data segment (and places where .byte is used in the .code segment).

Yes. So one is as good as the other, isn't it?

If you want, I can change it to "jmp".

oliverschmidt: > Yes. So one is as good as the other, isn't it?

No. If it were, we wouldn't need mnemonics and write all our code using .byte ;-))

> If you want, I can change it to "jmp".

Please do so - thanks in advance.

oliverschmidt: Trying to add a more helpful comment:

The idea is to use mnemonics everywhere the generated bytes are fetched/executed by the CPU and .byte everywhere the generated bytes are read/written by the CPU.

groessler: > No. If it were, we wouldn't need mnemonics and write all our code using .byte ;-))

Hey, c'mon, everybody reading the library soure code knows (or is supposed to know, at least) the meaning of $4C, isn't (s)he?  :-)

But I'll change it....

groessler: I've updated the pull request....

PR #309, opened 2016-06-11
Allow character code 0x00 to be remapped with other character codes.

greg-king5:
This PR gives Christian [what he wants](https://sourceforge.net/p/cc65/mailman/message/35130706/).  I do not see that it causes any problems for the tools.  [Of course, he could give himself a big headache with the Atari target's system string functions; but, I assume that he knows what he's doing.  ;-) ]


MonteCarlos:
doesn't break that all print functions if terminating zero is remapped?

Am 11.06.2016 um 16:13 schrieb Oliver Schmidt:

greg-king5:
> Doesn't that break all print functions if terminating zero is remapped?

Yes, that's one of the reasons why I mentioned the "big headache".  But, I assume that he will use custom display functions with remapped characters (also, remapping is limited to an individual source file).

PR #310, opened 2016-06-13
Implement exec() for Atari XDOS.

groessler:
- Adds new ENOEXEC error code, also used by Apple2 targets.
- Maximum command line length is 40, incl. program name. This is
  an XDOS restriction.
- testcode/lib/tinyshell.c has been extended to be able to run
  programs.

@oliverschmidt: This version doesn't yet include support for longer command lines as we were talking about in private email.

### DIFF
--- a/libsrc/atari/exec.s
+++ b/libsrc/atari/exec.s
@@ -33,10 +33,8 @@ notsupp:lda     #ENOSYS         ; "unsupported system call"
         .byte   $2C             ; bit opcode, eats the next 2 bytes
 noiocb: lda     #EMFILE         ; "too many open files"
         jsr     incsp2          ; clean up stack
-seterr: jsr     __directerrno
-        lda     #$FF
-        tax
-        rts                     ; return -1
+seterr: jmp     __directerrno
+
 
 ; entry point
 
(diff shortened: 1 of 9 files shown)


libsrc/atari/exec.s
```diff
@@ -0,0 +1,207 @@
...
+        .code
+
+notsupp:lda     #ENOSYS         ; "unsupported system call"
+        .byte   $2C             ; bit opcode, eats the next 2 bytes
+noiocb: lda     #EMFILE         ; "too many open files"
+        jsr     incsp2          ; clean up stack
+seterr: jsr     __directerrno
+        lda     #$FF
+        tax
+        rts                     ; return -1
```
greg-king5: `__directerrno` returns `-1`; so, you can just jump to it here.

testcode/lib/tinyshell.c
```diff
@@ -340,6 +354,23 @@ static void cmd_rename(void)
...
+
+    progname = strtok(args, " \t\n");
+    if (! progname) {
+        puts("usage: exec <progname> [arguments]");
+        return;
+    }
+    arguments = strtok(NULL, "");
+
+    /*printf("exec: %s %s\n", progname, arguments ? arguments : "");*/
+    st = exec(progname, arguments);
```
greg-king5: Assigning the return value to `st` is pointless -- it always is `-1`.  You just should cast it to `void`.

groessler:
Thanks for the tips. I've updated the pull request.

PR #313, opened 2016-06-18
Fix the Plus/4 (and C16) joystick driver.

greg-king5:
It now chooses a stick correctly.  And, it now reads the stick's fire button.

This fixes https://sourceforge.net/p/cc65/mailman/cc65-devel/thread/f2f66618-7a40-3d72-6c41-1b43216a0ed0%40suomi24.fi/#msg35162158 -- the parts that can be fixed.


oliverschmidt:
Thanks :-)

PR #315, opened 2016-06-23
Fix typo in samples makefile

skry:


oliverschmidt:
Thanks for the fix :-)

PR #318, opened 2016-06-30
Add support for HuC6280 zero-page at $2000 instead of $0000

jbrandwood:
CA65 currently assumes that zero-page/direct-page references resolve to absolute address $0000-$00FF.

This isn't correct on the HuC6280 where zero-page/direct-page references resolve to absolute addresses $2000-$20FF.

This leads CA65 to generate inefficient/wrong code for instructions referencing addresses $0000-$00FF and $2000-$20FF on the HuC6280.

This patch fixes this by allowing CA65 to differentiate between instructions with one-byte immediate operands that are always $0000-$00FF (handled by the existing old Emit1() function), and instructions that have one-byte direct-page/zero-page references where the top-byte of the address is different on different processors (handled by the new EmitZP() function).

EmitZP() is just a copy of the existing Emit1() function but with the check for the top-byte changing depending upon the currently-selected processor.

### DIFF
--- a/src/ca65/instr.c
+++ b/src/ca65/instr.c
@@ -1054,7 +1054,11 @@ static void EmitCode (EffAddr* A)
             break;
 
         case 1:
-            Emit1 (A->Opcode, A->Expr);
+            if (A->AddrModeBit & (AM65_ALL_ZP | AM65_DIR_IND_Y | AM65_DIR_IND_LONG | AM65_DIR_IND_LONG_Y)) {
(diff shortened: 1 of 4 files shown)


jbrandwood:
Whoops, looks like Travis failed ... building with mingw-w64 instead of Visual Studio now to track down what's gone wrong.

jbrandwood:
Oh, dear. It looks like there's a deeper issue with CA65's expression handling that I missed.

I'm withdrawing the pull request until this gets fixed.

Issue #235, opened 2015-11-26
ca65: expressions don't search for symbols in outer scopes


tokumaru-x:
```
.scope
	Symbol = 4
	.scope
		lda #Symbol ;<- this works
		.if Symbol = 4 ;<- this generates an error
			.out "Symbol is 4."
		.endif
	.endscope
.endscope
```

fo-fo:
This seems to be a "feature" that everybody runs into eventually. Strictly speaking it's not a bug. Ca65 assumes that `Symbol` might get defined later in the scope so it doesn't see it as a constant (it can't look forward to see whether it will get defined or not). The `lda` works because it doesn't require a constant.

In my opinion, since `.if` generates a compilation-stopping error in this case, it would be better behavior for it to use the value from the outer scope, and set up an assert to make sure that the symbol doesn't get redefined later in the same scope (if it does, that would generate an error because "wrong" value got used for the `.if` condition). However there might be some problems with this approach that I'm unaware of.

mrdudz:
On Thursday 26 November 2015, 05:03:51 fo-fo <notifications@github.com> wrote:
> This seems to be a "feature" that everybody runs into eventually. Strictly
> speaking it's not a bug. Ca65 assumes that `Symbol` might get defined later
> in the scope so it doesn't see it as a constant (it can't look forward to
> see whether it will get defined or not). The `lda` works because it doesn't
> require a constant.
> 
> In my opinion, since `.if` generates a compilation-stopping error in this
> case, it would be better behavior for it to use the value from the outer
> scope, and set up an assert to make sure that the symbol doesn't get
> redefined later in the same scope (if it does, that would generate an error
> because "wrong" value got used for the `.if` condition). However there
> might be some problems with this approach that I'm unaware of.

forward references in what essentially resembles a preprocessor expression 
would be totally NOT what i'd expect though.

-- 

http://www.hitmen-console.org    http://magicdisk.untergrund.net
http://www.pokefinder.org        http://ar.pokefinder.org

I have always wished that my computer would be as easy to use as my telephone. 
My wish has come true. I no longer know how to use my telephone. 
<Bjarne Stroustrup>

fo-fo:
> forward references in what essentially resembles a preprocessor expression
> would be totally NOT what i'd expect though.

[...]

PR #319, opened 2016-07-04
Add support for HuC6280 zero-page at $2000 instead of $0000

jbrandwood:
CA65 currently assumes that zero-page references resolve to absolute address $0000-$00FF.

This isn't correct on the HuC6280 where zero-page references resolve to absolute addresses $2000-$20FF.

This leads CA65 to generate inefficient/wrong code for instructions referencing addresses $0000-$00FF and $2000-$20FF on the HuC6280.

This patch fixes this by allowing CA65 to differentiate between instructions with one-byte immediate operands that are always $0000-$00FF (handled by the existing old Emit1() function), and instructions that have one-byte zero-page references where the top-byte of the address is different on different processors (handled by the new EmitZP() function).

EmitZP() is just a copy of the existing Emit1() function but with the check for the top-byte changing depending upon the currently-selected processor.

This is a "fixed" version of my previous Pull Request #318.

CA65 has way too much code/logic in the flow that requires zero-page to be at $0000, that I've been unable to tell it that the HuC6280 zero-page is at $2000 and NOT at $0000.

Without a major rewrite of some core internal CA65 logic, the best that I've been able to do is to tell CA65 that the HuC6280 ALSO has zero-page at $2000.

The only accesses to $0000-$00FF in a normal PC Engine are to the VDC hardware. This hardware is mirrored throughout the $0000-$03FF region, so I've just chosen to change the equates that are used in PCE.INC to one of the mirrored locations in order to avoid confusing CA65 with instructions that access $0000-$00FF.

While not perfect, this solves the main issue of the the HuC6280 using $2000-$20FF to access zero-page, and it does so without making any major changes in CA65 that could break other platforms.

### DIFF
--- a/asminc/pce.inc
+++ b/asminc/pce.inc
@@ -34,11 +34,11 @@ VDC_LENR        = 18            ; (DMA) Length Register
 VDC_SATB        = 19            ; Sprite Attribute Table
 
 ; VDC port
-; Note: absolute addressing mode must be used when writing to this port
+; Note: use mirror location to avoid CA65 zero-page logic
 
-VDC_CTRL        = $0000
-VDC_DATA_LO     = $0002
-VDC_DATA_HI     = $0003
+VDC_CTRL        = $0100
+VDC_DATA_LO     = $0102
+VDC_DATA_HI     = $0103
 
 ; huc6260 - Video Color Encoder (vce)
 
--- a/libsrc/pce/crt0.s
+++ b/libsrc/pce/crt0.s
@@ -74,7 +74,7 @@ start:
         ;tam     #%10000000      ; e000-fFFF  hucard/syscard bank 0
 
         ; Clear work RAM (2000-3FFF)
-        stz     <$00
+        stz     $2000
         tii     $2000, $2001, $1FFF
 
         ; Initialize hardware
--- a/libsrc/pce/vdc.s
+++ b/libsrc/pce/vdc.s
@@ -7,7 +7,7 @@ HIRES   = 1
         .export         vdc_init
 
 vdc_init:
-        lda     a:VDC_CTRL
+        lda     VDC_CTRL
 
         VREG    $00, $0000      ; MAWR
         VREG    $01, $0000      ; MARR
@@ -37,5 +37,5 @@ vdc_init:
 
 .endif
 
-        lda     a:VDC_CTRL
+        lda     VDC_CTRL
         rts
--- a/src/ca65/instr.c
+++ b/src/ca65/instr.c
@@ -1054,7 +1054,11 @@ static void EmitCode (EffAddr* A)
             break;
 
         case 1:
-            Emit1 (A->Opcode, A->Expr);
+            if (A->AddrModeBit & (AM65_ALL_ZP | AM65_DIR_IND_Y | AM65_DIR_IND_LONG | AM65_DIR_IND_LONG_Y)) {
+                EmitZP (A->Opcode, A->Expr);
+            } else {
+                Emit1 (A->Opcode, A->Expr);
+            }
             break;
 
         case 2:
--- a/src/ca65/instr.c
+++ b/src/ca65/instr.c
@@ -1054,7 +1054,7 @@ static void EmitCode (EffAddr* A)
             break;
 
         case 1:
-            if (A->AddrModeBit & (AM65_ALL_ZP | AM65_DIR_IND_Y | AM65_DIR_IND_LONG | AM65_DIR_IND_LONG_Y)) {
+            if (A->AddrModeBit & AM65_ALL_ZP) {
                 EmitZP (A->Opcode, A->Expr);
             } else {
                 Emit1 (A->Opcode, A->Expr);
--- a/src/ca65/instr.h
+++ b/src/ca65/instr.h
@@ -93,7 +93,7 @@
 #define AM65_SET_ABS    (AM65_ABS | AM65_ABS_X)
 
 /* Bitmask for all ZP operations */
-#define AM65_ALL_ZP     (AM65_DIR | AM65_DIR_X | AM65_DIR_Y | AM65_DIR_IND | AM65_DIR_X_IND)
+#define AM65_ALL_ZP     (AM65_DIR | AM65_DIR_X | AM65_DIR_Y | AM65_DIR_IND | AM65_DIR_X_IND | AM65_DIR_IND_Y | AM65_DIR_IND_LONG | AM65_DIR_IND_LONG_Y)
 
 /* Bitmask for all ABS operations */
 #define AM65_ALL_ABS    (AM65_ABS | AM65_ABS_X | AM65_ABS_Y | AM65_ABS_IND | AM65_ABS_X_IND)
--- a/src/ca65/objcode.c
+++ b/src/ca65/objcode.c
@@ -36,6 +36,9 @@
 #include <string.h>
 #include <errno.h>
 
+/* common */
+#include "cpu.h"
+
 /* cc65 */
 #include "error.h"
 #include "fragment.h"
@@ -129,6 +132,38 @@ void Emit3 (unsigned char OPC, ExprNode* Expr)
 {
     Emit0 (OPC);
     EmitFarAddr (Expr);
+}
+
+
+
+void EmitZP (unsigned char OPC, ExprNode* Value)
+/* Emit an instruction with an one byte direct-page argument */
+{
+    long V;
+    Fragment* F;
+
+    if (IsEasyConst (Value, &V)) {
+        /* Must be in byte range */
+        if ((CPU != CPU_HUC6280 && (V & ~0xFFL) != 0x0000) ||
+            (CPU == CPU_HUC6280 && (V & ~0xFFL) != 0x2000)) {
+            Error ("Range error (0x%lx not in zero-page)", V);
+        }
+
+        /* Create a literal fragment */
+        F = GenFragment (FRAG_LITERAL, 2);
+        F->V.Data[0] = OPC;
+        F->V.Data[1] = (unsigned char) V;
+        FreeExpr (Value);
+
+    } else {
+
+        /* Emit the opcode */
+        Emit0 (OPC);
+
+        /* Emit the argument as an expression */
+        F = GenFragment (FRAG_EXPR, 1);
+        F->V.Expr = Value;
+    }
 }
 
 
--- a/src/ca65/objcode.c
+++ b/src/ca65/objcode.c
@@ -137,15 +137,15 @@ void Emit3 (unsigned char OPC, ExprNode* Expr)
 
 
 void EmitZP (unsigned char OPC, ExprNode* Value)
-/* Emit an instruction with an one byte direct-page argument */
+/* Emit an instruction with an one byte zero-page argument */
 {
     long V;
     Fragment* F;
 
     if (IsEasyConst (Value, &V)) {
         /* Must be in byte range */
-        if ((CPU != CPU_HUC6280 && (V & ~0xFFL) != 0x0000) ||
-            (CPU == CPU_HUC6280 && (V & ~0xFFL) != 0x2000)) {
+        if (((V & ~0xFFL) != 0x0000L) &&
+            (CPU != CPU_HUC6280 || (V & ~0xFFL) != 0x2000L)) {
             Error ("Range error (0x%lx not in zero-page)", V);
         }
 
--- a/src/ca65/objcode.h
+++ b/src/ca65/objcode.h
@@ -62,6 +62,9 @@ void Emit2 (unsigned char OPC, ExprNode* Value);
 void Emit3 (unsigned char OPC, ExprNode* Expr);
 /* Emit an instruction with a three byte argument */
 
+void EmitZP (unsigned char OPC, ExprNode* Value);
+/* Emit an instruction with an one byte direct-page argument */
+
 void EmitSigned (ExprNode* Expr, unsigned Size);
 /* Emit a signed expression with the given size */
 
--- a/src/ca65/studyexpr.c
+++ b/src/ca65/studyexpr.c
@@ -443,11 +443,13 @@ static void StudyExprInternal (ExprNode* Expr, ExprDesc* D);
 
 
 
-static unsigned char GetConstAddrSize (long Val)
+// static unsigned char GetConstAddrSize (long Val)
+unsigned char GetConstAddrSize (long Val)
 /* Get the address size of a constant */
 {
-    if ((CPU != CPU_HUC6280 && (Val & ~0xFFL) == 0x0000) ||
-        (CPU == CPU_HUC6280 && (Val & ~0xFFL) == 0x2000)) {
+    if ((Val & ~0xFFL) == 0) {
+        return ADDR_SIZE_ZP;
+    } else if (CPU == CPU_HUC6280 && (Val & ~0xFFL) == 0x2000L) {
         return ADDR_SIZE_ZP;
     } else if ((Val & ~0xFFFFL) == 0) {
         return ADDR_SIZE_ABS;
--- a/src/ca65/studyexpr.c
+++ b/src/ca65/studyexpr.c
@@ -40,6 +40,7 @@
 #include "debugflag.h"
 #include "shift.h"
 #include "xmalloc.h"
+#include "cpu.h"
 
 /* ca65 */
 #include "error.h"
@@ -445,7 +446,8 @@ static void StudyExprInternal (ExprNode* Expr, ExprDesc* D);
 static unsigned char GetConstAddrSize (long Val)
 /* Get the address size of a constant */
 {
-    if ((Val & ~0xFFL) == 0) {
+    if ((CPU != CPU_HUC6280 && (Val & ~0xFFL) == 0x0000) ||
+        (CPU == CPU_HUC6280 && (Val & ~0xFFL) == 0x2000)) {
         return ADDR_SIZE_ZP;
     } else if ((Val & ~0xFFFFL) == 0) {
         return ADDR_SIZE_ABS;
--- a/src/ca65/studyexpr.c
+++ b/src/ca65/studyexpr.c
@@ -443,8 +443,7 @@ static void StudyExprInternal (ExprNode* Expr, ExprDesc* D);
 
 
 
-// static unsigned char GetConstAddrSize (long Val)
-unsigned char GetConstAddrSize (long Val)
+static unsigned char GetConstAddrSize (long Val)
 /* Get the address size of a constant */
 {
     if ((Val & ~0xFFL) == 0) {


groessler:
Just a note (I'm not a committer):

Your change could be an improvement for the PCE, but I think the problem should be solved in a more general way.

The 65816 has the same problem with a modified "direct page" setting. I think ca65 needs a directive to tell it where the "direct page" is located.

Something like this:

.zpaddr $2000
lda $2001  ; assembles to $A5 $01
lda $3001  ; $AD $01 $30
.zpaddr $3000
lda $2001  ; $AD $01 $20
lda $3001  ; $A5 $01

Then the HuC6280 is just a special case where "zpaddr" is always $2000.

Btw., I've searched the internet, but haven't found anything about HuC6280 using $2000 as "zero-page". Can you give a pointer?

jbrandwood:
You can see Charles MacDonald reverse-engineering document here ...
http://cgfm2.emuviews.com/txt/pcetech.txt

The original Hudson datasheets are "out there" if you can find them, but since they were never officially made public in English, I can't give you a link to them.

They were made public in Japanese if you can ever find the Develo Assembler book, but that's really, really hard to find.

*****************

I agree that a nice-and-general direct-page assembler directive would be the preferred way to handle this, but I'm not sure that that will ever happen in CA65.

I tried to "fix" things properly in a variety of ways, but CA65's addressing-mode logic starts falling down badly if you try to make $0000-$00FF NOT be zero-page.

It's baked very, very deeply into the CA65 code, and LD65 doesn't have anything to do link-time direct-page fixups anyway.

IMHO, the way that the addressing-mode logic handles symbols within "proc" scoping would also need to be rewritten.

If someone wants to implement a good long-term general fix, then I expect that they're going to have to spend weeks or months rewriting a whole bunch of CA65's low-level code and deal with lots of library breakage.

Since CC65 doesn't actually generate 65816 code anyway, then I suspect that's too much effort, for too little reward.

oliverschmidt:
Given Greg's recent comment on the mailing list (btw. I personally would like to see that discussion here) saying

>My conclusion is that they are the wrong functions for what you want to
do (but, I haven't found the better places).

I'm very reluctant to consider merging this pull request.

oliverschmidt:
> I agree that a nice-and-general direct-page assembler directive would be the preferred way to handle this, but I'm not sure that that will ever happen in CA65.
I tried to "fix" things properly in a variety of ways, but CA65's addressing-mode logic starts falling down badly if you try to make $0000-$00FF NOT be zero-page.

Maybe I'm missing the point but I understand that you have a hard time making page $00 a "non-zero-page" so you opted to not __move__ the single zero-page from page $00 to page $20 but instead __add__ a second zero-page at $20.

However I don't see why that means that the location of the second zero-page has to be hard-coded. Instead it could be configurable as Christian proposed, couldn't it?

jbrandwood:
Hi Oliver,

> Given Greg's recent comment on the mailing list (btw. I personally would like to see that discussion here) saying
> 
>     My conclusion is that they are the wrong functions for what you want to
>     do (but, I haven't found the better places).

I'll be interested to hear if these "better places" can be found.

The problem (IMHO) is that CA65's method for deciding the width of the addressing-mode is fundamentally inconsistent, and works differently in proc scope vs global scope.

The whole usage of the data size of individual numeric components within an expression to "guess" at the size of the eventual expression seems very dubious to me.

Apparently it seemed equally dubious to Ulrich because there's code in CA65 to throw away the result of that guess, and recalculate the addressing mode depending upon the locations of symbols (import vs importzp), or the final constant value of an expression, which seems like the correct way to figure it out.

Unfortunately, that code doesn't get used inside proc scope.


> However I don't see why that means that the location of the second zero-page has to be hard-coded. Instead it could be configurable as Christian proposed, couldn't it?

Yes, my patch could be made a little more general in that a new directive could be added to allow you to change exactly which page is considered to be the 2nd mirror for the actual direct page addressing, but I don't see that as an improvement.

The patch as it is currently written can only handle direct page addressing that falls on a 256-byte "page" boundary.

That's OK for direct-page addressing at $0000 (like the 6502), or $2000 (like the HuC6280), or even the 6809 (which CA65 doesn't support anyway), but it isn't good enough for the 65816 which has a 16-bit direct-page base address than can be at any byte-boundary location in the bottom 64KB of memory.

IMHO, getting that to work properly would require modifying the instruction's expression to insert a subtraction into every direct-page instruction that's emitted, and deferring the actual calculation until link time.

That's a much deeper patch than I am comfortable in writing for CA65 at this point, and if you already don't like where I'm modifying the code now, then my modifying the expression when a direct-page instruction is emitted is really unlikely to make you happy.

And even then, it would still leave the issue of still having the extra zero-page location at $0000-$00FF, which is actually going to be a serious problem on the 65816 because you are almost certainly going to be having real data located there.

jbrandwood:
> IMHO, getting that to work properly would require modifying the instruction's expression to insert a subtraction into every direct-page instruction that's emitted, and deferring the actual calculation until link time.
> 
> That's a much deeper patch than I am comfortable in writing for CA65 at this point, and if you already don't like where I'm modifying the code now, then my modifying the expression when a direct-page instruction is emitted is really unlikely to make you happy.
> 
> And even then, it would still leave the issue of still having the extra zero-page location at $0000-$00FF, which is actually going to be a serious problem on the 65816 because you are almost certainly going to be having real data located there.
> 

To complicate things further, when "<" is used at the start of an addressing mode, CA65 is treating it as a low-byte expression calculation, and not as an addressing-mode specification.

That's another thing on the list of issues that would have to be worked on in order to get CA65 to understand that direct addressing isn't always at $0000-$00FF.

IMHO, the patch that I've presented here is the smallest and safest way to get the HuC6280 support to work acceptably instead of being just-plain-wrong as it is now.

I can add half-assed support for a sort-of-direct-page for the 65816, but it would be pathetically limited, and it would lead to breakage and unintended side-effects and bugs in 65816 code from the dual zero-pages.

I honestly think that fixing the 65816 is a longer term thing that needs to be looked at separately.

When that is eventually done, then I believe that you're going to end up with different Emit1() and EmitZP() functions anyway, because that is just how the processor works.

"lda #42" has a different meaning to "lda <42", and the assembler needs to know different things about the expression in each case.

Since adding that distinction is the bulk of my patch, then I really don't see that there should be a problem in accepting it; but I'm certainly happy to consider modifications if they look like they'll actually improve the situation.

oliverschmidt:
>> However I don't see why that means that the location of the second zero-page has to be hard-coded. Instead it could be configurable as Christian proposed, couldn't it?


> [...]
That's OK for direct-page addressing at $0000 (like the 6502), or $2000 (like the HuC6280), or even the 6809 (which CA65 doesn't support anyway), but it isn't good enough for the 65816 which has a 16-bit direct-page base address than can be at any byte-boundary location in the bottom 64KB of memory.
[...]
And even then, it would still leave the issue of still having the extra zero-page location at $0000-$00FF, which is actually going to be a serious problem on the 65816 because you are almost certainly going to be having real data located there.

Thanks for the detailed explanation. I now see that my proposal indeed isn't useful.

oliverschmidt:
> That's a much deeper patch than I am comfortable in writing for CA65 at this point, and if you already don't like where I'm modifying the code now, then my modifying the expression when a direct-page instruction is emitted is really unlikely to make you happy.

Maybe a misunderstanding, maybe I'm (again) missing the point, but...

I'm not unhappy regarding the size nor complexity of the change. I'm unhappy because Greg stated that he thinks that the change isn't "at the right place". Maybe a larger, more complex change is better. Let me however be totally clear about that I don't see myself being able to assess this! So what are my criteria for a merge "in the core" of the tool chain?

- People like Greg don't complain about it
- If there's a user visible aspect then that is documented
- Regression tests pass
- Code style fits to the surrounding code

groessler:
On 07/07/16 18:26, jbrandwood wrote:

> That's OK for direct-page addressing at $0000 (like the 6502), or 
> $2000 (like the HuC6280), or even the 6809 (which CA65 doesn't support 
> anyway), but it isn't good enough for the 65816 which has a 16-bit 
> direct-page base address than can be at any byte-boundary location in 
> the bottom 64KB of memory.
>
> IMHO, getting that to work properly would require modifying the 
> instruction's expression to insert a subtraction into every 
> direct-page instruction that's emitted, and deferring the actual 
> calculation until link time.
>


Just thinking out loud, I'm not suggesting that you should implement 
that in this PR:

Without looking at the code that doesn't seem to be too difficult:

In the object file each reference to a zero-/direct-page variable needs 
to contain information
about the current zero-/direct-page location. Then the linker just needs 
to subtract that value
from the symbol address and check if it's < 256.


> And even then, it would still leave the issue of still having the 
> extra zero-page location at $0000-$00FF, which is actually going to be 
> a serious problem on the 65816 because you are almost certainly going 
> to be having real data located there.


I don't understand this. What "extra zero-page" location? After one sets 
the direct page to something
different, addresses in the 0-255 range can/will accessed with absolute 
(2 byte) instructions.

groessler:
Thinking  more about proper 65826 support...

While my earlier example should work fine in theory

`.dpage $3000`
`lda $3001  ; $A5 $01`

there is a problem when using symbols instead of absolute addresses.

file1:

`.export DPAGEADDR, DPAGESYM`
`.data`
`DPAGEADDR:`
`DPAGESYM .res 1`

file2:

`.import DPAGEADDR,DPAGESYM`
`.code`
`; add instructions to set direct page pointer to DPAGEADDR`
`.dpage DPAGEADDR`
`lda DPAGESYM`

In this case the assembler doesn't know that DPAGESYM can be accessed with a 2-byte instruction.
It somehow needs to be told about that fact.

Alternatively, the linker could change the opcode. For the linker to be able to change the opcode, every branch instruction and branch target would need to be exported as symbol reference and symbol in the object file.

`loop: lda DPAGESYM`
`bne loop`

The "lda" could be two or three bytes, depending on the current direct page setting. So the "bne" would have to be adapted accordingly by the linker, too.

jbrandwood:
>  oliverschmidt commented 7 hours ago 
> 
> I'm not unhappy regarding the size nor complexity of the change. I'm unhappy because Greg stated that he thinks that the change isn't "at the right place". Maybe a larger, more complex change is better. Let me however be totally clear about that I don't see myself being able to assess this! So what are my criteria for a merge "in the core" of the tool chain?
> 
>     People like Greg don't complain about it
>     If there's a user visible aspect then that is documented
>     Regression tests pass
>     Code style fits to the surrounding code

Hi Oliver,

I'll work with Greg here and on the mailing list to see if I can get his agreement that this patch is an improvement for HuC6280 users, and that it is the simplest solution to a real problem that HuC6280 users have.

I believe that a larger and more complex change will eventually have to be made to CA65 to fix direct-page addressing for 65816 users, but I don't think that this patch makes that task any more difficult than it is already going to be.

If I can persuade Greg and Christian of that, then I hope that I can work with you to make sure that the other 3 of your criteria can be met, and that some version of this patch can be accepted.

As it is now, my HuC6280 code has to be sprinkled with "<" addressing-overrides on every zero-page instruction in order to get them to work properly.


>  groessler commented
> 
> Without looking at the code that doesn't seem to be too difficult:
> 
> In the object file each reference to a zero-/direct-page variable needs
> to contain information
> about the current zero-/direct-page location. Then the linker just needs
> to subtract that value
> from the symbol address and check if it's < 256.
> 
> ...
> While my earlier example should work fine in theory
> there is a problem when using symbols instead of absolute addresses.
> ...
> Alternatively, the linker could change the opcode. For the linker to be able to change the opcode, every branch instruction and branch target would need to be exported as symbol reference and symbol in the object file.

Hi Christian,

While it doesn't *seem* too difficult in concept, I think that anyone that attempts this is going to start running into implementation details and bugs that are going to take weeks or months to go through.

As you just pointed out, it is possible to get into the potential situation of changing instructions in the linker, which would be a huge change in CA65/LD65.

The "ASxxxx" assembler implements direct-page addressing in a somewhat similar fashion to what you're proposing, with a special code in the object file to change the current direct-page value, and then one code for 1-byte relocatable immediate values (that don't care about the DP value), and a different code for 1-byte relocatable direct-page references (that do care about the DP value).

However it could eventually be implemented, I believe that CA65 would need to export more information to the linker than it currently does. CA65 seems to currently only export the low-byte of addresses in zero-page instructions (because "<" is processed as a byte-selector expression, and not an addressing-mode specifier), and this means that you're losing information that the linker is going to need in order to correctly subtract the current DP location and make sure that it remains within the zero-page addressing range.


>  groessler commented
> 
> I don't understand this. What "extra zero-page" location? After one sets
> the direct page to something different, addresses in the 0-255 range
> can/will accessed with absolute (3 byte) instructions.

I understand that that is the way that things should work.

My first patch that Travis CI rejected, was an attempt to implement things in that way.

IMHO, there's a lot of stuff going on under the hood in CA65 that is going to need to be changed in order to keep the information that is needed to remove CA65's assumption that zero-page must be at $0000-$00FF.

If you start tracing through CA65's processing of symbols, expressions and effective-addresses in a debugger, then you'll find that there are some "interesting" assumptions made in there.

For instance, it's going to take some serious changes to CA65 to stop "(zpsymbol + 4)" from being promoted into 2-byte addressing if "4" is no longer seen as a zero-page location.

oliverschmidt:
> [...]  I hope that I can work with you to make sure that the other 3 of your criteria can be met, and that some version of this patch can be accepted.

From what I see / understand so far I don't see issues there :-)

greg-king5:
I have implemented a directive (`.setdp`) that does what we want.  It works well, except for a long-standing bug in ca65:  lines that are inside a scope have difficulty seeing symbol names that are defined outside of that scope.  We have known that, sometimes, those symbols cannot be found.  I have discovered that ca65 always cannot see which of those symbols is constant.  The only way to make my patch work with those symbols is to use the name-space operator (`::`).

You can see the important pieces of the patch at https://github.com/greg-king5/cc65/blob/direct-page/src/ca65/instr.c#L985 and https://github.com/greg-king5/cc65/blob/direct-page/src/ca65/instr.c#L1197.

If you have a local copy of the cc65 repository, and you want to test the code, then you can use these two Git commands to get the code:
```sh
git checkout -b direct-page master
git pull https://github.com/greg-king5/cc65.git direct-page
```

jbrandwood:
Thanks Greg, I'll take a look at your patch and try it out.

At first glance it's a more generalized and much more elegant version of one of my earlier attempts that I rejected because I didn't think that you guys would accept it with the kind of scoping problems with symbol resolution that you're seeing.

It's good to know that a namespace can be used to get around the scoping, that's valuable news to me!

groessler:
On 07/10/16 14:01, greg-king5 wrote:

> I have implemented a directive (|.setdp|) that does what we want.
>


Looks fine to me.

It would be nice if you could add some tests for this to the testsuite.

regards,
chris

oliverschmidt:
@jbrandwood:

> Thanks Greg, I'll take a look at your patch and try it out.

Given that you seem(ed) pretty interested in the topic at hand and motivated to get something merged into cc65 upstream I'm wondering when Greg / we can receive feedback from you...

jbrandwood:
Sorry for the delay in responding, my time this week got taken up with other
priority work.

If I'm understanding Greg's patch thinking correctly, then I believe that his
patch has a small-but-critical bug in it that incorrectly promotes all 65816
zero-page addresssing modes to absolute-addressing.

If tried his original code the the "65816" replaced with "HuC6820", and CA65
couldn't assemble the output that CC65 produced because a whole bunch of
instructions didn't work with absolute-addressing.

I think that I've got a fix that changes it to do what he wanted, and here
are some tests with both his original code, and with my change.

In the tests below, the "<" addressing override is only added when it is needed
to make the test assemble without error, i.e. it points out things that CA65
either correctly or incorrectly thinks aren't legal ZP instructions without it.

This is the test setup ...


    000000r 1                               .exportzp _string
    000000r 1               
    000000r 1                               .importzp _ed_xfer_src
    000000r 1               
    000000r 1                               .setcpu "HuC6280"
    000000r 1               
    000000r 1               _al             =       $20f8
    000000r 1               member          =       $04
    000000r 1               
    000000r 1                               .zeropage
    000000r 1               
    000000r 1  00 00        _string:        .word   0
    000002r 1               
    000002r 1                               .code




***************************************************************************

Greg's Original src/ca65/instr.c line 995 ...

            if (CPU == CPU_65816 && ED.AddrSize == ADDR_SIZE_ZP) {
                ++ED.AddrSize;
            }

Note: "sta VDC_DATA_LO" is incorrectly evaluated as a ZP location.
      Symbol resolution in proc scope breaks things.

    000002r 1                               .code
    000000r 1               
    000000r 1  64 F8                        stz     _al
    000002r 1  64 FC                        stz     _al + 4
    000004r 1  64 FC                        stz     _al + member
    000006r 1  64 rr                        stz     _ed_xfer_src
    000008r 1  64 rr                        stz     _ed_xfer_src + 4
    00000Ar 1  85 02                        sta     VDC_DATA_LO
    00000Cr 1  85 AC                        sta     $20ac
    00000Er 1               
    00000Er 1  91 F8                        sta     (_al),y
    000010r 1  91 FC                        sta     (_al + 4),y
    000012r 1  91 rr                        sta     (_ed_xfer_src),y
    000014r 1  91 rr                        sta     (_ed_xfer_src + 4),y
    000016r 1  91 02                        sta     (VDC_DATA_LO),y
    000018r 1  91 AC                        sta     ($20ac),y
    00001Ar 1               
    00001Ar 1                               .proc   ug: near
    00001Ar 1               
    00001Ar 1  9C F8 20                     stz     _al
    00001Dr 1  9C FC 20                     stz     _al + 4
    000020r 1  9C FC 20                     stz     _al + member
    000023r 1  64 rr                        stz     _ed_xfer_src
    000025r 1  64 rr                        stz     _ed_xfer_src + 4
    000027r 1  85 02                        sta     VDC_DATA_LO
    000029r 1  85 AC                        sta     $20ac
    00002Br 1               
    00002Br 1  91 F8                        sta     (<_al),y
    00002Dr 1  91 FC                        sta     (<_al + 4),y
    00002Fr 1  91 rr                        sta     (_ed_xfer_src),y
    000031r 1  91 rr                        sta     (_ed_xfer_src + 4),y
    000033r 1  91 02                        sta     (VDC_DATA_LO),y
    000035r 1  91 AC                        sta     ($20ac),y
    000037r 1               
    000037r 1                               .endproc

***************************************************************************

Modified with "ED_Const() instead of 65816 test" ...

            if (ED_IsConst (&ED) && ED.AddrSize == ADDR_SIZE_ZP) {
                ++ED.AddrSize;
            }

Note: "sta VDC_DATA_LO" is fixed, but "sta (VDC_DATA_LO),y" is still wrong.
      Symbol resolution in proc scope still breaks things.

    000002r 1                               .code
    000000r 1               
    000000r 1  64 F8                        stz     _al
    000002r 1  64 FC                        stz     _al + 4
    000004r 1  64 FC                        stz     _al + member
    000006r 1  64 rr                        stz     _ed_xfer_src
    000008r 1  64 rr                        stz     _ed_xfer_src + 4
    00000Ar 1  8D 02 00                     sta     VDC_DATA_LO
    00000Dr 1  85 AC                        sta     $20ac
    00000Fr 1               
    00000Fr 1  91 F8                        sta     (_al),y
    000011r 1  91 FC                        sta     (_al + 4),y
    000013r 1  91 rr                        sta     (_ed_xfer_src),y
    000015r 1  91 rr                        sta     (_ed_xfer_src + 4),y
    000017r 1  91 02                        sta     (VDC_DATA_LO),y
    000019r 1  91 AC                        sta     ($20ac),y
    00001Br 1               
    00001Br 1                               .proc   ug: near
    00001Br 1               
    00001Br 1  9C F8 20                     stz     _al
    00001Er 1  9C FC 20                     stz     _al + 4
    000021r 1  9C FC 20                     stz     _al + member
    000024r 1  64 rr                        stz     _ed_xfer_src
    000026r 1  64 rr                        stz     _ed_xfer_src + 4
    000028r 1  85 02                        sta     VDC_DATA_LO
    00002Ar 1  85 AC                        sta     $20ac
    00002Cr 1               
    00002Cr 1  91 F8                        sta     (<_al),y
    00002Er 1  91 FC                        sta     (<_al + 4),y
    000030r 1  91 rr                        sta     (_ed_xfer_src),y
    000032r 1  91 rr                        sta     (_ed_xfer_src + 4),y
    000034r 1  91 02                        sta     (VDC_DATA_LO),y
    000036r 1  91 AC                        sta     ($20ac),y
    000038r 1               
    000038r 1                               .endproc

***************************************************************************

Changing VDC_DATA_LO equate in asminc/pce.inc line 36 ...

    ; VDC port
    ; Note: use mirror location to avoid CA65 zero-page logic
  
    VDC_CTRL        = $0100
    VDC_DATA_LO     = $0102
    VDC_DATA_HI     = $0103


Greg's Original & "ED_Const() instead of 65816 test" both correctly require
overrides on VDC_DATA_LO when used in an indexed instruction ...

    000002r 1                               .code
    000000r 1               
    000000r 1  64 F8                        stz     _al
    000002r 1  64 FC                        stz     _al + 4
    000004r 1  64 FC                        stz     _al + member
    000006r 1  64 rr                        stz     _ed_xfer_src
    000008r 1  64 rr                        stz     _ed_xfer_src + 4
    00000Ar 1  8D 02 01                     sta     VDC_DATA_LO
    00000Dr 1  85 AC                        sta     $20ac
    00000Fr 1               
    00000Fr 1  91 F8                        sta     (_al),y
    000011r 1  91 FC                        sta     (_al + 4),y
    000013r 1  91 rr                        sta     (_ed_xfer_src),y
    000015r 1  91 rr                        sta     (_ed_xfer_src + 4),y
    000017r 1  91 02                        sta     (<VDC_DATA_LO),y
    000019r 1  91 AC                        sta     ($20ac),y
    00001Br 1               
    00001Br 1                               .proc   ug: near
    00001Br 1               
    00001Br 1  9C F8 20                     stz     _al
    00001Er 1  9C FC 20                     stz     _al + 4
    000021r 1  9C FC 20                     stz     _al + member
    000024r 1  64 rr                        stz     _ed_xfer_src
    000026r 1  64 rr                        stz     _ed_xfer_src + 4
    000028r 1  8D 02 01                     sta     VDC_DATA_LO
    00002Br 1  85 AC                        sta     $20ac
    00002Dr 1               
    00002Dr 1  91 F8                        sta     (<_al),y
    00002Fr 1  91 FC                        sta     (<_al + 4),y
    000031r 1  91 rr                        sta     (_ed_xfer_src),y
    000033r 1  91 rr                        sta     (_ed_xfer_src + 4),y
    000035r 1  91 02                        sta     (<VDC_DATA_LO),y
    000037r 1  91 AC                        sta     ($20ac),y
    000039r 1               
    000039r 1                               .endproc

***************************************************************************

jbrandwood:
After doing the testing, my feeling is that Greg's patch is a much better
long-term solution for CA65 than my original patch at the start of this
pull-request.

Even with the ED_Const() fix, Greg's patch still seems to occasionally suffer
from the duplicate zero-page at $0000-$00FF problem that I was talking about
earlier in the thread.

As such, it's still likely to unpredicatbly break things for 65816 users
that use the new ".setdp", and asminc/pce.inc MUST be modified to change
the VDC_* equates into a non-zero-page mirror location.

But with the ED_Const() fix, and the change in pce.inc, it's going to be
perfectly usable for PCE programming, and it's going to be the better
base from which someone can eventually fix the whole scoped symbol
processing issue.

jbrandwood:
Waiting for some feedback. Do you guys want me to close the pull-request?

oliverschmidt:
Sorry but I can't provide usefull feedback. However I don't see why this pull request should be closed at this time.

greg-king5:
My Pull Request will change only the assembler part.  Your PR will be needed to change the `pce` library part.

fo-fo:
> The "lda" could be two or three bytes, depending on the current direct page setting. So the "bne" would have to be adapted accordingly by the linker, too.

Just a small comment: You'd have to completely break backwards compatibility to allow this change. Consider this example:

```
loop: lda DPAGESYM
  bne loop
end:
size = end-loop ; currently, this is an assembly-time constant
.out .sprintf("size is: %d", size) ; this works
```
Obviously this would not be possible anymore if linker is allowed to change the code/data size.

groessler:
The ".out" wouldn't work, that's true.
But I wouldn't call it "completely break backwards compatibility".

fo-fo:
Yeah maybe that was a too strong choice of words. Anyway, you can expect a lot of problems with that approach (not just with .out).

jbrandwood:
> greg-king5 referenced this pull request 3 days ago
>
> ca65: expressions don't search for symbols in outer scopes #235 

I added a comment about the current CA65 symbol-handling in that other PR.

From my POV, the current symbol-handling implementation in CA65 is inconsistent, and causes more  problems than its potential-benefits provide. But that's really a separate issue from the direct-page stuff here ... it's just the root-cause of some of the difficulties in engineering a good solution to this issue here that will work in all circumstances (i.e. within scopes).

groessler:
On 08/09/16 20:10, fo-fo wrote:
>
> Anyway, you can expect a lot of problems with that approach (not just 
> with .out).
>


Hmm, what kind of problems are you thinking of? I think it should work, 
in theory.

jbrandwood:
Hi Greg ... I can see that you've made some more changes to your
personal branch for the new ".setdp" instruction since I last tested
it out ... but those changes still haven't been applied to the CC65
master branch.

May I ask how things are going, and if there's anything that I can do
to help get that new functionality adding into the main branch?

...
...

Well, a build of Greg's latest source shows that some great progress
has been made!

It looks like the automatic ".setdp $2000" that's being done for the
HuC6280 processor is working wonderfully ... at least in the normal
top-level scope.

But things are totally hosed within a "proc" scope.

So, a giant step forward, and totally usable from my POV as an
assembly-language programmer that doesn't use/need CA65's scoping
capability.

IMHO, the next step toward finally nailing this thing is probably
going to be some work on CA65's symbol-table code to make those scopes
work better (which would probably also fix issue #235 at the same
time).

What are your thoughts Greg, now that you've dug into the source code
and got the ".setdp" working?

jbrandwood:
> It looks like the automatic ".setdp $2000" that's being done for the
> HuC6280 processor is working wonderfully ... at least in the normal
> top-level scope.

Uh-oh ... one problem that now exists in the top-level scope is that the "<" prefix to an addressing mode doesn't work on the HuC6280 (and presumably the 65816).

I'm going to guess that this is because the "<" is being treated in the addressing mode as a low-byte operator rather than as an addressing mode override (which it is supposed to be by convention and WDC documentation).

jbrandwood:
It's very quiet in here.

greg-king5:
You are correct about the "`<`" operator.

It can be used as an address-mode override only when the operand of `.setdp` is less than 256.  Then, (at least some) eight-bit constant numbers will be mapped to eight-bit addresses.  ca65 will default to zero-page addressing modes when it knows that the number won't be bigger than 255.

("a:", "f:", and "z:" are the official overrides in ca65's syntax.)

jbrandwood:
> You are correct about the "<" operator.
> ...
> ("a:", "f:", and "z:" are the official overrides in ca65's syntax.)

Since "<", "|", and ">" are the official addressing-mode override prefixes from the WDC 65816 Data Sheet, and were used on the Apple II GS and SNES, and "<" and ">" were commonly used on 6502 assemblers as the addressing-mode override prefix ...

... then perhaps it might be a good idea for CA65 to accept them as alternatives to the CA65-only "z:", "a:" and "f:"?

oliverschmidt:
Without really understanding the topic discussed here I'd like to mention that I believe that the ca65 design goal isn't primarily about compatibility but rather consistency. So if a decision needs to be made weighing those two I think it's appropriate to rather opt for the latter - just my two cents.

jbrandwood:
> Without really understanding the topic discussed here I'd like to mention that I believe that the ca65 design goal isn't primarily about compatibility but rather consistency.

The question is less one of compatibility (although I'm a big fan of that, myself) ... it's a question of correct operation.

As a programmer, when you write "lda <variable", your desire is to tell the assembler that you want it to use zero-page/direct-page addressing.

You're not doing it to say "I want the low-byte of the variable address".

But when you write "lda #<variable" ... then you ARE saying that "I want the low-byte of the variable address".

They are different uses.

The only reason that it has worked so far is that zero-page/direct-page is at location $0000-$00FF on the standard 6502.

Now that Greg is adding the ".setdp" feature, that isn't true anymore.

The way that CA65 is processing the "<" in an addressing-mode fundamentally loses the intent of what the programmer means, and it is causing the "<" to fail when using the ".setdp".

Just processing that leading "<" as an equivalent to your existing "z:" prefix should fix the situation.

bbbradsmith:
I don't think it's a good idea to have < take on different behaviours depending on different contexts. It should always mean "low byte", directly analogous to < for "high byte".

You may have been using < in existing PCE code to cope with the lack of DP support, but don't retroactively change the function to fit the new .setdp idea. Write new code that works this .setdp instead, and just stop using < to mean "force ZP instruction" in new code. (...and old code maintains backward compatibility simply by not using .setdp)

z: isn't a "force ZP" prefix, it's an "enforce ZP" prefix. It just creates an error if the address doesn't turn out to be in the ZP. So z: should now check the symbol's high byte against the DP, rather than for $00.

Maybe some disassemblers or the datasheet use < to indicated a ZP instruction, but I think it's a different problem looking at a single instruction vs. looking at symbolic code in context. In ca65 < and > are simple prefix operators, not instruction modifiers.

mrdudz:
totally agree, < should always (and by that i mean always) mean "take the low byte". giving it a different meaning depending on context is a terrible idea.

jbrandwood:
We can probably all agree that context-sensitive syntax isn't always ideal ... but it is common in most programming languages (including C).

Using "<" to force an instruction to use ZP addressing was common usage when all of these 6502-and-65816 systems were mainstream.

CA65 certainly doesn't need to follow that usage if the developers don't want it to, and the assembler should normally have all the information that it needs to determine if/when to use ZP addressing.

It just doesn't do so correctly at-the-moment, and won't until the .setdp functionality is integrated.

EtchedPixels:
.setdp doesn't actually work for all the 65c816 cases because you can be in bank 1 with DP in bank 0, so there is *no* valid value to put in .setdp..

mrdudz:
For some strange reason the automatic checking will not start... *shrug*

jbrandwood:
It's a VERY old set of patches. Greg's DP patch was a better starting-point, and I keep a semi-up-to-date fork of CA65 with it applied here ...
https://github.com/jbrandwood/cc65

Although FWIW, I don't personally use CA65/CC65 myself these days.

Issue #321, opened 2016-07-07
rand() uses only 23 bits of its state


pinobatch:
Currently `rand()` in the C library is a hand-optimized assembly language implementation of the following linear congruential generator:

    int rand(void) {
        seed = (seed * 16843009) + 826366247;
        return (seed >> 8) & 0x7FFF;
    }

However, it was discovered in [this post on NESdev](http://forums.nesdev.com/viewtopic.php?p=174895#p174895) that bits 31-23 of the state never propagate into the output. So it actually produces the same sequence as a 23-bit LCG:

    int rand_as_if(void) {
        seed = ((seed * 65793) + 4282663) & ((1 << 23) - 1);
        return seed >> 8;
    }

A reply to the NESdev post includes a spectral test, which shows correlations between adjacent numbers in the sequence. It turns out to be a lot more correlated than, say, a linear feedback shift register such as [Greg Cook's implementation of CRC16](http://6502.org/source/integers/crc-more.html).

So there are two options:

* **Repair:** Re-specify `rand()` to return a less correlated function of the state. One possibility is bits 31-17. Another, which I've heard described as "tempering", is to continue to return bits 22-16 in X, but change A to bits 31-24 XOR bits 15-8.
* **Optimize:** Keep the current `rand()` sequence, but make it take fewer cycles by not calculating bits 31-24 (`rand+3`) at all. I have done that. Do you want it as a pull request?

bbbradsmith:
The spectral test doesn't really apply to LFSRs, so it's not really the fairest comparison. It's really more of a thing to identify poorly chosen LCGs, but this particular LCG doesn't look bad at all to me.

Using bits 16-30 looks perfectly good to me. The simple tests I've done so far make it look slightly better than the current bits 8-22. It'd be a pretty simple change, and it slightly increases efficiency by getting rid of that pha/pla.

Technically you could use bits 17-31 for maximal length, but a 31-bit sequence is plenty good enough, I think.

bbbradsmith:
Pull request submitted here: https://github.com/cc65/cc65/pull/323

oliverschmidt:
Fixed with https://github.com/cc65/cc65/pull/323

PR #323, opened 2016-07-12
Fix broken rand() implementation.

bbbradsmith:
Fix broken rand() implementation. The high 8 bits were unused, reducing it to a 24-bit implementation (while still doing all the work for a 32-bit one). The best entropy is in the unused high byte, returning these bits in A instead of bits 8-15, which had considerably lower entropy (i.e. rand() & 255 was effectively a 16-bit LCG).

### DIFF
--- a/libsrc/common/rand.s
+++ b/libsrc/common/rand.s
@@ -44,7 +44,6 @@ _rand:  clc
         lda     rand+1
         adc     #$59
         sta     rand+1
-        pha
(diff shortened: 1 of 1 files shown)


bbbradsmith:
Response to issue raised here: https://github.com/cc65/cc65/issues/321

oliverschmidt:
Thanks for your contribution.

Issue #296, opened 2016-05-24
A bit more verbose error messages?


ssuukk:
My compilation fails with: "Error: Incompatible pointer types" and the function I am calling has 3 pointer arguments, all of them very cryptic (it's not my code...). It would be great if cc65 at leas told which pointer is failing, either by index or name of argument...

oliverschmidt:
Closed via https://github.com/cc65/cc65/pull/328

PR #328, opened 2016-08-02
Reporting sym name for incompatible pointer types.

cacciatc:
PR to address https://github.com/cc65/cc65/issues/296

Note: if an argument with an incompatible pointer type appears more than once in a function invocation, then this message is less helpful. However, the simplicity of this solution I think still outweighs that edge-case (unless there is an easy way to get the current param index inside the TypeConversion function).


oliverschmidt:

PR #332, opened 2016-08-10
Add C-code interfaces for the CBM Kernal functions SCNKEY and UDTIM.

greg-king5:
Normally, Commodore's Kernal firmware calls those functions while it is handling the "Jiffy" interrupt.  But, if a program usurps that interrupt, then it must call those functions.  Otherwise, library functions such as `clock()` and `cgetc()` cannot work.

Assembly Language `.interruptor` routines are the best way to call those functions.  But, if a programmer wants to write an Interrupt Service Routine in C, then those Kernal functions must be called through names that are visible to the C compiler.

This Pull Request adds those visible names to cc65's function library.  Also, it adds wrapper code that handles ROM/RAM banking on the Plus4.

### DIFF
--- a/doc/funcref.sgml
+++ b/doc/funcref.sgml
@@ -3,7 +3,7 @@
 <article>
 <title>cc65 function reference
 <author><url url="mailto:uz@cc65.org" name="Ullrich von Bassewitz">
-<date>2016-06-08
+<date>2016-08-07
 
 <abstract>
 cc65 is a C compiler for 6502 based systems. This function reference describes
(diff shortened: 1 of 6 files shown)


oliverschmidt:
May general understanding of cc65 interrupt handling is that the original handler is called after the cc65 is done with its business (aka interrupt chaining). As far as I understand this is true for the Plus4 too:
https://github.com/cc65/cc65/blob/master/libsrc/plus4/crt0.s#L164

So I generally don't understand why cc65 interrupt code (be it asm or C) should have the need to call anything explicitly that is usually called anyway.

greg-king5:
Unfortunately, the Plus4's Kernal uses _two_ raster interrupts; the upper one is used only to prepare for a general test for the lower one.  The lower raster interrupt calls those functions internally.  In order to move between those two raster lines, the Kernal must change the raster location of interrupts each time.

That interferes with programs that want to choose the locations.  Therefore, those programs must use and _cancel_ raster interrupts, so that the Kernal won't impose its own settings.  But, it means that the Kernal won't do _any_ of the things that it normally does while handling rasters -- the program must do them.

"Jiffy" interrupts are separate from raster interrupts on other models.  But even there, programs might stop Jiffy interrupts because the Kernal will interfere with the precise timing that fancy video effects need.  Strictly speaking, those platforms don't need this PR because their ISRs must be written in Assembly (to get that precise timing).  But, I want the patches to be generic across the `cbm` libraries -- for the sake of consistency.

oliverschmidt:
@greg-king5: Thanks for the detailed explanation :-)

Issue #331, opened 2016-08-09
Standard version options for all bin/*


cacciatc:
I noticed the following today:

```
bin/ld65 --version
ld65 V2.15
ld65: Error: No object files to link
```

I then tried the other programs in bin/* and it is not consistent:

```
./bin/ar65 --version
Usage: ar65 <operation ...> lib file|module ...
Operations are some of:
	a	Add modules
	d	Delete modules
	l	List library contents
	v	Increase verbosity (put before other operation)
	x	Extract modules
	V	Print the archiver version

./bin/ar65 V
ar65 V2.15

./bin/chrcvt65 --version
chrcvt65 V2.15
chrcvt65: No input files

./bin/cl65 --version
cl65 V2.15
cl65: No input files

./bin/co65 --version
co65 V2.15
co65: Error: No input file

./bin/da65 --version
da65 V2.15
da65: No input file

./bin/grc65 --version
grc65 V2.15
grc65: No input file

./bin/od65 --version
od65 V2.15
od65: No input files

./bin/sim65 --version
sim65 V2.15
sim65: No program file

./bin/sp65 --version
sp65 V2.15
```

It's a little thing, but IMO they should all exit successfully ala https://www.gnu.org/prep/standards/standards.html#g_t_002d_002dversion. Am happy to submit a PR for this.

mrdudz:
agreed

polluks:
BTW shall we use getopt_long?

oliverschmidt:
I noticed that too - but it wasn't important enough to me.  I'm in general fine to merge a PR changing it.

oliverschmidt:
Fixed with https://github.com/cc65/cc65/issues/334.

PR #334, opened 2016-08-12
All programs print version and exit successfully.

cacciatc:
* All programs are now using the ProgName variable as well to display their respective name.

PR to address https://github.com/cc65/cc65/issues/331


oliverschmidt:

Issue #330, opened 2016-08-09
switch


polluks:
This should be legal (but nonsense) C:
`int main()`
`{`
`switch (0);`
`return 0;`
`}`

a.c(3): Error: `{' expected
a.c(3): Warning: No case labels

cacciatc:
Started looking into this over the weekend.

oliverschmidt:
Fixed with https://github.com/cc65/cc65/pull/336.

PR #336, opened 2016-08-15
Switch error fix

cacciatc:
Removed the test for LCURLY inside the switch statement.

PR to address https://github.com/cc65/cc65/issues/330

Notes:

* The test for LCURLY was introduced at https://github.com/cc65/cc65/commit/64ec376140353367ed03cb7f294529f7536a1b9d. This commit mentions that those changes were made to support arbitrary code in switch statements such as Duff's Device. Accordingly, I added a test for Duff's Device.

* Also added tests for empty switch bodies. This test just needs to compile successfully to pass; although, I placed it inside the val dir since there doesn't appear to be an opposite of err dir (where these files *must* compile, rather than not compile).

* Unlike with gcc or clang an empty switch will still generate code in cc65 https://github.com/cc65/cc65/blob/master/src/cc65/swstmt.c#L116

### DIFF
--- a/test/val/switch2.c
+++ b/test/val/switch2.c
@@ -1,0 +1,39 @@
+/*
+  !!DESCRIPTION!! Testing empty bodied switch statements.
+  !!ORIGIN!!      
+  !!LICENCE!!     GPL, read COPYING.GPL
+*/
+
+#include <stdio.h>
+
+unsigned char success=0;
+unsigned char failures=0;
+unsigned char dummy=0;
+
+void done()
+{
+  dummy++;
+}
+
+void switch_no_body(void)
+{
+  switch(0);
+}
+
+void switch_empty_body(void)
+{
+  switch(0) {};
+}
+
+/* only worried about this file compiling successfully */
+int main(void)
+{
+  switch_no_body();
+  switch_empty_body();
+
+  success=failures;
+  done();
+  printf("failures: %d\n",failures);
+
+  return failures;
+}
--- a/test/val/switch2.c
+++ b/test/val/switch2.c
@@ -22,7 +22,7 @@ void switch_no_body(void)
 
 void switch_empty_body(void)
 {
-  switch(0) {};
+  switch(0) {}
 }
 
 /* only worried about this file compiling successfully */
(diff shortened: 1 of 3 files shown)


greg-king5:
`test/val/` _is_ the opposite of `test/err/` because the opposite of "doesn't compile" is "compiles into correct machine-code".

`switch2` must be executed because we need to know that cc65 won't translate "empty" switch statements into something that crashes.

greg-king5:
> Unlike with gcc or clang, an empty switch still will generate code in cc65: https://github.com/cc65/cc65/blob/master/src/cc65/swstmt.c#L116

That referenced comment mentions that the optimizer usually will remove that (unused) code.

---
What about "Warning: No case labels"?  Should that message be kept; or, should it, too, be removed?

---
There are a couple of comments in `SwitchStatement()` that imply that switch statements must have braces.  Those comments need to be fixed.

test/val/switch2.c
```diff
@@ -0,0 +1,39 @@
...
+}
+
+void switch_no_body(void)
+{
+  switch(0);
+}
+
+void switch_empty_body(void)
+{
+  switch(0) {};
```
greg-king5: You can remove that semicolon.

cacciatc:
> Unlike with gcc or clang, an empty switch still will generate code in cc65: https://github.com/cc65/cc65/blob/master/src/cc65/swstmt.c#L116
That referenced comment mentions that the optimizer usually will remove that (unused) code.

Yeah, just wanted to point it out, since as an end user I might not expect that behavior.

> What about "Warning: No case labels"? Should that message be kept; or, should it, too, be removed?

I think we should keep the message. It is standard among the other notable C compilers and the vast majority of the time there should be case labels.

> There are a couple of comments in SwitchStatement() that imply that switch statements must have braces. Those comments need to be fixed.

Good catch!

cacciatc:
I only saw one comment that referred to the previous curly brace behavior and fixed it. https://github.com/cc65/cc65/blob/master/src/cc65/swstmt.c#L202 refers to a closing brace, but that statement is still true given the changes.

greg-king5:
The word "(correctly)" says that it is a mistake not to have a closing brace -- but, that no longer is true.  Therefore, the word should be removed.

cacciatc:
Removed "correctly".

Yeah, I was reading that comment differently. The only way for the antecedent to be true is if there is a matching opening brace (as far as I know a closing curly brace without an opening one is never valid C syntax). In our case of a switch with no curly brace, it is impossible for a closing curly to "correctly" terminate the expression. Therefore the comment is not applicable to the case of ```switch(0);```

PR #338, opened 2016-08-19
Allow use of different charmaps on Atari target

IrgendwerA8:
Hi,

as you may can identify in the test program, cc65 has a problem with (wrong) string merging of identical source code strings under different char maps (at least to my limited understanding).
However, the shown use case in the test program demonstrates that the support of both mappings is useful even now (I had already a discussion with Chris Groessler). The merging problem should be added to the 2DO?

Another note: Wouldn't it make sense to put all target dependent includes in separate include directory? Like
include/target/atari/...
include/target/cbm/...  etc.


oliverschmidt:
It seems to me that some addition to the documentation would be desirable, wouldn't it?

include/atari_screen_charmap.h
```diff
@@ -0,0 +1,304 @@
...
+#pragma charmap (0x18, 0x58)
+#pragma charmap (0x19, 0x59)
+#pragma charmap (0x1A, 0x5A)
+#pragma charmap (0x1B, 0x5B)
+#pragma charmap (0x1C, 0x5C)
+#pragma charmap (0x1D, 0x5D)
+#pragma charmap (0x1E, 0x5E)
+#pragma charmap (0x1F, 0x5F)
+
+#pragma charmap (0x20, 0x00)
```
greg-king5: Both that `#pragma` and the first one in the list should be surrounded by `#pragma warn (remap-zero)` lines (we don't want cc65 to warn us constantly about something that we are doing deliberately):
```c
#pragma warn (remap-zero, push, off)
#pragma charmap (0x20, 0x00)
#pragma warn (remap-zero, pop)
```
(You don't need to change "atari_atascii_charmap.h".)

IrgendwerA8:
The latest change reflect both comments.

oliverschmidt:
Thanks for the contribution - especially coming with regression test.

PR #341, opened 2016-08-27
Update exehdr.s for Lynx target

alexthissen:
Fix for memory bank 1 which should be zero in almost all cartridges for emulators to work correctly.

### DIFF
--- a/cfg/lynx-bll.cfg
+++ b/cfg/lynx-bll.cfg
@@ -1,7 +1,8 @@
 SYMBOLS {
     __STACKSIZE__:        type = weak, value = $0800; # 2k stack
     __STARTOFDIRECTORY__: type = weak, value = $00CB; # start just after loader
-    __BLOCKSIZE__:        type = weak, value = $0400; # cart block size
+    __BANK0BLOCKSIZE__:   type = weak, value = 1024; # bank 0 cart block size
+    __BANK1BLOCKSIZE__:   type = weak, value = 0; # bank 1 block size
     __BLLHDR__:           type = import;
 }
 MEMORY {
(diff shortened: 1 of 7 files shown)


oliverschmidt:
You write "almost all" so it seems preferable to have another weak variable in the linker config file maybe  named `__BANK1SIZE__` that defaults to zero.

alexthissen:
A very good suggestion. It allows control over the second bank, which is hardly ever present. Then again, when it is, it is probably a custom chip that requires manual control over it.

oliverschmidt:
Great - so I await a new commit. Please provide a short comment here when you're done.

alexthissen:
Added a commit to the pull request. Think this should cover the __BANK1BLOCKSIZE__ changes. Please verify.

cfg/lynx.cfg
```diff
@@ -42,4 +43,4 @@ FEATURES {
             count   = __INTERRUPTOR_COUNT__,
             segment = RODATA,
             import  = __CALLIRQ__;
-}
+}
```
oliverschmidt: You seem to have removed the final linefeed here. It' scommon practise to end all source files with a linefeed.

cfg/lynx-bll.cfg
```diff
@@ -1,7 +1,8 @@
 SYMBOLS {
     __STACKSIZE__:        type = weak, value = $0800; # 2k stack
     __STARTOFDIRECTORY__: type = weak, value = $00CB; # start just after loader
-    __BLOCKSIZE__:        type = weak, value = $0400; # cart block size
+    __BANK0BLOCKSIZE__:   type = weak, value = 1024; # bank 0 cart block size
```
oliverschmidt: Could you please change the constants (back) to four-digit hex constants for consistency reasons. Thanks in advance for your patience with my nitpicking ;-)

alexthissen:
Fixed errors as identified during review.

oliverschmidt:

Issue #314, opened 2016-06-23
Wishlist: "regcall" calling convention


clbr:
Passing 2+th parameters on the stack is quite slow. It also happens that most programs do not use the register keyword, due to the saving overhead it's much better to manually allocate ZP vars. So, for those programs, we have 6 unused ZP bytes in regbank.

I propose a new calling convention, "regcall", where the rightmost param would be passed as in fastcall, and the next up to six bytes in regbank, with the remainder in the sw stack.

This would improve the code generated in the C function as well, with no unnecessary push and pop calls, the optimizer can handle the available data much better.

For the cost of one "jsr pushax", you can write and load all six ZP bytes and have cycles left over. It's also one of the things I've had to work around and manually optimize, making functions explicitly use ZP vars for param passing, when it's something the compiler could do for me.

PR #342, opened 2016-08-28
Demonstration code for "register call" feature, as base for a discussion

IrgendwerA8:
Most parts of the changes where lurking on my hard disks more than three years, where I proposed a new 'regcall' calling convention. (Unfortunately I'm unable to find the old proposal on the mailing list nor Uz reaction. (I wonder where wish #314 comes from - it goes in the same direction using the same vocabulary)).

Like Uz back then remarked: The problem when putting arguments into registers consists in subroutine calls taking arguments in the same registers.
I browsed through the compiler sources back then quite a lot and left with big respect for anybody who is able to service that beast.

So my solution targets the preprocessor, allowing the speed of registers calls without changing the compiler. The user is self responsible for applying the correct call. In the commits I've changed exemplary the memcpy() and memcmp() calls.

These are usable the standard (fastcall) way or using them in upper case letters "MEMCPY"/"MEMCMP" calls the register variant, where the user has to take care currently, that no argument is formed itself by a call destroying the registers.

The selection of this library function is sub-optimal, since register calls make the biggest sense when the runtime performance of the function itself is quite low compared to the performance of fulfilling the calling convention. (Like e.g. fixed point math functions.)

I did however a small test to show the potential: Thanks to the register calls, I've saved ~1/3 of the execution time (see code).
A drawback is that a regcall takes "numberOfParameters-1" more bytes to be executed ("lda var, ldx var+1, jsr pushax" vs. "lda var, sta zp, ldx var+1, sta zp+1")

Things I like to have discussed/answered:

Is there interest in supporting this kind of call or is it just me preferring lean parameter passing this way?

Currently every regcall function has two entry points, one for the fastcall and another one (inner) for the register call. I've created a calling converter which would allow to save these identical "stack to register" code for every individual function (they are currently commented). But here the overhead for the function pointer transfer to the stack (runtime/size) is quite high. A more lean variant could consist in supplying only the function pointer in a register running the same stack to register code and jumping to code of the pointer, but this would imply a (limited) change of the compiler (but also a 8 byte penalty per function call in the classic case (saved code for translation vs. costly calls). Any hints or remarks on this subject?

It would be very desirable if the compiler could detect on his own if a simple register call is possible (no sub calls for arguments) an could choose the correct calling variant. This would imply that the compiler must be able to detect that a function supports the register call (keyword?!).
Is there a chance for such support?

### DIFF
--- a/include/callconv.h
+++ b/include/callconv.h
@@ -1,0 +1,91 @@
+/*****************************************************************************/
+/*                                                                           */
+/*                             callconv.h                                    */
+/*                                                                           */
+/*                 Calling convertor, register call to fastcall              */
+/*                                                                           */
+/*                                                                           */
+/*                                                                           */
+/* (c) Christian Krüger, latest change: 22-Jul-2013                          */
+/*                                                                           */
+/*                                                                           */
+/* This software is provided 'as-is', without any expressed or implied       */
+/* warranty.  In no event will the authors be held liable for any damages    */
+/* arising from the use of this software.                                    */
+/*                                                                           */
+/* Permission is granted to anyone to use this software for any purpose,     */
+/* including commercial applications, and to alter it and redistribute it    */
+/* freely, subject to the following restrictions:                            */
+/*                                                                           */
+/* 1. The origin of this software must not be misrepresented; you must not   */
+/*    claim that you wrote the original software. If you use this software   */
+/*    in a product, an acknowledgment in the product documentation would be  */
+/*    appreciated but is not required.                                       */
+/* 2. Altered source versions must be plainly marked as such, and must not   */
+/*    be misrepresented as being the original software.                      */
+/* 3. This notice may not be removed or altered from any source              */
+/*    distribution.                                                          */
+/*                                                                           */
+/*****************************************************************************/
+
+#ifndef _CALLCONV_H
+#define _CALLCONV_H
+
+#include <stddef.h>
+
+typedef void __fastcall__ (* RegCallFuncRetV)(void);
+typedef int __fastcall__ (* RegCallFuncRetN)(void);
+typedef size_t __fastcall__ (* RegCallFuncRetSt)(void);
+typedef void* __fastcall__ (* RegCallFuncRetPv)(void);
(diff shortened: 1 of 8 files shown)


mrdudz:
i somehow dont like the overhead that is introduced even when this new stuff isnt used. wouldnt it make more sense to put these alternative library routines into seperate sourcefiles?
i also find uppercase names for library routines kindof ugly - to me those are reserved for constants, not functions. my proposal would be using memcpy_rc or whatever for them.

IrgendwerA8:
Since most of the function need to pull their data from stack to 'registers' even today the overhead is very low (compare the old & new version of the functions). Centralizing this code means to save some code & work.
I choose upper case names to indicate the usage of a macro - a coding standard which seems quite common to me. It also implies some attention which isn't bad when you have to follow 'special' parameter passing rules. But of course I would prefer some compiler smartness choosing the matching variant automatically too... (so there is only one entry point visible for the programmer).

mrdudz:
the overhead might be low when only those few functions are modified like that... but once more functions inherit this modification, it'll easily add up to an amount that seems unacceptable to me. (i am referring to size, not speed here). the speed penalty should be zero for someone who doesnt need/want this stuff, IMHO

polluks:
Dates should be ISO 8601. Last line is missing EOL.

theoldbloke:
> [...] using them in upper case letters "MEMCPY"/"MEMCMP" calls the register variant

Although I understand and can sympathise with IrgendwerA8's motivation, **please, please, NO**: don't stick band-aids onto CC65 which mutate it into some weird, non-standard C variant.

> It would be very desirable if the compiler could detect on his own if a simple register call is possible [...]

Agreed, but do it in the compiler (it's not easy), *not* the pre-processor.

> Is there interest in supporting this kind of call [...]

Sorry, not from me. But I do support discussion of ways to improve the generated code within the (severe) limitations of a Small-C descendant, and IrgendwerA8 has brought attention to an issue that is highly worthy of further examination.

IrgendwerA8:
Since nobody seems really interested I keep this as private feature for my developments.

Issue #168, opened 2015-07-03
cl65 writes object files into current work directory by default


mrdudz:
what the topic says. they should go into the system temp directory instead

greg-king5:
Actually, cl65 allows the other tools to do their default action of putting their output files next to their input files.  Only the final output file can be put into a different location.  Usually, that's what we want:  all related files with the same name (and different suffixes) are grouped together.

But occasionally, we want to avoid cluttering out input directory with any output files.  Therefore, we need a cl65 option that will put intermediate files into a separately named directory -- something like
`--temp-dir <dir>`.

I see this issue as a feature request, instead of a bug report.

oliverschmidt:
I disagree that this is usually what we want. Rather I don't see why any write operation should be performed in the directory containing the input/source file(s).

Anyway I agree with you to classify this issue as an enhancement (aka feature request) rather than a bug.

MonteCarlos:
If someone would like to add the object file to a library with ar65 he 
needs them in the source directory or subsequent directories.
This is what every ide does, i think.

MC


Am 04.07.2015 um 22:34 schrieb Oliver Schmidt:

mrdudz:
On Tuesday 07 July 2015, 04:42:11 MonteCarlos <notifications@github.com> 
wrote:
> If someone would like to add the object file to a library with ar65 he
> needs them in the source directory or subsequent directories.
> This is what every ide does, i think.

in that case you would explicitly instruct cl65 to produce an object file as 
output though, and _then_ it will go to the specified directory, of course.

-- 

http://www.hitmen-console.org    http://magicdisk.untergrund.net
http://www.pokefinder.org        http://ar.pokefinder.org

Aus Fehlern wird man klug. Darum ist einer nicht genug.

fo-fo:
I agree with @oliverschmidt and @mrdudz that system temporary directory would be preferable.

Related issue is that if you have two files, say `test.c` and `test.s` and you compile them with `cl65 test.c test.s`, cl65 ends up overwriting and deleting `test.s`. This actually happened to me a while ago.

MonteCarlos:
thats why .c.o or .s.o is seen in makefiles very often


Am 24.03.2016 um 10:40 schrieb fo-fo:

spiro-trikaliotis:
Hello,

* On Thu, Mar 24, 2016 at 02:40:31AM -0700 fo-fo wrote:
 
> Related issue is that if you have two files, say test.c and test.s and you
[...]

PR #348, opened 2016-09-07
ca65: added .P4510 and .IFP4510 pseudo commands

SvOlli:
ca65: added .P4510 and .IFP4510 analog to .PC02 and .IFPC02 commands
- added testcase to "test/assembler"
- updated documentation

getcpu.s: added detection for 4510 CPU


doc/ca65.sgml
```diff
@@ -3805,6 +3827,7 @@ Here's a list of all control commands and a description, what they do:
        <tt><ref id=".IFPSC02" name=".IFPSC02"></tt>,
        <tt><ref id=".P02" name=".P02"></tt>,
        <tt><ref id=".P816" name=".P816"></tt>,
+       <tt><ref id=".P4510" name=".P4510"></tt>
```
greg-king5: Typo: a comma is missing from the end of the line.

SvOlli:
added the missing ",", since no other topic has arisen over the weekend

greg-king5:
What are the problems that you had with `WORKDIR` in "test/assembler/Makefile" (what are the symptoms)?

SvOlli:
Not everything could be put in workdir, *.o was still left in the same directory as *.s, so building stuff out of tree seemed rather useless, since cleaning up the test/assembler directory still needs to be done. The Makefile is also a bit more readable that way.

oliverschmidt:
> *.o was still left in the same directory as *.s

See https://github.com/cc65/cc65/issues/168.

SvOlli:
I am now watching #168 so I can re-add WORKDIR, once the issue is fixed.

SvOlli:
I found a nice workaround: after building the assembler test target, the corresponding *.o file gets removed afterwards, so nothing has been left over. Also noted in a comment why this happens, so it can be removed once #168 is fixed. (Fixing #168 will not break this build, as the "-f" option to rm always returns 0, event if the file does not exist.)

SvOlli:
Has this one "aged enough" without objections to be acceptable? ;-)

oliverschmidt:
Sorry, I missed that you had answered Gregs question.

PR #349, opened 2016-09-12
Atari: fix lookup of default device on XDOS.

groessler:
Stefan Dorndorf, author of XDOS, pointed out that retrieving the
default device by looking at an undocumented memory location won't
work in future XDOS versions.
He also showed a way to get the default device in a compatible
manner.

This change implements his method and adds a version check (XDOS
versions below 2.4 don't support this -- for them the behaviour
will be the same as, for example, AtariDOS: no notion of a default
drive).

### DIFF
--- a/asminc/atari.inc
+++ b/asminc/atari.inc
@@ -1024,9 +1024,16 @@ XFILE        = $087D    ; XDOS filename buffer
 XLINE        = $0880    ; XDOS DUP input line
 XGLIN        = $0871    ; get line
 XSKIP        = $0874    ; skip parameter
+.ifdef __ATARIXL__
+.ifndef SHRAM_HANDLERS
+.import XMOVE_handler
+.endif
+.define XMOVE XMOVE_handler
+XMOVE_org    = $0877    ; move filename
+.else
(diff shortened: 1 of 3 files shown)


asminc/atari.inc
```diff
@@ -1024,9 +1024,16 @@ XFILE        = $087D    ; XDOS filename buffer
 XLINE        = $0880    ; XDOS DUP input line
 XGLIN        = $0871    ; get line
 XSKIP        = $0874    ; skip parameter
+.ifdef __ATARIXL__
+.ifndef SHRAM_HANDLERS
+.import XMOVE_handler
+.endif
+.define XMOVE XMOVE_handler
+XMOVE_org    = $0877    ; move filename
```
oliverschmidt: Just for curiosity: Why don't the other XDOS "entry points" require the same special considerations when using the "XL RAM" ?

groessler: Some of them are variables, not functions. And, I only added a wrapper for the function which is in fact used by the runtime lib.

oliverschmidt: Thanks for the background info :-)

PR #354, opened 2016-09-23
Add missing tag and remove duplicate text

sur5r:
The docbook toolchain as of Debian stretch seems to get more strict and no longer accepts the missing &lt;tt&gt; tag:

`onsgmls:<OSFD>0:3555:41:E: end tag for element "TT" which is not open`


oliverschmidt:

PR #358, opened 2016-11-19
scanner: Correct handling of \0101

EtchedPixels:
The C language has this oddity that octal constants are 3 bytes so the sequence
"\0101" is two bytes and well defined by the langage. cc65 currently misparses
this as a 1 byte octal code. Add a count to fix this.

Signed-off-by: Alan Cox <etchedpixels@gmail.com>

### DIFF
--- a/src/cc65/scanner.c
+++ b/src/cc65/scanner.c
@@ -267,6 +267,7 @@ static int ParseChar (void)
 {
     int C;
     int HadError;
+    int Count;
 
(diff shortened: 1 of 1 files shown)


greg-king5:
Only a three-digit octal number can be out of range.  Your patch prevents the range error from being seen twice.  Therefore, the `HadError` test is no longer needed.  Please add a commit that removes the `HadError` code from the octal parser.

PR #359, opened 2016-12-04
fix build break on da65

peterferrie:


oliverschmidt:
Hi Peter,

As you prove here I'm using VS only seldom to build the cc65 - basically only when wanting to use the VS debugger...

Let me take this opportunity to (again) express how much I appreciate your work on ProDOS ports!

And let me remark that I personally consider the replacement of ProDOS with a minimal DOS that's compatible both with the ProDOS filesystem as well as the ProDOS device driver interface only as a last resort because it breaks the compatibility with i.e. custom drivers replacing the Disk II driver inside ProDOS or file level (on contrast to block level) ProDOS extensions similiar to AppleTalk.

Regards,
Oliver

peterferrie:
Hi Oliver,

> Thanks for contributing to cc65@GitHub :-)
>
> As you prove here I'm using VS only seldom to build the cc65 - basically
> only when wanting to use the VS debugger...

I understand completely.
I'm looking to get back into contributing, so I needed a complete
build to be sure.

> Let me take this opportunity to (again) express how much I appreciate your
> work on ProDOS ports!

Thanks.  More will be coming as soon as I have time to work on them.

> And let me remark that I personally consider the replacement of ProDOS with
> a minimal DOS that's compatible both with the ProDOS filesystem as well as
> the ProDOS device driver interface only as a last resort because it breaks
> the compatibility with i.e. custom drivers replacing the Disk II driver
> inside ProDOS or file level (on contrast to block level) ProDOS extensions
> similiar to AppleTalk.

I agree, but when memory is at such a premium that there is no
alternative, it's good to have that alternative! :-)

greg-king5:
I'm curious to know, on which version of Visual Studio did you build the cc65 tools?

peterferrie:
> I'm curious to know, on which version of Visual Studio did you build the
> cc65 tools?

I'm using VS 2015 because it's what I have at work.
I'm also running GCC to build MAME.  The 64-bit version is emitting
some warnings which I will work to fix next.

oliverschmidt:
I'm btw using "Visual Studio Express 2013 for Windows Desktop".

Issue #339, opened 2016-08-25
CC65: Add "any" to --list-opt-steps  


cacciatc:
Since it is an accepted value for --enable-opt and --disable-opt.

oliverschmidt:
Fixed by https://github.com/cc65/cc65/pull/360

PR #360, opened 2016-12-15
Added 'any' to --list-opt-steps.

cacciatc:
Addresses: https://github.com/cc65/cc65/issues/339

### DIFF
--- a/src/cc65/codeopt.c
+++ b/src/cc65/codeopt.c
@@ -911,6 +911,8 @@ void ListOptSteps (FILE* F)
 /* List all optimization steps */
 {
     unsigned I;
+    
+    fprintf (F, "any\n");
     for (I = 0; I < OPTFUNC_COUNT; ++I) {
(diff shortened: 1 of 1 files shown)


src/cc65/codeopt.h
```diff
@@ -41,7 +41,7 @@
 /* cc65 */
 #include "codeseg.h"
 
-
+#define ANY "any"
```
oliverschmidt: Please remove this macro again and just use - yet another identical - string constant for the printf().

src/cc65/codeopt.c
```diff
@@ -911,6 +911,8 @@ void ListOptSteps (FILE* F)
 /* List all optimization steps */
 {
     unsigned I;
+    
+    fprintf (F, "%s\n", ANY);
```
greg-king5: I agree with Oliver;
```C
fputs ("any\n", F);
```
is good enough.

cacciatc:
Changed macro back to string constant.

PR #361, opened 2016-12-20
NES memory map amend (16k prg, 8k chr default)

marshallward:
The configuration file and runtime (crt0.s) provided for the default NES
ROM layout (2x16k PRG, 8k CHR) incorrectly added interrupts (IRQ1, IRQ2,
TIMERIRQ) which are not supported by the NES hardware.  For example, see
the NESdev wiki, which makes no reference to these interrupts.

    https://wiki.nesdev.com/w/index.php/CPU_memory_map

The VECTORS region was also incorrectly set to 0xFFF6, which would have
left the 0xFFF4 normally unspecified.  This did not result in any error,
however, since cc65 simply placed ROMV directly after ROM0 regardless of
start address.

(This layout may be due to a copy-and-paste from the PC-Engine
configuration, whose interrupt registers start at 0xFFF6, begins with
the three interrupts listed above, followed by NMI and START, and does
not end with a final IRQ interrupt.)

Despite the absence of any actual error, since START is still placed at
0xFFFC, this patch removes the nonexistent interrupts and also correctly
aligns the ROM0 and ROMV regions.  It also has the (admittedly very
minor) benefit of freeing up 3 additional bytes for ROM0.

### DIFF
--- a/cfg/nes.cfg
+++ b/cfg/nes.cfg
@@ -12,10 +12,10 @@ MEMORY {
     # - code
     # - rodata
     # - data (load)
-    ROM0:   file = %O, start = $8000, size = $7FF4, fill = yes, define = yes;
+    ROM0:   file = %O, start = $8000, size = $7FFA, fill = yes, define = yes;
 
     # Hardware Vectors at End of 2nd 8K ROM
-    ROMV:   file = %O, start = $FFF6, size = $000C, fill = yes;
+    ROMV:   file = %O, start = $FFFA, size = $0006, fill = yes;
 
     # 1 8k CHR Bank
     ROM2:   file = %O, start = $0000, size = $2000, fill = yes;
--- a/libsrc/nes/crt0.s
+++ b/libsrc/nes/crt0.s
@@ -159,9 +159,6 @@ nmi:    pha
 
 ; Interrupt exit
 
-irq2:
-irq1:
-timerirq:
 irq:
         rti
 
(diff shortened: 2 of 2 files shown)


fo-fo:
Seems good to me.

The only question is if there's value in maintaining backwards compatibility with code which assumes that the `VECTORS` segment is 12 bytes. I know for a fact there's code like that out there, but then again the required changes would be trivial.

oliverschmidt:
> Seems good to me.

Thanks for your quick feedback :-)

> The only question is if there's value in maintaining backwards compatibility with code which assumes that the VECTORS segment is 12 bytes.

I'm in general always much more into consistency and easy maintenance than backward compatibility. And apart from that the linker configs are known since "ages" to be the least backward compatible aspect of cc65 ;-)

bbbradsmith:
It also seems fine to me. The 3 symbols have no plausible use on the NES, and don't appear to be exported, so there should be no negative consequence here, and the benefit of 6 saved bytes.

marshallward:
Oops, six bytes, thanks for noting the correction :) (Every Byte Matters!).  I've amended the git log.

polluks:
BTW the PC-Engine target is newer.

mrdudz:
i think they are both based on my work.... and i just provided the PCE stuff much later (they were made pretty much at the same time)

oliverschmidt:
This pull request seems to be ready to be merged...

PR #365, opened 2017-01-08
Add C support for Atari 2600 (VCS)

FlorentFlament:


polluks:
Pretty cool, how about supporting a batari Basic kernel? Let's call it "bC".

include/_tia.h
```diff
@@ -0,0 +1,79 @@
...
+    unsigned char CXBLPF;
+    unsigned char CXPPMM;
+    unsigned char INPT0;
+    unsigned char INPT1;
+    unsigned char INPT2;
+    unsigned char INPT3;
+    unsigned char INPT4;
+    unsigned char INPT5;
+  };
+  unsigned char data[45]; // For an unknown reason, this array is required for
```
spiro-trikaliotis: Perhaps, this should be discussed a little bit. What erratic behaviour do you see without this array?

libsrc/atari2600/crt0.s
```diff
@@ -0,0 +1,46 @@
...
+        .include        "zeropage.inc"
+
+
+.segment "STARTUP"
+start:
+; Clear decimal mode
+        cld
+; Clear zeropage & registers
+; Also Initializes system stack pointer
+        ldx     #0
```
spiro-trikaliotis: Perhaps I am missing something obvious, but the dex / txs in the loop does not look right to me.

Why not:

ldx #0
txa
txs
clearLoop
pha
dex
bne clearLoop

I know, this is not completely the same, as I am writing the stack differently.

Even better, omitting the pha completely would save many cycles, although you would need two more byte:

ldx #0
txa
txs
clearLoop
sta $0100,x
dex
bne clearLoop

include/_tia.h
FlorentFlament: Here it is:
```
$ cat test.c
union {
  struct {
    int a;
    int b;
  };
  struct {
    int c;
    int d;
  };
  //int t[2];
} u;

void main(void) {
  u.a = 0;
}
$ gcc test.c
$ cc65 test.c
test.c(11): Error: Variable `u' has unknown size
test.c(14): Error: Struct/union has no field named `a'
test.c(14): Error: Invalid lvalue in assignment
$ 
```

libsrc/atari2600/crt0.s
FlorentFlament: I agree this loop is a bit tricky. Nothing obvious, but with your first approach, you are missing 1 pha (After the loop the stack pointer's value is 0 while it should be $ff), which would led to using 1 additional byte of ROM for the same result. And for the second approach, you consume more bytes and the stack pointer's value is still 0.

FlorentFlament: Though, we might want to sacrifice 1 byte in favour of more clarity. What would you prefer ?

FlorentFlament:
@polluks ; ) Looks like it has already been done there: http://bataribasic.com/

include/_tia.h
oliverschmidt: As far as I can see this is a bug in the cc65 compiler. However there seems to be a way to archive the desired result without a workaround - at least such an explicit one:

```
struct {
  union {
    int a;
    int c;
  };
  union {
    int b;
    int d;
  };
} u;

void main(void) {
  u.a = 0;
}
```

If you agree that this works I ask you to change your code to this approach.

oliverschmidt:
Are you aware of https://github.com/cc65/wiki/wiki/Atari-2600 ? The content was migrated from http://wiki.cc65.org/doku.php?id=cc65:atari_2600 and originally created in 2010 by 'kes'.

So do I understand it correctly that "everybody" agrees that the statement "For a machine with only 128 bytes of RAM for its data and stack, programming it using any language other than assembly is not very practical." on that Wiki page is wrong? Anyhow before I'm willing to merge this pull request it is necessary to clarify what is to be done with that wiki page.

Another aspect is the missing documentation: As far as I can tell the level of support in cc65 for the 2600 can be compared to the one for the Watara Supervision Console. So it might make sense to check that documentation to get an idea.

libsrc/atari2600/crt0.s
```diff
@@ -0,0 +1,46 @@
...
+        txa
+clearLoop:
+        dex
+        txs
+        pha
+        bne     clearLoop
+
+; Initialize data
```
oliverschmidt: I understand that we want to minimize the overhead for this target but nevertheless a call to `        jsr zerobss` (usually done before the `jsr copydata') seems obligatory to me.

oliverschmidt:
Another aspect I consider obligatory is a `ctype.s` file. I imagine that i.e. the Supervision one can be copied: https://github.com/cc65/cc65/blob/master/libsrc/supervision/ctype.s

include/_tia.h
FlorentFlament: Fair enough, this workaround seems reasonable.

libsrc/atari2600/crt0.s
FlorentFlament: The initialization loop clears the whole (128 bytes of) memory with 0s (including the BSS segment). I don't see the advantage of calling `zerobss` (unless it does something else than writing zeros to the BSS segment, that I've not seen).

FlorentFlament:
Obviously I don't agree with kes' statement on the wiki. Together with my team (http://www.pouet.net/groups.php?which=12080), we are currently working on an Atari 2600 demo mixing C with asm, which is perfectly fine.
I'll add some documentation.

FlorentFlament:
As for the `ctype.s`, I think that adding  the 256  bytes table to an Atari 2600 binary should be optional, since a programmer may not need to use the `ctype` helpers. In that case spending 256 bytes (out of 4096 in the case of a 4k cartridge) for an unused table wouldn't be reasonable.

FlorentFlament:
My bad. Regarding the ctype.s, the table won't be included in the binary file if it is not used. I'll add the file to the atari2600 libsrc

libsrc/atari2600/crt0.s
oliverschmidt: Thanks for pointing out. In that case I ask you to add a comment clarifying that your initialization loop zeros the BSS (too).

oliverschmidt:
> Obviously I don't agree with kes' statement on the wiki.

I see. Do you think the Wiki page has still any relevance after merging your pull request? I'm wondering if I can/should just delete it or if it needs to be modified.

include/_tia.h
greg-king5: I think that Uz called it "a partially implemented C99 feature".  He wrote only enough cc65 code to allow anonymous bit-fields in structs, and the code in <_vic2.h>.  Therefore, structs don't need any named fields; but, unions need at least one.

greg-king5: I think that Uz called it "a partially implemented C99 feature".  He wrote only enough cc65 code to allow anonymous bit-fields in structs, and the code in <_vic2.h>.  Therefore, structs don't need any named fields; but, unions need at least one.

libsrc/atari2600/crt0.s
greg-king5: You should aim for the smallest size everywhere that doesn't _require_ high speed.

greg-king5:
Our tradition is that I/O register names are upper-case in Assembly code, but lower-case in C code.  We use upper-case names for macro definitions.

libsrc/atari2600/crt0.s
FlorentFlament: Sounds good. Let's keep this code as is then

libsrc/atari2600/crt0.s
FlorentFlament: Fair enough, I'll add some comments there to provide more details

FlorentFlament:
@oliverschmidt IMO, the wikipage on the Atari 2600 doesn't help much. I'd just remove it.

FlorentFlament:
@greg-king5 Fair enough, I'll try to stick with your tradition and convert the register names to lower-case in the C headers.

greg-king5:
We almost forgot this:  We need an "`atari2600.inc`" include file that mirrors the "`atari2600.h`" header.  It should define the I/O register addresses (use the `:=` operator).  And, it should define some of the more common (default) values that can be stuffed into those registers (use the `=` operator).

cfg/atari2600.cfg
```diff
@@ -0,0 +1,22 @@
...
+#
+# Florent Flament (contact@florentflament.com), 2017
+
+SYMBOLS {
+    __STACKSIZE__: type = weak, value = $0010;   # 16 Bytes system stack
+}
+
+MEMORY {
+    RAM: file = "", start = $0080, size = $0080 - __STACKSIZE__, define = yes;
+    ROM: file = %O, start = $F000, size = $1000;
```
greg-king5: I think that the "`, fillval = $FF`" attribute needs to be in that line (unused locations in PROMs should have their bits set to one).

libsrc/atari2600/crt0.s
```diff
@@ -0,0 +1,49 @@
...
+
+
+.segment "STARTUP"
+start:
+; Clear decimal mode
+        cld
+
+; Initialization Loop:
+; * Clears Atari 2600 whole memory (128 bytes) including BSS segment
+; * Clears TIA registers
```
greg-king5: It would be nice if that TIA clearing was mentioned in the document (maybe, in a section about start-up conditions).

greg-king5:
> Pretty cool, how about supporting a batari BASIC kernel? Let's call it "bC".

@polluks, are you suggesting that we should be able to create our own kernels for batari BASIC; or, are you saying that we should be able to call into the existing kernels from C and ca65 Assembly source files?

cfg/atari2600.cfg
FlorentFlament: @greg-king5 Ok. For my information, why is it preferred to fill unused locations with $ff ?

libsrc/atari2600/crt0.s
FlorentFlament: Fair enough, I'll add that to the doc.

FlorentFlament:
@greg-king5 Regarding the atari2600.inc file, I actually planned to add it in a next pull request. But I can do that altogether if that's the recommended approach.

greg-king5:
> Regarding the "atari2600.inc" file, I actually planned to add it in a next pull request. But, I can do that altogether if that's the recommended approach.

The timing of it isn't vitally important; in my opinion, adding it by a later PR is OK.  (I just wanted to make sure that it had not been overlooked.)

cfg/atari2600.cfg
greg-king5: As I understand it, after a PROM (and flash memory, too) is erased, all of its bits are %1.  PROM programming software takes advantage of that.  They don't write $FF bytes; instead, they just skip the location.  Therefore, it is useful to fill unused space with $FF when creating image files for ROMs.

polluks:
@greg-king5 you can already create a kernel if you convert DASM to ca65. But right now you have no sprite or playfield macros.

cfg/atari2600.cfg
FlorentFlament: Thks, good to know. I updated the fillval in my last PR.

FlorentFlament:
@greg-king5 Regarding the atari2600.inc, I added it anyway in my last update.

Regards

asminc/atari2600.inc
```diff
@@ -0,0 +1,7 @@
+; Atari 2600 TIA & RIOT read / write registers
+;
+; Florent Flament (contact@florentflament.com), 2017
+
+; TIA & RIOT registers mapping
+.include "atari2600_tia.inc"
+.include "atari2600_riot.inc"
```
polluks: Please check EOL

FlorentFlament: Fixed. I also double checked the other files.

doc/atari2600.sgml
```diff
@@ -0,0 +1,124 @@
...
+
+
+<sect>Limitations<p>
+
+TBD
+
+
+<sect>Other hints<p>
+
+One may write her own linker configuration file to tune the memory
```
oliverschmidt: Is this "her" some new political correctness stunt that I'm not aware of? Anyhow the cc65 docutmentation tends to be used over a long period of time so I'd like it to rather not reflect some fashion that goes out of style soon. I guess any native speaker can come up with phrases at least as good as my proposals "... write a custom linker ..." and "... layout of the own program."

doc/atari2600.sgml
```diff
@@ -0,0 +1,124 @@
...
+  declaration of the structure. Also refer to the Stella Programmer's
+  Guide by Steve Wright for a detailed description of the chip and its
+  registers.
+
+</descrip><p>
+
+
+<sect>Loadable drivers<p>
+
+There is no drivers for the Atari 2600.
```
polluks: is --> are

doc/intro.sgml
```diff
@@ -298,7 +298,7 @@ in completing this section.
 Available at <url
 url="http://www.atari.org.pl/PLus/index_us.htm">:
 
-Emulates Atari 400/800/65XE/130XE/800XL/1200XL/5200, with stereo sound, disk
+Emulates Atari 400/800/65XE/130XE/800XL/1200XL/2600/5200, with stereo sound, disk
```
polluks: Are you sure?

doc/atari2600.sgml
FlorentFlament: ; ) That's a fashion that's been around for the last couple of years. I'll change that.

doc/atari2600.sgml
FlorentFlament: nice catch

doc/intro.sgml
FlorentFlament: Oups, I went too quickly on that one.

FlorentFlament: I'll document the usage of the Stella emulator

FlorentFlament:
@oliverschmidt Yes, I think that I have addressed all the current comments..

greg-king5:
:+1:

oliverschmidt:
Some more general words:

From my perspective this review process went very well :-)

* Several cc65 community members participated in providing constructive feedback.
* The pull request owner acted in a both very quickly and positive way on the feedback.

FlorentFlament:
thks ; )

PR #371, opened 2017-01-29
Create an Oric Telestrat target library

jedeoric:
Here is pull request for telemon 2.4 and telemon 3.0 targets. Both are roms for Oric telestrat.

https://en.wikipedia.org/wiki/Oric

Telestrat is the third Oric. Telemon 2.4 is the last rom released in 1986. Telemon 3.0 is released this year in 2017. For this last rom, you can see : http://orix.oric.org

I hope everything is OK :) I will do some update in theses targets in few weeks :)


polluks:
How about sharing code between 2.4 and 3.0 like this https://github.com/cc65/cc65/pull/370/files?

jedeoric:
Hello,

I can, but i don't know what is the best practice on cc65. 

What is the best case ? I don't know if it's bad to keep dependencies. Let me know :)

greg-king5:
I do not like the idea of having target platforms for ROMs.  I prefer that targets be machines
(in this case, `-t telestrat`).  The Telestrat library can be written for Telemon 3.  And, a special CFG file can be added for Telemon 2.

Besides the different amount of available RAM, is there anything that would prevent a program written for Telemon 2.4 from using the Telemon 3 library (e.g., variables with different addresses in page zero)?

jedeoric:
In fact, i prefer to have only one target (telestrat). I did 2 targets because i was thinking that it was the idea of cc65. But, i can rename target Telemon30 into telestrat and i can merge code which is the same as telemon24 . 

Telemon 3.0 = Telemon 2.4 without minitel videotex (https://en.wikipedia.org/wiki/Minitel) routines. The space  (earned with videotex routines removed) is used to add open/fread/fclose etc routines.

I doubt that someone will need these videotex primitives (and in cc65) in 2017.

Telemon3.0 and telemon2.4 has the same adresses (ZP, $2xx, $4xx, $5xx)

I modify this in order to have one target, if you are OK ?

greg-king5:
>I modify this in order to have one target, if you are OK.

Yes, thanks.

Here are some preliminary comments:

In Assembly code:
**`:=`**  should be used to assign addresses to names.
**`=`**  should be used to assign non-addresses (such as Telemon's command codes) to names.

The file `_read.s` should be named `read.s` (no underscore).

Everywhere, except in makefiles, tab characters should be replaced by enough spaces to keep the code alignment.  See the [code-style guidelines](https://github.com/cc65/wiki/wiki#this-cc65-projects-coding-guidelines).

jedeoric:
ok, i did all the changes. I think it's ok, if something is still wrong let me know.

oliverschmidt:
Some general remarks:

- At least from my perspective fewer target are better than more. Look i.e. at the number of Apple II machines https://github.com/cc65/cc65/blob/master/libsrc/apple2/get_ostype.s#L35 covered with two targets. Similiar situation for the ATARI 8-bit machines https://en.wikipedia.org/wiki/Atari_8-bit_family covered with two targets. I don't know if it's appropriate from that perspective to have so many CBM targets - but that's adifferent story.

- The granularity of the targets as they are presented to the cc65 user isn't necessarily reflected by the internal code organization. There's for all the CBM targets a directory https://github.com/cc65/cc65/tree/master/libsrc/cbm with code shared between them. Then there's the option to have two targets share code from one directory: https://github.com/cc65/cc65/tree/master/libsrc/apple2 is used for both `apple2` and `apple2enh` while https://github.com/cc65/cc65/tree/master/libsrc/atari is used for both `atari` and `atarixl`. And finally there's the option to "share" single files by replacing one instance with a one-liner including the other. So there are really many ways to stay clear from code duplication in cc65.

- Although not clearly specified a cc65 target needs to have some relevance (however you define that) in order to become a target in this cc65@GitHub upstream repo. I'm personally very reluctant to consider a "custom ROM" like telemon 3.0 qualifying for that. I.e. the http://singularcrew.hu/idedos/ has an API for writing reading/writing a block of memory from/to an IDE64 drive. This API would allow to greatly improve the performance of POSIX read()/write(). However in a former discussion the API was considered not relevant enough to become part of this cc65 repo.

doc/telestrat.sgml
```diff
@@ -0,0 +1,177 @@
...
+reference">. Even functions marked as "platform dependent" may be available on
+more than one platform. Please see the function reference for more
+information.
+
+
+
+<sect>Binary format<p>
+
+The standard binary output format generated by the linker for the Telemon 3.0 target
+is a machine language program with a 20 bytes header described here : http://orix.oric.org/doku.php?id=orix:header
```
oliverschmidt: - You can't presume the Telemon 3.0 thing to be known. You need to elaborate much more about it. The link given doesn't explain _ANYTHING_. I guess that the Telemon 3.0 uses the .o65 file format. If yes: Why? Is the file to be placed at different locations? Are references into ROM/OS/DOS routines not at fixed locations? I think you get the point of info I'd like to either see here or on a page @ orix.oric.org and linked to from here.

- If this isn't only about Telemon 3.0 then - as far as I understand - you need (at least) two different binary formats. One for the Telemon 2.4 (what ever that might look like) and the .o65 (?) one. Both need to be described here. And you need to find a way to have both generated. Usually this is done via a) parameters in the .cfg file or b) several .cfg files. Check out the Apple II and ATARI sections @cc65 to get an idea on multiple .cfg files for different binary formats. If it's only about the header then you might get away with one .cfg file and a command line parameter. See http://cc65.github.io/doc/apple2.html#ss4.4 for an example of an optional header.

- If there's a binary format compatible with both 2.4 and 3.0 that one should be the default. The 3.0 specific binary format needs to be the "additional option" - even if that file format (and the whole 3.0 thing in general) should be your personal motivation to add a target to cc65 in the first place.

doc/telestrat.sgml
```diff
@@ -0,0 +1,177 @@
...
+</descrip><p>
+
+
+
+<sect>Limitations<label id="limitations"><p>
+
+<sect1>Disk I/O<p>
+
+This version handles fopen, fread, fclose primitives. Because Telemon 3.0 handles these two primitives. By the way, 
+it uses an extension "ch376 card" which handles sdcard and FAT 32 usb key.
```
oliverschmidt: - We're under `Limitations` here. If disk I/O works as expected then there's no need for such a section at all. If disk I/O isn't fully available (as I understand it's not in 2.4) then this is what needs to be stated here. And if the 2.4 has the same limitations the ATMOS has (with the same "hack" to allow stdin / stdout) then the statements from the ATMOS should be used here.

- I understand that the file I/O we're talking about here requires the "ch376" thingy to provide an actual file system to access. If that is true you certainly need to elaborate on this. Again either here or on orix.oric.org and link to it from here. This requirement has nothing to do with the fact that this file I/O is - as far as I understand - new/uncommon. I.e. the Apple II and the ATARI have several DOSes if not all of them are supported by the cc65 file I/O then this needs to be clarified.

libsrc/telestrat/ch376.s
```diff
@@ -0,0 +1,344 @@
+
+; jede jede@oric.org 2017-01-22
+
+	; For XA65 compatibily in the future
+    .FEATURE 	c_comments,labels_without_colons,pc_assignment, loose_char_term
```
oliverschmidt: Sorry, but source code in the cc65 upstream repo is for the cc65 tool chain ;-)

libsrc/telestrat/ch376.s
```diff
@@ -0,0 +1,344 @@
...
+    ; error is here
+    rts
+no_error:
+    lda     #CH376_GET_STATUS
+    sta     CH376_COMMAND
+    lda     CH376_DATA
+    rts
+.endproc 
+
+.proc _ch376_fread
```
oliverschmidt: I'm pretty sure you are aware that a C library implements i.e. `fread()` using `read()`. So I'm wondering why there should be anything named `fread` in a target-specific source. Same goes for the other `f...` functions.

libsrc/telestrat/print.s
```diff
@@ -0,0 +1,16 @@
...
+; void print (char * str);
+    .export         _print
+    .import         popax
+    .importzp       tmp1
+    .include        "telestrat.inc"
+
+.proc   _print
+    stx tmp1
+    ldy tmp1
+    BRK_TELEMON XWSTR0
```
oliverschmidt: As far as I understand this is the same as a `write()` to fd 1. In that case please remove it. cc65 C libraries are not supposed to duplicate functionality already present through standard APIs. Just check out the other targets. All of them have some nice straight-forward ROM entry point for "just" printing a char on the screen, but non of them has a `print()` C function.

libsrc/telestrat/keyboard.s
```diff
@@ -0,0 +1,14 @@
...
+
+    .export         _key
+				
+    .include        "zeropage.inc"		
+    .include        "telestrat.inc"
+
+; char key(void);
+		
+.proc _key
+    BRK_TELEMON XRDW0 ; read keyboard
```
oliverschmidt: As far as I understand this is the same as a `read()` from fd 0. In that case please remove it. cc65 C libraries are not supposed to duplicate functionality already present through standard APIs. Just check out the other targets. Most of them have some nice straight-forward ROM entry point for "just" get a char from the keyboard, but non of them has some `key()` C function.

cfg/telestrat.cfg
```diff
@@ -0,0 +1,43 @@
+SYMBOLS {
+
+    __ORIXHDR__:   type = import;
+
```
oliverschmidt: Please remove this empty line.

cfg/telestrat.cfg
```diff
@@ -0,0 +1,43 @@
...
+    __STACKSIZE__: type = weak, value = $0800; # 2K stack
+
```
oliverschmidt: Please remove this empty line.

libsrc/telestrat/orixhdr.s
```diff
@@ -0,0 +1,35 @@
...
+        .import __AUTORUN__, __PROGFLAG__
+        .import __BASHEAD_START__, __MAIN_LAST__
+
+
+; ------------------------------------------------------------------------
+; Orix header see http://orix.oric.org/doku.php?id=orix:header for specs
+
+.segment        "ORIXHDR"
+
+    .byte   $01, $00          ;
```
oliverschmidt: Missing comment.

oliverschmidt:
- mym.s
I don't see a music player being part of the C library.

- ch376.s
Most of the functions seem to duplicate standard C library functions. Please elaborate on your intentions.

- graphics.s
As far as I can see this should be converted into a TGI driver.

oliverschmidt:
- The changes in the .vcxproj files are for sure not related to the new target so they don't belong in this pull request.

- As far as understand the changes in the .vcxproj files are the result of using a newer Visual Studio. The benefit of having them in the repo would be that one wouldn't be told by Visual Studio that they are of some older version and that Visual Studio will import/update them. However the downside would be that older versions of Visual Studio won't be able to open them anymore. As you surely guess by now I consider the benefit (much) smaller than the downside. So as long as the latest Visual Studio imports the old files without issues I don't see them "upgraded".

doc/telestrat.sgml
```diff
@@ -0,0 +1,177 @@
...
+
+The functions listed below are special for the Telemon 3.0. See the <url
+url="funcref.html" name="function reference"> for declaration and usage.
+
+<itemize>
+<item>explode
+<item>ping
+<item>shoot
+<item>zap
+<item>oupsx
```
jedeoric: The main problem for : 

atmos.h would include oric.h
telestrat.h would include oric.h

there are atmos_zap(), atmos_explode(). Theses functions are tje same for telestrat. I would like to have zap(), ping() instead. But, in this way, some code will be broken.

Let me know, because i don't know how to do this fix

doc/telestrat.sgml
```diff
@@ -0,0 +1,177 @@
...
+
+<!-- Begin the document -->
+
+<sect>Overview<p>
+
+This file contains an overview of the Telestrat runtime system as it comes with the
+cc65 C compiler. It describes the memory layout, Telemon3.0-specific header files,
+available drivers, and any pitfalls specific to that platform.
+
+Please note that Telemon3.0-specific functions are just mentioned here, they are
```
jedeoric: Yes i replaced Telemon into Telestrat, "push" is coming soon

doc/telestrat.sgml
jedeoric: - For the first point, i am explaining a bit more in the sgml doc, and i'll update orix.oric.org for the header section.

- For the second point, yes it needs 2 headers, but it's not simple,  I updated sgml too, but it's a bit difficult on Telestrat for Telemon 2.4. 

When you have a Telemon 2.4 rom, you can't start any file from disk. Telemon 2.4 loads floppy disk boot sector and loads Stratsed Operating system from disk. In this way, the binary must be inserted on a Stratsed disk. But this tool does not exist, and should never exists because Oric international died before releasing docs. I would like to do it, but i have too many things to do before.
The only way is to use Floppy Builder from Osdk (http://osdk.defence-force.org/) to insert the binary sector per sector, track per track, but nobody will use this trick, because it's too complicated when you need to test your code. I explain this in the sgml doc.

That is why i created Orix and Telemon 3.0. I did Orix header like o65 format, but it's not the same : I would like to use o65 format in the future, and the best way, is to have similar header to share code.

I can manage Stratsed header (and i have less header doc for this one than Orix header). Let me know, but nobody will use Stratsed feature, because it's too much complicated to use it (mainly because there is a lack of tools)

- For the last point, there is no common header. I did my own header to manage 6502, 65C02, 65C816, and others stuffs. That is why i wanted to add telestrat target. There is 2000 Telestrat in the world, but, we are maybe 5 persons who are using the real hardware. The video on orix.oric.org is on emulator. If you want to see telestrat with telemon 3.0 and Orix working, here is the link : https://www.youtube.com/watch?v=TkzdlqF6-x8&list=PLugWQmG9T0gr9hSZNtmvDWkv2jpSwOW48

doc/telestrat.sgml
```diff
@@ -0,0 +1,177 @@
...
+<item>ping
+<item>shoot
+<item>zap
+<item>oupsx
+</itemize>
+
+
+<sect1>Hardware access<p>
+
+The following pseudo variables declared in the <tt/atmos.h/ header file do allow
```
jedeoric: It's a bug, it's corrected here, i will push as soon as possible

doc/telestrat.sgml
jedeoric: I did "open, read, close" because i would like to add in Orix, mount command which could mount sedoric/ftdos/stratsed file system. I added some explanation in sgml too.

libsrc/telestrat/oserror.s
```diff
@@ -0,0 +1,17 @@
...
+; /* Map a system specific error into a system independent code */
+;
+
+        .include        "errno.inc"
+        .export         __osmaperrno
+
+.proc   __osmaperrno
+
+        lda     #<EUNKNOWN
+        ldx     #>EUNKNOWN
```
jedeoric: I did initial functions, because it's in my plans to do it, but it's not finished, i removed theses 2 files from my repo. I'll finished it.

README.md
```diff
@@ -30,6 +30,7 @@ including
 - the Nintendo Entertainment System (NES) console.
 - the Watara Supervision console.
 - the Oric Atmos.
+- the Oric telestrat.
```
jedeoric: I corrected here.

libsrc/telestrat/ch376.s
jedeoric: On the Oric, we are using xa65. I did this to share code between cc65 and C compiler in OSDK. Removed here.

asminc/telestrat.inc
```diff
@@ -0,0 +1,147 @@
...
+
+; ---------------------------------------------------------------------------
+; Zero page
+
+; ---------------------------------------------------------------------------
+; Page 00
+RES             := $00
+RESB            := $02
+
+TR0	            := $0C
```
jedeoric: Well i am looking to this here

libsrc/telestrat/ch376.s
jedeoric: It's a code to test loading from usbkey because i did a tool with cc65 to test all the functionnalities of the chip. I removed ch376.s file in order to link with my tool, instead of sharing the source code in cc65

libsrc/telestrat/print.s
jedeoric: Yes, it's quickest way to write a string to video string. I removed this code, and i'll see in few day how to manage it

libsrc/telestrat/keyboard.s
jedeoric: i removed this file too. I will have a look. I did this to share headers/functions between cc65 and Oric C compiler (osdk) that is why i did this. But, i understand that cc65 must keep standard C librairies.

cfg/telestrat.cfg
jedeoric: Corrected here

cfg/telestrat.cfg
jedeoric: Corrected here

libsrc/telestrat/orixhdr.s
jedeoric: Fixed here

jedeoric:
about : mym.s, ch376.s and graphics.s, i removed theses files, i wanted to share the code for anybody in cc65, but i'll distribute theses files in an other way. graphics.s is removed, because if i have to do TGI driver, it's some time that i don't have.

jedeoric:
for the .vcxproj, i am using Visual Studio 2015.

jedeoric:
Hello,

I think i have finished and fixed all things related in comments above. 

I tried yesterday and today to find a trick to insert binary on Stratsed disk (for Telemon 2.4), and there is no way to insert a binary. The only thing is to use FloppyBuilder in OSDK package "http://osdk.defence-force.org/". This tool insert a binary sector by sector, track by track and bootsector is read by Telemon 2.4.

I think that cc65 target should manage Telemon 3.0/Orix header.

If something is still wrong, let me know.

libsrc/telestrat/scrsize.s
```diff
@@ -0,0 +1,19 @@
+;
```
greg-king5: This file is an exception to the comment that I wrote earlier about a leading underscore in library source file names.  It should be named "`_scrsize.s`".  It is a target-dependent helper for "`libsrc/conio/scrsize.s`".

jedeoric:
i corrected scrsize.s into _scrsize.s

doc/telestrat.sgml
```diff
@@ -0,0 +1,185 @@
...
+
+Telestrat manages also mouse, but it had been no handled yet in this version.
+
+</descrip>
+
+<sect1>RS232 device drivers<p>
+
+<descrip>
+
+Telestrat has a RS232 port, but it's not used
```
polluks: atmos-acia.ser is ready for Telestrat :-)

jedeoric: Yes thanks, i will have a look. 

I would like to have the telestrat target merged, before working on others parts as TGI, RS232 and others stuffs

greg-king5:
> - ch376.s
> Most of the functions seem to duplicate standard C library functions. Please elaborate on your intentions.

As I see it, those functions would be a part of the hardware-level foundation of the POSIX I/O functions.  But, they need to be put into separate files, so that a program would link to only the functions that it actually needs.

I think that your `key()` function can be put back into the library; but, it must be renamed to `cgetc()`.  That function's prototype is declared in "`conio.h`".  The function should be implemented in a file called "`cgetc.s`".

doc/telestrat.sgml
polluks: My TGI uses six OS calls, JOY should work out of the box...

doc/telestrat.sgml
```diff
@@ -0,0 +1,185 @@
...
+
+</descrip><
+
+<sect>Limitations<label id="limitations"><p>
+
+<sect1>Disk I/O<p>
+
+Telemon 3.0 handles fopen, fread, fclose primitives. It means that this function will crash the Telestrat because Telemon 2.4 did not have not these primitives.
```
oliverschmidt: Telemon 2.4 did not have -> Telemon 2.4 doesn't have

src/ar65.vcxproj
```diff
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
```
oliverschmidt: Leave this file alone.

src/ca65.vcxproj
```diff
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
```
oliverschmidt: Leave this file alone.

src/cc65.vcxproj
```diff
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
```
oliverschmidt: Leave this file alone.

src/cl65.vcxproj
```diff
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
```
oliverschmidt: Leave this file alone.

src/co65.vcxproj
```diff
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
```
oliverschmidt: Leave this file alone.

src/common.vcxproj
```diff
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
```
oliverschmidt: Leave this file alone.

src/da65.vcxproj
```diff
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
```
oliverschmidt: Leave this file alone.

src/grc65.vcxproj
```diff
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
```
oliverschmidt: Leave this file alone.

src/ld65.vcxproj
```diff
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
```
oliverschmidt: Leave this file alone.

src/od65.vcxproj
```diff
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
```
oliverschmidt: Leave this file alone.

src/sim65.vcxproj
```diff
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
```
oliverschmidt: Leave this file alone.

src/sp65.vcxproj
```diff
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
```
oliverschmidt: Leave this file alone.

oliverschmidt:
> [...] When you have a Telemon 2.4 rom, you can't start any file from disk. [...]

Thanks for you detailed explanation. I assumed somehow that Telestrad was a superset of Atmos and that therefore http://cc65.github.io/doc/atmos.html#s2 would apply to Telemon 2.4 but that obviously isn't the case.

oliverschmidt:
Do we all agree that after removal of the .vcxproj file changes (and the minor doc change) this pull request is ready to be merged?

doc/telestrat.sgml
jedeoric: For TGI, i should have a look, because telemon (the right telestrat rom) has not same calls than the atmos, not the same parameters, and as i did a lot of reverse engineer for telemon, it require some work. For joysticks, there is also routines for joystick management in telemon, i will have a loop too.

doc/telestrat.sgml
jedeoric: ok corrected here, i will push it as soon as possible.

@greg-king5 : ch376 is chip which could be use in any 8 bits cpu, because it uses a 8 bits data bus. I can share this to community, because theses routines will work in any 6502 target platform, the main problem is that there a command port and a data port on this chip which could not be the same if the ch376 is used on the others target. 
For key() converted into cget, i will have a look like TGI too

jedeoric:
@oliverschmidt about telestrat/atmos. Telestrat is an atmos with : 2 cardridges port, another 6522 (for joysticks/bank switch) and a RS232 port. 

Telemon does not support tape, but if we introduce the atmos cardridge (which is the Atmos rom), tape will work has atmos (it means that we will use atmos target).

jedeoric:
Well, for merge request, i did all the changes. I will have a look to TGI/conio/rs232.

src/da65.vcxproj
```diff
@@ -86,6 +86,7 @@
     <ClCompile Include="da65\infofile.c" />
     <ClCompile Include="da65\labels.c" />
     <ClCompile Include="da65\main.c" />
+    <ClCompile Include="da65\opc4510.c" />
```
oliverschmidt: I guess this change isn't actually related to this pull request but rather a fix/addon to a 3rd party change. I generally prefer a separate pull request in such a case but I'm willing to merge this here.

src/da65.vcxproj
```diff
@@ -109,6 +110,7 @@
     <ClInclude Include="da65\handler.h" />
     <ClInclude Include="da65\infofile.h" />
     <ClInclude Include="da65\labels.h" />
+    <ClInclude Include="da65\opc4510.h" />
```
oliverschmidt: I guess this change isn't actually related to this pull request but rather a fix/addon to a 3rd party change. I generally prefer a separate pull request in such a case but I'm willing to merge this here.

src/da65.vcxproj
greg-king5: That "da65.vcxproj" diff is a Git artifact.  It shouldn't be shown because it was merged by PR #359.

PR #373, opened 2017-02-01
pce/memcpy.s: remove superfluous comma in comment

groessler:
I'm not a native English speaker but I think this comma is wrong...

### DIFF
--- a/libsrc/pce/memcpy.s
+++ b/libsrc/pce/memcpy.s
@@ -97,7 +97,7 @@ memcpy_getparams:
 
 ; ----------------------------------------------------------------------
 ; The transfer instructions use inline arguments.
-; Therefore, we must build the instruction, in the DATA segment.
+; Therefore, we must build the instruction in the DATA segment.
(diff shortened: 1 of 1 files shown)


mrdudz:
me neither.... but i think even the first one is questionable :)

oliverschmidt:
> even the first one is questionable

Greg adds those all over the place. I'm not in the mode to question them.

greg-king5:
Commas in english have two purposes:
1. separate items and phrases,
2. show where to pause (for a moment) when speaking.

I probably was thinking about the second one when I typed that second comma.  But, Christian is right; it really isn't needed.

The first comma is correct syntax.  It serves both purposes there.  The same thing goes for the "`But,`" above here.  With the comma, it applies to the entire sentence; without the comma, it would apply to only the word "`Christian`".  Examples:
- **"I would choose anyone but him for that job!"**
- **"I could choose anyone; but, he is the best one for that job."**

PR #381, opened 2017-02-13
"static" forward declarations

### LINKED ISSUE #204: using "static" to declare forward reference doesnt work
this basically describes what should happen: http://infocenter.arm.com/help/topic/com.arm.doc.faqs/ka9772.html

in my code i have something like this:

``` C
 typedef struct _DIRMENU
 {
     char *name;
     struct _DIRMENU *dest;
 } DIRMENU; 

 extern DIRMENU rmenu; /* <-- this currently "works" but gives a warning, GCC gives an error */
 /* static DIRMENU rmenu; */ /* <-- this SHOULD work, but gives an error in cc65 */

 static DIRMENU lmenu = {
     "left",
     &rmenu
 };

 static DIRMENU rmenu = {
     "right",
     &lmenu
 };
```

pfusik:
The middle commit fixes #204. The sample code from this ticket now compiles as expected.

While working on it, I noticed and fixed two other problems (in the other commits):

* The compiler silently upgraded a declaration with an initializer to a definition, e.g. in `extern int i = 42;` the `extern` was ignored. I made it a syntax error.
* Duplicate global variables were only signaled by the assembler. I made them a compiler error.

### DIFF
--- a/src/cc65/compile.c
+++ b/src/cc65/compile.c
@@ -154,7 +154,7 @@ static void Parse (void)
                  (Decl.StorageClass & (SC_EXTERN|SC_STATIC)) == SC_STATIC)) {
 
                 /* We will allocate storage */
-                Decl.StorageClass |= SC_STORAGE | SC_DEF;
+                Decl.StorageClass |= SC_STORAGE;
             }
 
             /* If this is a function declarator that is not followed by a comma
@@ -186,6 +186,9 @@ static void Parse (void)
 
                 /* Allow initialization */
                 if (CurTok.Tok == TOK_ASSIGN) {
+
+                    /* This is a definition */
+                    Entry->Flags |= SC_DEF;
 
                     /* We cannot initialize types of unknown size, or
                     ** void types in ISO modes.
@@ -234,18 +237,14 @@ static void Parse (void)
                         Entry->Flags &= ~(SC_STORAGE | SC_DEF);
                     }
 
-                    /* Allocate storage if it is still needed */
-                    if (Entry->Flags & SC_STORAGE) {
-
-                        /* Switch to the BSS segment */
-                        g_usebss ();
-
-                        /* Define a label */
-                        g_defgloblabel (Entry->Name);
-
-                        /* Allocate space for uninitialized variable */
-                        g_res (Size);
-                    }
+                    /* A global (including static) uninitialized variable
+                    ** is only a tentative definition. For example, this is valid:
+                    ** int i;
+                    ** int i;
+                    ** static int j;
+                    ** static int j = 42;
+                    ** Code for these will be generated in FinishCompile.
+                    */
                 }
 
             }
@@ -300,7 +299,7 @@ void Compile (const char* FileName)
     struct tm*  TM;
 
     /* Since strftime is locale dependent, we need the abbreviated month names
-    ** in english.
+    ** in English.
     */
     static const char MonthNames[12][4] = {
         "Jan", "Feb", "Mar", "Apr", "May", "Jun",
@@ -397,20 +396,26 @@ void Compile (const char* FileName)
 void FinishCompile (void)
 /* Emit literals, externals, debug info, do cleanup and optimizations */
 {
-    SymTable* SymTab;
-    SymEntry* Func;
+    SymEntry* Entry;
 
-    /* Walk over all functions, doing cleanup, optimizations ... */
-    SymTab = GetGlobalSymTab ();
-    Func   = SymTab->SymHead;
-    while (Func) {
-        if (SymIsOutputFunc (Func)) {
+    /* Walk over all global symbols:
+    ** - for functions do cleanup, optimizations ...
+    ** - generate code for uninitialized global variables
+    */
+    for (Entry = GetGlobalSymTab ()->SymHead; Entry; Entry = Entry->NextSym) {
+        if (SymIsOutputFunc (Entry)) {
             /* Function which is defined and referenced or extern */
-            MoveLiteralPool (Func->V.F.LitPool);
-            CS_MergeLabels (Func->V.F.Seg->Code);
-            RunOpt (Func->V.F.Seg->Code);
+            MoveLiteralPool (Entry->V.F.LitPool);
+            CS_MergeLabels (Entry->V.F.Seg->Code);
+            RunOpt (Entry->V.F.Seg->Code);
+        } else if ((Entry->Flags & (SC_STORAGE | SC_DEF | SC_STATIC)) == (SC_STORAGE | SC_STATIC)) {
+            /* Tentative definition of uninitialized global variable */
+            g_usebss ();
+            g_defgloblabel (Entry->Name);
+            g_res (SizeOf (Entry->Type));
+            /* Mark as defined, so that it will be exported not imported */
+            Entry->Flags |= SC_DEF;
         }
-        Func = Func->NextSym;
     }
 
     /* Output the literal pool */
--- a/src/cc65/compile.c
+++ b/src/cc65/compile.c
@@ -144,14 +144,17 @@ static void Parse (void)
             **
             **   - if it is not a typedef or function,
             **   - if we don't had a storage class given ("int i")
-            **     or the storage class is explicitly specified as static.
+            **   - if the storage class is explicitly specified as static,
+            **   - or if there is an initialization.
             **
             ** This means that "extern int i;" will not get storage allocated.
             */
             if ((Decl.StorageClass & SC_FUNC) != SC_FUNC          &&
                 (Decl.StorageClass & SC_TYPEMASK) != SC_TYPEDEF    &&
                 ((Spec.Flags & DS_DEF_STORAGE) != 0         ||
-                 (Decl.StorageClass & (SC_EXTERN|SC_STATIC)) == SC_STATIC)) {
+                 (Decl.StorageClass & (SC_EXTERN|SC_STATIC)) == SC_STATIC ||
+                 ((Decl.StorageClass & SC_EXTERN) != 0 &&
+                  CurTok.Tok == TOK_ASSIGN))) {
 
                 /* We will allocate storage */
                 Decl.StorageClass |= SC_STORAGE;
--- a/src/cc65/compile.c
+++ b/src/cc65/compile.c
@@ -144,17 +144,14 @@ static void Parse (void)
             **
             **   - if it is not a typedef or function,
             **   - if we don't had a storage class given ("int i")
-            **   - if the storage class is explicitly specified as static,
-            **   - or if there is an initialization.
+            **     or the storage class is explicitly specified as static.
             **
             ** This means that "extern int i;" will not get storage allocated.
             */
             if ((Decl.StorageClass & SC_FUNC) != SC_FUNC          &&
                 (Decl.StorageClass & SC_TYPEMASK) != SC_TYPEDEF    &&
                 ((Spec.Flags & DS_DEF_STORAGE) != 0         ||
-                 (Decl.StorageClass & (SC_EXTERN|SC_STATIC)) == SC_STATIC ||
-                 ((Decl.StorageClass & SC_EXTERN) != 0 &&
-                  CurTok.Tok == TOK_ASSIGN))) {
+                 (Decl.StorageClass & (SC_EXTERN|SC_STATIC)) == SC_STATIC)) {
 
                 /* We will allocate storage */
                 Decl.StorageClass |= SC_STORAGE | SC_DEF;
--- a/src/cc65/compile.c
+++ b/src/cc65/compile.c
@@ -188,6 +188,10 @@ static void Parse (void)
                 if (CurTok.Tok == TOK_ASSIGN) {
 
                     /* This is a definition */
+                    if (SymIsDef (Entry)) {
+                        Error ("Global variable `%s' has already been defined",
+                               Entry->Name);
+                    }
                     Entry->Flags |= SC_DEF;
 
                     /* We cannot initialize types of unknown size, or
--- a/src/cc65/symtab.c
+++ b/src/cc65/symtab.c
@@ -821,7 +821,7 @@ SymEntry* AddGlobalSym (const char* Name, const Type* T, unsigned Flags)
         }
 
         /* An extern declaration must not change the current linkage. */
-        if (IsFunc || (Flags & (SC_EXTERN | SC_DEF)) == SC_EXTERN) {
+        if (IsFunc || (Flags & (SC_EXTERN | SC_STORAGE)) == SC_EXTERN) {
             Flags &= ~SC_EXTERN;
         }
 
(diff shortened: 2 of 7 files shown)


src/cc65/compile.c
```diff
@@ -300,7 +299,7 @@ void Compile (const char* FileName)
     struct tm*  TM;
 
     /* Since strftime is locale dependent, we need the abbreviated month names
-    ** in english.
+    ** in English.
```
greg-king5: The name of a country and the name of the people who live there are "capitated" (have an initial capital letter).  But, the name of their language isn't capitated.  Example:

**The French (who live in France) speak french.**

pfusik: Do you mean that the [Cambridge Dictionary](http://dictionary.cambridge.org/dictionary/english/english?q=English) and [Wikipedia](https://en.wikipedia.org/wiki/English_language) are wrong?

oliverschmidt:
Hi,

First of all thanks for working on bugs :-)

Unfortunately a regression test fails:
```
make[1]: *** [../../testwrk/yaccdbg.prg] Error 1
make[1]: Leaving directory `/home/travis/build/cc65/cc65/test/ref'
```
See https://github.com/cc65/cc65/tree/master/test/ref for the source & Makefile.

Apart from that I'm wondering if I could motivate you to add three regression tests to https://github.com/cc65/cc65/tree/master/test/val each failing without your change and succeeding with it.

Regards,
Oliver

pfusik:
Hi Oliver,

Yes, I'll check why `yaccdbg` failed and add the tests.

Piotr

oliverschmidt:
> I'll check why yaccdbg failed and add the tests.

:-)

mrdudz:
thanks for having a look at this... its pissing me off for a while now :)

it might by a good idea to put your additional fixes into a seperate PR, so incase something breaks it can be checked/debugged seperately

src/cc65/compile.c
greg-king5: So ... according to Wikipedia, all of my english teachers -- oops, English teachers -- were wrong.  OK.

greg-king5:
```C
extern int i = 42;
```
I don't have access to a C90 document; but, [this C99 document](http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf) says that it is allowed.  It even gives an example of an "external definition", on document page 144 (PDF page 156).

pfusik:
Greg, you are right. gcc only gives a warning for this, which I somehow overlooked as an error.

I'll revert this one then.

oliverschmidt:
> I'll revert this one then.

Just to make sure we're on the same page: We're waiting for Piotr to remove some part of this pull request - correct ?

pfusik:
Oliver,

That is the easy part.
Additionally a failed test needs to be checked and new tests added. I want to do that, I'm just short of time.

oliverschmidt:
Thanks for the prompt intermediate feedback :-) We're certainly not in a hurry. I just wanted to make sure we have no communication deadlock here.

greg-king5:
> Unfortunately, a regression test fails:
> ```
> make[1]: *** [../../testwrk/yaccdbg.prg] Error 1
> make[1]: Leaving directory `/home/travis/build/cc65/cc65/test/ref'
> ```

Don't bother to look at that test; it isn't a regression.  This PR causes two arrays to be moved (defined earlier) in the memory map.  So, `yytext[]` no longer hides a bug in `"test/ref/yacc.c"`.  I haven't found the bug; but, I do have this work-around:
```diff
diff --git a/test/ref/yacc.c b/test/ref/yacc.c
--- a/test/ref/yacc.c
+++ b/test/ref/yacc.c
@@ -333,6 +333,10 @@ int yylineno =1;
 # define YYU(x) x
 # define NLSTATE yyprevious=YYNEWLINE
 char yytext[YYLMAX];
+
+/* a work-around for an array underflow bug (apparently on yylstate[]) */
+int dummy1;
+
 struct yysvf *yylstate [YYLMAX], **yylsp, **yyolsp;
 char yysbuf[YYLMAX];
 char *yysptr = yysbuf;
```
It stops `"yaccdbg.c"` from interfering with this PR.

---
There is a regression at `"test/err/static-4.c"`.  Your patch should not allow it to compile.  See PR #197 for details.

oliverschmidt:
@greg-king5: Thanks for your analysis !

pfusik:
Pushed the promised revert and a fix for the `static-4.c` regression.
Now that I locally removed the `yaccdbg` test, `limits.c` fails because of LF/CRLF mismatch (I'm on Windows, `limits.ref` is LF, `sim65` outputs CRLF)

groessler:
Hmm, limits.ref should be checked out with your Windows git client with CR/LF.

Do the other source files (*.c, *.h, etc.) have LF or CR/LF in your sandbox?

pfusik:
@groessler During git installation I chose an option to checkout as-is. I don't like the idea that the contents of files is different depending on which OS I'm using.

groessler:
> I don't like the idea that the contents of files is different depending on which OS I'm using.

But text files have to be different. Depending on your editor (if it adds CRLF to new files) the repository might end up with files with mixed line endings.

mrdudz:
On Thursday 09 March 2017, 12:12:04 Christian Groessler 
<notifications@github.com> wrote:
> > I don't like the idea that the contents of files is different depending on
> > which OS I'm using.
> But text files have to be different. Depending on your editor (if it adds
> CRLF to new files) the repository might end up with files with mixed line
> endings.

sounds like a broken editor to me :)

-- 

You think you know when you learn, are more sure when you can write, even more 
when you can teach, but certain when you can program. 
<Alan J. Perlis>

groessler:
Try notepad, it won't even display files with just LF ending correctly :-)

pfusik:
In the middle of an editor flame war, I added the regression tests.

The diagnostics are now sub-optimal for the mixed global/static definitions, e.g.

    duplicate-global-static.c(13): Warning: static declaration follows non-static declaration of `n'.
    duplicate-global-static.c(13): Error: Global variable `n' has already been defined
    duplicate-static-global.c(13): Warning: public declaration follows static declaration of `n'.
    duplicate-static-global.c(13): Error: Global variable `n' has already been defined

So the error is preceded by a warning. It could be fixed.

C still surprises me after 20 years. If this is correct C:

    static int n = 0;
    extern int n;           /* should not give an error */

is `n` externally visible or not?

groessler:
No editor war.

But on Windows the sources should be checked out with CRLF endings.

pfusik:
What do you mean by "should be" ? I've been professionally developing for the last fifteen years mostly on Windows and I use LF. Notepad is **not** a developer's tool. My editor auto-detects LF vs CRLF, so no need to worry about mixed line endings.

oliverschmidt:
> No editor war.

Thanks !

> But on Windows the sources should be checked out with CRLF endings.

One can easily be religious about it. But it just doesn't help to be so as we see here. sim6502 is built with MinGW default options and ends up - as expected - to create CR/LF on Windows (which is by the way necessary i.e. to be used interactively in a CMD.EXE shell). So in order for the tests to work the reference file needs to have CR/LF on Windows too. It's that simple. The Git guys were pretty surely not keen to create those special options but they are plainly necessary. No religion about it. Or do I miss some point?

oliverschmidt:
> What do you mean by "should be" ?

I'd interpret it that way: Sources should be checked out with CR/LF to have things work smoothly - like the cc65 regression tests.

You are of course free to do it otherwise - but then there's no point in complaining here about the regression tests not working. Simple - isn't it?

In example I personally - of course - always use the Git as-is option on Windows. But I don't bother others with the issues arising from doing so - because I'm a professional software developer and understand the pros and cons of my decisions ;-)))

pfusik:
Now it's clear. Thanks for explanation.

`yaccdbg` is now the only test that fails. It's because the compiler outputs global variables in a different order and the test compares pointers to them.

groessler:
I intended to let the CRLF discussion rest here.

But then I checked my cc65 sandbox, and, behold, there are some files checked in with CRLF line endings.

Specifically:
`

./include/osic1p.h
./include/zlib.h
./test/ref/divmod.c
./test/ref/switch2.c
./test/ref/varargs.c
./test/ref/init.c
./test/ref/pointer2.c
./asminc/smc.inc
./asminc/opcodes.inc
./libsrc/osic1p/cgetc.s
./libsrc/osic1p/osic1p.inc
./libsrc/osic1p/osiscreen.inc
./libsrc/c64/emd/c64-65816.s
./testcode/lib/mul-test.c
`

May I ask for the reason you are checking out "as is", as opposed with the correct system line ending?

Even if the editor auto-detects existing line endings, it might use CRLF for new files.

So Windows guys, please, checkout with correct line ending.

oliverschmidt:
> there are some files checked in with CRLF line endings

:-(

> May I ask for the reason you are checking out "as is", as opposed with the correct system line ending?

I personally noticed that sooner or later I end up in most projects to move files back and forth between Linux and Windows. Either by sending them via email or using a shared SMB drive or <...>.

According to my experience Windows is better in dealing with LF files than Linux in dealing with CR/LF files. So I use LF everywhere.

oliverschmidt:
> So Windows guys, please, checkout with correct line ending.

Git allows to check out files leaving them as LF but convert CR/LF to LF on check in (see ``core.autocrlf input``). So it's easy to have the best of both worlds if you (like me) prefer local LF files on Windows.

oliverschmidt:
> May I ask for the reason you are checking out "as is", as opposed with the correct system line ending?

Another reason is coming into my mind: Some projects come with shell scripts intended to be used on Windows too - and depending in the Windows port of the shell to be used the shell insists on LF shell scripts. One could surely fix that with file-extension-specific Git setting for that project but just using LF for all (text) files is path of least resistance.

groessler:
This appears to be a good setting. Now, how can we tell people to use this setting? I don't think we have a kind of "contributing" documentation, do we?

oliverschmidt:
> This appears to be a good setting. 

Yep. Therefore it's recommended in many Git tutorials. It's very purpose is to keep CR/LF out of the repo. I.e. http://adaptivepatchwork.com/2012/03/01/mind-the-end-of-your-line/ says

"This setting is generally used on Unix/Linux/OS X to prevent CRLFs from getting written into the repository. The idea being that if you pasted code from a web browser and accidentally got CRLFs into one of your files, Git would make sure they were replaced with LFs when you wrote to the object database."

> Now, how can we tell people to use this setting? I don't think we have a kind of "contributing" documentation, do we?

No, we don't. Maybe you want to create one. Or maybe you want to write "at least" a posting on the dev list. At least it was you who noticed the issue and contributed the fix.

mrdudz:
On Friday 10 March 2017, 02:49:21 Oliver Schmidt <notifications@github.com> 
wrote:
> > This appears to be a good setting.
> 
> Yep. Therefore it's recommended in many Git tutorials. It's very purpose is
> to keep CR/LF out of the repo. I.e.
> http://adaptivepatchwork.com/2012/03/01/mind-the-end-of-your-line/ says
> 
> "This setting is generally used on Unix/Linux/OS X to prevent CRLFs from
> getting written into the repository. The idea being that if you pasted code
> from a web browser and accidentally got CRLFs into one of your files, Git
> would make sure they were replaced with LFs when you wrote to the object
> database."
> > Now, how can we tell people to use this setting? I don't think we have a
> > kind of "contributing" documentation, do we?
> No, we don't. Maybe you want to create one. Or maybe you want to write "at
> least" a posting on the dev list. At least it was you who noticed the issue
> and contributed the fix.

isnt there some repo global setting for this in git? thats how its done in the 
VICE repo (on svn) at least...

-- 

The Christian religion not only was at first attended with miracles, but even 
at this day cannot be believed by any reasonable person without one. 
<David Hume>

silverdr:
There is the `.gitattributes` approach. Setting this up is probably easier and more effective than writing documentation how should everyone configure his git installation.

greg-king5:
> C still surprises me after 20 years. If this is correct C:
> ```C
> static int n = 0;
> extern int n;           /* should not give an error */
> ```
> is `n` externally visible or not?

The bottom of PDF-page 42 of [this document](http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf) says:
> 6.2.2 Linkages of identifiers
> **...**
>
> 3. If the declaration of a file scope identifier for an object or a function contains the storage-class specifier **`static`**, the identifier has internal linkage.
>
> 4. For an identifier declared with the storage-class specifier **`extern`** in a scope in which a prior declaration of that identifier is visible, if the prior declaration specifies internal or external linkage, the linkage of the identifier at the later declaration is the same as the linkage specified at the prior declaration.  If no prior declaration is visible, or if the prior declaration specifies no linkage, then the identifier has external linkage.

pfusik:
> there are some files checked in with CRLF line endings
> :-(
It's not causing any problems. And it's easy to cleanup with `dos2unix`.
> 
> May I ask for the reason you are checking out "as is", as opposed with the correct system line ending?
> I personally noticed that sooner or later I end up in most projects to move files back and forth between Linux and Windows. Either by sending them via email or using a shared SMB drive or <...>.
> 
> According to my experience Windows is better in dealing with LF files than Linux in dealing with CR/LF files. So I use LF everywhere.

That is also my reason and experience. Plus, LF takes less space. :)
And my editor is configured to LF for new files.

pfusik:
@greg-king5 Thanks. I see that cc65 does it right.

I think it's good to have tests for all variable definition/declaration combinations: `{,extern,static} int x [= 0]` that's 6 possibilies, meaning 36 tests. And to test the resulting linkage.

oliverschmidt:
> There is the .gitattributes approach.

First I was afraid that per-repo Git settings dictate the user if his local files are LF vs. CR/LF - which we might not want. However now I (believe to have) learned that it's possible to only mandate LF conversion on comit. So I'll add a `.gitattributes`.

Thanks for your input.

silverdr:
Exactly. You can set it this way. The important thing (which I had to learn the hard way some time ago) is to manually set files/patterns you want to include. Otherwise surprises may be expected, especially after some longer time of intensive repo use ;-)

oliverschmidt:
> The important thing (which I had to learn the hard way some time ago) is to manually set files/patterns you want to include.

While I in general see your point I don't see much risk for the cc65 repo. There's as far as I see only one non-text-file in the repo.

Therefore I see me doing the classic `* text=auto`.

greg-king5:
Travis-CI is green.  Do you plan to add any more tests to this pull request?

oliverschmidt:
> So I'll add a .gitattributes.

Done: https://github.com/cc65/cc65/commit/2e8d792f9a1b55bb19942e02b91b1294dbc37e87

oliverschmidt:
> Do you plan to add any more tests to this pull request?

In case there are more tests to come (which is always welcome) they can go into a new pull request...

pfusik:
Yes, I think I'll include the extra tests that I mentioned in a new pull request fixing the duplicate warning/error messages.

PR #382, opened 2017-02-13
Add testcode/lib/mouse-test.c. Rename samples/mousetest.c to mousedemo.c

groessler:
mouse-test.c can be use to test all available mouse drivers for a given target.
Currently supported targets are Atari, C64, and C128.
To avoid confusion, rename samples/mousetest.c to mousedemo.c.

This was the test program I was using back in 2014 when I added some mouse drivers for Atari and found the "phantom key presses" problem on C128.

I think it makes sense to have it as test program in the cc65 code base.

### DIFF
--- a/testcode/lib/mouse-test.c
+++ b/testcode/lib/mouse-test.c
@@ -1,10 +1,13 @@
 /*
-** Test/demo program for mouse usage.
-** Will work for the C64/C128/CBM510/Atari/Apple2.
+** Test program for mouse drivers.
+** Supportsthe C64/C128/CBM510/Atari/Apple2.
 **
 ** 2001-09-13, Ullrich von Bassewitz
 ** 2013-09-05, Greg King
 **
+** Compile with "-DSTATIC_MOUSE" to statically link all available drivers.
+** Compile with "-DMOUSE_DRIVER=<driver_sym>" to statically link the given driver.
+**     E.g., -DMOUSE_DRIVER=atrsts_mou to just link with the Atari ST mouse driver.
 */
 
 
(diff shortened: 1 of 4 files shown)


oliverschmidt:
Thanks :-)

Are you aware of https://github.com/cc65/cc65/commit/88c102b993d64fa9b74af6c1a74f992480d1d9a3 ?

groessler:
On 02/13/17 23:02, Oliver Schmidt wrote:

> Thanks :-)
>
> Are you aware of 88c102b 
> <https://github.com/cc65/cc65/commit/88c102b993d64fa9b74af6c1a74f992480d1d9a3> 
> ?
>


No. I wasn't aware of that.

Sorry, Greg, but citing part of your message from the link

---- begin

I want to put a |samples/mousedemo.c| file in its place that does 
something outside of the mouse library, but uses that library for input. 
Some things that spring to my mind are:

  * A child education program that shows simple shapes; and, asks the
    kid to point to the one that is named.
  * A brain game that shows, then hides, words and/or pictures; then,
    asks the player to "click" on two places that match.
  * A simple calculator. It draws little "buttons" on the screen. We
    "click" them, instead of tapping keys, to input numbers and operations.
  * (Or maybe, a mousable version of the menu program that's in the
    contributions directory.)

---- end

I think that is out of scope for a simple mouse demo program. I think 
the existing program shows how to use the mouse sufficiently enought, 
and I think that's the purpose of the files in samples/. Show how to use 
the API.

If you want make a nicer mouse demo, it's ok. But I don't know 
(depending how complex it is) if it should be part of "cc65" or more of 
an external "contrib" item.

regards,
chris

groessler:
I can also rename mousedemo.c back to mousetest.c. I ave no strong opinion on that, just thought it would prevent confusion. Or I can delete it altogether with the new testcode/lib/mouse-test.c.

oliverschmidt:
As nobody complained so far I'm going to merge this pull request as-is now.

groessler:
Thanks :-)

testcode/lib/mouse-test.c
```diff
@@ -0,0 +1,431 @@
...
+    bool Invisible = true, Done = false, Jailed = false;
+
+#ifdef __ATARIXL__
+    cprintf ("adding heap: $%04X bytes at $%04X\r\n",
+             &_HIDDEN_RAM_SIZE__ - (&_HIDDEN_RAM_LAST__ - &_HIDDEN_RAM_START__),
+             &_HIDDEN_RAM_LAST__);
+
+    _heapadd (&_HIDDEN_RAM_LAST__, (size_t)(&_HIDDEN_RAM_SIZE__ - (&_HIDDEN_RAM_LAST__ - &_HIDDEN_RAM_START__)));
+    cgetc ();
+#endif
```
greg-king5: Is that code needed when the static drivers are tested?

groessler: Probably not. Don't know if there are any allocations under the hood during the rest of the program flow.

PR #386, opened 2017-02-23
Improved display list instruction definition for more comfortable use…

IrgendwerA8:
… within void data definition.

### DIFF
--- a/include/_antic.h
+++ b/include/_antic.h
@@ -57,42 +57,42 @@ struct __antic {
 /* antic instruction set */
 
 /* absolute instructions (non mode lines) */
-#define DL_JMP  1
-#define DL_JVB  65
+#define DL_JMP  (unsigned char) 1
+#define DL_JVB  (unsigned char) 65
 
-#define DL_BLK1 0
-#define DL_BLK2 16
-#define DL_BLK3 32
-#define DL_BLK4 48
-#define DL_BLK5 64
-#define DL_BLK6 80
-#define DL_BLK7 96
-#define DL_BLK8 112
+#define DL_BLK1 (unsigned char) 0
+#define DL_BLK2 (unsigned char) 16
+#define DL_BLK3 (unsigned char) 32
+#define DL_BLK4 (unsigned char) 48
+#define DL_BLK5 (unsigned char) 64
+#define DL_BLK6 (unsigned char) 80
+#define DL_BLK7 (unsigned char) 96
+#define DL_BLK8 (unsigned char) 112
 
 /* absolute instructions (mode lines) */
-#define DL_CHR40x8x1    2       /* monochrome, 40 character & 8 scanlines per mode line (GR. 0) */
-#define DL_CHR40x10x1   3       /* monochrome, 40 character & 10 scanlines per mode line */
-#define DL_CHR40x8x4    4       /* colour, 40 character & 8 scanlines per mode line (GR. 12) */
-#define DL_CHR40x16x4   5       /* colour, 40 character & 16 scanlines per mode line (GR. 13) */
-#define DL_CHR20x8x2    6       /* colour (duochrome per character), 20 character & 8 scanlines per mode line (GR. 1) */
-#define DL_CHR20x16x2   7       /* colour (duochrome per character), 20 character & 16 scanlines per mode line (GR. 2) */
+#define DL_CHR40x8x1    (unsigned char) 2       /* monochrome, 40 character & 8 scanlines per mode line (GR. 0) */
+#define DL_CHR40x10x1   (unsigned char) 3       /* monochrome, 40 character & 10 scanlines per mode line */
+#define DL_CHR40x8x4    (unsigned char) 4       /* colour, 40 character & 8 scanlines per mode line (GR. 12) */
+#define DL_CHR40x16x4   (unsigned char) 5       /* colour, 40 character & 16 scanlines per mode line (GR. 13) */
+#define DL_CHR20x8x2    (unsigned char) 6       /* colour (duochrome per character), 20 character & 8 scanlines per mode line (GR. 1) */
+#define DL_CHR20x16x2   (unsigned char) 7       /* colour (duochrome per character), 20 character & 16 scanlines per mode line (GR. 2) */
 
-#define DL_MAP40x8x4    8       /* colour, 40 pixel & 8 scanlines per mode line (GR. 3) */
-#define DL_MAP80x4x2    9       /* 'duochrome', 80 pixel & 4 scanlines per mode line (GR.4) */
-#define DL_MAP80x4x4    10      /* colour, 80 pixel & 4 scanlines per mode line (GR.5) */
-#define DL_MAP160x2x2   11      /* 'duochrome', 160 pixel & 2 scanlines per mode line (GR.6) */
-#define DL_MAP160x1x2   12      /* 'duochrome', 160 pixel & 1 scanline per mode line (GR.14) */
-#define DL_MAP160x2x4   13      /* 4 colours, 160 pixel & 2 scanlines per mode line (GR.7) */
-#define DL_MAP160x1x4   14      /* 4 colours, 160 pixel & 1 scanline per mode line (GR.15) */
-#define DL_MAP320x1x1   15      /* monochrome, 320 pixel & 1 scanline per mode line (GR.8) */
+#define DL_MAP40x8x4    (unsigned char) 8       /* colour, 40 pixel & 8 scanlines per mode line (GR. 3) */
+#define DL_MAP80x4x2    (unsigned char) 9       /* 'duochrome', 80 pixel & 4 scanlines per mode line (GR.4) */
+#define DL_MAP80x4x4    (unsigned char) 10      /* colour, 80 pixel & 4 scanlines per mode line (GR.5) */
+#define DL_MAP160x2x2   (unsigned char) 11      /* 'duochrome', 160 pixel & 2 scanlines per mode line (GR.6) */
+#define DL_MAP160x1x2   (unsigned char) 12      /* 'duochrome', 160 pixel & 1 scanline per mode line (GR.14) */
+#define DL_MAP160x2x4   (unsigned char) 13      /* 4 colours, 160 pixel & 2 scanlines per mode line (GR.7) */
+#define DL_MAP160x1x4   (unsigned char) 14      /* 4 colours, 160 pixel & 1 scanline per mode line (GR.15) */
+#define DL_MAP320x1x1   (unsigned char) 15      /* monochrome, 320 pixel & 1 scanline per mode line (GR.8) */
 
 /* modifiers on mode lines */
-#define DL_HSCROL       16
-#define DL_VSCROL       32
-#define DL_LMS          64
+#define DL_HSCROL(x)    (unsigned char)((x) | 16)
+#define DL_VSCROL(x)    (unsigned char)((x) | 32)
+#define DL_LMS(x)       (unsigned char)((x) | 64)
 
 /* general modifier */
-#define DL_DLI          128
+#define DL_DLI(x)       (unsigned char)((x) | 128)
 
 /* End of _antic.h */
 #endif /* #ifndef __ANTIC_H */
--- a/testcode/lib/atari/displaylist.c
+++ b/testcode/lib/atari/displaylist.c
@@ -51,7 +51,7 @@ main(void)
 {
     // unfortunately "sizeof()" doesn't work with void data
     // (Error: Size of data type is unknown)
-    // so we trick with the adresses at front and behind...
+    // so we trick with the addresses at front and end...
 
     int returnValue = (((unsigned int)&dlend-(unsigned int)&DisplayList) != 28);  // assure only one byte per instruction!
 
--- a/testcode/lib/atari/displaylist.c
+++ b/testcode/lib/atari/displaylist.c
@@ -1,0 +1,69 @@
+/*
+** testprogram for ANTIC instructions as defined in "_antic.h"
+**
+** 23-Feb-2017, Christian Krueger
+*/
+
+#include <conio.h>
+#include <atari.h>
+#include <peekpoke.h>
+#include <string.h>
+
+// code is only for testing purposes, as screen and display list are not aligned
+// and jumps not set!
+
+unsigned char DummyScreen[400];
+
+void DisplayList = {
+	DL_BLK1,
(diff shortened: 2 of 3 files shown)


testcode/lib/atari/displaylist.c
```diff
@@ -0,0 +1,69 @@
...
+#include <atari.h>
+#include <peekpoke.h>
+#include <string.h>
+
+// code is only for testing purposes, as screen and display list are not aligned
+// and jumps not set!
+
+unsigned char DummyScreen[400];
+
+void DisplayList = {
```
groessler: What is this? Is this C?

What's wrong with
`unsigned char DisplayList[] = { ... };` ?

IrgendwerA8: This is a cc65 extension, see bullet no. 6: http://cc65.github.io/doc/cc65.html
You cannot define this as unsigned char array because 16 bit addresses are embedded in a display list too ( see "DummyScreen").
The (useful) void-initialization was the reason to change the DL-defines to explicitly "unsigned char" otherwise they would occupy 16 bit...

oliverschmidt:
@groessler: Is this clarified from your point of view?

testcode/lib/atari/displaylist.c
groessler: I see. I forgot about the 16bit address inside the DL.

include/_antic.h
```diff
@@ -57,42 +57,42 @@ struct __antic {
...
+#define DL_MAP160x1x2   (unsigned char) 12      /* 'duochrome', 160 pixel & 1 scanline per mode line (GR.14) */
+#define DL_MAP160x2x4   (unsigned char) 13      /* 4 colours, 160 pixel & 2 scanlines per mode line (GR.7) */
+#define DL_MAP160x1x4   (unsigned char) 14      /* 4 colours, 160 pixel & 1 scanline per mode line (GR.15) */
+#define DL_MAP320x1x1   (unsigned char) 15      /* monochrome, 320 pixel & 1 scanline per mode line (GR.8) */
 
 /* modifiers on mode lines */
-#define DL_HSCROL       16
-#define DL_VSCROL       32
-#define DL_LMS          64
+#define DL_HSCROL(x)    (unsigned char)((x) | 16)
```
IrgendwerA8: Of course I could, but I would like to have this as a breaking change:
- the names are good as they are and changing them leads IMHO to more confusion and does not really help (DL_HSCROL2 ?)
- the new definition is IMHO better than the old since multiple use doesn't do any harm
(DL_HSCROL(DL_HSCROL(...)) != DL_HSCROL + DL_HSCROL
- it is the first time that the usage is part of the documentation
- old usage is reported by the compiler with an error and therefore easy detectable

groessler: > I would like to have this as a breaking change

I would like to have both methods be able to coexist. For a static DL the "void" approach is simpler, but if the DL is to be changed at runtine, the "unsigned char array" method is simpler to use.

> the names are good as they are and changing them leads IMHO to more confusion and does not really help (DL_HSCROL2 ?)

maybe DL_HSCROLx or DH_HSCROLX?

> (DL_HSCROL(DL_HSCROL(...)) != DL_HSCROL + DL_HSCROL

(DL_HSCROL(DL_HSCROL(...)) == DL_HSCROL | DL_HSCROL

oliverschmidt: > but if the DL is to be changed at runtine, the "unsigned char array" method is simpler to use.

Could you please elaborate?

groessler: Let's say, you want to change the 3rd entry of the DL (typically the last of the first three entries which each just create eight empty screen lines) to now also give a DL-interrupt, you can simply write
DisplayList[2] |= DL_DLI;

How would this be done in the "void" approach?

IrgendwerA8: DisplayList[2] |= DL_DLI(0)

I really would appreciate a single definition as two concurrent ones can be confusing, mixed and misleading. I see no real benefit in having different ones.

>(DL_HSCROL(DL_HSCROL(...)) == DL_HSCROL | DL_HSCROL
Yes, but thanks to the new definition you aren't in danger to use a "wrong operator"...

groessler: > I really would appreciate a single definition as two concurrent ones can be confusing, mixed and misleading. I see no real benefit in having different ones.

Backwards compatibility.

> but thanks to the new definition you aren't in danger to use a "wrong operator"...

When writing C you're always in danger of using a wrong operator.

> What about a macro for that use case?

That would be nice.

> DisplayList[2] |= DL_DLI(0)

So, it boils down to having the same name or a different name for the new DL_DLI(x) macro with the existing DL_DLI macro (and of course the other modifiers), correct?

My POV is that if there is no real good reason to break backward compatibilty, don't do it. I don't think this case is a real good reason. As you stated before, previously there was no documentation for the Antic modes, so it could be documented that there is DL_DLI for the constant, and (say) DL_DLIx() for the modifier macro.
cc65 already seems to be difficult to newbies, so they tend to shy away, and if there are complaints that something that worked before suddenly doesn't work anymoore with the newest version will raise the acceptance level even more.

IrgendwerA8: Sorry that you don't appreciate the change, but here my final thoughts regarding the issue (I have no problem to cancel the pull request (it would still live in my fork) - it seems it happens to often that someone comes up with a little change request supplying a test and documentation too.. ;^) ):

Creating a display list on the C side isn't very comfortable. When using the unsigned char array approach, you aren't able to initialise the screen addresses, as all operations on addresses to extract a high- and a low byte are non-const in cc65 (e.g.  adr >> 8,  adr & 0xFF, (unsigned char)addr etc.). Only solution was to set up these addresses in C code (needs memory, takes performance).
AFAIK the only solution consists in using the void-data feature of cc65, requiring unsigned char data type for ANTIC instructions, otherwise you would go the uncomfortable route of writing:

void DisplayList = {
    (unsigned char) DL_BLK8,
    (unsigned char) DL_BLK8,
    (unsigned char) DL_BLK8,
    (unsigned char) (DL_CHR20x8x2 | DL_LMS),
    screen,
    (unsigned char) DL_CHR20x8x2,
    (unsigned char) DL_CHR20x8x2,
    (unsigned char) DL_CHR20x8x2,
    (unsigned char) DL_BLK4,
    (unsigned char) DL_CHR20x8x2,
    (unsigned char) DL_JVB
};

Not be able to initialise the addresses in a C display list forced at least me to do so on the assembler side. Only for that purpose I've often created an assembler file, which seems much more unfriendly for newbies to me than offering a neat solution for C.

Due to the documentation level, the reason I mentioned before and overall usage of cc65 on the Atari I doubt that the little change on modifier level implies many source files to break. I think ease the use (which this change IMHO does) is quite accepted.
BTW: Nobody seemed to care that a change once broke my SMC macros for ca65 which is even part of the suite: https://sourceforge.net/p/cc65/mailman/message/35225776/
I regard that anyway as "OK", since I also think that this is a rarely used feature.

Since this discussion is likely to exceed the creation time for the change (and of course my spare time I willing to spend to contribute), I have unfortunately have to say take it or leave it.

groessler: > Sorry that you don't appreciate the change 

I appreciate your change

> Only solution was to set up these addresses in C code (needs memory, takes performance).

That's why.

> void DisplayList = { ....

The "(unsigned char)" castings in your change are ok. So with them it would look like

void DisplayList = {
DL_BLK8,
DL_BLK8,
DL_BLK8,
(unsigned char) (DL_CHR20x8x2 | DL_LMS),
screen,
DL_CHR20x8x2,
DL_CHR20x8x2,
DL_CHR20x8x2,
DL_BLK4,
DL_CHR20x8x2,
DL_JVB
};

> Not be able to initialise the addresses in a C display list forced at least me to do so on the assembler side. Only for that purpose I've often created an assembler file, which seems much more unfriendly for newbies to me than offering a neat solution for C.

Agreed

> Due to the documentation level.... 

I also don't know how much users would be affected, but why potentially break something if there's no need (from my perspective, at least).

> ince this discussion is likely to exceed the creation time for the change (and of course my spare time I'm willing to spend to contribute), I unfortunately have to say take it or leave it.

It also takes up my spare time.

So, could you explain, why you are insisting to use the same names for the modifiers instead of using new names? In my view, when you'd use different names, the PR would be fine.

oliverschmidt:
I really want to help here - but unfortunately I just don't know enough about the ATARI to do well, but I'll try nevertheless...

- I see a subtopic about the question, if the proposed change makes it harder to modify display lists at run time. Is there an agreement about this aspect?

- I see a subtopic about the four macros DL_HSCROL, DL_VSCROL, DL_LMS and DL_DLI - right? So far they didn't have a parameter, with the proposed change they have one parameter. My question here is: If someone used them as they are now. What would be the minimal change he would have to apply to his code to make it compatible with the proposed ones?

- Are there other subtopics with disagreement I don't see?

Thanks in advance, Oliver

include/_antic.h
IrgendwerA8: >It also takes up my spare time.

Yes, I mean it not in a bad mood when I try to save our time!

Ok, really a last try: I find it much more error-prone to deliver an additional define which when it is not used breaks a correct looking display list which is hard to debug due to the 16 bit format.
Imagine a additional "DL_DLI_New()" but someone uses the old definition, thinking everything is ok:
void DisplayList = {
DL_BLK8 | DL_DLI,
...
}

You will have a hard time to locate the error in the hidden use of 16 bit values - especially true for newbies.
IMHO tweaking a display list at runtime is a more advanced operation, where I find I acceptable for the experienced user to OR with a DL_DLI(0) - I think he knows what he does.
Furthermore an advanced user is keen to write fast code, where
DisplayList[2] **=** DL_DLI(_DL_MODELINE_) would be faster anyway...

But I'm really running out of arguments now.

groessler: > Imagine a additional "DL_DLI_New()" but someone uses the old definition, thinking everything is ok:
void DisplayList = {
DL_BLK8 | DL_DLI,
...
}

A similar problem could happen if the user looks at some existing code which using the "char array" approach, and copy'n'paste something like

void DisplayList = {
DL_BLK8 | DL_DLI(0),
...
}

And, for the other argument,

> DisplayList[2] = DL_DLI(DL_MODELINE)

one can write

DisplayList[2] = DL_DLI | _DL_MODELINE_;

Also, with this change, the ASM defines (in `asminc/atari_antic.h`) are different to the C defines.

But, I'm stepping back. If you think this is the way to go, it's ok for me. But I'll direct complaints about compatibility to you, though :-)

So, if you don't want to add to this PR, please tell Oliver to merge.

IrgendwerA8: >A similar problem could happen if the user looks at some existing code which using the "char >array" approach, and copy'n'paste something like
>
>void DisplayList = {
>DL_BLK8 | DL_DLI(0),
>...
>}

And this actually emphasize my point: While it is impossible just to copy from existing code a e.g. "| DL_DLI" without getting a compiler error, I see no sense that new code is written in this way "DL_BLK8 | DL_DLI(0)" and not "DL_DLI(BLK8)" - the intrinsic safety is simply better.

>And, for the other argument,
>
>>    DisplayList[2] = DL_DLI(DL_MODELINE)
>
>one can write
>DisplayList[2] = DL_DLI | DL_MODELINE;

Yes, sure, but the point is, if it is really a _**drawback**_ that you have to write DL_DLI(DL_MODELINE) instead?

>Also, with this change, the ASM defines (in asminc/atari_antic.h) are different to the C defines.

No different meanings but a bit different to apply. Yes, I see this difference it in the same category, that you have to call a c-function __func_ inside assembler but _func_ outside.

>But, I'm stepping back. If you think this is the way to go, it's ok for me. But I'll direct complaints >about compatibility to you, though :-)

I'm fine with that. I doubt the discussion will be much longer than this one ;-). Just again for clarification: My intention was not to create a bad mood here, but just to ease the use without introduce hidden problems which are likely to happen if the new defines are applied without awareness.

And for creation of a macro to access void data byte-wise: I think this is display list independent and could make sense for all targets.

Thank you for your review and pointing out possible inconveniences.
**Oliver**, please merge my request.

groessler: I don't want to continue the discussion, just a note:

> Yes, sure, but the point is, if it is really a drawback that you have to write DL_DLI(DL_MODELINE) instead?

To me, as kind of low-level programmer, `DL_DLI | DL_DL_MODELINE` is simply more readable than `DL_DLI(DL_DL_MODELINE)`.

But to each his own....

IrgendwerA8: Absolutely understandable. I'm just willing to accept the other syntax if the alternative would look like

`(unsigned char)(DL_DLI | DL_DL_MODELINE)`

That's all.

oliverschmidt:
I still don't fully get the details but it seems that in general the proposed change is considerd an improvement and that the "only" question is if the improvement is big enough to justify an incompatible change.

I personally don't consider backward compatibility such a high value because I personally think that all users interested in "just some" cc65 that works are still on 2.13 and will stay there forever. The ones that did the move to cc65@GitHub already know that new versions come with incompatibilities here and there - just my two cents.

Having that said thanks to both of you trying to do the best for cc65 !

testcode/lib/atari/displaylist.c
```diff
@@ -0,0 +1,69 @@
...
+
+
+int
+main(void)
+{
+    // unfortunately "sizeof()" doesn't work with void data
+    // (Error: Size of data type is unknown)
+    // so we trick with the addresses at front and end...
+
+    int returnValue = (((unsigned int)&dlend-(unsigned int)&DisplayList) != 28);  // assure only one byte per instruction!
```
greg-king5: When that programming trick is used, keep in mind that the second symbol must be initialized too, and both symbols must have the same constantness qualification (i.e., if either one is `const`, then both must be `const`).  The reason is that initialized and uninitialized names, plus constant and non-constant names, are put into different address segments.  The two names won't be next to each other if they have different qualities.

IrgendwerA8: Thanks for the hint. Any chance to have sizeof() working on void data?

greg-king5: > Any chance to have sizeof() working on void data?

Yep, see pull request #401.

spiro-trikaliotis: Which makes the compiler even more non-standard. sizeof() on void data is undefined and should trigger a diagnostic. I do not like this non-conforming extension, as it supresses important diagnostics.

IrgendwerA8: `--standard c99` is your friend?

PR #387, opened 2017-02-24
atari_antic.inc: use "or" instead of "plus" in the usage example

groessler:
Noticed this after the discussion of PR #386.


oliverschmidt:
Thanks !

PR #388, opened 2017-02-25
Correcting typo, adding clrscr for conio and adding some $200 variables for conion function

jedeoric:


jedeoric:
I close it, because i have a crash with clrscr, there is a "jsr $ffff" somewhere and i don't know where it's

Issue #356, opened 2016-10-14
Wishlist: unroll inlined memset with PoT/div2/div4 sizes


clbr:
Currently, the inlined memset only writes once, creating very high loop overhead, more than half of the time is spent spinning the counter. When the size to write is a power of two, like it often is, the inlined memset should automatically unroll to 2, 4, or 8 writes per loop iteration, depending on array size.

PR #389, opened 2017-02-25
Optimization of string functions (size & speed).

IrgendwerA8:


oliverschmidt:
In the cc65 source code base we don't want to have any tab chars (0x08) - apart from Make recipes ;-)

Please remove the tab chars (and readjust formatting afterwards) - thanks.

IrgendwerA8:
Thanks for the notice. Fixed.

greg-king5:
I hope that you keep going with that optimization, `strncat()` for example.

IrgendwerA8:
>I hope that you keep going with that optimization, strncat() for example.

Yes, this and some others are still lurking on my disk (therefore my old modification date in the files).
I cannot say when I find the time to apply the changes
* all these have been created in respect to my new "regcall"-convention I presented here lately (without success) and I have to port them back 
* I like to supply them with a unit test which currently doesn't exist - to prove that they actually work correct (at least should be more convenient with the supplied macros now - if they are accepted)

Thanks for the feedback!

oliverschmidt:
I really like that test code is added. However I'm unhappy with adding the unittest header to the publich cc65 interface. I'd like to see it rather as cc65-internal tool.

IrgendwerA8:
So which location do you suggest?

oliverschmidt:
I think I'd prefer it in test/val (and a potential copy in test/ref). Or if you strongly prefer I'd agree with just test too.

IrgendwerA8:
Since these paths are not set per default the include for strcat-test.c has to look like "../../test/val/unittest.h" or I'm missing a point?

oliverschmidt:
Sorry, my fault. I misread the location of the test file :-(

Given the location of the test file I'd suggest testcode/lib or testcode.

oliverschmidt:
Thanks for your quick reaction. I don't see reasons not to merge this pull request. However I'd like to know if you are aware that the test located under 'test' are automatically executed as CI tests (i.e. https://travis-ci.org/cc65/cc65/builds/206080851). May you'd prefer to integrate your test code there?

IrgendwerA8:
I'm not familiar with the build configuration, but if library code is really build with cc65 and executed with a 6502 emulator (sim65?) on the build server here (?), why is there the "special" "testcode"-folder in a separate branch?
Are there special requirements for return codes and where can I inspect the outcome of the currently running tests?

oliverschmidt:
> I'm not familiar with the build configuration

 The overall orchestration is in https://github.com/cc65/cc65/blob/master/.travis.yml executing test at https://github.com/cc65/cc65/blob/master/.travis.yml#L9

> library code is really build with cc65 and executed with a 6502 emulator (sim65?) on the build server here (?)

That's exactly what happens (on Travis CI), see i.e. https://github.com/cc65/cc65/blob/master/test/val/Makefile

> why is there the "special" "testcode"-folder in a separate branch

At least from my personal perspective 'testcode' is a loose collection of code that someone classified as potentially useful for someone to test something at some point.

In contrast 'test' contains code that follows a rather strict formalism and is automatically testable. This automated test is done as CI test.

> Are there special requirements for return codes

I suggest putting new tests in test/val and there the test programs are required to return zero for success and non-zero for failure.

> where can I inspect the outcome of the currently running tests?

You can execute the tests locally using the Makefile. They are executed as CI test. The outcome is part of the job log linked to i.e. on pull request pages:

image

greg-king5:
> why is there the "special" "testcode"-folder in a separate branch (on the file system -- not a git branch)?

The first reason is that it's a legacy from the days before this project was moved to Github, before the `test/` directory tree was added, and before this repository was hooked to Travis-CI.

The other reason is what Oliver said.  There are fewer requirements about the code in that tree.  A program might -- or might not -- be suitable for automatic testing.  It might need to interact with a user.  It might produce output that must be inspected by a programmer.

>> Are there special requirements for return codes?
>
> I suggest putting new tests in `test/val/`; and there, the test programs are required to return zero for success and non-zero for failure.

Also, try to give the programs unique names -- within the entire `test/` tree.  The built programs are stored together in `testwrk/`.  Therefore, if two source files have the same name (but are put into different directories), then the second program won't be tested, because `make` will think that it already had been built and tested.

IrgendwerA8:
Added some more optimization. These are all I can currently submit.
BTW: Is there any kind of policy regarding speed/size balance?
I wondered that an additional byte is spend in the decspX implementations to save a single cycle in "no page crossing case".

E.g. decsp3:

`.proc   decsp3

        lda     sp
        sec
        sbc     #3
        sta     sp
        bcc     @L1
        rts

@L1:    dec     sp+1
        rts

.endproc`

Could be
`.proc   decsp3

        lda     sp
        sec
        sbc     #3
        sta     sp
        bcs     leave
	dec     sp+1
leave:  rts`

mrdudz:
i'd always vote for size over speed, unless perhaps for the core runtime stuff - because size is much more of a problem usually.

IrgendwerA8:
...and how can I fixed the: "Error: Maximum number of cycles reached."-error in Travis build running the lib_common_strcat-test?

greg-king5:
> BTW: Is there any kind of policy regarding the speed/size balance?

Ullrich aimed for high speed in run-time code that he believed is run often, in typical programs.  But, I don't think that there was an official policy.

mrdudz:
> ...and how can I fixed the: "Error: Maximum number of cycles reached."-error in Travis build running the lib_common_strcat-test?

have a look at the makefile in the respective directory... the maximum number of cycles is given to the simulator there, just increase it a bit :)

greg-king5:
>> ... and, how can I fix the: "Error: Maximum number of cycles reached."-error in Travis build running the lib_common_strcat-test?
>
> Have a look at the makefile in the respective directory.  The maximum number of cycles is given to the simulator there; just increase it a bit. :-)

It's a long-standing bug in the compiler.  The `-Os` command-line option is -- sort of -- "broken".  When `-O` is used, cc65 will expand calls to some string functions into inline code.  That code uses eight-bit counters; therefore, it can't handle strings that are longer than a RAM page.  Normally, cc65 won't expand those calls when it knows that the strings are too long -- but, `-Os` ignores that restriction!

That bug hasn't been a problem -- until now.  Your test programs deliberately use page-and-a-half strings.  So, they are guaranteed to fail the passes that compile with `-Os`.

I like @clbr's idea about unrolling some of the inline code's loops (issue #356).  I think that `-Os` can be used to switch that feature on/off (instead of its current implementation).  But for now, I think that we should remove the `s` from all of the makefiles' optimizer options.

mrdudz:
i strongly disagree - the CI tests should go through all optimizer functions (and that optimization is simply broken and should be removed/fixed)

greg-king5:
It will be replaced -- but, not right now (we must design the code that will replace it).  In the meantime, we don't want the regression tester to mislead @IrgendwerA8 into thinking that his code has failed.  (We will put those optimizer steps back in when they work again.)

IrgendwerA8:
Thanks for the clarification. I'll wait until an other commit clears the situation.

Regarding the "s"-optimization: I tend to agree with mrdudz. If it's known that the compiler produces invalid code in that case, stop the compiler from generating it until it is fixed. I'm pretty sure others aren't also happy to "discover" (debug) this "feature"...

For the size/speed balance: In the described "decspX"-cases I think the penalty isn't that bad. Spending only one cycle for a saved byte sounds quite comfortable to me...?

oliverschmidt:
The size/speed question is indeed a difficult one. However if it's actually one cycle vs. one byte I personally would spend the cycle and save the byte.

mrdudz:
i'd say... depends :=) is that an inlined function (so that one byte may add up to a lot of bytes?) - then save it. is it an often used subroutine? then prefer speed.

oliverschmidt:
As far as I can see everybody wants to have this pull request merged. And as far as I can see the only issue left is the incompatibility of the (highly welcome) test programs with the -Os compiler option.

Therefore I suggest that the owner of this pull request adds
```
// temporarily disable optimizations altogether until a fine grain control
// is implemented on Makefile level only disabling the compiler option -Os
#pragma optimize (off)
```
to the test programs requiring it.

mrdudz:
sounds sane to me

IrgendwerA8:
Added the "pragma" like advised, no change...

greg-king5:
> Added the "pragma", like advised; no change.

I didn't expect that.  Like Oliver, I thought that inlining is done only if optimization is enabled.  I looked at cc65's code that handles inlining.  I saw checks for `-Os`; I didn't see a check for `-O`.  So, I confirm that cc65 _always_ attempts to inline calls to `memcpy()`, `memset()`, `strcmp()`, `strcpy()`, and `strlen()`.

In order for an attempt to succeed:
- When an array argument is a pointer variable, it must have the `register` storage class.  Otherwise, it must be an array name.
- An array's size must be constant, known to the compiler, and less than 256 bytes -- unless the `-Os` option makes cc65 ignore that rule.
- When an array is on the C stack (automatic storage class), its entire length must be within 255 bytes of the top of the stack.

oliverschmidt:
> So, I confirm that cc65 always attempts to inline calls to memcpy(), memset(), strcmp(), strcpy(), and strlen().

I must admit that I personally consider this so surprising that I had to check myself to believe it.

I presume that everybody agrees that inlining is a classic optimization operation. At least from my perspective that means that without any optimization option given the compiler shouldn't do any inlining.

One of the many reasons is that if one puts i.e. a strlen() somewhere into a non-optimized program one can expect to have a) _strlen referenced/linked and b) that a breakpoint set at _strlen is hit at the time the program calls strlen().

Therefore I believe the current behavior (inlining although no optimization option is given) is to be classified as a bug. And as such I consider it appropriate to change the behavior in an incompatible way.

I've been looking at the code in https://github.com/cc65/cc65/blob/master/src/cc65/stdfunc.c a bit myself. As far as I understand one needs to AND the current expressions enabling the inlining with a check for optimizations being active. 

So far so good. However the question is which optimization option should enable the inlining: `-O` or `-Oi`? I think reading the corresponding section at http://cc65.github.io/doc/cc65.html#ss2.2 to find out which option favors speed and which option favors size one comes to the conclusion:
- `-O` and `-Or` favor size.
- `-Oi` and `-Os` favor speed.

From that perspective I think it would be by far least surprising that std function inlining is in general enabled with `-Oi` (or `--codesize`) and that `-Os` additionally enables "unsafe" std function inlining.

Conclusion: I suggest to make std function inlining depending on `-Oi` (or `--codesize`) being active. Doing so doesn't require to change the corresponding section at http://cc65.github.io/doc/cc65.html#ss2.2 in any way.

I personally have a hard time to invest the time necessary to implement the change in question. If someone is willing and able to contribute (soon) a pull request doing so it would be nice if he would answer here in order to avoid duplicate effort.

groessler:
Just for the record: I agree. Inlining shouldn't happen if no optimization flags are specified. Only "-Oi"  and probably also "-Os" should enable inlining.

mrdudz:
this surprises me as well - given how dedicated Uz was to meet the standard requirements

IMHO the "normal" optimization options (-Oxyz) should never result in code that violates the standard. however, i would still like to see the current behaviour preserved in something like -fuse-speedy-strings (or whatever). similar to gcc's "fast-math"

oliverschmidt:
> Only "-Oi" and probably also "-Os" should enable inlining.

Good point. I guess as `-Os` means enable even "unsafe" std function inlining it should implicitly enable "safe" std function inlining.

So let's presume "safe" std function inline should be dependent on `-Oi`. Then the question is which value of codesize should actually enable it. Looking at http://cc65.github.io/doc/cc65.html#option-codesize I see two viable approaches:
- codesize >= 200. This means that `-Oi` enables it.
- codesize > 100. This means that every non-default codesize asking for speed optimization enables it.

The latter approach could be interesting as it would allow to obtain todays behavior without setting any optimization option (-> just std function inlining) by setting `--codesize 101`.

BTW 1: The docs says for `--codesize`: "Accept larger code by factor x". As far as I understand it should rather read "Accept larger code by x percent".

BTW 2: I never realized in all the years that one can set codesize to a value < 100. There's one location in the compiler code looking out for that: https://github.com/cc65/cc65/blob/master/src/cc65/coptsize.c#L882

clbr:
FWIW, I don't think -Os should do inlining it knows will fail, or will
likely fail. That's just asking for trouble, it should always be safe
to use maximum optimizations.

oliverschmidt:
> I don't think -Os should do inlining it knows will fail, or will likely fail. That's just asking for trouble, it should always be safe to use maximum optimizations.

My first thought was that I don't see why we should change the behavior of `-Os`. But thinking about it some more I'm wondering it it wouldn't indeed be a good idea to have `-Os` enable std function inlining in general (independently from `-Oi`) and drop forced std function inline altogether...

mrdudz:
> FWIW, I don't think -Os should do inlining it knows will fail, or will likely fail. That's just asking for trouble, it should always be safe to use maximum optimizations.

yes. this. any "unsafe" optimizations should go into seperate options and they should never be used by default by the "normal" optimizer options.

greg-king5:
> My first thought was that I don't see why we should change the behavior of `-Os`. But thinking about it some more, I'm wondering if it wouldn't indeed be a good idea to have `-Os` enable std. function inlining in general (independently from `-Oi`), and drop forced std. function inline altogether.

That's what `-Os` did when Uz first put it into cc65.  But then later, he changed it so that inlining is "normal", and `-Os` does something extra.

oliverschmidt:
> That's what -Os did when Uz first put it into cc65. But then later, he changed it so that inlining is "normal", and -Os does something extra.

Thanks for that extra info. I'm inclined to want to go back to how it was before. Thoughts?

silverdr:
> On 2017-03-09, at 14:41, Oliver Schmidt <notifications@github.com> wrote:
> 
> That's what -Os did when Uz first put it into cc65. But then later, he changed it so that inlining is "normal", and -Os does something extra.
> 
> Thanks for that extra info. I'm inclined to want to go back to how it was before. Thoughts?

+1 from here.

-- 
SD!

oliverschmidt:
So if no one has good arguments against it I suggest to go back to Uz' original design: `-Os` enables std function inlining.

It would obviously nice to have a "special" option to activate unsafe std function inlining (what currently `-Os` does) but after all I leave this decision to the contributor of the necessary pull request.

greg-king5:
The plan is:
1. Function inlining defaults to **disabled**.
2. `-Os` and `--inline-funcs` enable cautious inlining.
3. `--force-inline-funcs` enables non-cautious inlining.

Question:
Should switch `#3` imply switch `#2`?  Or, should it be ignored unless we _explicitly_ use switch `#2`?

The pragmas will be `#pragma inline-funcs` and `#pragma force-inline`.  They will be independent of each other.  A programmer will be able to use `#pragma force-inline` where he knows better that cc65 can know; but, he still can control overall inlining from the command-line.

silverdr:
For what my 1.3¢ is worth here - no strong opinion but it seems natural to me that #3 implies #2

oliverschmidt:
Hi Greg,

Thanks for laying out a plan :-)

> `--inline-funcs`

I'm wondering if we shouldn't find a name that contains 'std'. Because it's not about functions written by the user but standard functions and because it fits better to the 's' in -Os.

> They will be independent of each other.

Does 'each' refer to each pragma or does 'each' refer to each way of controlling std func inlining (cmdline options vs. pragmas). Your following sentence seems to lead to the latter interpretation.

> Should switch #3 imply switch #2?

I'd say 'yes'.

greg-king5:
> Thanks for laying out a plan.

I did it so that I can get people's opinions about good names for the switches and the pragmas.

>> `--inline-funcs`
>
> I'm wondering if we shouldn't find a name that contains 'std'. Because it's not about functions written by the user, but standard functions; and, because it fits better to the 's' in `-Os`.

OK, `--inline-std-funcs` and `#pragma inline-std-funcs`.

>> They will be independent of each other.
>
> Does 'each' refer to each pragma; or, does 'each' refer to each way of controlling std. func. inlining (cmdline options vs. pragmas)? Your following sentence seems to lead to the latter interpretation.

"Each other" refers to the pragmas.  The idea is that the second pragma is a hint from a programmer to cc65.  It says, "**IF** you happen to be inlining functions, then it's safe to be more aggressive in this section.  If you aren't inlining, then ignore me."  Unlike the command-line switch, it won't imply the other pragma.  Because of that, a programmer can put it in his source code permanently; he won't need to remove it if he needs to disable inlining (e.g., when debugging).

Hmm, after I typed the above paragraph, I started to wonder about the pragma's name.  "Force" doesn't seem appropriate for a hint.  But, I don't want to stray far away from the name of its command-line switch.  Maybe `--more-inline-std-funcs` and `#pragma more-inline-std-funcs`.

oliverschmidt:
>"IF you happen to be inlining functions, then it's safe to be more aggressive in this section. If you aren't inlining, then ignore me."

Okay, now I got the point.

> "Force" doesn't seem appropriate for a hint. But, I don't want to stray far away from the name of its command-line switch. 

It seems to me that the semantics of the switch in question and the pragma in question are rather different (hoping vs. knowing and implying inlining vs. passive without inlining). Therefore I don't think that their names have to be similiar.

What about `--force-inline-std-funcs` and `#pragma inline-std-funcs-safe` ?

IrgendwerA8:
`--inline-std-aggressive` and `#pragma permit-std-aggressive` ?

oliverschmidt:
greg-king5 wrote:

> The plan is: [...]

Hi Greg,

I interpreted your statement above in the way that you intend to provide a pull request that actually implements the changes in question. I certainly don't want to appear as pushing you but I'd like to understand if my interpretation is correct.

Thanks, Oliver

greg-king5:
> I interpreted your statement above in the way that you intend to provide a pull request that actually implements the changes in question. I certainly don't want to appear as pushing you; but, I'd like to understand if my interpretation is correct.

It is correct.  I have implemented
```sh
--inline-std-funcs
```
```C
#pragma inline-stdfuncs ([push,] on|off)
#pragma allow-eager-inline ([push,] on|off)
```
But, I still haven't picked a good name for the other command-line switch.  I'm leaning toward
```sh
--eagerly-inline-funcs
```

oliverschmidt:
> #pragma allow-eager[...]
> --eagerly[...]

At least to me this seems to be a pretty nice way to both...
* express that both (the pragma and option) are related to the same thing
* express that they behave differently ("allow eager" vs. "[do] eagerly")

At least for me personally those names are perfectly fine :-)

oliverschmidt:
With https://github.com/cc65/cc65/commit/02daf9f8b5c1ae2267561ee05ce67b2d0393c12d the cc65 change discussed was implemented.

So the new tests added with this pull request finally succeed. The `#pragma optimize (off)` added to the tests should be removed again but given the patience the author of this pull request showed so far I consider it appropriate to take care of that after merging.

PR #395, opened 2017-03-05
Remove a pointless cc65 command-line option from "samples/Makefile".

greg-king5:
See [this PR comment](https://github.com/cc65/cc65/pull/389#issuecomment-284243168).

### DIFF
--- a/samples/Makefile
+++ b/samples/Makefile
@@ -90,7 +90,7 @@ LDFLAGS_tgidemo_atarixl    = --start-addr 0x4000
 %: %.s
 
 .c.o:
-	$(CC) $(CFLAGS) -Oirs --codesize 500 -T -g -t $(SYS) $<
+	$(CC) $(CFLAGS) -Ors --codesize 500 -T -g -t $(SYS) $<
 	$(AS) $(<:.c=.s)
 
 .s.o:


mrdudz:
as said in the other PR, i strongly disagree with this. the right way would be disabling the optimizer step *in the compiler*, thus preventing bad code from being generated for everyone. patching the tests is just wrong.

oliverschmidt:
As far as I understand the behavior of the compiler is like this for many years. We've seen no error reports on it. On the other hand people require all existing optimizer options. I don't follow your argument about removing the option from the compiler.

oliverschmidt:
I'd like to see the old, correct Makefile lines being commented out (with a comment saying that they are commented out temporarily due to a compiler issue) rather than being "just" changed.

mrdudz:
> As far as I understand the behavior of the compiler is like this for many years. We've seen no error reports on it. On the other hand people require all existing optimizer options. I don't follow your argument about removing the option from the compiler.

the point is that removing the options from the CI tests does more harm than good. you are right, leaving the options in the compiler doesnt make things worse. however, no more testing it certainly does. and this kind of thing should never be done only to make a certain test work (there are other testcases that will only break with optimizations enabled - do we also add them now and remove more optimizations from the tests so these can work?)

oliverschmidt:
>  you are right, leaving the options in the compiler doesnt make things worse

Good that we agree on that :-)

> there are other testcases that will only break with optimizations enabled

That's indeed a very strong point. If I understand things correct I see two options:

- Temorarily change the new tests to uses shorter strings.
- Temorarily don't build the new tests with the option in question. Maybe a relatively easy way to do that would be to use a #pragma to override the command line option (?)

greg-king5:
I looked at the cc65 code again.  My description was wrong.  It isn't broken; it's risky.  `-Os` is meant to be used only where a programmer can be certain that a program never will use strings that are too long.  Then, one can force cc65 to inline some functions even when it can't know about those strings.  `-Os` isn't a generic optimization feature; it's designed for a special condition.

We don't want to shorten the strings because Christian's tests must verify that the library code can handle strings of any length.

I think that we have three choices:
1. Move those tests back to `testcode/lib/`.
2. Don't test any combination of `-O` with `s` in it.
3. Add exception rules, for those programs, to the makefile.  (I think that we should make a distinction between tests that break because cc65 currently makes some bad code, and tests that are supposed to break in certain conditions.  Christian's tests are supposed to disallow the `-Os` option.)

oliverschmidt:
> I looked at the cc65 code again.

Thanks.

> -Os is meant to be used only where a programmer can be certain that a program never will use strings that are too long.

I don't want to apper as a know-it-all but that's nothing new. It's even documented: "The inlined string and memory functions will not handle strings or memory areas larger than 255 bytes. Similarly, the inlined is..() functions will not work with values outside the char. range (such as EOF)."

> We don't want to shorten the strings because Christian's tests must verify that the library code can handle strings of any length.

I see.

> Add exception rules, for those programs, to the makefile.

Given the feedback / opinions received so far this seems the only "good" option. The question is: Who will take care of that ;-)

I could imagine to add `#pragma optimize (off)` to the tests in question as a quick workaround in order to allow to move forward with https://github.com/cc65/cc65/pull/389

mrdudz:
i would much prefer:

4. create another optimizer option that *only* enables such "risky" optimizations. -Os is not exclusively handling those, is it?

groessler:
regarding option #3:

When I look at val/Makefile there are already some compiler overrides for K&R. Couldn't we add simiar overrides for the failing string tests?  For this to work, multiple -O options should be handled by the driver that only the last one is used. gcc for example does that. I don't know if cc65 does that.

greg-king5:
I decided that a merge of groepaz's and Oliver's ideas is a better way to go.  The test makefiles will continue to test all available combinations of the `-O` command-line option.  We will use a `#pragma` to make an exception directly in the programs that need it.

I still removed `-Os` from the samples makefile.  That directory is meant to give people some examples of how to utilize cc65.  I think that it shouldn't give them the impression that `-Os` is a standard optimization switch.

samples/Makefile
```diff
@@ -90,7 +90,7 @@ LDFLAGS_tgidemo_atarixl    = --start-addr 0x4000
 %: %.s
 
 .c.o:
-	$(CC) $(CFLAGS) -Oirs --codesize 500 -T -g -t $(SYS) $<
+	$(CC) $(CFLAGS) -Oir --codesize 500 -T -g -t $(SYS) $<
```
oliverschmidt: As far as I understand `-Oi` equals `--codesize 200` so if there's a `--codesize 500` following we should rather drop `-Oi`, shouldn't we?

MonteCarlos:
Depends, which option we give higher priority. Is this defined?

oliverschmidt:
> Depends, which option we give higher priority. Is this defined?

If I'm not totally mistaken there's nothing like priorities. The option are read from left to right. The most right option "wins".

mrdudz:
i'd like it stay like this too - because this way it is very simple to determine what is being used. once special priority rules are introduced it will be much harder

samples/Makefile
greg-king5: Yep, it's redundant.

greg-king5:
I'm frugal; I don't want to waste a pull request.  So, this one has morphed twice.  Now, it takes out a redundant switch.

oliverschmidt:
> I'm frugal; I don't want to waste a pull request.

;-)

PR #398, opened 2017-03-09
Add a pragma for finer control of forced inlining of some C functions.

greg-king5:
`#pragma force-inline(on | off)` can be used to control where cc65 will inline calls to certain array and string functions.  Inlining can be controlled on an expression-by-expression basis.  (The pragma doesn't affect what `<ctype.h>` does.  But, you can use `#undef` to disable that header's inlining.)

I added a command-line switch that lets us enable the forcing independently of the `-O` switch.  But, `-O` will disable `--force-inline-funcs` if `-O` follows that long switch.

This Pull Request will allow C library string function tests to be added to the automated regression test suite, without needing to alter the makefile.  See the comments at PRs #389 and #395.

### DIFF
--- a/src/cc65/main.c
+++ b/src/cc65/main.c
@@ -88,7 +88,7 @@ static void Usage (void)
             "  -O\t\t\t\tOptimize code\n"
             "  -Oi\t\t\t\tOptimize code, inline more code\n"
             "  -Or\t\t\t\tEnable register variables\n"
-            "  -Os\t\t\t\tInline some known functions\n"
+            "  -Os\t\t\t\tForce inlining of some known functions\n"
             "  -T\t\t\t\tInclude source as comment\n"
             "  -V\t\t\t\tPrint the compiler version number\n"
             "  -W warning[,...]\t\tSuppress warnings\n"
@@ -119,6 +119,7 @@ static void Usage (void)
             "  --dep-target target\t\tUse this dependency target\n"
             "  --disable-opt name\t\tDisable an optimization step\n"
             "  --enable-opt name\t\tEnable an optimization step\n"
+            "  --force-inline-funcs\t\tForce inlining of some known functions\n"
             "  --help\t\t\tHelp (this text)\n"
             "  --include-dir dir\t\tSet an include directory search path\n"
             "  --list-opt-steps\t\tList all optimizer steps and exit\n"
@@ -254,7 +255,7 @@ static void SetSys (const char* Sys)
         case TGT_TELESTRAT:
             DefineNumericMacro ("__TELESTRAT__", 1);
             break;
-                                
+
         case TGT_NES:
             DefineNumericMacro ("__NES__", 1);
             break;
@@ -556,7 +557,7 @@ static void OptDebugOpt (const char* Opt attribute ((unused)), const char* Arg)
 
 
 
-static void OptDebugOptOutput (const char* Opt attribute ((unused)), 
+static void OptDebugOptOutput (const char* Opt attribute ((unused)),
                                const char* Arg attribute ((unused)))
 /* Output optimization steps */
 {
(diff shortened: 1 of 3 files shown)


src/cc65/main.c
```diff
@@ -951,6 +962,9 @@ int main (int argc, char* argv[])
 
                 case 'O':
                     IS_Set (&Optimize, 1);
+                    IS_Set (&CodeSizeFactor, 100);
+                    IS_Set (&EnableRegVars, 0);
+                    IS_Set (&InlineStdFuncs, 0);
```
oliverschmidt: I must admit that I don't understand why you added these three lines. I imagine that i.e.
```
-Or -O
```
should not turn register vars off, should it?

oliverschmidt:
Hi Greg,

First of all thanks for providing this pull request.

I fully understand that the discussion about std function inlining is sort of separate from this pull request: Even if we fix the compiler to not inline std functions without being asked to do so the test Makefile still asks him to do so - so it's necessary to have a way to turn it explicitly off vie #pragma.

However if we would decide to drop forced std function inlining then the #pragma in question would need a different name. So I'd like to have this question sorted out first before merging this pull request. Does that make sense to you?

Regards,
Oliver

src/cc65/main.c
greg-king5: > I must admit that I don't understand why you added those three lines. I imagine that, i.e.,
> ```
> -Or -O
> ```
> should not turn register vars. off, should it?

It was suggested by @groessler in [this comment](https://github.com/cc65/cc65/pull/395#issuecomment-284393260).

greg-king5:
> However, if we would decide to drop forced std. function inlining, then the `#pragma` in question would need a different name. So, I'd like to have this question sorted out first before merging this pull request. Does that make sense to you?

Yes.

src/cc65/main.c
groessler: At least on gcc, the last -O option in the command line "wins". So -O2 -O0 will disable optimizations.

mrdudz: but this is different... -O2 and -O0 are optimization levels, which enable/disable a whole lot of seperate options. -Os and -Oi are not like that, those enable detailed options, and should add up on the commandline, IMHO

oliverschmidt: I agree with mrdudz: -O, -Oi, -Os and -Or are different option which just happen be be related to optimization and and are therefore allowed to be written as -Oisr. But I don't see their setting to influence each other.

groessler: I don't see -O2 and -O3 etc. different from -Osi or -Osir. Internally the gcc level also just enable or disable different optimization steps.

If cc65 would behave like gcc it would be rather simple to change the Makefile to opit specific optimizations for "problem files".

But I noticed that cc65 has a --disable-opt switch. This could be used in this example without changing the existing behaviour of the -O switches.

greg-king5: OK, those lines will disappear.

oliverschmidt: @groessler:

The doc explizitly says "It is possible to concatenate the modifiers for -O. For example, to enable register variables and inlining of known functions, you may use -Ors." And this is exactly what one intuitively expects. I'm having a hard time to believe that you really consider(ed) to make
`-Os -Or` mean something different than `-Osr`. This would be the opposite from all compilers I used in my life so far...

groessler: So you obviously haven't used gcc :-)

oliverschmidt: Are you refering to the -O0 vs. -O2 thingy above? Or some other gcc command line option? If the latter please tell me which one(s).

groessler: With all options in gcc AFAIK, if conflicting, the last one succeeds. Yes, I was referring to the -O options in gcc (-O<num>, -Ofast, -Os, -Og).
But, as I said, I think the current cc65 behaviour should be retained.

oliverschmidt:
Hi Greg,

As far as I understand https://github.com/cc65/cc65/commit/02daf9f8b5c1ae2267561ee05ce67b2d0393c12d (which picked up your latest ideas) makes this pull request obsolete.

Regards,
Oliver

PR #399, opened 2017-03-10
Fix line endings (CRLF -> LF) on all affected files.

groessler:
Some files were checked in with CRLF line endings.


oliverschmidt:
Thanks :-)

PR #402, opened 2017-03-12
few 6502 and some 65SC02 runtime optimizations

IrgendwerA8:


oliverschmidt:
Hi,

Maybe you're interested in but not aware of this:

The sim6502 and sim65C02 targets write a 1 byte header into the output file allowing to distinguish between both targets: https://github.com/cc65/cc65/blob/master/libsrc/sim6502/exehdr.s

The sim65 program reads that byte and activates the respective CPU: https://github.com/cc65/cc65/blob/master/src/sim65/main.c#L147

There are already two virtual CPUs defined: https://github.com/cc65/cc65/blob/master/src/sim65/6502.h#L48

However right now the 65C02 CPU is just a copy of the 6502 CPU: https://github.com/cc65/cc65/blob/master/src/sim65/6502.c#L2483

Could you imagine to enhance the virtual 65C02 CPU enough to actually successfully execute code for from the sim65C02 target? That would allow to add regression tests for the 65SC02 optimizations you added...

Regards,
Oliver

IrgendwerA8:
I'll see what I can do to raise the emulator at least to 65SC02 level. I know that this is not your first attempt to convince someone to do this... ;-)

oliverschmidt:
> I'll see what I can do to raise the emulator at least to 65SC02 level.

Great !

> I know that this is not your first attempt to convince someone to do this... ;-)

As long as you don't feel offended that I didn't ask you first... ; -))

IrgendwerA8:
Done (see #405). I leave changing the build configuration to have test runs with the 65C02 mode up to you ;-). My dry runs here ran well, e.g. the stroul test:
6502:
Failures: 0
624981 cycles
65c02:
Failures: 0
623482 cycles

oliverschmidt:
> I leave changing the build configuration to have test runs with the 65C02 mode up to you ;-)

That's perfectly fine. I suggest to have that build configuration change in place before merging this pull request.

libsrc/runtime/along.s
```diff
@@ -9,16 +10,12 @@
...
-        bcc     aulong          ; Yes, handle like unsigned type
-        ldx     #$ff
-        stx     sreg
+aulong: ldx     #0
+store:  stx     sreg
         stx     sreg+1
         rts
-
-
+
```
greg-king5: We don't want "dangling" lines (lines without newline characters).

libsrc/runtime/lsave.s
```diff
@@ -1,5 +1,6 @@
 ;
 ; Ullrich von Bassewitz, 08.08.1998
+; Christian Krueger, 11-Mar-2017, optimization
 ;
 ; CC65 runtime: save ax into temp storage/restore ax from temp storage
```
greg-king5: Would you mind changing those "`ax`" words into "`eax`" words?

oliverschmidt:
>> I leave changing the build configuration to have test runs with the 65C02 mode up to you ;-)

> That's perfectly fine.

Done: https://github.com/cc65/cc65/commit/4a6bca0b560ae2c1a7cfe3b14f75ee38094b9b76

IrgendwerA8:
Cool. Seems that the work was worth it.
(Style complaints handled.)

oliverschmidt:
Now that the 65C02 regression tests are in place - and don't indicate any issues with this pull request - it seems to be the right time to (finally) merge...

IrgendwerA8:
Thanks!
Here the updated results:
6502:
Failures: 0
624927 cycles
65c02 (50 bytes smaller):
Failures: 0
621326 cycles

oliverschmidt:
Thanks for the update.

polluks:
OPC_6502_F8 needs
```
    Cycles = 2;
    Regs.PC += 1;
```

polluks:
Please take care of
http://www.6502.org/tutorials/65c02opcodes.html#6 and
http://www.6502.org/tutorials/65c02opcodes.html#7

IrgendwerA8:
Feel free to create an own pull request as I have currently no time.
As added below the header I'm aware of e.g.
http://www.6502.org/tutorials/65c02opcodes.html#7
but thought it would be better to have some 65C02 support than none.
OPC_6502_F8 is IMHO really bad, as I think the simulator would hang up.
Seems nobody to use ever since...

oliverschmidt:
> [...] thought it would be better to have some 65C02 support than none.

Full ACK. Sim65's task is to allow for automated regression tests. Anything beyond that is purely nice-to-have.

PR #411, opened 2017-03-21
Fix regression of #pragma bss-name

### LINKED ISSUE #409: C ZP vars regressed
The current master produces invalid assembly for C files that put variables in ZP. I have a known working version from Jan 2016, seeing if I could bisect this.

```c
#pragma bss-name (push,"ZEROPAGE")
#pragma data-name (push,"ZEROPAGE")

char oam_off;

#pragma data-name(pop)
#pragma bss-name (pop)
```

Today's master produces this:
```asm
;
; File generated by cc65 v 2.15 - Git fe59626
;
        .fopt           compiler,"cc65 v 2.15 - Git fe59626"
        .setcpu         "6502"
        .smart          on
        .autoimport     on
        .case           on
        .debuginfo      off
        .importzp       sp, sreg, regsave, regbank
        .importzp       tmp1, tmp2, tmp3, tmp4, ptr1, ptr2, ptr3, ptr4
        .macpack        longbranch
        .export         _oam_off

.segment        "DATA"

.segment        "ZEROPAGE"
.segment        "DATA"

.segment        "BSS"

.segment        "ZEROPAGE"
.segment        "BSS"
_oam_off:
        .res    1,$00
```

As you can see, the var does not go to ZP. Here's the correct output, from cc65 from Jan 2016.

```asm
;
; File generated by cc65 v 2.15 - Git 24c2da9
;
        .fopt           compiler,"cc65 v 2.15 - Git 24c2da9"
        .setcpu         "6502"
        .smart          on
        .autoimport     on
        .case           on
        .debuginfo      off
        .importzp       sp, sreg, regsave, regbank
        .importzp       tmp1, tmp2, tmp3, tmp4, ptr1, ptr2, ptr3, ptr4
        .macpack        longbranch
        .export         _oam_off

.segment        "DATA"

.segment        "ZEROPAGE"
.segment        "DATA"

.segment        "BSS"

.segment        "ZEROPAGE"
_oam_off:
        .res    1,$00
.segment        "BSS"
```

(note the git sha is wrong, because the build system does not force-update version.o when the git rev changes)

pfusik:
Closes #409 

Added tests, also for the two corner cases I mentioned in #409.
The sim6502 linker scripts are modified because of the `bss-name.c` test. I'm not sure if it's ok.

### DIFF
--- a/cfg/sim6502.cfg
+++ b/cfg/sim6502.cfg
@@ -3,7 +3,7 @@ SYMBOLS {
     __STACKSIZE__: type = weak, value = $0800; # 2k stack
 }
 MEMORY {
-    ZP:     file = "",               start = $0000, size = $001A;
+    ZP:     file = "",               start = $0000, size = $001B;
     HEADER: file = %O,               start = $0000, size = $0001;
     MAIN:   file = %O, define = yes, start = $0200, size = $FDF0 - __STACKSIZE__;
 }
--- a/cfg/sim65c02.cfg
+++ b/cfg/sim65c02.cfg
@@ -3,7 +3,7 @@ SYMBOLS {
     __STACKSIZE__: type = weak, value = $0800; # 2k stack
 }
 MEMORY {
-    ZP:     file = "",               start = $0000, size = $001A;
+    ZP:     file = "",               start = $0000, size = $001B;
     HEADER: file = %O,               start = $0000, size = $0001;
     MAIN:   file = %O, define = yes, start = $0200, size = $FDF0 - __STACKSIZE__;
 }
--- a/src/cc65/compile.c
+++ b/src/cc65/compile.c
@@ -242,16 +242,24 @@ static void Parse (void)
                             Error ("Variable `%s' has unknown size", Decl.Ident);
                         }
                         Entry->Flags &= ~(SC_STORAGE | SC_DEF);
+                    } else {
+                        /* A global (including static) uninitialized variable
+                        ** is only a tentative definition. For example, this is valid:
+                        ** int i;
+                        ** int i;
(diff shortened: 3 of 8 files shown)


oliverschmidt:
> The sim6502 linker scripts are modified because of the bss-name.c test. I'm not sure if it's ok.

Maybe I'm missing the point but as far as I understand the test wants to place an additional byte in the ZEROPAGE segment. So it seems natural to me to increase this segment to allow for that. After all sim65's primary purpose to act as regression test vehicle...

pfusik:
Yes, that is the point. My doubt was because the ZP size was exactly the size of cc65 "registers", perhaps that was to catch in tests that no additional "registers" are added by mistake. Otherwise I would expect a much relaxed ZP limit.

greg-king5:
> My doubt was because the ZP size was exactly the size of cc65 "registers"; perhaps that was to catch, in tests, that no additional "registers" are added by mistake. Otherwise, I would expect a much relaxed ZP limit.

It probably was just cut-and-paste from other CFG files.

oliverschmidt:
> My doubt was because the ZP size was exactly the size of cc65 "registers", perhaps that was to catch in tests that no additional "registers" are added by mistake.

No I see your point. I'm pretty sure there's no such test.

> Otherwise I would expect a much relaxed ZP limit.

Hm, so far there was no issue with that limit so it can't have been "that wrong", can it? ;-)

I personally currently see no reason against expanding it to the full $FF...

oliverschmidt:
> It probably was just cut-and-paste from other CFG files.

Sorry for my ignorance but I just saw no reason for a larger size...

oliverschmidt:

PR #412, opened 2017-03-22
Add makefile code to update the tools' Git revision string.

### LINKED ISSUE #410: Build system doesn't force-update version.o when git rev changes
As in title. This means an incremental build may have the old commit in the version string.

greg-king5:
Closes #410.

This code updates a tool's version string only when that tool actually needs to be rebuilt.  So, the string will be accurate for each tool.  And, it should handle building inside and outside of Git repositories.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -139,7 +139,17 @@ define PROG_template
 
 $$(eval $$(call OBJS_template,$1))
 
-../bin/$1$(EXE_SUFFIX): $$($1_OBJS) ../wrk/common/common.a | ../bin
+ifeq ($(GIT_SHA),N/A)
+$1_VERSION =
+else
+$1_VERSION = ../wrk/$1/version.o
+
+$$($1_VERSION): $$($1_OBJS) ../wrk/common/common.a
+	@echo $1 -- Git $(GIT_SHA)
+	@$(CC) -c $(CFLAGS) -o $$@ common/version.c
+endif
+
+../bin/$1$(EXE_SUFFIX): $$($1_OBJS) $$($1_VERSION) ../wrk/common/common.a | ../bin
 	$$(CC) $$(LDFLAGS) -o $$@ $$^ $$(LDLIBS)
 
 $1: ../bin/$1$(EXE_SUFFIX)


oliverschmidt:
I've spent quite some time by now staring at the diff but I still don't get it :-(

- As far as I understand https://github.com/cc65/cc65/issues/410 complains about the binaries not showing the Git revision present when building them. That means from my perspective that if i.e. a new Git revision only changes some C library stuff for platform <you name it> still all binaries should be rebuilt in order to reflect that Git revision. Mayby I'm missing the point but I don't see this pull request changing that situation - and therefore I don't see this pull request closing https://github.com/cc65/cc65/issues/410

- Maybe I'm missing the point even more but as far as I can see this pull request strives to have each binary show the Git revision that that changed something relevant to that binary. To me this looks sort of the opposite of what https://github.com/cc65/cc65/issues/410 wants to have. Apart from that I must admit that I personally don't the the Git revision shown by a binary as an attribute of that individual binary but rather as an attribute of the cc65 installation as a whole that binary is part of.

So again I might be misunderstanding things - but if I don't then I see this pull request moving things in the very wrong direction :-(

greg-king5:
I read in #410 that he is disappointed that the version string doesn't change -- at all.  He didn't say "incremental build" of what; but, the timing of that issue makes me conclude that he was thinking about the recent changes in the compiler.  They didn't cause a change in
`cc65 --version`.

I don't like the idea that all of the tools should be rebuilt simply because we change some small thing in one of the platform libraries.

I see the Git number as a time-stamp.  It tells us what the state of the source code was when it went into that particular binary.  It is a debugging aid.  If we put a bug into a program -- but don't notice it immediately -- then the Git string can help us to see where in the log to start looking for what was done.

Also:
Suppose a person discovers that he has two copies of a tool.  He can see if they are different or the same by looking at the version display.  And, he can see how different they are by looking at the Git log.  If that version string were being updated by things that have nothing to do with that program, then those copies would look different -- even when they have the same code!

On the other side:
When the project is built for distribution, the build usually starts with empty "`*wrk/`" directories.  Then, everything is built from the current Git HEAD; everything has the same Git SHA; and, the version number actually is an attribute of the entire distro.  So, I think that this PR can give both of us what we want -- depending on the circumstance of the build.

MonteCarlos:
Whats about checking only the major version number when using libraries?

This way small changes which are represented by the minor and revision 
part of the version number would not cause linking issues.

Only if the major part of the version changes libraries would need to be 
rebuild.


Monte Carlos

oliverschmidt:
Hi Greg,

I certainly dislike when we have diverging opinions on cc65 topics and therefore I deliberately took some time before I answer here, but unfortunately I nevertheless totally disagree...

The user aspect: cc65 translates source code to a binary. The binary is influenced by several tools and the library. If the binary doesn't work as expected and one wants to track down the root cause one needs to check all used tools and the library. If one wants to know from the user what he used to create the binary one wants to ask for one SHA, not for n SHAs of all tools and the library (oops, the library doesn't have an SHA at all). I'm with you seeing the SHA as a timp-stamp. There's only one cc65 time-stamp, not several time-stamps of several tools (and none for the libary).

The Git aspect: A Git commit SHA includes the top level tree SHA. So it's by its very nature a property of the whole repo, not one file or one directory. If one is interested in the commits touching one directory only then `git log` has all the options to filter that as desired.

The build aspect: (One of) the primary design goal of an incremental build is to yield the same result as a scratch build - just faster. It's a bad idea (tm) to intentionally have an incremental build yield a different result as you suggest.

Conclusion: If there's a simple-enough way to have an incremental cc65 build work like a scratch cc65 build in that all tools show the (same) SHA of the last Git commit (which I think is what https://github.com/cc65/cc65/issues/410 asks for) then I'm willing to merge that.

Regards,
Oliver

PR #418, opened 2017-04-07
Fix the handling of the C construct "while (0) { }".

greg-king5:
It's a dark corner case; nobody will type it directly.  But, conditional macroes might create it.


oliverschmidt:
> dark corner case

Nice phrase ;-)

PR #424, opened 2017-05-02
Trampoline support

### LINKED ISSUE #422: Wishlist: pragma trampoline
```c
// header

void mytrampoline(void);
// mytrampoline is in a common bank, written in asm

// these functions are in bank 0
#pragma trampoline (push, mytrampoline, 0)
void func(void);
u8 func2(u16 in);
#pragma trampoline (pop)
```

I propose a setup like this. The header would mark those functions as having this trampoline and this userdata value. When calling such a function, cc65 would setup the parameters as normal, write the function's address to ptr4 (probably with ldy/sty would be most optimal at that point), and the userdata value to tmp4. Then jsr mytrampoline.

The simplest trampoline would be an indirect jump, while the intended one would switch banks, callptr4, restore banks, rts. The trampoline must be in a common bank (or RAM, in which case it would be an absolute pointer value instead of a function name?).

As the target is NES, the userdata value is an u8, the maximum number of banks is 256. A common bank is also available in most cases. These assumptions may not hold for other platforms, please say if they behave differently.

clbr:
Fixes #422. Sponsored by Mega Cat Studios.

### DIFF
--- a/src/cc65/funcdesc.h
+++ b/src/cc65/funcdesc.h
@@ -67,8 +67,8 @@ struct FuncDesc {
     unsigned            ParamCount;     /* Number of parameters              */
     unsigned            ParamSize;      /* Size of the parameters            */
     struct SymEntry*    LastParam;      /* Pointer to last parameter         */
-    struct SymEntry*    Trampoline;     /* Pointer to the trampoline         */
-    unsigned char       TrampolineData; /* The trampoline's user data        */
+    struct SymEntry*    WrappedCall;     /* Pointer to the WrappedCall         */
+    unsigned char       WrappedCallData; /* The WrappedCall's user data        */
 };
 
 
--- a/src/cc65/funcdesc.h
+++ b/src/cc65/funcdesc.h
@@ -67,6 +67,8 @@ struct FuncDesc {
     unsigned            ParamCount;     /* Number of parameters              */
     unsigned            ParamSize;      /* Size of the parameters            */
     struct SymEntry*    LastParam;      /* Pointer to last parameter         */
+    struct SymEntry*    Trampoline;     /* Pointer to the trampoline         */
+    unsigned char       TrampolineData; /* The trampoline's user data        */
 };
 
 
--- a/src/cc65/funcdesc.h
+++ b/src/cc65/funcdesc.h
@@ -67,8 +67,8 @@ struct FuncDesc {
     unsigned            ParamCount;     /* Number of parameters              */
     unsigned            ParamSize;      /* Size of the parameters            */
     struct SymEntry*    LastParam;      /* Pointer to last parameter         */
-    struct SymEntry*    WrappedCall;     /* Pointer to the WrappedCall         */
-    unsigned char       WrappedCallData; /* The WrappedCall's user data        */
+    struct SymEntry*    WrappedCall;    /* Pointer to the WrappedCall        */
+    unsigned char       WrappedCallData;/* The WrappedCall's user data       */
 };
 
 
--- a/src/cc65/pragma.c
+++ b/src/cc65/pragma.c
@@ -509,7 +509,7 @@ static void WrappedCallPragma (StrBuf* B)
     Entry = FindSym(Name);
 
     /* Check if the name is valid */
-    if (Entry && Entry->Flags & (SC_FUNC | SC_STORAGE)) {
+    if (Entry && Entry->Flags & SC_FUNC) {
 
         PushWrappedCall(Entry, Val);
         Entry->Flags |= SC_REF;
--- a/src/cc65/pragma.c
+++ b/src/cc65/pragma.c
@@ -125,7 +125,7 @@ static const struct Pragma {
     { "static-locals",          PRAGMA_STATIC_LOCALS      },
     { "staticlocals",           PRAGMA_STATICLOCALS       },      /* obsolete */
     { "warn",                   PRAGMA_WARN               },
-    { "wrapped-call",		PRAGMA_WRAPPED_CALL	  },
+    { "wrapped-call",           PRAGMA_WRAPPED_CALL	  },
     { "writable-strings",       PRAGMA_WRITABLE_STRINGS   },
     { "zpsym",                  PRAGMA_ZPSYM              },
 };
--- a/src/cc65/pragma.c
+++ b/src/cc65/pragma.c
@@ -51,6 +51,7 @@
 #include "scanstrbuf.h"
 #include "symtab.h"
 #include "pragma.h"
+#include "trampoline.h"
 
 
 
@@ -87,6 +88,7 @@ typedef enum {
     PRAGMA_SIGNEDCHARS,                                 /* obsolete */
     PRAGMA_STATIC_LOCALS,
     PRAGMA_STATICLOCALS,                                /* obsolete */
+    PRAGMA_TRAMPOLINE,
     PRAGMA_WARN,
     PRAGMA_WRITABLE_STRINGS,
     PRAGMA_ZPSYM,
@@ -122,6 +124,7 @@ static const struct Pragma {
     { "signedchars",            PRAGMA_SIGNEDCHARS        },      /* obsolete */
     { "static-locals",          PRAGMA_STATIC_LOCALS      },
     { "staticlocals",           PRAGMA_STATICLOCALS       },      /* obsolete */
+    { "trampoline",		PRAGMA_TRAMPOLINE	  },
     { "warn",                   PRAGMA_WARN               },
     { "writable-strings",       PRAGMA_WRITABLE_STRINGS   },
     { "zpsym",                  PRAGMA_ZPSYM              },
@@ -437,6 +440,84 @@ static void SegNamePragma (StrBuf* B, segment_t Seg)
 
         /* Segment name is invalid */
         Error ("Illegal segment name: `%s'", Name);
+
+    }
+
+ExitPoint:
+    /* Call the string buf destructor */
+    SB_Done (&S);
+}
+
+
+static void TrampolinePragma (StrBuf* B)
+/* Handle the trampoline pragma */
+{
+    StrBuf      S = AUTO_STRBUF_INITIALIZER;
+    const char *Name;
+    long Val;
+    SymEntry *Entry;
+
+    /* Check for the "push" or "pop" keywords */
+    switch (ParsePushPop (B)) {
+
+        case PP_NONE:
+            Error ("Push or pop required");
+            break;
+
+        case PP_PUSH:
+            break;
+
+        case PP_POP:
+            PopTrampoline();
+
+            /* Done */
+            goto ExitPoint;
+
+        case PP_ERROR:
+            /* Bail out */
+            goto ExitPoint;
+
+        default:
+            Internal ("Invalid result from ParsePushPop");
+
+    }
+
+    /* A symbol argument must follow */
+    if (!SB_GetSym (B, &S, NULL)) {
+        goto ExitPoint;
+    }
+
+    /* Skip the following comma */
+    if (!GetComma (B)) {
+        /* Error already flagged by GetComma */
+        Error ("Value required for trampoline data");
+        goto ExitPoint;
+    }
+
+    if (!GetNumber (B, &Val)) {
+        Error ("Value required for trampoline data");
+        goto ExitPoint;
+    }
+
+    if (Val < 0 || Val > 255) {
+        Error ("Value must be between 0-255");
+        goto ExitPoint;
+    }
+
+    /* Get the string */
+    Name = SB_GetConstBuf (&S);
+    Entry = FindSym(Name);
+
+    /* Check if the name is valid */
+    if (Entry && Entry->Flags & (SC_FUNC | SC_STORAGE)) {
+
+        PushTrampoline(Entry, Val);
+        Entry->Flags |= SC_REF;
+
+    } else {
+
+        /* Segment name is invalid */
+        Error ("Trampoline does not exist or is not a function or array");
 
     }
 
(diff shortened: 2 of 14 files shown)


oliverschmidt:
Hi,

You surely noticed that neither your issue nor your pull request received any feedback ;-)

About every cc65 target supports some type of banking. And C code tends to generate large code compared to hand crafted 6502 code. So I certainly see the use case.

However I don't like the amount of code added for that nevertheless pretty specific feature. So you know without Uz being around the maintenance of the cc65 compiler isn't exactly easy and I'm therefore very reluctant to add a significant amount of code to it.

I for sure don't understand everything in this pull request but it seems to me that quite some code is for supporting the pragma push/pop. Do you consider it really necessary to be able to nest these pragma usages?

Apart from that I neither understand from your issue nor your pull request what the alternative to the function (that in "in RAM" ... array thingy) is supposed to be. Please elaborate.

It's good that you provide a test program. But that test program too doesn't cover that "in RAM" ... array thingy.

And finally I'm rather unhappy with the name "trampoline". That seems to be just one of the potential use cases. To me this seems to be a "wrapper". You could i.e. use it as well to implement some instrumentation for profiling. And the 8-bit "data" going into tmp4 is something I'd call "identifier" - because that's pretty much what you can do with just 8 bits.

clbr:
A single level would be enough, however I added the push/pop since almost every other pragma in cc65 also supports it, so it would be consistent. IntPtrStack is copy-paste from the existing IntStack - old, well tested code - other than that, the push/pop support counts only for a couple lines. If you prefer less code, I can certainly change that.

A RAM trampoline is intended for the case where there is no common bank, but everything must be switched. Thus, the trampoline code must be in RAM. This is achieved usually manually, writing the trampoline in asm (not via asm statements in C, because cc65 can optimize those and thus change the code size), then in init code copying it to that memory location. I didn't see any way for tests to include separate asm files, and as my target platform is NES, a common bank is almost always available.

In generated code terms, using an array instead of a function simply changes the jsr target.

The naming is due to being the primary use, that's what everyone will be googling for. If it was called "function wrappers", many potential users would miss the feature's existence. There's lots of google hits, even on the cc65 mailing list, for trampolines used with banked memory. It's true you could make a simple function profiler with it, a call graph similar to what gprof does.

The doc can absolutely be changed to call the data identifier instead. As mentioned in the issue, on NES there can not be more than 256 banks, thus it's not an identifier there (to be mapped into a more complex pointer, etc) - it's the direct value that can be sent to the bank-switching functionality.

oliverschmidt:
Thanks for your comprehensive reply :-)

> IntPtrStack is copy-paste from the existing IntStack - old, well tested code

Which proves how little I actually know about the cc65 compiler. I follow your arguments and would now like to see the push/pop things stay as-is.

> A RAM trampoline is intended for the case where there is no common bank, but everything must be switched. Thus, the trampoline code must be in RAM.

I'm having a hard time to follow you. You seem to make quite some assumptions which are maybe NES-specific. "no common bank, but everything must be switched" assumes that was is banked is ROM, correct? "the trampoline must be in RAM" assumes that the memory addresses that are not banked correspond to RAM, correct?

Given my assumptions I'd say that from a cc65 perspective this means the trampoline's RUN address differs from the trampoline's LOAD address. So from the perspective of this pull request a "RAM trampoline" means that instead of a trampoline symbol a trampoline address is provided, correct?

> In generated code terms, using an array instead of a function simply changes the jsr target.

Is that array "variant" related to the RAM variant? I presume not. Please elaborate on the array variant.

> The naming is due to being the primary use [...]

I have no problem at all if the documentation is targeted towards the trampoline use case (and therefore makes use of the term trampoline). However I don't see me merging this pull request if the pragma itself is called trampoline.

To me this is similiar to http://cc65.github.io/doc/ca65.html#ss17.1 - it's obvious what the ca65 .feature is usable for. Nevertheless it isn't called .tassmode or alike.

clbr:
Here's pseudo-C illustrating the RAM trampoline, maybe it makes things clearer:
```c
void trampoline(void); // asm function in a separate file, 10 bytes in size
unsigned char array[10]; // BSS/RAM location that holds the code at runtime

#pragma trampoline(push, array, 0)
void somefunc(int foo);
#pragma trampoline(pop)

int main() {
    memcpy(array, trampoline, 10); // Copy the code to RAM
    somefunc(); // This generates the trampoline setup and "jsr _array" - it jumps to the RAM address
}
```

Indeed, on NES most banking only concerns ROM, RAM banking is rare. So on a setup where all ROM must be switched, the trampoline must be in RAM.

On naming, would "pragma func-wrapper" be ok?

It is identifying the bank yes, but I thought you meant via an indirection, not a direct one. Will change.

oliverschmidt:
> Here's pseudo-C illustrating the RAM trampoline [...]

Thanks, I think now I see. So you are obusing a data structure which happens to be a char array as place holder for code. This looks a lot like "assembler-in-C" to me.

I haven't tried but from my perspective it would be appropriate to use `#pragma code-name ("DATA")` for the trampoline, rely on `copydata` to place it in RAM and just use the function name as if it were in ROM.

Apart from that: Can you point out the code in your pull request that is necessary to make the "array hack" work?

> On naming, would "pragma func-wrapper" be ok?

Let's see: "func-wrapper" sounds a bit like code would be added to the function, doesn't it? What about "call-wrapper" or "wrapped-call"? Another idea would be "indirect-call".

clbr:
> I haven't tried but from my perspective it would be appropriate to use #pragma code-name ("DATA") for the trampoline, rely on copydata to place it in RAM and just use the function name as if it were in ROM.

This does work on default cc65, just tested. However, on NES there may be more than one RAM area, and not all init code calls copydata. So I'd like to keep the array option, as that allows explicit positioning and works with all kinds of init code.

> Can you point out the code in your pull request that is necessary to make the "array hack" work?
https://github.com/cc65/cc65/pull/424/commits/2890b3a810219ad512f9fdd2432811f94c0b121e#diff-a81901b4930a088e0d6aa3c999b5bd73R512
The STORAGE option allows arrays. So less than one line of code for that functionality.

I don't like indirect-call, as that has asm implications. Wrapped-call sounds fine, shall I go with that?

greg-king5:
I prefer "call-wrapper" because it's what that pragma creates.  A "wrapped-call" is the Assembly code that the created object generates when cc65 sees an appropriate function call (in a source file).

oliverschmidt:
> A "wrapped-call" is the Assembly code that the created object generates when cc65 sees an appropriate function call (in a source file).

I see your point (at least I think so) but then "fastcall" would be wrong too. I think one needs to interpret both as "prepare the subject (here: the function) to be used (here: to be called) in that way".

oliverschmidt:
>> [...] #pragma code-name ("DATA")  [...]

> This does work on default cc65, just tested.

Thanks for verifying :-)

> [...] works with all kinds of init code.

This isn't necessarily a good argument as one of the differences between Asm and C is the type safety of C which we break here deliberately.

> The STORAGE option allows arrays. So less than one line of code for that functionality.

I see. So that for sure isn't an issue. However I'm still unsure because the distinction between data pointers and code pointers is general principle in C.

> I don't like indirect-call, as that has asm implications.

That appeared to me too.

> Wrapped-call sounds fine, shall I go with that?

Yep.

greg-king5:
You reminded me of something.  Didn't Uz write in an email that he wanted to use `__attribute__()`, instead of `#pragma`, for this sort of thing?

mrdudz:
On Thursday 11 May 2017, 15:12:22 greg-king5 <notifications@github.com> wrote:
> You reminded me of something.  Didn't Uz write in an email that he wanted to
> use `__attribute__()`, instead of `#pragma`, for this sort of thing?

that'd make sense... at least eg GCC does it like that too

-- 

Universitäten sind schöne Misthaufen, auf denen gelegentlich einmal eine edle 
Pflanze gedeiht. 
<Albert Einstein>

clbr:
I specifically used pragma instead of attribute because it's very
annoying to type/copypaste the thing into 100 functions, let alone when
you have to change it.

GCC has a similar pragma for visibility, in addition to the attribute,
for precisely that reason.

oliverschmidt:
> Didn't Uz write in an email that he wanted to use __attribute__(), instead of #pragma, for this sort of thing?

Thanks for the reminder.

> I specifically used pragma instead of attribute because it's very
annoying to type/copypaste the thing into 100 functions, let alone when
you have to change it.

I tend to agree with that. I imagine the pragma wrapped-call being used together with the pragma code-name. The latter wouldn't be fun as attribute too.

oliverschmidt:
I still dislike that "array hack" for multiple reasons so I thought about a more cc65'ish way of doing it (not actually tested):

There's a custom linker config for sure anyway so one can add a specific segment - named TRAMPOLINE ;-) - with `load` from ROM, `run` in RAM and `define = yes`.

Then one can have `extern void _TRAMPOLINE_LOAD__[], _TRAMPOLINE_RUN__[], _TRAMPOLINE_SIZE__[];` and `memcpy(_TRAMPOLINE_LOAD__[], _TRAMPOLINE_RUN__[], _TRAMPOLINE_SIZE__[])`.

One could even go one step further: If one the NES there's one (or not more than a handful of) possible actions a trampoline that needs to sit in RAM could have to do, then that (or those) trampoline(s) could be part of the NES C library. They could assume that there's a segment named TRAMPOLINE, be written in "pure" Asm and have a `.constructor`, that does the necessary copy of the TRAMPOLINE segment.

clbr:
> `memcpy(_TRAMPOLINE_LOAD__[], _TRAMPOLINE_RUN__[], _TRAMPOLINE_SIZE__[])`.

Swapped dest and source here?

> One could even go one step further: If one the NES there's one (or not more than a handful of) possible actions a trampoline that needs to sit in RAM could have to do, then that (or those) trampoline(s) could be part of the NES C library. They could assume that there's a segment named TRAMPOLINE, be written in "pure" Asm and have a `.constructor`, that does the necessary copy of the TRAMPOLINE segment.

Alas, every mapper has a different way of switching banks.

oliverschmidt:
>> `memcpy(_TRAMPOLINE_LOAD__[], _TRAMPOLINE_RUN__[], _TRAMPOLINE_SIZE__[])`

> Swapped dest and source here?

Yes.

clbr:
I quoted from the email I got. This thread has a lot of edited replies, which I didn't even see, as I followed via email.

src/cc65/pragma.c
```diff
@@ -123,6 +125,7 @@ static const struct Pragma {
     { "static-locals",          PRAGMA_STATIC_LOCALS      },
     { "staticlocals",           PRAGMA_STATICLOCALS       },      /* obsolete */
     { "warn",                   PRAGMA_WARN               },
+    { "wrapped-call",		PRAGMA_WRAPPED_CALL	  },
```
oliverschmidt: Please adjust alignment.

oliverschmidt:
Sorry if there was a misunderstanding: My idea was to not only rename the pragma itself but also "everything else": C functions, source file names. The term 'trampoline' should only appear as use case in the docs.

Additionally I'd like to learn if there's something wrong with my `extern void _TRAMPOLINE_LOAD__[], _TRAMPOLINE_RUN__[], _TRAMPOLINE_SIZE__[]; memcpy(_TRAMPOLINE_RUN__, _TRAMPOLINE_LOAD__, _TRAMPOLINE_SIZE__);` approach. If there isn't then I ask you to remove that "storage" support and adjust the docs - just because the trampoline must live in RAM that doesn't mean it has to be hack using "pseudo-data". We have i.e. self-modifying code in the C library that is always copied to RAM. That's what the LOAD vs. RUN location support in ld65 is for.

If you want can add a directory 'samples/nes' and put a real world sample for a RAM trampoline there showing how to actually do it. Or you can create a page in the Wiki explaining it.

src/cc65/pragma.c
```diff
@@ -125,7 +125,7 @@ static const struct Pragma {
     { "static-locals",          PRAGMA_STATIC_LOCALS      },
     { "staticlocals",           PRAGMA_STATICLOCALS       },      /* obsolete */
     { "warn",                   PRAGMA_WARN               },
-    { "wrapped-call",		PRAGMA_WRAPPED_CALL	  },
+    { "wrapped-call",           PRAGMA_WRAPPED_CALL	  },
```
oliverschmidt: Please align the closing brace too ;-)

clbr:
They look exactly the same in my editor, only github marks tabs as 7 spaces. :p

src/cc65/pragma.c
oliverschmidt: Good that we follow up on this: Apart from Makefiles there should be no tab (ascii 9) char in any cc65 source file! See https://github.com/cc65/wiki/wiki/C-and-Assembly-sources-style-guide

src/cc65/funcdesc.h
```diff
@@ -67,6 +67,8 @@ struct FuncDesc {
     unsigned            ParamCount;     /* Number of parameters              */
     unsigned            ParamSize;      /* Size of the parameters            */
     struct SymEntry*    LastParam;      /* Pointer to last parameter         */
+    struct SymEntry*    WrappedCall;     /* Pointer to the WrappedCall         */
```
oliverschmidt: Please adjust alignment of comment-end.

src/cc65/pragma.c
```diff
@@ -446,6 +449,84 @@ static void SegNamePragma (StrBuf* B, segment_t Seg)
...
+    /* Check if the name is valid */
+    if (Entry && Entry->Flags & SC_FUNC) {
+
+        PushWrappedCall(Entry, Val);
+        Entry->Flags |= SC_REF;
+
+    } else {
+
+        /* Segment name is invalid */
+        Error ("Wrapped-call target does not exist or is not a function or array");
```
oliverschmidt: Please adjust text to removal of storage support.

test/val/trampoline.c
```diff
@@ -0,0 +1,49 @@
+/*
+  !!DESCRIPTION!! trampoline pragma
```
oliverschmidt: Please adjust to recent rename.

oliverschmidt:
Thanks for your cooperation :-)

PR #425, opened 2017-05-06
Add fast paths for char postinc/dec

clbr:


oliverschmidt:
While I imagine that char postinc/postdec is already present in the test programs I'm wondering if it wouldn't nevertheless be a good idea to add an explicit test program just to make sure that change works as expected and that there are no future regressions.

test/val/postincdec.c
```diff
@@ -0,0 +1,23 @@
...
+	val = 0;
+	array[0] = array[1] = 10;
+
+	array[val++] = 2;
+	array[val++] = 2;
+	--val;
+	array[val--] = 0;
+	array[val--] = 0;
+
+	return (array[0] == array[1] && array[0] == 0) ? 0 : 1;
```
greg-king5: I think that the line should include a test of the final value of the `val` variable:
```C
&& val == 0xFF
```

oliverschmidt:
All wishes have been taken care of and there are no additional comments recently - time to merge.

PR #429, opened 2017-05-17
Optimize the inlined strlen/memcpy/memset

pfusik:


clbr:
The first one looks like it'd break the branch? bne only works on the A register.

clbr:
Sorry, ldx does set the Z flag, nevermind.

src/cc65/stdfunc.c
```diff
@@ -284,7 +284,7 @@ static void StdFunc_memcpy (FuncDesc* F attribute ((unused)), ExprDesc* Expr)
             Label = GetLocalLabel ();
 
             /* Generate memcpy code */
-            if (Arg3.Expr.IVal <= 127) {
+            if (Arg3.Expr.IVal <= 129) {
```
oliverschmidt: Please explain a bit why this can be changed from 127 to 129.

pfusik: These conditions are for loops like:

         ldy #IVal-1
    loop ...
         dey
         bpl loop

They will work for `ldy #128` and lower, hence IVal<=129.

oliverschmidt: Thanks for the explanation :-)

PR #430, opened 2017-05-18
VIC's kernal does not support CTRL+[

polluks:


doc/vic20.sgml
```diff
@@ -188,7 +188,7 @@ No VIC1011 drivers are currently available for the VIC20.
 
 <sect1>Escape code<p>
 
-For an Esc, press CTRL and the <tt/[/ key.
+For an Esc, pressing CTRL and the <tt/[/ key is not supported.
```
greg-king5: > In case I'm right, I think the text should rather read "Entering an ESC is not possible." or alike.

It's "worse" than that.  The VIC-20 doesn't have a traditional control key.  CTRL-R is the only traditional key combo that produces anything.  On the VIC-20, the CTRL key is used with the number keys to change the cursor's color.

So ...
that "Escape code" subsection should be removed; and, a note should be put in the "Limitations" section.  It should say that the `CTRL` key can't be used to type most control characters.

doc/vic20.sgml
```diff
@@ -219,8 +219,7 @@ The program return code (low byte) is passed back to BASIC by use of the
 
 <sect1>Using extended memory<p>
 
-The extended memory at $A000 may be added to the heap by using the following
-code:
+BLK5 memory may be added to the heap by using the following code:
```
polluks: It's common practice, see [http://sleepingelephant.com/ipw-web/bulletin/bb/search.php?keywords=BLK5](url).

oliverschmidt: @silverdr: I initially thought the same way too but the $A000 appears right below in the source code so one can easily make the connection between the two terms "$A000" and "BLK5". And Googling them together (even without "VIC20") work just fine. So from that perspective I considered polluks' change as an improvement even for non VIC20 experts (like i.e. me).

PR #433, opened 2017-05-19
Trampoline fixes

### LINKED ISSUE #432: #pragma wrapped-call fails on fastcall with parameters and -O.
Functions with `__cdecl__` / `--standard` seem to be fine, since they do not use the registers to pass arguments, but the new wrapped-call feature seems to break the default `__fastcall__` convention when optimizations (`-O`) are used:
```
void trampoline(void);
#pragma wrapped-call (push, trampoline, 32)
int func_c(long arg)
{
   return arg +3;
}
#pragma wrapped-call (pop)

; without -O func_c(7) produces:
   ldx     #$00
   stx     sreg
   stx     sreg+1
   lda     #$07
   ldy     #32
   sty     tmp4
   ldy     #<(_func_c)
   sty     ptr4
   ldy     #>(_func_c)
   sty     ptr4+1
   jsr     _trampoline

; with -O func_c(7) produces:
   ldy     #32
   sty     tmp4
   ldy     #<(_func_c)
   sty     ptr4
   ldy     #>(_func_c)
   sty     ptr4+1
   jsr     _trampoline
```
Notice that somehow the optimizer had completely omitted the argument setup. It seems like it's not properly connected to the wrapped-call and it just drops them?

There is a similar problem with variadic functions. Since the wrapped-call preamble clobbers Y, they lose the information about the stack size that should be passed in Y. (Same for functions without prototype, but that's an obscure case.)

While I'm asking about this feature, I think the document could use some more description of the responsibilities of the trampoline function since it is a bit intimately tied with the calling conventions e.g. it will likely have to preserve arguments/return values from A/X while it does its bankswitching...

I left some notes about that at the NESDev forum here:
[https://forums.nesdev.com/viewtopic.php?p=196106#p196106](https://forums.nesdev.com/viewtopic.php?p=196106#p196106)

clbr:
Fixes #432.

### DIFF
--- a/src/cc65/codeinfo.c
+++ b/src/cc65/codeinfo.c
@@ -403,7 +403,8 @@ void GetFuncInfo (const char* Name, unsigned short* Use, unsigned short* Chg)
                 *Use = REG_Y;
             } else if (D->Flags & FD_CALL_WRAPPER) {
                 /* Wrappers may go to any functions, so mark them as using all
-                  registers */
+                ** registers.
+                */
                 *Use = REG_EAXY;
             } else if (D->ParamCount > 0 &&
                        (AutoCDecl ?
--- a/src/cc65/codeinfo.c
+++ b/src/cc65/codeinfo.c
@@ -401,6 +401,10 @@ void GetFuncInfo (const char* Name, unsigned short* Use, unsigned short* Chg)
             */
             if ((D->Flags & FD_VARIADIC) != 0) {
                 *Use = REG_Y;
+            } else if (D->Flags & FD_CALL_WRAPPER) {
+                /* Wrappers may go to any functions, so mark them as using all
+                  registers */
+                *Use = REG_EAXY;
             } else if (D->ParamCount > 0 &&
                        (AutoCDecl ?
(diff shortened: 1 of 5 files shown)


src/cc65/codeinfo.c
```diff
@@ -401,6 +401,10 @@ void GetFuncInfo (const char* Name, unsigned short* Use, unsigned short* Chg)
             */
             if ((D->Flags & FD_VARIADIC) != 0) {
                 *Use = REG_Y;
+            } else if (D->Flags & FD_CALL_WRAPPER) {
+                /* Wrappers may go to any functions, so mark them as using all
+                  registers */
```
oliverschmidt: Please use the predominant multi-line comment style.

test/val/trampoline-varargs.c
```diff
@@ -0,0 +1,48 @@
+/*
+  !!DESCRIPTION!! wrapped-call pragma w/ varags
```
greg-king5: I think that the line should say,
```
!!DESCRIPTION!! wrapped-call pragma w/ variadic function
```

PR #435, opened 2017-05-20
Add more info to the docs about "#pragma wrapped-call".

greg-king5:
- The documents will show that `push` must be used verbatim.  It isn't optional; and, nothing must be sustituted for it.
- The documents will describe the circumstances when the wrapper function must protect the CPU registers.
- The documents will show how the wrapper should call the wrapped function.


oliverschmidt:
Thanks :-)

PR #436, opened 2017-05-23
Small improvement to ca65 macro documentation

bbbradsmith:
Adding a brief note about how parentheses are problematic with ca65 macros, and replacing an unclear example with one about accidental parentheses.

### DIFF
--- a/doc/ca65.sgml
+++ b/doc/ca65.sgml
@@ -4282,6 +4282,12 @@ different:
 	some things may be done with both macro types, each type has special
 	usages. The types complement each other.
 
+<item>	Parentheses work differently from C macros.
+	The common practice of wrapping C macros in parentheses may cause
+	unintended problems here, such as accidentally implying an
+	indirect addressing mode. While the definition of a macro requires
+	parentheses around its argument list, when invoked they should not be included.
+
 </itemize>
 
 Let's look at a few examples to make the advantages and disadvantages
@@ -4314,19 +4320,17 @@ Macros with parameters may also be useful:
         DEBUG   "Assembling include file #3"
 </verb></tscreen>
 
-Note that, while formal parameters have to be placed in braces, this is
-not true for the actual parameters. Beware: Since the assembler cannot
-detect the end of one parameter, only the first token is used. If you
-don't like that, use classic macros instead:
+Note that, while formal parameters have to be placed in braces,
+the actual parameters used when invoking the macro should not use braces.
+The invoked parameters are separated by commas only, if parentheses are
+used by accident they will become part of the replaced token:
 
 <tscreen><verb>
-.macro  DEBUG   message
-        .out    message
-.endmacro
+.define COMBINE(ta,tb,tc) ta+tb*10+tc*100
+
+        COMBINE 5,6,7     ; 5+6*10+7*100 = 765 correct
+        COMBINE(5,6,7)    ; (5+6*10+7)*100 = 7200 incorrect!
 </verb></tscreen>
-
-(That is an example where a problem can be solved with both macro types).
-
 
 <sect1>Characters in macros<p>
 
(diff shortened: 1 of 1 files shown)


nicklausw:
So is it considered safe to use C-style macros in expressions at all? Situations like `lda #get 0,1,2 + 3` where the + 3 is supposed to be added to the result of a 3-arg function named 'get' would probably cause problems, because the parser would probably just read `lda #get 0,1,5`. Would it be more safe to say `lda #get{0,1,2} + 3`?

doc/ca65.sgml
```diff
@@ -4314,19 +4320,17 @@ Macros with parameters may also be useful:
...
-</verb></tscreen>
-
-(That is an example where a problem can be solved with both macro types).
+.define COMBINE(ta,tb,tc) ta+tb*10+tc*100
 
+        COMBINE 5,6,7     ; 5+6*10+7*100 = 765 correct
+        COMBINE(5,6,7)    ; (5+6*10+7)*100 = 7200 incorrect!
+</verb></tscreen>
 
 <sect1>Characters in macros<p>
```
greg-king5: We want to have two blank lines between `<sect1>` subsections.  (See [this wiki article](https://github.com/cc65/wiki/wiki/LinuxDoc-sources-style-guide).)

greg-king5:
> So, is it considered safe to use C-style macros in expressions, at all?  Situations like `lda #get 0,1,2 + 3` where the `+ 3` is supposed to be added to the result of a 3-arg. function named "get" probably would cause problems because the parser probably would just read `lda #get 0,1,5`. Would it be safe to say `lda #get{0,1,2} + 3`?

Nope, that would give only one argument to "get".  But, it's the right idea -- just move the left brace over to the last argument:
```asm
lda #get 0, 1, {2} + 3
```

bbbradsmith:
Hm, there is more to this than I thought. I'd never seen {} in ca65 code before. Do they work like parentheses, except without the unfortunate overload for indirect addressing? I asked a question about this a while back on the mailing list but nobody mentioned {} as a possible solution. !! I notice a few of the directives mention curly braces in their description, but there's no explicit description of them as an element of syntax. What do they do?

I'm also now confused about how it takes arguments when invoking. Except for the last argument I presumed it picks up everything before the comma, but does the last argument continue to the end of the line, or does it end earlier than this?

Clearly I need to do a bit of research and revision before this is a viable pull request. It doesn't seem like these things work quite like I believed.

bbbradsmith:
Ah, I think I understand now. {} can be used to encapsulate inputs to a directive or macro? Thus they're really only useful on the last argument of a macro invocation as a way to mark the end of that argument, right?

bbbradsmith:
Have added some more examples and a little more text to try and cover the common caveats about these things. (Also tried to meet the style guideline linked above.)

oliverschmidt:
@Greg: I see that you commented on this pull request before it was re-opened. Did you have an opportunity to check its current state? Should it be merged from your point of view?

greg-king5:
> @Greg: I see that you commented on this pull request before it was re-opened. Should it be merged, from your point of view?

Yes , I'm satisfied.

greg-king5:
> Ah, I think I understand now. `{}` can be used to encapsulate inputs to a directive or macro? Thus, they're really only useful on the last argument of a macro invocation, as a way to mark the end of that argument, right?

That's an added bonus.  The original purpose is to allow us to put commas into arguments.  A simplistic example:
```asm
.macro stxa param
    sta param
    stx 1 + param
.endmac

    stxa var  ; single absolute argument
    stxa var,y  ; two arguments
    stxa {var,y}  ; single indexed argument
```

oliverschmidt:
@greg-king5: Thanks for your quick feedback.

@bbbradsmith: Thanks for your contribution.

PR #440, opened 2017-05-25
Also Oricutron mentioned

polluks:


doc/ld65.sgml
```diff
@@ -247,10 +247,11 @@ Here is a description of all of the command-line options:
   <tag><tt>-Ln</tt></tag>
 
   This option allows you to create a file that contains all global labels and
-  may be loaded into the VICE emulator using the <tt/ll/ (load label) command. You
+  may be loaded into the VICE emulator using the <tt/ll/ (load label) command
+  or into the Oricutron emulator using the <tt/sl/ (symbols load) command. You
   may use this to debug your code with VICE. Note: Older versions had some
   bugs in the label code. If you have problems, please get the latest <url
-  url="http://vice-emu.sourceforge.net/" name="VICE"> version.
+  url="http://vice-emu.sourceforge.net" name="VICE"> version.
```
spiro-trikaliotis: Why did you remove the trailing slash from the URL? Just because the browsers are fault tolerant and add the slash if it is missing, the correct URL must have the slash. In fact, you changed the URL so that it is not a valid URL anymore.

polluks: URL normalization, see https://www.npmjs.com/package/normalize-url
of course we not talking about trailing slashes of directories.

spiro-trikaliotis: Interesting. I wanted to say that RFC 3986 changed the behaviour of RFC 1738, but I am wrong here, too (3.3):

If neither <path> nor <searchpart> is present, the "/" may also be omitted.

So, I stand corrected.

silverdr: Yeah - it's "always" been valid. The way we used to use it with trailing slash comes AFAIR from the fact (or myth - don't remember now exactly) that browsers were making two requests instead of one when there was no trailing slash due to... something. So we learned to optimise it with telling them explicitly that we want the "DocumentRoot" being served by always adding the trailing slash.

oliverschmidt: Sorry but I'm not good at reading (not even talking about understanding) RFCs. Are we talking here "only" about the slash at the end of the host name/addr? Because in "http://my.host/my/path/" I thought the trailing slash would be important to make the relative URL "mydoc.html" being requested as "http://my.host/my/path/mydoc.html" and not as ""http://my.host/my/mydoc.html", or am I completely missing the point here?

silverdr: It is about the trailing slash after the FQDN (aka host name). It's optional (implicitly asking for DocumentRoot's index) when there is neither path nor query but we (at least Spiro and me :-) got used to always add it. So much that an URL without one looks wrong to us. If there is path, then the trailing slash indicates the request for default (index) document at given path. Anyway - it is not incorrect and even if we don't like this "normalisation" (as much as everything that smells node.js :-) we can't argue about it being a violation of the appropriate RFCs.

oliverschmidt: > It is about the trailing slash after the FQDN (aka host name).

I see :-)

> but we (at least Spiro and me :-) got used to always adding it.

Me too ;-)

spiro-trikaliotis: IMHO, it belongs there.
Note also that normalization does not require to remove the trailing slash. Normalization can also mean that the slash is always added, for more consistency. ;)

PR #443, opened 2017-05-27
More POSIX.2, staying compatible

polluks:
The old syntax was "ar65 v l foo.lib", now "ar65 tv foo.lib" also works (maybe ar's syntax was always a bit crude because of its age).

### DIFF
--- a/src/ar65/main.c
+++ b/src/ar65/main.c
@@ -62,7 +62,7 @@ static void Usage (void)
 {
     fprintf (stderr, "Usage: %s <operation ...> lib file|module ...\n"
             "Operations are some of:\n"
-            "\ta\tAdd modules\n"
+            "\tr\tAdd modules\n"
             "\td\tDelete modules\n"
             "\tt\tList library table\n"
             "\tv\tIncrease verbosity (put before other operation)\n"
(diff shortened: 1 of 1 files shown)


greg-king5:
If you are going to change the way that verbosity is set, then you need to change it for all of the operations.  Currently, the Add-object-file operation responds to one and two Vs.

polluks:
Well, in the past ar's verbosity was a dummy and the POSIX syntax not really fits in a single switch. However, I don't want to break compatibility.

greg-king5:
The problem is that you changed the usage message.  You say that `v` should be put _after_ the other operation.  That new position will work for the `t` operation; but, it won't work for the `a Add modules` operation!  The `a` operation still needs `v` to be put _before_ the `a`.  Both
```sh
ar65 v a foo.lib bar.obj baz.obj
```
and
```sh
ar65 v v a foo.lib bar.obj baz.obj
```
show additional progress messages.  [The functions `LibOpen()` and `LibCheckExports()` call `Print()`.]

polluks:
I see, it's smarter than I thought...

spiro-trikaliotis:
So, now we have two options that are stll there for backward compatibility, but they are not documented anymore. Wouldn't it be better to document them, but mark them as deprecated and refer to the POSIX style options?

polluks:
Ok, it's in the document.

spiro-trikaliotis:
I would expect the old commands to be mentioned in the list of operations, too, with a remark that they are deprecated.

PR #446, opened 2017-06-06
Add LZ4 decompression support

clbr:
LZ4 is a fast compression format. While zlib is my usual go-to, the RAM requirements for it are too high for many NES games. LZ4 does not require any RAM.


oliverschmidt:
- Please remove all tab ($8) chars from cc65 source files.
- Looking at the label names and the general layout the source file looks like generated from the compiler. I guess you are aware that C files are allowed in the C library too.

clbr:
The assembly has been heavily hand-optimized. It's almost 7 times faster, uses no RAM (vs 14 bytes BSS), and takes 1/4 the space.

include/lz4.h
```diff
@@ -0,0 +1,44 @@
...
+#ifndef _LZ4_H
+#define _LZ4_H
+
+void __fastcall__ decompress_lz4 (const unsigned char* src, unsigned char* dst,
+                                  const unsigned short uncompressed_size);
+/*
+   Decompresses the source buffer into the destination buffer.
+   The size of the decompressed data must be known in advance, LZ4
+   does not include any terminator in-stream.
+*/
```
oliverschmidt: Please adjust this multi-line comment to match the style used elsewhere in this repo.

libsrc/common/lz4.s
```diff
@@ -0,0 +1,283 @@
...
+_out = regsave
+_written = regsave + 2
+_tmp = tmp1
+_token = tmp2
+_offset = ptr3
+_in = sreg
+_outlen = ptr4
+
+; ---------------------------------------------------------------
+; void decompress_lz4 (const u8 *in, u8 * const out, const u16 outlen)
```
oliverschmidt: The second const here doesn't match the declaration in the C header file.

libsrc/common/lz4.s
```diff
@@ -0,0 +1,283 @@
...
+        .import         memcpy_upwards,pushax,popax
+        .export         _decompress_lz4
+
+_out = regsave
+_written = regsave + 2
+_tmp = tmp1
+_token = tmp2
+_offset = ptr3
+_in = sreg
+_outlen = ptr4
```
oliverschmidt: Is there a specific reason why you added an '_' prefix to all the symbols? An '_' prefix sort of implies the intend the make something visible from C - which isn't the case here.

libsrc/common/lz4.s
```diff
@@ -0,0 +1,283 @@
...
+        lda     #$00
+        sta     _written
+;
+; while (written < outlen) {
+;
+        jmp     L0046
+;
+; token = *in++;
+;
+L0004:  ldy     #0
```
oliverschmidt: Please decide on a constant style. Either '#$00' or '#0'.

oliverschmidt:
> The assembly has been heavily hand-optimized. It's almost 7 times faster, uses no RAM (vs 14 bytes BSS), and takes 1/4 the space.

Thanks for the explanation. Maybe you want to add those facts to some comment in the assembler source file...

oliverschmidt:
Are you aware of http://pferrie.host22.com/misc/appleii/lz4unp02.zip ?

oliverschmidt:
...and/or http://xxl.atari.pl/lz4-decompressor/ ?

include/lz4.h
```diff
@@ -0,0 +1,44 @@
...
+/*                                                                           */
+/*****************************************************************************/
+
+
+
+#ifndef _LZ4_H
+#define _LZ4_H
+
+void __fastcall__ decompress_lz4 (const unsigned char* src, unsigned char* dst,
+                                  const unsigned short uncompressed_size);
```
oliverschmidt: Looking at https://github.com/lz4/lz4/blob/master/lib/lz4.h I suspect that it doesn't make sense to try to map this this to something there. However there they use 'LZ4_...'. Maybe you want to change you '..._lz4' to that style.

oliverschmidt:
I like that you added a test program. However I'd really appreciate if you would add docs too. Things like http://cc65.github.io/doc/funcref.html#ss2.49 are bad enough ;-)

clbr:
I knew of the Apple II version, it wasn't usable as it made some platform-specific assumptions, and starting from scratch was faster than trying to understand it. I also knew of the 65816 version, which used more advanced instructions. Did not know about that Polish one, it seems to use 65c02 instructions, as well as likely to be much slower with all those jumps.

clbr:
I didn't rename the function, as no other func in cc65 has such capitals, and this is a custom one anyway. It most closely matches LZ4_decompress_fast, but as it doesn't return a value, it's not safe to use that name.

libsrc/common/lz4.s
```diff
@@ -0,0 +1,286 @@
...
+; copysrc = out + written - offset;
+;
+L002F:  lda     out
+        clc
+        adc     written
+        pha
+        lda     out+1
+        adc     written+1
+        tax
+        pla
```
greg-king5: You can store the low byte temporarily in the .Y index register instead of the stack.

greg-king5:
> Did not know about that Polish one; it seems to use 65c02 instructions.

Do you intend to add any 65sc02 codes to the function?

clbr:
> Do you intend to add any 65sc02 codes to the function?

No. I only care about the NES.

libsrc/common/lz4.s
clbr: 3+4 cycles would be replaced by 2+2+2 cycles, as there would also need to be a ldy #0 before the next memcpy_upwards call (it assumes y is zero on entry). Not sure if worth it?

greg-king5: Both calls to `memcpy_upwards` are preceeded by calls to `pushax`; it's guaranteed to return with a zero in the .Y register.  Therefore, `ldy #$00` isn't needed before either call (comment out line 140 -- with a comment about why it can be a comment).

oliverschmidt:
>> Do you intend to add any 65sc02 codes to the function?

> No. I only care about the NES.

;-)

oliverschmidt:
@clbr Thanks for answering all questions and taking care of all comments.

@greg-king5 Thanks for the optimization suggestions.

pfusik:
> Did not know about that Polish one, it seems to use 65c02 instructions, as well as likely to be much slower with all those jumps.

It's plain 6502, co-authored by me. `inw foo` is short for

        inc foo
        bne skip
        inc foo+1
    skip

as supported by [xasm](https://github.com/pfusik/xasm)

IrgendwerA8:
Would like to see a size/speed comparison to have the best implementation in cc65...

PR #447, opened 2017-06-07
Add a missing -O configuration to some regression-test make-files.

greg-king5:
I created this PR completely in Github's web interface.  (That's why there are three commits for the three files -- one commit would have been enough if I had done it in my local computer's Git.)


oliverschmidt:
Nice find :-)

PR #448, opened 2017-06-14
cpeekc function for c64

MonteCarlos:
Small assembly cpeekc implementation for c64 including test program.
The description of the -E flag is originally not related to this pull request. However i had a merge conflict in cl65.sgml with this description. I chose the one which coincided with the cc65 documentation. Hopefully this agrees with other intention.


libsrc/c64/cpeekc.s
```diff
@@ -0,0 +1,29 @@
+;
+; Ullrich von Bassewitz, 1998-08-06, 2009-09-26
```
mrdudz: Uz should probably get removed here :)

mrdudz:
the mix with the other file is a bit ... :)

as for the cpeekc, i almost prefer this over mine :) this very same routine should work for most other cbm targets too (minus c128, cbm2.. see my other pull request)

MonteCarlos:
I see that the ldx #0 is obligatory. I will include that.
I'll have a look on the preamble, too. Sorry for missing that.

mrdudz:
btw, perhaps you missed the post on the ml - gregg already merged this (and my stuff) together and worked more on it - this PR can probably be closed in favour of his work (i did the same with mine).

oliverschmidt:
> btw, perhaps you missed the post on the ml - gregg already merged this (and my stuff) together and worked more on it - this PR can probably be closed in favour of his work (i did the same with mine).

I guess you are refering to https://github.com/greg-king5/cc65/commit/44cc7632eba073bda0cb821bb943b466370724e0 - correct?

@greg-king5: Could you please create a pull request to the cc65 upstream repo so we can discuss things here? Or is there a specific reason not to do so that I don't see?

greg-king5:
> greg-king5: Could you please create a pull request to the cc65 upstream repo, so we can discuss things here? Or, is there a specific reason not to do so that I don't see?

My only reason is that I usually don't create a PR until I think that I have finished committing things to it.

oliverschmidt:
Maybe you can make an exception here: I think it would be beneficial if we could discuss "general, conceptual" aspects that nevertheless come up when when looking at actual code. An example: I seem to remember that I learned that a "correct" PETSCII<->ASCII conversion is done with tables. However the last cpeekc() code that I'm aware of does that offset-by-$20 thingy.

polluks:
What is lynx-comlynx.s doing in this commit?!

oliverschmidt:
@greg-king5 Do you have an estimate for us when you will be willing to share your work for this topic in one way or another?

mrdudz:
@greg-king5 yes please push that peekc stuff a bit! i need to finally make my tetris thingy work with a current cc65 :)

oliverschmidt:
> my tetris thingy work with a current cc65

[pretty off topic]
I presume that Tetris depends on clock() because the blocks must fall in some give speed. Unfortunately this means that it's not portable to the Apple II. However there's at least a little light ahead. There are several - incompatible - hardware clocks for the Apple II and the ProDOS OS uses them to stamp file access. ProDOS also provides an API to get the current time but that delivers only minutes. But someone started to create new versions of ProDOS - and one idea for an upcoming version is to provide an extend current time API with seconds and milliseconds. That would mean that with _any_ hardware clock and that new ProDOS a clock() implemenation is supposed to be pretty straight forward :-)
[/pretty off topic]

polluks:
Well, you don't need a full clock but an IRQ Manager
http://mirrors.apple2.org.za/ground.icaen.uiowa.edu/MiscInfo/Hardware/interrupt.card

oliverschmidt:
> Well, you don't need a full clock but an IRQ Manager

Surely but those devices are extremely rare so it doesn't help at all. Having a mouse card would provide the same type of vsync interrupt source but nobody cared enough until today to provide an Apple II clock() implementation based on vsync interrupt counting. While surely not necessary for Tetris but in general such an implementation would need to take care of 50Hz vsync vs 60Hz vsync which would need to be detected by cycle counting - which would need to be implemented to take care to temporarily turn off all types of accelerators...

I think by now it's clear why I consider an OS call to get the wall clock time in millisecond resolution beneficial ;-))

mrdudz:
i am closing this - https://github.com/cc65/cc65/pull/545 makes it obsolete

oliverschmidt:
Thanks.

PR #450, opened 2017-06-15
Add a missing label to some conio cursor movement functions.

greg-king5:
I noticed that @Fabrizio-Caruso had seen the linking problem while he was developing his https://github.com/Fabrizio-Caruso/ASCII-CHASE/commit/e5e35191ccfb547f4e97ecac94935bd5b5c96362 game.


oliverschmidt:
I see that `gotoxy` is imported by conio/cputs.s so cputs() - and not only cputsxy() - can never have worked on the targets in question. Maybe they are all using cprintf() even for plain string constants...

PR #452, opened 2017-06-16
kbrepeatdelay and kbrepeatrate for cbm targets

mrdudz:
quick PR to get things going... (not really ready yet)

since it came up on the list recently - do we want these functions? :) are the names OK or not? should prototypes go into conio.h ?


asminc/plus4.inc
```diff
@@ -33,6 +33,11 @@ FKEY_COUNT      := $55D         ; Characters for function key
 FKEY_SPACE      := $55F         ; Function key definitions
 FKEY_ORIG       := $F3D2        ; Original definitions
 
+;FIXME: he?! these ok? :o)
```
polluks: No, they are $540, $541 and $542.

asminc/pet.inc
```diff
@@ -31,6 +31,11 @@ BASIC_BUF_LEN   = 81            ; Maximum length of command-line
 
 KEY_BUF         := $26F         ; Keyboard buffer
 
+;FIXME: these are wrong?
```
polluks: Indeed.
40 columns: $3ee, $3ea, $3e9
80 columns: $e4, $e5, $e6

mrdudz:
more?

how should we handle the difference between 40- and 80- column models? they are different targets anyway, right? so the different constants could be achived using some ifdef magic? or another .inc? or what?

polluks:
No ... targtet c128 uses PRINT (jsr $C322) for 40 **and** 80, i.e. no hardware banging like c64's conio.

mrdudz:
i was thinking of the PET targets :)

polluks:
Oh I see. Check in target pet for SCR_LINELEN.

polluks:
Just for Uz...
cbm510: $D7 and $D8
http://www.von-bassewitz.de/uz/oldcomputers/p500/rom500.s.html

greg-king5:
> How should we handle the difference between 40- and 80- column PET/CBM models?

(cc65 has a single `-t pet` target.)  cputc() tests it during run-time:
https://github.com/cc65/cc65/blob/master/libsrc/pet/cputc.s#L84-L86

So, you will need to define constants like `KBDREPEAT40` and `KBDREPEAT80`, and to choose the right one, at run-time.

---
I think that it makes sense to put kbdrepeat()'s prototype in `<cbm.h>`.  The others are tick count-down timers; the Kernal changes them!

mrdudz:
> I think that it makes sense to put kbdrepeat()'s prototype in <cbm.h>.

mmmh - but that would indicate they are CBM specific (which they hopefully wont be on the long run)

polluks:
Well, the functions are something like get_tv(), for example the NES will never get a kbdrepeat().

mrdudz:
by that logic, also kbhit and cgetc (and probably more) shouldnt be in conio either :)

polluks:
I see, there's something inconsistent.
Take a look at hello.c our conio reference
`#if defined(__NES__) || defined(__PCE__) || defined(__GAMATE__)`
conio expects an input, therefore its name, but you don't have a keyboard this time.

Because many targets have an implementation, get_tv() should move to conio?
The question is whether kbdrepeat() will be useful without conio or not IMHO.

oliverschmidt:
Hi,

Maybe I don't have a (good enough) overview but...

I think this isn't a precise science. It's rather about a sense of what is appropriate.

I'd say something that is available on (almost) all targets (think keyboard) and/or can be emulated on (almost) all targets (think F-keys) belongs into a shared header structured by topics (conio, tgi, mouse, ...). Something that is available on one (or few) targets belongs into the target header(s).

Let's not discuss things here which have already found their places - even if those places might not always be optimal and/or follow the guide I laid out above.

Those keyboard repeat setting functions seem pretty target specific to me. I'd guess no non-CBM target supplies something similar. So I personally clearly see them in cbm.h.

Given that they are not considered part of conio I personally think that names starting with 'kb' are fine. I would however rather use the prefix 'kbd' instead of just 'kb'. Or are there things in the CBM world (i.e. BASIC commands or alike) that make the names suggested a premier choice?

oliverschmidt:
Ooops, didn't want to close!

greg-king5:
Also, ... portable programs that use **stdio** might want to use `kbdrepeat()` to make CBM keyboards act the way that other keyboards act.

We should have also some defined constants for `kbdrepeat()`:
```c
#define KBDREPEAT_CURSOR 0x00
#define KBDREPEAT_NONE 0x40
#define KBDREPEAT_ALL 0x80
```

oliverschmidt:
> Also, ... portable programs that use stdio might want to use kbdrepeat() to make CBM keyboards act the way that other keyboards act.

I think I understand that this statement refers to you rproposal to "set" "something" on the keyboard by default on conio initialization. There were however strong oppsite opinions.

Is it possible to elaborate on the topic in a way a non-CBM insider understand the issue? I.e. I presume that a "normal" computer in a "normal" scenario (like i.e. working in the BASIC screen editor) behaves like this: I press a key and the key is accepted once - but if I continue to hold down the key then after some delay (!) the key starts to auto repeat until I release the key.

mrdudz:
on CBM targets the regular alphanumeric keys do not have any repeat at all - only the "editor" keys (cursor, space, insert/delete). why space belongs to them, i have no idea :)

oliverschmidt:
Thanks - and do I understand it correctly that it is possible to have repeat for the "regular" keys as well? Does the repeat have the "usual" delay before starting? If both question are answered with 'yes' then why isn't it desirable to activate repeat for all keys by default (just like we i.e. switch to lower case output by default) ?

mrdudz:
yes its possible. and yes they will have the initial delay. however i dont think it should be default, since its not what a typical c64 users expects - most programs do not work this way.

oliverschmidt:
Thanks for the explanation. So it's about soft facts and meeting expectations...

greg-king5:
I'm not aware of any programs that _demand_ that the alpha-numerics _never_ repeat.  I find it difficult to imagine that a program would fail if a key was to repeat -- after we held it down for awhile.  If a program like that does exist, then it's very unusual; so, it should tell people, "don't hold down any keys".  And, it should disable repeating on keyboards that support that switch.

I still think that -- portable -- programs should be able to have some consistant expectations about the environments on which they run.  So, ... programs should expect to be able to print lower-case text, with capital letters where they're needed.  And maybe, they should expect that everybody's keyboards repeat every key.

By the way, Commodore wasn't consistant with their models!  The C128 and the Plus4, for example, do repeat every key, by default.

oliverschmidt:
I personally think that Greg's argument is valid. However to be honest it's just not important enough to me to enter a dispute about that topic.

Maybe a reasonable compromise would be to have a cc65 constructor that activates the repeat feature and that is as easily as possible to be activated by the user who wants it. Ideally from the build command line...

asminc/pet.inc
```diff
@@ -31,6 +31,18 @@ BASIC_BUF_LEN   = 81            ; Maximum length of command-line
 
 KEY_BUF         := $26F         ; Keyboard buffer
 
+;FIXME: we must somehow handle the difference between the two - how?
```
oliverschmidt: Is this pull request supposed to be merged with this FIXME still in place?

mrdudz: i don't know.... would that be acceptable? i don't like the idea a lot myself.... but i don't know to fix this either. are there preprocessor symbols for 40cols vs 80cols PETs? or even different libraries? can it be checked at runtime, and how? i have no idea about either :)

mrdudz:
added some docs. i cant test if its ok because linuxdocs fail to build the docs here, for whatever reason

oliverschmidt:
> added some docs. 

:-)

> i cant test if its ok because linuxdocs fail to build the docs here, for whatever reason

At least there are no errors/warnings according to Travis CI...

mrdudz:
no, but here there are - and when i fix them it will fail on CI :) different linuxdoc versions, i guess

greg-king5:
Those functions are fastcall.  Their prototypes must be marked with `__fastcall__`, so that they will be called properly even if `--all-cdecl` is used.  And, their `funcref` descriptions must have the usual note that says that a fastcall function must be used only in the presence of a prototype.

----
The repeat-delay and the repeat-rate variables aren't "settings".  Instead, they are active counters.  When a key is pressed, the Kernal changes them 60 times every second.  Therefore, there is no "original value" that should be restored by a program.  So, their functions should return `void`.

mrdudz:
> The repeat-delay and the repeat-rate variables aren't "settings". Instead, they are active counters. When a key is pressed, the Kernal changes them 60 times every second. Therefore, there is no "original value" that should be restored by a program. So, their functions should return void.

ewww. i could have sworn those are persistant settings. like this the two functions are actually not quite useful at all (you'd have to call them after each keyboard scanning) - i'd rather just remove them. no idea why i have even added them back then :=)

mrdudz:
mmmh *shrug*

oliverschmidt:
Sorry - what is wrong? Should I not have this merged?

mrdudz:
as gregg pointed out, the functions that are supposed to set the repeat delay and repeat rate do not work as advertised (and i see no simple way to fix that unfortunately) - so they should probably get removed before someone uses them and wonders that they dont do what they are supposed to do :=)

oliverschmidt:
Ah, now I see. You last two commits are listed below your comment staring with 'ewww' so I presumed your commits were meant to "adjust" what was "wrong". I admit that I was too busy trying to make everybody happy regarding the joystick interface to really try to understand what Greg and you were up to.

So do I get it right this time that this pull request should just have been closed instead of being merged? So a full revert is the right thing to do now?

mrdudz:
closed? no.... "someone" should delete the two useless functions, then it can be merged :)

greg-king5:
Maybe they aren't totally useless.  I haven't tested it; but, maybe they can be used to zero those counters immediately after calling `cgetc()`.  It might make the keyboard look more like a joystick; that is, we might get immediate and very rapid key repeating.

mrdudz:
that doesnt really work as intended... and resetting the live counters the hard way like that inside your keyboard polling loop.... urks. that sounds like ugly hackery to me that i dont want to encourage :)

oliverschmidt:
> "someone" should delete the two useless functions

Done, see https://github.com/cc65/cc65/commit/4aa19494f5346a79d4c940903ad93f06be79915e

mrdudz:
Am Mittwoch, 30. August 2017, 16:41:04 CEST schrieb Oliver Schmidt:
> > "someone" should delete the two useless functions
> 
> Done, see
> https://github.com/cc65/cc65/commit/4aa19494f5346a79d4c940903ad93f06be79915
> e

great :)

so ... what happened to greggs cpeekxy()? :=)

-- 

In C++, friends can access each others' private members.

PR #453, opened 2017-06-16
waitvblank for cbm targets

mrdudz:
quick PR to get things going... (not really ready yet)

despite the fact that these functions can be shortened and improved (waiting for line 0 makes more sense... oh and there are again broken cbm2 things =)) - do we want these functions? :) are the names OK or not? where should the prototype go?

### DIFF
--- a/libsrc/c128/waitvblank.s
+++ b/libsrc/c128/waitvblank.s
@@ -1,0 +1,20 @@
+
+        .export         _waitvblank
+
+        .include        "c128/c128.inc"
+
+_waitvblank:
+        lda     PALFLAG
+        beq     @ntsc
+        ldx     #(312-24)-256
+        .byte $2c
+@ntsc:
+        ldx     #(262-4)-256
+@l1:
+        lda     VIC_CTRL1
+        and     #$80
+        beq     @l1
+@l2:
+        cpx     VIC_HLINE
+        bcs     @l2
+        rts
--- a/libsrc/c128/waitvblank.s
+++ b/libsrc/c128/waitvblank.s
@@ -5,7 +5,7 @@
 
 _waitvblank:
 
-        lda     MODE
+        bit     MODE
         bmi     @c80
 
 @l1:
--- a/libsrc/c128/waitvblank.s
+++ b/libsrc/c128/waitvblank.s
@@ -1,20 +1,13 @@
 
         .export         _waitvblank
 
-        .include        "c128/c128.inc"
+        .include        "c128.inc"
 
 _waitvblank:
-        lda     PALFLAG
-        beq     @ntsc
-        ldx     #(312-24)-256
-        .byte $2c
-@ntsc:
-        ldx     #(262-4)-256
 @l1:
         lda     VIC_CTRL1
-        and     #$80
-        beq     @l1
+        bpl     @l1
 @l2:
-        cpx     VIC_HLINE
-        bcs     @l2
+        lda     VIC_CTRL1
+        bmi     @l2
         rts
--- a/libsrc/c128/waitvblank.s
+++ b/libsrc/c128/waitvblank.s
@@ -4,10 +4,22 @@
         .include        "c128.inc"
 
 _waitvblank:
+
+        lda     MODE
+        bmi     @c80
+
 @l1:
-        lda     VIC_CTRL1
+        bit     VIC_CTRL1
         bpl     @l1
 @l2:
-        lda     VIC_CTRL1
+        bit     VIC_CTRL1
         bmi     @l2
+        rts
+
+@c80:
+        ;FIXME: do we have to switch banks?
+@l3:
+        lda     VDC_INDEX
+        and     #$20
+        beq     @l3
         rts
--- a/doc/funcref.sgml
+++ b/doc/funcref.sgml
@@ -415,7 +415,7 @@ function.
 
 <!-- <itemize> -->
 <!-- <item><ref id="get_tv" name="get_tv"> -->
-<!-- <item><ref id="waitvblank" name="waitvblank"> -->
+<!-- <item><ref id="waitvsync" name="waitvsync"> -->
 <!-- </itemize> -->
 
 (incomplete)
--- a/doc/nes.sgml
+++ b/doc/nes.sgml
@@ -1,4 +1,4 @@
-<!-- <!doctype linuxdoc system> //-->
+<!doctype linuxdoc system>
 
 <article>
 
--- a/doc/nes.sgml
+++ b/doc/nes.sgml
@@ -69,8 +69,8 @@ Programs containing NES specific code may use the <tt/nes.h/ header file.
 <sect1>NES specific functions<p>
 
 <itemize>
-<item>waitvsync - wait until the start of the next frame
-<item>get_tv
+<item>waitvsync - wait until the start of the next frame</item>
+<item>get_tv</item>
 </itemize>
 
 
--- a/doc/nes.sgml
+++ b/doc/nes.sgml
@@ -1,4 +1,4 @@
-<!doctype linuxdoc system>
+<!-- <!doctype linuxdoc system> //-->
 
 <article>
 
@@ -69,7 +69,7 @@ Programs containing NES specific code may use the <tt/nes.h/ header file.
 <sect1>NES specific functions<p>
 
 <itemize>
-<item>waitvblank - wait until the start of vblank
+<item>waitvsync - wait until the start of the next frame
 <item>get_tv
 </itemize>
 
--- a/doc/pce.sgml
+++ b/doc/pce.sgml
@@ -77,7 +77,7 @@ Programs containing PCE specific code may use the <tt/pce.h/ header file.
 <sect1>PCE specific functions<p>
 
 <itemize>
-<item>waitvblank</item>
+<item>waitvsync</item>
 <item>get_tv (since all PCE systems are NTSC, this always returns TV_NTSC)</item>
 </itemize>
 
--- a/doc/supervision.sgml
+++ b/doc/supervision.sgml
@@ -66,7 +66,7 @@ Programs containing Supervision specific code may use the <tt/supervision.h/ hea
 <sect1>Supervision specific functions<p>
 
 <itemize>
-<item>waitvsync
+<item>waitvsync</item>
 </itemize>
 
 
--- a/doc/supervision.sgml
+++ b/doc/supervision.sgml
@@ -66,7 +66,7 @@ Programs containing Supervision specific code may use the <tt/supervision.h/ hea
 <sect1>Supervision specific functions<p>
 
 <itemize>
-<item>waitvblank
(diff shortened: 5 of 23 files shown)


libsrc/c128/waitvblank.s
```diff
@@ -0,0 +1,13 @@
...
+        .include        "c128.inc"
+
+_waitvblank:
+@l1:
+        lda     VIC_CTRL1
+        bpl     @l1
+@l2:
+        lda     VIC_CTRL1
+        bmi     @l2
+        rts
```
polluks: What about VDC $D600 bit 5?

MonteCarlos:
Should be

+_waitvblank:
+@l1:
+        bit     VIC_CTRL1
+        bpl     @l1
+@l2:
+        bit     VIC_CTRL1
+        bmi     @l2
+        rts

to preserve accu.

mrdudz:
"to preserve accu."
since when is that necessary for functions? :)

MonteCarlos:
Maybe someone calls that function from assembly level and favors to keep accu? Or in some future times for the optimizer?

mrdudz:
ok, anything else? :)

mrdudz:
ok :=)

polluks:
BTW you may adapt your fire.c.

MonteCarlos:
In which header files these scanline functions should be integated?

Using an existing header like cbm.h or make a new one like crt.h 
(cathode ray tube.h)?

I understand the relevance of these functions for retro computing. 
However i just cannot

think of where they belong to.

If is would be of interest i could contribute a waitrasterline (9 bit) 
and a waitnframes function.


Your Monte Carlos

oliverschmidt:
Hi,

First of all sorry for being "away" for so long!

> do we want these functions?

I'd say 'yes'. My understanding is that several cc65 targets allow them to be implemented - and their use case(s) are out of question. As far as I know there are machines which "just" have some register to be watched and some machines allow an interrupt to be generated. So a wait function seems to be the owest common denominator as one can for sure have an interrupt function set a flag polled on by the wait function.

On the Apple II the newer models have (unfortunately each a different) way to detect the vblank. So it is possible to implement a wait function for the 'apple2enh' target that works on all machines the target is intended for. It's however quite some effort so nobody should hold its breath until an Apple II implementation is available.

> are the names OK or not? 

Do I miss a point? I only see one name in question: waitvblank(). I personally always thought that 'blanking' would be the right term but comparing the number of hits in Google of 'vblank' vs. 'vblanking' shows that 'vblank' seem by far the more popular term. So I think that the name is fine. And not containing any '_' is in line with the cc65 naming style.

> where should the prototype go?

I think waitvblank() falls into the very same category as videomode(): Rather low level, not to be expected on (even nearly) all targets, related to the video dispaly. And for videomode() it was decided to duplicate it in all target header files supporting the function. So I suggest to do the same here.

Regards,
Oliver

oliverschmidt:
> Maybe someone calls that function from assembly level and favors to keep accu? Or in some future times for the optimizer?

I don't see use spending _ANY_ bytes nor cycles for that. Only if it's possible without _ANY_ overhead.

oliverschmidt:
> If is would be of interest i could contribute a waitrasterline (9 bit)
and a waitnframes function.

Thanks for the kind offer. However they seem to specific in my personal opinion.

mrdudz:
> And for videomode() it was decided to duplicate it in all target header files supporting the function. So I suggest to do the same here.

as in c64.h? or cbm.h? :)

> Thanks for the kind offer. However they seem to specific in my personal opinion.

yeah i think so too

oliverschmidt:
> as in c64.h? or cbm.h? :)

If it's available on (almost) all CBMs then CBM. Other wise on the machine header files.

In the case of "almost all" I guess  some `#ifndef` would be appropiate.

polluks:
> And not containing any '_' is in line with the cc65 naming style.

Indeed? :smile: 
I know what you mean, funcref is already prepared for `waitvblank`...

MonteCarlos:
That referred to exchanging "lda" with "bit", which is memory wise and 
cycle wise the same.

Of course impl is system dependent and not all systems have vblank in 
bit 7. However if you go on assembly level you most probably do have 
targeted a specific system.

Another thing is what we refer to as "vblank". The short interval during 
retrace or the upper/lower border area? Assuming most people do not turn 
off borders when they program "c" it would probably make much more sense 
to wait for line 250 (c64 example) instead of line 0. This would support 
the second interpretation and it would double the rastertime you have to 
do things invisibly.

Someone can tell how many rasterlines the upper/lower border area 
accounts on other systems?


Yours Monte Carlos

oliverschmidt:
> Another thing is what we refer to as "vblank". The short interval during
retrace or the upper/lower border area? Assuming most people do not turn
off borders when they program "c" it would probably make much more sense
to wait for line 250 (c64 example) instead of line 0. This would support
the second interpretation and it would double the rastertime you have to
do things invisibly.

I understand that there are two use cases for waitvblank():

1. Generally pace an animation without counting cycles and without being independent from potential accelerated CPUs and without employing a timer (which might cause additional overhead or might just not be available (i.e. on the Apple II).

2. Do screen drawing during vblank to avoid "tearing" artifacts.

For 1.) the point when waitvblank() returns is of course unimportant. But for 2.) it is desirable that it returns right after after drawing the last line so that the app has as much time as possible to draw.

Um just not enough of a CBM guy to judge what that means for the implementation of waitvblank().

polluks:
The Atmos requires a (common) hardware hack http://www.pouet.net/prod.php?which=24190

mrdudz:
perhaps the function should really be called waitvsync() instead. vsync is typicall defined as waiting for the start of a new frame - which is what this function does.

*actual* waitvblank is very system dependent, often more complicated to implement (try making one that works on all types of C64s ... and then you know why i did wait for line 0 instead) and last not least doesnt even make a lot of sense for a typical portable cc65 program. once you require to wait for the bottom of the display window to do as much calculations until the start of the next display window you have written a very target specific program anyway, and you can just as well spend three more lines of code to do the waiting for you. on many videochips its not even possible, because all you do is waiting for vsync (VDC for example, i bet there are more cc65 targets)

oliverschmidt:
So if I understand you correctly you want to say that my use case 2.) above means:

- That the wait function is hard to implement
- That it is hard for a (portable) C program to make use of it (aka draw quickly enough)

Both doesn't surprise me. So if I got things right I'd vote for:

- focus on use case 1.)
- stick to a rather simple implementation on CBM
- allow for implementations on targets which don't allow for "real" waitvblank
- call it waitvsync()

mrdudz:
yesyesyes :=)

oliverschmidt:
:-)

oliverschmidt:
First of all thanks for bringing up the question.

For cc65 I always favor consistency over backward compatibility. cc65@GitHub is for enthusiasts who (hopefully) appreciate  active development - and are willing to adjust their projects when necessary. And the others - as already said - will sty with Uz' last version forever anyway.

So the short answer is: Yes, please adjust the name of the existing implementations.

mrdudz:
ok so... mmmh :)

two things:

1) how is plus4 vs c16 handled? right now i have put a waitvsync.s only into the plus4 lib - is that the right thing to do?
2) cbm510 doesnt work yet. i am not sure how to do this. any hints? :)

polluks:
1. just an include, see get_tv.s
1. it's an ordinary vic2

mrdudz:
uhm, i should make this more clear :)

1) c16 and plus4 are mostly the same thing (a plus4 is basically a c16 with more ram). do we really have to duplicate all and every library function? that seems the wrong thing to do (and looking at the respective directories that doesnt happen ... except for some functions. for whatever reason. i am wondering why that is the case, and what functions HAVE to be duplicated and where those go that dont have to be)
2) the VICII is in another bank, so the function needs to do banking (see colors.s). i just copied the respective code, it *should* work. :)

mrdudz:
ok, that makes it ready to merge... imho :) no clue about PET or cbm610, so thats something for whoever wants to implement it :)

oliverschmidt:
> c16 and plus4 are mostly the same thing (a plus4 is basically a c16 with more ram). do we really have to duplicate all and every library function? that seems the wrong thing to do (and looking at the respective directories that doesnt happen ... except for some functions. for whatever reason. i am wondering why that is the case, and what functions HAVE to be duplicated and where those go that dont have to be)

General Remark: The current build system allows to override source files. So i.e. a source file in `common` can be overridden by a source file with the same name in `cbm`. And that one can be overridden by yet another file in (let's say) `c64`. However when introducing the that build system feature I did _not_ check all the existing source files to see where the feature could be used to reduce the number of identical files. I only used it for new files added afterwards. Maybe using that feature consequently for all the CBM related source files (for use cases like "all CBM machine but one or two can share that code") would reduce the number for source files for the `c16` and the `plus4` and thus reduce the issue discussed here. Needless to say that I'd appreciate if someone would check the CBM related source files to see what duplicates could be removed/reduced...

There are currently two mechanisms in use for such scenarios:

- If two targets are really very similiar they can just share their whole C libary sources - see https://github.com/cc65/cc65/blob/master/libsrc/Makefile#L150 for the currently three target pairs. The few differences are then to be implemented with `.ifdef`. If one of the two target has things the other just doesn't have this leads to completely `.ifdef`ed source files - see i.e. https://github.com/cc65/cc65/blob/master/libsrc/apple2/textframe.s#L10

- If two targets share only a few things one can include the source file one the target in the source file of the other target - see i.e. https://github.com/cc65/cc65/blob/master/libsrc/atari5200/cclear.s

mrdudz:
somehow that answer didnt help me at all to determine how to add waitvsync to c16 =P

oliverschmidt:
> somehow that answer didnt help me at all to determine how to add waitvsync to c16 =P

Sorry, I just wrote all I could that seemed related to the question...

mrdudz:
ok i tried... its NOT pulled into the c16 lib - which i think is kinda weird, the c16 lib as stuff plus4 doesnt and the other way around. i'll open an issue about it (as said, they are pretty much the same thing, no idea why there are even two different libs - another .cfg should do the trick)

polluks:
An idea of Uz: C16 lib has no bank switching overhead, a config is not sufficient.
Did you check my get_tv hint?

mrdudz:
i did, but that didnt answer my question (its quite obvious that you can do this). and to avoid the "bank switching" (ie enabling/disabling the ROM when calling into the kernal) can be solved by adding a single file that overrides the respective kernal calls. anyway lets discuss it in the issue -> https://github.com/cc65/cc65/issues/461

polluks:
I see.

groessler:
I think there should be some documentation be added for this function.
Looking at the source it seems now to be supported by

- most CBM machines
- gamate
- nes
- pce

But it's documented only in the 

- nes
- pce
- supervision (??!)

specific docs. The supervision doc is wrong, I don't see such a function in its library there. It should also added to the general "funcref" document.

PR #454, opened 2017-06-17
Stop cc65 from putting redundant .segment directives into its output file.

greg-king5:
It isn't a bug; but, it annoyed me.  A side-effect of PR #411 is that a `.segment` assembler directive is put into cc65's output file for _every_  uninitiated global (type=bss) variable that's in the file.  Most of them are the same; they don't _change_ to a different segment.  They just clutter the file.

So, this patch inserts a `.segment` only where the BSS segment name actually needs to change.  The resulting file is easier to examine.


oliverschmidt:
As you likely expected I wasn't even aware of that undesirable side effect of https://github.com/cc65/cc65/pull/411. I fully agree with your assessment. Thanks for the fix :-)

PR #455, opened 2017-06-26
Add functions that make it easier to write portable programs for OSI computers.

greg-king5:
I learned that we need those functions, while testing the conio peek functions on the `osic1p` target.


oliverschmidt:
Thanks.

PR #456, opened 2017-06-28
Add a makefile dependency for the libraries' "extra" files.

greg-king5:
The new dependency rules must be "bootstrapped".  You must force a rebuild of the extra object files, by using these commands:
```sh
rm lib/*.o
make lib
```


oliverschmidt:
The Travis Ci failure seems related to this change:
```
Fatal error: Cannot open dependency file `../libwrk/osic1p/osic1p-screen-s3-32x28.d': No such file or directory
```

greg-king5:
> The Travis Ci failure seems related to this change:
> ```
> Fatal error: Cannot open dependency file `../libwrk/osic1p/osic1p-screen-s3-32x28.d': No such file or directory
> ```

Yep.  I forgot about the consequence of putting the dependency file in a place that's different from the object file's place.  And, I hadn't tested it with a completely clean build.

Fixed, now.

oliverschmidt:
Thanks for the explanation - and the fix.

PR #458, opened 2017-07-17
c64: add dummy waitvblank();

pmjdebruijn:
having to #ifdef waitvblank makes for ugly code, having a dummy
waitvblank for systems lacking it seems like a lesser evil.

### DIFF
--- a/include/c64.h
+++ b/include/c64.h
@@ -158,6 +158,9 @@ extern void c64_hi_tgi[];               /* Referred to by tgi_static_stddrv[] */
 
 
 
+void waitvblank (void);
+/* Wait for the vertical blanking */
+
 unsigned char get_ostype (void);
 /* Get the ROM version. Returns one of the C64_OS_xxx codes. */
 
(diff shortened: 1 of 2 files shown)


pmjdebruijn:
It seems I missed #453, which a much better approach of course...

But the general issue raised here holds true, that it might be considered to have a stub waitvblank for platforms that do not have implemented it yet. As ifdeffing each waitvblank is a bit of a PITA. 

Ideally, whenever a stub function is used, the compiler should info/warn about the stub (which is where #460) might fit in.

Which leaves the question what's the nicer way to define a stub, so that may or may not be optimized away. In case of waitvblank throwing in a few nops might make sense as waitvblank usually causes some delay as well.

Feel free to close, however I would be interested in your view on this....

mrdudz:
a proper dummy could just be like

#define waitvsync()

... however, this hides the fact that the function does not exist, and it will result in a program that doesnt work as expected. not sure i like this idea :)

pmjdebruijn:
The critical thing here would be adding a compiler warning, so programmers will notice this, even if they didn't read the docs.

oliverschmidt:
At least so far the rule was not to add "fake/dummy" things to the C library. While I see your point that waitvsync is a bit special as it doesn't "do that much" I don't see us changing our rule. To me it seems better that cross target program contains the necessary #ifdef stuff. As i.e. done in https://github.com/cc65/cc65/blob/master/samples/sieve.c#L14

PR #459, opened 2017-07-17
pragma: bring pragma_t enum in line with Pragmas struct

pmjdebruijn:
fixes an inconsequential inconsistency, might prevent some
confusion in the future.


greg-king5:
Yes, a matching sorted order always is desirable.  Thanks.

PR #460, opened 2017-07-17
Pragma message

pmjdebruijn:


pmjdebruijn:
Somehow ed3d8f8 causes this, which makes little sense to me:

c128 - cbm/exec.c
cbm/exec.c(69): Error: Unknown pragma `zpsym'
cbm/exec.c(71): Error: Unknown pragma `zpsym'
cbm/exec.c(73): Error: Unknown pragma `zpsym'
cbm/exec.c(76): Error: Unknown pragma `zpsym'
make[2]: *** [../libwrk/c128/exec.o] Error 1
make[1]: *** [c128] Error 2
make: *** [all] Error 2

greg-king5:
We use `bsearch()` to find the pragma name in the table.  That function needs an alphabetically _sorted_ table.

greg-king5:
I wish to see a description for this PR.  It should tell us what `#pragma message` does in other compilers, and what you anticipate that cc65 should do about that pragma.

spiro-trikaliotis:
I am not the original author, but it seems he wants to be able to output informational messages:
https://msdn.microsoft.com/de-de/library/x7dkzch2(v=vs.90).aspx

pmjdebruijn:
DOH with regard to bsearch. Thanks for the pointer (I think we nearly glanced a pun there).

And indeed as @spiro-trikaliotis mentioned, I would like to be able to use pragma message similar to what's described on the microsoft page, gcc/clang also support this.

If I recall correctly some platform don't have a function textcolor/bgcolor/bordercolor, in which case they should probably be annotated with pragma message, so the compiler is explicit about this.

That said, I noticed lots of functions are implemeted as asm, so presumably pragma message support in cc65 won't help with that. Presumably something similar would need to be added to ca65?

PR #463, opened 2017-07-18
joy: expand/standardize JOY_ consts on NES

pmjdebruijn:
Standardize on A/B as opposed to FIRE/FIRE2 as per NES.

expand to include SELECT/START for the NES

PC Engine seems to be similar to the NES.

Most older platforms seems to be effectively single action button.

### DIFF
--- a/include/joystick.h
+++ b/include/joystick.h
@@ -60,8 +60,10 @@
 #define JOY_DOWN        1
 #define JOY_LEFT        2
 #define JOY_RIGHT       3
-#define JOY_FIRE        4
-#define JOY_FIRE2       5               /* Second fire button if available */
+#define JOY_A           4
+#define JOY_B           5               /* Second fire button if available */
+#define JOY_SELECT      6               /* Select or third button if available */
+#define JOY_START       7               /* Start or fourth button if avaiable */
 
 /* Array of masks used to check the return value of joy_read for a state */
 extern const unsigned char joy_masks[8];
@@ -71,8 +73,11 @@ extern const unsigned char joy_masks[8];
 #define JOY_BTN_DOWN(v)         ((v) & joy_masks[JOY_DOWN])
 #define JOY_BTN_LEFT(v)         ((v) & joy_masks[JOY_LEFT])
 #define JOY_BTN_RIGHT(v)        ((v) & joy_masks[JOY_RIGHT])
-#define JOY_BTN_FIRE(v)         ((v) & joy_masks[JOY_FIRE])
-#define JOY_BTN_FIRE2(v)        ((v) & joy_masks[JOY_FIRE2])
+#define JOY_BTN_A(v)            ((v) & joy_masks[JOY_A])
+#define JOY_BTN_B(v)            ((v) & joy_masks[JOY_B])
+#define JOY_BTN_SELECT(v)       ((v) & joy_masks[JOY_SELECT])
+#define JOY_BTN_START(v)        ((v) & joy_masks[JOY_START])
+
 
 /* The name of the standard joystick driver for a platform */
 extern const char joy_stddrv[];
--- a/include/nes.h
+++ b/include/nes.h
@@ -90,16 +90,6 @@
 /* No support for dynamically loadable drivers */
 #define DYN_DRV         0
 
-/* The joystick keys - all keys are supported */
-#define KEY_A           0x01
-#define KEY_B           0x02
-#define KEY_SELECT      0x04
-#define KEY_START       0x08
-#define KEY_UP          0x10
-#define KEY_DOWN        0x20
-#define KEY_LEFT        0x40
-#define KEY_RIGHT       0x80
-
 /* Define hardware */
 
 /* Picture Processing Unit */
--- a/libsrc/atari/joy/atrmj8.s
+++ b/libsrc/atari/joy/atrmj8.s
@@ -41,8 +41,8 @@
         .byte   $04                     ; JOY_DOWN
         .byte   $08                     ; JOY_LEFT
         .byte   $10                     ; JOY_RIGHT
-        .byte   $01                     ; JOY_FIRE
-        .byte   $00                     ; JOY_FIRE2 not available
+        .byte   $01                     ; JOY_A
+        .byte   $00                     ; JOY_B not available
         .byte   $00                     ; Future expansion
         .byte   $00                     ; Future expansion
 
--- a/libsrc/atari/joy/atrstd.s
+++ b/libsrc/atari/joy/atrstd.s
@@ -40,8 +40,8 @@
         .byte   $02                     ; JOY_DOWN
         .byte   $04                     ; JOY_LEFT
         .byte   $08                     ; JOY_RIGHT
-        .byte   $10                     ; JOY_FIRE
-        .byte   $00                     ; JOY_FIRE2 not available
+        .byte   $10                     ; JOY_A
+        .byte   $00                     ; JOY_B not available
         .byte   $00                     ; Future expansion
         .byte   $00                     ; Future expansion
 
--- a/libsrc/atari5200/joy/atr5200std.s
+++ b/libsrc/atari5200/joy/atr5200std.s
@@ -33,8 +33,8 @@
         .byte   $02             ; JOY_DOWN
         .byte   $04             ; JOY_LEFT
         .byte   $08             ; JOY_RIGHT
-        .byte   $10             ; JOY_FIRE
-        .byte   $20             ; JOY_FIRE2
+        .byte   $10             ; JOY_A
+        .byte   $20             ; JOY_B
         .byte   $00             ; Future expansion
         .byte   $00             ; Future expansion
 
@@ -99,7 +99,7 @@ READJOY:
         lda     #0              ; Initialize return value
         cmp     TRIG0,y
         bne     @notrg
-        lda     #$10            ; JOY_FIRE
+        lda     #$10            ; JOY_A
 
 ; Read joystick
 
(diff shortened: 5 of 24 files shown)


pmjdebruijn:
I'm aware it's a bit of a matter of preference what to standardize on, but the NES seems to be the most common supported platform which actually uses 8 "buttons". Which is why I would find it less awkward to map simpler cases onto the NES model, as opposed to the other way around.

pmjdebruijn:
Another note, I'd like for the JOY_BTN_SELECT etc to be available on other platforms too, so that it does not forcibly need to be ifdeffed. Obviously depending on the particular it might still need to be.

mrdudz:
i dont like the idea of calling the fire-button "A". its "fire" on most platforms....

and what would this JOY_BTN_SELECT defined as for example on c64?

please understand that the libraries are meant to represent the least common denominator (for portable programs). a portable program cant just use every feature present on some platform - in that case it must handle the incompatible things by itself. just defining a bunch of symbols to make things compile (but most likely not work as expected) is not the solution.

oliverschmidt:
Sorry but I need to second mrdudz.

The conceptial aspect: This falls from my perspective again into that "fake/dummy" scenario. If the joystick of a target doesn't have i.e. a 'Select' button I don't see the point in pretending it does.

The pratical aspect: One cannot judge beforehand how the buttons available on a target should be mapped to your "normalized" target. Only the author of the program can know that. Imagine i.e. the program needs three actions. Then it may want to use three buttons when available but may want both buttons pressed together on targets with only two buttons available. All this is up to the program. And the program might contain a help screen explaining how to trigger the actions. That help screen needs to name the buttons in the way the users of that target are used to name them.

pmjdebruijn:
Fair enough. I'd be happy to rework it...

But the current implementation doesn't actually properly follow that philosphy either. As FIRE2 is accessible to many platforms that only have a single FIRE button if I'm not mistaken.

I'd be happy the file a new PR, to bring everything in line with the described philosphy:
From include/joystick.h remove FIRE2, and move it to platforms which actually have it.
Add NES/FCE specific stuff to the relevant system specific files, in a way that is consistent with the JOY_BTN_ approach.
From include/nes.h remove KEY_ as it seems different from JOY_BTN for no obvious reason.

Would that be considered to be merged, or at least a new point for further discussion?

oliverschmidt:
> But the current implementation doesn't actually properly follow that philosphy either. As FIRE2 is accessible to many platforms that only have a single FIRE button if I'm not mistaken.

Very good point indeed!

I'd even go one step further: I'd suggest to rename:

- `JOY_FIRE` to `JOY_PUSH`.
- `JOY_BTN_FIRE` to `JOY_BTN_PUSH`.

I have two reasons:

- We have several targets where the joystick device(s) doesn't follow like the classic "fighter plane stick" design and where therefore the it's perfectly unclear which button is the "Fire Button".
- I personally don't need references to military vocabulary where not necessary or even appropriate.

So I'd generally appreciate a pull request that:

- Removes FIRE2...
- Renames FIRE... to PUSH...
- Adds appropriate defines to target-specific headers for targets with more than one button
- Adjusts https://github.com/cc65/cc65/blob/master/testcode/lib/joy-test.c to test only one button (named PUSH)

mrdudz:
military vocabulary? oh please... ^_^ - at least call it "trigger" then, that term is at least used elsewhere too (unlike push...)

btw, the C64 (and VIC20, and Atari8-bit, perhaps more) CAN use 2 (even 3) buttons... perhaps the driver should be updated for that :) (see http://wiki.icomp.de/wiki/DB9-Joystick )

oliverschmidt:
I thought about "trigger" too.  But "push" seems more neutral (and shorter). Look i.e. at the typical Apple II joystick:
applejoystick
Can you tell me which button is the trigger (or fire) button? There are just two equal buttons. An experienced user knows which one is the "primary" and which one is the "secondary". So if a program just talks about "the" button he knows which one is meant. I would like to call it just "button" but as the four direction thingies are called button too I opted for "push".

Regarding the number of buttons this is clearly an unclear thing. We have at least potentially several drivers for each target and therefore we don't even know the exact number of actually available buttons for a single target. I see two options:

- The generic header file defines PUSH1 to PUSH9 and the target header files optional provide aliases like A and B or OPTION and SELECT.
- The generic header file defines just PUSH and the target header files provide the rest. From a conceptional pov this is impossible as the number of buttons can potentially vary between different joysticks (-> joystick drivers). But from a realistic pov there seems to be a "usual" number of buttons on joysticks for a certian target.

And given that the driver can tell how many joysticks it supports it may very well have been a good idea to allow it to tell how many buttons the joystick(s) it supports have.

However for many years there was no momentum for improvement and now Pascal seems interested (or even eager) to invest. So I don't see me turning this effort down by saying we should change everything in general ;-)

oliverschmidt:
> PUSH1 to PUSH9

As far as I understand this should rather read PUSH1 to PUSH4 as we seem to be limited to the bits left in one byte.

silverdr:
> On 2017-07-20, at 21:39, Oliver Schmidt <notifications@github.com> wrote:
> 
> But the current implementation doesn't actually properly follow that philosphy either. As FIRE2 is accessible to many platforms that only have a single FIRE button if I'm not mistaken.
> 
> Very good point indeed!
> 
> I'd even go one step further: I'd suggest to rename:
> 
> 	• JOY_FIRE to JOY_PUSH.
> 	• JOY_BTN_FIRE to JOY_BTN_PUSH.

PUSH? Can't it all be just globally be BTN0, 1, ... and target specific aliases for targets that have well defined controllers with named buttons?

-- 

oliverschmidt:
For all of you wondering about my "push" idea:

I'd call the thing just "button" but if you look at  https://github.com/cc65/cc65/blob/master/include/joystick.h#L70-L73 you see that the four directions are already called JOY_BTN_... so if everything is called a button how to call THE button(s)? Maybe push buttons...

Apart from the naming I think by now too, that it is preferable to have 4 neutral named definitions in the generic header and optional aliases in the target headers. The fact that there are usually no 4 fire/trigger/push/<...> buttons is acceptable as even the target header doesn't really know how many buttons are actually present (and the drivers don't even tell at run time).

silverdr:
> On 2017-07-21, at 10:15, Oliver Schmidt <notifications@github.com> wrote:
> 
> For all of you wondering about my "push" idea:
> 
> I'd call the thing just "button" but if you look at https://github.com/cc65/cc65/blob/master/include/joystick.h#L70-L73 you see that the four directions are already called JOY_BTN_... so if everything is called a button how to call THE button(s)? Maybe push buttons...

Now it is clear where you get the idea from. Just, to me it looks like actually the lines (macros) you refer to are misleadingly named. I'd rather deprecate the macros and give them new, less ambiguous names like - say - JOY_DIR_UP, JOY_DIR_LEFT, ...

Then there would be clear namespace for buttons JOY_BTN_0, 1, ... and target specific aliases. As you noticed there shouldn't be anything wrong with more buttons defined even if the target doesn't support all of them. It would be target specific implementation detail what would be returned.

-- 

oliverschmidt:
> Now it is clear where you get the idea from.

:-)

> Just, to me it looks like actually the lines (macros) you refer to are misleadingly named.

Full ACK.

> JOY_DIR_UP, JOY_DIR_LEFT, ...

Yep.

> JOY_BTN_0, 1, ...

Yep.

> As you noticed there shouldn't be anything wrong with more buttons defined even if the target doesn't support all of them.

To clarify further:

In general I think we should stick to "fail early". So rather fail to build because a certain function/macro doesn't exsist for a certain target than experience that it just doesn't work/does nothing. This is especially true as several of us build programs for targets we don't actually know so we don't know how to test if things actually work as expected.

However in this case even the target header files can't contain the "only the correct" number of buttons as different drivers for a certain target can very well support a different number of buttons. And unfortunately we can't even "fail late" as the driver doesn't tell how many buttons it supports (in contrast to the number of joysticks). Given that level of uncertainty we can as well have the maximum number of technically possible buttons (as far as I see 4) defined right in the generic header file.

As discussed the target header files can additionally add aliases with target specific names.

It seems that we get to a common perspective here so maybe Pascal (or in fact anybody) wants to create a pull request doing that.

greg-king5:
I agree with the idea of putting generic names in the API header; and, of putting the platforms' typical names in the target headers (the typicals would be mapped to the generics).

I think that the names of the evaluation macroes should be swapped with the names of the indices.  Then, they will be easier for programmers to type:
```C
/* The following codes are *indices* into the joy_masks array */
#define JOY_DIR_UP      0
#define JOY_DIR_DOWN    1
#define JOY_DIR_LEFT    2
#define JOY_DIR_RIGHT   3
#define JOY_BUTTON1     4
#define JOY_BUTTON2     5
#define JOY_BUTTON3     6
#define JOY_BUTTON4     7

/* Macros that evaluate the return code of joy_read() */
#define JOY_UP(v)       ((v) & joy_masks[JOY_DIR_UP])
#define JOY_DOWN(v)     ((v) & joy_masks[JOY_DIR_DOWN])
#define JOY_LEFT(v)     ((v) & joy_masks[JOY_DIR_LEFT])
#define JOY_RIGHT(v)    ((v) & joy_masks[JOY_DIR_RIGHT])
#define JOY_BTN1(v)     ((v) & joy_masks[JOY_BUTTON1])
#define JOY_BTN2(v)     ((v) & joy_masks[JOY_BUTTON2])
#define JOY_BTN3(v)     ((v) & joy_masks[JOY_BUTTON3])
#define JOY_BTN4(v)     ((v) & joy_masks[JOY_BUTTON4])
```

Question:
Should those "index" names be macroes or an enumeration?

pmjdebruijn:
I just filed #467 with what I gathered from this thread in what I think would be a more consistent approach as opposed to @greg-king5's example above.

PR #467, opened 2017-07-22
joy: refactor generic joy_masks array indices and macros

pmjdebruijn:
Resulting out of the discussion in #463


oliverschmidt:
I must admit that I (actually rather strongly) prefer Greg's idea:

```
/* The following codes are *indices* into the joy_masks array */
#define JOY_DIR_UP      0
#define JOY_DIR_DOWN    1
#define JOY_DIR_LEFT    2
#define JOY_DIR_RIGHT   3
#define JOY_BUTTON1     4
#define JOY_BUTTON2     5
#define JOY_BUTTON3     6
#define JOY_BUTTON4     7

/* Macros that evaluate the return code of joy_read() */
#define JOY_UP(v)       ((v) & joy_masks[JOY_DIR_UP])
#define JOY_DOWN(v)     ((v) & joy_masks[JOY_DIR_DOWN])
#define JOY_LEFT(v)     ((v) & joy_masks[JOY_DIR_LEFT])
#define JOY_RIGHT(v)    ((v) & joy_masks[JOY_DIR_RIGHT])
#define JOY_BTN1(v)     ((v) & joy_masks[JOY_BUTTON1])
#define JOY_BTN2(v)     ((v) & joy_masks[JOY_BUTTON2])
#define JOY_BTN3(v)     ((v) & joy_masks[JOY_BUTTON3])
#define JOY_BTN4(v)     ((v) & joy_masks[JOY_BUTTON4])
```

> Should those "index" names be macroes or an enumeration?

I don't see strong benefits in one or the other solution so I'd maybe just stay with the macros...

pmjdebruijn:
But that arrange is internally inconsistent...

Why would the direction key indices or macros have an explicitly _DIR_ added to them, but not the buttons? In my view, they should either be both there, or neither.

As for the macros, that flows forth from the indices... JOY_BTN_BUTTON1 would look a bit strange...

And in my view there should be a direct link between the index and macro, so the JOY_NAME index should result in a JOY_TYPE_NAME macro, where NAME is consistent, not BTN1 that completely arbitrarily translates to BUTTON1 somehow.

Essentially the macros should be trivial to generate by script based on the indices. Even if it's not a real world concern here. Trivial scriptability is usually a sign of good internal consistency.

OTOH:

If the concern is that you don't want the _DIR_ or _BTN_ type indicator at all, maybe this would work:

#define JOY_UP 0
#define JOY_UP_PRESSED ((v) & joy_masks[JOY_UP])

#define JOY_BUTTON1 5
#define JOY_BUTTON1_PRESSED ((v) & joy_masks[JOY_BUTTON1])

PRESSED could be ACTIVE too, in the end that's just a nomenclature argument, as opposed to structure consistency.

pmjdebruijn:
Another sidenote.

Keep in mind that on system like the NES and FCE, buttons are often just referred to as A or I, and text usually say "press A", not "Press the A Button" if I recall correctly at least.

Keeping that I mind, I don't see the issue with JOY_1 as opposed to JOY_BUTTON1.

pmjdebruijn:
Which reminds me, how do most C64 games actually call the first (and practically only) button, is that where FIRE came from?

In which case, I guess I would need to define FIRE in the commodore relevant header...

oliverschmidt:
I think Gregs idea is (or that's at least what I like about it is):

The "return code macros" are the ones that I see being used. They should be short/nice/descriptive. The index macros are more like helpers to me. But they need to have names on their own - and those names have to be different from the return code ones. So BTN becomes BUTTON and the directions get a DIR infix. Not necessarily consistent but practical.

pmjdebruijn:
Doesn't something like this make more sense then:

#define INDEX_JOY_UP
#define JOY_UP ((v) & joy_masks[INDEX_JOY_UP])

#define INDEX_JOY_1
#define JOY_1 ((v) & joy_masks[INDEX_JOY_1])

oliverschmidt:
Using an INDEX prefix for all index macros is fine for me. But I dislike JOY_1. It's about a button (be it push or fire or trigger). Apart from that JOY_1 could be mistaken as joystick 1.

pmjdebruijn:
For sake of argument I've amended the commit, including the NES/FCE bits.

I do get the JOY_1 conflict, I briefly forgot about that. to avoid that keeping _DIR_ and _BTN_ makes a lot of sense.

greg-king5:
I don't like the `IDX_` prefix; I want everything in "joystick.h" to have the `JOY_` and `joy_` prefixes.

However, I now wonder if we need to define the indices, at all.  Would they ever be used outside of "joystick.h"?  (The target headers should not use them.)  Maybe, we can define only:
```C
/* Macros that evaluate the return code of joy_read() */
#define JOY_DIR_UP(v)           ((v) & joy_masks[0])
#define JOY_DIR_DOWN(v)         ((v) & joy_masks[1])
#define JOY_DIR_LEFT(v)         ((v) & joy_masks[2])
#define JOY_DIR_RIGHT(v)        ((v) & joy_masks[3])
#define JOY_BTN_1(v)            ((v) & joy_masks[4])
#define JOY_BTN_2(v)            ((v) & joy_masks[5])
#define JOY_BTN_3(v)            ((v) & joy_masks[6])
#define JOY_BTN_4(v)            ((v) & joy_masks[7])
```

For the sake of encapsulation, the target headers should "know" as little as possible about the joystick API and its code.  Their only job is to give some typical platform names.  Therefore, they should map "typical" macroes directly to the above generic macroes.  E.g.,
```C
#define JOY_BTN_A(v)            JOY_BTN_1(v)
#define JOY_BTN_B(v)            JOY_BTN_2(v)
#define JOY_BTN_SELECT(v)       JOY_BTN_3(v)
#define JOY_BTN_START(v)        JOY_BTN_4(v)
```
(The target headers don't need to know about `joy_mask[]`.)

I accept the idea of splitting the names into a `_DIR` group and a `_BTN` group.

oliverschmidt:
> Presumably you might want to keep BTN and drop DIR, if so, I'll amend it for final merging.

Yep, that's my opinion.

> However, I now wonder if we need to define the indices, at all. Would they ever be used outside of "joystick.h"? (The target headers should not use them.)

I'm fully with you. I never saw the point in them. Let's drop them.

> I accept the idea of splitting the names into a _DIR group and a _BTN group.

Hm, I'm not for that (see above). It's obvious that LEFT is a direction, so there's no need for DIR_LEFT, but it's not obvious that 1 is a button, so JOY_BTN_1 is necessary.

> #define JOY_BTN_SELECT(v)       JOY_BTN_3(v)

I'm fine with that as well as with JOY_SELECT.

groessler:
>>  However, I now wonder if we need to define the indices, at all. Would they ever be used outside of "joystick.h"? (The target headers should not use them.)

> I'm fully with you. I never saw the point in them. Let's drop them.

I think they can be beneficial inf one want to test for more than one thing. Let's say to detect pressing of any of two (or more) trigger keys, or to test for having both keys pressed simulateneously. It can be done with the JOY_BTN_xxx macros, but I think cc65 will generate better code if one accesses joy_masks[] directly in this case.

pmjdebruijn:
So I just amended again with some of the feedback.

Questions that are left are...

... keep IDXes or not? As you might noticed I've added them as a suffix now, instead of a prefix, to alleviate the we want to keep JOY_ as the common prefix concern.


Also, I noticed Gamate had some stuff left as well, which I included now, however include/gamate.h had some duplication that's also present in asminc/gamate.inc.

silverdr:
> On 2017-07-23, at 00:09, greg-king5 <notifications@github.com> wrote:
> 
> I don't like the IDX_ prefix; I want everything in "joystick.h" to have the JOY_ and joy_ prefixes.
> 
> However, I now wonder if we need to define the indices, at all. Would they ever be used outside of "joystick.h"? (The target headers should not use them.) Maybe, we can define only:
> 
> /* Macros that evaluate the return code of joy_read() */
> 
> #define JOY_DIR_UP(v)           ((v) & joy_masks[0])
> #define JOY_DIR_DOWN(v)         ((v) & joy_masks[1])
> #define JOY_DIR_LEFT(v)         ((v) & joy_masks[2])
> #define JOY_DIR_RIGHT(v)        ((v) & joy_masks[3])
> #define JOY_BTN_1(v)            ((v) & joy_masks[4])
> #define JOY_BTN_2(v)            ((v) & joy_masks[5])
> #define JOY_BTN_3(v)            ((v) & joy_masks[6])
> #define JOY_BTN_4(v)            ((v) & joy_masks[7])

Just a nitpick - I don't know about you guys but at least /me will find it... odd to have the primary button bearing index "1" rather than "0". We're talking programmers here, right? And the target specific aliases will have descriptive names anyway. Be it "FIRE", "SELECT", "START", whatever..

-- 

pmjdebruijn:
That's not really the case, only a few targets have target specific aliases, namely NES/FCE/Gamate.

And wouldn't it be totally weird if you'd have code that said:

printf('please press button 1')
if (JOY_BTN_0(joy1)) { ... }

Counting from 0, "because we're all programmers", seems misplaced here.

mrdudz:
i second this... its "joystick number 1" and "button number 1"

silverdr:
> On 2017-07-23, at 17:44, Pascal de Bruijn <notifications@github.com> wrote:
> 
> That's not really the case, only a few targets have target specific aliases, namely NES/FCE/Gamate.
> 
> And wouldn't it be totally weird if you'd have code that said:
> 
> printf('please press button 1')
> if (JOY_BTN_0(joy1)) { ... }
> 
> Counting from 0, "because we're all programmers", seems misplaced here.

As I wrote it's a nitpick and I am not going to fight for this but I always find it hard to justify indexing things in a *programmers'* interface starting with anything else than 0. And I've seen more than enough of those to be convinced that it's always confusing and leading to more than necessary ;-) "off-by-one" errors. Since we're making substantial changes to this interface it seems like a good moment to do it "right". I don't know about all targets but on those I am familiar with the lines would read:

printf('Please press FIRE button');
if (JOY_BTN_0(joy0))
{ ...

or

printf('Please press START button');
if (JOY_BTN_0(joy0))
{ ...

Do we have targets that have "officially" the buttons numbered on their controllers? Or are those usually named like "FIRE", "START", "SELECT", "A", "B", "UP", "RESET", etc.

-- 

silverdr:
> On 2017-07-23, at 18:17, Bob Andrews <notifications@github.com> wrote:
> 
> i second this... its "joystick number 1" and "button number 1"

Which means second button on the second joystick? ;-) At least that what comes to programmer's mind.

Anyway - I don't see it as anything major but rather a minor cleanup while we are at it. If not - well, just one more oddly indexed stuff to deal with.

-- 

pmjdebruijn:
PCE has I and II buttons, which is essentially 1 and 2, but in that case I made defines that literally said I and II.

silverdr:
> On 2017-07-23, at 19:02, Pascal de Bruijn <notifications@github.com> wrote:
> 
> PCE has I and II buttons, which is essentially 1 and 2, but in that case I made defines that literally said I and II.

Exactly - and that's also how I see it. Generic indexes start at 0 (for programmer's peace of mind/mental health) and "common names" are whatever is the most - well - common or most understandable in the context of a particular target. In your example it is "I" and "II", which you also decided to use. As before though - all of this is basically only a suggestion that came to my mind immediately after noticing the proposed indices.

-- 

greg-king5:
>>>  However, I now wonder if we need to define the indices, at all. Would they ever be used outside of "joystick.h"? (The target headers should not use them.)
>
>> I'm fully with you. I never saw the point in them. Let's drop them.
>
> I think they can be beneficial if one wants to test for more than one thing. Let's say to detect the pressing of any of two (or more) trigger keys, or to test for having both keys pressed simulateneously. It can be done with the `JOY_BTN_xxx()` macros; but, I think cc65 will generate better code if one accesses `joy_masks[]` directly, in this case.

That's a good point.

Currently, it's even worse!  I tested two ways to combine buttons.  This (partial) statement generates broken code:
```C
if (JOY_BTN_FIRE(v) | JOY_BTN_FIRE2(v))
```
The compiler creates 16-bit code.  The optimizer sees that the high byte always is zero; so, it tries to squeeze that code into 8 bits.  But, the final test for zero/non-zero isn't changed; a garbage high byte is tested -- so, that `if` statement usually will fail to tell the truth!

The statement
```C
if (v & (joy_masks[JOY_FIRE] | joy_masks[JOY_FIRE2]))
```
optimizes into correct, efficient 8-bit code.

So, we do need to have those index macroes; they actually might be used outside of "joystick.h".

pmjdebruijn:
I didn't remove the indices yet :)

But in the end presumably this is something that _ideally_ should be fixed in the optimizer right?

I find the former code much more readable than the latter.

With regard to the BTN_0 vs BTN_1 thing, the point there that's it's not an index, they're names, it could have been BTN_A, etc. 

Also starting with BTN_0, BTN_1, BTN_2, etc creates the problem that if there is a console that actually has a 1 BUTTON, BTN_1 is already defined and points to the second button instead of the first.

The SEGA Master System for example has this, though I don't think it has 6502 CPU though.

From what I gather my last amend should be fit for merge... If not, please do tell me what needs to change to achieve a merge...

greg-king5:
> Which reminds me, how do most C64 games actually call the first (and practically only) button; is that where FIRE came from?  In which case, I guess I would need to define FIRE in the Commodore-relevant header.

Yes.  And, yes -- in `<cbm.h>`.

greg-king5:
The Lynx's buttons are labelled "A" and "B".

pmjdebruijn:
Amended, presuming I got the indices right, can anybody verify that?

groessler:
> Just a nitpick - I don't know about you guys but at least /me will find it... odd to have the primary button bearing index "1" rather than "0". We're talking programmers here, right?

I agrre that numbering from 0 is the "natural" way programmers (me, at least) think.

groessler:
> But in the end presumably this is something that ideally should be fixed in the optimizer right?

Of course the optimized shoult be fixed...

> I find the former code much more readable than the latter.

... but it's not us to decide what is more "reabable" to a given programmer. Additionally, for optimization, the programmer could create his own bit mask to test beforehand, and the use this one during the program flow, at least one "or" operation less to do during runtime.

include/gamate.h
```diff
@@ -80,17 +80,6 @@
  bit 3:
 */
 
-#define JOY_DATA        0x4400
```
oliverschmidt: Does anybody know what this was supposed to be used for?

include/joystick.h
```diff
@@ -52,27 +52,31 @@
...
-#define JOY_LEFT        2
-#define JOY_RIGHT       3
-#define JOY_FIRE        4
-#define JOY_FIRE2       5               /* Second fire button if available */
+#define JOY_UP_IDX              0
+#define JOY_DOWN_IDX            1
+#define JOY_LEFT_IDX            2
+#define JOY_RIGHT_IDX           3
+#define JOY_BTN_1_IDX           4       /* Universally available */
+#define JOY_BTN_2_IDX           5       /* Second fire button if available */
```
oliverschmidt: As discussed there are on generic "fire" buttons.

include/joystick.h
```diff
@@ -52,27 +52,31 @@
...
+#define JOY_BTN_3_IDX           6       /* Third fire button if available  */
```
oliverschmidt: As discussed there are on generic "fire" buttons.

include/joystick.h
```diff
@@ -52,27 +52,31 @@
...
+#define JOY_BTN_4_IDX           7       /* Fourth fire button if available */
```
oliverschmidt: As discussed there are on generic "fire" buttons.

clbr:
This change broke my existing NES code, and replacing it with the generic replacements would make my code slower.

Please restore the direct KEY_ defines to nes.h.

oliverschmidt:
Sorry but I don't see any KEY_ defines being removed from nes.h by this pull request.

clbr:
2d9922a46e6112ed901dc675d37a6bfc4f928c4d was a precursor to this pull request.

oliverschmidt:
Thanks - Now I see. Could you please provide a snippet of of your current code so I can understand your exact usage scenario?

clbr:
```
if (ctrl & (KEY_A | KEY_B | KEY_START))
```

Before that commit, it compiled to a single AND with a constant. Using
the current generic way, it would take several indirect loads and
combines at runtime. Besides being ugly, naturally.

```
if (ctrl & (joy_masks[JOY_BTN_A_IDX] | joy_masks[JOY_BTN_B_IDX] |
joy_masks[JOY_BTN_START_IDX]))
```

oliverschmidt:
Okay, my understanding is this. We have two issues here:

- Ugly code (minor): The current setup presumes that every button is tested individually. If that scenario one can use the JOY_...(v) macros so the joy_masks[] array isn't visible at all. To me the idea to test several buttons together seems esoteric enough to not justify explicit support.

- Ineffective code (major): This seems to me like just another aspect of a more general problem: We have those dynamically loadable drivers and their architecture brings quite some overhead. But we don't have a benefit from the loadable drivers most of the time because
a) the hardware in question is always available so there's no need to be able to resort to a different hardware (think of almost all TGI drivers)
b) there's only instance for given hardware "class" for a given target (think of most JOY, MOU, SER drivers)
Coming back to this specific issue: For JOY drivers there is the idea that different drivers can put different values in the joy_masks array. So in order to be independent from different loadable drivers the indirection is necessary. Now you can of course say that
a) You don't care about the other drivers beside your favorite one.
b) You don't load the driver dynamically anyway.
c) You couldn't even load the driver dynamically because there's no file i/o in the first place
d) There's only one driver available anyway for your target
But nevertheless it's not really appropriate to put fixed constants in the header files as there - conceptually - might be a different driver with wanting to set different values in the joy_masks.

So what to do about it:

- Easy solution: Put the constants in question in your own sources.

- Hard solution: Maybe I'm wrong and someone want's to clarify but the way I see thing we could/should get rid of that whole joy_masks thingy. We could/should define fixed values for the buttons in the target header files. The values for those buttons could/should be derived from the "primary" JOY driver for that target. If a target has more than one driver then the "secondary" JOY drivers would have to add some code to "emulate" the values of the primary driver.

The benefit would be that all programs could avoid the overhead introduced by the compiled code necessary for the joy_masks lookup. And only very few drivers would have to - potentially - add one assembly language lookup.

I'm definitively open for contributions (and willing to adjust the Apple II part myself as usually).

clbr:
> To me the idea to test several buttons together seems esoteric enough to not justify explicit support.

This happens all the time in games, it's a very common operation.

> a) You don't care about the other drivers beside your favorite one.
> b) You don't load the driver dynamically anyway.
> c) You couldn't even load the driver dynamically because there's no file i/o in the first place
> d) There's only one driver available anyway for your target

All these points, indeed.

> - Easy solution: Put the constants in question in your own sources.

It was added to the platform-specific header exactly to reduce
duplication in all of our sources.

> - Hard solution: Maybe I'm wrong and someone want's to clarify but the way I see thing we could/should get rid of that whole joy_masks thingy. We could/should define fixed values for the buttons in the target header files. The values for those buttons could/should be derived from the "primary" JOY driver for that target. If a target has more than one driver then the "secondary" JOY drivers would have to add some code to "emulate" the values of the primary driver.
> 
> The benefit would be that all programs could avoid the overhead introduced by the compiled code necessary for the joy_masks lookup. And only very few drivers would have to - potentially - add one assembly language lookup.
> 
> I'm definitively open for contributions (and willing to adjust the Apple II part myself as usually).

This is a regression, and reverting the commit in question is the fix.
If you were going to require a joy overhaul, IMHO you should have asked
it from the OP of this pull request, given he wanted to overhaul the
joy_masks.

oliverschmidt:
> All these points, indeed.

I listed them for the very reason to point that out ;-)

> This is a regression, and reverting the commit in question is the fix.

I think I invested enough time to explain why this isn't the case.

> If you were going to require a joy overhaul, IMHO you should have asked
it from the OP of this pull request, given he wanted to overhaul the
joy_masks.

I see your position but I'm far from agreeing to it.

clbr:
Very well. Then we will freeze cc65 at the working version.

oliverschmidt:
Or even better - revert to ftp://ftp.musoftware.de/pub/uz/cc65/ ;-))

silverdr:
> On 2017-08-18, at 13:31, Oliver Schmidt <notifications@github.com> wrote:
> 
> 	• Hard solution: Maybe I'm wrong and someone want's to clarify but the way I see thing we could/should get rid of that whole joy_masks thingy. We could/should define fixed values for the buttons in the target header files. The values for those buttons could/should be derived from the "primary" JOY driver for that target. If a target has more than one driver then the "secondary" JOY drivers would have to add some code to "emulate" the values of the primary driver.

To me the idea of a "driver" is a form of abstraction layer with consistent interface in the first place. So the code using that abstraction layer should be - well - abstracted from the underlying implementation. Having said that, the definition of the interface is another process that might not be fully obvious or straightforward and will often lead to some compromises one way or another. With this, I wouldn't call the potential addition being needed to "emulate" anything but rather to "implement" the defined interface. Basing the interface specification on the requirements of the most commonly used (aka "primary") driver seems like an understandable compromise.

> The benefit would be that all programs could avoid the overhead introduced by the compiled code necessary for the joy_masks lookup. And only very few drivers would have to - potentially - add one assembly language lookup.

I haven't deeply analysed your claims here but it seems at least "fair enough" to me.

-- 

oliverschmidt:
> With this, I wouldn't call the potential addition being needed to "emulate" anything but rather to "implement" the defined interface. Basing the interface specification on the requirements of the most commonly used (aka "primary") driver seems like an understandable compromise.

Thanks for clarifying. The term "emulating" was meant to over-articulate the point I wanted to make - but it likely was an unfortunate term here.

mrdudz:
i have to agree with clbr here. if anything, the whole joystick driver thing should be optimized for that case that essentially one specific driver is linked statically, because that _IS_ the common case (i have yet to see one real world program that works differently).
and yes, testing some buttons at once is common, and has to be be fast. using indirect lookups is just silly. i wonder why that was ever proposed anway - if anything the damn driver could just return a "normalized" value. iirc those i looked at get the value from a table anyway.
last not least this shows that too much issues were adressed in a single pull request here. the renaming was ok. the rest should have been denied instantly. oh well.

greg-king5:
On 2017-08-18, Oliver Schmidt wrote:
>
> • Hard solution:  Maybe I'm wrong, and someone wants to clarify; but, the way I see things is that we could/should get rid of that whole joy_masks thingy. We could/should define fixed values for the buttons in the target header files. The values for those buttons could/should be derived from the "primary" JOY driver for that target. If a target has more than one driver, then the "secondary" JOY drivers would have to add some code to "emulate" the values of the primary driver.

That's how the joystick drivers worked originally (the mouse drivers still work that way).  I don't remember exactly why Uz moved the mapping out of the drivers, and into the applications.  I suppose that he thought the same way that you did.  He didn't realize that some platforms/games love button combinations.

include/gamate.h
greg-king5: > Does anybody know what this was supposed to be used for?

It's the address of the joypad's I/O port.

mrdudz:
i always considered this extra mapping optional... if it really works like i think it works now... its really a regression and should be made like it worked before

oliverschmidt:
Trying to sort things out...

- Obviously I'm not creative enough to see why one checks for several buttons at once. However I believe that this aspect isn't really important because the indirection over the joy_masks array is undesirable for checking a single button just as well.

- The current driver architecture means quite some overhead both measured by bytes as well as by cycles. I.e. every single driver call is indirected. I've written about that many times so I don't want to reiterate all this here. Let me just say that I'm in general very interested in getting rid of all this (at least for most scenarios).

- There are quite some places where cc65 sacrifices efficiency to get consistency and abstraction. I don't considered it a viable approach to nominally have the abstractions in place but then just bypass them. It's either the one _OR_ the other way. So given the current driver architecture the constants in nes.h were at best a hack. And from that perspective the pull request removing them was a correction - period.

- I strongly suggest that we stop this bashing and instead put our energy in implementing what I believe we all agree on as being desirable: Removing joy_masks !

clbr:
> I strongly suggest that we stop this bashing and instead put our energy in implementing what I believe we all agree on as being desirable: Removing joy_masks !

Breaking working code, and then saying "accept worse code or do lots of work for us" is not the way to get work from others.

oliverschmidt:
> Breaking working code

If you can't accept change stay with Uz' code.

> not the way to get work from others

You're right. I'll better take your attitude as model on how to best motive others ;-))

silverdr:
> On 2017-08-19, at 09:45, clbr <notifications@github.com> wrote:
> 
> I strongly suggest that we stop this bashing and instead put our energy in implementing what I believe we all agree on as being desirable: Removing joy_masks !
> 
> Breaking working code, and then saying "accept worse code or do lots of work for us" is not the way to get work from others.

C'mon - for myself I have to deal with changes that break some perfectly working (commercial, in production!) code regularly. I don't say this particular change is good nor am willing to take sides but the fact that a working code no longer works after some changes to the framework/toolset is very common (that's _also_ why we write automated tests, isn't it?) and yes, it takes time to adapt and it sometimes requires worse code too. This change may be bad from your point of view or may even be objectively bad but it shouldn't cripple the ability go forward anyway.

-- 

oliverschmidt:
Hm, maybe I should make it clear: I will not bring back the NES hack into this repo. I'll rather stop my work on cc65. In fact everybody is welcome to take over - and has always been !

oliverschmidt:
https://github.com/cc65/cc65/commit/7f52a770d9c6a43dde67689032bb3a4848eb6b30 removes joy_masks. Almost all driver code could stay untouched. There are however three drivers that need adjustmnet - which is _not_ part of the commit. I'd appreciate if someone else would take care of adjusting the three drivers:
- atrmj8.s
- c64-numpad.s
- vic20-stdjoy.s

clbr:
This works, thanks.

oliverschmidt:
Okay, I adjusted the three drivers in question myself (https://github.com/cc65/cc65/commit/0c54a19a0b482e5498c0c95ab1588731d34623d4, https://github.com/cc65/cc65/commit/a817da7c6bc0db40cbe23866b4b7a04b280ab612, https://github.com/cc65/cc65/commit/acf831658f591675b890c9c191fcc540094ae77e). I'm however not able to test my changes (that I consider complex enough to have bugs). I'd appreciate if someone else would take care of testing them.

PR #468, opened 2017-07-22
make changes

pmjdebruijn:
This branch does two things:
1. it installs the samples to their own subdirectory, which makes the installation result a bit more tidy.
2. it changes the prefix env var, to uppercase, since this seems to be in more common use, as this confused me while creating the Debian packaging as well.

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -6,8 +6,8 @@ endif
 
 .SUFFIXES:
 
-htmldir = $(prefix)/share/doc/cc65$(DESTPACKAGE_SUFFIX)/html
-infodir = $(prefix)/share/info
+htmldir = $(PREFIX)/share/doc/cc65$(DESTPACKAGE_SUFFIX)/html
+infodir = $(PREFIX)/share/info
 
 ifdef CMD_EXE
 
@@ -46,7 +46,7 @@ clean:
 	$(RM) -r ../html ../info
 
 install:
-	$(if $(prefix),,$(error variable `prefix' must be set))
+	$(if $(PREFIX),,$(error variable `PREFIX' must be set))
 ifeq ($(wildcard ../html),../html)
 	$(INSTALL) -d $(DESTDIR)$(htmldir)
 	$(INSTALL) -m0644 ../html/*.* $(DESTDIR)$(htmldir)
--- a/libsrc/Makefile
+++ b/libsrc/Makefile
@@ -72,7 +72,7 @@ endif
 
 ifndef TARGET
 
-datadir = $(prefix)/share/cc65
+datadir = $(PREFIX)/share/cc65
 
 all lib: $(TARGETS)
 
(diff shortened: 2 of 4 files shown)


oliverschmidt:
Hm, my reference (as Makefile author) was/is https://www.gnu.org/software/make/manual/html_node/Directory-Variables.html#Directory-Variables and there it's `prefix`, not `PREFIX`.

pmjdebruijn:
Hmmm, I've rarely come acros prefix=, while I regularly have to use PREFIX=, practically speaking...

Also keep in mind that the GNU documentation heavily assumed you're having a full autotools setup, which isn't the case for cc65 at this point (and given it's lack of dependancies, probably isn't needed either).

PREFIX has always made more sense to me, as DESTDIR is in uppercase too.

If you want I can remove this particular commit, so at least the samples commit might still pass?

oliverschmidt:
> Also keep in mind that the GNU documentation heavily assumes you're having a full autotools setup, which isn't the case for cc65 at this point (and given it's lack of dependancies, probably isn't needed either).

Please note that I didn't want to imply that I'm "right" and you are "wrong". I just wanted to point out that it is deliberately lowercase and the reason why. Maybe others have opinions too and/or maybe you have a "better" reference than I have.

> PREFIX has always made more sense to me, as DESTDIR is in uppercase too.

I'd never dare to try to make sense out of this stuff ;-))

> [...] at least the samples commit might still pass?

I'm pretty sure the current arrangement is the result of input of others, not my idea. Maybe someone (Spiro ?) wants to speak up on this.

spiro-trikaliotis:
> I'm pretty sure the current arrangement is the result of input of others, not my idea. Maybe someone (Spiro ?) wants to speak up on this.

I do not object against this. It means I have to commit a change on OBS, but that's all. So, whatever you like the most is good for me.

greg-king5:
I introduced "`prefix`" to cc65's makefiles, back when Uz ran the project.  I chose lower-case because I was used to seeing it in the Autotools system, and because I'm lazy (I wanted to avoid using Shift while typing it).  It was a long time before the other build/install variables were added; consistency wasn't an issue back then.

Now, it makes sense to use upper-case for those names.  They customize the packaging process.  They're passed between scripts and makefiles.  They're environment variables; such variables are upper-case traditionally.

oliverschmidt:
@Spiro, @Greg: Thanks for your quick and helpful feedback. I deduct from it that the pull request okay to be merged as-is.

@Spiro: I'd of course appreciate if you could make the necessary adjustments on OBS.

spiro-trikaliotis:
> I'd of course appreciate if you could make the necessary adjustments on OBS.

done, at least for Debian. The RPM part does not build since some time. Again, if there is anyone with experience in generating RPM .spec files, I would appreciate any help!

oliverschmidt:
>> I'd of course appreciate if you could make the necessary adjustments on OBS.

> done, at least for Debian.

Thanks :-)

> The RPM part does not build since some time.

Okay, good to know...

PR #471, opened 2017-07-28
Added ProDOS file types and auxiliary types

gungwald:
I created apple2_filetype.h following the existing pattern in cbm_filetype.h. apple2_filetype.h is included in apple2.h, just as in cbm.h. I also updated the .gitignore to exclude the swap files that vim creates.

I have not yet completed the documentation updates I suggested. I will do those next.

All the types are prefixed with PRODOS, rather than APPLE2, to differentiate them from DOS33 types, which potentially have different values. I have not added the DOS33 types, but that could be done. Although it wouldn't be useful because the open/fopen functions only work for ProDOS as far as I know. But, that could change in the future, so I didn't want to create a limitation that had to be fixed later.

I tried to connect the file types with their associated auxiliary types by including the 3 character file type in each of its auxiliary types. This is a "self-documenting" feature to help the programmer know which auxiliary types are associated with a file type. 

For example, the auxiliary types for SND include the SND 3 letter code.

`#define PRODOS_T_SND    0xD8 /* Apple IIgs audio             */`

`#define PRODOS_AUX_T_SND_AIFF           0x0000 /* AIFF                     */`
`#define PRODOS_AUX_T_SND_AIFF_C         0x0001 /* AIFF-C                   */`

If there's something that doesn't fit with your vision of how this should work or something that you don't like, let me know. I can change it.


.gitignore
```diff
@@ -7,3 +7,6 @@
 /testwrk/
 /wrk/
 /cc65.zip
+
+# Vim swap files
+.*.swp
```
oliverschmidt: We only ignore files created by the tool chain(s) mentioned in the Makefiles. Not files created by random tools. Please revert your change to .gitignore.

gungwald:
I wasn't sure what to specify for the author(s) in apple2_filetypes.h so I left it the same as the other files.

oliverschmidt:
> I created apple2_filetype.h following the existing pattern in cbm_filetype.h.

Great :-) Looking at https://github.com/cc65/cc65/blob/master/include/cbm_filetype.h#L94 I wonder if it wouldn't be good to move https://github.com/cc65/cc65/blob/master/include/apple2.h#L138-L143 to the new file...

.gitignore
gungwald: OK, will do.

oliverschmidt:
> All the types are prefixed with PRODOS, rather than APPLE2, to differentiate them from DOS33 types, which potentially have different values.

Ok.

> I have not added the DOS33 types, but that could be done. Although it wouldn't be useful because the open/fopen functions only work for ProDOS as far as I know.

Correct.

> But, that could change in the future, so I didn't want to create a limitation that had to be fixed later.

I see and follow your thoughts.

> I tried to connect the file types with their associated auxiliary types by including the 3 character file type in each of its auxiliary types. This is a "self-documenting" feature to help the programmer know which auxiliary types are associated with a file type.

Nice :-)

oliverschmidt:
> I wasn't sure what to specify for the author(s) in apple2_filetypes.h so I left it the same as the other files.

I see you fully entitled to put your name in there.

oliverschmidt:
You might consider to change the comment in https://github.com/cc65/cc65/blob/master/include/apple2.h#L142 to refer to the corresponding macro instead of the "magic constant" `6`.

include/apple2_filetype.h
```diff
@@ -0,0 +1,314 @@
...
+#define PRODOS_AUX_T_LBR_APPLELINK_ACU  0x8001  /* AppleLink ACU            */
+#define PRODOS_AUX_T_LBR_SHRINKIT       0x8002  /* ShrinkIt                 */
+
+/* LBR auxiliary types */
+#define PRODOS_AUX_T_ATK_EASYMNT_ALIAS  0x0000  /* EasyMount alias          */
+
+/* BAS auxiliary types */
+#define PRODOS_AUX_T_BAS_PGM_LOAD_ADDR  0x0801  /* Applesoft pgm load addr  */
+
+/* The file stream implementation and the POSIX I/O functions will use the
```
oliverschmidt: Please prepend the `Variables` section header (with all its newlines ;-) here.

include/apple2_filetype.h
```diff
@@ -0,0 +1,322 @@
...
+/*****************************************************************************/
+
+
+
+/* The file stream implementation and the POSIX I/O functions will use the
+** following variables to determine the file type and auxiliary type to use.
+** This applies specifically to the fopen and open functions.
+*/
+extern unsigned char _filetype;  /* Default: PRODOS_T_BIN */
+extern unsigned int _auxtype;    /* Default: 0 */
```
oliverschmidt: As you did so nicely in the rest of this file you might want to align the closing comments here too.

gungwald:
i have made all the suggested updates and updated related comments as appropriate.

include/apple2_filetype.h
gungwald: Right. Will do.

greg-king5:
> We ignore only files created by the tool chain(s) mentioned in our Makefiles, not files created by random tools. Please revert your change to `.gitignore`.

Anyway, you want to do that kind of ignoring system-wide (for all of your repositories).  Therefore, for Git, you should put that pattern into
**`~/.config/git/ignore`**

oliverschmidt:
> Anyway, you want to do that kind of ignoring system-wide (for all of your repositories).

Thanks for the clarification. The editor used is an attribute of you as person using your system. The tool chain used is an attribute of the project and its Makefiles.

gungwald:
I did not know about ~/.config/git/ignore. I thought everything had to be done in .gitignore. :-) Thanks for the explanations. Makes sense.

greg-king5:
The same is true for `.gitattributes`.  Your personal site-wide Git attributes should be put into
`~/.config/git/attributes`

PR #472, opened 2017-07-29
Added doc for file types and aux types

gungwald:


doc/apple2enh.sgml
```diff
@@ -517,6 +517,14 @@ url="ca65.html" name="assembler manual">.
 </descrip><p>
 
 
+<sect1>Specifying file types for fopen<p>
+
+See section 
+<url url="apple2.html#ss9.4" name="Specifying file types for fopen">
+in the apple2 docoumentation.
```
greg-king5: Typo: `docoumentation` -> `documentation`.

greg-king5:
I see that `write()` changes line-feeds into carriage-returns.  But, that action will cause trouble when programs actually do want to write binary files!  Maybe instead, the Apple targets should use a translation table for character/string literals, as some other targets do.

oliverschmidt:
> I see that write() changes line-feeds into carriage-returns.

I presume you are refering to https://github.com/cc65/cc65/blob/master/libsrc/apple2/write.s#L77 but this code is only for stdout/stderr.

> But, that action will cause trouble when programs actually do want to write binary files!

Sorry, but do you really think that I'd create a C library that can't read/write binary files? 

> Maybe instead, the Apple targets should use a translation table for character/string literals, as some other targets do.

- I don't see how such a table would help getting the decision between text output vs. binary output right.

- I seem to remember that we agreed so far that for the C library it's rather about size than about speed. So I don't see why I should spend 256 bytes (or 512 bytes for both directions) for a translation table.

- Some people may think that it would be a good idea to a have Apple specific char set used by cc65 for literals (like PETSCII for the CBMs) that has the MSB set. However I consider the strings with MSB set rather as using "screen code".

Bottom line: I'm pretty sure that everything in the Apple C library regarding char sets is exactly the way it should be ;-)

doc/apple2.sgml
```diff
@@ -511,6 +511,109 @@ url="ca65.html" name="assembler manual">.
...
+<sect1>Specifying file types for fopen<p>
+
+<descrip>
+
+  <tag>Explanation of File Types</tag>
+
+  ProDOS associates a file type and an auxiliary type with each file.
+  These type specifications are separate from the file's name, unlike
+  Windows and UNIX-like systems which use the file name's suffix (a.k.a.
+  extension) to specify the file type. For example, .exe, .doc, or .bat.
```
oliverschmidt: UNIX-like systems don't really care much about the file name suffix. I.e. https://en.wikipedia.org/wiki/Filename_extension says:

> Programs transforming or creating files may add the appropriate extension to names inferred from input file names (unless explicitly given an output file name), but programs reading files usually ignore the information; it is mostly intended for the human user. It is more common, especially in binary files, for the file itself to contain internal metadata describing its contents.

doc/apple2enh.sgml
oliverschmidt: The `apple2enh` docs deliberately don't refer to the `apple2` docs but duplicate all content. Please adhere to this rule for new content.

doc/apple2.sgml
silverdr: @oliverschmidt Very much true. I recall I once wrote this as a response to some discussion that involved "extensions": “traditional” UNIX (or Amiga) way of recognising what the file contains, is… to check what the file contains (sic!). Strange? [...]

gungwald: Extensions in UNIX are not required or used by the system, but some programs do use them: gcc, make, java, Eclipse, vim, gnotepad. Make in particular has implicit rules that are based on file extensions. Although you could say they are just file name patterns, rather than extensions. The patterns though usually have to do with the suffix of the file. The "file" command does look at the contents of the file, rather than its extension, to determine it's type. Regardless I can remove the reference to UNIX. It's not that important.

doc/apple2.sgml
```diff
@@ -511,6 +511,109 @@ url="ca65.html" name="assembler manual">.
...
+  In contrast, the ISO C function <tt/fopen()/ and the POSIX function
+  <tt/open()/ have no parameter to specify either a file type or an
+  auxiliary type. Therefore, some additional mechanism for specifying
+  the file types is needed. 
+  
+  <tag>Specifying the File Type and Auxiliary Type</tag>
+
+  There are two global variables provided that allow the file type
+  and auxiliary type to be specified before a call to <tt/fopen()/
+  or <tt/open/. They are defined in <tt/apple2_filetype.h/:
```
gungwald: OK

doc/apple2.sgml
```diff
@@ -511,6 +511,109 @@ url="ca65.html" name="assembler manual">.
...
+  extern unsigned char _filetype;  /* Default: PRODOS_T_BIN */
+  extern unsigned int _auxtype;    /* Default: 0            */
+  </verb>
+  </tscreen>  
+
+  The header file <tt/apple2_filetype.h/ also defines many values
+  that can be used to set these variables. It is included in
+  <tt/apple2.h/, which is in turn included in <tt/apple2enh.h/.
+  So it
+  not necessary to include it directly. Just
```
gungwald: Good catch.

doc/apple2.sgml
```diff
@@ -511,6 +511,109 @@ url="ca65.html" name="assembler manual">.
...
+  <tt/PRODOS_T_BIN/. The <tt/_filetype/ variable must be set to
+  <tt/PRODOS_T_TXT/ to create a text file. 
+  
+  For a text file, 
+  <tt/_auxtype/ specifies the record length. A zero record
+  length text file is referred to as a sequential text file.
+  This is equivalent to text files on 
+  other operating systems, except that the line terminator is a
+  carriage return instead of a new line (Linux/BSD/MacOS) or 
+  carriage return, new line pair (Windows).
```
gungwald: UNIX & C documentation typically use "new line" as the "\n" character is an "n" because of that terminology. But, line-feed is fine too.

doc/apple2enh.sgml
gungwald: I don't like copying code or documentation because it increases the amount of work required for changes and increases the possibilities for inconsistencies when you update one and forget about the other. But, I will do it if that's they way you want it.

doc/apple2.sgml
gungwald: I do agree that "new line" is more abstract than line-feed, so you're probably right here.

doc/apple2.sgml
```diff
@@ -511,6 +511,109 @@ url="ca65.html" name="assembler manual">.
...
+  which is defined as zero.
+
+  <tscreen>
+  <verb>
+    &num;include &lt;stdio.h&gt;
+    &num;include &lt;string.h&gt;
+    &num;include &lt;errno.h&gt;
+    &num;include &lt;apple2.h&gt;
+
+    void main()
```
gungwald: Will do.

doc/apple2.sgml
gungwald: I guess you're correct. Officially UNIX doesn't recognize extensions, not like Windows does. So my statement  really wasn't correct with the reference to UNIX in it.

gungwald:
I have made all suggested updates. Thanks for everyone's comments. :-)

gungwald:
I think it would be best to try to find a way to support the standard '\n' translation properly without increasing memory. Maybe there's a way? Rather than a shared translation table, maybe a few additional lines of code in each platform-specific section? I realize I'm not an expert on this and I might be saying something silly, but there still could be a good solution that you haven't thought of yet. I'm just saying, give it some thought.

gungwald:
Adding fuel to the fire... :-) Backspace (\h) is not handled in fgets(s, stdin). I'm not sure what the standard says on this but other C compiles/libraries resolve backspaces on input. cc65 does handle it on output: fprintf, fwrite. Or that is, something is handling it on output. Maybe it's the OS. Not sure.

PR #477, opened 2017-08-31
Add constructors to make the C64 and the VIC-20 repeat all keys.

greg-king5:
This PR makes the Commodore 64 and VIC-20 keyboards repeat all of their keys when those machines execute cc65-compiled programs.  That makes them consistent with cc65's other platforms.  Portable programs can assume that if a machine has a keyboard, then _all_ of its keys probably will repeat.

I couldn't do it for the "pet" platform.  The models with "graphics" keyboards cannot repeat, at all.  The models with "business" keyboards have a REPEAT-key that actively controls the `KBREPEAT_ALL` flag (by means of the firmware).  That flag bit is set to zero while the key isn't pressed.

### DIFF
--- a/asminc/c64.inc
+++ b/asminc/c64.inc
@@ -31,11 +31,10 @@ BASIC_BUF_LEN   = 89            ; Maximum length of command-line
 
 CHARCOLOR       := $286
 CURS_COLOR      := $287         ; Color under the cursor
+KBDREPEAT       := $28A         ; $00 = cursor, $40 = none, $80 = all
+KBDREPEATRATE   := $28B
+KBDREPEATDELAY  := $28C
 PALFLAG         := $2A6         ; $01 = PAL, $00 = NTSC
-
-KBDREPEAT       := $28a
-KBDREPEATRATE   := $28b
-KBDREPEATDELAY  := $28c
 
 ; ---------------------------------------------------------------------------
 ; Kernal routines
--- a/asminc/vic20.inc
+++ b/asminc/vic20.inc
@@ -30,10 +30,9 @@ BASIC_BUF_LEN   = 89            ; Maximum length of command-line
 
 CHARCOLOR       := $286
 CURS_COLOR      := $287         ; Color under the cursor
-
-KBDREPEAT       := $28a
-KBDREPEATRATE   := $28b
-KBDREPEATDELAY  := $28c
+KBDREPEAT       := $28A         ; $00 = cursor, $40 = none, $80 = all
+KBDREPEATRATE   := $28B
+KBDREPEATDELAY  := $28C
 
 ; ---------------------------------------------------------------------------
 ; Screen size
--- a/libsrc/c64/cgetc.s
+++ b/libsrc/c64/cgetc.s
@@ -5,7 +5,9 @@
 ;
 
         .export         _cgetc
+
         .import         cursor
+        .forceimport    keys_repeat
 
         .include        "c64.inc"
 
--- a/libsrc/c64/keyrepeat.s
+++ b/libsrc/c64/keyrepeat.s
@@ -1,0 +1,52 @@
+;
+; Normally, the keyboards on some Commodore models repeat only their "editor"
+; keys (cursor movers, delete, insert, and space).  The keyboards on other CBM
+; models, and many other platforms, repeat all keys, normally.  This constructor
+; makes Commodore keyboards repeat, so that portable programs can assume that
+; all platforms with keyboards repeat all keys.
+;
+; This module is linked by the conio and POSIX input functions.
+;
+; 2017-08-20, Greg King
+;
+
+        .constructor    keys_repeat
+        .destructor     keys_normal
+
+        .include        "c64.inc"
+
+
+;--------------------------------------------------------------------------
+; Put this constructor into a segment whose space
+; will be re-used by BSS, the heap, and the C stack.
+;
+.segment        "ONCE"
+
+; Make all keys repeat.
+
+keys_repeat:
+        lda     KBDREPEAT
+        sta     rptsave
+        lda     #%10000000
+        sta     KBDREPEAT
+        rts
+
+
+;--------------------------------------------------------------------------
+
+.code
+
+; Restore the old keyboard-repeat state.
+
+keys_normal:
+        lda     rptsave
+        sta     KBDREPEAT
+        rts
+
+
+;--------------------------------------------------------------------------
+
+.segment        "INIT"
+
+rptsave:
+        .res    1
--- a/libsrc/c64/soft80_cgetc.s
+++ b/libsrc/c64/soft80_cgetc.s
@@ -7,9 +7,11 @@
 ;
 
         .export         soft80_cgetc
+
         .import         soft80_internal_cellcolor, soft80_internal_cursorxlsb
         .import         cursor
         .importzp       tmp1
+        .forceimport    keys_repeat
 
         .include        "c64.inc"
         .include        "soft80.inc"
--- a/libsrc/c64/soft80mono_cgetc.s
+++ b/libsrc/c64/soft80mono_cgetc.s
@@ -7,10 +7,12 @@
 ;
 
         .export         soft80mono_cgetc
+
         .import         soft80mono_internal_cellcolor, soft80mono_internal_cursorxlsb
         .import         soft80mono_internal_nibble
         .import         cursor
         .importzp       tmp1
+        .forceimport    keys_repeat
 
         .include        "c64.inc"
         .include        "soft80.inc"
--- a/libsrc/cbm/read.s
+++ b/libsrc/cbm/read.s
@@ -1,6 +1,6 @@
 ;
 ; 2002-11-16, Ullrich von Bassewitz
-; 2013-12-23, Greg King
+; 2017-08-07, Greg King
 ;
 ; int read (int fd, void* buf, unsigned count);
 ;
@@ -12,6 +12,10 @@
         .import         rwcommon
         .import         popax
         .importzp       ptr1, ptr2, ptr3, tmp1, tmp2, tmp3
+
+.if     .def(__C64__) || .def(__VIC20__)
+        .forceimport    keys_repeat
+.endif
 
         .include        "cbm.inc"
         .include        "errno.inc"
--- a/libsrc/vic20/cgetc.s
+++ b/libsrc/vic20/cgetc.s
@@ -5,7 +5,9 @@
 ;
 
         .export         _cgetc
+
         .import         cursor
+        .forceimport    keys_repeat
 
         .include        "vic20.inc"
 
--- a/libsrc/vic20/keyrepeat.s
+++ b/libsrc/vic20/keyrepeat.s
@@ -1,0 +1,52 @@
+;
+; Normally, the keyboards on some Commodore models repeat only their "editor"
+; keys (cursor movers, delete, insert, and space).  The keyboards on other CBM
+; models, and many other platforms, repeat all keys, normally.  This constructor
+; makes Commodore keyboards repeat, so that portable programs can assume that
+; all platforms with keyboards repeat all keys.
+;
+; This module is linked by the conio and POSIX input functions.
+;
+; 2017-08-20, Greg King
+;
+
+        .constructor    keys_repeat
+        .destructor     keys_normal
+
+        .include        "vic20.inc"
+
+
+;--------------------------------------------------------------------------
+; Put this constructor into a segment whose space
+; will be re-used by BSS, the heap, and the C stack.
+;
+.segment        "ONCE"
+
+; Make all keys repeat.
+
+keys_repeat:
+        lda     KBDREPEAT
+        sta     rptsave
+        lda     #%10000000
+        sta     KBDREPEAT
+        rts
+
+
+;--------------------------------------------------------------------------
+
+.code
+
+; Restore the old keyboard-repeat state.
+
+keys_normal:
+        lda     rptsave
+        sta     KBDREPEAT
+        rts
+
+
+;--------------------------------------------------------------------------
+
+.segment        "INIT"
+
+rptsave:
+        .res    1


oliverschmidt:
Hi Greg,

I know that this question for sure isn't specific to this PR but nevertheless: Could you please check if the C64 Contiki Soft80 webbrowser still builds/works - given its extremely tight RAM usage. I know that you are familiar with Contiki so I presume this isn't a lot effort for you.

Thanks in advance, Oliver

mrdudz:
not sure if its a good idea to make it so conio behaves different than stdio in this regard - confusing as hell :)

and where is that peekc stuff now? tse :)

groessler:
I also don't know if this is a good idea.

These are just the inherent differences btw. the system, and I don't think we should tape over that. We should keep the default for each platform, that's the way the platform users expect it.

If we _do_ want to tape over that, IMO the repeat should be disabled, since some targets don't support key repeat, and then it's again inconsistent.

And, thought further, shouldn't all platforms initialize the screen colors to the same? A "cross-platform" program might expect a given color scheme....

oliverschmidt:
> not sure if its a good idea to make it so conio behaves different than stdio in this regard

Maybe I'm missing the point but where do you get the impression from that conio behaves different than stdio? One comment explicitly says "This module is linked by the conio and POSIX input functions.". And I see the .forceimport both in cgetc.s and read.s.

silverdr:
I tend to second that what Christian wrote. I'd rather see an explicit call made by the cross-target program to enable those when needed (can effectively be noop on targets that repeat all keys by default) rather than being a default thing, which is not what users of a specific target are well used to. And the example about colours seems to be cogent too.

oliverschmidt:
> If we do want to tape over that, IMO the repeat should be disabled, since some targets don't support key repeat, and then it's again inconsistent.

- If we shut down the screen, disconnect the keyboard and the drives then all machines behave perfectly consistent - in doing nothing anymore ;-)

- There are targets where you can't disable keyboard repeat - i.e. the Apple II.

> And, thought further, shouldn't all platforms initialize the screen colors to the same? A "cross-platform" program might expect a given color scheme....

I'm pretty sure that this was meant to be irony, but I personally don't consider it that way ;-)

oliverschmidt:
I think there are two different aspects on the topic:

1) Cross platform vs. non-cross-platform: Maybe one can compare it to cross-platform-UI tools. The programs built with them tend to not feel native to the platform they run on. I'd say the point one has to face as developer is, that usually a single person uses a program only on a single platform. So the person has no benefit whatsoever from the program behaving consistent over several platforms. He only sees inconsistency with other programs on "his" platform.

2) Expectation vs. Optimization: Again an example: If a C64 program comes with some Turbo DOS thingy that speeds up operations on some drives but doesn't cause problems on drives it doesn't speed up (and doesn't cause other incompatibilities - I guess you get the point) then it likely doesn't match the expectation but the optimization is surely welcome.

So now let's look at the keyboard repeat regarding those two aspects:

For 1) I agree that the keyboard repeat is rather something not asking to be made identical. Why? A program will rather likely not become unusable or even very hard to use just because the presumption made by the author regarding the state of keyboard repeat isn't met. So I personally think that the consistency reason given by Greg as motivation isn't such a strong point. And the fact that the setting doesn't work on all CBMs already shows that the goal of consistency can't be really reached: Some targets never repeat and some targets always repeat.

But for 2) I'm asking the ones who dislike this PR what they think about all cc65 programs switching to lowercase on startup. Isn't that again something not really expected but "just a good thing to do"? Maybe I'm naive but I see that people usually don't keep their fingers lying on the keys of their keyboard. So the experienced C64 users expecting that there's not repeat will not hit the point where the repeat starts. And the other less experienced C64 users (like i.e. me) will be happy to find the keyboard working nicely and consistently for all keys.

So to sum it up: I tend to like the PR but not for the cross platform consistency but for the pure improvement and "cross key" consistency (and I see it similar to the switch to lowercase).

groessler:
>  If we shut down the screen, disconnect the keyboard and the drives then all machines behave perfectly consistent - in doing nothing anymore ;-)

What do you want to say with that? Seems to be an argument _not_ to have forced consisteny.

> There are targets where you can't disable keyboard repeat - i.e. the Apple II.

Good. So we cannot get consistent over all targets. So, what's the point?

> I'm pretty sure that this was meant to be irony, but I personally don't consider it that way ;-)

Yes, it was meant as irony. I would hate to have my Atari programs changing color without me explicitly asking for it.

groessler:
> For 1) I agree that the keyboard repeat is rather something not asking to be made identical. Why? A program will rather likely not become unusable or even very hard to use just because the presumption made by the author regarding the state of keyboard repeat isn't met. So I personally think that the consistency reason given by Greg as motivation isn't such a strong point. And the fact that the setting doesn't work on all CBMs already shows that the goal of consistency can't be really reached: Some targets never repeat and some targets always repeat.

and

> So to sum it up: I tend to like the PR but not for the cross platform consistency but for the pure improvement and "cross key" consistency (and I see it similar to the switch to lowercase).

?? So what do you mean? It seems these are conflicting statements.

oliverschmidt:
>>> If we do want to tape over that, IMO the repeat should be disabled, since some targets don't support key repeat, and then it's again inconsistent.

>> If we shut down the screen, disconnect the keyboard and the drives then all machines behave perfectly consistent - in doing nothing anymore ;-)

> What do you want to say with that? Seems to be an argument not to have forced consisteny.

I wanted to say that even if it would be possible to disable the repeat where it is present by default I'd be against doing so because this is no question of lowest common denominator.

>> There are targets where you can't disable keyboard repeat - i.e. the Apple II.

> Good. So we cannot get consistent over all targets. So, what's the point?

Easy. The point is that your proposal can't be realized.

oliverschmidt:
> ?? So what do you mean?

Hm, rereading my comment I don't see how I can make it any clearer

>  It seems these are conflicting statements.

Sorry, but they are not. Not at all.

silverdr:
The switch to lowercase is a different thing. Many programs switch to lowercase whenever they need to use Upper/Lowercase text, so the user assumes this program to be one of those that do this (even if no use is made of uppercase text afterwards). Changing the repeating mode is another thing and only programs with very specific need tend to do it. In such relatively rare cases it is typically for something very different than text input. An explicit call to enable this when a program needs it would be better as all keys repeating tends to be highly confusing on the targets in question.

groessler:
> Easy. The point is that your proposal can't be realized.

Hmm, I didn't propose anything. Greg proposed this PR to have consistency with the other platforms. Since that cannot work (not supprted by all platforms), I don't see the point of this PR. Maybe I know too little about C64 and VIC20, and it's plain obvious for everyone else...

groessler:
>> It seems these are conflicting statements.

> Sorry, but they are not. Not at all.

Looks linke I'm dense today, but

> I agree that the keyboard repeat is rather something not asking to be made identical. 

and 

> I tend to like the PR but not for the cross platform consistency but for the pure improvement and "cross key" consistency

seem to be conflicting statements. Although, I don't exactly know what you mean with "cross key" consistency.

oliverschmidt:
>> Easy. The point is that your proposal can't be realized.

> Hmm, I didn't propose anything.

You originally wrote "IMO the repeat should be disabled" which is a phrase that I classified as a proposal. Sorry for getting this wrong.

oliverschmidt:
>> I agree that the keyboard repeat is rather something not asking to be made identical.

> and

>> I tend to like the PR but not for the cross platform consistency but for the pure improvement and "cross key" consistency

> seem to be conflicting statements. Although, I don't exactly know what you mean with "cross key" consistency.

My first sentence says that I don't think that - in contrast to Greg - it is important to have all targets behave identical. My second sentence says that I like the PR, _BUT_ _NOT_ because of cross platform consistency reasons. So the two sentences fit perfectly together.

Or to put it in other words: It's the right PR - just for the wrong reason.

> Although, I don't exactly know what you mean with "cross key" consistency.

Cross target consistency means that all targets behave the same - so "cross key" consistency means that all keys of the keyboard behave the same (they all do auto repeat).

oliverschmidt:
> Changing the repeating mode is another thing and only programs with very specific need tend to do it. In such relatively rare cases it is typically for something very different than text input.

Thanks for your input :-)

It surely wouldn't be the first time I just don't have enough CBM background to understand - but I want to understand this !

So I presume we all spend much more time in front of modern machines than we spend in front of retro machines. And I presume that every modern machine keyboard does have auto repeat for all keys. And further I presume that most of the time we (in contrast i.e. to my son ;-) use our keyboard to enter text - i.e. for this discussion. So my maybe stupid question is: What is wrong with having the C64 do just the very same? It can't be that it is a retro machine because other retro machines tend to have auto repeat for all keys and it works just fine.

I already asked in another context: Does the auto repeat we discuss here behave differently than i.e. the one on modern machine keyboards? Or do you exchange your expectation for a different one because of the 'Commodore' on the upper left of the "keyboard" in front of you?

silverdr:
> And further I presume that most of the time we (in contrast i.e. to my son ;-) use our keyboard to enter text - i.e. for this discussion.

:-))

> What is wrong with having the C64 do just the very same? It can't be that it is a retro machine because other retro machines tend to have auto repeat for all keys and it works just fine.

No, it is not about it being a retro machine. It is about it being a specific one, where the default behaviour is different than on other machines, including the retro ones. As an example to tell what I mean in the more modern context - I spent hours (cumulatively - sic!) to teach my ex-wife that pushing the close gadget in the corner of the (current machine, current OS) window in Mac does NOT necessarily close the program and therefore does NOT necessarily free the resources used by the program. After switching out of Windows she just does this automatically and expects that if there is no more windows open, the program must have quit. This is not true and she gets irritated when the machine becomes unresponsive and I get irritated when I get back to her machine and start closing dozens of programs, which weren't closed. She got used to this very much and has hard times remembering (or finds it inconvenient) to use the key combination or menu entry to quit programs. There are such relatively small nuances, which make difference even on non-retro machines. Might sound silly today but it actually took me some time to unlearn pushing shift-cursor to move cursor to the left ;-)

>I already asked in another context: Does the auto repeat we discuss here behave differently than i.e. the one on modern machine keyboards? Or do you exchange your expectation for a different one because of the 'Commodore' on the upper left of the "keyboard" in front of you?

It basically behaves the way we have it currently on modern machines. The question is do we want to make the retro ones behave like the modern ones? If the answer is "yes" then this is all fine. If the answer is "we want to preserve the original feel" then I can only say that back in the days I found it highly annoying. I even had a cheat sheet, where one of the items was a POKE to turn this autorepeat OFF for such occasional situations when something turned it on and didn't clean up after itself, while I - for one reason or another - didn't want to reset the machine. And I certainly wasn't alone in this as even my favourite [DolphinDOS](http://e4aws.silverdr.com/projects/dolphindos2/) added a key combination to switch the mode, without me asking them to do so ;-)

silverdr:
> It basically behaves the way we have it currently on modern machines

One more thing - programs that use this mode, often change also the repeat rate (to high), which - if not cleaned up properly - make typing text difficult.

groessler:
>> seem to be conflicting statements. Although, I don't exactly know what you mean with "cross key" consistency.

> My first sentence says that I don't think that - in contrast to Greg - it is important to have all targets behave identical. My second sentence says that I like the PR, BUT NOT because of cross platform consistency reasons. So the two sentences fit perfectly together.
> Or to put it in other words: It's the right PR - just for the wrong reason.

I don't think it's right. But then, it's just me...


>> Although, I don't exactly know what you mean with "cross key" consistency.

> Cross target consistency means that all targets behave the same - so "cross key" consistency means that all keys of the keyboard behave the same (they all do auto repeat).

Why should they (if they don't do default on the platform)?

I can sse you point a bit, I think. You say, that platforms which cannot repeat, don't repeat (obviously). And platforms, where _some_ keys repeat by default, should repeat _all_ keys. Is this a correct assessment?

But I still don't think the runtile lib should do that unless the programmer asks for it.

oliverschmidt:
> Maybe I wasn't clear. Yes, if we want to have this for the sake of consistency, then we should make it non-repeating,

That's exactly how I understood you in the first place.

> There I didn't know yet that that's not possible on the Apple.

Therefore I pointed it out - nothing more :-)

oliverschmidt:
>> Or to put it in other words: It's the right PR - just for the wrong reason.

> I don't think it's right.

Surely. I just wanted to clarify my reasoning. I did't want to imply that it's "the truth".

> But then, it's just me...

No, it's everybody but Greg and me ;-)

oliverschmidt:
> It basically behaves the way we have it currently on modern machines.

That's what I thought I had understood. So it's nothing "wrong" with it technically.

>  The question is do we want to make the retro ones behave like the modern ones?

That's not the interesting question from my pov. Modern machines more or less just happen to behave that way.

My thoughts are more like this: I guess the C64 Kernal initializes the I/O location at some point during boot with that "cursor only repeat" code. Now let's imagine the programmer back then had decided to rather put the "all repeat" code there. Then _the_ _expectation_ would of course be that all keys repeat. Would we then potentially have now a discussion here asking to "finally get rid of that repeat all bullshit at least during running cc65 programs" ? Or in other words: Do you really think that cursor only repeat is "better"? Or is it "just" because you're used to it?

Maybe it's just me who considers this an interesting question: I know that I have an "uncommon" understanding of things when it comes to "retro feeling". I was maintainer for some years the most popular Apple II emulator. I added a video rendering that looks better than that of the real machine.  After handing over the project the maintainers went into the opposite direction. But likely that's the reasons why emulators have so many video rendering options ;-)

After all this is probably just a question of taste. Again referring to emulators: Some like it that the emulator loads stuff in some Warp mode. Other want to sit there and wait for some minutes listening to emulated drive sounds to get that retro feeling. So you surely guessed by now I'm into the Warp mode ;-)

So how to proceed here? If we can agree that it's rather a question of taste it might make sense to ask on the cc65 mailing list to get some more opinions.

silverdr:
> Or in other words: Do you really think that cursor only repeat is "better"? Or is it "just" because you're used to it?

First - both answers to the question you find not so interesting from your pov are valid. Now - to say whether it is objectively better, we would need to do a thorough research, which we won't do I presume. I personally think it is better (even if now I type on a machine that behaves differently [*]) but I certainly can't be objective here as the machine I used most when I was getting used to computing was a Commodore. I surely am used to it. But so is every user of this "platform". BTW - it's not only cursor keys. It's cursor and space (both forward and back) keys.

\* - to alleviate this problem :-) on current machines I set the repeat delay noticeably higher than default setting. It's not the same as on the Commodore but better. Again, might be objective but might also be a personal habit. We won't know that. One thing is for me sure - a typical CBM user is not expecting all keys to repeat without a clear, specific reason. This comes often as annoying when doing text entry and it surely stems from the fact how those machines behave by default.

mrdudz:
> Maybe I'm missing the point but where do you get the impression from that conio behaves different than stdio? One comment explicitly says "This module is linked by the conio and POSIX input functions.". And  I see the .forceimport both in cgetc.s and read.s.

sorry, i didnt see the read.s :)

as for including it or not - i am not a big friend of it either. not only because it makes c64 programs behave differently than a c64 user would expect - but for the most part because it pulls code into all programs compiled with cc65 that only very few programs would need or even benefit from. IMHO that should never happen.

Marrin:
Which kind of programs rely on the automatic repeat of all keys (_at an arbitrary/non specified rate_!) that would benefit from this?

As long as I don't see a (potential) benefit for a substantial amount of projects, I would opt for a solution with less code, i.e. in this case not guaranteeing consistent key repeat across all targets.

oliverschmidt:
> it might make sense to ask on the cc65 mailing list to get some more opinions.

I did so (https://sourceforge.net/p/cc65/mailman/message/36021530/) and feedback was the very same I already received here.

I now understand that there are Kernals / DOSes which set the auto-repeat for every key. This means two things to me:

1. Greg and I are in general right that it is something that might be desirable.
2. It is something that is rather to be considered an attribute of the environment than an attribute of the program running.

Therefore I'm closing this PR - at least in its current form.

PR #480, opened 2017-09-02
At most 256 bytes will be reserved

phs:


doc/ca65.sgml
```diff
@@ -1695,7 +1695,7 @@ either a string or an expression.
   Example:
 
   <tscreen><verb>
-        ; Reserve space for some data, but 256 bytes minimum
+        ; Reserve space for some data, but 256 bytes maximum
        	savearea:       .min (.sizeof (foo), 256)
```
greg-king5: Well, well ...
I had noticed, before, that the comment describes the wrong limit.  But, I hadn't noticed, until now, that the next line has a syntax mistake: the `.res` directive is missing.  Please put it into that line.

oliverschmidt:

PR #487, opened 2017-09-13
Fixed address in CBM 510 document.

polluks:
See cbm510.cfg


greg-king5:
Yes, the address is set [here](https://github.com/cc65/cc65/blob/master/libsrc/cbm510/crt0.s#L89-L95), and used [there](https://github.com/cc65/cc65/blob/master/libsrc/cbm510/crt0.s#L279-L284).

oliverschmidt:
Thanks :-)

PR #494, opened 2017-09-24
[telestrat] Adding some vars, adding getcwd management, changing ZP address and ZP length

jedeoric:
Theses variables are used in telemon rom


oliverschmidt:
Hi,

The idea of a GitHub pull request is that it's a about acertain topic / feature / attribute / <...>. According to the title this pull request is about adding variable definitions.

From that perspective changes regarding the zero page usage and current working directory support don't fit here.

Regards,
Oliver

libsrc/telestrat/initcwd.s
```diff
@@ -0,0 +1,28 @@
+;
+; Oliver Schmidt, 18.04.2005
```
oliverschmidt: If you don't want to update this section of the file then please remove it altogether. I'm pretty sure I didn't write `BRK_TELEMON XVARS` ;-)

jedeoric:
I changed the title of this pull request. 

This pull request contains some changes for Telestrat target
* getcwd management 
* some variables availables in Telestrat ROM

Please note that i changed stdio.inc in order to fix right FILENAME_MAX for Telestrat target.

cfg/telestrat.cfg
```diff
@@ -4,7 +4,7 @@ SYMBOLS {
     __RAMEND__:    type = weak, value = $9800;
 }
 MEMORY {
-    ZP:      file = "", define = yes, start = $00E0,            size = $001A;
+    ZP:      file = "", define = yes, start = $00B0,            size = $003A;
```
oliverschmidt: Why was $B0-$DF not used before? Was that just a bug that is now fixed? Or does using $B0-$DF imply some limitation / compatibility reduction?

jedeoric: Hello,

$00 to $8B are used by the main telestrat bank (telemon : main 16 KB bank.)
$8c to $CF are used by Hyperbasic language (another 16KB bank)
$D0 to $FF are used by Telematic (for minitel routines).

The first time, i used telematic cardridge ZP.  Now, i grab also Hyperbasic ZP because this cardridge won't work (i don't have source code yet). For instance, i manage this target with telemon primitives only for instance.

If you want more info, you can check this : http://orix.oric.org/doku.php

or 

https://www.youtube.com/watch?v=VKgHpLKjl9c&t=38s

oliverschmidt: Thanks for the clarification.

Is there a specific reaon why you then don't go  for `start = $008C, size = $0074` ?

Note: The reason why other target tend to have a `size = $001A` is (or should be) that they save and restore the ZP locations used. So they want to minimize the save area.

oliverschmidt: Ooops, I just noticed that you save and restore too. That answers my question ;-)

PR #507, opened 2017-10-22
[telestrat] tgi_line & tgi_outtext are available, toascii.s added, doc updated

jedeoric:
tgi_line is now working
tgi_outtext is now working.
toascii added.
telestrat doc updated


libsrc/telestrat/toascii.s
```diff
@@ -0,0 +1,14 @@
+;
+; char __fastcall__ toascii (char c);
```
oliverschmidt: Please remove this file again.

Background: I wasn't aware until now that ctype.h contans a toascii() declaration. However most target libraries don't have a (null-)implementation so far. This is actually good because the current build system allows to have a common (null-)implementation which is automatically overridden by target-specific (non-null-)implementations. I will provide that common (null-)implementation (and remove the already existing (null-)implementations).

oliverschmidt: >  I will provide that common (null-)implementation (and remove the already existing (null-)implementations).

Done: https://github.com/cc65/cc65/commit/bd9b4ef60c9de6c4b2f0bad2a50d624daf066536

jedeoric:
Done, 

Comments changed except for VIA/VIA2 & ACIA because the comment are the register's full name

jedeoric:
Build failed, i need to have a look

jedeoric:
Corrected, sorry for the mistake

jedeoric:
Corrected here in my local repository, i will sent it in the next pull request (joysticks, serial and another tgi function)

PR #510, opened 2017-11-02
Alternative joystick interface for Atmos

polluks:


oliverschmidt:
First of all thanks for the contribution !

I'd consider an addition to the documentation at http://cc65.github.io/doc/atmos.html#ss5.3 as obligatory.

PR #516, opened 2017-11-09
Fix joy mask in gamate

Fabrizio-Caruso:


oliverschmidt:
First of all thanks for providing a pull request to cc65 :-)

I don't want to appear as arrogant but I'm having a really hard to time to believe that your proposed change is correct. JOY_xxx_MASK values are _ANDed_ with the parameter. From that perspective I'm pretty sure that the JOY_xxx_MASK values need to represent binary values with a single bit set:
https://github.com/cc65/cc65/blob/afcfba935594879742d85469d3cb1c38a25fa6f4/include/joystick.h#L59-L66 

Apart from that before my refactoring the actual mask values were provided by the driver itself. Looking that the GAMATE joystick driver before the refactoring I see - as expected - single bit values: https://github.com/cc65/cc65/blob/5dcebbccb65924197ac9078fb90b6d08bd3f6823/libsrc/gamate/joy/gamate-stdjoy.s#L29-L36

Fabrizio-Caruso:
You are not arrogant at all. You are simply right...
What is wrong is joy_read which needs to get inverted bits (bit-wise NOT applied to its output).
The Gamate buttons and gamapad UNset bits instead of setting them.
Sorry for the bad PR...

Fabrizio-Caruso:
I am deleting my branch. What fixes the problem in my tests is taking the bitwise NOT of joy_read. I need to figure out how to do this for the gamate and test it.

oliverschmidt:
If this is in fact the issue / fix, then I'm having a hard time to see how I introduced that issue with my refactoring. I didn't touch the GAMATE joy_read at all.

Fabrizio-Caruso:
The "issue" may have existed before the re-factoring. The issue does not prevent the joystick to be used. It prevents to use the macros as with all oher targets. 
Second tentative solution: May be an EOR #$FF on the joy_read function for the GAMATE may fix it. I am going to do some tests and re-do a PR ONLY if my tests are OK.

oliverschmidt:
I'm in general of course willing to have that EOR#$FF added to the joy_read function. I'm just wondering if really everybody implementing/testing/using it so far never cared that it is totally incompatible with the masks ?!?

Fabrizio-Caruso:
EOR #$FF fixes the problem.
Without EOR #$FF you can still use the masks but in a contorted way: you have to pass the bit-wise not of the output of joy_read. This is inconsistent with what you do with all other targets.
With the EOR fix, you use the masks with GAMATE as with all the other targets.

Fabrizio-Caruso:
I have created a new PR: 
Correct bits in gamate joystick detection #517

Fabrizio-Caruso:
I have tested the fix with CROSS CHASE where, for all targets, I use something like:

	void movePlayerByJoystick(unsigned char joyInput)
	{
		if(JOY_UP(joyInput))
		{
			_DO_MOVE_UP
		}
		else if(JOY_DOWN(joyInput))
		{
			_DO_MOVE_DOWN
		}
		else if(JOY_LEFT(joyInput))
		{
			_DO_MOVE_LEFT
		}
		else if(JOY_RIGHT(joyInput))
		{
			_DO_MOVE_RIGHT
		}
		else if(JOY_BTN_1(joyInput) && guns>0 && !missile._status)
		{
			playerFire = 1;
		}
			_DRAW_PLAYER	
	}

PR #517, opened 2017-11-09
Correct bits in gamate joystick detection

Fabrizio-Caruso:
This fixes GAMATE joystick input: joy_read was producing inverted bits.
This fixes makes GAMATE fully compatible with the other targets.

### DIFF
--- a/libsrc/gamate/joy/gamate-stdjoy.s
+++ b/libsrc/gamate/joy/gamate-stdjoy.s
@@ -79,7 +79,7 @@ COUNT:
 
 READJOY:
         lda     JOY_DATA
-        eor     #$ff
+        eor     #$FF
         ldx     #0
         rts
 
--- a/libsrc/gamate/joy/gamate-stdjoy.s
+++ b/libsrc/gamate/joy/gamate-stdjoy.s
@@ -79,7 +79,7 @@ COUNT:
 
 READJOY:
         lda     JOY_DATA
-	eor     #$ff
+        eor     #$ff
         ldx     #0
         rts
 
--- a/libsrc/gamate/joy/gamate-stdjoy.s
+++ b/libsrc/gamate/joy/gamate-stdjoy.s
@@ -79,6 +79,7 @@ COUNT:
 
 READJOY:
         lda     JOY_DATA
+	eor     #$ff
         ldx     #0
         rts
 


Fabrizio-Caruso:
Remark: With this fix the following code works as expected with the GAMATE target as well as with all other targets:

	void movePlayerByJoystick(unsigned char joyInput)
	{
		if(JOY_UP(joyInput))
		{
			_DO_MOVE_UP
		}
		else if(JOY_DOWN(joyInput))
		{
			_DO_MOVE_DOWN
		}
		else if(JOY_LEFT(joyInput))
		{
			_DO_MOVE_LEFT
		}
		else if(JOY_RIGHT(joyInput))
		{
			_DO_MOVE_RIGHT
		}
		else if(JOY_BTN_1(joyInput) && guns>0 && !missile._status)
		{
			playerFire = 1;
		}
		#if defined(FULL_GAME)
			_DRAW_PLAYER	
	}

oliverschmidt:
Are you aware of https://github.com/cc65/cc65/blob/master/testcode/lib/joy-test.c ? Maybe you want to consider to confirm with it that things behave as you expect.

Fabrizio-Caruso:
@oliverschmidt Thanks! I just tested my fix and it works :-)
Pressing the first button or moving the joystick make the corresponding word become -----
Remark: The test only checks the directions and the first button.

libsrc/gamate/joy/gamate-stdjoy.s
```diff
@@ -79,6 +79,7 @@ COUNT:
 
 READJOY:
         lda     JOY_DATA
+	eor     #$ff
```
oliverschmidt: - We don't want to have tab chars in the cc65 code base.
- The predominant style is that hex chars are uppercase.

I can surely fix those things myself but if you want to see your pull request merged as-is then you need to adjust them...

Fabrizio-Caruso: Ok! I will fix!

oliverschmidt:
> I just tested my fix with joy-test.c (as well as my game) and it works :-)

Thanks for the clarification !

libsrc/gamate/joy/gamate-stdjoy.s
```diff
@@ -79,6 +79,7 @@ COUNT:
 
 READJOY:
         lda     JOY_DATA
+        eor     #$ff
```
oliverschmidt: The predominant style is that hex chars are uppercase.

oliverschmidt:
Thanks !

Fabrizio-Caruso:
Thanks @oliverschmidt !

How frequently are the snapshots in https://sourceforge.net/projects/cc65/files/ created?
Will my fix be available in the next days?

oliverschmidt:
>How frequently are the snapshots in https://sourceforge.net/projects/cc65/files/ created?

They are automatically created on every change.

> Will my fix be available in the next days?

It's available since ~10 mins.

PR #519, opened 2017-11-10
[Telestrat] mkdir is now available

jedeoric:
This pull request contains : 
* Doc correction
* mkdir management

### DIFF
--- a/libsrc/telestrat/sysmkdir.s
+++ b/libsrc/telestrat/sysmkdir.s
@@ -1,0 +1,30 @@
+;
+; Jede (jede@oric.org), 2017-10-27
+;
+; unsigned char _sysmkdir (const char* name, ...);
+;
+
+        .export         __sysmkdir
+        .import         addysp, popax
+
+        .include        "telestrat.inc"
+        .include        "zeropage.inc"
+        
+
+__sysmkdir:
+        ; Throw away all parameters except the name
+        dey
+        dey
+        jsr     addysp
+
+        ; Get name
+        jsr     popax
+        
+        ; Call telemon primitive
+        
+        BRK_TELEMON(XMKDIR)
+        
+        rts
+
+
+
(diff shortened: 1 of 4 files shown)


libsrc/telestrat/sysmkdir.s
```diff
@@ -0,0 +1,30 @@
...
+        dey
+        dey
+        jsr     addysp
+
+        ; Get name
+        jsr     popax
+        
+        ; Call telemon primitive
+        
+        BRK_TELEMON(XMKDIR)
```
oliverschmidt: Does this take care of setting A to a suitable value for https://github.com/cc65/cc65/blob/master/libsrc/common/_mappederrno.s ?

jedeoric: For instance, XMKDIR return 0

It did not manage any error code for instance. Let me know if i need to change something

oliverschmidt: Unfortunately I don't really understand you. Do you want to express that BRK_TELEMON doesn't support any way at all to determine, if an operation was successful?

jedeoric: XMKDIR primitive handles only "EOK" ($00). I added ENODEV when we can not access to SDCARD.

But yes, i did not handle many case in this primitive yet (XMKDIR is only available in Telemon 3.x, which is the version i am doing).

oliverschmidt: Okay, basically I only wanted to make sure that you are aware that the A you return from `sysmkdir` will be fed into `__mappederrno` for further process (inkl. generating the return code for the caller of `mkdir`). Have you checked that the things in `__mappederrno` happen as you expect?

jedeoric: I am not aware of all things to do when i do a pull request, You can continue to ask me if i checked everything, because i don't know all stuffs in cc65 :)

About __mappederrno, i tested here, and it works

oliverschmidt:
> About __mappederrno, i tested here, and it works

Thanks for the feedback.

PR #529, opened 2017-11-21
Added standard joystick driver for the Supervision.

polluks:
Fix docs for #453.


libsrc/supervision/joy_stat_stddrv.s
```diff
@@ -0,0 +1,14 @@
...
+;
+; Oliver Schmidt, 2012-11-01
+;
+; const void joy_static_stddrv[];
+;
+
+        .export         _joy_static_stddrv
+        .import         _supervision_stdjoy_joy
+
+.rodata
```
oliverschmidt: Please remove this superfluous directive.

libsrc/supervision/joy/supervision-stdjoy.s
```diff
@@ -0,0 +1,93 @@
...
+; Driver signature
+
+        .byte   $6A, $6F, $79           ; "joy"
+        .byte   JOY_API_VERSION         ; Driver API version number
+
+; Library reference
+
+        .addr   $0000
+
+; Button state masks (8 values)
```
oliverschmidt: Since the joytick driver interface refactoring a few months ago drivers don't contain state masks anymore. The bit-to-button assignment is now done in the target header file, see e.g. https://github.com/cc65/cc65/blob/master/include/nes.h#L85-L93

If you need/want (for A, B, SELECT, START) button names beyond the the defaults in https://github.com/cc65/cc65/blob/master/include/joystick.h#L59-L66 then you need add the necessary defines as in e.g. https://github.com/cc65/cc65/blob/master/include/nes.h#L95-L103

include/supervision.h
```diff
@@ -105,9 +105,30 @@ extern unsigned char sv_nmi_counter;
 extern unsigned char sv_timer_irq_counter;
 extern unsigned char sv_timer_dma_counter;
 
+/* Masks for joy_read */
+#define JOY_UP_MASK     0x08
+#define JOY_DOWN_MASK   0x04
+#define JOY_LEFT_MASK   0x02
+#define JOY_RIGHT_MASK  0x01
+#define JOY_BTN_1_MASK  0x20
```
oliverschmidt: That JOY_BTN_x_MASK thingy is by no means obligatory. As far as I can see from Supervision pics it's likely no appropriate. I personally would rather define A, B, START and SELECT right away based on the numeric values.

oliverschmidt:

As far as I can see you didn't take care so far of my comment:

> You should define that there aren't dynamic drivers like in e.g. https://github.com/cc65/cc65/blob/master/include/nes.h#L110-L111

polluks:
Please bump JOY_API_VERSION because of incompatibility.

include/supervision.h
polluks: What do you mean? I don't see any difference to the NES...

oliverschmidt: Yes, you are right, please just ignore my comment ;-)

oliverschmidt:
> Please bump JOY_API_VERSION because of incompatibility.

Done with https://github.com/cc65/cc65/commit/6034b68f0681c2edcf999a7f71cbb006f3b6e3e8.

PR #530, opened 2017-11-22
Subversion fallback

polluks:
If you use don't use Git to access the repository, you will still get a revision number.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -55,7 +55,7 @@ endif
 ifdef GIT_SHA
   $(info GIT_SHA: $(GIT_SHA))
 else
-  GIT_SHA := $(shell git rev-parse --short HEAD 2>$(NULLDEV))
+  GIT_SHA := $(shell git rev-parse --short HEAD 2>$(NULLDEV) || svnversion -n)
   ifneq ($(words $(GIT_SHA)),1)
     GIT_SHA := N/A
     $(info GIT_SHA: N/A)
--- a/src/Makefile
+++ b/src/Makefile
@@ -55,7 +55,7 @@ endif
 ifdef GIT_SHA
   $(info GIT_SHA: $(GIT_SHA))
 else
-  GIT_SHA := $(shell git rev-parse --short HEAD 2>$(NULLDEV) || svnversion -n)
+  GIT_SHA := $(shell git rev-parse --short HEAD 2>$(NULLDEV) || svnversion)
   ifneq ($(words $(GIT_SHA)),1)
     GIT_SHA := N/A
     $(info GIT_SHA: N/A)


oliverschmidt:
Quite some effort has been spent to make the cc65 Makefiles compatible with CMD.EXE. I don't see us giving this up for what looks to me like a corner case.

greg-king5:
Windows CMD.EXE can handle the patch.

However, the description needs to be changed.  It implies that this patch is for people who don't put cc65 into any version control system.  But, it really is for people who put cc65 into Subversion instead of Git.

polluks:
@oliverschmidt corner case = me :persevere:

oliverschmidt:
> corner case = me 😣

Rather: corner case = use SVN for a *Git*Hub project ;-)

oliverschmidt:
> Windows CMD.EXE can handle the patch.

Thanks for pointing out!

So I'd like to understand what `svnversion -n` writes to stdout when it succeeds and when it fails.

polluks:
At least the bridge is officially supported, I already changed Oricutron's makefile like this.

oliverschmidt:
> At least the bridge is officially supported

If it weren't I would have closed this pr right away ;-)

But seriously: Please tell me what `svnversion -n` writes to stdout if it fails. Nothing?

polluks:
"exported" i.e. no version control https://groups.google.com/forum/m/#!topic/tortoisesvn/GAPMIv1mvPA

oliverschmidt:
So if I'm not wrong again then the Makefile logic will accept that as version as it's exactly one word. So the reported version becomes "exported" and not "N/A". I personally don't have a problem with that, I just want to understand.

greg-king5:
"`exported`" came from an old revision (1.6) of Subversion.  The latest revision (1.9) outputs either two or six words (usually "`Unversioned directory`").  When the directory is in a Subversion repository, the command returns one (possibly compound) word.  The `-n` option strips off the new-line character (not really needed because `make` also does it).

oliverschmidt:
> "exported" came from an old revision (1.6) of Subversion. The latest revision (1.9) outputs either two or six words (usually "Unversioned directory"). When the directory is in a Subversion repository, the command returns one (possibly compound) word.

Thanks for the detailed description - that's the very info I was asking for :-)

> The -n option strips off the new-line character

I read that in its docs...

> (not really needed because make also does it).

Exactly, so let's drop it.

@polluks: Please provide a short feedback, if you are willing to update this pr with `-n` removed or if I shall adjust that afterwards.

oliverschmidt:
@greg-king5: Looking at the commit message of https://github.com/cc65/cc65/commit/c9438ae1a74c7564f98b5c995ccc1a8f212053fd I guess that the  `2>$(NULLDEV)` isn't necessary anymore now that the `||` forces shell usage. What do you think?

spiro-trikaliotis:
When do we add other VCS? What about cvs, mercurial, bazaar, ...?
What value does this "version" have if one cannot even map it to a version in the official git repository? I mean, if we know that ca65 was built from r24544, what can we make from it?

You see, I do not see any benefit in this PL. However, I see more work for support in the future.
Thus, from my point of view, I do not like the patch and I would reject it.

Oliver, one from your goals when you took cc65 over was to get rid of any superfluous parts, wasn't it? You did not care about things others want to use, if you do not see any value in it.
Isn't this patch a step in the wrong direction, back to where you started from?

polluks:
@spiro-trikaliotis Please calm down! Since years I was using subversion and you didn't notice it.
The makefiles are already transparent of the version control, it was just this single line.
Don't judge me for using a "retro" tool, if GitHub stops svn support I'll revert it, okay?

spiro-trikaliotis:
@polluks  ;)
I don't mind if you use subversion. Its your choice. I am using "git svn" on others people's svn myself, I know this "I like my tool" attitude.
But, I would never fathom to supply a patch in order to support git instead of svn in their build system. If the official repository uses the VCS "X", then I respect this. If I talk with them about their repository, I look up the svn revision and speak about it (instead of using git hashes, which would not make sense to them).
I am in their house, I am their guest, and I do not start to change the color of their carpet (cf. https://blogs.msdn.microsoft.com/oldnewthing/20091202-00/?p=15823).

polluks:
A link to a Micro$oft blog? Sorry, I'm part of the free world (at least at home). :-P

oliverschmidt:
Hi,

The only reason why I considered to merge this pr in the first place was
that SVN is officially supported by GitHub! That fact together with the
size of the patch made me merge it.

If however „all“ contributors but one are strongly against this patch I‘m
perfectly willing to revert it.

Regarding this patch not fitting to my policy of what belongs in this repo
and what not I‘d like to remember everybody that I personally didn‘t like
the whole version thingy - but I was convinced of the opposite. So to me
personally this SVN support isn‘t a matter of black and white but a matter
of shades of grey ;-)

Regards,
Oliver

greg-king5:
> Looking at the commit message of c9438ae1a74c7564f98b5c995ccc1a8f212053fd, I guess that the  `2>$(NULLDEV)` isn't necessary anymore, now that the `||` forces shell usage. What do you think?

`$(shell )` doesn't suppress stderr; we see error messages.  I think that Git errors would confuse people -- especially "`fatal: Not a git repository`".  If they are like me, then they assume that the word "**fatal**" means that the build process crashed.  But, in this context, it has almost no consequence.  Therefore, I believe that we should hide those messages.  People should see only the progress of the build, and problems that interfere with the build.

spiro-trikaliotis:
What if git is not installed, and svnversion is not installed, either? Will we see an error message from the shell that svnversion cannot be found?

polluks:
Indeed, we may also add `2>$(NULLDEV)` for svnversion.

oliverschmidt:
> $(shell ) doesn't suppress stderr; we see error messages.

Ah, I forgot about that, I'm getting old. I should have written that info in that commit message as well...

> What if git is not installed, and svnversion is not installed, either? Will we see an error message from the shell that svnversion cannot be found?

I've done some tests:

With CMD.EXE every failure to execute a program results in
```
'<program>' is not recognized as an internal or external command,
operable program or batch file.
```
to be printed to stderr.

With "another" Shell it results in
```
/bin/sh: <program>: command not found
```
to be printed to stderr.

So without git nor svnversion available we see now two error messages instead of one.

oliverschmidt:
> But, in this context, it has almost no consequence. Therefore, I believe that we should hide those messages. People should see only the progress of the build, and problems that interfere with the build.

Thanks for your thoughts. I of course agree with you.

oliverschmidt:
> Indeed, we may also add 2>$(NULLDEV) for svnversion.

@greg-king5: Do you have an opinion on this proposal?

greg-king5:
> > Indeed, we may also add 2>$(NULLDEV) for svnversion.
>
>@greg-king5: Do you have an opinion on this proposal?

I agree with him.  Svnversion's error messages aren't as alarming as Git's; but still, we shouldn't see them.

oliverschmidt:
Done with https://github.com/cc65/cc65/commit/e485e96de09b3913465af5e942488e202327c7ad.

Issue #526, opened 2017-11-18
Feature request: wait beam function


spektro:
I'm requesting a function for waiting the vertical beam to reach a given raster line on the Commodore Plus/4. I'd like to synchronize drawing with the vertical beam to avoid flickering.

groessler:
Nice attitude "I'm requesting"...

Use `waitvsync()`...

spektro:
Not making any demands, just asking.

I found the waitvsync() macro from the fire.c sample program, thanks!

Here is my C16/Plus/4 version of the macro:
`#define waitvsync() while(!(TED.rasterline_hi&0x01))`

It seems that drawing to bitmap takes so long that beam syncing doesn't eliminate flickering :(

greg-king5:
`waitvsync()` is declared by `<cbm.h>`.

polluks:
@spektro an alternative method may be double buffering...

spektro:
@greg-king5
So it is :) I missed the `cbm.h`'s `waitvsync()` because I didn't have the latest cc65 version installed.

@polluks
I'm trying to avoid double buffering because it would require over 20kB of memory (two bitmap screens).

PR #531, opened 2017-11-22
Fix docs for #526

polluks:
funcref.sgml is still growing and growing, how about outsourcing all TGI functions into tgi.sgml?


oliverschmidt:
> funcref.sgml is still growing and growing, how about outsourcing all TGI functions into tgi.sgml?

So this would mean another entry in the section "3. Library information and other references"? That seems like a good idea to me. One could extend that approach to the other driver-based functions (joy_, mouse_, ser_, ...) as well. After that funcref.sgml should be back to a healthy size.

I'd appreciate pull requests moving in that direction.

PR #532, opened 2017-11-22
Add peek functions to the conio library.

greg-king5:
Here is my Work-In-Progress effort to add character and string peeking functions to cc65's **conio** library.  The functions are named:
```C
cpeekc()        cpeekrevers()
cpeekcolor()    cpeeks()
```
So far, I have a lot of stuff for the checked targets; and, (very) few things for the not-checked targets:
- [x] Apple II -- [merged]
- [x] Atari
- [ ] Atari 5200
- [x] Atmos
- [x] CBM -- [merged]
- [x] Commander X16 -- [merged]
- [ ] Creativision
- [x] Documentation -- [merged]
- [ ] Gamate
- [ ] GEOS
- [ ] NES
- [x] OSI Challenger 1P
- [x] PC Engine
- [ ] Supervision
- [ ] Telestrat


polluks:
Atari 2600? You may ignore platforms without conio.

groessler:
I'm wondering why we don't consider these improvements:

1. scroll can scroll up or down
2. cpeekc can retrieve more than one char

Number 1) would be handy when writing a full screen file viewer or editor or something similar.
Number 2) might actually be a different function, cpeeks. This could be added when demand arises.

I mentioned these things in the discussion then, but they were somehow ignored.

groessler:
>>    I mentioned these things in the discussion then, but they were somehow ignored.

> Really?

I think so. I'll have to dig up the email or comment.

> Maybe it wasn't expressed clearly enough but at least to me scroll() is something we want to do for sure. It's the very sweet spot between the current situation (no scrolling at all) and adding implicit scrolling to the output functions (considered to expensive). With an explicit scroll() the necessary code comes only in when needed.

So you agree with a scroll() function capable of scrolling up or down?

> Regarding cpeeks() I personally can't say much because I personally understand neither the use case of cpeekc() nor cpeeks() well ;-)

Me neither :-)  Let's ignore cpeeks() here.

>However I seem to understand that both proposals don't contradict what is under discussion here. So it can as well go into another discussion - and this discussion:

>    is already big enough
>    contains what is necessary for that Tetris project

>so it seems preferable to me to not mix those proposals in here.

But this PR implements a scroll() which can only scroll up. I think, at least, the function definitions needs an argument to indicate the direction. To have the API in place. Hmm, but if not all directions are implemented it would also need a return code to indicate "not implemented" or such. Better have it completely implemented from the start.

oliverschmidt:
Sorry I didn't look close enough at this pull request...

> Me neither :-) Let's ignore cpeeks() here.

I just noticed that cpeeks() _IS_ part of this pull request - so everything is just fine.

> But this PR implements a scroll() which can only scroll up.

Now I see your point. My initial thought is that some people might argue that scrolling one line up might easily implementable with some ROM/OS routine. If one the other hand scrolling is implemented by manually moving bytes around then it is likely just as easy to allow scrolling in both directions and even multiple lines at once. That would ask for a signed char parameter. But maybe it would be best to remove scrolling from this pull request and really discuss this question in another one...

groessler:
> Sorry I didn't look close enough at this pull request...

Me neither, obviously...

> I just noticed that cpeeks() IS part of this pull request - so everything is just fine.

... missed that. Sorry for the noise.

>Now I see your point. My initial thought is that some people might argue that scrolling one line up might easily implementable with some ROM/OS routine. If one the other hand scrolling is implemented by manually moving bytes around then it is likely just as easy to allow scrolling in both directions and even multiple lines at once. That would ask for a signed char parameter. But maybe it would be best to remove scrolling from this pull request and really discuss this question in another one...

Would be fine for me.

mrdudz:
i agree, scrolling should be handled in a seperate PR. it wasnt part of the original cpeek...() discussion either. i am not sure what cpeekrevers() is or why it is necessary either (it was not in my original code/proposal and i dont remember it from the discussion).

that said, i am a bit confused about what is implemented for the different targets and what isnt.

i am completely lost when it comes to how i'd contribute to this too - i'd like to get it merged ASAP however :)

oliverschmidt:
> i am not sure what cpeekrevers() is or why it is necessary either (it was not in my original code/proposal and i dont remember it from the discussion).

Fortunately we don't depend on your memory let alone mine ;-)
https://sourceforge.net/p/cc65/mailman/message/35873183/

Below the full text:

---

Hi,

I've been thinking about this once more and it seems to me that it's more appropriate to our resource limited machines to rather offer lower level, finer granularity functions that the user can than combine to build higher level functions from on his own.

Of course those C written higher level functions don't perform as well as shrink wrapped ones from the C library but that seems nevertheless less problematic than "abusing" higher level C library functions for lower level tasks. And the quick idea of "let's just have both" simply doesn't fill the bill when looking into the implementation details.

Looking at movetext() brought up the idea in my mind that an _explicit_ scroll function might be the right thing to add to cc65 CONIO. So cputc() and friends don't get bloated but there's some high performance scroll available to the users needing it. However the cc65 CONIO doesn't have any notion of windows/viewports or alike so far. So movetext() allowing to move any rect anywhere is maybe over-engineered and its flexibility means that any 6502 implementation makes the typical use case of complete-screen-vertical-scrolling-up-one-line larger/slower than necessary.

As a result I don't consider the <...>text() functions as the best approach anymore but rather see some "peek" functions as discussed in the beginning of this thread.

Given that now some motivation for the details:

The cc65 CONIO has textcolor() and revers() so it seems desirable to be able to read back those very attributes on that very abstraction level beside the chars written. However I don't see some color or reverse peeking to advance the cursor. From a consistency perspective (and my personal gut feeling ;-) I don't see char peeking to advance the cursor too. But what does that mean for some cpeek<...>xy() function? Does it leave the cursor where it was before or does it leave the cursor at the location given? I think the former is more appropriate but that could/would mean that cpeek<...>xy() can't easily just internally call gotoxy(). The simple solution is to just not offer cpeek<...>xy(). Instead all peeking (char, color, reverse) happens at the current cursor position - which is left untouched. Peeking at a specific position means for the user a first gotoxy() then cpeek<...>() sequence. This doesn't actually imply that much overhead over a single cpeek<...>xy() call. But it makes obvious to the user what happens. And it allows to use gotox() / gotoy() which might result even in a performance gain as gotox() is at least on some targets significantly cheaper than the other/implicit gotos.

The primary reason for wanting cursor advancement was reading several adjacent chars. I see that we can can easily have a string peek function that takes care of that even more efficiently.

So we end up with this set of functions:

char cpeekc();
unsigned char cpeekcolor();
unsigned char cpeekrevers();
void \_\_fastcall\_\_ cpeeks(char *s, unsigned char length);

The values returned can be described pretty easily: Feeding them as-is back into cputc(), textcolor(), revers() and cputs() should result in the original screen content. In other words: The peek functions are supposed to implement the inverse of any transformation done in cputc(), textcolor(), revers() and cputs().

Details:

- cputs() wraps to the next line if the strings is too long to fit in the current line. I don't know if it's worth the effort to allow cpeeks() to continue reading from the next line. I'd like to discuss this aspect with the actual implementers.

- I know at least one target that can't convert chars put on the screen via hline()/vline() back to something that would again result in the same lines when using it with cputc(). I suggest it's just undefined what cpeekc() returns for chars put via hline()/vline().

The explicit scroll function moves the whole screen content one line up, adds an empty line at the bottom and moves the cursor one line up. In line with the naming of clrscr() I see it as:

void scrollscr();

A final remark: When the mouse callbacks don't use sprites/overlays/<...> but exchange chars on the screen then there's a risk that the peek functions read the mouse cursor and that scrolling creates a second stale mouse cursor. I suggest to ignore these issues in the implementation of the functions discussed here and add a hint in the documentation that in this scenario the functions needs to be wrapped by mouse_hide() and mouse_show().

I'd like the ones who actually want to implement the functions above for some target to speak up now. Maybe not surprising I see myself implementing them for the Apple II - but iff there's a moment for other targets.

Regards,
Oliver

oliverschmidt:
Hi,

> that said, i am a bit confused about what is implemented for the different targets and what isnt.

Surely. This pull request was created because we asked eagerly for the current work-in-progress content. So that's what we got: Work-in-progress meaining we have to look at the source for an answer to that question.

> i am completely lost when it comes to how i'd contribute to this too - i'd like to get it merged ASAP however :)

Okay, here's my proposal:

This pull request contains scroll stuff (which is by the way perfectly correct as it _was_ part of our discussion in summer) which we'd rather like to see (re-)discussed in a different thread. That fact and the partially unclear implementation state on some targets makes me consider this pull request as a resource to pull stuff from.

So you/someone could decide to go for the CBM machines:

- Get all cross target files like documentation, header file, test etc.
- Remove scroll stuff from those files
- Get all CBM specific files
- Check if things work as expected on the various CBM machines (including but maybe not limited to https://github.com/greg-king5/cc65/blob/38a20f71d92ba265a8c4c8535302552c20a82a17/testcode/lib/cpeek-test.c)
- Create a new pull request from the content listed so far

Then others could pick their favorite target(s), pull relevant content from this pull request and create additional pull requests...

Just my two cents,
Oliver

greg-king5:
I think that we should have _two_ scrollers.
1. `scrollscr()` is smaller because it will do only what Oliver described -- most people won't need more than that.
2. If someone does need the "bells and whistles", then he will call a different, and bigger, function,
`scroll2()` for example.

I now realize that this is supposed to be a **_peek_** Pull Request -- `scrollscr()` doesn't belong here.  I'll extract it; and, move it to another PR.

mrdudz:
ok, so i stripped out the CBM related things and here they are: https://github.com/cc65/cc65/pull/545

mrdudz:
ok so.... perhaps greg could now delete the docs and CBM related stuff from this PR, so we can see whats left? :)

oliverschmidt:
...or the other contributors feeling responsible for "their" machine go ahead, get the files for their target from this pull request, test the code and provide individual pull requests. As far as I see the implementations for

* Atari
* Atmos
* OSI Challenger 1P
* PC Engine

look complete.

I imagine that those contributors would be (even) more motivated if that Tetris game would become available. According to my understanding it should now work on (several of) the CBM machines.

BTW: I committed to contribute implementations for the Apple II. I'll certainly stick to that. But not just right now...

mrdudz:
> I imagine that those contributors would be (even) more motivated if that Tetris game would become available. According to my understanding it should now work on (several of) the CBM machines.

it works on all of them (vic20 and c16 have to be expanded).... yes, i will push it to github, not sure if i will find the time to clean it up a bit more etc before xmas though :)

btw, whats also missing is an implementation for the c64 soft80 stuff. i actually have that laying around from the original code, i will have a look at making it work with what is in the tree now

oliverschmidt:
@mrdudz: This whole conio peek<...> functions momentum was driven by your Tetris clone requiring them. I seem to remember that you wanted to contribute that Tetris clone using them. And I stated that if that Tetris clone working on the C64 with the conio peek<...> functions would be actually available, I would implement at least enough of the peek<...> functions on the Apple II to make the Tetris clone run there too.

Is there any chance that this Tetris clone using the conio peek<...> functions will surface at some point?

To be honest, I'm rather unhappy that
* the current amount of functionality is pretty different for the "main" targets
* this pull request comprising quite some effort from @greg-king5 hangs around here year after year

polluks:
Anyway conio peek should be no requirement if you have a local copy of the playfield. You just "peek" the playfield instead of the video RAM.

mrdudz:
> Is there any chance that this Tetris clone using the conio peek<...> functions will surface at some point?

sure, i am working on it from time to time... it just doesnt have very high priority for me, and time is limited :) perhaps i should at least put it in a public repo for the time being, so everyone can see how ugly it is =)

> Anyway conio peek should be no requirement if you have a local copy of the playfield. You just "peek" the playfield instead of the video RAM.

i'm not using it for that - it's used for generic "screen fader" type of effects (like something similar to the popular "doom melting" effect)

mrdudz:
@oliverschmidt here it is: https://github.com/mrdudz/portris ... very messy right now, but compiles with current cc65

oliverschmidt:
@mrdudz: ...and here's my contribution: https://github.com/cc65/cc65/commit/04cc463452de739262b91ef1fa4757ad249d6762 ;-)

mrdudz:
i comitted my conio peek functions for PCE. I now see they were also dangling around in this PR - and i wonder why.

What's the deal with the other stuff here?

oliverschmidt:
> What's the deal with the other stuff here?

In https://github.com/cc65/cc65/pull/532#issuecomment-352202951 I wrote:

> the other contributors feeling responsible for "their" machine go ahead, get the files for their target from this pull request, test the code and provide individual pull requests

Nobody opposed so that's still what I consider agreed upon. From that POV your new PCE PR is just fine. Regarding the other targets someone could try to motivate the usual suspects...

mrdudz:
I will look at them one by one sooner or later, now that the tetris thing is there.... but someone else doing it independently from that would obviously be preferred :)

mrdudz:
@greg-king5  could you comment on the current state of all this and perhaps update the first post (maybe it'd also make sense to close this one alltogether?)

that tetris thingy i wrote is also on github now, btw: https://github.com/mrdudz/portris - it may serve as a testbed for this stuff (and obviously i am open for PRs to cover targets i didnt consider yet)

mrdudz:
So, i updated the PR a bit. The remaining stuff appears to be ok for me - although perhaps the NES and/or PCE things should be tested first? :) (Why was it a good idea again to use a: for vdc regs on pce?)

mrdudz:
ok so, i am closing this one (finally) - #1810 has everything from this pr (a bit massaged and tested even)

PR #541, opened 2017-11-30
Fix bugs in the Tiny Graphics Interface's document.

greg-king5:
This Pull Request fixes a bug that causes warning messages from LinuxDoc Tools.

And, it puts some descriptions into proper order, sorted by function name.


oliverschmidt:
Thanks for - once again - taking care of the details :-)

Issue #511, opened 2017-11-05
OSIC1P bug on kbhit: it probably consumes the input


Fabrizio-Caruso:
The OSIC1P target has an issue with the keyboard input.

My game CROSS CHASE would simply freeze waiting in vain for a key press.
The game can be made sort of playable (by avoiding some code) only in turn-based mode.

It looks like as if kbhit would detect the key press and consumes the key press.
A subsequent cgetc waits in vain.

smuehlst:
What I can reproduce is that kbhit() blocks in all calls but the first one. I will try to analyze the problem.

Fabrizio-Caruso:
kbhit should not consume the first key-press, i.e., cgetc should read it without waiting.
This is the normal correct behaviour:
1. the key-board is pressed (e.g., key "A")
2. kbhit becomes TRUE
3. cgetc reads "A"

cgetc is correctly supposed to wait for a key-press if no key press is in the buffer but kbhit should NOT remove the detected key press from the buffer

TEST: The following code should do what its name suggests
``
		void WAIT_PRESS(void)
		{
			while(kbhit())
				cgetc();
			while(!kbhit())
			{ 
			}; 
			cgetc();
                   }
``

P.S.: In the latest version of CROSS CHASE for the OSI C1P target I have removed some code in order to check whether the rest of the game works, and it DOES! :-) So I only see a problem with the keyboard input.

Fabrizio-Caruso:
@smuehlst you can simply test with the function WAIT_PRESS above. Any progress?

Fabrizio-Caruso:
@smuehlst libsrc/osic1p/knhit.s does more than detecting the input, it also reads the value and this is not what kbhit is supposed to do. kbhit should simply return true if any key is pressed (including still pressed and not just newly pressed) and the input buffer should remain unchanged.

Fabrizio-Caruso:
@smuehlst I do not know how the Ohio Scientific C1P works.
I am only trying to guess: the keyboard detection may not be fully supported by the hardware and if this is the case we should simulate what is missing...
I mean that it could be as for the Apple II where the key-release is not detected by the hardware. I am only trying to help you to investigate this issue...

smuehlst:
@Fabrizio-Caruso: There is no progress yet. The implementation of kbhit() was done in collaboration with an other person, and I contacted the person for help. I got feedback, but it will take a while.

[...]

PR #543, opened 2017-12-02
New OSI input routine that fixes issue #511.

jefftranter:
See my comments in the issue for more details.


libsrc/osic1p/cgetc.c
```diff
@@ -0,0 +1,167 @@
...
+ */
+char cgetc (void)
+{
+    unsigned char key;
+    
+    if (lastKey != 0) { // Did we previously get a key?
+        key = lastKey; // If so, use it.
+        lastKey = 0; // Then clear it.
+    } else {
+        while ((key = getkey()) != 0)
```
smuehlst: The loop must be:

    while ((key = getkey()) == 0)
        ; // Wait for key to be pressed.

Otherwise cgetc() will not block.

smuehlst:
cscanf() does not yet work as expected with this new implementation. A single keypress will result in multiple characters being read. My first guess is that some logic is needed to return a character only after the key has been released again.

libsrc/osic1p/cgetc.c
```diff
@@ -0,0 +1,167 @@
...
+// Address of hardware keyboard register.
+#define KEYBOARD 0xdf00
+
+// Some special keys.
+#define ESC 0x1b
+#define CR 0x0d
+#define LF 0x0a
+
+// Lookup table of keys, given keyboard row and bit numbers.
+const unsigned char keys[8][8] = {
```
oliverschmidt: static

libsrc/osic1p/cgetc.c
```diff
@@ -0,0 +1,167 @@
...
+    { 0, ',', 'm',  'n',  'b',  'v', 'c', 'x' }, // Row 2
+    { 0, 'k', 'j',  'h',  'g',  'f', 'd', 's' }, // Row 3
+    { 0, 'i', 'u',  'y',  't',  'r', 'e', 'w' }, // Row 4
+    { 0,   0,   0,   CR,   LF,  'o', 'l', '.' }, // Row 5
+    { 0,   0, '_',  '-',  ':',  '0', '9', '8' }, // Row 6
+    { 0, '7', '6',  '5',  '4',  '3', '2', '1' }, // Row 7
+};
+
+// Table of bitmasks for active keys in each row.
+const unsigned char columnMask[8] = {
```
oliverschmidt: static

libsrc/osic1p/cgetc.c
```diff
@@ -0,0 +1,167 @@
...
+    0xfe, // Row 4
+    0xf8, // Row 5
+    0xfc, // Row 6
+    0xfe, // Row 7
+};
+
+// Bit masks for modifier keys.
+const unsigned char controlMask = 0x40; // Control
+const unsigned char shiftMask = 0x06;   // Left Shift, Right Shift
+const unsigned char lockMask = 0x01;    // Shift Lock
```
oliverschmidt: 3 x static

libsrc/osic1p/cgetc.c
```diff
@@ -0,0 +1,167 @@
...
+
+// Holds last key pressed.
+unsigned char lastKey = 0;
```
oliverschmidt: static

jefftranter:
I will work on addressing these issues and submit a new request.

oliverschmidt:
@smuehlst: Do you consider this pull request in its current state as desirable to be merged?
@jefftranter: Have you tried what was suggested in https://github.com/cc65/cc65/issues/511#issuecomment-348728980? As far as I understand the code with the `!=` instead of `==` can't have worked so any positive test result with that code can't be relevant.

smuehlst:
@oliverschmidt:
> Do you consider this pull request in its current state as desirable to be merged?
No, it has undesired side effects. See my earlier comment about cscanf() not working correctly.

Fabrizio-Caruso:
@smuehlst, repeated key-presses is what people want sometimes at least for cgetc(). 
In most targets cgetc() detects multiple key-presses if a key is kept pressed.
Could you at least leave this behaviour for cgetc() or at leave an option for auto-repeat?
This automatic repeat makes writing a game much simpler.
Without it the player would have to hammer the key to keep moving.

smuehlst:
@Fabrizio-Caruso: The problem is that with the current update the repeat starts **immediately**. If you try a program that uses cscanf() with the new cgetc() implementation, you will see that it is not possible to type a single character. You will get several characters for a single normal keypress, and this completely breaks normal typing to enter text.

oliverschmidt:
> The problem is that with the current update the repeat starts immediately.

This is for obvious reasons a show stopper.

libsrc/osic1p/cgetc.c
greg-king5: Scalar values that never change should be used as either literals, macroes, or enumerations.  Using a variable to hold a constant makes cc65 generate inefficient code.

Fabrizio-Caruso:
So a solution would require a sort of initial delay or even better a different routine for cscanf. Right?

oliverschmidt:
> So a solution would require a sort of initial delay

Yes.

> or even better a different routine for cscanf.

No. cgetc() either doesn't repeat at all or after a delay. Apart from that you wouldn't technically be able to provide a different routine for cscanf.

Fabrizio-Caruso:
@oliverschmidt, not repeat at all, means that no action game could use cgetc and this would mean that a coder would have to re-implement a non-trivial routine that will be there for the most part.
Please provide some sort of detection of keys kept pressed (i.e., auto-repeat) at 
least just as an option to be enabled somehow or implement an alternative solution like a small delay (not necessarily a big initial one) that allows both keys kept to be detected and cscanf to work as expected.

mrdudz:
reminds me of https://github.com/cc65/cc65/pull/452

groessler:
> not repeat at all means that no action game could use cgetc and this would imply that a coder would have to re-implement a non-trivial routine (that is already implemented for the most part).

I think the first and main purpose of cgetc() is to get keyboard input. User typing something on the keyboard.
To use it as joystick "ersatz" is not a general use case. I think best would be to write a joystick driver for the target which handles pressing some keys as joystick input.

Fabrizio-Caruso:
@groessler, I agree, cgetc is meant, above all, to get keyboard input. 
If having auto-repeat costs almost nothing as a default or optional feature, I think it is good to have it. Those early computers often explicitly used the keyboard as input and detecting keys that are kept pressed can be useful even though it is an ersatz. Some of these computers had auto-repeat as default for all keys or some keys. Some other had ways to enable it (even when using standard rom routines).

oliverschmidt:
> I agree, cgetc is meant, above all, to get keyboard input.

:-)

> If having auto-repeat costs almost nothing as a default or optional feature, I think it is good to have it. 

Please read https://github.com/cc65/cc65/pull/452 to see that this opinion is by no means the popular one! Following the predominant opinion on that discussion would mean to conclude: If the OSI ROM input routine doesn't implement an auto-repeat then the cc65 cgetc() shouldn't have an auto-repeat as well, even one with some delay as it is found on other targets - just because the user on this very target has the expecetation to have no auto-repeat.

oliverschmidt:
@Fabrizio-Caruso:  You wrote:

> or even better a different routine for cscanf.

So you seem to see that the requirements of "usual" keyboard input are different from the requirements of an action game. Maybe you haven't realized that this perspective is pretty close to what Chris wrote:

> I think best would be to write a joystick driver for the target which handles pressing some keys as joystick input.

The "only" difference is that you see cgetc() optimized for action games and "usual" keyboard input going for something different while Chris sees cgetc() being optimized for the "usual" keyboard input and action games going for something different ;-)

I presume that you already guess by now which perspective makes more sense to me. If you want to have delay-free auto-repeat keys (which are desirable/necessary for an action game) then a pseudo-joystick driver seems indeed the way to go.

Fabrizio-Caruso:
@oliverschmidt, I agree cgetc should work well for standard input.
Using it for action game is a secondary goal, which should not interfere with the standard input.
I would not want a cgetc that makes impossible to get a normal string input.
cgetc should work by default as standard keyboard input.
I only meant that at least an optional delayed auto-repeat (if obtained at nearly no cost) would be good in my opinion. A reason for this is that using additional code that does almost the same thing is a waste of memory (a few hundreds bytes) on a machine that came with 4k of ram.

oliverschmidt:
Sorry to insist: Have you actually read the discussion on https://github.com/cc65/cc65/pull/452 ?

Fabrizio-Caruso:
By looking at some BASIC games (e.g., a space invaders clone) for the OSI C1P, it looks like the machine had some default auto-repeat.

@oliverschmidt, I had read that thread but maybe I have missed some important points. I understood that that solution to toggle auto-repeat was not working as expected.

oliverschmidt:
> I had read that thread but maybe I have missed some important points. I understood that that solution to toggle auto-repeat was not working as expected.

You're right. The important point is rather in this mailing list thread: https://sourceforge.net/p/cc65/mailman/message/36021530/

smuehlst:
This pull request is superseded by pull request #567 and should be discarded.

Issue #544, opened 2017-12-11
Enhancements to multiplication runtime - Booth's Algorithm for 6502


prandeamus:
First time contributor - be nice, please!
It's been 30 years since I worked with the 6502 but it was a part of my life in those days. On thing I did when I was young was write an implementation of Booth's algorithm for integer multiplication, and it was somewhat faster than conventional multiplication algorithms (on average, there are a few corner cases).
For fun, I've recreated some of that work in some ca65 assembly code and done some simple benchmarks. Is anyone interested in refining this work? I lack the patience to figure out all the cycle times, but my implementation of an 8x8 signed multiply does seem faster that the built-in version in the standard runtime.
I would appreciate some guidance and maybe a collaborator. Yes, I know git moderately well, but am new to the project.

spiro-trikaliotis:
I know the Booth's algorithm, as I wrote. But I always thought that the handling of 2 bits at a moment would need more cycles, even if you use something like a state machine where you just know what the previous bit was.

As far as you do not try to implement Schönhage-Strassen multiplication (https://en.wikipedia.org/wiki/Sch%C3%B6nhage%E2%80%93Strassen_algorithm), everything should be good ;)
Imay be biased, I had some courses from Prof. Schönhage.

I think the best way is to create a pull request, so we can use the mechanisms of Github for the review.

prandeamus:
Ha. I didn't know git and github as well as I thought. I have a local branch with the files, but I can't push that branch. 
"remote: Permission to cc65/cc65.git denied to prandeamus"
I'm a new account; does that make a difference?

mrdudz:
you'll have to create a pull request, you cant just push to the main repo

silverdr:
> On 2017-12-12, at 00:17, prandeamus <notifications@github.com> wrote:
> 
> Ha. I didn't know git and github as well as I thought. I have a local branch with the files, but I can't push that branch.
> "remote: Permission to cc65/cc65.git denied to prandeamus"
> I'm a new account; does that make a difference?

The difference is most probably in your workflow expectations. The typical process looks as follows:

1. you "fork" the main repository to your account
2. you clone your forked repository to your local workstation
3. you make a "feature branch" in your cloned local copy
4. you make your changes/commits in the branch created in [3.]
5. you push changes to your repository
6. you create "pull request" from your feature branch to the upstream repository
7. people review your pull request, discuss it, comment and eventually accept or reject it.

[...]

PR #547, opened 2017-12-12
Implementation of 8x8 and 16x16 booth multiplication for discussion

prandeamus:


prandeamus:
As per issue #544, 8x8 booth multiply implementation for discussion.

prandeamus:
@spiro-trikaliotis for you!

greg-king5:
<cc65.h> declares functions that will give a better comparison between current cc65 code and your Booth implementations.

prandeamus:
Thanks @greg-king5 I'll take a look.

prandeamus:
I've established that the booth algorithm for 8x8->16 bit multiply works, but is slower than the existing code. It's possible that the booth 16x16->32 code can still be optimised, but right now it's just an academic curiosity.

I'm going to close this pull request for now but I will return to my branch some time and see if it can be improved. Code is still available on my fork if anyone is interested.

oliverschmidt:
Thanks for the explanation.

spiro-trikaliotis:
I am sorry that I did not have time yet to have a closer look into your implementation. However, it seems your implementation is not faster than the "school's method", and I am not surprised about this. Years ago, I came to the same conclusion, that it should not be possible to be faster on the 6502 for such small multiplications.

prandeamus:
Well, it was good practice in re-learning 6502 assembly! Booth will only be more efficient if additional code that is required to detect runs of 1s and 0s is offset by fewer addition/subtractions. There's a chance that I can optimise the 16x16 case to get it closer to school method. I guess that the tipping point might be at 32x32->16 bits ... Thanks for your input, though, it was helpful.

Issue #550, opened 2017-12-13
[info][atmos][oric1] Which Atmos-specifc routines to avoid for Orc 1 compatibility?


Fabrizio-Caruso:
Hi

Is the use of rom routines the only reason why the Atmos target is just an Atmos target?
The binary runs on Oric 1 and then crashes (as expected) but it crashes less if certain routines are avoided.

Are arithmetic operations on anything bigger than unsigned char relying on the rom routines?
Maybe one could make it run on Oric 1 if one knew what needs to be avoided.
Probably all conio and input/output routines rely on the Atmos rom and need to be re-implemented at low level to have Oric1 compatibility.

polluks:
Indeed, the reason are the rom routines and their low memory adresses.
The funct key is also missing. http://cc65.github.io/doc/atmos.html#ss7.1

Fabrizio-Caruso:
What I would like to know is what I need to avoid. 

Are rom routines ONLY used for input/output? or also used for arithmetic operations?
If the problem is only input/output, it is possible to use CC65 to code for the plain Oric 1.

I am assuming that the memory configuration is compatible or very similar. 
My game runs (for a few seconds) on the Oric 1.

polluks:
Well, arithmetics are in libsrc/common/ and hello.tap seems to work with Oric-1 :-)

Fabrizio-Caruso:
Thanks! Indeed what breaks on the Oric 1 are the printf directives. If a generic non-target specific implementation exists, it could be used to produce an Oric 1 target with little effort.

Fabrizio-Caruso:
My game works perfectly well on Oric 1 if I remove calls to the predefined sound effects.
I guess the address of the sound routines differ between the two models. It looks to me that the
the atmos target is pretty much an oric target. There could be other cases that are not cover by my game, though.

Fabrizio-Caruso:
So, if no other conflicts between Oric 1 and Oric Atmos roms, it would suffice to rewrite:
```
void atmos_explode (void);
/* Bomb sound effect */

void atmos_ping (void);
/* Bell or ricochet sound effect */

void atmos_shoot (void);
/* Pistol sound effect */

void atmos_tick (void);
/* High-pitch click */

void atmos_tock (void);
/* Low-pitch click */

void atmos_zap (void);
/* Raygun sound effect */
```

oliverschmidt:
Fixed with https://github.com/cc65/cc65/pull/552.

[...]

PR #554, opened 2017-12-17
oops #550 again

polluks:
I made a mistake.


oliverschmidt:
Ah, that adjusts my 6502 world :-) I didn't understand the lda/bvs combo at all.

polluks:
This happens if you just test the first case :-)

PR #557, opened 2017-12-18
Optimizing the stack size

polluks:


oliverschmidt:
Thanks - especially for the creativision.cfg bug fix.

groessler:
The creativision.cfg fix could explain some intermittend problems I saw in the past. Thanks.
But as it is now, this leave only $86 bytes for data + bss. Could we decrease the default stack size? What would be a sane value? $100? Or even $80?

PR #562, opened 2017-12-28
Implementation of getfirstdevice/getnextdevice for Atari.

groessler:
Tested briefly on an emulator (atari800) and more thoroughly on the real thing...


oliverschmidt:
Thanks :-)

Looking forward to a getcurrentdevice() and a getdevicedir()...

PR #564, opened 2017-12-31
Add none.lib

bauen1:
Adds `none.lib` which contains the cc65 library without any platform specific code.
Its only purpose is to be used for porting cc65 to a new platform out-of-tree and the user needs to supply a `crt0.o` module for the library to work.

### DIFF
--- a/cfg/none.cfg
+++ b/cfg/none.cfg
@@ -7,6 +7,7 @@ MEMORY {
 }
 SEGMENTS {
     ZEROPAGE: load = ZP,   type = zp;
+    STARTUP:  load = MAIN, type = ro,  optional = yes;
     LOWCODE:  load = MAIN, type = ro,  optional = yes;
     ONCE:     load = MAIN, type = ro,  optional = yes;
     CODE:     load = MAIN, type = rw;
--- a/cfg/none.cfg
+++ b/cfg/none.cfg
@@ -8,7 +8,7 @@ SYMBOLS {
 }
 MEMORY {
     ZP:   file = "", define = yes, start = __ZPSTART__, size = $001F;
-    MAIN: file = %O, define = yes, start = %S,          size = __STACKSTART__ - __STACKSIZE__;
+    MAIN: file = %O,               start = %S,          size = __STACKSTART__ - __STACKSIZE__ - %S;
 }
 SEGMENTS {
     ZEROPAGE: load = ZP,   type = zp;
--- a/cfg/none.cfg
+++ b/cfg/none.cfg
@@ -1,9 +1,14 @@
+FEATURES {
+    STARTADDRESS: default = $1000;
+}
 SYMBOLS {
-    __STACKSIZE__: type = weak, value = $0800; # 2k stack
+    __STACKSIZE__:  type = weak, value = $0800; # 2k stack
+    __STACKSTART__: type = weak, value = $8000;
+    __ZPSTART__:    type = weak, value = $0080;
 }
 MEMORY {
-    ZP:   file = "", define = yes, start = $0000, size = $0001F;
-    MAIN: file = %O, define = yes, start = %S,    size = $10000 - __STACKSIZE__;
+    ZP:   file = "", define = yes, start = __ZPSTART__, size = $001F;
+    MAIN: file = %O, define = yes, start = %S,          size = __STACKSTART__ - __STACKSIZE__;
 }
 SEGMENTS {
     ZEROPAGE: load = ZP,   type = zp;
--- a/cfg/none.cfg
+++ b/cfg/none.cfg
@@ -3,7 +3,7 @@ SYMBOLS {
 }
 MEMORY {
     ZP:   file = "", define = yes, start = $0000, size = $0001F;
-    MAIN: file = %O,               start = %S,    size = $10000 - __STACKSIZE__;
+    MAIN: file = %O,               start = %S,    size = $10000 - __STACKSIZE__, define = yes;
 }
 SEGMENTS {
     ZEROPAGE: load = ZP,   type = zp;
--- a/cfg/none.cfg
+++ b/cfg/none.cfg
@@ -3,7 +3,7 @@ SYMBOLS {
 }
 MEMORY {
     ZP:   file = "", define = yes, start = $0000, size = $0001F;
-    MAIN: file = %O,               start = %S,    size = $10000 - __STACKSIZE__, define = yes;
+    MAIN: file = %O, define = yes, start = %S,    size = $10000 - __STACKSIZE__;
 }
 SEGMENTS {
     ZEROPAGE: load = ZP,   type = zp;
--- a/libsrc/Makefile
+++ b/libsrc/Makefile
@@ -27,13 +27,13 @@ TARGETS = apple2       \
           gamate       \
           lynx         \
           nes          \
+          none         \
           osic1p       \
           pce          \
           sim6502      \
           sim65c02     \
           supervision  \
-          telestrat    \
-          none
+          telestrat
 
 DRVTYPES = emd \
            joy \
--- a/libsrc/Makefile
+++ b/libsrc/Makefile
@@ -33,7 +33,7 @@ TARGETS = apple2       \
           sim65c02     \
           supervision  \
           telestrat    \
-          none
+          no-platform
 
 DRVTYPES = emd \
            joy \
--- a/libsrc/Makefile
+++ b/libsrc/Makefile
@@ -33,7 +33,7 @@ TARGETS = apple2       \
           sim65c02     \
           supervision  \
           telestrat    \
-          no-platform
+          none
 
 DRVTYPES = emd \
            joy \
--- a/libsrc/Makefile
+++ b/libsrc/Makefile
@@ -32,7 +32,8 @@ TARGETS = apple2       \
           sim6502      \
           sim65c02     \
           supervision  \
-          telestrat
+          telestrat    \
+          none
 
 DRVTYPES = emd \
            joy \
--- a/libsrc/none/crt0.s
+++ b/libsrc/none/crt0.s
@@ -1,0 +1,25 @@
+       .export         _exit
+       .export         __STARTUP__ : absolute = 1      ; Mark as startup
+       .import         zerobss, _main
+       .import         initlib, donelib
+       .import         __MAIN_START__, __MAIN_SIZE__   ; Linker generated
+       .import         __STACKSIZE__                   ; Linker generated
+
+       .include "zeropage.inc"
+
+       .segment "STARTUP"
+
+       cld
+       ldx #$FF
+       txs
+       lda #<(__MAIN_START__ + __MAIN_SIZE__ + __STACKSIZE__)
+       ldx #<(__MAIN_START__ + __MAIN_SIZE__ + __STACKSIZE__)
+       sta sp
+       stx sp+1
+       jsr zerobss
+       jsr initlib
+       jsr _main
+_exit: pha
+       jsr donelib
+       pla
+       brk
--- a/libsrc/none/crt0.s
+++ b/libsrc/none/crt0.s
@@ -2,7 +2,7 @@
        .export         __STARTUP__ : absolute = 1      ; Mark as startup
        .import         zerobss, _main
        .import         initlib, donelib
-       .import         __STACKSTART__, __STACKSIZE__   ; Linker generated
+       .import         __STACKSTART__                  ; Linker generated
 
        .include "zeropage.inc"
 
--- a/libsrc/none/crt0.s
+++ b/libsrc/none/crt0.s
@@ -9,9 +9,6 @@
 
        .segment "STARTUP"
 
-       cld
-       ldx #$FF
-       txs
        lda #<(__MAIN_START__ + __MAIN_SIZE__ + __STACKSIZE__)
        ldx #<(__MAIN_START__ + __MAIN_SIZE__ + __STACKSIZE__)
        sta sp
@@ -22,4 +19,4 @@
 _exit: pha
        jsr donelib
        pla
-       brk
+       rts
--- a/libsrc/none/crt0.s
+++ b/libsrc/none/crt0.s
@@ -2,15 +2,14 @@
        .export         __STARTUP__ : absolute = 1      ; Mark as startup
        .import         zerobss, _main
        .import         initlib, donelib
-       .import         __MAIN_START__, __MAIN_SIZE__   ; Linker generated
-       .import         __STACKSIZE__                   ; Linker generated
+       .import         __STACKSTART__, __STACKSIZE__   ; Linker generated
 
        .include "zeropage.inc"
 
        .segment "STARTUP"
 
-       lda #<(__MAIN_START__ + __MAIN_SIZE__ + __STACKSIZE__)
-       ldx #<(__MAIN_START__ + __MAIN_SIZE__ + __STACKSIZE__)
+       lda #<__STACKSTART__
+       lda #>__STACKSTART__
        sta sp
        stx sp+1
        jsr zerobss
--- a/doc/customizing.sgml
+++ b/doc/customizing.sgml
@@ -81,6 +81,7 @@ SEGMENTS {
     ZEROPAGE: load = ZP,  type = zp,  define   = yes;
     DATA:     load = ROM, type = rw,  define   = yes, run = RAM;
     BSS:      load = RAM, type = bss, define   = yes;
+    HEAP:     load = RAM, type = bss, optional = yes;
     STARTUP:  load = ROM, type = ro;
     ONCE:     load = ROM, type = ro,  optional = yes;
     CODE:     load = ROM, type = ro;
@@ -94,6 +95,7 @@ The meaning of each of these segments is as follows.
 <p><tt>       ZEROPAGE:  </tt>Data in page 0, defined by ZP as starting at $0 with length $100
 <p><tt>       DATA:      </tt>Initialized data that can be modified by the program, stored in RAM
 <p><tt>       BSS:       </tt>Uninitialized data stored in RAM (used for variable storage)
+<p><tt>       HEAP:      </tt>Uninitialized C-level heap storage in RAM, optional
 <p><tt>       STARTUP:   </tt>The program initialization code, stored in ROM
 <p><tt>       ONCE:      </tt>The code run once to initialize the system, stored in ROM
 <p><tt>       CODE:      </tt>The program code, stored in ROM
@@ -300,20 +302,23 @@ also forcing a BRK instruction into the CPU.
 <sect>Custom Run-Time Library Creation<p>
 
 The next step in customizing the cc65 toolset is creating a run-time
-library for the targeted hardware.  The recommended way to do this is to
-modify the platform-independent standard library of the cc65 distribution.
-It is named &quot;none.lib&quot; in the lib directory of the distribution.
+library for the targeted hardware.  The easiest way to do this is to
+modify a standard library from the cc65 distribution.  In this example,
+there is no console I/O, mouse, joystick, etc. in the system, so it is
+most appropriate to use the simplest library as the base, which is for
+the Watara Supervision and is named &quot;supervision.lib&quot; in the
+lib directory of the distribution.
 
-When using &quot;none.lib&quot; we need to supply our own <tt>crt0</tt>
-module with custom startup code.  This is simply done by first copying the
-the library and giving it a new name, compiling the startup code with ca65,
-and finally using the ar65 archiver to add the module to the new library.
-The steps are shown below:
+The only modification required is to replace the <tt>crt0</tt> module in
+the supervision.lib library with custom startup code.  This is simply
+done by first copying the library and giving it a new name, compiling
+the startup code with ca65, and finally using the ar65 archiver to
+replace the module in the new library.  The steps are shown below:
 
 <tscreen><verb>
-        cp /usr/local/share/cc65/lib/none.lib sbc.lib
-        ca65 crt0.s
-        ar65 a sbc.lib crt0.o
+$ copy &quot;C:\Program Files\cc65\lib\supervision.lib&quot; sbc.lib
+$ ca65 crt0.s
+$ ar65 a sbc.lib crt0.o
 </verb></tscreen>
 
 <sect>Interrupt Service Routine Definition<p>
@@ -701,14 +706,14 @@ that can be used as the initialization data for the Xilinx Block RAM
 used for code storage:
 
 <tscreen><verb>
-        cc65 -t none -O --cpu 65sc02 main.c
-        ca65 --cpu 65sc02 main.s
-        ca65 --cpu 65sc02 rs232_tx.s
-        ca65 --cpu 65sc02 interrupt.s
-        ca65 --cpu 65sc02 vectors.s
-        ca65 --cpu 65sc02 wait.s
-        ld65 -C sbc.cfg -m main.map interrupt.o vectors.o wait.o
-                  rs232_tx.o main.o sbc.lib
+$ cc65 -t none -O --cpu 65sc02 main.c
+$ ca65 --cpu 65sc02 main.s
+$ ca65 --cpu 65sc02 rs232_tx.s
+$ ca65 --cpu 65sc02 interrupt.s
+$ ca65 --cpu 65sc02 vectors.s
+$ ca65 --cpu 65sc02 wait.s
+$ ld65 -C sbc.cfg -m main.map interrupt.o vectors.o wait.o rs232_tx.o
+          main.o sbc.lib
 </verb></tscreen>
 
 During the C-level code compilation phase (<tt>cc65</tt>), assumptions
--- a/doc/customizing.sgml
+++ b/doc/customizing.sgml
@@ -302,17 +302,16 @@ also forcing a BRK instruction into the CPU.
 The next step in customizing the cc65 toolset is creating a run-time
 library for the targeted hardware.  The recommended way to do this is to
 modify the platform-independent standard library of the cc65 distribution.
-It is named &quot;no-platform.lib&quot; in the lib directory of the
-cc65 distribution.
+It is named &quot;none.lib&quot; in the lib directory of the distribution.
 
-When using &quot;no-platform.lib&quot; we need to supply our own <tt>crt0</tt>
+When using &quot;none.lib&quot; we need to supply our own <tt>crt0</tt>
 module with custom startup code.  This is simply done by first copying the
 the library and giving it a new name, compiling the startup code with ca65,
 and finally using the ar65 archiver to add the module to the new library.
 The steps are shown below:
 
 <tscreen><verb>
-        cp /usr/local/share/cc65/lib/no-platform.lib sbc.lib
+        cp /usr/local/share/cc65/lib/none.lib sbc.lib
         ca65 crt0.s
         ar65 a sbc.lib crt0.o
 </verb></tscreen>
--- a/doc/customizing.sgml
+++ b/doc/customizing.sgml
@@ -302,23 +302,20 @@ also forcing a BRK instruction into the CPU.
 <sect>Custom Run-Time Library Creation<p>
 
 The next step in customizing the cc65 toolset is creating a run-time
-library for the targeted hardware.  The easiest way to do this is to
-modify a standard library from the cc65 distribution.  In this example,
-there is no console I/O, mouse, joystick, etc. in the system, so it is
-most appropriate to use the simplest library as the base, which is for
-the Watara Supervision and is named &quot;supervision.lib&quot; in the
-lib directory of the distribution.
+library for the targeted hardware.  The recommended way to do this is to
+modify the platform-independent standard library of the cc65 distribution.
+It is named &quot;none.lib&quot; in the lib directory of the distribution.
 
-The only modification required is to replace the <tt>crt0</tt> module in
-the supervision.lib library with custom startup code.  This is simply
-done by first copying the library and giving it a new name, compiling
-the startup code with ca65, and finally using the ar65 archiver to
-replace the module in the new library.  The steps are shown below:
+When using &quot;none.lib&quot; we need to supply our own <tt>crt0<tt>
+module with custom startup code.  This is simply done by first copying the
+the library and giving it a new name, compiling the startup code with ca65,
+and finally using the ar65 archiver to add the module to the new library.
+The steps are shown below:
 
 <tscreen><verb>
-$ copy &quot;C:\Program Files\cc65\lib\supervision.lib&quot; sbc.lib
-$ ca65 crt0.s
-$ ar65 a sbc.lib crt0.o
+        cp /usr/local/share/cc65/lib/none.lib sbc.lib
+        ca65 crt0.s
+        ar65 a sbc.lib crt0.o
 </verb></tscreen>
 
 <sect>Interrupt Service Routine Definition<p>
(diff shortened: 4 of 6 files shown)


oliverschmidt:
I must admit that I don't see any point in the none.lib:
* It's as-is non-functional but unexperienced cc65 users won't understand that cause additional issues.
* Without very precise documentation what i.e. a crt0.s is supposed to contain in which order etc. the only way for someone to create a new cc65 target is to look (very closely) at existing targets anyway so it seems pretty natural to me to start from an existing target in the first place.
* I don't see anybody creating - and maintaining (!) - that precise documentation.

bauen1:
* I am looking for a better name (eg. `no-platform.lib`)
 * Porting by using an already existing target's library brings the danger of forgetting to remove  platform-specific modules (eg. forgetting about the joystick for the supervision target), `none.lib`  only contains platform-independent modules
* The documentation already exists, but its outdated and unmaintained: https://github.com/cc65/cc65/blob/master/doc/customizing.sgml

`none.lib` was created to reduce that chance of errors happening when porting the cc65 library out-of-tree.

oliverschmidt:
> Porting by using an already existing target's library brings the danger of forgetting to remove platform-specific modules (eg. forgetting about the joystick for the supervision target), none.lib only contains platform-independent modules

I clearly see your point. However, it doesn't invalidate what I wrote.

> The documentation already exists, but its outdated and unmaintained: https://github.com/cc65/cc65/blob/master/doc/customizing.sgml

That's what I'm refering to. That documentation was a "one-time-contribution". To me it was already at the time it was created clear that it would become outdated.

mrdudz:
i would appreciate none.lib - in fact i have had this in my local copy since a decade or so :)

oliverschmidt:
Sorry but I don't follow you both here. Maybe it's because I think that http://cc65.github.io/doc/customizing.html#s4 is in general worthless. If I want to add a target to cc65 I for sure change the relevant Makefile to have it create the new library from the sources. I certainly won't start with "patching" some binary library. Instead of further pushing that obscure library patching approach by supplying a none.lib rather the documentation should be changed to advocate a meaningful approach...

Marrin:
I don't think this is what this pull request is about. It is about your approach, but instead of starting with a target specific library source, there will be a "clean" library with no target specific bits. So you don't end up with joystick drivers for a different platform - or worse.

oliverschmidt:
We go round in circles here. I don't see any point in shipping a non-functional library. If someone want to create a new library the right way is to create it from scratch. I haven't seen any argument in this discussion so far invalidating my perspective.

greg-king5:
I agree with Oliver.  I don't see any advantage in creating a new platform/target library out-of-tree.  It makes more sense to create a directory "next door" to existing code, use that code as examples/templates for new code, and use `libsrc/Makefile` to build a fresh (and clean) library.  A developer doesn't need to wrestle with compiler, assembler, and librarian commands; that's the job of `make`.

If someone wants to put the new code into a separate project, then that one can move/copy the directory to anywhere one wants, later after it has been debugged.

mrdudz:
i don't agree that such a library would be "non functional". it could just contain the bare compiler runtime functions and still be useful. for example in my chameleon menu code i do exactly that - the bare minimum compiler runtime comes from the cc65 libs, and pretty much everything else is custom. and it wouldnt be useful for anyone to put this code into the cc65 tree either, since noone will be writing menu code for it anytime soon :) a none.lib in the tree however would save me from manually tracking compiler updates that need a matching runtime. this is also pretty much what most other compilers i have seen do (i dont like the everything-is-in-one-lib approach either, for exactly this reason)

greg-king5:
But, a `none.lib` isn't needed for that case.  _Any_ library will work there.  The platform-specific stuff simply will be ignored.

mrdudz:
yes and no. i also use custom versions of stdlib functions, for example (eg string functions that only handle 255 chars at max). and when eg conio things come into play things will get really messy if you arent super careful. (another problem of the all-in-one approach).

bauen1:
This only goes well until you accidentally link against platform-specific parts (which you can't remove since you don't know the name the platform-specific modules use)

mrdudz:
yep, which is exactly why i use this none.lib that only has the runtime in it

greg-king5:
Ah, OK.  "Debugging assistance" is a legitimate justification for `none.lib`.  Specific diagnostics are better than mysteriously misbehaving programs. ;-)

Putting "none" in the Makefile makes sense; but, the descriptions in "customizing.sgml" still are unreasonable.  I think that you should remove the SGML patch from this PR.

(That document should get its own PR with more practical instructions.)

oliverschmidt:
By now we're messing up several quite different scenarios.

We started out with the idea of a non-functional library purely intended as building block for a new target library. I think I made clear that I don't see any value in this and that I'm not interested in this.

However, now we're discussing a target-neutral library that can be used in environments without explicit cc65 target support. Such a library wouldn't include any I/O whatsoever. I/O capabilities would have to be supplied by the user using custom asm modules, inline asm, _sys() or function pointer constants.

I _can_ see the point of such a target-neutral library and am in general interested in it. But I'm afraid we have (very ?) different opinions on the details.

What I have in mind  is a `-t none` option for cc65 and cl65. The corresponding none.lib includes a crt0.s. That crt0.s does the usual things like clearing BSS, calling constructors, main() and destructors. It ends with an rts. The binary created with none.cfg doesn't contain any header or alike. The start address defaults to let's say $1000 and is supposed to be set via `--start-addr 0x????`. The start of the software stack defaults to let's say $8000 and is supposed to be set via `-D __STACKSTART__=0x????`. The ZP memory area starts by default at let's say $80 and can be moved via some -D too.

libsrc/Makefile
```diff
@@ -32,7 +32,8 @@ TARGETS = apple2       \
           sim6502      \
           sim65c02     \
           supervision  \
-          telestrat
+          telestrat    \
+          none
```
greg-king5: "`none`" should be moved to its alphabetically sorted position in the list.

mrdudz:
putting a crt0 into this lib isnt useful IMHO. most certainly you will want a custom one anyway.

bauen1:
While a crt0 doesn't exactly seem usefull for me, the following should do (with a small addition to none.cfg)
```
       .export         _exit
       .export         __STARTUP__ : absolute = 1      ; Mark as startup
       .import         zerobss, _main
       .import         initlib, donelib
       .import         __MAIN_START__, __MAIN_SIZE__   ; Linker generated
       .import         __STACKSIZE__                   ; Linker generated

       .include "zeropage.inc"

       .segment "STARTUP"

       cld
       ldx #$FF
       txs
       lda #<(__MAIN_START__ + __MAIN_SIZE__ + __STACKSIZE__)
       ldx #<(__MAIN_START__ + __MAIN_SIZE__ + __STACKSIZE__)
       sta sp
       stx sp+1
       jsr zerobss
       jsr initlib
       jsr _main
_exit: pha
       jsr donelib
       pla
       brk
```
Would that be ok with everyone ?

oliverschmidt:
> putting a crt0 into this lib isnt useful IMHO. most certainly you will want a custom one anyway.

But without a crt0.s the library isn't functional - and I won't add a non-functional library ;-)

oliverschmidt:
My question regarding different opinions/scenarios/environments was more in the direction if we agree that there's a main() being the (single) entry point into the C code.

oliverschmidt:
> While a crt0 doesn't exactly seem usefull for me, the following should do [...]

What is the conceptual difference between your proposal and a crt0.s?

bauen1:
I just wanted to see if said implementation would be ok for you, It might have sounded a little bit different that what I meant to say.

groessler:
Wouldn't it make more sense instead of creating a (non-functional) none.lib, to split the cc65 libraries into two? One machine independent runtime.lib and a machine dependend <target>.lib. I think it would fix the OP's problem.

oliverschmidt:
> I just wanted to see if said implementation would be ok for you,

In general it's exactly what I have in mind. However, I'd suggest a few changes:
* Instead of initializing the hardware stack and ending with a brk I'd leave the hardware stack as-is and end with an rts.
* I'd add the __exit symbool.

mrdudz:
> But without a crt0.s the library isn't functional

again, yes and no. the problem is - again - the concept of putting everything into a single library. if you look at eg gcc that isnt the case, and its totally normal for embedded stuff to provide its own crt0 and then link against the runtime.

> Wouldn't it make more sense instead of creating a (non-functional) none.lib, to split the cc65 libraries into two?

to me, it'd make sense to have a seperate crt0, runtime, libc, conio and other stuff all in their own libraries :)

oliverschmidt:
> Wouldn't it make more sense instead of creating a (non-functional) none.lib, to split the cc65 libraries into two? One machine independent runtime.lib and a machine dependend .lib. I think it would fix the OP's problem.

I'm in general not at all against modularizing the libraries. But I guess we agree that we'd need to keep the user experience of just providing `-t <target>` and having every thing work automagically, don't we?

oliverschmidt:
> [...] totally normal for embedded stuff to provide its own crt0 and then link against the runtime.

Just to be sure I'm not missing the point: Doing that is still possible if the library already contains a crt0.s, right?

bauen1:
> Instead of initializing the hardware stack and ending with a brk I'd leave the hardware stack as-is and end with an rts.

I see your reason for this and will implement it shortly
> I'd add the __exit symbool.

I'm not sure, you probably mean `void _exit()`, because the code already definse `_exit` (`exit` for c right ?)
> Just to be sure I'm not missing the point: Doing that is still possible if the library already contains a crt0.s, right?

It is possible, but `none.lib` should **ONLY** contain `crt.o` otherwise we will be turning in a circle again.
> I'm in general not at all against modularizing the libraries. But I guess we agree that we'd need to keep the user experience of just providing -t <target> and having every thing work automagically, don't we?

I think the user already knows what hes doing when using `-t none`, but again, you have a pretty valid point here.

mrdudz:
> Just to be sure I'm not missing the point: Doing that is still possible if the library already contains a crt0.s, right?

sure. but the result depends on linking order so you have to be super careful. i prefer a simple error instead (eg when i didnt link the crt0).

groessler:
Sure, the user experience shouldn't suffer. I guess it just would need some small changes in cl65.

oliverschmidt:
Okay, trying to sort things...

One idea is to (finally) modularize the cc65 libraries. As far as I know one of the reasons not doing that so far was that ld65 requires `--start-group` / `--end-group` to behave the way it behaves with a monolithic library. However, when using `ld65 -t <target>` or `cl65 -t <target>` we want to hide the complexity of two libraries from the user anyway so we can as well hide the group thingy in just the same way.

Currently the common part of the monolithic library differs between 6502 and 65C02 targets. We want to keep that optimization so intuitive names for the two libraries (containing both the Runtime Library and the common parts of the C Library) would be 6502.lib and 65c02.lib.

ld65 and(?) cl65 would need to learn which target needs to link 6502.lib and which target needs 65c02.lib to behave as before when using `-t`. If we go that route I don't see any `-t none` as the guys going for a custom target/environment/scenario surely don't use `-t` anyway but directly name 6502.lib or 65c02.lib.

Is anybody willing/able/interested to contribute the necessary changes for that approach? Do we agree that this approach makes the pull request at hand obsolete?

oliverschmidt:
> It is possible, but none.lib should ONLY contain crt.o otherwise we will be turning in a circle again.

Sorry, but now you completely lost me :-( I _DO_ understand that both @mrdudz and you don't want to have a crt0.s and the "desired" parts of "the" library in the same file in order to avoid to unintentionally use the builtin crt0.s. However, if none.lib only contains crt0.s then how does one access the "desired" parts of "the" library?

Apart from that lack of understanding on my side I'd like to mention that a none.lib known to contain exactly one well known module (crt0.o) that isn't desirable in a custom scenario would still allow to do a single `ar65 d none.lib crt0.o` to get a "safe" library. That could still be a valid compromise, couldn't it?

groessler:
On 01/04/18 17:40, Oliver Schmidt wrote:

> However, when using |ld65 -t <target>| or |cl65 -t <target>| we want 
> to hide the complexity of two libraries from the user anyway so we can 
> as well hide the group thingy in just the same way. 


I don't know if we want to do that. If someone uses ld65 directly and 
not cl65, he's anyway (supposed to be) knowing what he does, and can be 
bothered to add the new library and start/endgroup commands to the 
linker. If the linker would do something automatically one doesn't 
exaclty know what it is doing and it would be annoying if for some 
reason it should _not_ do this.

We also don't keep backwards compatibility in other areas, so there is 
precedence for such a change.


> ld65 and(?) cl65 would need to learn which target needs to link 
> 6502.lib and which target needs 65c02.lib to behave as before when 
> using |-t|.

In my view we shouldn't change ld65, just cl65 would need to know that.


> If we go that route I don't see any |-t none| as the guys going for a 
> custom target/environment/scenario surely don't use |-t| anyway but 
> directly name 6502.lib or 65c02.lib. 


Doesn't -t also change character substition in the compiler and assembler?


> Is anybody willing/able/interested to contribute the necessary changes 
> for that approach? Do we agree that this approach makes the pull 
> request at hand obsolete?
>

I don't answer the first part (yet). But I agree on the second part.

oliverschmidt:
> If someone uses ld65 directly and
not cl65, he's anyway (supposed to be) knowing what he does, and can be
bothered to add the new library and start/endgroup commands to the
linker. If the linker would do something automatically one doesn't
exaclty know what it is doing and it would be annoying if for some
reason it should _not_ do this.

To me the line between "do things automagically" and "do what I want without surprising me" isn't between using `cl65` vs. using `ld65` but rather between using `ld65 -t <target>` vs. using `ld65 -C <some_config> --lib <some_library>`.

silverdr:
> On 2018-01-04, at 17:56, Christian Groessler <notifications@github.com> wrote:
> 
> On 01/04/18 17:40, Oliver Schmidt wrote:
> 
> > However, when using |ld65 -t <target>| or |cl65 -t <target>| we want 
> > to hide the complexity of two libraries from the user anyway so we can 
> > as well hide the group thingy in just the same way. 
> 
> I don't know if we want to do that.

I believe we do. We shouldn't break the long established behaviour unless really necessary.

> If someone uses ld65 directly and 
> not cl65, he's anyway (supposed to be) knowing what he does, and can be 
> bothered to add the new library and start/endgroup commands to the 
> linker. If the linker would do something automatically one doesn't 
> exaclty know what it is doing and it would be annoying if for some 
> reason it should _not_ do this.

If I use my custom config then I want every tool to behave predictably and avoid any behind-the-scenes "optimisations". When I use -t <target> without custom configs then I consciously or subconsciously agree to use the (hopefully) reasonable defaults that come with it. Breaking that would be perceived negatively IMHO.

> We also don't keep backwards compatibility in other areas, so there is 
> precedence for such a change.

Again - we shouldn't break the long established behaviour unless really necessary.

> Doesn't -t also change character substition in the compiler and assembler?

I would even say "unfortunately". IMHO this should imply using a different default config, where this behaviour would be defined. I remember I got bitten by this some time ago.

-- 

silverdr:
> On 2018-01-04, at 18:03, Oliver Schmidt <notifications@github.com> wrote:
> 
> To me the line between "do things automagically" and "do what I want without surprising me" isn't between using cl65 vs. using ld65 but rather between using ld65 -t <target> vs. using ld65 -C <some_config> --lib <some_library>.

+1
-- 

bauen1:
> Sorry, but now you completely lost me :-( I DO understand that both @mrdudz and you don't want to have a crt0.s and the "desired" parts of "the" library in the same file in order to avoid to unintentionally use the builtin crt0.s. However, if none.lib only contains crt0.s then how does one access the "desired" parts of "the" library?

By building your own platform-specific parts just like any other target in `libsrc`

> Apart from that lack of understanding on my side I'd like to mention that a none.lib known to contain exactly one well known module (crt0.o) that isn't desirable in a custom scenario would still allow to do a single ar65 d none.lib crt0.o to get a "safe" library. That could still be a valid compromise, couldn't it?

Because it only contains `crt0.o` one can assume that after `crt0.o` is replaced the only thing left is platform-independent code, so yes it is a valid compromise.
I didn't want to add it because it opens the opertunity to add more code specific to `none.lib`, defeating its original purpose.

groessler:
On 01/04/18 18:03, Oliver Schmidt wrote:
>
> To me the line between "do things automagically" and "do what I want 
> without surprising me" isn't between using |cl65| vs. using |ld65| but 
> rather between using |ld65 -t <target>| vs. using |ld65 -C 
> <some_config> --lib <some_library>|.
>

I'm not sure I understand what you are trying to say.

Let's put it in other words:

- cl65 -t handles the change to two libraries "under the hood", no user 
visible change
- ld65 -C needs to have the two libraries explicitly specified, within 
--start-group/--endgroup
- ld65 -t knows about the two libraries and adds them within 
--start-group/--endgroup "under the hood"


This would be my proposal.

oliverschmidt:
```
- cl65 -t handles the change to two libraries "under the hood", no user
visible change
- ld65 -C needs to have the two libraries explicitly specified, within
--start-group/--endgroup
- ld65 -t knows about the two libraries and adds them within
--start-group/--endgroup "under the hood"
```
That's exactly my perspective :-) But it requires changes to ld65 (beside changes to cl65), doesn't it? You wrote "In my view we shouldn't change ld65, just cl65 would need to know that." which made me think you have a different opinion...

oliverschmidt:
> By building your own platform-specific parts just like any other target in libsrc

Please elaborate. I have no clue what you want to express :-(

oliverschmidt:
> I didn't want to add it because it opens the opertunity to add more code specific to none.lib, defeating its original purpose.

I see. We'd surely stay clear from that.

bauen1:
>>  By building your own platform-specific parts just like any other target in libsrc
>
> Please elaborate. I have no clue what you want to express :-(

`none.lib` contains the runtime (the desired platform-independent parts) and `crt0.o` so it can be used like any other library.
When porting `none.lib` one would normally replace `crt0.o` with a custom implementation and add more platform-specific modules.

I am still talking about this pull-request and not your future plans for modularizing all libraries, which would make `none.lib` and this pull-request obsolete.

silverdr:
> On 2018-01-04, at 19:26, Oliver Schmidt <notifications@github.com> wrote:
> 
> - cl65 -t handles the change to two libraries "under the hood", no user
> visible change
> - ld65 -C needs to have the two libraries explicitly specified, within
> --start-group/--endgroup
> - ld65 -t knows about the two libraries and adds them within
> --start-group/--endgroup "under the hood"
> 
> That's exactly my perspective :-) But it requires changes to ld65 (beside changes to cl65), doesn't it? You wrote "In my view we shouldn't change ld65, just cl65 would need to know that." which made me think you have a different opinion...

I had the same thought, which even made me join the discussion and oppose it ;-) There must have been a strange misunderstanding then.

-- 

oliverschmidt:
Further above you wrote:

> [...] none.lib should ONLY contain crt.o [...]

Now you wrote:

> none.lib contains the runtime (the desired platform-independent parts) and crt0.o [...]

You see where the misunderstanding came from? ;-))

Anyhow, I now think we're on the same page :-)

oliverschmidt:
Okay, I think I have a nice solution for the crt0.o "issue":

A usual crt0.s contains
```
.export         __STARTUP__ : absolute = 1
```
If we omit this from the none.lib crt0.s then the crt0.s becomes passive and won't be found/used unintentionally. Instead we add
```
__STARTUP__ : type = weak, value = 1;
```
to the none.cfg. This keeps the linker happy. If one doesn't supply a custom crt0.s there will be linker warnings - as desired.

The trick is now to add another custom export to the none.lib crt0.s - like e.g.
```
.export crt0 : absolute = 1
```
Then one use cl65 or ld65 with `-u crt0` to bring in the none.lib crt0.s. If we document that as part of the none target documentation I think everybody is pretty happy.

bauen1:
What exactly is the point of `crt0.o` then ?

oliverschmidt:
>>I'd add the __exit symbool.

>I'm not sure, you probably mean void _exit(), because the code already defines _exit (exit for c right ?)

That's what I meant. However, I mixed something up between cc65 and some other project ;-) Sorry for the noise. There's nothing to do for you.

oliverschmidt:
> What exactly is the point of crt0.o then ?

Yet again: Without it we deliver a library that nobody can use without adding his own crt0.s Doing so requires intimate knowledge of the cc65 internals. Apart from that the interface between crt0.s and the rest of the cc65 toolchain changed quite some times in the past and will change again in the future. Any user supplied crt0.s needs adjustments sooner or later. Therefore I want to have a crt0.s that is delivered with the library and maintained by the cc65 team and known to work.

Building a working binary with none.lib (using its builtin crt0.s) is by no means a hypothetical thing. E.g. the Apple II allows to load and run such a none-binary out-of-the-box. With a crt0.s inside none.lib the none target is of use out-of-the-box for users. Using it as building block to create something different is a mere side effect then.

Apart from that the crt0.s inside non.lib is an "active documentation" what a correct crt0.s needs to contain. Having it there makes it likely to be maintained. Having something like it in the docs effectively "makes sure" to have it not maintained - see the customizing docs.

groessler:
On 01/04/18 19:26, Oliver Schmidt wrote:
> But it requires changes to ld65 (beside changes to cl65), doesn't it? 
> You wrote "In my view we shouldn't change ld65, just cl65 would need 
> to know that." which made me think you have a different opinion...


Yes, sorry. I wasn't aware that "-t" of ld65 automatically adds the 
target library. I only ever linked asm programs with ld65.
So, yes, ld65 changes are required.

bauen1:
> Okay, I think I have a nice solution for the crt0.o "issue":
>
> A usual crt0.s contains
> ```
> .export         __STARTUP__ : absolute = 1
> ```
> If we omit this from the none.lib crt0.s then the crt0.s becomes passive and won't be found/used unintentionally. Instead we add
> ```
> __STARTUP__ : type = weak, value = 1;
> ```
> to the none.cfg. This keeps the linker happy. If one doesn't supply a custom crt0.s there will be linker warnings - as desired.
> 
> The trick is now to add another custom export to the none.lib crt0.s - like e.g.
> ```
> .export crt0 : absolute = 1
> ```
> Then one use cl65 or ld65 with -u crt0 to bring in the none.lib crt0.s. If we document that as part of the none target documentation I think everybody is pretty happy.

I'm against implementing these changes:
 * requires more documentation
 * `none.lib` becomes non-functional by default
 * `cc65` assumes that importing `__STARTUP__` includes apropiate startup code, which may breaks C programs unless you actually use the build-in, hidden, `crt0.o`.
 
> Apart from that the crt0.s inside non.lib is an "active documentation" what a correct crt0.s needs to contain. Having it there makes it likely to be maintained. Having something like it in the docs effectively "makes sure" to have it not maintained - see the customizing docs.

You have a very valid point.

oliverschmidt:
> I'm against implementing these changes [...]

Those changes were meant as a concession for you and @mrdudz - I personally would like much more to have none.lib to default to being usable like any other library and requiring either careful linking or manual removal of crt0.o to be used as builing block with own crt0 !

bauen1:
>     I'm against implementing these changes [...]
> Those changes were meant as a concession for you and @mrdudz - I personally would like much more to have none.lib to default to being usable like any other library and requiring either careful linking or manual removal of crt0.o to be used as builing block with own crt0 !

I'm sorry for the trouble.

So, is everyone OK with the current state of `none.lib` ?

mrdudz:
Am Donnerstag, 4. Januar 2018, 17:56:39 CET schrieb Christian Groessler:
> On 01/04/18 17:40, Oliver Schmidt wrote:
> > However, when using |ld65 -t <target>| or |cl65 -t <target>| we want
> > to hide the complexity of two libraries from the user anyway so we can
> > as well hide the group thingy in just the same way.
> 
> I don't know if we want to do that. If someone uses ld65 directly and
> not cl65, he's anyway (supposed to be) knowing what he does, and can be
> bothered to add the new library and start/endgroup commands to the
> linker. If the linker would do something automatically one doesn't
> exaclty know what it is doing and it would be annoying if for some
> reason it should _not_ do this.
> 
> We also don't keep backwards compatibility in other areas, so there is
> precedence for such a change.
> 
> > ld65 and(?) cl65 would need to learn which target needs to link
> > 6502.lib and which target needs 65c02.lib to behave as before when
> > using |-t|.
> 
> In my view we shouldn't change ld65, just cl65 would need to know that.
> 
> > If we go that route I don't see any |-t none| as the guys going for a
> > custom target/environment/scenario surely don't use |-t| anyway but
> > directly name 6502.lib or 65c02.lib.
> 
> Doesn't -t also change character substition in the compiler and assembler?
> 
> > Is anybody willing/able/interested to contribute the necessary changes
> > for that approach? Do we agree that this approach makes the pull
> > request at hand obsolete?
> 
> I don't answer the first part (yet). But I agree on the second part.

i agree to all of that :)

-- 

I should warn you that underneath these clothes I'm wearing boxer shorts and I 
know how to use them. 
<Robert Orben>

cfg/none.cfg
```diff
@@ -3,10 +3,11 @@ SYMBOLS {
 }
 MEMORY {
     ZP:   file = "", define = yes, start = $0000, size = $0001F;
-    MAIN: file = %O,               start = %S,    size = $10000 - __STACKSIZE__;
+    MAIN: file = %O,               start = %S,    size = $10000 - __STACKSIZE__, define = yes;
```
oliverschmidt: Please format the 'define = yes' consistently with the line above.

oliverschmidt:
> I'm sorry for the trouble. So, is everyone OK with the current state of none.lib ?

Okay, we seem to have a common understanding and agreement now :-)

I'd like to ask for some changes in none.cfg to realize what I wrote before:

> The start address defaults to let's say $1000 and is supposed to be set via `--start-addr 0x????`. The start of the software stack defaults to let's say $8000 and is supposed to be set via `-D __STACKSTART__=0x????`. The ZP memory area starts by default at let's say $80 and can be moved via some -D too.

The start addr default (of $1000) requires something like https://github.com/cc65/cc65/blob/master/cfg/apple2.cfg#L3-L5

As far as I see your software stack initialization isn't working well, you write in your crt0.s `__MAIN_START__ + __MAIN_SIZE__ + __STACKSIZE__`. MAIN is a memory area (not a segment). So `__MAIN_SIZE__ + __STACKSIZE__` is always 64kB. But 6502 systems usually have ROM at high addresses because of the 6502 vectors. Therefore I suggest to change the $10000 in the none.cfg to reference a new symbol named `__STACKSTART__` and have that symbol defined and set to a default (of $8000) just like `__STACKSIZE__`.

The first two memory locations serve a special purpose on the 6510. Therefore $0000 isn't a good general purpose location for the cc65 zeropage memory area. I suggest to replace the $0000 in the none.cfg with a reference to a new symbol named `__ZPSTART__` and have that symbol defined and set to a default (of $0080) just like `__STACKSIZE__`.

cfg/none.cfg
```diff
@@ -3,10 +3,11 @@ SYMBOLS {
 }
 MEMORY {
     ZP:   file = "", define = yes, start = $0000, size = $0001F;
-    MAIN: file = %O,               start = %S,    size = $10000 - __STACKSIZE__;
+    MAIN: file = %O, define = yes, start = %S,    size = $10000 - __STACKSIZE__;
```
oliverschmidt: Sorry, I didn't notice right away that the none.cfg had (another) bug before you started your pull request. Anyhow, you need to subtract the start address from the size to have and "end address". See https://github.com/cc65/cc65/blob/master/cfg/apple2.cfg#L16 as an example how this is meant to be done (incl. using a symbol).

bauen1:
Some code would be quite nice 🤣

oliverschmidt:
> Some code would be quite nice

https://github.com/cc65/cc65/blob/master/cfg/apple2.cfg contains every method / pattern I talked about.

I can surely do the changes myself ;-) I just thought you want/need to understand cc65 linker config files anyway to do with cc65 what you seem to want to do...

bauen1:
I do understand them but your explaination was a bit confusing.

cfg/none.cfg
```diff
@@ -1,12 +1,18 @@
...
-    __STACKSIZE__: type = weak, value = $0800; # 2k stack
+    __STACKSIZE__:  type = weak, value = $0800; # 2k stack
+    __STACKSTART__: type = weak, value = $8000;
+    __ZPSTART__:    type = weak, value = $0080;
 }
 MEMORY {
-    ZP:   file = "", define = yes, start = $0000, size = $0001F;
-    MAIN: file = %O,               start = %S,    size = $10000 - __STACKSIZE__;
+    ZP:   file = "", define = yes, start = __ZPSTART__, size = $001F;
+    MAIN: file = %O, define = yes, start = %S,          size = __STACKSTART__ - __STACKSIZE__;
```
oliverschmidt: - `define = yes` not necesarry anymore
- `__STACKSTART__ - __STACKSIZE__ - %S`

libsrc/none/crt0.s
```diff
@@ -0,0 +1,21 @@
+       .export         _exit
+       .export         __STARTUP__ : absolute = 1      ; Mark as startup
+       .import         zerobss, _main
+       .import         initlib, donelib
+       .import         __STACKSTART__, __STACKSIZE__   ; Linker generated
```
oliverschmidt: `__STACKSIZE__` not necesary anymore

oliverschmidt:
> I do understand them but your explaination was a bit confusing.

I see :-)

oliverschmidt:
Thanks for your extremely prompt reaction on my proposals :-) The change looks good to me.

I'd like to give the other community members a chance to comment before actual merging...

oliverschmidt:
Given that https://github.com/cc65/cc65/blob/master/src/common/target.c#L184 already defines that the 'none' target uses the ASCII char set it seems no additional restriction to add an ASCII char spec table to the 'none' target. This also enables sscanf(). E.g. https://github.com/cc65/cc65/blob/master/libsrc/sim6502/ctype.s can be copied.

greg-king5:
Side topic:
ld65 shouldn't be as smart as cl65; it should need to be told what libraries to link, by us or cl65.

I like Silverdr's proposal.  Currently, `-t` and `-C` differ about only the ".cfg" suffix in their operands.  His proposal says that `-t` should do both `-C` and `-(`.  That change would give us backwards compatibility in both directions (default-grouping for those who are used to monolithic libraries, and no-default-grouping for those who want total control over ld65).

silverdr:
@greg-king5 yes - thanks for noticing - I envision the `-t` as a shortcut for choosing predefined defaults for a given target that could otherwise be selected with several "lower level" options. Also the character substitution should be delegated to those predefined default configs and invoked via `-t` only indirectly. This way the compatibility with e. g. build scripts is retained, while additional flexibility is given to "power users".

oliverschmidt:
> Currently, -t and -C differ about only the ".cfg" suffix in their operands.

Thanks for pointing out! For some reason or another I was under the impression that -t would select both a linker config _and_ a library. Sort of interesting that I still know cc65 so little after all those years...

However, wouldn't it preferable if ld65 would actually behave as I expected it to do already?

oliverschmidt:
Time to merge...

groessler:
> All in all I'm wondering if it's really a good idea to try to move away from the monolithic libraries !

There are quite some strings/chars in the C files in libsrc/common. They could be moved to target libraries, but would then render the 6502.lib not to be self-contained. So I think it would be a pointless exercise to split the libraries. I'm withdrawing my proposal.

mrdudz:
Am Freitag, 5. Januar 2018, 11:24:28 CET schrieb greg-king5:
> Side topic:
> ld65 shouldn't be as smart as cl65; it should need to be told what libraries
> to link, by us or cl65.
> 
> I like Silverdr's proposal.  Currently, `-t` and `-C` differ about only the
> ".cfg" suffix in their operands.  His proposal says that `-t` should do
> both `-C` and `-(`.  That change would give us backwards compatibility in
> both directions (default-grouping for those who are used to monolithic
> libraries, and no-default-grouping for those who want total control over
> ld65).

+1 - it would also be in line with how other toolchains work

-- 

I see in C64.h a list of defines. 
<Wildstar>

oliverschmidt:
> I'm withdrawing my proposal.

I see - and agree.

So after having that discussed and merged this pull request the only question left is if it is desirable to change ld65 (in an incompatible way) to have -t not only fetch a config file but also a library.

oliverschmidt:
...or to have ld65 -t _not_ fetch a library but have it automatically group all libraries given.

groessler:
I would leave ld65 alone.

greg-king5:
Ever since built-in configurations were removed from ld65, I have wanted `-t` to be more than an echo
of `-C`.  I don't want it to go as far as cl65 goes; but, I do want it to have some kind of "added value".

oliverschmidt:
Full ACK.

PR #567, opened 2018-01-05
New OSI input routine based on disassembly of ROM code.

jefftranter:
New OSI input routine based on disassembly of ROM code. Equivalent to the ROM code but uses different storage locations to avoid corruption of memory used by cc65 programs. Typically the symptom of the corruption is not correctly returning to the OSI ROM monitor after returning from the main() function.

### DIFF
--- a/libsrc/osic1p/cgetc.s
+++ b/libsrc/osic1p/cgetc.s
@@ -4,17 +4,34 @@
 
         .constructor    initcgetc
         .export         _cgetc
+        .export         inputc
         .import         cursor
 
         .include        "osic1p.inc"
         .include        "extzp.inc"
         .include        "zeropage.inc"
 
+; Internal state that needs to be preserved across calls.
+; These variables are named after the original memory locations used by the
+; 65V PROM MONITOR, until the actual meaning is determined.
+.segment        "EXTZP" : zeropage
+
+KB0213:         .res    1
+KB0214:         .res    1
+KB0215:         .res    1
+KB0216:         .res    1
+
 ; Initialize one-character buffer that is filled by kbhit()
         .segment        "ONCE"
 initcgetc:
         lda     #$00
         sta     CHARBUF         ; No character in buffer initially
+
+        sta     KB0213          ; Initialize keyboard state
+        sta     KB0214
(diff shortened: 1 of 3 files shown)


oliverschmidt:
@greg-king5: Thanks for looking after this - unfortunately complex - issue :-)

polluks:
Pretty much overhead, maybe crt0 should preserve the locations and use the ROM call?

jefftranter:
Those locations would have to be saved before and restored after calling the ROM routine.
I think the hope is that eventually the code will be modified to address some issues like #511, so we want the routines to be part of the cc65 library.

polluks:
@jefftranter How about --start-addr 0x217 and give it a try?

greg-king5:
(Sorry, I completely forgot about it.)

The "new" `inputc` must be moved to "kbhit.s".  It must replace the keyboard scanner that currently is in "kbhit.s".  The line just above the `LFD47` label says, "`bne  LFD04  ; Go back and scan keyboard again`".  (That branch instruction is why we can't call the ROM function.)  It must be changed to some code that returns zero to the caller [`kbhit()` returns "false"].

`kbhit()` must be adjusted so that it behaves in this way:
1. Check `CHARBUF`; if it's not zero, then return the character (it's seen as "true").
2. Otherwise, scan the keyboard; if no character key is pressed, then return "false".
3. Else, debounce and decode the character key that is pressed.  Save the character in `CHARBUF`; and, return it.

`cgetc()` must be changed to work in this way:
1. If `CHARBUF` has a character, then jump to step 5.
2. Else, obey `cursor()`.
3. Then, repeatedly call `kbhit()` until it returns "true".  (Save the character in the X register.)
4. Hide/remove the cursor if needed.
5. Zero `CHARBUF`; and, return the character.

P.S.,
`kbhit()` doesn't need to preserve any registers.
`initcgetc:` must be moved and renamed to `initkbhit:`.

groessler:
Who has the "token" here? Anybody waiting for anybody else's changes? Is there a deadlock?

jefftranter:
smuehlst has a better pull request that addresses the review comments.
Closing mine.

PR #568, opened 2018-01-07
Guard against ld65 crashes when LineInfos is empty

benvanik:
Certain code paths in error handling can attempt to get source line info with empty LineInfos lists. This causes a precondition failure on the collection lookup.

The specific code path that was triggering this for me was a data segment containing a single condes table as the only data, meaning that there were no lines in the LineInfos collection. When the error handling in segments.c for SEG_EXPR_RANGE_ERROR (and others) triggered on the segment (for unrelated reasons) I got a precondition failure instead of a helpful error message. Other cases that create segments with no lines would also likely trigger similar errors elsewhere.

### DIFF
--- a/src/ld65/lineinfo.h
+++ b/src/ld65/lineinfo.h
@@ -155,11 +155,13 @@ INLINE const char* GetSourceNameFromList (const Collection* LineInfos)
 /* Return the name of a source file from a list of line infos */
 {
     /* The relevant entry is in slot zero */
-    return GetSourceName (CollConstAt (LineInfos, 0));
+    return CollCount(LineInfos) ? GetSourceName (CollConstAt (LineInfos, 0)) :
+                                  "<unknown>";
 }
 #else
 #  define GetSourceNameFromList(LineInfos)      \
-        GetSourceName ((const LineInfo*) CollConstAt ((LineInfos), 0))
+        (CollCount(LineInfos) ? GetSourceName ( \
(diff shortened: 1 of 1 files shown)


src/ld65/lineinfo.h
```diff
@@ -155,23 +155,27 @@ INLINE const char* GetSourceNameFromList (const Collection* LineInfos)
 /* Return the name of a source file from a list of line infos */
 {
     /* The relevant entry is in slot zero */
-    return GetSourceName (CollConstAt (LineInfos, 0));
+    return CollCount(LineInfos) ? GetSourceName (CollConstAt (LineInfos, 0)) :
+                                  "<unknown>";
```
oliverschmidt: Looking at e.g. https://github.com/cc65/cc65/blob/master/src/ca65/lineinfo.c#L253 or https://github.com/cc65/cc65/blob/master/src/co65/o65.c#L386 it seems to me that "unknown" would better match the predominant style than "<unknown>".

oliverschmidt:
Okay, the author of this pull request seems to be not interested to adjust his code according to my comment. Additionally he didn't comment my questions regarding his enhancement issue https://github.com/cc65/cc65/issues/569 so it seems to me his interest in contributing to cc65 is already gone. Given that so far nobody else complained about the issue supposed to be fixed here I'd rather not merge this pull request. If someone else opposes please reopen.

PR #570, opened 2018-01-07
Complain if unsupported flags are provided after -O

prandeamus:


prandeamus:
The compiler ignores meaningless options after -O. This change means that any meaningless options cause the compiler to stop with an error (there was no default handler in the switch).

oliverschmidt:

PR #571, opened 2018-01-09
Change the type of a cc65 variable that holds both integers and pointers.

greg-king5:
The change avoids possible pointer truncation on 64-bit MS Windows.

### DIFF
--- a/src/cc65/codegen.c
+++ b/src/cc65/codegen.c
@@ -40,6 +40,7 @@
 /* common */
 #include "check.h"
 #include "cpu.h"
+#include "inttypes.h"
 #include "strbuf.h"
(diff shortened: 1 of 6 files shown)


prandeamus:
src/cc65/exprdesc.c line 363, you've added a cast to (unsigned long). Did you mean uintptr_t , since many of the other changes are moving from unsigned long to uintptr_t ?

oliverschmidt:
> In "src/cc65/exprdesc.c", line 363, you've added a cast to (unsigned long). Did you mean uintptr_t , since many of the other changes are moving from unsigned long to uintptr_t?

No, that cast is supposed to match the "%...lX".

PR #573, opened 2018-01-10
Change none.cfg's "__STACKSTART__" to "__HIMEM__".

greg-king5:
Other configure files that let the command line control the size of the `main:` memory area use the name `__HIMEM__` to set where the top of RAM is.  This PR makes `none.cfg` be consistent with them.

### DIFF
--- a/cfg/none.cfg
+++ b/cfg/none.cfg
@@ -2,13 +2,13 @@ FEATURES {
     STARTADDRESS: default = $1000;
 }
 SYMBOLS {
-    __STACKSIZE__:  type = weak, value = $0800; # 2k stack
-    __STACKSTART__: type = weak, value = $8000;
-    __ZPSTART__:    type = weak, value = $0080;
+    __STACKSIZE__: type = weak, value = $0800; # 2k stack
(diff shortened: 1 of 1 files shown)


oliverschmidt:
I have a different perspective on this:

1. The term HIMEM was from my perspective never a general term for the end of the RAM. Rather I see the term related to a specific family of BASIC interpretes which use it to define the last address used by a BASIC program. In some aspects a cc65 is a replacement for a BASIC program. So it seemed natural to me to use the same term to define the last address used by a cc65 _on_ _that_ _platform_. If `__HIMEM__` is currently used (likely because of C&P) in linker configs for targets where HIMEM has no special meaning to users of that platform then I consider this unfortunate.

2. For the none target we want to make as little presumptions as possible. One of those presumptions is that the software stack is starting at the highest free address moving towards the end of the BSS. Maybe someone has reasons to locate the software stack just below the start address. The only thing we know for sure is that if someone uses the none library he will need a software stack and that he needs to define where that software stack is to start from.

oliverschmidt:
As there seem to be no new aspects I#m closing this...

Issue #572, opened 2018-01-09
Enhancement (minor): sim65 should call Paravirtualisation hooks on JMP (ind)


prandeamus:
This isn't hugely significant, but for the record: sim65 checks for paravirtualisation hooks on JMP abs and on JSR calls but doesn't check on JMP (indirect)

OPC_6502_20 ends with a call to ParaVirtHooks (&Regs);
OPC_6502_4C ends with a call to ParaVirtHooks (&Regs);
OPC_6502_6C has no such call, and I think it should, for completeness.

To be fair, this lack isn't breaking anything much today, but if a C program has pointer-to-function variables and one possible value of those variables is a paravirtualisation entry point like the file open call, it will fail to hook the call correctly.

prandeamus:
I think I can resolve this, but I'll get a test case together first.

oliverschmidt:
> I think I can resolve this,

I was about to ask you for that...

> but I'll get a test case together first.

Great !

prandeamus:
Modification on the impact of this: cc65 doesn't actually generate jmp (indirect) statements. If there ever is a call to pointer-to-function, it generates code to modify a JSR xx yy code stub in the library. Therefore it is less significant that I thought, and only affects indirect calls in assembly. But, for the record

In sim65, the open and close entry points are direct entry points to paravirtualised code. The code below contains code to open a file for reading and close immediately. It does it twice, once by regular calls to open and close, and the second time by calling indirectclose, which jumps via a variable jmpvec.

Note the inline assembler here . __asm__("jmp close") does work fine, but __asm("jmp (jumpvec)" fails with invalid opcode. In both cases, the JMP can be considered a shorthand for JSR xxx and RTS.

Compile and run with:
cl65 -t sim6502 jmpind.c sim6502.lib && sim65 -verbose -verbose jmpind

Results:
Loaded `jmpind' at $0200-$0369
PVOpen ("testdata", $0001)
PVClose ($0003)
PVOpen ("testdata", $0001)
Error: Illegal opcode $FF at address $FF4C

I think that without the paravirtualisation check the virtual CPU is jumping to the _close entry point without recognising it as a paravirtualisation entry point, then blundering around until it hits an illegal opcode. If the paravirtualisation worked, the trace would show a second PVClose() trace message and not an illegal opcode.

I know this all a bit obscure ... 

#include <stdio.h>
#include <stdint.h>
#include <stdlib.h>
#include <string.h>
#include <fcntl.h>
[...]

PR #574, opened 2018-01-13
Call ParaVirtHooks on JMP (indirect)

prandeamus:
Re discussion at https://github.com/cc65/cc65/issues/572
Ensure that Paravirtualisation hooks are called on JMP (indirect) calls


greg-king5:
JMP has three addressing modes.  Fix "(ind,X)" too.

prandeamus:
Ah. Another 65C02 thing. I'm showing my age now.  Yeah,  I can do that.

PR #575, opened 2018-01-19
atari5200: put default display list into its own memory area

### LINKED ISSUE #560: Meaning of "ld65: Error: atari5200/conioscreen.s(81): Display list crosses 1K boundary"
I am now getting a linker error I cannot really understand.

I

groessler:
This avoids 1K boundary crossing of the display list. Fixes #560.

### DIFF
--- a/cfg/atari5200.cfg
+++ b/cfg/atari5200.cfg
@@ -1,16 +1,18 @@
 SYMBOLS {
     __CARTSIZE__:        type = weak,   value = $4000;   # possible values: $4000 and $8000
+    __DLISTSIZE__:       type = weak,   value = $0000;
     __CART_ENTRY__:      type = import;
     __STACKSIZE__:       type = weak,   value = $0400;   # 4 pages stack
     __RESERVED_MEMORY__: type = export, value = $01E0;   # space for 20x24 screen buffer (default display list is in ROM)
 }
 MEMORY {
-    ZP:        file = "", start = $001D,                size = $00E3,                                               define = yes;
-    RAM:       file = "", start = $021C,                size = $4000 - __STACKSIZE__ - __RESERVED_MEMORY__ - $021C, define = yes;
-    ROM:       file = %O, start = $C000 - __CARTSIZE__, size = __CARTSIZE__ - $18,                                  define = yes, fill = yes, fillval = $FF;
-    CARTNAME:  file = %O, start = $BFE8,                size = $0014                                                              fill = yes, fillval = $40;
-    CARTYEAR:  file = %O, start = $BFFC,                size = $0002                                                              fill = yes, fillval = $59;
-    CARTENTRY: file = %O, start = $BFFE,                size = $0002;
+    ZP:        file = "", start = $001D,                 size = $00E3,                                               define = yes;
+    RAM:       file = "", start = $021C,                 size = $4000 - __STACKSIZE__ - __RESERVED_MEMORY__ - $021C, define = yes;
+    ROM:       file = %O, start = $C000 - __CARTSIZE__,  size = __CARTSIZE__ - __DLISTSIZE__ - $18,                  define = yes, fill = yes, fillval = $FF;
+    DLIST:     file = %O, start = $BFE8 - __DLISTSIZE__, size = __DLISTSIZE__;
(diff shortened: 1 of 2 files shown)


groessler:
Hmm, I just noticed that an empty program, not using conio still pulls in the display list. I can disable it by specifying -D__DLISTSIZE__=0 when linking, but shouldn't this be automatic? Somehow the linker pulls in the object file although __DLISTSIZE__ in the linker config is "weak".

oliverschmidt:
> Somehow the linker pulls in the object file although DLISTSIZE in the linker config is "weak".

You seem to have a very different understand of what "weak" means then I do. Me understand is that without "weak" two identical symbols are treated as errror. With one normal symbol and one "weak" symbole the "weak" symbol just gests igonored.

groessler:
I would not have expected the linker to draw in an object file for the sole purpose of resolving this symbol. After all, it already has a definition for it.

greg-king5:
I had thought that "`type = weak`" means an export that can be overridden.  But, it seems that it means an import with a default value.  Although, the ld65 document claims,
> `weak` is similar as `export`. However, the symbol is only defined if it is not defined elsewhere.

greg-king5:
Putting it at the beginning of the cart is much simpler than putting it at the end:
- Put `DLIST:` at the beginning of `ROM:`.  Its symbol names don't need to be defined (the CART* segments, also, don't need "define = yes").
- Remove `__DLISTSIZE__` from both files ("size doesn't matter").

groessler:
@greg-king5: Don't remember why I insisted to put it at the end of ROM. Maybe I considered it "cool". I also removed the "define=yes" from the CART* segments, but not from DLIST: Might come handy if someone wants to copy it to RAM in order to modifiy it.

groessler:
@greg-king5, @oliverschmidt: Seems that the behaviour of a weak symbol is weakly defined (pun intended).
Would a change be accepted that changes the behaviour to match the documentation? Specifically:

- no object file of a library is included for the sole purpose to resolve a weak symbol
- if an object file which gets included to resolve a non-weak symbol defines a non-weak version of a weak symbol, this file's value for the weak symbol is used

oliverschmidt:
@greg-king5: You wrote

> I had thought that "type = weak" means an export that can be overridden. But, it seems that it means an import with a default value. Although, the ld65 document claims,

I'm certainly not claiming that you're wrong but could you please elaborate what makes you think so (Im mean that your original understanding is wrong?). Because your original understanding seems to be identical to mine and identical to what I believe is documented.

@groessler: You wrote

> I would not have expected the linker to draw in an object file for the sole purpose of resolving this symbol. After all, it already has a definition for it.

I must admit that you totally lost me. I did the following: I took none.cfg and added
```
__ctype: type = import;
```
Just as expected a
```
cl65 -t none - C test.cfg -m test.map empty.c
```
yields a map file containing
```
c:\wrk\cc65\lib/none.lib(ctype.o):
```

Now changing the cfg to
```
__ctype: type = weak, value = 0;
```
makes the `ctype.o` disappear from the map file. This seems to be exactly the expected behavior...

groessler:
@oliverschmidt: I haven't tried to reproduce your changes yet. But look at the original version of this PR, it seems that conioscreen.o of atari5200 is pulled in just because it exports the `__DLIST_SIZE__` variable. If you compile an empty program, `int main(void) { }`.

greg-king5:
@oliverschmidt, go one step further.  Use `__ctype` in an expression in that configuration file:
```ini
AREA: file = "", start = $2000, size = __ctype;
```
It pulls "ctype.o" into the program.

groessler:
FWIW, I think the PR could be committed. It now doesn't use weak symbols and the problem we're discussing doesn't affect it anymore.

But we should continue to discuss the "weak" problem. Either here, or in a new "Issue", or on the ML.

oliverschmidt:
@greg-king5:

First of all thanks for experimenting :-)

You sound like you now successfully reproduced the "weak" issue. But that usage in an expression pulls in the symbol is expected / correct, right?

oliverschmidt:
> FWIW, I think the PR could be committed.

Right you are ;-)

Issue #497, opened 2017-09-29
[ATMOS] Bug in conio.h: inverted colors (maybe due to interference by invisible cursor)


Fabrizio-Caruso:
This is general issue and it is the cause of #492 (which I will soon close in order to keep this one)

PROBLEM:
In some targets (see below) conio functions (gotoxy, cputc) trigger the cursor to appear for an instant.
(Trying to disable the cursor, e.g., POKE(752,1) will not work because the cursor is re-activated by the next conio call).

 This causes: 
- random rapid character flicker (in ATARI, ATMOS and most probably ATARI5200 and others) and 
- random inversed colors (in ATMOS)

IMPACTED:
- ATARI (in 20x24 text mode 1 and probably any other text mode)
- ATMOS (in standard text mode)
Most probably ATARI 5200 in its standard text mode (same as 20x24 text mode 1 in ATARI).
Probably other targets are impacted.

CURRENT WORKAROUND:
I have been able to solve #492 in both ATARI and ATMOS by removing conio.h gotoxy + cputc and substituting them with direct writes onto video memory.

Fabrizio-Caruso:
```#include <cc65.h>
#include <conio.h>
#include <tgi.h>

int main(void)
{
	_graphics(1+16);
	_setcolor_low(0, TGI_COLOR_RED);
	_setcolor_low(1, TGI_COLOR_WHITE);
	_setcolor_low(2, TGI_COLOR_CYAN);
	_setcolor_low(3, TGI_COLOR_BROWN);
	_setcolor_low(4, TGI_COLOR_BLACK);
	while(1)
	{
	  char i; 
	  gotoxy(0,0);
	  for(i=0;i<8;++i)
	  {
		cputc('a'+64);
		cputc('b'+160);
	  } 
	  cgetc();
	  gotoxy(0,0); 
	  cprintf("4444");
	  cgetc();
	}
	return 0;
}
```

polluks:
If you put `cursor(1);` after cprintf it will work, strange.

Fabrizio-Caruso:
In my code for the Atari 8-bit I am using a workaround suggested by Christian.
The ATMOS case is more puzzling because I cannot reproduce it easily. It happens at random times.
So for the ATMOS version of my game I am not using conio.

oliverschmidt:
[...]

PR #576, opened 2018-01-19
Fix conio cursor, see issue #497

polluks:
Without cgetc's constructor the cursor is simply not turned off.

### DIFF
--- a/libsrc/atmos/crt0.s
+++ b/libsrc/atmos/crt0.s
@@ -1,8 +1,9 @@
 ;
 ; Startup code for cc65 (Oric version)
 ;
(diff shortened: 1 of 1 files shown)


greg-king5:
`stdin` needs the cursor.  It assumes that the cursor wasn't turned off; therefore, it doesn't enable the cursor.

polluks:
I see, it's more complex:
1. conio out
2. conio in+out
3. stdio out
4. stdio in+out

polluks:
I see, it's more complex:
1. conio out
2. conio in+out
3. stdio out
4. stdio in+out

PR #577, opened 2018-01-23
Moved initcgetc to initcputc

polluks:
joy-test.c doesn't use cgetc therefore the output is corrupted by the cursor.
Related to #497


polluks:
Related to #497

oliverschmidt:
About to merge as nobody opposed so far - and it seems to make sense to me ;-)

polluks:
Indeed, an old mistake or... challenge of Uz.

PR #588, opened 2018-02-08
Cleanup notes

polluks:


doc/funcref.sgml
```diff
@@ -3513,6 +3510,8 @@ be used in presence of a prototype.
 It allows to access the device on filesystem level by calling chdir() with
 the directory returned.
 <tag/Notes/<itemize>
+<item>Calling getdevicedir() <em/does/ check for a (formatted) disk in a
```
oliverschmidt: That phrase seems to come from my demo program (https://github.com/cc65/cc65/blob/master/samples/enumdevdir.c#L88-L90).

The emphasis on __does__ only makes sense in the context of:

*Calling getfirstdevice()/getnextdevice() does __not__ turn on the motor of a drive-type device and does __not__ check for a disk in the drive.*
(https://github.com/cc65/cc65/blob/master/samples/enumdevdir.c#L88-L90).

So it would be great if you would get that into the docs too...

oliverschmidt:
Thanks :-)

PR #592, opened 2018-02-14
Improved GeOS version detection.

blackystardust:
Improved the GEOS detection code to be able to detect gateway and megapatch3, also added a sample to get the GEOS version (geosver).

### DIFF
--- a/include/geos/gsys.h
+++ b/include/geos/gsys.h
@@ -31,6 +31,8 @@ char get_ostype(void);
 #define GEOS4           0x04    /* plus4 geos is not or'ed with version */
 #define GEOS128         0x80
   /* version flags */
+#define MEGAPATCH3      0x03
+#define GATEWAY         0x08
 #define GEOS_V10        0x10
 #define GEOS_V11        0x11
 #define GEOS_V12        0x12    /* ??? not sure */
--- a/libsrc/geos-cbm/geossym2.inc
+++ b/libsrc/geos-cbm/geossym2.inc
@@ -4,8 +4,10 @@
 ;4-2-99
 
 bootName                =       $c006
+gatewayFlag             =       $c007
 version                 =       $c00f
 nationality             =       $c010
 sysFlgCopy              =       $c012
 c128Flag                =       $c013
+mp3Flag                 =       $c014
 dateCopy                =       $c018
--- a/libsrc/geos-cbm/geossym2.inc
+++ b/libsrc/geos-cbm/geossym2.inc
@@ -11,3 +11,4 @@ sysFlgCopy              =       $c012
 c128Flag                =       $c013
 mp3Flag                 =       $c014
 dateCopy                =       $c018
+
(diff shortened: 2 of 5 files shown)


polluks:
BTW a lowercase "e" in GEOS?

blackystardust:
Change where needed, this change and other future improvements to cc65 are more part of either vice or my other projects, the pull requests are more of a 'I might as well make the changes available' than a dedicated contribution.

I got burned that way with some previous pull requests.

If the pull request is useful, please change and merge, ifnot, just get rid of it, either way the code will wind up in vice or my other projects.

oliverschmidt:
@blackystardust: Some general comments seem appropriate...

Uz spent at least at certain times quite some effort in reworking contributions to cc65 to have them fit to the existing code base. I think we all agree that this type of work isn't fun and I wouldn't be surprised if that was one of the aspects that made him drop cc65.

I didn't want to go down the same road so my policy has always been to rather invest in educating contributors then to invest in adjusting code. I generally only see two exceptions: A) obvious one-time contributors (as an investment in their education would be wasted) and B) real bug fixes (as see my self obligated in my role as maintainer to get that code in).

Your past contributions didn't fall in one of those two categories. Rather you even explicitly held out upcoming contributions so I acted as usual (which btw also seemed fair towards other contributors).

However, afterwards there was some discussion and the result was that @mrdudz joined as additional merger. He committed to assess your future contributions, merge them and do cleanup where necessary.

So as you see it wasn't only you who was unhappy with your past contributing-to-cc65 experience. And steps have been taken to pave the way for an improvement. Therefore I'm happy to see you back so we can see if things really work out as planned. However, I don't see me in the driver seat here (anymore).

Oliver

blackystardust:
@oliverschmidt 

I understand your viewpoint, and maybe I jumped the gun a bit, I'm used to a faster way of developing, when I make drivers/changes/improvements I don't wait around for days/weeks for the changes to become mainstream, and therefor I just tend to continue, even if this means building on top of changes/improvements that are not in the mainstream yet.

This fast pace of making improvements keeps me interested in doing them, and the slow pace of my previous improvements, where I couldn't continue in a reasonable amount of time put me off from making any further drivers/improvements.

I do agree that the style of a project needs to be observed, and that review is needed, a balance between fast pace and style/review might be the best way to go, but I have no idea how to realize something like that.

Besides the name (GeOS -> GEOS) I made sure to keep the codestyle the same as what was already present, so afaik I stayed within the bounds.

I would like to contribute more, especially where (missing) drivers (and improvements) are concerned, if this is appreciated and a slightly slower pace needs to be used, I will try to do so.

mrdudz:
ok, looks good to me, so i will click merge. debut \o/ :)

oliverschmidt:
Thanks for the positive feedback and explanation :-)

>  I made sure to keep the codestyle the same as what was already present, so afaik I stayed within the bounds.

Indeed !

> I would like to contribute more, especially where (missing) drivers (and improvements) are concerned, if this is appreciated and a slightly slower pace needs to be used, I will try to do so.

Great :-)

I hope that with somewhat adjusted expectations and the help of @mrdudz we'll get it done.

oliverschmidt:
@blackystardust: Thanks for the adjustments.
@mrdudz: Thanks for the action.

PR #594, opened 2018-02-23
Change the PC-Engine's config., so that we can make 8K, 16K, and 32K carts.

### LINKED ISSUE #586: [PCE][info] Does anyone know how to activate memory banks other than $E000-$FFFF?
Hi 

I am still struggling with the PCE. 
I could not find any information on how to use more than 8k in the $E000-$FFFF.
I think I know how to tell the linker to use more.
I think I know how to tell the linker what code goes where.
But the produced 16k rom (or bigger) won't start if I use anything different than a single 8k section.

I wonder whether I have to initialize the machine somehow to activate some other memory banks?
Should I use the bank attribute? How?
Does anyone here know where I can get more information?

Thanks in advance for any hint!

  Fabrizio

greg-king5:
This new code uses the method, of making a proper cartridge image, that Oliver described in issue https://github.com/cc65/cc65/issues/586#issuecomment-364428044.

The PC-Engine's document describes the post-processing that's needed to make that scheme work.  And, the Makefile, in the PCE example in the library test-code directory, shows it.

Closes #586.


oliverschmidt:
Thanks for this nice cleanup, incl. harmonization with other targets (`MAIN`) and comprehensive documentation :-)

PR #595, opened 2018-02-23
Fix sector-number conversions in the CBM GEOS DIO code.

greg-king5:
These fixes were sitting on my hard drive for about two and a half years!

[ I found those bugs while writing code for a CBM DIO implementation -- also two and a half years old. :-/ ]

### DIFF
--- a/libsrc/geos-cbm/disk/dio_cts.s
+++ b/libsrc/geos-cbm/disk/dio_cts.s
@@ -1,6 +1,6 @@
 ;
-; Maciej 'YTM/Elysium' Witkowiak
-; 2.7.2001
+; 2001-07-02, Maciej 'YTM/Elysium' Witkowiak
+; 2015-08-26, Greg King
 ;
 ;
 ; unsigned char __fastcall__ dio_phys_to_log (dhandle_t handle,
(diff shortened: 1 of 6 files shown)


oliverschmidt:
Thanks :-)

oliverschmidt:
> I found those bugs while writing code for a CBM DIO implementation -- also two and a half years old. :-/

Is there enough of that code present that others could benefit from it - or do you consider to complete it "now"?

greg-king5:
I want to add a few things to my CBM DIO code; then, I'll put it in my public Github repository, so that others can see my design choices.

Also, I need to change `dio_open()`.  Currently, it uses the same zero-origin drive numberring system that other platforms use.  But, you want it to be compatible with the getdevice functions (which use Commodore's "eight-origin" device numbers).

oliverschmidt:
> it uses the same zero-origin drive numberring system that other platforms use.

The Apple II DIO always used a getdevice functions compatible numbering. I think the zero-origin numbering wasn't a fortunate design decision anyway as it would never be clear if the numbering would only include the actual attached drives or also include potentially attached drives.

oliverschmidt:
> I want to add a few things to my CBM DIO code; then, I'll put it in my public Github repository, so that others can see my design choices.

Do you have an estimate when this will be done?

greg-king5:
> I want to add a few things to my CBM DIO code; then, I'll put it in my public Github repository, so that others can see my design choices.

It's published at https://github.com/greg-king5/cc65/tree/cbm-dio.

oliverschmidt:
Thanks !

Issue #498, opened 2017-10-02
cc65 breaks on devices or configurations where the 'zero page' is not at 0


EtchedPixels:
In various pieces of compiler generated code you find patterns of the form

STA $08
STX $09
JMP $0008

which breaks on the HuC (DP at $2000) and also on 65c816 if you are running code outside of bank 0.

I was thinking about adding a compiler option --unaliased-zp which would
1. Make the linker warn if asked to use a .zp variable for 16bit addressing
2. Tweak the compiler so in that case it generates

PSHX
PSHA
RTS   (1 clock longer)

for this case *and* for -Os  on 65C02 as it's also half the size

(it needs something uglier on 6502 - but you'd not normally build 6502 with this option...)

Thoughts ?

groessler:
What's the current status of PR #319? Anyone working on this?

EtchedPixels:
Having dug a bit the offender seems only to be in the runtime. so for now I've patched my runtime and I will link my fixed routine in crt0 so it gets used and found before the buggy cc65 one.

Using the existing jmpvec helper fixes the problem nicely (but leaves the jmpvec and interrupt bug even more of a problem)

oliverschmidt:
> Thoughts ?

A pull request with the described change would be in general welcome.

> Using the existing jmpvec helper fixes the problem nicely.

I can't follow you. Please elaborate.

EtchedPixels:
I'll send you a pull request once I've tested a bit more and looked for other cases.

EtchedPixels:
@oliverschmidt  My current status is that I have a 'sufficiently rigged demo' that I can build my Fuzix kernel on 65C816 with the kernel code in bank1, data in bank2 and ZP / S in bank 0.

To do that I've had to
- Do some hacks around jmpvec and the C support code
- Add a custom setjmp
- Ensure I don't place any data in 00xx otherwise the compiler/assembler optimises it to ZP
- Some tweaks where the compiler tries to write self modifying code and assumes code == data

To do it all properly also needs a compiler tweak. I want to spend the time to get it all working and right and tidy first

oliverschmidt:
Thanks for the quick feedback :-) So we'll just keep this issue open as-is you'll come back with news when you're ready...

EtchedPixels:
https://github.com/EtchedPixels/FUZIX/wiki/Using-CC65-on-the-65C816

is where I am today

pinobatch:
Related bugs include #194 and #317.

[...]

PR #599, opened 2018-02-27
Fix how the PC-Engine calls functions through pointers.

### LINKED ISSUE #596: [PCE][function pointers](With isolated example) Crashes whenever pointers to functions are used.
My game crashes ONLY with the PCE target whenever in my code I call a function through a pointer to function.
I get different crash/freeze behavior depending on the size of the stack.
I have tried with stack size from 0x20 to 0x300.
The problem could be stack-related.
I will try to isolate this issue.
I use the standard cfg with cart size 16k and with different stack sizes:
```
SYMBOLS {
    __CARTSIZE__:  type = weak, value = $4000; # $2000, $4000, or $8000
    __STACKSIZE__: type = weak, value = $0300; # 3 pages stack
}
```
I compile and build my game with:
```
pce_16k:
	$(CC65_PATH)$(MYCC65) -O -t pce \
	-DFULL_GAME \
	-DEND_SCREEN -DBETWEEN_LEVEL \
	--config $(SOURCE_PATH)/../cfg/pce_16k_less_stack.cfg \
	$(SOURCE_PATH)/horizontal_missile.c $(SOURCE_PATH)/rocket.c \
	$(SOURCE_PATH)/item.c $(SOURCE_PATH)/end_screen.c \
	$(SOURCE_PATH)/display_macros.c \
	$(SOURCE_PATH)/enemy.c $(SOURCE_PATH)/invincible_enemy.c \
	$(SOURCE_PATH)/level.c $(SOURCE_PATH)/character.c $(SOURCE_PATH)/text.c \
	$(SOURCE_PATH)/missile.c $(SOURCE_PATH)/strategy.c \
	$(SOURCE_PATH)/input_macros.c $(SOURCE_PATH)/main.c \
	-o $(BUILD_PATH)/FULL_PCE.bin
	dd if=$(BUILD_PATH)/FULL_PCE.bin ibs=1 count=8192 of=$(BUILD_PATH)/FULL_PCE_LOW.bin
	dd if=$(BUILD_PATH)/FULL_PCE.bin ibs=1 skip=8192 of=$(BUILD_PATH)/FULL_PCE_HIGH.bin
	rm $(BUILD_PATH)/FULL_PCE.bin 
	cat $(BUILD_PATH)/FULL_PCE_HIGH.bin $(BUILD_PATH)/FULL_PCE_LOW.bin > $(BUILD_PATH)/FULL_PCE_SWAPPED.pce
	rm $(BUILD_PATH)/FULL_PCE_LOW.bin
	rm $(BUILD_PATH)/FULL_PCE_HIGH.bin	
```
(I know I am not using dd in the most efficient way but it is super fast anyway)

greg-king5:
The problem was mentioned by @EtchedPixels in #498.

Two call functions in the `runtime` library are used to call functions through pointers.  They store the pointer in zero-page variables; then, jump indirectly through those variables.

But, the HuC6280 CPU makes zero-page addressing modes go to addresses $2000 - $20FF.  Most other addressing modes aren't redirected.  Therefore, the indirect jumps don't access the same locations that the zero-page store instructions access.  Programs that use function pointers will crash on the PC-Engine.

This patch substitutes, on the PC-Engine, library functions that jump through locations in $2000 - $20FF instead of $0000 - $00FF.

Fixes #596.

### DIFF
--- a/libsrc/pce/call.s
+++ b/libsrc/pce/call.s
@@ -1,0 +1,16 @@
+;
+; CC65 runtime: call function via pointer in ax
+;
+; 1998-08-06, Ullrich von Bassewitz
+; 2018-02-28, Greg King
+;
+
+        .export         callax
+        .importzp       ptr1
+
+callax: sta     ptr1
+        stx     ptr1+1
+
+; The PC-Engine puts the zero-page at $2000.
+
+        jmp     (ptr1 + $2000)  ; go there
--- a/libsrc/pce/callptr4.s
+++ b/libsrc/pce/callptr4.s
@@ -1,0 +1,14 @@
+;
+; CC65 runtime: call function via pointer in ptr4
+;
+; 2018-02-28, Greg King
+;
+
+        .export         callptr4
+        .importzp       ptr4
+
+callptr4:
+
+; The PC-Engine puts the zero-page at $2000.
+
+        jmp     (ptr4 + $2000)


oliverschmidt:
As far as I understand the Makefile it should be possible to do this in the same style as https://github.com/cc65/cc65/blob/master/libsrc/apple2/_sys.s, https://github.com/cc65/cc65/blob/master/libsrc/pce/memcpy.s, https://github.com/cc65/cc65/blob/master/libsrc/pce/memmove.s and https://github.com/cc65/cc65/blob/master/libsrc/pce/memset.s although the shared files are here not located in `common` but in `runtime`. If I'm correct I ask you to apply that style here too.

greg-king5:
OK, two substitute files instead of two changed files.

oliverschmidt:
Thanks for adjusting the change according to my comment :-)

Fabrizio-Caruso:
This fix does not fix #596, at least not with the currently available snapshot (10:29 p.m. C.E.T.).

Fabrizio-Caruso:
The very simple example I posted, still crashes.

greg-king5:
I used your test code to verify the fix.  It shows "Hello world!" to me.

Fabrizio-Caruso:
@greg-king5 
The test program would show a flickering hello world (as if the program is continuously restarted?). It would eventually creash showing random characters on the screen. (Tested with mednafen).
CROSS CHASE immediately hangs whenever the portion of the code with pointers to functions is entered.

Fabrizio-Caruso:
@greg-king5 I should have written a better test program that shows the problem in an immediate way.

Fabrizio-Caruso:
@greg-king5 do you get the same error after seeing the flickering hello?

Fabrizio-Caruso:
@all, @oliverschmidt, @greg-king5 
The following example will prove that there is a problem, by only displaying "world" and not "Hello"

```
#include <stdio.h>
#include <conio.h>

void hello(void)
{
	gotoxy(3,5);
	textcolor(COLOR_WHITE);
	cprintf("Hello\n");
}

struct objectStruct
{
	void (* foo) (void);
};

typedef struct objectStruct Object;



int main()
{
	Object obj;
	obj.foo = hello;
	
	obj.foo();
	
	gotoxy(3,7); cprintf("world!\n");
	
	while(1){};
	
	return 0;
}

```

PR #600, opened 2018-02-28
Update docu

polluks:
More conio targets


oliverschmidt:
Thanks for your continuous care for the docs :-)

PR #602, opened 2018-03-01
 Added 2a03/2a07 and HuC6280 detection. 

blackystardust:
Added 2a03/2a07 and HuC6280 detection.


oliverschmidt:
The predominant style for hex digits in cc65 is upper case. Afaik this is consistent at least in the shared code. It would be nice if this file would be adjusted either before merging or after merging...

blackystardust:
ok, will fix, and will keep it in mind for future contributions.

blackystardust:
Fixed the hex digit cases in my latest commit.

oliverschmidt:
> ok, will fix, and will keep it in mind for future contributions.

I appreciate that :-)

PR #603, opened 2018-03-05
fix build in MSVC

peterferrie:


oliverschmidt:
Wow, what a coincidence !

After going undetected for many months I fixed this very issue with same two changes this very day... https://github.com/cc65/cc65/commit/19f3229f45262da83c910650c382f5bdd510d022

Anyhow, thanks for your contribution - not only to cc65 but to the Apple II communicty in general !!

PR #606, opened 2018-03-07
Optimize deceaxy

pfusik:


oliverschmidt:
I think that if this change would have totally broken the things than our regression running on Travis CI would show it. However, I'd still like to have an additional opinion on this change before merging it.

mrdudz:
i suggest some comments :) besides that, looks OK to me though

oliverschmidt:
Sorry for my usual nitpicking but...
- Please have blank line before and after comment lines, see e.g. https://github.com/cc65/cc65/blob/master/libsrc/runtime/lasr.s
- Please use upper case chars for hex constants ($ff -> $FF)

oliverschmidt:
@pfusik : Thanks for your contribution (and your patience)
@mrdudz : Thanks for your feedback

PR #608, opened 2018-03-09
Improved tv timing detection for the c64.

blackystardust:
Improved tv timing detection for the c64.

This does not work correctly when there is a cpu that has a turbo mode, this will be dealt with in a future pull request.

### DIFF
--- a/asminc/get_tv.inc
+++ b/asminc/get_tv.inc
@@ -11,6 +11,8 @@
 .enum TV
     NTSC
     PAL
+    NTSC_OLD
+    PAL_N
     OTHER
 .endenum
 
--- a/include/cbm.h
+++ b/include/cbm.h
@@ -150,7 +150,9 @@ struct cbm_dirent {
 
 #define TV_NTSC         0
 #define TV_PAL          1
-#define TV_OTHER        2
+#define TV_NTSC_OLD     2
+#define TV_PAL_N        3
+#define TV_OTHER        4
 
 unsigned char get_tv (void);
 /* Return the video mode the machine is using. */
--- a/libsrc/c64/get_tv.s
+++ b/libsrc/c64/get_tv.s
@@ -4,6 +4,11 @@
 ; unsigned char get_tv (void);
 ; /* Return the video mode the machine is using */
 ;
+; Changed to actually detect the mode instead of using a flag
+; Marco van den Heuvel, 2018-03-08
+;
+; The detection goes wrong on accelerated machines for now.
+;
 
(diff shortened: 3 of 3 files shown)


oliverschmidt:
Sorry if anybody but me understands that I'm just missing the point but...

I'm wondering if a typical cc65 users sees this change as desirable. It increases the code size by ~45 bytes while I imagine that for the average user the current "naive" 50/60 Hz is just good enough.

This btw also happens inside the C library itself: The C64 time() implementation just wants to distinguish between 50 and 60 Hz (https://github.com/cc65/cc65/blob/709ee6a28b1375222d44590a57e93d732966f8bf/libsrc/c64/systime.s#L75) but becomes ~45 bytes larger.

The comment '_The detection goes wrong on accelerated machines for now._' sort of implies that you intend to "add even more code" to the function. I'm unsure but maybe it would make sense to rather have get_tv() keep just reading the flag and add a new function that gives more precise information - where it is needed/desired (?).

mrdudz:
Functions dealing with TOD time only need 50Hz vs 60Hz - and infact, these functions should NOT use the tv mode at all! for example a PAL C64SX has 50Hz video (PAL), but the TOD ticks are 60Hz (usually they would come from the power grid, but SX generates them on board, always 60Hz regardless of video system).

so to get this 100% right there must be a new function that determines the TOD tick frequency (not sure how to do this, probably needs some kind of loop that measures the ticks).

the tv-mode is an entirely different thing, and to universally work for everyone, of course it should detect pal/ntsc/ntsc-old/pal-n - this info is needed for whoever wants to write code that relies on frame timing. for this i suggest the "sokrates variant" shown on codebase: http://codebase64.org/doku.php?id=base:detect_pal_ntsc

silverdr:
> On 2018-03-09, at 17:37, Bob Andrews <notifications@github.com> wrote:
> 
> Functions dealing with TOD time only need 50Hz vs 60Hz - and infact, these functions should NOT use the tv mode at all! for example a PAL C64SX has 50Hz video (PAL), but the TOD ticks are 60Hz (usually they would come from the power grid, but SX generates them on board, always 60Hz regardless of video system).

Precisely. Detecting video system is one thing (timing-wise there is more than just PAL and NTSC) and detecting TOD frequency is another thing. In contrary to popular belief they are completely unrelated. I have a set of routines to detect virtually every piece of hardware config. I can supply those if there is interest. I _always_ detect TOD freq whenever I need to use it. Otherwise you may get useless results.

-- 

PR #609, opened 2018-03-11
Added missing PCE conio functions.

polluks:


libsrc/pce/cclear.s
```diff
@@ -0,0 +1,25 @@
...
+_cclearxy:
+        pha                     ; Save the length
+        jsr     gotoxy          ; Call this one, will pop params
+        pla                     ; Restore the length and run into _cclear
+
+_cclear:
+        cmp     #0              ; Is the length zero?
+        beq     L9              ; Jump if done
+        sta     tmp1                                 
+L1:     lda     #$20            ; Blank - screen code
```
oliverschmidt: Maybe "screen code" is a term that doesn't apply (well) to this target (?).

polluks: Not an official term but it fits IMHO.

PR #617, opened 2018-03-19
Added the vic20 georam emd.

blackystardust:
Added the vic20 georam emd.


oliverschmidt:
A new EM driver doesn't involve the risk of causing a regression in existing code. So given that the contribution seems both complete and follow our style guides I tend to jump the gun here...

PR #619, opened 2018-03-20
Added isfast function which returns a 1 when the C128 is in 2MHz mode.

blackystardust:
Added an 'unsigned char isfast(void)' function for the C128 target which returns 1 when the C128 is in 2Mhz mode.

### DIFF
--- a/libsrc/c128/isfast.s
+++ b/libsrc/c128/isfast.s
@@ -1,0 +1,20 @@
+;
+; Marco van den Heuvel, 2018-03-19
+;
+; unsigned char isfast (void);
+; /* Returns 1 if the CPU is in 2MHz mode. */
+;
+
+        .export         _isfast
+
+        .include        "c128.inc"
+
+
+.proc   _isfast
+
+        lda     VIC_CLK_128
+        and     #$01
+        rts
+
+.endproc
+
--- a/libsrc/c128/isfast.s
+++ b/libsrc/c128/isfast.s
@@ -14,6 +14,7 @@
 
         lda     VIC_CLK_128
         and     #$01
+        ldx     #$00
         rts
 
 .endproc
--- a/doc/c128.sgml
+++ b/doc/c128.sgml
@@ -87,6 +87,7 @@ url="funcref.html" name="function reference"> for declaration and usage.
 <item>c64mode
 <item>fast
 <item>slow
+<item>isfast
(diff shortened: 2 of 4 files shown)


libsrc/c128/isfast.s
```diff
@@ -0,0 +1,20 @@
...
+        .export         _isfast
+
+        .include        "c128.inc"
+
+
+.proc   _isfast
+
+        lda     VIC_CLK_128
+        and     #$01
+        rts
```
oliverschmidt: You need to set X, see https://github.com/cc65/wiki/wiki/Parameter-passing-and-calling-conventions#Overview

blackystardust: Ah, I see, ok, will fix in a few.

blackystardust:
Set X register to 0 so that 16bit promotion works as expected.

polluks:
Maybe fast() should simply return the speed bit?

mrdudz:
those are completely unrelated, videomode returns pal/ntsc, this function returns 1mhz/2mhz

to me it looks ok, merge? :)

oliverschmidt:
ACK.

polluks:
@mrdudz C'mon, I was talking about the return value. videomode returns the old value, that's it.

oliverschmidt:
It seemes to me that videomode is more about wanting to set some mode at program start and wanting to set the mode back at program exit. There the "return old value pattern" seems natural.

However, with the speed I rather see a (or the even the usual) usecase in "just" wanting to know the current speed to adjust some program logic to it - and not necessarily change it / assert some setting.

But maybe I'm missing the point...

mrdudz:
> @mrdudz C'mon, I was talking about the return value. videomode returns the old value, "isvideomode" would not be necessary.

that still doesnt make sense to me. but neither makes the implementation of videomode to be honest :)

polluks:
Well, I just like to fuse tiny functions foo and isfoo.

greg-king5:
`fast()` is for programs that want to change the CPU speed.  `isfast()` is for programs that _don't_ want to change the current speed.

polluks:
@greg-king5 For sure, do you have any real world example?

greg-king5:
> @greg-king5 For sure, do you have any real world example?

No, it's just common sense.

PR #620, opened 2018-03-23
Added c16/plus4 fast(), isfast() and slow() functions, and updated the documentation accordingly.

blackystardust:
Added c16/plus4 fast(), isfast() and slow() functions, and updated the documentation accordingly.

### DIFF
--- a/include/plus4.h
+++ b/include/plus4.h
@@ -59,6 +59,14 @@
 extern void plus4_stdjoy_joy[];   /* Referred to by joy_static_stddrv[] */
 extern void plus4_stdser_ser[];
 
+void fast (void);
+/* Switch the CPU into double clock mode. */
+
+void slow (void);
+/* Switch the CPU into single clock mode. */
+
+unsigned char isfast (void);
+/* Returns 1 if the CPU is in double clock mode. */
 
 
 /* End of plus4.h */
--- a/asminc/plus4.inc
+++ b/asminc/plus4.inc
@@ -73,6 +73,7 @@ TED_CURSLO      := $FF0D
 TED_V1FRQLO     := $FF0E
 TED_V2FRQLO     := $FF0F
 TED_V2FRQHI     := $FF10
+TED_CLK         := $FF13
 TED_BGCOLOR     := $FF15
 TED_COLOR1      := $FF16
 TED_COLOR2      := $FF17
--- a/doc/c128.sgml
+++ b/doc/c128.sgml
@@ -85,9 +85,6 @@ url="funcref.html" name="function reference"> for declaration and usage.
 <itemize>
 <item>videomode
 <item>c64mode
-<item>fast
-<item>slow
-<item>isfast
 </itemize>
 
 
@@ -120,6 +117,19 @@ declaration and usage.
 <item>cbm_save
 <item>cbm_write
 <item>get_tv
+</itemize>
+
+
+<sect1>CBM specific CPU functions<p>
+
+Some CPU related functions are available for some of the Commodore
+machines. See the <url url="funcref.html" name="function reference"> for
+declaration and usage.
+
+<itemize>
+<item>fast
+<item>slow
+<item>isfast
 </itemize>
 
 
--- a/doc/c16.sgml
+++ b/doc/c16.sgml
@@ -126,6 +126,18 @@ declaration and usage.
 </itemize>
 
 
+<sect1>CBM specific CPU functions<p>
+
+Some CPU related functions are available for some of the Commodore
+machines. See the <url url="funcref.html" name="function reference"> for
+declaration and usage.
+
+<itemize>
+<item>fast
+<item>slow
+<item>isfast
+</itemize>
+
 <sect1>Hardware access<p>
 
 The following pseudo variables declared in the <tt/c16.h/ header file do
--- a/doc/funcref.sgml
+++ b/doc/funcref.sgml
@@ -140,6 +140,12 @@ function.
 
 <sect1><tt/c16.h/<label id="c16.h"><p>
 
+<itemize>
+<item><ref id="fast" name="fast">
+<item><ref id="isfast" name="isfast">
+<item><ref id="slow" name="slow">
+</itemize>
+
 (incomplete)
 
 
@@ -501,6 +507,12 @@ It does not declare any functions.
 
 
 <sect1><tt/plus4.h/<label id="plus4.h"><p>
+
+<itemize>
+<item><ref id="fast" name="fast">
+<item><ref id="isfast" name="isfast">
+<item><ref id="slow" name="slow">
+</itemize>
 
 (incomplete)
 
@@ -3292,21 +3304,21 @@ program, it may not be able to read it.
 
 <quote>
 <descrip>
-<tag/Function/Switch the C128 into 2MHz mode.
-<tag/Header/<tt/<ref id="c128.h" name="c128.h">/
+<tag/Function/Switch the CPU into fast mode (C128: 2MHz mode, C16/Plus4: double clock mode).
+<tag/Header/<tt/<ref id="c128.h" name="c128.h">,
+<ref id="c16.h" name="c16.h">, <ref id="plus4.h" name="plus4.h">/
 <tag/Declaration/<tt/void fast (void);/
-<tag/Description/The function will switch the clock of the C128 to 2MHz. This
-will nearly double the speed compared to slow mode.
+<tag/Description/The function will switch the clock of the CPU to fast mode. For the C128
+target it means switching the CPU into 2MHz mode. For the C16/Plus4 target it means
+switching the CPU into double clock mode.
 <tag/Notes/<itemize>
-<item>The function is specific to the C128.
-<item>2MHz clock will not work in 40 column mode.
+<item>The function is specific to the C128, C16 and Plus4.
+<item>On the C128 the 2MHz clock will not work in 40 column mode.
 </itemize>
-<tag/Availability/C128
+<tag/Availability/cc65 (not all platforms)
 <tag/See also/
 <ref id="isfast" name="isfast">,
 <ref id="slow" name="slow">,
-<ref id="toggle_videomode" name="toggle_videomode">,
-<ref id="videomode" name="videomode">
 <tag/Example/None.
 </descrip>
 </quote>
@@ -3960,19 +3972,18 @@ fastcall function, so it may only be used in presence of a prototype.
 
 <quote>
 <descrip>
-<tag/Function/Check if the C128 is in 2MHz mode.
-<tag/Header/<tt/<ref id="c128.h" name="c128.h">/
+<tag/Function/Check if the CPU is in fast mode (C128: 2MHz mode, C16/Plus4: double clock mode).
+<tag/Header/<tt/<ref id="c128.h" name="c128.h">,
+<ref id="c16.h" name="c16.h">, <ref id="plus4.h" name="plus4.h">/
 <tag/Declaration/<tt/unsigned char isfast (void);/
-<tag/Description/The function returns a 1 if the C128 is in 2MHz mode.
+<tag/Description/The function returns a 1 if the CPU is in fast mode (C128: 2MHz mode, C16/Plus4: double clock mode).
 <tag/Notes/<itemize>
-<item>The function is specific to the C128.
+<item>The function is specific to the C128, C16 and Plus4.
 </itemize>
-<tag/Availability/C128
+<tag/Availability/cc65 (not all platforms)
 <tag/See also/
 <ref id="fast" name="fast">,
 <ref id="slow" name="slow">,
-<ref id="toggle_videomode" name="toggle_videomode">,
-<ref id="videomode" name="videomode">
 <tag/Example/None.
 </descrip>
 </quote>
@@ -6086,20 +6097,20 @@ be used in presence of a prototype.
 
 <quote>
 <descrip>
-<tag/Function/Switch the C128 into 1MHz mode.
-<tag/Header/<tt/<ref id="c128.h" name="c128.h">/
+<tag/Function/Switch the CPU into slow mode (C128: 1MHz mode, C16/Plus4: single clock mode).
+<tag/Header/<tt/<ref id="c128.h" name="c128.h">,
+<ref id="c16.h" name="c16.h">, <ref id="plus4.h" name="plus4.h">/
 <tag/Declaration/<tt/void slow (void);/
-<tag/Description/The function will switch the clock of the C128 to 1MHz. This
-will halve the speed compared to fast mode.
+<tag/Description/The function will switch the clock of the CPU to slow mode. for the C128
+target it means switching the CPU into 1MHz mode. for the C16/Plus4 target it means
+switching the CPU into single clock mode.
 <tag/Notes/<itemize>
-<item>The function is specific to the C128.
+<item>The function is specific to the C128, C16 and Plus4.
 </itemize>
-<tag/Availability/C128
+<tag/Availability/cc65 (not all platforms)
 <tag/See also/
 <ref id="fast" name="fast">,
 <ref id="isfast" name="isfast">,
-<ref id="toggle_videomode" name="toggle_videomode">,
-<ref id="videomode" name="videomode">
 <tag/Example/None.
 </descrip>
 </quote>
--- a/doc/plus4.sgml
+++ b/doc/plus4.sgml
@@ -124,6 +124,19 @@ declaration and usage.
 </itemize>
 
 
+<sect1>CBM specific CPU functions<p>
+
+Some CPU related functions are available for some of the Commodore
+machines. See the <url url="funcref.html" name="function reference"> for
+declaration and usage.
+
+<itemize>
+<item>fast
+<item>slow
+<item>isfast
+</itemize>
+
+
 <sect1>Hardware access<p>
 
 The following pseudo variables declared in the <tt/plus4.h/ header file do
--- a/libsrc/c16/fast.s
+++ b/libsrc/c16/fast.s
@@ -1,0 +1,22 @@
+;
+; Marco van den Heuvel, 2018-03-20
+;
+; void fast (void);
+; /* Switch the CPU into double clock mode. */
+;
+
+        .export         _fast
+
+        .include        "plus4.inc"
+
+
+.proc   _fast
+
+        lda     TED_CLK
+        and     #%11111101
+        sta     TED_CLK
+        rts
+
+.endproc
+
+
--- a/libsrc/c16/isfast.s
+++ b/libsrc/c16/isfast.s
@@ -1,0 +1,22 @@
+;
+; Marco van den Heuvel, 2018-03-20
+;
+; unsigned char isfast (void);
+; /* Returns 1 if the CPU is in double clock mode. */
+;
+
+        .export         _isfast
+
+        .include        "plus4.inc"
+
+
+.proc   _isfast
+
+        lda     TED_CLK
+        lsr
+        and     #$01
+        ldx     #$00
+        rts
+
+.endproc
+
--- a/libsrc/c16/slow.s
+++ b/libsrc/c16/slow.s
@@ -1,0 +1,22 @@
+;
+; Marco van den Heuvel, 2018-03-28
+;
+; void slow (void);
+; /* Switch the CPU into single clock mode. */
+;
+
+        .export         _slow
+
+        .include        "plus4.inc"
+
+
+.proc   _slow
+
+        lda     TED_CLK
+        ora     #%00000010
+        sta     TED_CLK
+        rts
+
+.endproc
+
+
--- a/libsrc/plus4/fast.s
+++ b/libsrc/plus4/fast.s
@@ -1,0 +1,22 @@
+;
+; Marco van den Heuvel, 2018-03-20
+;
+; void fast (void);
+; /* Switch the CPU into double clock mode. */
+;
+
+        .export         _fast
+
+        .include        "plus4.inc"
+
+
+.proc   _fast
+
+        lda     TED_CLK
+        and     #%11111101
+        sta     TED_CLK
+        rts
+
+.endproc
+
+
--- a/libsrc/plus4/isfast.s
+++ b/libsrc/plus4/isfast.s
@@ -1,0 +1,22 @@
+;
+; Marco van den Heuvel, 2018-03-20
+;
+; unsigned char isfast (void);
+; /* Returns 1 if the CPU is in double clock mode. */
+;
+
+        .export         _isfast
+
+        .include        "plus4.inc"
+
+
+.proc   _isfast
+
+        lda     TED_CLK
+        lsr
+        and     #$01
+        ldx     #$00
+        rts
+
+.endproc
+
--- a/libsrc/plus4/slow.s
+++ b/libsrc/plus4/slow.s
@@ -1,0 +1,22 @@
+;
+; Marco van den Heuvel, 2018-03-28
+;
+; void slow (void);
+; /* Switch the CPU into single clock mode. */
+;
+
+        .export         _slow
+
+        .include        "plus4.inc"
+
+
+.proc   _slow
+
+        lda     TED_CLK
+        ora     #%00000010
+        sta     TED_CLK
+        rts
+
+.endproc
+
+


oliverschmidt:
I like the consistency improvements for the documentation :-)

If no-one opposes I'll merge this soon.

plbyrd:
Is there an isfast for the 128?

Sent from Mail for Windows 10

groessler:
Hmm, why have fast(), isfast() and slow() functions? Where is isslow()?

Wouldn't it be better to have setspeed() and getspeed() functions?

include/plus4.h
```diff
@@ -59,6 +59,14 @@
...
 extern void plus4_stdser_ser[];
 
+void fast (void);
+/* Switch the CPU into double clock mode. */
+
+void slow (void);
+/* Switch the CPU into single clock mode. */
+
+unsigned char isfast (void);
+/* Returns 1 if the CPU is in double clock mode. */
```
greg-king5: Those declarations belong in `<cbm264.h>`.

oliverschmidt: @greg-king5: Are you willing to adjust this (either via direct push to master or via pull rquest) ?

oliverschmidt:
> Hmm, why have fast(), isfast() and slow() functions? Where is isslow()?

fast() and slow() seem to be present of the C128 for quite some time. So adding isfast() seemed the most appropriate way. I presume that you question about isslow() isn't meant for real ;-)

Then the C128 approach was ported 1:1 to other CBM machines. Again the most appropriate approach from my perspective.

> Wouldn't it be better to have setspeed() and getspeed() functions?

Yep. i'm personally willing to merge a pull request that changes that for all three machines. However, usually you are more about backward compatibility than I am so maybe you're hesitant given the exsisting fast() / slow() on the C128. I'm open...

include/plus4.h
greg-king5: > Are you willing to adjust this (either via direct push to master, or via pull request)?

Yes, very easy to move.

greg-king5:
A little bit of history:
Someone wanted cc65 programs to be able to do what C128 BASIC programs can do.
The `fast()` and `slow()` names came directly from that BASIC.

But, `setspeed()` does seem to be a more appropriate name for a system library function.

include/plus4.h
oliverschmidt: :-)

blackystardust:
I would have no problem in rewriting both the c128 and plus4 fast, slow and isfast functions into setspeed() and getspeed() functions, however, maybe setcpuspeed() and getcpuspeed() would be more appropriate and descriptive.

oliverschmidt:
> The fast() and slow() names came directly from that BASIC.

I see.

> But, setspeed() does seem to be a more appropriate name for a system library function.

It seems that we're all into the setcpuspeed() / getcpuspeed() approach. I presume setcpuspeed() returns the previous speed setting. Maybe it would be good to have macros along the lines of the VIDEOMODE macros:
https://github.com/cc65/cc65/blob/master/include/c128.h#L101-L105
https://github.com/cc65/cc65/blob/master/include/apple2enh.h#L76-L80

This would allow two things:
- Maybe/likely save a few bytes of code by not requiring that "slow" maps to 0 and "fast" maps to 1.
- Allowing for cross target macros like CPUSPEED_HIGH and CPUSPEED_LOW on the one hand and target specific macros like CPUSPEED_2MHZ and and CPUSPEED_1MHZ on the other hand.

blackystardust:
@oliverschmidt 

That would work, and allow for CPUSPEED_20MHZ for SuperCPU, and quite a selection of speeds for multispeed carts like the chameleon.

oliverschmidt:
> That would work, and allow for CPUSPEED_20MHZ for SuperCPU, and quite a selection of speeds for multispeed carts like the chameleon.

Hm, but support for all that would make the implementation much larger. And in general the cc65 approach is to support only stock hardware / firmware (apart from drivers of course).

blackystardust:
My thinking was:
- for native cpu's have/use get/setcpuspeed()
- for accelerator cpu's have cart specific functions like set/get_supercpu_speed(), set/get_chameleon_speed()...etc
- and generic accelerator functions that detect and uses the cart specific code, so get/set_accelerator_speed()

That way any native cpu code is small, any accelerator specific code is small, and only code written to be able to use any cpu will be larger.

polluks:
> I presume setcpuspeed() returns the previous speed setting...

I'll get my return value! ☺

oliverschmidt:
> My thinking was: [...]

I see your point. Code not interested in supporting custom hardware isn't bloated by custom hardware support :-)

However, I'm still hesitant to generally add custom hardware/firmware (think alternative DOS) support to the C Libarary.

So what about not having cart-specific functions but only generic accelerator functions. That is something that could be mapped to Apple II (and ATARI ?) accelerators too.

My proposal:

* getcpuspeed()/setcpuspeed() with generic CPUSPEED_LOW/CPUSPEED_HIGH (with target specific values) mapped to CPUSPEED_1MHZ/CPUSPEED_2MHZ or CPUSPEED_SINGLE/CPUSPEED_DOUBLE depending on the target. setcpuspeed() returns the same value as getcpuspeed().

* getacceleratorspeed()/setacceleratorspeed() which detects which accelerator is present (if any at all).

  * getacceleratorspeed() returns ACCELERATORSPEED_NONE if no (known) accelerator is present. Other possible values are ACCELERATORSPEED_LOW if a (known) accelerator was found but is inactive and ACCELERATORSPEED_CHAMELION or ACCELERATORSPEED_SUPERCPU for an active (known) accelerator. 

    * ACCELERATORSPEED_HIGH is *_never_* returned.

  * setacceleratorspeed() takes the values ACCELERATORSPEED_LOW to deactivate any (known) accelerator and ACCELERATORSPEED_HIGH to activate any (known) accelerator. Additional values are ACCELERATORSPEED_CHAMELION or ACCELERATORSPEED_SUPERCPU to only active the names accelerator. The value ACCELERATORSPEED_NONE means a NOP. setacceleratorspeed() returns the same value as getacceleratorspeed().

    * A successful operation with ACCELERATORSPEED_LOW or ACCELERATORSPEED_HIGH can always be directly detected by a return value different from ACCELERATORSPEED_NONE.

    * A successful operation with ACCELERATORSPEED_CHAMELION or ACCELERATORSPEED_SUPERCPU can only be detected by issuing the identical operation twice and checking the second return value to be identical to the parameter. These operations are therefore not recommended for general use. Rather they are (together with ACCELERATORSPEED_NONE) primarily provided to allow to simply re-enable any previously enabled mode by code like:

```
oldspeed = setacceleratorspeed(ACCELERATORSPEED_HIGH);

...

setacceleratorspeed(oldspeed);
```

polluks:
@oliverschmidt Nice draft :-)
How about some indicator if fast turns off video?

mrdudz:
not so sure that makes sense - you really want more specific things. for example SCPU has various optimization modes that you might want to enable (or not). and the chameleon can run at various speeds (2,3,4,5,6,7mhz).

greg-king5:
I think that
* `ACCELERATORSPEED_NONE` should be spelled `ACCELERATORSPEED_NO`.
* If a specific accelerator is named with `setacceleratorspeed()`, then it should be a filter.  Only it should be detected; all other devices should be the same as "no device" -- they should return `ACCELERATORSPEED_NO` (a failed operation).

greg-king5:
> How about some indicator if fast turns off video?

As far as I know, the C128's 40-column screen is the only place where that matters.  So, maybe a preprocessor guard and `videomode()` might be enough.

polluks:
@greg-king5 also c16 and plus/4

oliverschmidt:
> you really want more specific things

I don‘t think so. It‘s generally questionable if an accelerator interface is appropriate for cc65. But if it is, then it certainly doesn‘t reach beyond turning the accelerator on and off. If a certain accelerator can‘t be operated on that basis in a reasonable way then I question support for that accelerator in general.

Maybe I‘m once more too naive too understand the C64 specifics but from my experience an accelerator is either active on cold boot with some default config (or some persistently saved config). Or it supports some short cut to turn it on (with default/saved config). The goal of the interface discussed here is to programatically activate that very default/saved config and deactivate the accelerator altogether - nothing more.

oliverschmidt:
> If a specific accelerator is named with setacceleratorspeed(), then it should be a filter. Only it should be detected; all other devices should be the same as "no device" -- they should return ACCELERATORSPEED_NO (a failed operation).

That approach seems intuitive at first thought - in fact I had already written it that way myself ;-)

However, the overall rule of the interface is that
* set...speed() returns the previous speed - and that speed can‘t depend on the parameter of set...speed().
* set...speed() returns the same value as get...speed().

Both rules would be violated by the „filter approach“.

polluks:
An example: plasma.c
After "Making charset, mompls" we should call set...speed()

mrdudz:
> I don‘t think so. It‘s generally questionable if an accelerator interface is appropriate for cc65. But if it is, then it certainly doesn‘t reach beyond turning the accelerator on and off. If a certain accelerator can‘t be operated on that basis in a reasonable way then I question support for that accelerator in general.

i wholeheartly disagree. i dont see a single reason for any of that either. yes, there should be a simple and no-overhead way to enable/disable acceleration.  (for this fast/slow functions are sufficient) - however, for example when doing stuff for scpu you really always want to choose an optimization mode that suits your needs (else the generic function would have to disable all of them, wasting a lot of potential speedup).

making a function to get the current speed, or even returning the previous speed on setting it, sounds like overengineering without any practical use on the other hand :)

actually, it could be argued, that any cc65 compiled program would simply call fast() at startup. because really, why not?

oliverschmidt:
> making a function to get the current speed, or even returning the previous speed on setting it, sounds like overengineering without any practical use on the other hand :)

It seems to me I gave a pratical use in my comment https://github.com/cc65/cc65/pull/620#issuecomment-376246179, didn't I ?

Anyhow, let me propose an alternative in my next comment...

> actually, it could be argued, that any cc65 compiled program would simply call fast() at startup. because really, why not?

Maybe I'm too naive to see the irony but to me there's really no reason not to do so.

oliverschmidt:
Okay, regarding returning the previous setting...

The role model is from my perspective videomode(). I didn't decide to have it return the previous mode - Uz did that. Anyhow, I can see it making sense here.

Then we have getcpuspeed()/setcpuspeed(). Given that there's an explicit get... function there's no strict reason to have the set... function return the previous speed. I think there are nevertheless two reasons why it still might make sense:

* If it is presumed that usually the previous speed is necessary at least once then one can save a function call.
* If it is presumed that read access to the speed setting and write access to the speed setting share code then avoiding to link the get... function can save more than just a function call.

Given that setcpuspeed() is supposed to not be able to fail one can say: "Instead of a void function it can as well return the previous speed." However, I personally don't think that these are strong arguments so if there's a momentum here to say setcpuspeed should just be void I'm personally fine with that too.

So last but not least we have getacceleratorspeed()/setacceleratorspeed(). I think in general the same (rather weak arguments) for returning the previous speed apply here too. However, the set... function may very well fail, so that alone might be a good reason to drop returning the previous speed but rather return zero for successful operation and some implementation defined non-zero values(s) for failure(s). A cross target program is supposed to only test for zero. If there's a momentum to favor this alternative I'm totally open.

mrdudz:
> Maybe I'm too naive to see the irony but to me there's really no reason not to do so.

- overhead. eg on the c64 there are perhaps 5 or so accelerators which all work slightly different and need special care.
- on the c128 when the CPU is in fast (2mhz) mode, the VICII will display garbage. so you can either use it for 80 columns programs, or you can enable 2mhz only in the borders (needs some raster interrupt)
- IEC I/O will potentially no more work when the accelerator is enabled. however in a plain old program file i/o should work, out of the box :)

that said, it'd be interesting how much overhead those generic set/get functions would create.

also, IMHO, there should still be extra functions per accelerator to set up the options. this should be almost zero overhead, since its all about setting up a register most of the time.

blackystardust:
I'm watching the comments roll by, never figured this would cause such a stir.

My preference still goes to what I've detailed before, native cpu functions, specific per accelerator functions, and generic accelerator functions that can recog which is present.

I'm not familiar with the hierarchy of this project, so, to know which road to take, who decides which direction this needs to go in?

polluks:
> Maybe I'm too naive to see the irony but to me there's really no reason not to do so.

**fast+vic=blind** you maybe also check this behavior with your good ol' ZX-81

oliverschmidt:
> I'm watching the comments roll by, never figured this would cause such a stir.

Just grab some popcorn and enjoy ;-)

> I'm not familiar with the hierarchy of this project, so, to know which road to take, who decides which direction this needs to go in?

I see your point. I just hope you're not in a hurry. There's no formalized decision-making process (at least none that I'm aware of). I tend to believe I have sort of a right of veto. Apart from that I'm naive enough to imagine usually the best arguments win...

> My preference still goes to what I've detailed before, native cpu functions, specific per accelerator functions, and generic accelerator functions that can recog which is present.

Do you see the generic accelerator functions internally actually call the per accelerator functions? Or share the code in some other way? Do you see a way to somehow encapsulate the per accelerator code although the generic functions have to recog the accelerators? Or do the generic functions just call each  per accelerator functions in turn until one succeeds?

oliverschmidt:
> overhead. eg on the c64 there are perhaps 5 or so accelerators which all work slightly different and need special care.

Likely a misunderstanding: You were mentioning fast() so I was under the impression we were only talking about setcpuspeed() and not about setacceleratorspeed().

> on the c128 when the CPU is in fast (2mhz) mode, the VICII will display garbage. so you can either use it for 80 columns programs, or you can enable 2mhz only in the borders (needs some raster interrupt)

So would it from you perspective desirable to couple the C128 implementation of setcpuspeed() with videomode() so that only "the right thing" happens?

> IEC I/O will potentially no more work when the accelerator is enabled. however in a plain old program file i/o should work, out of the box :)

Unfortunately I have no idea what you refer to. Does "plain old program file i/o" refer to using the ROM functions? Or to rephrase my question: Does the C64 C Library offer functions beyond "plain old program file i/o" ?

mrdudz:
> So would it from you perspective desirable to couple the C128 implementation of setcpuspeed() with videomode() so that only "the right thing" happens?

well. no. what IS the right thing? :)

> Does the C64 C Library offer functions beyond "plain old program file i/o" ?

no, thats the point. regular stdlib i/o should work by default - and that means you cant enable (some) accelerators by default.

silverdr:
@mrdudz "actually, it could be argued, that any cc65 compiled program would simply call fast() at startup. because really, why not?" - AFAIR on some targets that might imply blanking the screen. Also, some parts might need to be "cycle exact" and "cycled" on the regular CPU. Not that I am against or something. Just it would be good to have that in mind.

oliverschmidt:
>> So would it from you perspective desirable to couple the C128 implementation of setcpuspeed() with videomode() so that only "the right thing" happens?

> well. no. what IS the right thing? :)

From your description this seems easy to me: When in VIC mode stay at 1MHz even if the user asks for 2MHz, when in VDC mode let the user have its way.

> [...] regular stdlib i/o should work by default - and that means you cant enable (some) accelerators by default.

Sorry, but now you lost me completely. I can't make any sense of the statement above :-(

polluks:
> From your description this seems easy to me: When in VIC mode stay at 1MHz even if the user asks for 2MHz, when in VDC mode let the user have its way.

Did you read my plasma comment? Sometimes it's smart to boost VIC apps.

blackystardust:
> Do you see the generic accelerator functions internally actually call the per accelerator functions? Or share the code in some other way? Do you see a way to somehow encapsulate the per accelerator code although the generic functions have to recog the accelerators? Or do the generic functions just call each per accelerator functions in turn until one succeeds?

Code sharing between get and set versions of a function is possible, as for the generic one, to keep things modular and easily extendable it would call a is_<turbocartname>_present function which returns 1 if it is present and then uses that cart specific set/get functions.

Marrin:
>> [...] regular stdlib i/o should work by default - and that means you cant enable (some) accelerators by default.
>
> Sorry, but now you lost me completely. I can't make any sense of the statement above :-(

On some targets I/O to drives, printers and so on only works with the slow/normal CPU speed, so you can't use stdlib I/O on those devices when in fast mode.  The C64's IEC bus protocol is one example.

Marrin:
BTW I think it would make sense to offer `fast()`/`slow()` on the C64 with the C128 implementation.  On a C64 this has no effect on the CPU speed, but on the C128 in C64 mode it has. CPU intensive C64 applications, like compression programs, are doing this to take advantage of the higher CPU speed when run on a C128.

groessler:
Oops. I hadn't expected to stir up so much discussion with my "function name" comment (setspeed/getspeed). :-)

May I suggest that we don't overengineer the whole thing? My proposal would be that we provide "setspeed" and "getspeed" functions for the systems which support that out-of-the-box (C128, and maybe? C16/Plus4). Which only support the built-in speeds (no 3rd party accelerators).

I think "setspeed" should return the old speed setting. Why not?

Why "setcpuspeed/getcpuspeed"? Which other speeds could potentially be changed?

For support for 3rd party accelerators, I think those could be supported by target dependent functions.

oliverschmidt:
Hmmmmm, it seems that the only thing we all agree upon is, that the topic seems to be broader/wider than expected ;-)

I must admit that I don't even really know where to start responding to the comments. So let me take one step back and state what the C library is about in the first place (at least from my pov). I see primarily three reasons to add some functionality to the C library:

1. It's part of some "official spec". Having e.g. open/read/write/close allows to share (existing) C code between cc65 and other toolchains.

2. It allows for cross target programs. Having e.g. the various driver interfaces or the device functions allows to share code between several cc65 targets.

3. It significantly simplifies usage of the feature from the perspective of the C programmer. Having e.g. the textframe function on the Apple II allows to simply draw rectangles on the text screen.

When this thread started I thought there would be a potential for reason 2.) but by now I understand that I was naive. Let me briefly describe how I came to be that naive:

The most advanced Apple II machine is the Apple IIgs. Its CPU can be operated at 2.7 MHz or at 1 MHz (which is the standard Apple II operation speed). So one could implement fast()/slow() for the Apple IIgs. However, on the IIgs the 2.7 MHz mode is compatible with about everything imaginable. Because it is automatically temporarily turned off for everything incompatible. So the only reason to turn the 2.7 MHz off is that one just doesn't want the speedup. This is usually the case for games as Apple II games almost never sync to something but just run at full speed. The same goes for Apple II accelerators. They temporarily slow down where necessary so there are about no compatibility issues - beside unplayable games.

So now I understand that even if we had something like fast()/slow() for the C128 and the Apple IIgs with the same general idea and the same interface it would never allow to be used in the same way in a cross target program because its sematics are way to different :-(( And as far as I understand the same holds true for CBM accelerators vs. Apple II accelerators.

So reason 2.) from my list above doesn't apply. But what about reason 3.)? When I look at the implementations of fast()/slow() present/suggested so far they all just set/clear some bit in some memory mapped register. If I'm not mistaken this corresponds to a single `|=` / `&=` operation in C.

To summarize: I'm unsure about the general value of all those speed setting functions when they just look identical on different targets but have to be used in a totally target specific way! I don't see a value per se in wrapping a memory mapped register manipulation in a C function if using that C function requires intimate knowledge about the register semantics because I don't see the audience for such a function.

Coming from there my opinion is:

* I personally currently don't see much benefit in the fast()/isfast()/slow() functions. However, Marco is willing to contribute them to cc65 - which is of course always nice and welcome! Regarding the actual names I follow Chris' perspective that setspeed()/getspeed() (in contrast to setcpuspeed()/getcpuspeed() or fast()/isfast()/slow()) are the most appealing names. I would be happy if Marco would be willing to adjust this pull request to the setspeed()/getspeed() approach (and adjust the exsisting C128 code). However, if he doesn't want to do that I'm generally willing to merge the pull request as-is too!

* Regarding the accelerator support I come back to my three potential reasons above: Detecting several accelerators and using the one present seems like a complex-enough task to me to justify C functions for reason 3.). Therefore I personally see the point in generic setaccelerator<...>()/getaccelerator<...>() functions. However, I know nothing about CBM accelerators and in this thread it was said that the SuperCPU needs detailed configuration to be used effectively. So I wonder if it is possible at all to use CBM accelerators effectively with a generic interface at all. But I tend to presume that Marco knows more than enough about this and that he wouldn't propose a generic interface if it wouldn't make sense ?!?

* Furthermore it was said in this thread that configuring a CBM accelerator just means setting some register so by analogy with the CPU speed functions I personally currently don't see much benefit in C functions fro setting up a specific CBM accelerator. However, for that reason I asked Marco if there's a code sharing between those accelerator specific functions and the generic accelerator functions. I interpret Macro's reply (thanks, Marco!) as a clear 'yes'. So I see the accelerator specific functions as a side effect / prerequisite for the generic accelerator functions - and from that perspective they are welcome even if I personally don't see their value per se.

I hope that my thoughts make at least some sense and hope they help to flesh out this discussion...

blackystardust:
@oliverschmidt 

The C128 and C16/Plus4 are the easiest to change, some of the others have hidden i/o registers that need to be setup 1st, and/or special opcodes that change the cpu speed, so the methods used are sometimes not that simple.

Implementing the functions in a multi-level way (modular), where there are cpu/accelerator specific functions and a generic function that uses those specific functions is to keep the code small when you are just coding for either the native cpu or a specific accelerator. Ofcourse when you don't care which cpu is present and just want generic turbo/slow speed switch that works with any accelerator (if present) then you get some overhead because of all the modules being present and extra detection code.

IMHO this will allow for:
- native cpu coding with the same overhead as now
- specific cpu coding with small code size
- generic cpu coding with a larger overhead

mrdudz:
my proposal would be like this:

* have ACCELERATOR_fast()/ACCELERATOR_slow() that is defined for each ACCELERATOR (including non accelerated) as a non-portable function which does no more no less than switching between "slow" and "fast" CPU speed. for accelerators that allow setting more than these speeds it would switch between "non-accelerated" and "maximum" speed. likely most if not all of these functions can be implemented as simple macros that just write one register, so its really zero overhead.
* have ACCELERATOR_option(value) function(s) to set specific options. again, most can be macros, so no overhead
* the above goes, per accelerator, into seperate ACCELERATOR.h or .inc respectively. we can put all sorts of related/specific stuff in there without invading the rest of the libs :)

* have accelerator_detect() as a standalone helper function. i wouldnt mind if it ended up in that initcode area which is later reused, ie we call it from crt0 before main and just keep a variable that contains the detected accelerators ID. again, (almost) zero overhead that way

now a program that wants to support accelerators would do sth like this:

   switch (get_global_accelerator_id()) {
      case ACCELERATOR_SCPU: scpu_fast(); scpu_optbits(SCPU_0400_WRITEBACK); break;
      case ACCELERATOR_FLASH8: flash8_fast(); break;
   }

....which infact awfully looks like something that could be provided in a rather compact library function written in asm :) so _also_:

* have accelerator_fast(unsigned char id)/accelerator_slow(unsigned char id)
* the above two go into a generic accelerator.h and will infact provide a portable interface. however it needs to be documented precisely. it will, in a portable way, allow to switch the CPU/accelerator speed. however, unexpected side effects may occur and some precautions must be made in a program that wants to be truly portable across cc65 target. however doing this with a handful ifdefs in the program itself is a better idea than cluttering the runtime with it IMHO.

...this kinda matches what blacky was going to do, if i understood it right :=)

oliverschmidt:
> The C128 and C16/Plus4 are the easiest to change, some of the others have hidden i/o registers that need to be setup 1st, and/or special opcodes that change the cpu speed, so the methods used are sometimes not that simple.

I see.

oliverschmidt:
> have accelerator_detect() as a standalone helper function. i wouldnt mind if it ended up in that initcode area which is later reused, ie we call it from crt0 before main and just keep a variable that contains the detected accelerators ID. again, (almost) zero overhead that way

While I'm not sure that I understnad most of the comment I'm pretty sure that this is a great idea :-)

@blackystardust: You maybe aren't aware of what @mrdudz is referring to here: A simple example is https://github.com/cc65/cc65/blob/master/libsrc/apple2/get_ostype.s - all content of the segment "ONCE" is only available during startup and is later overwritten by other segments. In order to save some value computed during startup for later usage during runtime that value is to be placed in the segment "INIT". See https://github.com/cc65/wiki/wiki/Segment-usage-of-constructors for some more info.

blackystardust:
@oliverschmidt 

I see, yes, that could be used to check the current cpu, and put it in a var, that will cut down on the overhead for detection, and make the generic accelerator set/get functions pretty small.

greg-king5:
> case ACCELERATOR_SCPU: scpu_fast(); scpu_optbits(SCPU_0400_WRITEBACK); break;

A program that does graphics shouldn't need to know how to change that RAM mirroring -- write-back shouldn't be limited by a _general_ speed function.  `setspeed()` should do only `scpu_fast()`; it should change only the CPU's clock rate.

It might be OK for `setaccelspeed()` to do more -- especially if it can tell the program which accelerator it found.

blackystardust:
@oliverschmidt 

From what I understand my original plan still stands, expect for the detection code, it could be in one of those 'once' sections.

So if everyone agrees I'll start making the accelerator specific functions, and once those have been reviewed and merged I'll do the 'generic/don't care which accelerator is present' functions.

The plus4 fast/slow/isfast functions of this pull request still stand on their own merit and are not part of the accelerator functions, so I'm mainly waiting for this pull request to get merged.

oliverschmidt:
> The plus4 fast/slow/isfast functions of this pull request still stand on their own merit and are not part of the accelerator functions, so I'm mainly waiting for this pull request to get merged.

I see. So I'm going to merge this now, primarily because:

* Without this lengthy discussion it generally qualifies for a quick merge in the first place:

  * It "only" adds something so it can't introduce regressions.

  * It isn't so complex that it adds additional  maintenance burden - in case the contributor isn't interested in potentially necessary maintenance.

* With this discussion because I'm simply not able to assess all arguments given, sorry :-( But the code added with this pull request is so small that everybody who feels strongly uncomfortable with it can easily send another pull request suggesting a modification.

* Greg already agreed to contribute the movement of declarations into another header file he suggested.

I thank everybody for contributing thoughts to this discussion and hope that one can follow my arguments for merging as-is now.

PR #621, opened 2018-04-01
Fixed cpu detection.

blackystardust:
Fixed the CPU detection code; `bne` after the decimal mode check should have been a `beq`.

### DIFF
--- a/libsrc/common/getcpu.s
+++ b/libsrc/common/getcpu.s
@@ -53,7 +53,7 @@ _getcpu:
         adc     #1              ; $01+$09 = $10 on 6502, $01+$09 = $0A on 2a03/2a07
         cld
         cmp     #10
-        bne     @L5
+        beq     @L5
         lda     #0              ; CPU_6502 constant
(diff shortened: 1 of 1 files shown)


libsrc/common/getcpu.s
```diff
@@ -53,7 +53,7 @@ _getcpu:
         adc     #1              ; $01+$09 = $10 on 6502, $01+$09 = $0A on 2a03/2a07
         cld
         cmp     #10
```
greg-king5: I am so used to seeing hexadecimal operands in ca65 source files that I was convinced that I see `$10` on that line.  I even had typed a comment (that claimed that the old code is correct) before I realized that a dollar-sign _isn't_ there.

I recommend that you change that operand to `$0A`, so that no one else will make the same mistake that I (and probably you originally) made.

blackystardust:
@greg-king5 

Good suggestion, done so in my latest commit.

mrdudz:
merged :)

PR #624, opened 2018-04-10
Simpler get_tv

polluks:
You don't have to count the beam, just check the KERNAL.


oliverschmidt:
No SME (Subject Matter Expert) complained so far so...

PR #625, opened 2018-04-12
VIC-20: Add missing VIA register names

ops:
Added missing VIA registers to vic20.inc. Register names from the COMPUTE!'s book 'Mapping the VIC'.

Old definitions kept for compatibility but marked as deprecated.


polluks:
How about sharing with pet.inc, atmos.inc and telestrat.inc?

ops:
pet.inc has only one VIA, atmos.inc and telestrat.inc use struct to define VIA. So, I'm not sure how to share definitions without breaking backward compatibility. Guidance needed. I'm new to cc65 and I haven't touched my VIC since 1985 but I'm now planning to write something for it.

asminc/vic20.inc
```diff
@@ -66,16 +66,43 @@ VIC_COLOR       := $900F        ; Border and background color
 ; ---------------------------------------------------------------------------
 ; I/O: 6522 VIA1
 
-VIA1            := $9110
-VIA1_JOY        := $9111
-VIA1_DDRB       := $9112
-VIA1_DDRA       := $9113
+VIA1            := $9110        ; *** Deprecated ***
```
greg-king5: The `VIA1` and `VIA2` labels shouldn't be deprecated.  They can be used to refer to the entire chip.  
For example, they can be used in a loop to read/write many registers:
```asm
; Initiate the second VIA's registers.
        ldx #$0F - 1
@L1:    lda via2_data,x    
        sta VIA2,x
        dex
        bpl @L1
```

greg-king5:
I don't know what Stefan meant by the word "sharing"; but, I do think that it would be nice if you submitted a second Pull Request with the title, "Pet: Add missing VIA register names" -- do the same thing to `pet.inc` that you did to `vic20.inc`.

asminc/vic20.inc
```diff
@@ -66,16 +66,43 @@ VIC_COLOR       := $900F        ; Border and background color
...
+VIA1_JOY        := $9111        ; *** Deprecated ***
+VIA1_PB         := $9110        ; Port register B
+VIA1_PA1        := $9111        ; Port register A
```
greg-king5: I would like to see the "R" put into those names (`VIA1_PRB`, `VIA1_PRA1`, and `VIA1_PRA2`).  It would be consistent with other register names (such as `VIA1_IFR`) and the port names in other ".inc" files (such as `CIA1_PRB` in "c64.inc").

mrdudz: ugh no, that would be totally unlike what they are commonly called

asminc/vic20.inc
ops: Right, good point.

ops:
I can update pet.inc after this one is finished.

asminc/vic20.inc
mrdudz: yeah, terrible. the other headers should be fixed accordingly

ops: Guys, I'm lost here. Do we want VIA1_PB or VIA1_PRB in vic20.inc?

oliverschmidt: > Guys, I'm lost here. [...]

So at least I'm not the only one ;-)

polluks:
Well, I prefer the MOS datasheet. Maybe atmos.inc is closer to these defines.

asminc/vic20.inc
mrdudz: my point is: naming things differently than the official datasheets do is a bad idea. and tacking an extra R to every define because its a register (doh, what else?) is questionable at best. oh well :)

greg-king5:
> My point is: naming things differently than the official datasheets do is a bad idea. and tacking an extra "R" to every define because it's a register (doh, what else?) is questionable at best. Oh well. :)

(That's like saying, "it's questionable to tack the letters "IA" to the names CIA and VIA".)

I looked at several manufacturers' official datasheets for the PIA (6520), the VIA (6522), the TIA/TPI (6523/6525), and the CIA (6526).  Where they use labels like `PB`, they refer to the entire peripheral port, which includes a data register, a direction register, a buffer, data lines, and control lines.  Where the datasheets refer to only the data register, they use labels like `IRB`, `ORB` (Input Register B, Output Register B), and `PRB` (Peripheral Register B or Port Register B).

The official references use the letter "R" in their labels; I want `vic20.inc` to do the same thing.

mrdudz:
Am Dienstag, 17. April 2018, 21:43:51 CEST schrieb greg-king5:
> (That's like saying, "it's questionable to tack the letters "IA" to the
> names CIA and VIA".)

dont be silly.

> I looked at several manufacturers' official datasheets for the PIA (6520),
> the VIA (6522), the TIA/TPI (6523/6525), and the CIA (6526).  Where they
> use labels like `PB`, they refer to the entire peripheral port, which
> includes a data register, a direction register, a buffer, data lines, and
> control lines.  
>
> Where the datasheets refer to only the data register, they
> use labels like `IRB`, `ORB` (Input Register B, Output Register B), and
> `PRB` (Peripheral Register B or Port Register B).

in what VIA datasheet did you see PRB?

> The official references use the letter "R" in their labels; I want
> `vic20.inc` to do the same thing.

they probably also use all kinds of other letters in their labels - however 
that doesnt make it less silly to not just use the official names, verbatim. 
everything else is just confusing, for no reason at all. 

and for VIA its PB that is port B. or IRB/ORB if you actually want to refer to 
the register file in the chip (this would be very unusual for a header like 
this however). 

there is not a single instance of PRB in the MOS datasheet for that matter. 
(not in synertek one either apparently, but i dont have a OCRd version of that 
so that might be wrong)

-- 

Information's pretty thin stuff unless mixed with experience. 
<Clarence Day>

polluks:
@oliverschmidt We seem to need a referee...

Marrin:
C64 literature seems to use PRA/PRB for the CIA port registers.  For instance the „C64 Programmer's Reference Guide“, published by Commodore, uses those names in appendix M in the block diagram of the CIA chip (page 421), in the register map (page 428) and also in the following pages in the texts.

Reading that I don't think the R is redundant or unnecessary because the text talks about PA when the data lines/pins are meant and PRA when the register is meant.  Those are different things after all and it makes sense to be able to talk about them in the same sentence without naming confusion.

The text in the book seems to match this data sheet: http://www.datasheet4u.com/pdf-down/6/5/2/6526-Commodore.pdf by Commodore Semiconductors.

And here is a data sheet for the VIA by Rockwell which has the name IRA and ORA for the data registers: https://www.princeton.edu/~mae412/HANDOUTS/Datasheets/6522.pdf

oliverschmidt:
> We seem to need a referee...

I hope that's a joke. As you all know I tend to prefer approaches which experienced CBM asm developers regularly consider the worst case...

greg-king5:
Other sources for 65xx datasheets:
http://www.6502.org/documents/datasheets/
http://www.zimmers.net/anonftp/pub/cbm/documents/chipdata/

The TIA/TPI (6523/6525) datasheets use the names `PRA`, `PRB`, and `PRC`.

I concede that the VIC-20 Programmer's Reference Guide uses the names "Port A" and "Port B", in some places, to refer to the data registers.  But, in other places, those same names seem to refer to more of a parallel I/O system than only those particular registers.

I again state my desire for consistency in the naming style of I/O chip registers -- consistency between the names used for the VIA, and consistency between the names used among all of the various I/O chips.

> And for VIA, it's PB, that is port B, or IRB/ORB if you actually want to refer to the register file in the chip (that would be very unusual for a header like this, however).

In my opinion, `PB` is a bad substitute for `IRB/ORB`.  A better substitute keeps the letters that are the same in those two names.  Only the "I" and the "O" should be replaced (by "P").

Everything, that I see, points to my recomendation (`VIA1_PRB`, and so on).

mrdudz:
we'll first have to make sure what we want. 
what i want is: stick to the datasheet, literally, no exceptions, no own interpretations, corrections, recommandations, whatsoever. that makes it as easy as possible for everyone to look up those names. (and whats written in CIA or TPI or FBI datasheets is irrelevant to me when it comes to naming the VIA labels. actually it tells quite a bit that these have to be mentioned here)
gregg wants to retroactively make things more consistant, more logical, i dont know. IMHO this only creates confusion because the labels will not match whats in the datasheet.

my suggestion would be: merge this as is. PRB vs PB doesnt matter much right now.

then gregg makes a seperate pull request where this can be discussed seperately. and without mentioning unrelated datasheets perhaps :)

oliverschmidt:
@mrdudz Thanks for trying to bring a solution to this discussion.

> my suggestion would be: merge this as is. PRB vs PB doesnt matter much right now.
> then gregg makes a seperate pull request where this can be discussed seperately.

I like this approach a lot - for a reason that maybe not everybody sees: I think this thread is already by now a pretty strange experience for the author of this pull request - and first-time contributor to cc65.

PR #636, opened 2018-04-30
 Added C64 Turbo Master accelerator code and documentation. 

blackystardust:
Added C64 Turbo Master accelerator code and documentation.

### DIFF
--- a/doc/funcref.sgml
+++ b/doc/funcref.sgml
@@ -73,16 +73,19 @@ function.
 <item><ref id="detect_c65" name="detect_c65">
 <item><ref id="detect_chameleon" name="detect_chameleon">
 <item><ref id="detect_scpu" name="detect_scpu">
+<item><ref id="detect_turbomaster" name="detect_turbomaster">
 <item><ref id="get_c128_speed" name="get_c128_speed">
 <item><ref id="get_c64dtv_speed" name="get_c64dtv_speed">
 <item><ref id="get_c65_speed" name="get_c65_speed">
 <item><ref id="get_chameleon_speed" name="get_chameleon_speed">
 <item><ref id="get_scpu_speed" name="get_scpu_speed">
+<item><ref id="get_turbomaster_speed" name="get_turbomaster_speed">
 <item><ref id="set_c128_speed" name="set_c128_speed">
 <item><ref id="set_c64dtv_speed" name="set_c64dtv_speed">
 <item><ref id="set_c65_speed" name="set_c65_speed">
 <item><ref id="set_chameleon_speed" name="set_chameleon_speed">
 <item><ref id="set_scpu_speed" name="set_scpu_speed">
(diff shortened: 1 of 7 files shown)


testcode/lib/accelerator/Makefile
```diff
@@ -24,3 +24,6 @@ chameleon-test.prg: chameleon-test.c
 
 c65-test.prg: c65-test.c
 	$(CL) -t c64 c65-test.c -o c65-test.prg
+
+turbomaster-test.prg: turbomaster-test.c
+	$(CL) -t c64 turbomaster-test.c -o turbomaster-test.prg
```
greg-king5: `turbomaster-test.c` is missing.

libsrc/c64/acc_detect_turbomaster.s
```diff
@@ -0,0 +1,47 @@
...
+        .byte   $6B                    ; NOP on 65(S)C02 and TZA on 65CE02 and 4510
+        cmp     #$A3
+        beq     not_found
+
+; Check for turbo master basic replacement
+        ldy     TURBOMASTER_DETECT
+        cpy     #$A2
+        bne     not_found
+found:
+        lda     #$01
```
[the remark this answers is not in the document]
blackystardust: Well, in the line

.byte   $E2,$01

which is a

NOP #$01 on 65(S)C02

yes, it would be a $01 because of the INA, however, on a 65CE02/4510 it would be a

LDA $(01,S),Y

which depends on the stack, and the Y register, and then whatever is at that location is loaded into the A register, which most likely is not $01

doc/funcref.sgml
```diff
@@ -6320,6 +6364,28 @@ clean-up when exitting the program.
...
+<tag/Availability/cc65 (not all platforms)
+<tag/See also/
+<ref id="detect_turbomaster" name="detect_turbomaster">,
+<ref id="get_turbomaster_speed" name="get_turbomaster_speed">,
+<tag/Example/None.
+</descrip>
+</quote>
+
+
 <sect1>set_scpu_speed<label id="set_scpu_speed"><p>
```
greg-king5: The "set..." functions, too, have the same sort order problem.

libsrc/c64/acc_detect_turbomaster.s
```diff
@@ -0,0 +1,45 @@
...
+        .export         _detect_turbomaster
+
+        .include        "accelerator.inc"
+
+_detect_turbomaster:
+        lda     #$00
+        tax
+
+; Make sure the current CPU is not a 6510
+        .byte   $1A                   ; NOP on 8502, INA on 65(S)C(E)02, 4510 and 65816
```
oliverschmidt: No prerequisite for a merge but maybe you want to harmonize the the comment-start-column. It differs here from the ones below.

PR #643, opened 2018-05-13
The opcode for BS should be 0x0C.

mikeakohn:
For the SWEET16 instruction set, BS in src/ca65/instr.c was 0x0B when this should be 0x0C.


oliverschmidt:
@mikeakohn: Thanks for contributing to cc65 :-)

mikeakohn:
Glad to help :).

PR #645, opened 2018-05-14
Update c64.sgml

polluks:
soft80 is NMI ready

### DIFF
--- a/doc/c64.sgml
+++ b/doc/c64.sgml
@@ -5,7 +5,7 @@
 <title>Commodore 64-specific information for cc65
 <author><url url="mailto:uz@cc65.org" name="Ullrich von Bassewitz"><newline>
 <url url="mailto:greg.king5@verizon.net" name="Greg King">
-<date>2017-01-18
+<date>2018-05-14
 
 <abstract>
 An overview over the C64 runtime system as it is implemented for the cc65 C
@@ -122,8 +122,8 @@ since the program must be loaded to the BASIC start address.
 <sect1>80 Columns conio driver<p>
 
 The C64 package comes with an alternative software driven 80 columns
-module <tt/c64-soft80.o/ which uses the memory under I/O between &dollar;d000
-and &dollar;ffff.
+module <tt/c64-soft80.o/ which uses the memory under I/O between &dollar;D000
+and &dollar;FF3F.
 
 In memory constrained situations the memory from &dollar;400 to &dollar;7FF
 can be made available to a program by calling <tt/_heapadd ((void *) 0x0400, 0x0400);/


oliverschmidt:
So just to make sure I get this right: The soft80 code never actually used $FF40-$FFFF. The recent NMI issue now drives you to make this explicitly clear in the documentation.

polluks:
Exactly. Next question: How do we trigger the NMI catcher? At least it should be shared by the linker.

oliverschmidt:
> Next question: How do we trigger the NMI catcher? At least it should be shared by the linker.

Can you please elaborate? To help you see where to start here's what I know:

- As the 6502 NMI (nomen est omen) doesn't react on SEI/CLI there are only two options to handle memory banking in the $FFF0 area:
  - Temporarily disable all NMI sources on the "chip level".
  - Make sure there's a valid NMI vector in all memory banks.

- Video memory is only/mostly written too. ROM is only read from. Therefore it was a nice idea of the C64 designers to allow video memory and ROM to share address space.

I guess to have things like tgi_getpixel() working the C64 TGI driver temporarily banks in the video memory not write-only but read/write. This breaks NMIs triggered during that time.

I guess the "NMI catcher" is some code the NMI vector in the video memory bank points to. It likely switches back to ROM, simulates an NMI and when the ROM NMI handler returns switches back to the video memory bank and returns.

polluks:
1. The restore key is directly connected to the CPU's NMI pin, you cannot disable it. For example run hello.c with soft80 and press restore - bad luck.
2. c64-ram.emd uses the last page, press restore - bad luck.
You see, TGI was just the obvious case.

Imagine this worst case scenario: Many years ago an Intel engineer wrote a note "Branch prediction may disable system security" and nothing happened ;-)

mrdudz:
actually you CAN "disable" the Restore key with a little trick... program CIA2 so it triggers an NMI, then in the NMI handler do NOT acknowledge the NMI. since NMIs are edge triggered, no more NMIs will occur (until the CIA2 NMI was acknowledged)

however, that is not a good generic solution for cc65 libs i think. instead, code that would bank out the kernal should install trampoline code for the NMI like the one gregg posted in the rs232 issue. that code (and the code that installs it) code perhaps be in a seperate object that is pulled in by the respective modules that need it.

as for c64-ram.emd - perhaps it should simply not use the last page at all. afterall, using it to buffer text that comes from rs232 is one of the few useful applications, and it should work with that :)

oliverschmidt:
> code that would bank out the kernal should install trampoline code for the NMI like the one gregg posted in the rs232 issue.

Given that RESTORE is an omnipresent NMI source that seems so obvious that I don't understand why this isn't in place since many years.

> that code (and the code that installs it) could perhaps be in a seperate object that is pulled in by the respective modules that need it.

As far as I understand the from my pov over-engineered cc65 driver architecture hits us here again :-(

It seems natural to use a cc65 CONSTRUCTOR for the trampoline installation and just pull it in where necessary. But that means that **_all_** TGI drivers and **_all_** EMD drivers will pull it in. For TGI that's no actual issue as there's only one C64 TGI driver anyway. But for EMD it's not nice as only 1 out of 10 drivers needs it.

So maybe it would be appropriate to put a copy of the NMI trampoline right into c64-ram.emd - instead of making all EMD drivers pull in the "shared" NMI trampoline. Currently 64-hi.tgi and c64-ram.emd can be used sequentially in the same program - at least as far as I understand. If c64-ram.emd comes with its own NMI trampoline it should make sure to restore the previous setting to allow the shared NMI trampoline to work again - or detect the presence of the shared NMI trampoline and skip installing its copy.

Another viable option might be to "just" document that c64-ram.emd needs an NMI trampoline if NMI support is desired. And provide the user with the necessary linker command line option to pull it in - similiar to __EXEHDR__ in http://cc65.github.io/doc/c64.html#ss4.2. This option would have the benefit of using the shared NMI trampoline with its installation code being In the ONCE segment and thus being reused.

oliverschmidt:
> i think C64 users are simply used to "dont touch RESTORE, it will crash" - because that is valid for oh so many programs (probably most of them =P).

I see ;-) How does a "normal" (aka non-banking) cc65 program react on pressing RESTORE ?

polluks:
It's getting interesting.
Maybe we lost 10% of all cc65 users, "this software is unreliable", because they hit restore accidentally (remember it's above return)...
Indeed, constructors are inefficient what about foo_init and a library reference?

The idea of our dynamic drivers is to exchange a driver without recompile. Of course a driver does not know if another driver has already has a trampoline, so everyone needs its own :(
Maybe we can avoid this overhead while statically compiling?

"Over-engineered"? Which driver system do you think is better: ProDOS or SOS? :)

mrdudz:
> How does a "normal" (aka non-banking) cc65 program react on pressing RESTORE ?

it is only a problem when the kernal is banked out - when its banked in the kernal contains an NMI handler (which eg checks for run/stop, and performs BASIC warmstart when it is pressed). so in general its not an issue, because kernal is banked in.

oliverschmidt:
> so in general its not an issue, because kernal is banked in.

That's what I presumed. I just wanted to learn what happens.

> checks for run/stop, and performs BASIC warmstart when it is pressed

I see. Thanks.

polluks:
@oliverschmidt it's like Apple's Ctrl + Reset

polluks:
ordinary program + soft80 = bad luck (well, we never made a public opinion poll, I suppose many users simply use ca65 and ignore all drivers...)
I thought a constructor is used by every driver but a library reference would be more specific.
TGI: replace c64 by atari :)

mrdudz:
> replace c64 by atari

thats just mean

polluks:
Killjoy! :)
At the beginning Uz made a TGI "feature table" (for example if someone wants best colours), it was dropped. This is already stripped down!

oliverschmidt:
Okay, as we already got OT that far...

The raison d'être for cc65 driver is to support optional hardware available in multiple incarnations. However, when we look at the currently existing TGI drivers there's afaik only a single one that requires hardware which isn't always available: The C128 640x480 driver.

So put it in other words: The TGI drivers offer 'dynamic polymorphism' (incl. its overhead) while what we actually need/want is 'static polymorphism'. Either there's only one TGI driver for the current target in the first place or the program selects a TGI driver on preprocessor level. This selection may very well work based on features like 'high resolution', 'high number of colors' or 'page flipping support' allowing for nice cross-target code. Things like tgi_getmaxx(void) can then be implemented as macros expanding to constants allowing the cc65 optimizer to create more efficient code.

On the C128 there can be two drivers. One that tries 640x480 and falls back to 640x200 if 640x480 aren't possible. tgi_getmaxy(void) is a real function for that driver. There's likely a lot of code that be shared for both modes in that driver. The second driver is fixed to 640x200.

For all other driver classes (EMD, SER, MOU, JOY) there are targets with only one driver for a certain class.  For target-class tuples with several drivers the programmer can choose if he wants the static polymorphism and select a driver at compile time. Or if he wants dynamic polymorphism (similar to what is available today). However, for dynamic polymorphism there's an "auto select" feature missing today: It tries to initialize the available drivers in a predefined order until one initialization works. The predefined order is for SER drivers starting with the highest maximal port speed. For EMD drivers it's the largest number of pages.

This is how a driver architecture appropriate for cc65 looks like from my pov.

oliverschmidt:
> At the beginning Uz made a TGI "feature table" (for example if someone wants best colours)

I was already with cc65 when that thingy was still there...

groessler:
I'm chiming in rather late, and, have to admit, just cursory followed 
the thread.

I think the current "loadable drivers" implementation is quite good. For 
example (see testcode/lib/mouse-test.c),
one can write a program which loads a selected driver at runtime (needs 
disk I/O), or a program which has all
available drivers compiled in, still selectable at runtime, or have a 
singe driver compiled in.

The driver architecture has some overhead, agreed, but still not bad for 
an 8-bit machine.

The NMI problem: Couldn't there be "disable_rom()" and "enable_rom()" 
(or however you want to call them) runtime
functions where "disable_rom()" disables the ROM and _at_the_same_time_ 
installs an NMI vector to a trampoline which
temporarily enables the ROM and calls the handler? Might need a counter 
if disable_rom() is called while disable_rom()
is active.

Just food for thought...

regards,
chris

oliverschmidt:
Hm, I don‘t see how that approach solves any of the problems mentioned so
far. And apart from that it leads to the largest memory requirement of all
solutions discussed so far. I‘m wondering why nobody (wants to) see that
aspect...

mrdudz:
i dont see a huge memory overhead there... 20 bytes perhaps? so be it :)

groessler:
Doesn't _any_ program which disables the ROM have to deal with the NMI issue?
Probably normally it's not dealt with and causes crashes when the user presses RESTORE at the wrong time.
So it seems that providing an NMI trampoline is always a requirement when disabling the ROM (within cc65, there are asm programs which copy the ROM to the underlying RAM and only then disable the ROM).

mrdudz:
> Doesn't any program which disables the ROM have to deal with the NMI issue?

yes.

oliverschmidt:
Still I don‘t see the benefit of dynamically installing/uninstalling the
NMI trampoline over installing it once on program initialization - while
the later saves some memory and some cycles during runtime.

mrdudz:
it wouldnt have to be "dynamically" - just the first time disable_rom() is called, it will install the trampoline. that way you have a single point where its done, including the case when a user wants to bank out the rom himself (for whatever reason).
the drawback i see there is - it would have to do the same for IRQ - and i dont know if that makes sense, but perhaps it does.

oliverschmidt:
And how is that better than a constructor?

mrdudz:
it saves a few bytes (like, 3) =P

oliverschmidt:
In the binary file or in memory? I guess we agree that the later is more
relevant - and there the constructor allows to reuse the memory of the
trampoline installer. I still can‘t how it gets better than that !?!

mrdudz:
sometimes i wish we had some IRC channel for this kind of stuff :)

groessler:
Where woud you put the constructor?

oliverschmidt:
As suggested above in the „normal“ C library. To be referred to by either a
driver kernel library reference or by a command line option.

groessler:
On 05/15/18 23:47, Oliver Schmidt wrote:
> Still I don‘t see the benefit of dynamically installing/uninstalling the
> NMI trampoline over installing it once on program initialization - while
> the later saves some memory and some cycles during runtime.

Uninstalling is a no-op. enable_rom() would just enable the ROM (sic!), 
which is just 6 bytes (LDA/ORA/STA zp), and the
installed handler is then defunct (and not needed).

Installing (disable_rom()) is just a little bit more than the above 
sequence (but doing the opposite) and setting the CPU NMI vector to the
trampoline. And, of course, the trampoline itself.

Counters to handle recursive calls will need some more space...

regards,
chris

oliverschmidt:
> constructor in the "ONCE" area is better :)

Okay, so we're on the same page here :-)

>> To be referred to by either a driver kernel library reference or by a command line option.

> or disable_rom() - which then would be a very small function :)

That would be feasible for sure. But does it really make sense to offer that functionality (disable_rom() / enable_rom() ) on C level at all?

mrdudz:
> The C64 emd_libref gets its own file forceimporting that constructor meaning that all EMD drivers pull in the NMI trampoline.

this. at least i don't see a drawback there (except some 20 bytes longer binary, which doesn't matter much)

oliverschmidt:
> except some 20 bytes longer binary, which doesn't matter much

Only the NMI trampoline installer gets into the ONCE segment. The actual NMI trampoline is placed in the LOWCODE segment using memory during runtime.

mrdudz:
ok, then 5 bytes overhead? or 10? who cares :)

oliverschmidt:
I'm fine with that. So now that everything is clarified we "only" need someone willing/able to actually create it and contribute it ;-)

oliverschmidt:
BTW: If it would feasible/reasonable to access the C128 VDC from C64 mode then this would be a true use-case for TGI dynamic polymorphism: A C64 program presumes that it runs on a C128 in C64 mode and tries to initialize a C64 VDC driver. If that fails it falls back to the default VIC driver.

oliverschmidt:
Yet an OT: On the Apple II the startup code changes the Ctrl+Reset handler to point to exit(). Given that there's a destructor to close all open files for CBM ...
https://github.com/cc65/cc65/blob/709ee6a28b1375222d44590a57e93d732966f8bf/libsrc/cbm/open.s#L12
... I'm wondering if it wouldn't be desirable to have RunStop+Restore call exit() too before doing it's default thingy.

groessler:
> .. I'm wondering if it wouldn't be desirable to have RunStop+Restore call exit() too before doing it's default thingy.

This sounds like a good idea.

oliverschmidt:
>> [...] have RunStop+Restore call exit() [...]

> This sounds like a good idea.

I have created https://github.com/cc65/cc65/issues/650 to follow up on that.

BTW: Does the Atari have some RESET/IRQ/NMI key too? I just noticed that the Atari C Library has a closeallfiles destructor too...

groessler:
> Does the Atari have some RESET/IRQ/NMI key too? I just noticed that the Atari C Library has a closeallfiles destructor too...

It has a "reset" key which is generating an NMI on the (old) 400/800 computers, and a RESET on the newer XL+ machines (that's what I remember).

I think the OS does a warmstart then. Probably destructors won't be called. I'll check if it's possible to hook that vector.

polluks:
[Reset proof Atari](https://atariwiki.org/wiki/Wiki.jsp?page=Zeile%20per%20CIO%20einlesen%20und%20Programme%20resetfest%20machen)

groessler:
Thanks!

PR #646, opened 2018-05-14
Update c64.sgml

polluks:
Do not use last page!


oliverschmidt:
> uhm, shouldnt the driver really make sure that the last page isnt used? =)

See https://github.com/cc65/cc65/pull/647.

PR #647, opened 2018-05-14
Make NMI ready

polluks:
see #645

### DIFF
--- a/libsrc/c64/emd/c64-ram.s
+++ b/libsrc/c64/emd/c64-ram.s
@@ -44,7 +44,7 @@
 ; Constants
 
 BASE    = $D000
-PAGES   = ($10000 - BASE) / 256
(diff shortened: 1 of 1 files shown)


oliverschmidt:
I only notices after merging that as far as I understand http://cc65.github.io/doc/c64.html#ss7.2 needs to be changed from "Supports 48 256 byte pages." to "Supports 47 256 byte pages.". If that is correct it would be nice if you could send another pull request for the adjustment.

oliverschmidt:
I now saw https://github.com/cc65/cc65/pull/646.

polluks:
Fine, I had to use the single file webinterface because of [a github change](https://morph.zone/modules/newbb_plus/viewtopic.php?forum=12&topic_id=12286&post_id=135888&viewmode=flat&sortorder=0&showonepost=1).

PR #652, opened 2018-05-20
Added Popptr1 to save some bytes and ease parameter extraction

IrgendwerA8:
Further savings are possible by exchange of assignments to ptr-variables  like libsrc/cbm/rwcommon.s "ptr1<->ptr2"  (but here you have to take care of dependent files).


IrgendwerA8:
Just noticed that there is may an issue with "strpbrk.s" (use of X-register). This leads to the question what happened to the function anyway: The description in the "funcref.sgml" is commented, there is no prototype to include and even the fastcall convention isn't applied.
Time to fix or remove?

IrgendwerA8:
Revert change on "strpbrk.s". Together with documentation update, calling convention change and prototype introduction this is a nice extra story.

oliverschmidt:
Nice :-))

> Together with documentation update, calling convention change and prototype introduction this is a nice extra story.

Full ACK.

libsrc/geos-common/drivers/fio_module.s
```diff
@@ -13,7 +13,7 @@
 FILEDES         = 3             ; first free to use file descriptor
 
             .importzp ptr1, ptr2, ptr3, tmp1
-            .import addysp, popax
+            .import addysp, popax, poptr1
```
greg-king5: typo: popptr1

IrgendwerA8:
Sh... don't mess around with foreign targets (wonder why there wasn't a compile problem).
Do you mind if I would fix this in my "Saved 6 bytes by using popptr1" pull request?

oliverschmidt:
> I have to hold back myself to mess around with foreign targets.

You might have a point here after all ;-)) ... see  https://github.com/cc65/cc65/commit/e949a3e64283e65b00ba8311bc9cec7fcb6ff681

PR #654, opened 2018-05-21
Add an util script that prints static-able symbols

clbr:
Sample output:
```
Unique symbols in gas.o:
        _gas_y
        _gas_dir
        _gas_make_explode
        _gas_speed
        _gas_x
Unique symbols in enemy.o:
        _enemy_cell_x
        _enemy_cell_y
```

I've long used a similar script for ELF. It comes in handy when finding unused symbols.

### DIFF
--- a/util/findstatic.awk
+++ b/util/findstatic.awk
@@ -1,0 +1,49 @@
+#!/usr/bin/awk -f
+#
+#	List all variables/functions that can be made static.
+#	findstatic.awk *.o
+#
+#	(C) Lauri Kasanen, under the cc65 license
+#
+
+BEGIN {
+	delete ARGV[0]
+	importbase = "od65 --dump-imports "
+	exportbase = "od65 --dump-exports "
+
+	for (i in ARGV) {
+		importcmd = importbase ARGV[i] " "
+		exportcmd = exportbase ARGV[i] " "
+
+		while ((importcmd | getline) > 0) {
+			if (! /Name:/) continue
+			sub(/Name:/, "")
+			gsub(/\"/, "")
+
+			imports[$1] = 1
+		}
+
+		while ((exportcmd | getline) > 0) {
+			if (! /Name:/) continue
+			sub(/Name:/, "")
+			gsub(/\"/, "")
+
+			exports[ARGV[i]][$1] = 1
+		}
+	}
+
+	for (file in exports) {
+		printed = 0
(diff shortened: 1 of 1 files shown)


util/findstatic.awk
```diff
@@ -0,0 +1,49 @@
...
+
+	for (file in exports) {
+		printed = 0
+		for (sym in exports[file]) {
+			if (sym in imports)
+				continue
+
+			if (!printed) {
+				printed = 1
+				print "Unique symbols in " file ":"
```
greg-king5: I think that `"Unused exports in "` is a better description than `"Unique symbols in "`.  (The word "unique" would make me infer that only one file exported the name.)

clbr:
The "unique symbols" wording matches the ELF script, findstatic.pl by Andrew Tridgell. If two files exported the same symbol, it would fail linking, no?

oliverschmidt:
To be honest I'm not keen to have any scripts in this repo. https://github.com/cc65/cc65/blob/master/util/ca65html is sort of an exception as it has been "already" there.

If you look e.g. at https://github.com/cc65/cc65/tree/master/test there's a mix of GNU Make logic (e.g. https://github.com/cc65/cc65/blob/master/test/ref/Makefile) and C code (https://github.com/cc65/cc65/blob/master/test/bdiff.c). These are the two tools I see used for utils.

If there's no strong 3rd party momentum urging me to merge this pull request I'll rather not merge it.

groessler:
I personally wouldn't have a problem if this script gets added, but I'm running a "read" OS :-).
OTOH, Windows users don't have awk readily available.

Maybe it can be put on the contrib page?

oliverschmidt:
I don't see a strong momentum for that script.

From my POV a Wiki page could be added with some explanation and the actual script source code. That Wiki page could be linked e.g. at https://github.com/cc65/wiki/wiki#tips-and-tricks

@clbr: If you are interested in that option then just let me know and I'll provide you with write access to the cc65 Wiki and you can create/edit that page (and others).

clbr:
I'm fine with it being on the wiki, but not sure I want write access there...

oliverschmidt:
> I'm fine with it being on the wiki

:-)

> but not sure I want write access there...

AFAIK there's no pull request mechanism for Wiki pages. However, there's a version control system so I don't see a significant risk to break things badly unintentionally.

PR #657, opened 2018-05-22
Various speed size optimizations

IrgendwerA8:
Does anybody know why in several places the combination
pha/pla is used instead of tay/tya even the Y-reg is not in use (in the function domain)?
Would this break some kind of optimization?
Or should I just try it out to save this 3 cycles e.g. in "negax" or "leaaxsp"...?

### DIFF
--- a/libsrc/common/ltoa.s
+++ b/libsrc/common/ltoa.s
@@ -6,11 +6,11 @@
 ;
 
         .export         _ltoa, _ultoa
-        .import         popax
+        .import         popax, popptr1, negeax
         .import         __hextab, __longminstr
         .importzp       sreg, ptr1, ptr2, ptr3, tmp1
 
-
+        .macpack        cpu
 
 .code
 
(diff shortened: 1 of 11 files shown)


oliverschmidt:
Travis CI says:

```
val/lib_common_strcat.g.6502.prg
make[1]: *** [../../testwrk/val/lib_common_strcat.g.6502.prg] Error 1
```

When you change target independent code (like here) I suggest that you locally run the regression tests as those offer at least some coverage - as we see above...

IrgendwerA8:
Already inspecting the issue, but yes, would be nice to do so.
Is there a way of running the CI check on the server without applying for a pull request? (As clearly visible now, I have problems running the tests locally...)

oliverschmidt:
> Is there a way of running the CI check on the server without applying for a pull request?

Not that I'm aware of.

> I have problems running the tests locally...

Maybe you want to elaborate and they can be fixed...

IrgendwerA8:
To beef this a bit up here a benchmark running the "dhrystone"-test from here:
http://atariage.com/forums/topic/276141-gcc-6502-vs-cc65/#entry3976080

The improvement is more or less cosmetic, but measurable:

Old:
9.480 sec.
New (due to multiplication savings):
9.440 sec.

Still a long way to go... (gcc-6502: 7.56 sec.)

oliverschmidt:
> Fixed my local environment to run tests now too.

:-)

PR #658, opened 2018-05-22
Indicate a dummy outtext & some clean-up

polluks:


libsrc/apple2/tgi/a2.lo.s
```diff
@@ -58,8 +58,8 @@ Y2      :=      ptr4
         .word   48              ; Y resolution
         .byte   16              ; Number of drawing colors
         .byte   1               ; Number of screens available
-        .byte   8               ; System font X size
-        .byte   8               ; System font Y size
+        .byte   0               ; System font X size
+        .byte   0               ; System font Y size
```
oliverschmidt: As far as I can see the same is true for the C64 / C128 TGI drivers. Please change those too for the sake of consistency.

libsrc/nes/tgi/nes-64-56-2.s
```diff
@@ -242,6 +225,7 @@ SETCOLOR:
 ;
 
 SETVIEWPAGE:
+;       rts
```
oliverschmidt: So far this type of optimization was from my perspective deliberately not done in drivers because the functions are access via a jump table. I can however follow you that it is appropriate to apply the optimization to drivers too. However I'd appreciate if it would be applied to all drivers for the sake of consistency.

libsrc/apple2/tgi/a2.lo.s
polluks: No, it's just not implemented. I was thinking about integrating the Plato routine if the serial code ever works...

oliverschmidt: > No, it's just not implemented.

I can't follow you. Isn't exactly this true for the C64 / C128 TGI drivers too?

> I was thinking about integrating the Plato routine if the serial code ever works...

Now you totally lost me. I thought PLATO is about 512x512 pixels. How  does that relate to this 40x48 pixel TGI ?

oliverschmidt: Ah, now I see (?). For C64 / C128 TGI it's just a bug that they don't have text support, but for Apple II LoRes it's a hardware limitation.

libsrc/nes/tgi/nes-64-56-2.s
oliverschmidt: Is your thumb-up to be interpreted as "Yes, I'll check out the other drivers too and remove those RTSes." ?

polluks: Maybe a misunderstanding, I just want to mark the fall-through. It's not an incomplete implementation like outtext.

oliverschmidt: No, just this one time no misunderstanding ;-) This is totally unrelated to OUTTEXT. I'm simply trying to motivate you to apply the same fall-through to all cc65 drivers. I'm pretty sure that there are quite some places...

oliverschmidt: I see...

PR #659, opened 2018-05-23
Fixed typos and added references for #655

polluks:
Some references for easier finding the sample code.

PR #660, opened 2018-05-23
More references

polluks:


doc/tgi.sgml
```diff
@@ -9,7 +9,8 @@
 
 <abstract>
 The cc65 library provides functions for platform independent graphics.
-Include the tgi.h header file to get the necessary definitions.
+Include the tgi.h header file to get the necessary definitions, see also
+<tt/samples/tgidemo.c/ and <tt/samples/mandelbrot.c/.
```
oliverschmidt: I presume that Gregs comment on https://github.com/cc65/cc65/pull/659 saying `Too many slashes! You need to use the "long form" of the tag: <tt></tt>.` is relevant here too.

polluks: Indeed, the parser cannot get it.

Issue #525, opened 2017-11-17
No cbm_k_tksa


dmantione:
Hello,

I am missing another essential KERNAL call in cbm.h: There is no cbm_k_tksa(). This time it seems to be really missing, there is no trace of it in the source. The call is essential because after a TALK, the next step to do in serial bus communication is to transmit a secondary address on the serial bus, i.e. cbm_k_talk() is useless without it.

For example, a routine to read the error channel 15 from a floppy drive is as follows:
```
char drive_error_code;

char device_read_error(char device) {
  char* p;
  cbm_k_talk(device);
  asm ("lda #$6F"); /* $60 + $0F commands secondary channel $0F to talk to the computer */
  asm ("jsr $FF96");
  p=(char *)0x200;
  do
    *p=cbm_k_acptr();
  while (*p++!=13);
  *p=0;
  cbm_untlk(device);
//  printf("%s\n",0x200);
  p=(char *)0x200;
  drive_error_code=10*(p[0]-0x30)+(p[1]-0x30);
  return drive_error_code;
}
```
The use of inline assembler is ugly, so it would be usefull to have a cbm_k_tksa().

spiro-trikaliotis:
You're right.
And cbm_k_second (for the secondary address after listen) is also missing.
And cbm_k_untlk
And some others, too.

It seems one should go through the list of KERNAL routines and add the ones that are missing.

oliverschmidt:
Contributions welcome...

Compyx:
This looks like something I might be able to help with, doesn't appear to be too complicated. I already implemented the cbm_k_tska() function once to get a feel for the cc65 codebase.

But before I start, any pointers to documentation on how the cc65 codebase is structured, code style guidelines, that sort of stuff?

oliverschmidt:
For of all thanks for your interest in contributing to cc65 :-)

If you check out https://github.com/cc65/cc65/tree/master/libsrc/cbm there are a bunch of files named c_*.s - I'd say whatever you do should fit in there seamlessly and that's all...

Compyx:
Looks like adding these missing Kernal calls will mostly involve copy-pasting from already existing files. And of course some proper testing.

I'll have a go at it :)

greg-king5:
The style guides are in the wiki:
https://github.com/cc65/wiki/wiki#this-cc65-projects-coding-guidelines

Compyx:
Thanks for the Wiki link, that should be enough to get me started.

[...]

PR #662, opened 2018-05-23
Add CBM kernal call SECOND to CBM library (see issue #525)

Compyx:
This will add the CBM kernal call SECOND to the CBM library.

### DIFF
--- a/doc/c128.sgml
+++ b/doc/c128.sgml
@@ -125,6 +125,7 @@ declaration and usage.
 <item>cbm_k_bsout
 <item>cbm_k_clrch
 <item>cbm_k_tksa
+<item>cbm_k_second
 <item>cbm_load
 <item>cbm_open
 <item>cbm_opendir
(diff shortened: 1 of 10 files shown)


Compyx:
I actually used that page to figure out which Kernal calls are supported by which machine, after checking various KERNAL's provided with VICE. The table appears to be correct.

For example, TKSA ($FF96) is correct for anything expect the old PET, the PET has the BASIC DOPEN call at $FF96.

Unfortunately I only have three C64's (two breadboxes, one C64C) and a 1541-II and an rrnetmk3 to test code. I never had a PET, I was born three years before the PET was released.
Doesn't really matter much for adding KERNAL calls, they're basically aliases for SYS ${ROUTINE}, but it might make a difference for less obvious code.

spiro-trikaliotis:
Ok, I was not aware if you know it or if not, that's why I though I might give you a hint at it.

Compyx:
Thanks for the hint anyway, just confirms I found a page with correct info :)

PR #663, opened 2018-05-24
Adapted div & mod for popptr1.

IrgendwerA8:


libsrc/runtime/absvaludiv.s
```diff
@@ -0,0 +1,22 @@
...
+;
+; When negating values, we will ignore the possibility here, that one of the
+; values is $8000, in which case the negate will fail.
+
+        .export         absvaludiv
+        .import         _abs, popax, udiv16
+        .importzp       ptr1, ptr4
+
+
+absvaludiv:
```
greg-king5: That name should be changed to `absvaludiv16` because someone might want to make an `absvaludiv32` function, in the future.

greg-king5:
After this PR is merged, everyone must "clean" away your libraries.  You must build fresh libraries, so that you will avoid a clash between the old "abs.s" and the new "negabs.s".

IrgendwerA8:
Changed naming. I had a hard time deciding if the 2x _abs calls in absvaludiv16 are a good idea, as they clearly favour size before speed (8 bytes vs. +22 cycles in worst case). Since I saved some cycles in other places, the worst case is only a little bit slower now. Anyway not sure if changing "idiv32by16r16" in the same way to save four bytes is really desired?

This just puts more emphasis on doing unsigned math/types with cc65 whenever possible, as the unsigned version is now faster.

libsrc/runtime/div.s
```diff
@@ -1,37 +1,29 @@
...
-
-; Result is negative
-
-        lda     sreg            ; Load low byte of result
-        jmp     negax           ; Adjust the sign
-
-; Result is positive
-
-Pos:    lda     sreg
+        pha                         ; check if high-bytes indicate
```
oliverschmidt: The /libsrc/runtime pretty consistently uses Uppercase for the first letter of comments. Please keep to that style for the sake of consistency.

oliverschmidt:
> -8 bytes vs. +24 cycles in worst case

Can you please elaborate a bit what that worst case is like and the probability of the different possible cycle counts?

> Anyway not sure if changing "idiv32by16r16" in the same way to save four bytes is really desired?

Can you please elaborate a bit on the number of additional cycles and their probability?

> This just puts more emphasis on doing unsigned math/types with cc65 whenever possible, as the unsigned version is now faster.

This seems in general an appropriate perspective :-)

IrgendwerA8:
* The worst case is unfortunately the (I guess) most usual one, where both operands are positive. The previous version checked each operand and only called in the negative case "negax" while this here calls "abs" to perform the check and negax in line -> saves the code for the separate checks, but is a function call for each operand in any case.

* the same could be applied for e.g. "idiv32by16r16" (line 25/26): Just call "abs" to do the check and correction (-4 bytes but every time a subroutine call).

I've written a small benchmark program to illustrate the change: Calling 100000 times a (signed) int division took (-Oirs) 18.92 sec (Atari target) vs. 19.62 now. (File size 3400 vs. 3380 bytes).
(The "unsigned int" case is now 17.8 sec BTW.)

I can also remove the adaptations for the signed div/mod and having (part) of the size savings (no need for popsreg in div/mod) intact, but since I worked on these files anyway I thought it would be a good opportunity to have a look for savings too. In the end all comes down here what to favour: size or speed - and where to draw the line... (this is a honest request, asking for a second view on this)

IrgendwerA8:
..and to give a complete picture (the former numbers are from the 8-bit div case) here the numbers from the 16 bit case, where my second optimization is accounted:
                             old  |  new
---------------------------------------
unsigned          25.84  |  24.58
signed              26.68  |  25.90

Times in seconds.
As visible, the abs-call draw-back is fully compensated by the other savings.

oliverschmidt:
I'm really sorry but now you have totally lost me. I have no clue what number relates to which scenario. I'd really like to try to help making  good decisions, but I need the data in an easier-to-consume form :-(

IrgendwerA8:
```
10000 Division (all times in seconds, "-Oirs -t atari")

                   current master | pull request #663
-----------------------------------------------------
 8 bit unsigned             18.06 | 17.88 
 8 bit signed               18.92 | 19.62 (slower)
16 bit unsigned             25.84 | 24.58
16 bit signed               26.68 | 25.90

Executable size for signed division -20 bytes.
Abs() three cycles faster.
```
I hope this summary helps...

oliverschmidt:
Thanks :-)

If we would say: "We don't want anything to be slower. Period." What options would we have then?

oliverschmidt:
> Like said: Size or speed...

For sure. The question is how much size and how much speed ;-)

My personal opinion would be here to go for the second option: No slowdown for 8 bit signed and less size reduction.

> but then the decision has to be made if a 4 bytes bigger negax with (faster) abs() is still a good idea.

Sorry, but I'd need again some table or alike to make sure I really get it right.

oliverschmidt:
> To reduce the drawback when using abs() I've joined "negax" and "abs" in the new file "negabs.s". If a program uses now abs() the code will be 3 bytes shorter and 3 cycles quicker. But if a program doesn't need abs and only uses negax, the code size is increased by 4.

That seems a very reasonable tread-off to me.

IrgendwerA8:
So code size will be then in my sample prg only 12 bytes shorter.

I apply the changes, like discussed?

oliverschmidt:
> allows partially inclusion of code and not only file-wise one. Quite a lot functions could benefit (speed & size) from that change - but that's future...

Again sorry, but... Are you saying that today you have to chose between splitting some code in two files in order to be able to link only parts of it or keeping it in one file in order to be able to use relative branches between the parts of it (instead of absolute JMPs/JSRs). But with the improved library format one could have both.

oliverschmidt:
> I apply the changes, like discussed?

Please do so. Thanks in advance!

IrgendwerA8:
Exactly.

IrgendwerA8:
Going back (partially) to the master-sources, I found nevertheless an operation to optimize. So these are the numbers now:
```
                    current master | current pull request
--------------------------------------------------------
 8 bit unsigned             18.06 | 17.88 
 8 bit signed               18.92 | 18.66
16 bit unsigned             25.84 | 24.58
16 bit signed               26.68 | 24.92
```
File size is reduced by 13.

oliverschmidt:
> Going back (partially) to the master-sources

I hope you're not disappointed. As far as I understand you explicitly asked for feedback and I did my best to provide reasonable feedback...

> I found nevertheless an operation to optimize. So these are the numbers now: [...] File size is reduced by 13.

Wow, all operations are faster __AND__ the size is reduced by more than 10 bytes :-))

greg-king5:
Oops, I forgot that the `div()` function's implementation is tied intimately to the way that the division operator's runtime uses the cc65 pseudo-registers.  The remainder part of "libsrc/common/divt.s" needs to be adjusted.

@oliverschmidt, that bug is why `mktime()` fails.

IrgendwerA8:
I'll examine the issue - give me some time.

PR #668, opened 2018-05-28
Mul second thoughts

IrgendwerA8:
Finally found a way to regain the 2 lost bytes here.


oliverschmidt:
@IrgendwerA8:

I just noticed that I seem to have no way to just get in touch with you. Therefore I comment on this PR presuming that you get notified...

The other day I discovered a regression almost certainly related to your arithmetic optimizations. It manifests in the code compiled from https://github.com/cc65/cc65/blob/master/libsrc/common/localtime.c

I created a regression test for it that currently fails, see https://github.com/cc65/cc65/commit/b04028b5d8ea4ab9984f9ab648791013004dcefb

Obviously I / we would appreciate it very much if you could find and fix the issue.

PR #673, opened 2018-06-05
Fix `make` failing in parent directories containing spaces

bengdahl:
Fails to build when running `make` while the repo is in a parent directory containing a space.

Output for directory `/example dir/cc65`:
```
ar65/del.c
gcc: error: dir/cc65/asminc: No such file or directory
gcc: error: dir/cc65/include: No such file or directory
gcc: error: dir/cc65/target: No such file or directory
gcc: error: dir/cc65/lib: No such file or directory
gcc: error: dir/cc65/lib: No such file or directory
gcc: error: dir/cc65/cfg: No such file or directory
make[1]: *** [Makefile:151: ../wrk/ar65/del.o] Error 1
make: *** [Makefile:6: all] Error 2
```

To replicate:
1. Clone into a directory with a space in its name
2. `cd cc65`
3. `make`

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -64,8 +64,8 @@ endif
 
 CFLAGS += -MMD -MP -O3 -I common \
           -Wall -Wextra -Wno-char-subscripts $(USER_CFLAGS) \
-          -DCA65_INC=$(CA65_INC) -DCC65_INC=$(CC65_INC) -DCL65_TGT=$(CL65_TGT) \
-          -DLD65_LIB=$(LD65_LIB) -DLD65_OBJ=$(LD65_OBJ) -DLD65_CFG=$(LD65_CFG) \
+          -DCA65_INC="$(CA65_INC)" -DCC65_INC="$(CC65_INC)" -DCL65_TGT="$(CL65_TGT)" \
+          -DLD65_LIB="$(LD65_LIB)" -DLD65_OBJ="$(LD65_OBJ)" -DLD65_CFG="$(LD65_CFG)" \
           -DGIT_SHA=$(GIT_SHA)
 
 LDLIBS += -lm
@@ -148,7 +148,7 @@ endef # PROG_template
 
 ../wrk/%.o: %.c
 	@echo $<
-	@$(CC) -c $(CFLAGS) -o $@ $<
+	$(CC) -c $(CFLAGS) -o $@ $<
 
 ../bin:
 	@$(call MKDIR,$@)
--- a/src/Makefile
+++ b/src/Makefile
@@ -148,7 +148,7 @@ endef # PROG_template
 
 ../wrk/%.o: %.c
 	@echo $<
-	$(CC) -c $(CFLAGS) -o $@ $<
+	@$(CC) -c $(CFLAGS) -o $@ $<
 
 ../bin:
 	@$(call MKDIR,$@)


oliverschmidt:
From my POV the issue you mention is rather a conceptual issue because of spaces in directory paths in dependency files. I don't see how a Makefile could prevent those issues. So if you can't prove me wrong there I don't consider your change a fix - and will therefore not merge it.

groessler:
The gcc 7.3.0 on my Linux box seems to handle spaces in dependency file correctly. The spaces are escaped with a backslash, and subsequent "make" calls don't report unsatisfied dependencies.

But, spaces in file or directory names are evil, we shouldn't support that bad habit. I hate them.

Disregarding personal opinions, I think the change itself is good. But it's up to you, Oliver, to merge or not.

mrdudz:
i'd say a makeifle that fails because of spaces in some path is just broken :)

oliverschmidt:
> i'd say a makeifle that fails because of spaces in some path is just broken :)

You're missing the point. At least the last time I checked it, it was the way the compiler and the make tool interact for dependency handling that failed. So the only way to write a Makefile that doesn't fail because of spaces in some path was to drop incremental builds altogether. I'd say it is/was a no-brainer what to prefer :-)

oliverschmidt:
> The gcc 7.3.0 on my Linux box seems to handle spaces in dependency file correctly. The spaces are escaped with a backslash, and subsequent "make" calls don't report unsatisfied dependencies.

That's a pretty interesting information. However, could you please verify that not only the dependency files don't cause build issues but that they still actually work? Meaning to change some header file ans see if the correct source files are recompiled/relinked? Thanks in advance.

Additionally I'm wondering how a "good" gcc alone helps. As far as I remember the C libraries Makefile has the cc65 compiler generate dependency files too !?!

oliverschmidt:
> As far as I remember the C libraries Makefile has the cc65 compiler generate dependency files too !?!

Yep, indeed:
https://github.com/cc65/cc65/blob/master/libsrc/Makefile#L266
https://github.com/cc65/cc65/blob/master/libsrc/Makefile#L273

oliverschmidt:
> you make it sound like fixing cc65 to not output broken dependencies is a problem :)

No, of course not. It makes me just wonder if the test conducted by Chris is the test I have in my mind or if he tested something different - because the test that I think of would be doomed to fail because of the dependency files generated by cc65.

If I wouldn't see the option to adjust cc65 (and ca65) to do the same the (recent ?) gcc dosn't I wouldn't have classified Chris' findings as "pretty interesting" ;-)

groessler:
> i'd say a makeifle that fails because of spaces in some path is just broken :)

I'd say that if a makefile fails because of spaces in some path the user is broken. :-)  (s)he uses spaces in her/his path names.

groessler:
> That's a pretty interesting information. However, could you please verify that not only the dependency files don't cause build issues but that they still actually work? Meaning to change some header file and see if the correct source files are recompiled/relinked? 

Yes, seems to work.
Also apparently works in cc65 generated dependency files (cc65 escapes spaces with backslashes there, too).

Just have the bad feeling that with this message I'm going support something I don't actually like (spaces in path names). But I guess, it's better for the overall experience of "random cc65 user"...

bengdahl:
I made this pull request because I expect to be able to clone a repository, run `make` or whatever the repo uses, install, and be finished with it. The failure to build disrupts that and wastes time trying to get things working. 

The failure I'm getting seems to originate in CFLAGS in `src/Makefile`. The flags containing paths are not escaped so when `gcc` is called on line 151 it fails.

> The gcc 7.3.0 on my Linux box seems to handle spaces in dependency file correctly. The spaces are escaped with a backslash, and subsequent "make" calls don't report unsatisfied dependencies.

The dependencies seem to use escaped paths on my machine as well and recompile correctly when changing a header. I'm using gcc 8.1.0. 

> I'd say that if a makefile fails because of spaces in some path the user is broken. :-) (s)he uses spaces in her/his path names.

Yes, I know something's wrong with me for using spaces :P

oliverschmidt:
Okay...

So I played a bit on my own:

- Both gcc and cc65 are asked to create dependencies only for `#include "?.h"` and not for `#include <?.h>`. All files include via the former syntax are foun via relative paths so there's no reason for any C dependency to contain a path with spaces in the first place.

- ca65 creates dependencies with absolute paths (for whatever reason) but spaces in those paths are escaped with `\` and seem to work just fine. Both on *IX and on Windows with MinGW-w64 gnumake running in cmd.exe.

I didn't encounter any problem with a full build nor with an incremental build both on *IX and on Windows (with above setup) :-) So I just don't know anymore what type of issue I had with dependency files in the past. The minimalist change proposed here in fact DOES allow to build cc65 in a path with spaces!

However...

We don't only want to successfully build cc65 but also use it, don't we?

I know of three conceptually different areas where cc65 needs "its own path":

1. cl65 spawning cc65 and friends.

2. Any cc65 binary accessing files like headers and libraries.

3. `cl65 --print-target-path` allowing access to drivers and target tools from within Makefiles.

- On *IX both 1.) and 2.) seem to work just fine - although I don't fully understand why ;-)

- On Windows 1.) fails and 2.) seems to work.

- For 3.) it seems safe to just always enclose the output path in a pair of `"`. Or does somebody disagree?

groessler:
> On Windows 1.) fails and 2.) seems to work.

So this seems to have to be fixed anyway.

> For 3.) it seems safe to just always enclose the output path in a pair of ". Or does somebody disagree?

I'm not sure. I don't know the use cases. I think better would be to escape special chars with "\". Should work on *IX, don't know about Windows.

groessler:
... escape with backslash ... (again stupid github editor)

oliverschmidt:
> ... escape with backslash ...

I see - and agree.

oliverschmidt:
Having this PR merged together with https://github.com/cc65/cc65/commit/d861be8ad569d3928727a8c8162fc843835035db should allow cc65 to build and work in parent directories containing spaces on *IX.

oliverschmidt:
https://github.com/cc65/cc65/commit/b74ab9de899edd35738cd6b08693f3f2a9a17dae should allow cc65 to build and work in parent directories containing spaces on Windows too.

PR #676, opened 2018-06-06
Fix joystick for the Supervision target

Fabrizio-Caruso:
This PR fixes the Supervision joystick detection (same fix as the one I found for the Gamate target)

I have tested with my game (CROSS CHASE).
I have tested the Fire 1 button and  UP, DOWN, LEFT, RIGHT movements.

### DIFF
--- a/libsrc/supervision/joy/supervision-stdjoy.s
+++ b/libsrc/supervision/joy/supervision-stdjoy.s
@@ -77,5 +77,6 @@ COUNT:
 
 READJOY:
         lda     sv_control
+	eor     #$FF
         ldx     #0
         rts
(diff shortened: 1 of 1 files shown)


libsrc/supervision/joy/supervision-stdjoy.s
```diff
@@ -77,5 +77,6 @@ COUNT:
 
 READJOY:
         lda     sv_control
+	eor     #$FF
```
oliverschmidt: Sorry for the nitpicking but you need to use spaces instead of a tab char for indentation to have me merge this pr as-is.

Fabrizio-Caruso:
I have replaced the tab with spaces.

PR #680, opened 2018-06-10
cmake toolchain file added to project

Grandmother:
file have been gotten from
https://github.com/fo-fo/ngin/blob/master/cmake/toolchains/cc65-toolchain.cmake

### DIFF
--- a/cc65-toolchain.cmake
+++ b/cc65-toolchain.cmake
@@ -1,0 +1,41 @@
+# from https://github.com/fo-fo/ngin
+# CMake toolchain file for cc65
+# This is largely a result of experimentation, so some things may be done
+# suboptimally/wrong. Some compilation options/CMake features may not work.
+# Some generators will also not work (like MSVS). Ninja has been tested to work.
+# What is supported: C, assembly, static libraries
+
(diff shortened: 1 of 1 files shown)


oliverschmidt:
You ask for that file to be placed in the root directory of cc65. As far as I understand a .cmake file placed there implies that cc65 __itself__ could be built using that file. But from the contents of the file I seem to understand that - at best - the file is supposed to allow some project be be built __using__ cc65.

When I understand things right then I could image to have a Wiki page created the is linked at https://github.com/cc65/wiki/wiki#Tips_and_tricks containing the source code of the .cmake file and some hints on how to use it and/or what type of project it can be used for. In case you're interested to do so please let me know and I provide you with write access to the Wiki.

Grandmother:
Yes, it would be the bast way to connect this information with project. Now I'm trying to solve some problems I met and to chat with author. When I at least solve my problems, I'll add information to wiki.

oliverschmidt:
I just invited you to the cc65 Wiki Editors team. Please make sure to create new cc65 Wiki pages in the MediaWiki syntax (instead of the pupular default MarkDown syntax).

Let me close this PR now...

PR #681, opened 2018-06-10
Support for "inline parameters" of subroutines

shinra-jp:
Example of 2 bytes virtual operands:
`jsr	SomeProc`
`.byte	$00, $01 ; argument to SomeProc`
`; return here from SomeProc`
`bit	$3F`

### DIFF
--- a/doc/da65.sgml
+++ b/doc/da65.sgml
@@ -542,6 +542,18 @@ code. The following attributes are recognized:
   range, where <tt/label/ is the label name given with the <tt/NAME/
   attribute, and <tt/offs/ is the offset within the data.
 
+  <tag><tt>VOPERAND</tt></tag>
+  This optional attribute is followed by a numerical value. It tells the
+  assembler that subroutine calls to this label follow "virtual operands"
+  of the given bytes like this:
+
+<tscreen><verb>
+       	JSR	LabelWith2BytesOfVoperand
+	.byte	$00, $10    ; virtual operands
+	; return here
+	BIT	$0F
+</verb></tscreen>
+
 </descrip>
 
 
--- a/doc/da65.sgml
+++ b/doc/da65.sgml
@@ -542,16 +542,16 @@ code. The following attributes are recognized:
   range, where <tt/label/ is the label name given with the <tt/NAME/
   attribute, and <tt/offs/ is the offset within the data.
 
-  <tag><tt>VOPERAND</tt></tag>
+  <tag><tt>PARAMSIZE</tt></tag>
   This optional attribute is followed by a numerical value. It tells the
-  assembler that subroutine calls to this label follow "virtual operands"
+  assembler that subroutine calls to this label follow "inline parameters"
   of the given bytes like this:
 
 <tscreen><verb>
-       	JSR	LabelWith2BytesOfVoperand
-	.byte	$00, $10    ; virtual operands
-	; return here
-	BIT	$0F
+        JSR     LabelWithParamSize2
+        .byte   $00, $10
+        (return here)
+        code...
 </verb></tscreen>
 
 </descrip>
--- a/doc/da65.sgml
+++ b/doc/da65.sgml
@@ -544,8 +544,8 @@ code. The following attributes are recognized:
 
   <tag><tt>PARAMSIZE</tt></tag>
   This optional attribute is followed by a numerical value. It tells the
-  assembler that subroutine calls to this label follow "inline parameters"
-  of the given bytes like this:
+  assembler that subroutine calls to this label are followed by
+  "inline parameters" with the given number of bytes, like this:
 
 <tscreen><verb>
         JSR     LabelWithParamSize2
--- a/src/da65/handler.c
+++ b/src/da65/handler.c
@@ -748,19 +748,19 @@ void OH_JmpAbsoluteXIndirect (const OpcDesc* D)
 
 void OH_JsrAbsolute (const OpcDesc* D)
 {
-    unsigned ParamSize = SubroutineParamSize[GetCodeWord(PC+1)];
+    unsigned ParamSize = SubroutineParamSize[GetCodeWord (PC+1)];
     OH_Absolute (D);
     if (ParamSize > 0) {
         unsigned RemainingBytes;
-	unsigned BytesLeft;
+        unsigned BytesLeft;
         PC += D->Size;
-        RemainingBytes = GetRemainingBytes();
+        RemainingBytes = GetRemainingBytes ();
         if (RemainingBytes < ParamSize) {
             ParamSize = RemainingBytes;
         }
         BytesLeft = ParamSize;
         while (BytesLeft > 0) {
-            unsigned Chunk = (BytesLeft > BytesPerLine)? BytesPerLine : BytesLeft;
+            unsigned Chunk = (BytesLeft > BytesPerLine) ? BytesPerLine : BytesLeft;
             DataByteLine (Chunk);
             BytesLeft -= Chunk;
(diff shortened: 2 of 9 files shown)


doc/da65.sgml
```diff
@@ -542,6 +542,18 @@ code. The following attributes are recognized:
   range, where <tt/label/ is the label name given with the <tt/NAME/
   attribute, and <tt/offs/ is the offset within the data.
 
+  <tag><tt>VOPERAND</tt></tag>
+  This optional attribute is followed by a numerical value. It tells the
+  assembler that subroutine calls to this label follow "virtual operands"
+  of the given bytes like this:
+
+<tscreen><verb>
+       	JSR	LabelWith2BytesOfVoperand
```
oliverschmidt: Your sample code makes (inconsistent) use of tab chars. Please replace all tab chars with blanks (aka space chars).

src/da65/handler.c
```diff
@@ -741,3 +743,31 @@ void OH_JmpAbsoluteXIndirect (const OpcDesc* D)
...
+    OH_Absolute (D);
+    if (VOperandSize > 0) {
+        unsigned RemainingBytes;
+        PC += D->Size;
+        RemainingBytes = GetRemainingBytes();
+        if (RemainingBytes < VOperandSize) {
+            VOperandSize = RemainingBytes;
+        }
+        if (VOperandSize > 0) {
+            DataByteLine (VOperandSize); /* FIXME: follow BytesPerLine */
```
oliverschmidt: Please elaborate why you consider this ready to be merged when there's a FIXME comment.

doc/da65.sgml
```diff
@@ -542,6 +542,18 @@ code. The following attributes are recognized:
...
+	.byte	$00, $10    ; virtual operands
```
oliverschmidt: As far as I understand your pull request doesn't make da65 ouput the "; virtual operands" so from my pov it should rather no be part of this sample, should it?

doc/da65.sgml
```diff
@@ -542,6 +542,18 @@ code. The following attributes are recognized:
...
+	; return here
```
oliverschmidt: As far as I understand your pull request doesn't make da65 ouput the "; return here" so from my pov it should rather no be part of this sample, should it?

doc/da65.sgml
```diff
@@ -542,6 +542,18 @@ code. The following attributes are recognized:
   range, where <tt/label/ is the label name given with the <tt/NAME/
   attribute, and <tt/offs/ is the offset within the data.
 
+  <tag><tt>VOPERAND</tt></tag>
+  This optional attribute is followed by a numerical value. It tells the
+  assembler that subroutine calls to this label follow "virtual operands"
```
oliverschmidt: As you can see from my comments below on your sample code that code can be misinterpreted easily. Therefore I suggest that you try to actually describe what VOPERAND does rather just say "like this".

doc/da65.sgml
```diff
@@ -542,6 +542,18 @@ code. The following attributes are recognized:
...
+  of the given bytes like this:
+
+<tscreen><verb>
+       	JSR	LabelWith2BytesOfVoperand
+	.byte	$00, $10    ; virtual operands
+	; return here
+	BIT	$0F
```
oliverschmidt: BIT instructions are sometimes used to skip parts of instructions (see http://www.6502.org/tutorials/6502opcodes.html#BIT). Therefore I suggest that your sample code uses something else to avoid a misleading association for the reader.

mrdudz:
> Apart from my inline comments I'm wondering about the name "virtual operands".

i find it a weird name as well. however, although i am quite familiar with that technique, i am not sure how to call it. something like "arguments_follow_jsr"? "args_follow_inlined"?

oliverschmidt:
Okay, so primarily we need a name for the LABEL attribute. I understand that "the technique" is familiar and that there's nothing like a well known ABI for the 6502. Therefore we can very well proclaim "the technique" as THE way to pass operands/parameters/arguments.

- The term 'operand' is usually used for asm op-codes.
- The term 'argument' is used by da65 for asm op-codes.

So we're left with 'parameter'. Therefore I suggest the name PARAM_SIZE for the new LABEL attribute.

And coming from there I suggest to call "the technique" 'Inline Parameters'.

doc/da65.sgml
shinra-jp: Fixed.

src/da65/handler.c
shinra-jp: I found a simple implementation to honor the BytesPerLine without big changes to the data.c.

doc/da65.sgml
shinra-jp: Changed the sample not to be misinterpreted as real codes.

shinra-jp:
Thank you for your reviews. I changed the term to "inline parameters" and the attribute name to "PARAMSIZE" (w/o an underscore) as suggested.

doc/da65.sgml
```diff
@@ -542,6 +542,18 @@ code. The following attributes are recognized:
   range, where <tt/label/ is the label name given with the <tt/NAME/
   attribute, and <tt/offs/ is the offset within the data.
 
+  <tag><tt>PARAMSIZE</tt></tag>
+  This optional attribute is followed by a numerical value. It tells the
+  assembler that subroutine calls to this label follow "inline parameters"
+  of the given bytes like this:
```
greg-king5: That sentence needs to be reworded to say, '... subroutine calls to this label are followed by "inline parameters" with the given number of bytes, like this:'.

src/da65/handler.c
```diff
@@ -741,3 +743,35 @@ void OH_JmpAbsoluteXIndirect (const OpcDesc* D)
     }
     SeparatorLine ();
 }
+
+
+
+void OH_JsrAbsolute (const OpcDesc* D)
+{
+    unsigned ParamSize = SubroutineParamSize[GetCodeWord(PC+1)];
```
oliverschmidt: Please follow the (unusal) style with a space before the parameter list.

src/da65/handler.c
```diff
@@ -741,3 +743,35 @@ void OH_JmpAbsoluteXIndirect (const OpcDesc* D)
...
+    OH_Absolute (D);
+    if (ParamSize > 0) {
+        unsigned RemainingBytes;
+	unsigned BytesLeft;
+        PC += D->Size;
+        RemainingBytes = GetRemainingBytes();
```
oliverschmidt: Please follow the (unusal) style with a space before the parameter list.

src/da65/infofile.c
```diff
@@ -380,13 +381,15 @@ static void LabelSection (void)
...
         {   "SIZE",     INFOTOK_SIZE    },
+        {   "PARAMSIZE", INFOTOK_PARAMSIZE },
     };
 
     /* Locals - initialize to avoid gcc warnings */
     char* Name    = 0;
     char* Comment = 0;
     long Value    = -1;
     long Size     = -1;
+    long ParamSize = -1;
```
oliverschmidt: Please re-align the '=' on the lines above.

src/da65/infofile.c
```diff
@@ -380,13 +381,15 @@ static void LabelSection (void)
         {   "ADDR",     INFOTOK_ADDR    },
         {   "NAME",     INFOTOK_NAME    },
         {   "SIZE",     INFOTOK_SIZE    },
+        {   "PARAMSIZE", INFOTOK_PARAMSIZE },
```
oliverschmidt: Please re-align the lines above.

PR #687, opened 2018-06-21
Smaller driver code

polluks:


greg-king5:
I suspect that the original code was `jmp EXIT`.  But then, it was replaced by that inline exit code, in order to gain a little extra speed (and, re-enable interrupts a few cycles sooner).

oliverschmidt:
@greg-king5: Should I interpret your comment as rather not wanting to see this pull request merged?

polluks:
At least EXIT should removed.

greg-king5:
Yes, I think that the code path should keep its directly inlined exit instructions.

Stefan's advice about the EXIT label is wise, IMHO.

oliverschmidt:
Thanks for coming to an agreement, gentlemen :-)

PR #688, opened 2018-06-21
RFC: Common kernal inc file for Commodore targets

ops:
Based on info found in

https://www.pagetable.com/?p=926

I have created a common kernal include file for CBM targets. These definitions are available for asm programmers and files in libsrc could use this file to define kernel entry points instead of defining their own.


oliverschmidt:
Okay, so it's again me having no idea of CBM stuff...

It seems to me that the symbols defined here are already "handled" elsewhere in the cc65 code base. How does this file relate to the existing cc65 code base. Is it supposed to replace something? Or is it supposed to be an addition?

When I look e.g. at `IOINIT := $FF84` I notice that there's already https://github.com/cc65/cc65/blob/master/libsrc/c64/kernal.s#L8 and https://github.com/cc65/cc65/blob/master/libsrc/c64/kernal.s#L51 so how's that supposed to co-exsist?

Apart from that I see e.g.
```
.elseif .def(__CBM510__) || .def(__CBM610__)
  IOINIT       := $FF7B
```
but both https://github.com/cc65/cc65/blob/master/libsrc/cbm510/kernal.s#L45 and https://github.com/cc65/cc65/blob/master/libsrc/cbm610/kernal.s#L45 say `IOINIT          = $FF84`. Is that supposed to mean that those two files are wrong?

Sorry, but I think nobody can expect me to decide on this pull request without having answers to those questions...

spiro-trikaliotis:
The CBM-II (510, 610) have IOINIT and CINT at different addresses than the C64 and later. This is what https://www.pagetable.com/?p=926 talks about.

That is, libsrc/cbm510/kernal.s and libsrc/cbm610/kernal.s are wrong in this aspect.

In fact, I don't think this is the only error in the other includes.

Note, howver, that although the VIC 20 does not have CINT and IOINIT in the jump table, the file libsrc/vic20/kernal.s defines them, nevertheless, as direct calls into the ROM (that is, not through the jump table). This is something that this pull request has left out.

Thus, you are right, these are double info, and it should be cleaned up. 

However, I am not sure how to do this:
1. Leave this file here out, and fix all the other libsrc/*/kernel.s files
2. Remove the libsrc/*/kernel.s files, and replace them by this file here.
3. like 2, but also include any additional addresses (like CINT, IOINIT in the case of the VIC 20)

Somehow, I tend to no. 3, but YMMV.

ops:
My primary goal was to make these definitions available for asm programmers so it is an addition in that sense. If this get merged, libsrc/kernel.s can use of this file by include and not repeat the definitions. This is not meant to replace libsrc files since those contain exports too.

VIC-20 CINT and IOINIT are already handled by this PR. And there's BSOUT/CHROUT aliases too.

oliverschmidt:
Thanks for the feedback on my comment :-)

So it seems to me that I didn't get things wrong after all. It's actually true that this pull request is can't just be merged as-is and things are fine. Rather it opens up quite some questions around handling of Kernal entry points...

Maybe it's a good thing to start with the beginning:

> My primary goal was to make these definitions available for asm programmers [...]

I think it's a no-brainer that a tool chain that is used as CBM assembler should have an include file in its "public" asm include file directory with the Kernal entry points. But cc65 has none, so I'm wondering why nobody has complained so far.

The next strange thing is this completely commented-out section: https://github.com/cc65/cc65/blob/master/libsrc/cbm/cbm.inc#L6-L48

And finally we have those machine specific kernal.s files like https://github.com/cc65/cc65/blob/master/libsrc/c64/kernal.s but those are from my pov strange too. Why do they contain `.export` statements requiring individual `.import` statements? I would have expected those files to be split into an .inc file with `.global` statements (see http://cc65.github.io/doc/ca65.html#ss11.46) and an .s file with the label assignment operators (`:=`).

So at least to me the whole thing looks rather broken right now :-(

Without looking at the pull request at hand there are several general options:

1. A single .inc file in asminc with a bunch of `:=` operators using '.ifdef`.
2. A single .inc file in asminc with a bunch of `.global` statements and several machine specific .s files with  `:=` operators (without .ifdef).
3. A single .inc file like in 1.) __*plus*__ a single .inc like in 2.) but the latter placed in libsrc/cbm.

The benefit of 1.) is obvious. It's simple and there's no duplication.

The benefit of 2.) is that it's more "cc65-ish" but it requires the user to link against the C library to get the actual symbols. My personal asm projects tend to link against the C library anyway because I lend the exe/BASIC/... header from there but that's a different story...

The benefit of 3.) is that a C library could be modified to use banking and then the Kernal entry point symbols wouldn't be `:=` operators anymore but the addresses of short chunk routines. See e.g. https://github.com/cc65/cc65/blob/master/libsrc/apple2/cout.s (which doesn't use `.global` either). In this scenario a straight asm program surely doesn't want to use banking just because the C library does.

Opinions? Thoughts?

ops:
This is getting a bit more complicated than I expected. I simply wanted make kernal entries available for asm programmers. Since entries are the mostly same across different CBM platforms I wanted to define them in one place. OK, done.

Now we have definitions for all CBM platforms and we can reuse these definitions in libsrc/xxx/kernal.s files and not to repeat them in every file.

Please see
https://github.com/ops/cc65/commit/f7fbac4c6b409182a09b1c1ac493d04d8e1d6fe8

oliverschmidt:
> This is getting a bit more complicated than I expected.

;-)

> The CBM 510 and CBM 610 have an arcane memory-banking scheme. Machine language programs on those computers cannot see the Kernal ROM and the I/O chips directly.

Do I get it right that this isn't an optimization for the C library but necessary for every little asm program?

My general understanding so far:

The needs of a "simple" standalone asm program and C library may be / are different. I wrote that as theoretical option 3.) in my last comment and Greg pointed out that it's already reality on the plus4.

So trying to re-phrase what I wrote above:

Either we consider "simple" standalone asm programs and the C library two separate worlds. Then we can add the content of this pull request as-is (mybe with a diffrent file name) and that's it. The file is ignored by the C library.

Or we see that there's a big overlap between "simple" standalone asm programs and the C libary. Then we want to use the content of this pull request in the C library where appropriate to reduce duplication/multiplication of defines.

When I get things right the latter means that the plus4 C library does _not_ make use of the new content. 

For the CBM510/CBM610 things depend on the Greg's answer to my question above.

ops:
I think that issues Greg and you raised are not in the scope of this PR and my subsequent commit https://github.com/ops/cc65/commit/f7fbac4c6b409182a09b1c1ac493d04d8e1d6fe8

I'm not familiar with CBM510/CBM610 or Plus4 but if the support for these platforms were broken, it still is. If they previously worked OK they are still OK. The only change is that standard kernal entry points for asm programmer are now provided by ca65.  Additionally, the source code of library is cleaned from duplicate definitions. The binaries are still same as before.

oliverschmidt:
I think the point is that I (and Greg ?) would have loved to see a header file that can be used to replace e.g. the line
```
.import         SETLFS, OPEN, CHKIN, BASIN, CLRCH, BSOUT, READST
```
at https://github.com/cc65/cc65/blob/master/libsrc/cbm/read.s#L11 with a single `.include` statement.

Your header file can't serve that purpose for the reasons given above. You __are__ right in that this use case just isn't in the scope of the PR from your POV. However, you are wrong are __not__ right that it's related to boken vs. working targets.

Anyhow, based on my current understanding I'm in general willing to merge this PR. However, please rename the file to `cbm_kernal.inc`.

And please provide a PR for (an adjusted) https://github.com/ops/cc65/commit/f7fbac4c6b409182a09b1c1ac493d04d8e1d6fe8.

ops:
> I think the point is that I (and Greg ?) would have loved to see a header file that can be used to replace > e.g. the line
>
> .import         SETLFS, OPEN, CHKIN, BASIN, CLRCH, BSOUT, READST

I see. I was only thinking about pure asm stuff and above is more or less library related. Maybe I can work on this as I get more familiar with the libs.

> Your header file can't serve that purpose for the reasons given above. You are right in that this use
> case just isn't in the scope of the PR from your POV. However, you are wrong are not right that it's
> related to boken vs. working targets.

My point was that this PR just adds and cleans something. It does not fix or broke anything.

I'll prepare PRs later today. Thanks for taking care of this.

PR #692, opened 2018-06-29
Update library to use cbm_kernal.inc for kernal entry points

ops:
See preliminary discussion about this PR at https://github.com/cc65/cc65/pull/688


libsrc/c128/kernal.s
```diff
@@ -4,6 +4,8 @@
 ; C128 kernal functions
 ;
 
+	.include "cbm_kernal.inc"
```
oliverschmidt: There should be no /t chars whatsoever in the cc65 code base.

ops: Fixed

PR #707, opened 2018-07-25
Quick fix for missing _div() adaptation after 95223be.

IrgendwerA8:
For time being this should fix the issue. Since 4 bytes more are needed, I try to find a better solution later.
I guess I'm free to change the order of the members

```
typedef struct {
     int rem, quot;
} div_t;
```
if needed?


greg-king5:
> I guess I'm free to change the order of the members.

Yep!

oliverschmidt:

PR #712, opened 2018-08-10
sim65: Fix "$2C: BIT abs" to access the correct address.

ppelleti:
The BIT instruction (opcode 2C) does not work correctly in sim65.  The problem is demonstrated by this example:
```asm
        .export _main
        .import _exit

        .rodata
foo:    .byte $F0
        .code

        .proc _main
        lda #$0F
        bit foo
        beq zero
        lda #1
        ldx #0
        jmp _exit
zero:   lda #2
        ldx #0
        jmp _exit
        .endproc        ; _main
```

The test program should return 2; but, with the current version of sim65, it returns 1.

This pull request fixes sim65 so that the BIT instruction works correctly.

### DIFF
--- a/src/sim65/6502.c
+++ b/src/sim65/6502.c
@@ -739,7 +739,7 @@ static void OPC_6502_2C (void)
     unsigned Addr;
     unsigned char Val;
     Cycles = 4;
-    Addr = MemReadByte (Regs.PC+1);
+    Addr = MemReadWord (Regs.PC+1);
     Val = MemReadByte (Addr);
     SET_SF (Val & 0x80);
     SET_OF (Val & 0x40);


greg-king5:
Welcome to the project.  Thanks for this fix.

The thoughts that you posted on http://forum.6502.org/viewtopic.php?f=2&t=5242 are correct.  Our test suite looks for regressions in the compiler.  We use sim65 during the tests; but, we don't have anything that tests the tester.

spiro-trikaliotis:
> We use sim65 during the tests; but, we don't have anything that tests the tester.

Someone might want to adjust Wolfgang Lorenz test suite? ;)

mrdudz:
Am Samstag, 11. August 2018, 09:02:13 CEST schrieb spiro-trikaliotis:
> > We use sim65 during the tests; but, we don't have anything that tests the
> > tester.
> Someone might want to adjust Wolfgang Lorenz test suite? ;)

it should be doable even without modifying the testsuite - in the vice 
testrepo there already is everything you need. just add the $d7ff debug 
register to sim65 :)

-- 

Programmiersprachen die es einem nicht erlauben sich selber in den Fuss zu 
schiessen sind vor allem für die Programmierer geeignet die Gefahr laufen sich 
bei ihrem Tun selber zu verletzen.

PR #714, opened 2018-08-14
Add CHKOUT and CLRCHN to cbm_kernal.inc

ops:
These names for kernal entry points are used in VIC-20 and C64 Programmer's Reference Guide.


greg-king5:
Those are the spellings that are used in most Commodore guides.

Issue #716, opened 2018-08-16
BRK <arg> is output as opcode $05


fadden:
This:

```
        .setcpu "65816"
        lda #$11
        brk
        nop
        brk $22
        rts
```

is assembled to: `a911 00 ea 0522 60`.  Somehow BRK <arg> turned into ORA dp.

greg-king5:
Fixed by commit d13d068e71fb7cc08734e2c17e67e83f48d28d77.

PR #721, opened 2018-08-19
WDM support

### LINKED ISSUE #715: WDM not supported by ca65
The table of instructions in src/ca65/instr.c doesn't have WDM in it.  This causes assembly to fail when it encounters something like "WDM $00":

    Checkcc.S(3): Error: `:' expected
    Checkcc.S(3): Error: Unexpected trailing garbage characters

The disassembler sources know about WDM, treating it similarly to COP.  src/da65/opc65816.c has:

```
    {   "cop",  2,  flNone,                   OH_Implicit              }, /* $02 */
   ....
    {   "wdm",  2,  flNone,                   OH_Implicit              }, /* $42 */
```

I would argue that the assembler should treat it the same way.

polluks:
Fixes  #715.

### DIFF
--- a/src/ca65/instr.c
+++ b/src/ca65/instr.c
@@ -659,7 +659,7 @@ static const struct {
         { "CLI",  0x0000001, 0x58, 0, PutAll },
         { "CLV",  0x0000001, 0xb8, 0, PutAll },
         { "CMP",  0x0b8f6fc, 0xc0, 0, PutAll },
-        { "COP",  0x0000004, 0x02, 0, PutAll },
+        { "COP",  0x0000004, 0x02, 6, PutAll },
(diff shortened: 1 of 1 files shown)


src/ca65/instr.c
```diff
@@ -659,7 +659,7 @@ static const struct {
         { "CLI",  0x0000001, 0x58, 0, PutAll },
         { "CLV",  0x0000001, 0xb8, 0, PutAll },
         { "CMP",  0x0b8f6fc, 0xc0, 0, PutAll },
-        { "COP",  0x0000004, 0x02, 6, PutAll },
+        { "COP",  0x0000004, 0x02, 0, PutAll },
```
greg-king5: That change will cause a bug that's similar to issue #716.

PR #723, opened 2018-08-19
Fix ftell() on Apple II to return the correct value.

### LINKED ISSUE #722: Apple II: ftell() returns 512 at beginning of file
I am finding that immediately after I open a file for reading, `ftell()` returns `512` when I would expect it to return `0`.

This is on an unenhanced Apple //e running ProDOS 1.9.

Here is the test program I am using:

```
/* cl65 -W-unused-param -t apple2 -C apple2-system.cfg -o ftell.system test-ftell.c */

#include <stdio.h>
#include <conio.h>

int main (int argc, char **argv) {
    const char *fname = "/ram/foo";
    FILE *f;
    long pos;

    f = fopen (fname, "w");
    if (!f)
        goto err;

    fprintf (f, "Hello, World!");
    fclose (f);

    f = fopen (fname, "r");
    if (!f)
        goto err;

    pos = ftell (f);
    printf ("At offset %lu\n", pos);

    fclose (f);
    cgetc ();
    return 0;

 err:
    printf ("Couldn't open %s\n", fname);
    cgetc ();
    return 1;
}
```

This program prints `AT OFFSET 512`, when I would expect it to print `AT OFFSET 0`.

ppelleti:
Fixes #722: `ftell()` returns the value returned by `lseek()`, and `lseek()` for the
Apple II wasn't returning a value.

### DIFF
--- a/libsrc/apple2/lseek.s
+++ b/libsrc/apple2/lseek.s
@@ -22,6 +22,7 @@ _lseek:
         jsr     popptr1
         jsr     popax
         sta     ptr2
+        stx     ptr2+1
 
         ; Get and process fd
         jsr     popax
@@ -78,6 +79,9 @@ seek_common:
         tya
         adc     ptr2
         sta     mliparam + MLI::MARK::POSITION+2
+        lda     #$00
+        adc     ptr2+1
+        bne     einval          ; less than 0 or greater than 2^24 - 1
 
         ; Set file pointer
         lda     #SET_MARK_CALL
@@ -103,13 +107,13 @@ seek_common:
 einval: lda     #EINVAL
 
         ; Set __errno
-errno:  ldx     #$FF
-        stx     sreg
+errno:  jsr     __directerrno   ; leaves -1 in AX
+        stx     sreg            ; extend return value to 32 bits
         stx     sreg+1
-        jmp     __directerrno
+        rts
 
         ; Set __oserror
-oserr:  ldx     #$FF
-        stx     sreg
+oserr:  jsr     __mappederrno   ; leaves -1 in AX
+        stx     sreg            ; extend return value to 32 bits
         stx     sreg+1
-        jmp     __mappederrno
+        rts
--- a/libsrc/apple2/lseek.s
+++ b/libsrc/apple2/lseek.s
@@ -84,6 +84,14 @@ seek_common:
         jsr     callmli
         bcs     oserr
 
+        ; Need to return the position in EAX
+        lda     #0
+        sta     sreg+1
+        lda     mliparam + MLI::MARK::POSITION+2
+        sta     sreg
+        ldx     mliparam + MLI::MARK::POSITION+1
+        lda     mliparam + MLI::MARK::POSITION
+
         rts
 
         ; Load errno code
--- a/libsrc/apple2/lseek.s
+++ b/libsrc/apple2/lseek.s
@@ -6,6 +6,7 @@
 
         .export         _lseek
         .import         popax, popptr1
+        .macpack        cpu
 
         .include        "zeropage.inc"
         .include        "errno.inc"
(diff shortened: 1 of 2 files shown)


ppelleti:
> How about an .ifdef __APPLE2ENH__ and stz?

OK, I've added that.

oliverschmidt:
> How about an .ifdef __APPLE2ENH__ and stz?

Using stz isn't primarily an attribute of the Apple II model, rather it's an attribute of the CPU inside the model. There's a predominant way to do that in the cc65 code base. E.g. like here: https://github.com/cc65/cc65/blob/master/libsrc/runtime/land.s

oliverschmidt:
__directerrno and __mappederrno return -1 as an 16 bit value. Here we want to return -1 as 32 bit value. So we need to set sreg/sreg+1 to $FF before jumping to them.

libsrc/apple2/lseek.s
```diff
@@ -85,8 +85,12 @@ seek_common:
         bcs     oserr
 
         ; Need to return the position in EAX
+        .ifdef  __APPLE2ENH__
+        stz     sreg+1
+        .else
         lda     #0
         sta     sreg+1
+        .endif
```
oliverschmidt: The Apple2 specific code uses `#$00` istead of `#0`. Please adjust the line above.

ppelleti:
OK, I believe I've fixed those three items.

oliverschmidt:
> OK, I believe I've fixed those three items.

Yes, you did - thanks for the very quick reaction :-)

Not directly related but maybe worth changing (as part of this pull request ?): The code as it stands ignores bits 25-32 of the offset parameter. Given that the code explicitly checks that whence is valid would it be consistent to check that bits 25-32 of the offset are actually zero - and just use the same exit used with a wrong whence when they don't?

ppelleti:
> would it be consistent to check that bits 25-32 of the offset are actually zero

I think that we would need to check that bits 25-32 are a sign extension of bit 24, since offset could be negative.

ppelleti:
I think what we need to do is check that bits 24-31 are 0, *after* adding `offset` to the 24-bit quantity in A, X, and Y in `seek_common`.

Sorry, I haven't been thinking very clearly at this time of night.  :)

oliverschmidt:
> [...] check that bits 24-31 are 0, after adding offset to the 24-bit quantity in A, X, and Y in seek_common.

So just saving X to ptr2+1 at the top and change the 3-byte-addition into a 4-byte-addition using ptr2+1 and afterwards check the 4th byte to be zero, correct?

That seems reasonable to me. Are you willing to implement and test that approach?

libsrc/apple2/lseek.s
```diff
@@ -84,13 +85,31 @@ seek_common:
...
+        stx     sreg
+        stx     sreg+1
+        jmp     __directerrno
 
         ; Set __oserror
-oserr:  jmp     __mappederrno
+oserr:  ldx     #$FF
+        stx     sreg
+        stx     sreg+1
+        jmp     __mappederrno
```
greg-king5: A slightly better way:
```asm
oserr:  jsr     __mappederrno
        stx     sreg
        stx     sreg+1
        rts
```
It saves a byte.

ppelleti:
> That seems reasonable to me. Are you willing to implement and test that approach?

I've gone ahead and implemented that, and I implemented @greg-king5's suggestion to save a byte on each of the two error paths.

I'll let you know when I've finished testing.

ppelleti:
OK, I verified that `testcode/lib/seek.c` passes on my unenhanced Apple //e running ProDOS 1.9.  I also added some improvements to the test.

oliverschmidt:
Thanks for the contribution - and the feedback :-)

PR #727, opened 2018-08-20
CMOS optimisation 3rd

polluks:
"The decimal flag D is cleared" [...]
http://shu.emuunlim.com/download/pcedocs/pce_cpu.html


libsrc/pce/crt0.s
```diff
@@ -39,7 +39,6 @@ start:
         nop
         csh                     ; Set high speed CPU mode
         nop
-        cld
         nop
```
greg-king5: Remove the `nop` too.

PR #740, opened 2018-08-26
Add segment type "overwrite".

laubzega:
This patch adds a new segment type - the overlay. It is useful in situations where disassembled 3rd party code is used as a foundation that is subsequently patched by your code, yet you want to avoid modifying the disassembled source itself (because you expect it to change as you progress with reverse engineering, or maybe due to updates being released by said 3rd party).

Example use:
```ini
MEMORY {
    ROM: file = %O, start = $8000, size = $8000;
}

SEGMENTS {
    CODE: load = ROM, type = ro;
    MAXWEAPON: load = ROM, start=$8342, type = overlay;
    NODEATH: load = ROM, offset=$0500, type = overlay;
}
```

### DIFF
--- a/doc/ld65.sgml
+++ b/doc/ld65.sgml
@@ -655,8 +655,8 @@ SEGMENTS {
 </verb></tscreen>
 
 Segment named ORIGINAL contains the original code, disassembled or provided in
-a binary form (i.e. using <tt><ref id=".INCBIN" name=".incbin"></tt>
-directive).  Subsequent four segments will be relocated to addresses specified
+a binary form (i.e. using <tt/.INCBIN/ directive; see the <tt/ca65/ assembler
+document).  Subsequent four segments will be relocated to addresses specified
 by their "start" attributes ("offset" can also be used) and then will overwrite
 whatever was at these locations in the ORIGINAL segment. In the end, resulting
 binary output file will thus contain original data with the exception of four
--- a/doc/ld65.sgml
+++ b/doc/ld65.sgml
@@ -519,7 +519,7 @@ specify a segment attribute. There are five possible segment attributes:
         rw          means read/write
         bss         means that this is an uninitialized segment
         zp          a zeropage segment
-        overwrite   a segment that overwrites another one
+        overwrite   a segment that overwrites (parts of) another one
 
 </verb></tscreen>
 
@@ -635,8 +635,8 @@ Here is an example:
 
 <tscreen><verb>
 MEMORY {
-    RAM: file = "", start = $6000, size = $2000, type=bss;
-    ROM:  file = %O, start = $8000, size = $8000, type=ro;
+    RAM: file = "", start = $6000, size = $2000, type=rw;
+    ROM: file = %O, start = $8000, size = $8000, type=ro;
 }
 </verb></tscreen>
 
@@ -655,14 +655,15 @@ SEGMENTS {
 </verb></tscreen>
 
 Segment named ORIGINAL contains the original code, disassembled or provided in
-a binary form. Subsequent four segments will be relocated to addresses
-specified by their "start" attributes ("offset" can also be used) and then will
-overwrite whatever was at these locations in the ORIGINAL segment. In the end,
-resulting binary output file will thus contain original data with the exception
-of four sequences starting at $9000, $f7e8, $8000 and $e5b7, which will sport
-code from their respective segments. How long these sequences will be depends
-on the lengths of corresponding segments - they can even overlap, so think what
-you're doing.
+a binary form (i.e. using <tt><ref id=".INCBIN" name=".incbin"></tt>
+directive).  Subsequent four segments will be relocated to addresses specified
+by their "start" attributes ("offset" can also be used) and then will overwrite
+whatever was at these locations in the ORIGINAL segment. In the end, resulting
+binary output file will thus contain original data with the exception of four
+sequences starting at $9000, $f7e8, $8000 and $e5b7, which will sport code from
+their respective segments. How long these sequences will be depends on the
+lengths of corresponding segments - they can even overlap, so think what you're
+doing.
 
 Finally, note that OVERWRITE segments should be the final segments loaded to a
 particular memory area, and that they need at least one of "start" or "offset"
--- a/doc/ld65.sgml
+++ b/doc/ld65.sgml
@@ -512,13 +512,15 @@ What we are doing here is telling the linker, that all segments go into the
 the linker will first write the <tt/CODE/ segment, then the <tt/RODATA/
 segment, then the <tt/DATA/ segment - but it will not write the <tt/BSS/
 segment. Why? Here enters the segment type: For each segment specified, you may also
-specify a segment attribute. There are four possible segment attributes:
+specify a segment attribute. There are five possible segment attributes:
 
 <tscreen><verb>
-        ro      means readonly
-        rw      means read/write
-        bss     means that this is an uninitialized segment
-        zp      a zeropage segment
+        ro          means readonly
+        rw          means read/write
+        bss         means that this is an uninitialized segment
+        zp          a zeropage segment
+        overwrite   a segment that overwrites another one
+
 </verb></tscreen>
 
 So, because we specified that the segment with the name BSS is of type bss,
@@ -617,6 +619,54 @@ would write two files that start with the name of the output file specified on
 the command line, with "-1.bin" and "-2.bin" appended respectively. Because
 '%' is used as an escape char, the sequence "%%" has to be used if a single
 percent sign is required.
+
+<sect1>OVERWRITE segments<p>
+
+There are situations when you may wish to overwrite some part (or parts) of a
+segment with another one. Perhaps you are modifying an OS ROM that has its
+public subroutines at fixed, well-known addresses, and you want to prevent them
+from shifting to other locations in memory if your changed code takes less
+space. Or you are updating a block of code available in binary-only form with
+fixes that are scattered in various places. Generally, whenever you want to
+minimize disturbance to an existing code brought on by your updates, OVERWRITE
+segments are worth considering.
+
+Here is an example:
+
+<tscreen><verb>
+MEMORY {
+    RAM: file = "", start = $6000, size = $2000, type=bss;
+    ROM:  file = %O, start = $8000, size = $8000, type=ro;
+}
+</verb></tscreen>
+
+Nothing unusual so far, just two memory blocks - one RAM, one ROM. Now let's
+look at the segment configuration:
+
+<tscreen><verb>
+SEGMENTS {
+    RAM:       load = RAM, type = bss;
+    ORIGINAL:  load = ROM, type = ro;
+    FASTCOPY:  load = ROM, start=$9000, type = overwrite;
+    JMPPATCH1: load = ROM, start=$f7e8, type = overwrite;
+    DEBUG:     load = ROM, start=$8000, type = overwrite;
+    VERSION:   load = ROM, start=$e5b7, type = overwrite;
+}
+</verb></tscreen>
+
+Segment named ORIGINAL contains the original code, disassembled or provided in
+a binary form. Subsequent four segments will be relocated to addresses
+specified by their "start" attributes ("offset" can also be used) and then will
+overwrite whatever was at these locations in the ORIGINAL segment. In the end,
+resulting binary output file will thus contain original data with the exception
+of four sequences starting at $9000, $f7e8, $8000 and $e5b7, which will sport
+code from their respective segments. How long these sequences will be depends
+on the lengths of corresponding segments - they can even overlap, so think what
+you're doing.
+
+Finally, note that OVERWRITE segments should be the final segments loaded to a
+particular memory area, and that they need at least one of "start" or "offset"
+attributes specified.
 
 <sect1>LOAD and RUN addresses (ROMable code)<p>
 
--- a/src/ld65/bin.c
+++ b/src/ld65/bin.c
@@ -193,8 +193,8 @@ static void BinWriteMem (BinDesc* D, MemoryArea* M)
                     NewAddr += M->Start;
                 }
                 if (DoWrite || (M->Flags & MF_FILL) != 0) {
-                    /* Seek back for "overlay" segments */
-                    if (NewAddr < Addr) {
+                    /* Seek in "overlay" segments */
+                    if (S->Flags & SF_OVERLAY) {
                         fseek(D->F, NewAddr - M->Start, SEEK_SET);
                     } else {
                         WriteMult (D->F, M->FillVal, NewAddr-Addr);
--- a/src/ld65/bin.c
+++ b/src/ld65/bin.c
@@ -193,8 +193,13 @@ static void BinWriteMem (BinDesc* D, MemoryArea* M)
                     NewAddr += M->Start;
                 }
                 if (DoWrite || (M->Flags & MF_FILL) != 0) {
-                    WriteMult (D->F, M->FillVal, NewAddr-Addr);
-                    PrintNumVal ("SF_OFFSET", NewAddr - Addr);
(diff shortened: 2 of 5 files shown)


oliverschmidt:
First of all thanks for your interest in contributing to cc65!

I must admit that I don‘t like this pull request:

* I think the usecase is (way) too specific to be supported by the (upstream) cc65 tool chain.

* I think the segment type isn‘t a good place to express what you want to express. Rather I see the type obused - as it was the simplest way to do it. This looks like a hack to me.

However, if you can create by one means or another a momentum in the cc65 community active here or the dev mailing list to ask for the feature then I‘m willing to think about a good way to express it.

silverdr:
This PR is a direct response to the problem I wrote about recently. I was looking (admittedly not too deep) for an existing way to do overlays during link time and a friend of mine seem to have made the way to do it. If I may vote for finding a good way to have this kind of functionality available, then please count it. I see (and have) numerous projects, where this would be very useful.

oliverschmidt:
> This PR is a direct response to the problem I wrote about recently.

For everybody's reference:
https://sourceforge.net/p/cc65/mailman/message/36398448/

>  I was looking (admittedly not too deep) for an existing way to do overlays during link time [...]

I'm personally not convinced that a linker is in general an appropriate tool to do what (I understood) you want to do.

spiro-trikaliotis:
@oliverschmidt, why shouldn't the linker be the right tool?
I am looking, for example, at the C64 KERNAL or the floppy firmwares. There, Commodore often patched functions by adding a JMP PATCH_SOMETHING ... instead of a STA $xxxx (for example), and then having something like

PATCH_SOMETHING: 
    ... do something more ...
   STA $xxxx
   JMP PATCH_RETURN

Unfortunately, the PATCH_SOMETHING are not consistent in all versions of the ROM. For example, for the 1541 floppy ROMs (the early one for the C64), they were in the $E000-$FFFF area (because that is the only ROM area that was changed), while in the 1570 and 1571 floppy (the one of the C128), they were in the $8000-$BFFF area (which was not present in the 1541 yet).

This patch would make my work *much* more easier in expressing this behaviour. For me, I would upvote it specifically.

I am still struggling if there is an already existing way to do it, but so far, I have not found it.

silverdr:
> I'm personally not convinced that a linker is in general an appropriate tool to do what (I understood) you want to do.

At the moment I don't see any more appropriate one. All in all this is the tool that takes compiled/assembled object files and outputs the final product (binary in this case).

silverdr:
@spiro-trikaliotis you just described a very similar case to what I have at hand! We are writing patches that need to be applied on top of the existing firmware in various versions. And they are supposed to be self-cotained "binary" patches so that nothing vital of the existing code changes its position and we can position the patches around the original codebase easily. Also we need to be able to build a preselected set of binaries out of the sources (including selected, self-contained patches) preferably again without resorting to source-based techniques with ever-expanding and quickly growing beyond what is reasonable set of `#IFs`. While this is not exactly the case for the project at hand, there can be other situations where the baseline source code changes are to be avoided. Having an easy way to handle this kind of overlays, without sacrificing anything from the existing functionality is a win-win IMHO.

oliverschmidt:
We need to decide where we want this discussion to happen. I suggest the cc65 dev mailing list. So please repost there what you consider relevant.

oliverschmidt:
> I think the usecase is (way) too specific to be supported by the (upstream) cc65 tool chain.

Okay, I understand by now that this seems to be a viable use case.

> I think the segment type isn‘t a good place to express what you want to express. Rather I see the type obused - as it was the simplest way to do it. This looks like a hack to me.

After giving it quite some thought I can't come up with a better approach.

So I'm generally willing to merge this pull request. However:

* I think the term 'overlay' doesn't fit well. And even if I personally would see/agree/understand I still see cc65 users having totally wrong ideas when they read something about 'overlay segments'. I propose the term 'replace' instead: "It's a replace segment, it replaces something already present at that address."

* There needs to be documentation. The ld65 documentation works a lot with small examples (e.g. in http://cc65.github.io/doc/ld65.html#ss5.1) so I think an explanation with an example would fit in well.

src/ld65/bin.c
```diff
@@ -193,8 +193,13 @@ static void BinWriteMem (BinDesc* D, MemoryArea* M)
                     NewAddr += M->Start;
                 }
                 if (DoWrite || (M->Flags & MF_FILL) != 0) {
-                    WriteMult (D->F, M->FillVal, NewAddr-Addr);
-                    PrintNumVal ("SF_OFFSET", NewAddr - Addr);
+                    /* Seek in "overlay" segments */
+                    if (S->Flags & SF_OVERLAY) {
+                        fseek(D->F, NewAddr - M->Start, SEEK_SET);
```
oliverschmidt: Blank before parameter list is missing.

src/ld65/config.c
```diff
@@ -757,6 +758,7 @@ static void ParseSegments (void)
                         case CFGTOK_RW:    /* Default */                    break;
                         case CFGTOK_BSS:   S->Flags |= SF_BSS;              break;
                         case CFGTOK_ZP:    S->Flags |= (SF_BSS | SF_ZP);    break;
+                        case CFGTOK_OVERLAY: S->Flags |= SF_OVERLAY;        break;
```
oliverschmidt: Please align the other four cases above (if still necessary after the rename).

src/ld65/config.c
```diff
@@ -1848,6 +1851,29 @@ unsigned CfgProcess (void)
...
+            ** segment types following them in current memory region.
+            */
+            if (S->Flags & SF_OVERLAY) {
+            {
+                if (S->Flags & (SF_OFFSET | SF_START)) {
+                    ++Overlays;
+                } else {
+                    CfgError (GetSourcePos (M->LI),
+                              "Segment `%s' of type `overlay' requires either"
+                              " `Start' or `Offset' argument to be specified.",
```
oliverschmidt: No period at the end of the message string.

src/ld65/config.c
```diff
@@ -1896,22 +1922,33 @@ unsigned CfgProcess (void)
...
-                        if (S->Flags & SF_OFFSET) {
-                            CfgWarning (GetSourcePos (S->LI),
-                                        "Segment `%s' offset is too small in `%s' by %lu byte%c",
-                                        GetString (S->Name), GetString (M->Name),
-                                        Addr - NewAddr, (Addr - NewAddr == 1) ? ' ' : 's');
+
+                    if (S->Flags & SF_OVERLAY) {
+                        if (NewAddr < M->Start) {
+                            CfgError (GetSourcePos (S->LI),
+                                      "Segment `%s' begins before memory area `%s'.",
```
oliverschmidt: No period at the end of the message string.

src/ld65/config.c
greg-king5: Also, please say that they are an `attribute` instead of an `argument`.

src/ld65/config.c
greg-king5: The constant needs to be `(SF_RO | SF_REPLACE)`.  It will avoid an error message if the memory area is marked as Read-Only.

laubzega:
Hi Oliver and greg-king5,

Thank you for the review and suggestions. I implemented the changes you requested, and while I am not too happy about the name change, I agree that OVERLAY is not a great fit either. I will get to work on the docs now, and if in the meantime something better pops up in my mind, I will let you know.

src/ld65/config.c
```diff
@@ -1854,17 +1854,15 @@ unsigned CfgProcess (void)
...
             ** segment types following them in current memory region.
             */
-            if (S->Flags & SF_OVERLAY) {
-            {
+            if (S->Flags & SF_REPLACE) {
                 if (S->Flags & (SF_OFFSET | SF_START)) {
                     ++Overlays;
                 } else {
                     CfgError (GetSourcePos (M->LI),
                               "Segment `%s' of type `overlay' requires either"
```
silverdr: It is no longer type 'overlay'.

src/ld65/bin.c
```diff
@@ -193,8 +193,13 @@ static void BinWriteMem (BinDesc* D, MemoryArea* M)
                     NewAddr += M->Start;
                 }
                 if (DoWrite || (M->Flags & MF_FILL) != 0) {
-                    WriteMult (D->F, M->FillVal, NewAddr-Addr);
-                    PrintNumVal ("SF_OFFSET", NewAddr - Addr);
+                    /* Seek in "overlay" segments */
```
silverdr: No longer 'overlay'..

src/ld65/config.c
```diff
@@ -1795,6 +1797,7 @@ unsigned CfgProcess (void)
     for (I = 0; I < CollCount (&MemoryAreas); ++I) {
         unsigned J;
         unsigned long Addr;
+        unsigned Overlays = 0;
```
silverdr: Shouldn't we adjust the name of the label too?

src/ld65/config.c
```diff
@@ -1848,6 +1851,27 @@ unsigned CfgProcess (void)
...
 
+            /* Take note of overlayed segments and make sure there are no other
+            ** segment types following them in current memory region.
+            */
+            if (S->Flags & SF_REPLACE) {
+                if (S->Flags & (SF_OFFSET | SF_START)) {
+                    ++Overlays;
+                } else {
+                    CfgError (GetSourcePos (M->LI),
+                              "Segment `%s' of type `overlay' requires either"
```
silverdr: No longer type 'overlay'

oliverschmidt:
> Thank you for the review and suggestions.

And thank you for your continuous interest in this pull request given my initial negative feedback ;-)

>>  I propose the term 'replace' instead.

> [...] and while I am not too happy about the name change, I agree that OVERLAY is not a great fit either. I will get to work on the docs now, and if in the meantime something better pops up in my mind, I will let you know.

Please do so. As I wrote above I see 'replace' as proposal to avoid the term 'overlay' which is already overlaid with a different meaning (pun intended).

silverdr:
@oliverschmidt I am truly glad to read that. As for the name - full ACK that 'overlay' could potentially be confusing :-) "Replace" OTOH suggests more like a full replacement (of another segment), which - as we know - doesn't have to be the case. So if something better comes up as the type name - great! Maybe something along "overwrite" line?

oliverschmidt:
Given that we have read-only and read/write segments I personally think that anything having 'write' in the name even more leads to the assumption that the segment behaves more like other segments (incl. that full replacement aspect you brought up).

groessler:
I looked at a dictionary for English words for "überlagern", and got "superpose" and "superimpose" in addition to "overlay".

silverdr:
@oliverschmidt yeah.. admittedly I am not convinced to the "overwrite" either. Therefore if something better comes up before the PR is merged - great. If not, well.. we'll live with "replace".

silverdr:
@groessler - hm... "superimpose" seems rather lengthy but "superpose" is more reasonable in terms of length and means: "place (something) on or above something else, especially so that they coincide", which would fit well.

cacciatc:
I've been following this PR and am looking forward to the new segment! Thanks to everyone's work so far. My two cents for naming:

* I would vote no for "superimpose" since I believe that denotes that the thing underneath still somehow shows through.
* Why not "__patch__"? The word has been used a couple times this convo in describing the use case.

silverdr:
@cacciatc - "superimpose" no. "superpose" seems the best so far to me. I believe we don't have to decide today in any case so maybe something still better shows up.

laubzega:
My vote would be on "overwrite" or "patch", if only for the reason that they both exactly reflect what is being done, while "Replace" indeed suggests that something is removed entirely, and "superimpose" that things are still visible from underneath. "Superpose" is too contrived, IMHO. Never seen this word in the context of compilation/linking (or software engineering for that matter), so I'm not sure anyone will get the meaning without reading the docs. Both "overwrite" or "patch" are near-obvious.

Where are liberal arts majors when you need them?!

oliverschmidt:
> Both "overwrite" or "patch" are near-obvious.

I agree. However, "patch" is maybe a bit too "laid-back" (sorry, I'm no native speaker). I'm perfectly fine with "overwrite". In fact, that was my second-best idea behind "replace" in the first place.

So if that's okay with you please go for "overwrite".

doc/ld65.sgml
```diff
@@ -618,6 +620,54 @@ the command line, with "-1.bin" and "-2.bin" appended respectively. Because
...
+fixes that are scattered in various places. Generally, whenever you want to
+minimize disturbance to an existing code brought on by your updates, OVERWRITE
+segments are worth considering.
+
+Here is an example:
+
+<tscreen><verb>
+MEMORY {
+    RAM: file = "", start = $6000, size = $2000, type=bss;
+    ROM:  file = %O, start = $8000, size = $8000, type=ro;
```
oliverschmidt: Please align properly with line above.

doc/ld65.sgml
```diff
@@ -618,6 +620,54 @@ the command line, with "-1.bin" and "-2.bin" appended respectively. Because
...
+    ORIGINAL:  load = ROM, type = ro;
+    FASTCOPY:  load = ROM, start=$9000, type = overwrite;
+    JMPPATCH1: load = ROM, start=$f7e8, type = overwrite;
+    DEBUG:     load = ROM, start=$8000, type = overwrite;
+    VERSION:   load = ROM, start=$e5b7, type = overwrite;
+}
+</verb></tscreen>
+
+Segment named ORIGINAL contains the original code, disassembled or provided in
+a binary form. Subsequent four segments will be relocated to addresses
```
oliverschmidt: I think a very brief hint towards .incbin (with a link to it) would be very helpful, especially as you explicitly write "in a binary form".

doc/ld65.sgml
```diff
@@ -512,13 +512,15 @@ What we are doing here is telling the linker, that all segments go into the
...
 <tscreen><verb>
-        ro      means readonly
-        rw      means read/write
-        bss     means that this is an uninitialized segment
-        zp      a zeropage segment
+        ro          means readonly
+        rw          means read/write
+        bss         means that this is an uninitialized segment
+        zp          a zeropage segment
+        overwrite   a segment that overwrites another one
```
oliverschmidt: I think "a segment that overwrites (parts of) another one" is necessary to avoid some head scratching here.

doc/ld65.sgml
```diff
@@ -618,6 +620,54 @@ the command line, with "-1.bin" and "-2.bin" appended respectively. Because
...
+space. Or you are updating a block of code available in binary-only form with
+fixes that are scattered in various places. Generally, whenever you want to
+minimize disturbance to an existing code brought on by your updates, OVERWRITE
+segments are worth considering.
+
+Here is an example:
+
+<tscreen><verb>
+MEMORY {
+    RAM: file = "", start = $6000, size = $2000, type=bss;
```
greg-king5: Memory areas have only two types: `rw` and `ro` (`rw` is the default).

oliverschmidt:
The doc fails to build with `/tmp/linuxdoc-tools.ffTuQfmp4d/sgmltmp.ld65.2.texi:923: @ref reference to nonexistent node '.INCBIN'` - see Travis CI log.

greg-king5:
Linuxdoc-Tools doesn't support automatic cross-linking between documents.  But, it is risky to refer to a specific section of another document; its section numbers might change if text is added to it.  Therefore, it's safer to give a general mention, similar to this:
```html
... a binary form (i.e., using the <tt/.INCBIN/ directive; see the <tt/ca65/
assembler document).
```

oliverschmidt:
> But, it is risky to refer to a specific section of another document; its section numbers might change if text is added to it. 

I'd say the risk actually differs between documents. But the funcref is for sure the one with a too high risk. From that perspective thanks for your input !

oliverschmidt:
To me this looks ready to be merged - right?

laubzega:
I hope so. :)

PR #611, opened 2018-03-11
TGI driver (160x192x2) for the VIC-20.

SvenMichaelKlose:

PR #741, opened 2018-08-27
cbm-lo.tgi: one driver, eight targets

polluks:
This driver uses PETSCII for lores graphics. For example:
```basic
load "tgidemo",8
run:rem cbm-lo.tgi
```
(Sorry for the commit mess, I still get "Permission denied (publickey)" Git errors, so I used the web interface.)


doc/c128.sgml
```diff
@@ -200,15 +200,19 @@ memory drivers using the VDC memory!
 <descrip>
   <tag><tt/c128-vdc.tgi (c128_vdc_tgi)/</tag>
```
oliverschmidt: The predominant style seems to be to have blank lines after `<descrip>` and before `</descrip>`. I see that you follow that style for newly added descriptions. It would be nice if you would additionally modify descriptions you edit anyway to follow that style too.

libsrc/cbm/tgi/cbm-lo.s
```diff
@@ -0,0 +1,456 @@
...
+
+; ------------------------------------------------------------------------
+; SETVIEWPAGE: Set the visible page. Called with the new page in A (0..n).
+; The page number is already checked to be valid by the graphics kernel.
+;
+; Must set an error code: NO (will only be called if page ok)
+;
+
+SETVIEWPAGE:
+;       rts
```
oliverschmidt: My personal opinion was/is that we deliberately keep all entry points to drivers unique. However, recently it came into fashion to fold them together. But that change of style wasn't done throughout the code base, so it's now less consistent then before :-(

Let's see what rolling the dice will result in this driver ;-))

polluks:
tgidemo now uses outtext (welcome "y") because I got this function for free ;-)

libsrc/cbm/tgi/cbm-lo.s
polluks: Did you see my optimisation for DONE? I didn't want to repeat the code.

oliverschmidt:
> load "tgidemo",8 run:rem cbm-lo.tgi

I'm wondering how this can work because as far as I can see the driver you implemented doesn't get built.

I'm generally wondering if it is desirable to extended the already pretty complex libsrc/Makefile to be able to do so.

And even if it would be able I'm wondering if it would be desirable to have a file 'cbm-lo.tgi'. Where would it be placed? In a new directory target/cbm/ ? E.g. https://github.com/cc65/cc65/blob/master/samples/Makefile#L84 doesn't find it there. Maybe users don't find it as well.

So I'm pretty sure it rather desirable to have target-specific drivers. Internally they share the implementation like done at https://github.com/cc65/cc65/blob/master/libsrc/c16/joy/c16-stdjoy.s or at https://github.com/cc65/cc65/blob/master/libsrc/atari/tgi/atr10.s

I think the actual implementation (which is included eight times) should live at /cbm/tgi_lo.inc.

And finally please add the lines like https://github.com/cc65/cc65/blob/master/include/c64.h#L158 to the eight header files in question.

oliverschmidt:
> Indeed, this is the hard part, it doesn't fit in the scheme very well.

Yep.

> ".inc"?

Yes, like https://github.com/cc65/cc65/blob/master/libsrc/tgi/tgidrv_line.inc and https://github.com/cc65/cc65/blob/master/libsrc/atari/tgi/atari_tgi_common.inc

> Assemble once, run everywhere!

Your "everywhere" is just every CBM ;-) My "everywhere" with the Contiki .o65 CS8900A driver is Apple II, Atari and C64 !

oliverschmidt:
This looks nice and complete as far as I can tell :-)

Interesting that after all this pull request (and not https://github.com/cc65/cc65/pull/611) is bringing the first TGI driver to the VIC20 ;-)

There's just one nit-picking thingy left. I wrote:

> I think the actual implementation (which is included eight times) should live at /cbm/tgi_lo.inc.

and

>Yes, like https://github.com/cc65/cc65/blob/master/libsrc/tgi/tgidrv_line.inc and https://github.com/cc65/cc65/blob/master/libsrc/atari/tgi/atari_tgi_common.inc

Now you named the file cbm/cbm-lo.inc (with `-` instead of `_`). I think we pretty consistently use `_` in source file names (and only `-` in "output" file names because `-` is more likely to work in target file systems).

I'm not going to block this pull request from being merged because of this detail. However, if you agree with me and want to adjust the name, I'd appreciate it.

Regards,
Oliver

oliverschmidt:
Anyhow, there seem to be merge conflicts to be solved...

polluks:
I suppose my biggest commit... the code is based on my NES driver of 2009.
The PET kernal makes a little bit trouble because of the table in https://www.pagetable.com/?p=926.

libsrc/cbm/tgi_lo.inc
```diff
@@ -129,6 +135,24 @@ INSTALL:
 UNINSTALL:
         rts
 
+; Runtime PET patch
```
oliverschmidt: Likely I'm missing the point but wouldn't it be more appropriate to just use `.if .def(__PET__)` in SETPIXEL ?

oliverschmidt: As it stands the driver source is assembled 8 times to build 8 drivers. But you consider it important that they are identical instead of saying "if we build the driver 8 times anyway we can as well make them slightly different". You surely understand that this seems somewhat strange to me, don't you ;-)

If it is however important to you to showcase that the same file works for all CBMs I propose you this "compromise": Keep the runtime patch, but .ifdef it with PET. Then you know that you can use the PET driver for all CBMs while the usual user has a smaller driver for the only target he's interested in (apart from the PET).

libsrc/cbm/tgi_lo.inc
```diff
@@ -0,0 +1,488 @@
...
+; ------------------------------------------------------------------------
+; Constants
+
+RVS_ON          = 18
+RVS_OFF         = 18+128
+
+; ------------------------------------------------------------------------
+; Header. Includes jump table and constants.
+
+        module_header   _cbm_lo_tgi
```
greg-king5: Change the label:
`module_header MODULE_LABEL`

greg-king5:
The targets that didn't have a TGI driver before now need `TGI_COLOR_ ...` constants.

greg-king5:
Two drivers, seven targets -- the Plus4 driver cannot work because the firmware is disabled.

oliverschmidt:
> This looks nice and complete as far as I can tell :-)

- librefs missing
- TGI_COLOR_s missing
- MODULE_LABEL missing

I'm too old for this shit (https://www.imdb.com/title/tt0093409/quotes).

> Anyhow, there seem to be merge conflicts to be solved...

Still open...

polluks:
librefs were already added.
I'm using the colors of https://github.com/cc65/cc65/blob/cdc1c2cbd9aeb05373c5176c17afbed5cf634399/include/vic20.h
 to reduce future conflicts.
Apropos conflicts, what I have to do? Travis seems happy.
However Plus/4 is critical...

oliverschmidt:
> librefs were already added.

I know - but only after I wrote that the pull request is complete ;-)

> However Plus/4 is critical...

I can't comment on that. I have no clue what you're talking about.

> Apropos conflicts, what I have to do? Travis seems happy.

image

polluks:
aha https://stackoverflow.com/questions/50012036/github-cant-rebase-my-feature-branch-this-branch-cannot-be-rebased-due-to-con

libsrc/cbm/tgi_lo.inc
```diff
@@ -0,0 +1,488 @@
...
+; Graphics driver for a max. 160x50x2 (but 4 pixels) mode on all CBMs.
+;
+; 2018-08-30, Stefan Haubenthal
+;
+
+        .include        "zeropage.inc"
+
+        .include        "tgi-kernel.inc"
+        .include        "tgi-error.inc"
+        .include        "cbm_kernal.inc"
```
oliverschmidt: I must admit that I don't have the overview here. Is this line to be changed?

And maybe there's a bigger problem behind it:

I seem to understand that you need to call into those Kernal wrapper thingies on the Plus/4. However that code is part of the C library and .o65 based driver can't just call into the C library as the cc65 .o65 loader doesn't do any resolving of "outgoing" symbol references.

The usual two options to solve this is to either duplicate the code in question from the C library into the driver or use the libref mechanism: The cc65 driver loader makes the address of the libref symbol available to the loaded driver. That libref symbol is the same for all drivers of a certain type (TGI, SER, JOY, ...) on a certain target. However, as there's - at least currently - only one TGI driver for the Plus/4 I don't see a problem with this driver making use of the Plus/4 tgi_libref as it sees fit. Technically one would create some jump table like e.g. in https://github.com/cc65/cc65/blob/master/libsrc/atari/serref.s

As there's a dummy tgi_libref present on every target this approach doesn't keep the Plus/4 driver from working on other CBM machines. It just must make sure at runtime to not make use of the jump table.

oliverschmidt:
> I get no rebase error, so what to do?

I don't know. However, we'll find a solution one way or another at the time that's the only thing left.

libsrc/cbm/tgi_lo.inc
polluks: Exactly, I don't want cbm_kernal.inc anymore, so I'll use libref.

oliverschmidt: I see. However, the libref overhead might be larger than a duplication of the functionality in the driver if it's only about hitting some softswitch, doing the Kernal JSR and hitting some other softswitch...

greg-king5:
>> I get no rebase error; so, what to do?
>
> I don't know. However, we'll find a solution one way or another, at the time that it's the only thing left.

It's not a problem.  Two of this PR's commits conflict with my c39d07dd177a23079c9660d65f974861f331e29a commit.  But, the last merge commit fixed them.  Therefore, this PR can be merged automatically (which is what GitHub tells us).

oliverschmidt:
@greg-king5: What aspect of the GitHub web UI tells us that this pull request can be merged automatically? And if that is the case, why isn‘t there a merge button for me to press? And why isn‘t a rebase done by @polluks the appropriate approach?

greg-king5:
I can pull this PR into my local Git repository without any conflicts.  On GitHub,I get this:

image

oliverschmidt:
That's pretty amazing, because I still see the screen below. I suggest that we simply agree that you merge this pull request when it's finalized (that Plus/4 Kernal calls issue).

image

oliverschmidt:
What a stupid UI - putting an activatable „pull down menu“ into a disabled push button !

Anyhow, thanks for your patience to point it out to me until ecen I finally got it ;-)

polluks:
Is it possible to add LOWCODE to module.cfg?

oliverschmidt:
Maybe I misunderstand you but a module is in general intended to be loaded by a loader. So the loader decides on the placement in memory. What would LOWCODE express in that situation?

polluks:
@oliverschmidt The bankswitching of the plus4 requires a lowcode segment in the module which is not part of the config right now.

oliverschmidt:
As explained there can't be no LOWCODE segment in the module config for conceptual reasons. If you need to execute code in the LOWCODE segment from your driver then the libref approach is the only (reasonable) approach.

greg-king5:
- The CBM510, CBM610, C128, C16, and Plus4 models insert a line on the screen when a character is printed at the right edge -- even while a program is running!  Therefore, their drivers must not plot in the entire right edge column (not only the bottom right).
- The current PET patch can't work because the PLOT function also must calculate a pointer into the screen RAM.  https://github.com/cc65/cc65/blob/master/libsrc/pet/cputc.s#L78-L91

polluks:
@greg-king5 Thanks for testing, I will check it out.
My patching is not reasonable if we need a special plus4 version anyway.

libsrc/cbm/tgi_lo.inc
```diff
@@ -21,22 +23,28 @@ PLOT           := $FFF0
...
+.endif
+
 RVS_ON          = 18
 RVS_OFF         = 18+128
 
 ; ------------------------------------------------------------------------
 ; Header. Includes jump table and constants.
 
-        module_header   MODULE_LABEL
+        module_header   _c16_lo_tgi;MODULE_LABEL
```
shaubenthal: oops

polluks:
Travis fails? Here C128 works.
c128

greg-king5:
Dynamic drivers aren't linked to a library.  `.import` can't work -- unless the object files that export those names are linked explicitly to the driver.  See https://github.com/cc65/cc65/blob/master/libsrc/nes/Makefile.inc

The "libref" method will avoid that problem.  The references will be in each program; they will be linked to a library.

libsrc/plus4/Makefile.inc
```diff
@@ -0,0 +1,3 @@
+../target/plus4/drv/tgi/plus4-lo.tgi:   \
+                ../libwrk/plus4/kplot.o \
+                ../libwrk/plus4/kbsout.o
```
oliverschmidt: There seems to be a misunderstanding:

There are two mutually exclusive ways to get C library functionality into a driver: **Either** duplicate it in the driver **or** reference it.

* This commit has the title 'Prepare Plus/4 libref.' but actually it prepares duplication.
* As discussed duplication will never place the duplicated code in LOWCODE.

libsrc/pet/kernal.s
```diff
@@ -6,11 +6,10 @@
 
         .include "cbm_kernal.inc"
 
-        .export         CHKIN
-        .export         CKOUT, CHKOUT
+        .export         CHKOUT
         .export         CLRCH, CLRCHN
         .export         BASIN, CHRIN
-        .export         BSOUT, CHROUT
+        .export         CHROUT
```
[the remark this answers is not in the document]
mrdudz: time to make a regression test for it :)

greg-king5:
I suggest that you use libref for all CBM targets.  It will avoid a lot of `.IFDEFs`.
`PLOT` and `BSOUT` can be addressed directly.  `SCREEN` can point to the conio function `screensize`.
```asm
        .export tgi_libref

        .import screensize, PLOT, BSOUT

tgi_libref:
        .addr   screensize, PLOT, BSOUT
```

polluks:
I'm a bit lost in complexity, there should be three drivers: pet, plus4 and other  cbm.
How does one libref deliver two functions?

oliverschmidt:
See my comment further above:

> However, as there's - at least currently - only one TGI driver for the Plus/4 I don't see a problem with this driver making use of the Plus/4 tgi_libref as it sees fit. Technically one would create some jump table like e.g. in https://github.com/cc65/cc65/blob/master/libsrc/atari/serref.s

oliverschmidt:
This PR now appears to me to want to change in 50 files things not related at all to a TGI driver. Am I doing something wrong or is something wrong with this PR?

asminc/cbm_kernal.inc
```diff
@@ -1,7 +1,7 @@
 ;
 ; Olli Savia <ops@iki.fi>
 ;
-; Commodore kernal functions
+; Commodore kernal functions http://web.archive.org/web/20180720093634/https://www.pagetable.com/?p=926
```
silverdr: FWIW - while I hold the author of "pagetable" in high esteem, I don't see putting such links here as something adding value.

polluks: I know what you mean but you won't find a more complete list.

polluks:
> This PR now appears to me to want to change in 50 files things not related at all to a TGI driver. Am I doing something wrong or is something wrong with this PR?

I did a rebase, not a good idea :-(

oliverschmidt:
@polluks: Is there any momentum to have this PR brought into a mergable state (see our discussions above)? Otherwise I'm tempted to close it...

polluks:
This PR was overambitious for me, it would conflict with vic20-hi for example. Maybe I'll give it a second try later.

oliverschmidt:
> Maybe I'll give it a second try later.

:-)

PR #742, opened 2018-08-28
static driver

polluks:


oliverschmidt:
Maybe a misunderstanding? My proposal was/is that the eight drivers have their traditional names, e.g. c64-lo.

polluks:
I see. Maybe the name should depend who builds the driver?

oliverschmidt:
Yep, therefore me referencing the c16/Plus4 joy driver sharing and the multiple Atari tgi driver sharing. There all that happens.

polluks:
You cannot move/delete files in the browser... maybe I have to compile a git with TLS 1.2 myself.

oliverschmidt:
> maybe I have to compile a git with TLS 1.2 myself.

Given the role GitHub plays in open source project hosting by now I'd say that any effort allowing you to fully participate is justified.

PR #745, opened 2018-09-02
Atari: fix interrupt handling if extended memory is banked in

groessler:
Extendend memory is mapped over the main memory in the 0x4000..0x7FFF
area. Many DOSes disable interrupts while extended memory is banked in,
but not all (e.g. SpartaDOS-X).
This change modifies the initial interrupt handler to map in main memory
before chaining to the "worker" handlers.
Since the initial interrupt handler uses a data segment to store the
trampoline to chain to the original handler, introduce a new "LOWBSS"
segment to hold this trampoline. Otherwise the trampoline may end up
inside the 0x4000..0x7FFF area.

Add a link time warning if "LOWCODE" segment lays within the extended
memory window.

### DIFF
--- a/cfg/atari-cart.cfg
+++ b/cfg/atari-cart.cfg
@@ -10,25 +10,24 @@ SYMBOLS {
     __CARTFLAGS__:       type = weak,   value = $01;   # see documentation for other possible values
 }
 MEMORY {
-    ZP:      file = "", define = yes, start = $0082,                size = $007E;
-    MAIN:    file = "", define = yes, start = %S,                   size = __CARTSIZE__;
-    ROM:     file = %O, define = yes, start = $C000 - __CARTSIZE__, size = __CARTSIZE__ - 6, fill = yes, fillval = $FF;
-    CARTID:  file = %O,               start = $BFFA,                size = $0006;
-    DISCARD: file = "",               start = $0100,                size = $FF00;
+    ZP:     file = "", define = yes, start = $0082,                size = $007E;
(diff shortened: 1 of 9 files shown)


cfg/atari-cart.cfg
```diff
@@ -10,23 +10,25 @@ SYMBOLS {
...
-    CODE:     load = ROM,             type = ro,  define = yes;
-    RODATA:   load = ROM,             type = ro,                optional = yes;
-    DATA:     load = ROM, run = MAIN, type = rw,  define = yes, optional = yes;
-    INIT:     load = MAIN,            type = bss,               optional = yes;
-    BSS:      load = MAIN,            type = bss, define = yes, optional = yes;
-    CARTHDR:  load = CARTID,          type = ro;
+    ZEROPAGE: load = ZP,                  type = zp,                optional = yes;
+    EXTZP:    load = ZP,                  type = zp,                optional = yes;
+    STARTUP:  load = ROM,                 type = ro,  define = yes, optional = yes;
+    LOWBSS:   load = DISCARD, run = MAIN, type = rw,                optional = yes;  # not zero-initialized
```
greg-king5: The `DISCARD:` memory area isn't needed.
`load = MAIN, type =` has the same effect that `load = DISCARD, run = MAIN, type =` has.

That `load = MAIN,` means `load = MAIN, run = MAIN,`.
The "load" parts have no effect (because of `file = ""`); the "run" parts are the only things that matter there (they are the same).

oliverschmidt:
As far as I understand this is ready to be merged...

PR #749, opened 2018-09-05
Add LGTM code quality badges

xcorail:
Hi there!

I thought you might be interested in adding these [code quality badges](https://lgtm.com/projects/g/cc65/cc65/ci/#badges) to your project. They will indicate a very high code quality to potential users and contributors.
To get an idea of the analyses reflected by these grades, check the alerts discovered by [LGTM](https://lgtm.com/projects/g/cc65/cc65).

N.B.: I am on the team behind LGTM.com, I'd appreciate your feedback on this initiative, whether you're interested or not, if you find time to drop me a line. Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,8 @@
 [Wiki](http://github.com/cc65/wiki/wiki)
 
 [![Build Status](https://api.travis-ci.org/cc65/cc65.svg?branch=master)](https://travis-ci.org/cc65/cc65/builds)
+[![Code Quality: Cpp](https://img.shields.io/lgtm/grade/cpp/g/cc65/cc65.svg?logo=lgtm&logoWidth=18)](https://lgtm.com/projects/g/cc65/cc65/context:cpp)
+[![Total Alerts](https://img.shields.io/lgtm/alerts/g/cc65/cc65.svg?logo=lgtm&logoWidth=18)](https://lgtm.com/projects/g/cc65/cc65/alerts)
 
 cc65 is a complete cross development package for 65(C)02 systems, including
 a powerful macro assembler, a C compiler, linker, librarian and several


groessler:
With the "Short global name" warnings the tool wants to enforce its idea of code style to the project. Can you run the tool with these warnings disabled?

The other warnings "Local variable hides global variable", "Declaration hides parameter", and "Comparison result is always true" might be worth investigating. Did you investigate these? And, depending on the findings, fix the code or disable the warning yet?

IMO, it makes no sense to add such a badge, which at this point in time would only indicate that the tool has been run over the code base.

oliverschmidt:
>IMO, it makes no sense to add such a badge, which at this point in time would only indicate that the tool has been run over the code base.

Me too.

xcorail:
Hello @groessler 

> With the "Short global name" warnings the tool wants to enforce its idea of code style to the project. 

Most of the maintainability rules are indeed subject to debate. We don't have our own ideas of code styles. For this problem, we do 2 things:
* We document all our rules with the rationale and references behind. for this one : https://lgtm.com/rules/2153510661/. So that you can chose to ignore it or no.
* We rank the alerts according to what developers tend to fix or to ignore. We analyse tens of thousands of github projects, and if an alert is never or rarely fixed by developers, it will automatically go down, from error to warning to recommendation to disappearance ... 

> Can you run the tool with these warnings disabled?

Sure. You can suppress these warnings in all places by adding a configuration file containing this exclusion: https://lgtm.com/help/lgtm/showing-hiding-query-results. **Do you want me to propose a PR for this?**

Or you can also suppress them in **selected code** by adding a suppression comment in the code: https://lgtm.com/help/lgtm/alert-suppression

> The other warnings "Local variable hides global variable", "Declaration hides parameter", and "Comparison result is always true" might be worth investigating. Did you investigate these? And, depending on the findings, fix the code or disable the warning yet?

I understand from this that you would prefer, as an approach, to propose some relevant fixes, rather than just displaying a badge. Thanks for this valuable feedback, I appreciate.

> IMO, it makes no sense to add such a badge, which at this point in time would only indicate that the tool has been run over the code base.

I thought that the badge could also have a benefit by driving your regular contributors to fix some alerts, and to globally care about keeping quality high. From your feedback and @oliverschmidt 's, I understand that it's not the case.

Thanks a lot again for the feedback. Note that even if you don't display this badge, the LGTM alerts will be always here for you.

mrdudz:
what exactly does this do that simple compiler warnings dont do already?

xcorail:
Hi @mrdudz 

Perhaps that the alerts shown in your case could have been all found by a compiler or a grep, but our product runs much deeper analysis than a compiler, or standard linters ... We extract the code as data in a DB, and run sophisticated queries on this DB thanks to an object oriented language.

You can have a look at some examples in our blog, for example [here](https://lgtm.com/blog/apache_struts_CVE-2018-11776), to see that it would have been very difficult to find this with a compiler or some grep.

For the complete list of queries, they are open source and available at https://github.com/Semmle/ql

Hope this answers your question, but I am not sure ... so don't hesitate to ping me back if not.
Cheers

xcorail:
@mrdudz 
you can check also this list of security vulnerability discovered thanks to this techno : https://lgtm.com/security/disclosures
(These are focused on security, but can give you an idea of what it can do)

oliverschmidt:
As far as I see there's no momentum to add the badge in the way supposed by this pull request.

PR #758, opened 2018-09-25
Implementation of clock_[gs]ettime() functions for Atari

groessler:
Only SpartaDOS-X is supported. On other DOSes ENOSYS is returned.

This PR also adds a simple test program, testcode/lib/clock-test.c.

### DIFF
--- a/testcode/lib/clock-test.c
+++ b/testcode/lib/clock-test.c
@@ -1,0 +1,100 @@
+/* Clock test program
+ *
+ * 25-Sep-2018, chris@groessler.org
+ */
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <time.h>
+#include <errno.h>
+
+#ifdef __CC65__
+#include <conio.h>
+#include <cc65.h>
+
+static void exitfn(void)
+{
+    if (doesclrscrafterexit()) cgetc();
+}
+#endif /* #ifdef __CC65__ */
+
+static void print_time(void)
+{
+    struct tm *cur_tm;
+    time_t cur_time = time(NULL);
+    if (cur_time == -1) {
+        printf("time() failed: %s\n", strerror(errno));
+        return;
+    }
+    cur_tm = localtime(&cur_time);
+
+    printf("time: %s\n", asctime(cur_tm));
+    // DEBUG:
+    printf("mday=%d mon=%d year=%d\nhour=%d min=%d sec=%d\n", cur_tm->tm_mday, cur_tm->tm_mon, cur_tm->tm_year, cur_tm->tm_hour, cur_tm->tm_min, cur_tm->tm_sec);
+}
+
+int main(int argc, char **argv)
+{
+    static char c;
+    static int s;
+    static struct tm cur_time;
+    static struct timespec new_time;
+
+#ifdef __CC65__
+    atexit(exitfn);
+#endif
+
+    if (argc <= 1) {
+        print_time();
+        return 0;
+    }
+
+    if (argc != 3 || strcasecmp(*(argv + 1), "set")) {
+        printf("usage: CLOCKTST [set DD-MM-YY-HH-MM-SS]\n");
+        return 1;
+    }
+
+    s = sscanf(*(argv + 2), "%d-%d-%d-%d-%d-%d", &cur_time.tm_mday, &cur_time.tm_mon, &cur_time.tm_year, &cur_time.tm_hour, &cur_time.tm_min, &cur_time.tm_sec);
+    if (s != 6 || cur_time.tm_year > 99 /* other input values aren't being verified... */) {
+        printf("invalid time/date format\n");
+        return 1;
+    }
+    --cur_time.tm_mon;
+    if (cur_time.tm_year < 79)
+        cur_time.tm_year += 100;  /* adjust century */
+
+    memset(&new_time, 0, sizeof(new_time));
+    new_time.tv_sec = mktime(&cur_time);
+
+    printf("\nyou are about to set the time to\n-->  %s\n\nContinue (y/n)?", ctime(&new_time.tv_sec));
+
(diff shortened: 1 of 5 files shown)


testcode/lib/clock-test.c
```diff
@@ -0,0 +1,100 @@
...
+
+int main(int argc, char **argv)
+{
+    static char c;
+    static int s;
+    static struct tm cur_time;
+    static struct timespec new_time;
+
+#ifdef __CC65__
+    atexit(exitfn);
```
oliverschmidt: I sugesst to do this in the style of https://github.com/cc65/cc65/blob/master/libsrc/atari/targetutil/w2cas.c#L47-L49 and https://github.com/cc65/cc65/blob/master/samples/diodemo.c#L127-L130

testcode/lib/clock-test.c
```diff
@@ -0,0 +1,100 @@
...
+    // DEBUG:
+    printf("mday=%d mon=%d year=%d\nhour=%d min=%d sec=%d\n", cur_tm->tm_mday, cur_tm->tm_mon, cur_tm->tm_year, cur_tm->tm_hour, cur_tm->tm_min, cur_tm->tm_sec);
+}
+
+int main(int argc, char **argv)
+{
+    static char c;
+    static int s;
+    static struct tm cur_time;
+    static struct timespec new_time;
```
oliverschmidt: If I'm not mistaken these four `static` don't make sense here, do they?

groessler: See https://cc65.github.io/doc/cc65.html#toc7.16
Here I wrote the "static" explicitly instead of using the pragma.

testcode/lib/clock-test.c
groessler: I can change that. I initially wrote it the way it is currently, and didn't change it so users can see that there are different ways to accomplish the same thing. Dou you want me to change it?

testcode/lib/clock-test.c
oliverschmidt: But why do you want to have those variables in the BSS segment instead of the stack in the first place?

groessler:
If a RTC (and its driver) is present, the time is set in the RTC. Verified on real hardware.

The default CONFIG.SYS which comes with SD-X tries to load a hardware clock driver and then JIFFY.SYS. So, if no RTC is present, JIFFY.SYS is loaded by default. I haven't checked what happens if no clock driver is loaded at all. Can do this tonight.

testcode/lib/clock-test.c
oliverschmidt: To be honest I'd rather like to see us present a canonical way to do things.

oliverschmidt:
> If a RTC (and its driver) is present, the time is set in the RTC. Verified on real hardware.

Nice :-)

> The default CONFIG.SYS which comes with SD-X tries to load a hardware clock driver and then JIFFY.SYS. So, if no RTC is present, JIFFY.SYS is loaded by default.

Cool :-)

> I haven't checked what happens if no clock driver is loaded at all. Can do this tonight.

If the default CONFIG.SYS loads JIFFY.SYS then there's no real need to check this out from my pov.

groessler:
> If the default CONFIG.SYS loads JIFFY.SYS then there's no real need to check this out from my pov.

I will probably check this anyway, but it shouldn't prevent you from merging this PR when the other issue are fixed.
If something bad happens without clock driver, I'll send a follow-up PR.

testcode/lib/clock-test.c
oliverschmidt: Thanks for pointing out the cc65 docs to me ;-))

With that argument all cc65 programs should be built that way. I think that this type of "general optimizationl" doesn't belong to test/sample code. I'd prefer if you'd  remove it.

groessler: Nitpicking, eh? What about these:


    $ find samples/ testcode/ -type f -name \*.c | xargs grep -n pragma | grep static
    samples/sieve.c:41:#pragma static-locals(1);
    samples/fire.c:59:#pragma static-locals (1);
    samples/mandelbrot.c:43:#pragma static-locals (1);
    samples/plasma.c:54:#pragma static-locals (1);
    $


Anyway, I've changed it...

oliverschmidt: > Nitpicking?

I don't think so. What you referenced...
* does not change the source to deviate from what one expects from a normal C program - and therefore don't distract the user from understanding it. The exsistance of this thread proves that the 'static's in the source code do so. They make readers like me wonder why your program may depend on those variables sitting in the BSS - maybe because of yet another banking mystery?
* does actually need the speed they get out static locals - while the program at hand doesn't.

groessler: Fair enough.

oliverschmidt: Thanks for your understanding - and the change :-)

oliverschmidt:
You surely noticed that there's no documentation of the clock_...() functions so far. As soon as there's general documentation (plus Apple II & CBM specific) I'll get in touch with you to ask for Atari specific doc...

oliverschmidt:
What about clock_getres() - https://github.com/cc65/cc65/blob/master/include/time.h#L143 ?

Maybe/likely pretty easy to implement as it's just a constant (or ENOSYS)...

groessler:
Why does the Apple2 implementation of clock_getres() just have a minute-resolution? Aren't there seconds?

oliverschmidt:
Correct - no seconds :-(

groessler:
Ok. Poor Apple users :-)
So for Atari, with one second resolution, I guess I can return (tv_sec = 1, tv_usec = 0), correct?

oliverschmidt:
> So for Atari, with one second resolution, I guess I can return (tv_sec = 1, tv_usec = 0), correct?

That's likely the "natural" thing. But only if no ENOSYS.

PR #757, opened 2018-09-21
Add checks for risky goto statements.

laubzega:
Emit a warning in a situation described in https://github.com/cc65/cc65/issues/756

PR #759, opened 2018-09-26
Fix for #757 (broken gotos between blocks with local variables)

laubzega:
This should behave, but more testing is always welcome. Same for improvement suggestions.


greg-king5:
cc65 has three stages:
1. Read the input file; and, compile it into an intermediate representation of Assembly code.
2. Optimize that intermediate code.
3. Translate the intermediate stuff into Assembly code; and, write it into the output file.

Stages 1 and 3 are single-pass; but, stage 2 is multiple-pass.  The optimizer will pass repeatedly over a function's intermediate code until it cannot optimize anymore.  It can remove code and _insert_ code.

It might be possible for the C label parser to generate a simple `JMP` instruction for a forward `goto`; then, optionally insert stack adjustment code, in front of the `JMP`, when the target label is found.

laubzega:
Ok, but is the 2nd stage performed even when -O is not specified? 

That would indeed be helpful in generating smaller code.

oliverschmidt:
> Ok, but is the 2nd stage performed even when -O is not specified?
 
Afaik: No. But I personally would nevertheless generate naive code in the 1st stage. And "only" make sure that the (optional) 2nd stage can deal well with it.

Regards,
Oliver

laubzega:
Thanks for the feedback. I changed the multi-line comments to the same format as used elsewhere.

oliverschmidt:
Thanks for the style adjustments. As no one else had remarks / suggestions I'm going to merge.

PR #762, opened 2018-09-27
Adds suport for Atari XEX file format to LD65

dmsc:
The standard binary format in Atari DOS supports multiple loadable sections with address and size.

By adding support for this file format to the linker, we can simplify linker scripts and remove the headers from assembly files, this simplifies creation of custom linker scripts that load data at non-default addresses.

This pull request is for commenting on the feature, first commits add support for XEX format without changing the default liner scripts, the last one modifies all the Atari linker scripts.

### DIFF
--- a/src/ld65/config.c
+++ b/src/ld65/config.c
@@ -1002,6 +1002,66 @@ static void ParseO65 (void)
 
 
 
+static void ParseXex (void)
+/* Parse the o65 format section */
+{
+    static const IdentTok Attributes [] = {
+        {   "RUNAD",    CFGTOK_RUNAD            },
+    };
+
+    /* Remember the attributes read */
+    /* Bitmask to remember the attributes we got already */
+    enum {
+        atNone          = 0x0000,
+        atRunAd         = 0x0001,
+    };
+    unsigned AttrFlags = atNone;
+    Import *RunAd = 0;
+
+    /* Read the attributes */
+    while (CfgTok == CFGTOK_IDENT) {
+
+        /* Map the identifier to a token */
+        cfgtok_t AttrTok;
+        CfgSpecialToken (Attributes, ENTRY_COUNT (Attributes), "Attribute");
+        AttrTok = CfgTok;
+
+        /* An optional assignment follows */
+        CfgNextTok ();
+        CfgOptionalAssign ();
+
+        /* Check which attribute was given */
+        switch (AttrTok) {
+
+            case CFGTOK_RUNAD:
+                /* Cannot have this attribute twice */
+                FlagAttr (&AttrFlags, atRunAd, "RUNAD");
+                /* We expect an identifier */
+                CfgAssureIdent ();
+                /* Generate an import for the symbol */
+                RunAd = InsertImport (GenImport (GetStrBufId (&CfgSVal), ADDR_SIZE_ABS));
+                /* Remember the file position */
+                CollAppend (&RunAd->RefLines, GenLineInfo (&CfgErrorPos));
+                /* Eat the identifier token */
+                CfgNextTok ();
+                break;
+
+            default:
+                FAIL ("Unexpected attribute token");
+
+        }
+
+        /* Skip an optional comma */
+        CfgOptionalComma ();
+    }
+
+    /* Set the RUNAD import if we have one */
+    if ( RunAd )
+        XexSetRunAd (XexFmtDesc, RunAd);
+}
+
+
+
 static void ParseFormats (void)
 /* Parse a target format section */
 {
@@ -1009,6 +1069,7 @@ static void ParseFormats (void)
         {   "O65",      CFGTOK_O65      },
         {   "BIN",      CFGTOK_BIN      },
         {   "BINARY",   CFGTOK_BIN      },
+        {   "ATARI",    CFGTOK_ATARIEXE },
     };
 
     while (CfgTok == CFGTOK_IDENT) {
@@ -1029,8 +1090,11 @@ static void ParseFormats (void)
                 ParseO65 ();
                 break;
 
-            case CFGTOK_BIN:
             case CFGTOK_ATARIEXE:
+                ParseXex ();
+                break;
+
+            case CFGTOK_BIN:
                 /* No attribibutes available */
                 break;
 
--- a/src/ld65/config.c
+++ b/src/ld65/config.c
@@ -68,6 +68,7 @@
 #include "objdata.h"
 #include "scanner.h"
 #include "spool.h"
+#include "xex.h"
 
 
 
@@ -149,6 +150,7 @@ static Collection       CfgSymbols = STATIC_COLLECTION_INITIALIZER;
 /* Descriptor holding information about the binary formats */
 static BinDesc* BinFmtDesc      = 0;
 static O65Desc* O65FmtDesc      = 0;
+static XexDesc* XexFmtDesc      = 0;
 
 
 
@@ -543,6 +545,7 @@ static void ParseFiles (void)
         {   "FORMAT",   CFGTOK_FORMAT   },
     };
     static const IdentTok Formats [] = {
+        {   "ATARI",    CFGTOK_ATARIEXE },
         {   "O65",      CFGTOK_O65      },
         {   "BIN",      CFGTOK_BIN      },
         {   "BINARY",   CFGTOK_BIN      },
@@ -605,6 +608,10 @@ static void ParseFiles (void)
 
                         case CFGTOK_O65:
                             F->Format = BINFMT_O65;
+                            break;
+
+                        case CFGTOK_ATARIEXE:
+                            F->Format = BINFMT_ATARIEXE;
                             break;
 
                         default:
@@ -1023,6 +1030,7 @@ static void ParseFormats (void)
                 break;
 
             case CFGTOK_BIN:
+            case CFGTOK_ATARIEXE:
                 /* No attribibutes available */
                 break;
 
@@ -1559,6 +1567,7 @@ void CfgRead (void)
     /* Create the descriptors for the binary formats */
     BinFmtDesc = NewBinDesc ();
     O65FmtDesc = NewO65Desc ();
+    XexFmtDesc = NewXexDesc ();
 
     /* If we have a config name given, open the file, otherwise we will read
     ** from a buffer.
@@ -2096,6 +2105,10 @@ void CfgWriteTarget (void)
 
                     case BINFMT_O65:
                         O65WriteTarget (O65FmtDesc, F);
+                        break;
+
+                    case BINFMT_ATARIEXE:
+                        XexWriteTarget (XexFmtDesc, F);
                         break;
 
                     default:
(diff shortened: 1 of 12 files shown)


mrdudz:
IMHO the linker should be as "dumb" as possible (that keeps complexity low and reduces errors) - so when this is possible already (with more complicated linker configs) then i'd vote against adding it.

polluks:
Well, XEX is not exotic... but you cannot create plain EXE anymore?

groessler:
I very much appreciate your effort, but as already posted on AtariAge, I don't like that change very much.

As @polluks pointed out, you won't be able to not have "__AUTOSTART__" (a RUNAD setting), or the __EXEHDR__ at all by just adding parameters to the link command line. You will have to modify the cfg file, which is not the case currently.

And, in fact, your *.cfg files are (to me) less readable than the current ones. They break the symmetry (lack of a better word): Currently everything what is written out to the EXE file is stated in the cfg file.

With your cfg files, some things are written automatically (chunk headers), but there's still the need to have special segments/memory areas to set an INITAD vector. That's a strange mixture of doing some things automatically and some things manually.

What exactly do you want to fix/improve? Do you have an example?

dmsc:
Hi!

> I very much appreciate your effort, but as already posted on AtariAge, I don't like that change very much.
> 
> As @polluks pointed out, you won't be able to not have "**AUTOSTART**" (a RUNAD setting), or the **EXEHDR** at all by just adding parameters to the link command line. You will have to modify the cfg file, which is not the case currently.
> 
> And, in fact, your *.cfg files are (to me) less readable than the current ones. They break the symmetry (lack of a better word): Currently everything what is written out to the EXE file is stated in the cfg file.
> 
> With your cfg files, some things are written automatically (chunk headers), but there's still the need to have special segments/memory areas to set an INITAD vector. That's a strange mixture of doing some things automatically and some things manually.
> 
> What exactly do you want to fix/improve? Do you have an example?

My idea is to facilitate adding data to specified addresses from a C program.

Currently, you can write from C:
```
#pragma data-name (push, "MYDATA")
static char mydata[64] = {1,2,3,4,5}; // Put this in page 6
#pragma data-name (pop)
```

This locates the data at a new segment. You then have to modify the linker script to add the location
```
# "main program" load chunk
    MAINHDR:    file = %O,               start = $0000, size = $0004;
    MAIN:       file = %O, define = yes, start = %S,    size = $BC20 - __STACKSIZE__ - __RESERVED_MEMORY__ - %S;
    MYMEM:      file = %O, start=$600, size=$100;
    TRAILER:    file = %O,               start = $0000, size = $0006;
```
But this does not work, as you also have to put a header:
```
# "main program" load chunk
    MAINHDR:    file = %O,               start = $0000, size = $0004;
    MAIN:       file = %O, define = yes, start = %S,    size = $BC20 - __STACKSIZE__ - __RESERVED_MEMORY__ - %S;
    MYMEMHDR:   file = %O, start=0, size=4;
    MYMEM:      file = %O, start=$600, size=$100;
    TRAILER:    file = %O,               start = $0000, size = $0006;
```
But you can't put the header in the C portion, as this wont work (non-constant initiallizer):
```
extern char _MYRAM_START__, MYRAM_LAST__
#pragma data-name (push, "MYDATAHDR")
static char *mydatahdr[4] = {&_MYRAM_START__, &_MYRAM_LAST__ - 1}; // header
#pragma data-name (push, "MYDATA")
static char mydata[64] = {1,2,3,4,5}; // Put this in page 6
#pragma data-name (pop)
```

So, you need to create at least one ASM file, with the header:
```
.segment        "MYRAMHDR"
        .word   __MYRAM_START__
        .word   __MYRAM_LAST__ - 1
```

With my patch, you simply add the MEMORY and SEGMENT sections to the config file ant it just works, you don't need to mess with the headers.

About not having AUTOSTART and EXEHDR without changing the config file, I don't know how it is possible now with only command line parameters, how so?

IrgendwerA8:
I value the change a lot and for me the PR seems to cover the more common case of building a compound file than the header-less case (which would be still possible).
Due to several alignment requirements for the Atari target a multi segment file is quite common and this change eases the handling (deriving from the standard configuration) very much!

IrgendwerA8:
> Well, XEX is not exotic... but you cannot create plain EXE anymore?

I'm not sure what you mean but for the Atari XEX==EXE==COM. (The XEX was introduced to allow opening of those file with emulators assigned to this extension and not interfere with DOS/Windows EXE-files.)

IrgendwerA8:
>As I said, what I don't like in this change is the introduced asymmetry in the cfg file, and the reduced flexibility.

As far as I understood you are still able to create a handmade config file without using the xex definition and with all the flexibility you like, so I do not understand this argument.

> With this change one has to add two new lines to cfg file for a new load chunk (one in MEMORY, one in SEGMENTS). Is it such a big deal to add four lines instead?
> One will need an asm source file, that's true. But one doesn't need to code or understand assembly language, there are just data definitions needed.

Yes, this additional two lines + an asm source file referring to the outside defined segment limits including the steeper learning curve and loss of comfort is making the difference for me. This is a tedious task, so automating this step is highly appreciated by me. You still have a hard time to produce a C file only Atari project when no wasting memory for alignment. Not (especially but not only) beginner friendly.

groessler:
> As far as I understood you are still able to create a handmade config file without using the xex definition and with all the flexibility you like, so I do not understand this argument.

I'm not opposing a "xex" backend per se, but I don't like to have the default cfg files changed to it.

I very much prefer to be able to see and have control over what exactly is happening. To me the changed cfg files become less understandable.

I'm not against to adding the xex backend and maybe provide an "atari-xex.cfg" file. But with this the question arises who will maintain this backend.

> Yes, this additional two lines + an asm source file referring to the outside defined segment limits including the steeper learning curve and loss of comfort is making the difference for me.

How often do you change these things? Typically once or twice or maybe even three times during development. But not much more often (that's my experience at least).

> You still have a hard time to produce a C file only Atari project without wasting memory for alignment.

Yes. But there are anyway only two (or so) good places to put aligned memory. One is page 6 (or page 5 + page 6) and the other one is in front of the program (while raising the start address accordingly). Anything else will never be "beginner friendly".
And I had offered you to think about some default segments at the beginning of the program just for this use case.

groessler:
Hi Daniel,

> But, I do worry that if it is not the default, people will simply not use it.

Define "default"? There are many Atari related config files in the `cfg` directory. People already now need to choose one to use as base point. And `atari-xex.cfg` looks obvious and is the simplest one on the surface, so I don't see that as a big road block.

But cc65 is not just "Atari". When you look at the cfg files for other targets you'll notice that all of them build the executable with means "out of the box". And I would dare to state, that as soon as one grasps the linker syntax, all of them are understandable. Changing all of the Atari (file-related) cfgs to the xex backend would make Atari more complicated than the other targets in this respect. ("All targets do it the same way, but Atari does it differently...")

The "initad" and "runad" changes I would leave out. If someone wants this capability he's anyway more sophisticated and can be bothered to use the normal cfgs. And they would introduce target specific code to the general ld65 cfg/cmdline parser.

Btw, if you know what you're doing you can already now get by with just two additional lines in the cfg file for a new load section. Place the memory area 4 bytes before the intended start address and put the header data there. If you want an INIT function call, you can put it at the end of the section data. But as I understand you want to simplify things for "newbies".

To summarize, I would approve your pull request (pending other cc65 team member's optinions), if it would only add one new cfg file (equivalent of atari.cfg), a simple test-case, and a documentation update.

PS: I simplified the state of things wrt to cfg files above. There is also the "o65" target which is different to all others. This is a relocatable file format. A backend for SpartaDOS-X relocatable executables would very much qualify to have its own ld65 backend. (hint, hint :-) )

groessler:
> There is also the "o65" target 

Sorry, meant `lunix.cfg` and `module.cfg` which use the "o65" backend.

oliverschmidt:
> Changing all of the Atari (file-related) cfgs to the xex backend would make Atari more complicated than the other targets in this respect. ("All targets do it the same way, but Atari does it differently...")

Full ACK.

> There is also the "o65" target which is different to all others. This is a relocatable file format. A backend for SpartaDOS-X relocatable executables would very much qualify to have its own ld65 backend.

Full ACK.

groessler:
@dmsc: What is the status? In case I wasn't clear, I like the idea to have an `atari-xex.cfg` file for beginners. But I don't like all Atari cfgs changed to the xex backend.
A documentation update would be required, and (optionally, but preferred) a testsuite update.

dmsc:
I added documentation and removed all modified linker config files, but I don't know how to add a test, seems all tests are for the compiler of assembler, but none for the linker.

groessler:
> but I don't know how to add a test

Maybe, at least, add a test program in `testcode/atari/lib` or in a subdirectory therein, which uses some (best: "all") features of the "xex" backend of the linker.

dmsc:
Added two test programs, one using the included linker script and another using a custom script.

src/common/target.c
```diff
@@ -183,10 +183,10 @@ static const TargetEntry TargetMap[] = {
 static const TargetProperties PropertyTable[TGT_COUNT] = {
     { "none",           CPU_6502,       BINFMT_BINARY,      CTNone  },
     { "module",         CPU_6502,       BINFMT_O65,         CTNone  },
-    { "atari",          CPU_6502,       BINFMT_BINARY,      CTAtari },
+    { "atari",          CPU_6502,       BINFMT_XEX,         CTAtari },
```
groessler: Why do you change default format for Atari?

src/common/target.c
```diff
@@ -183,10 +183,10 @@ static const TargetEntry TargetMap[] = {
...
     { "atari2600",      CPU_6502,       BINFMT_BINARY,      CTNone  },
     { "atari5200",      CPU_6502,       BINFMT_BINARY,      CTAtari },
-    { "atarixl",        CPU_6502,       BINFMT_BINARY,      CTAtari },
+    { "atarixl",        CPU_6502,       BINFMT_XEX,         CTAtari },
```
groessler: Ditto for atarixl

src/ld65/config.c
```diff
@@ -543,6 +545,8 @@ static void ParseFiles (void)
         {   "FORMAT",   CFGTOK_FORMAT   },
     };
     static const IdentTok Formats [] = {
+        {   "ATARI",    CFGTOK_XEX      },
+        {   "XEX",      CFGTOK_XEX      },
```
groessler: Is "atari" format an alias for "xex"? If yes, I'd prefer to get rid of "xex" inside the cfg file (and source code) and just use "atari", or "atariexe". The *.cfg file names can still contain "xex" if you prefer. It's a handler for "Atari EXE" file format.

groessler:
Thanks for your contribution and patience.
Now it would be nice to have a "xex" config for C programs :-)

groessler:
> Now it would be nice to have a "xex" config for C programs :-)

Is there any momentum towards this from your side?

regards,
chris

dmsc:
Hi!

> > Now it would be nice to have a "xex" config for C programs :-)
> 
> Is there any momentum towards this from your side?
> 
> regards,
> chris

Ok, look at #848 .

Regards,
    Daniel.

PR #771, opened 2018-10-06
da65: properly scan empty strings

DankRank:
Previously, doing something like `LABEL { NAME ""; ADDR $1000; };` would result in $1000 being labeled as NAME: instead of being unnamed.


src/da65/scanner.c
```diff
@@ -413,6 +413,9 @@ void InfoNextTok (void)
         case '\"':
             NextChar ();
             I = 0;
+            if (C == '\"') {
+                InfoSVal[0] = '\0';
```
greg-king5: You don't need to use that test.  `InfoSVal` always should be preset to an empty string there.

DankRank: Fixed

PR #776, opened 2018-10-17
Warning about alignment when section address is defined by START or OFFSET

vrubleg:
I propose to don't output a warning about alignment when section address is defined by START or OFFSET and it fits alignment requirements.

Motivation: I have a project for NES. It has a segment where START is set, and it is not allowed to set ALIGN for such sections (where START or OFFSET is set). Using `.ALIGN` in this section causes an annoying warning. I propose to not display this warning if the segment address is already aligned well.

### DIFF
--- a/src/ld65/config.c
+++ b/src/ld65/config.c
@@ -1881,9 +1881,14 @@ unsigned CfgProcess (void)
                 */
 
                 /* Check if the alignment for the segment from the linker
-                ** config. is a multiple for that of the segment.
+                ** config is a multiple for that of the segment.
+                ** If START or OFFSET is provided instead of ALIGN, check
+                ** if its address fits alignment requirements.
                 */
-                if ((S->RunAlignment % S->Seg->Alignment) != 0) {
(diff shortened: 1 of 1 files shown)


oliverschmidt:
Full ACK. Either there should be no warning on a (matching) `.ALIGN` or a (matching) ALIGN attribute should be allowed alongside a START/OFFSET attribute.

greg-king5:
> Either there should be no warning on a (matching) `.ALIGN`; or, a (matching) ALIGN attribute should be allowed alongside a START/OFFSET attribute.

I think that it should be a third choice:  they should be independent of each other:
1. The start of the memory area should initiate the address, then
2. the end of the previous segment (in that area) should adjust the address, then
3. `start=` should "pad" the address, then
4. `offset=` should "pad" the address, then
5. `align=` should "pad" the address.

If each one is done in sequence, then the final address will be what we expect -- without any conflicts.

oliverschmidt:
A change implementing what you describe would be welcome from my POV.

PR #777, opened 2018-10-17
CL65: --no-rtl option for disabling default runtime library

vrubleg:
Sometimes it is useful to have an ability to disable default runtime library. All modern toolsets have this option. For example, `/NODEFAULTLIB` in the MSVC, and `-nodefaultlibs -nostdlib` in Clang. I propose to add such an option to the CL65. It could be useful for pure assembly projects, for example.

### DIFF
--- a/src/cl65/main.c
+++ b/src/cl65/main.c
@@ -139,6 +139,7 @@ static int Module = 0;
 
 /* Name of the target specific runtime library */
 static char* TargetLib  = 0;
+static int   NoRTL      = 0;
 
 
 
@@ -413,6 +414,12 @@ static void SetTargetFiles (void)
     /* Get a pointer to the system name and its length */
     const char* TargetName = GetTargetName (Target);
     unsigned    TargetNameLen = strlen (TargetName);
+
+    if (NoRTL)
+    {
+        /* Default RunTime Library is disabled */
+        return;
+    }
 
     /* Set the library file */
     TargetLib = xmalloc (TargetNameLen + 4 + 1);
@@ -812,6 +819,7 @@ static void Usage (void)
             "  --memory-model model\t\tSet the memory model\n"
             "  --module\t\t\tLink as a module\n"
             "  --module-id id\t\tSpecify a module ID for the linker\n"
+            "  --no-rtl\t\t\tDon't link default runtime library\n"
             "  --o65-model model\t\tOverride the o65 model\n"
             "  --obj file\t\t\tLink this object file\n"
             "  --obj-path path\t\tSpecify an object file search path\n"
@@ -1167,6 +1175,15 @@ static void OptModuleId (const char* Opt attribute ((unused)), const char* Arg)
 
 
 
+static void OptNoRTL (const char* Opt attribute ((unused)),
+                      const char* Arg attribute ((unused)))
+/* Disable default runtime library */
+{
+    NoRTL = 1;
+}
+
+
+
 static void OptO65Model (const char* Opt attribute ((unused)), const char* Arg)
 /* Handle the --o65-model option */
 {
@@ -1369,6 +1386,7 @@ int main (int argc, char* argv [])
         { "--memory-model",      1, OptMemoryModel    },
         { "--module",            0, OptModule         },
         { "--module-id",         1, OptModuleId       },
+        { "--no-rtl",            0, OptNoRTL          },
         { "--o65-model",         1, OptO65Model       },
         { "--obj",               1, OptObj            },
         { "--obj-path",          1, OptObjPath        },
--- a/src/cl65/main.c
+++ b/src/cl65/main.c
@@ -139,7 +139,7 @@ static int Module = 0;
 
 /* Name of the target specific runtime library */
 static char* TargetLib  = 0;
-static int   NoRTL      = 0;
+static int   NoCrtLib   = 0;
 
 
 
@@ -415,12 +415,6 @@ static void SetTargetFiles (void)
     const char* TargetName = GetTargetName (Target);
     unsigned    TargetNameLen = strlen (TargetName);
 
-    if (NoRTL)
-    {
-        /* Default RunTime Library is disabled */
-        return;
-    }
-
     /* Set the library file */
     TargetLib = xmalloc (TargetNameLen + 4 + 1);
     memcpy (TargetLib, TargetName, TargetNameLen);
@@ -498,8 +492,11 @@ static void Link (void)
         CmdSetTarget (&LD65, Target);
     }
 
-    /* Determine which target libraries are needed */
-    SetTargetFiles ();
+    if (!NoCrtLib)
+    {
+        /* Determine which target libraries are needed */
+        SetTargetFiles ();
+    }
 
     /* Add all object files as parameters */
     for (I = 0; I < LD65.FileCount; ++I) {
@@ -819,7 +816,7 @@ static void Usage (void)
             "  --memory-model model\t\tSet the memory model\n"
             "  --module\t\t\tLink as a module\n"
             "  --module-id id\t\tSpecify a module ID for the linker\n"
-            "  --no-rtl\t\t\tDon't link default runtime library\n"
+            "  --no-crt-lib\t\t\tDon't link default C runtime library\n"
             "  --o65-model model\t\tOverride the o65 model\n"
             "  --obj file\t\t\tLink this object file\n"
             "  --obj-path path\t\tSpecify an object file search path\n"
@@ -1175,11 +1172,11 @@ static void OptModuleId (const char* Opt attribute ((unused)), const char* Arg)
 
 
 
-static void OptNoRTL (const char* Opt attribute ((unused)),
+static void OptNoCrtLib (const char* Opt attribute ((unused)),
                       const char* Arg attribute ((unused)))
 /* Disable default runtime library */
 {
-    NoRTL = 1;
+    NoCrtLib = 1;
 }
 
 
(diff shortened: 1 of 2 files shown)


mrdudz:
i think "no runtime" is misleading. it should be "no standard libraries" - without the runtime library nothing works (you probably always want to link it). and the libc etc isnt part of the runtime library :)

vrubleg:
So, what name do you suggest? `--no-default-lib` would look like other options, but it looks too long for me =)

This could be used in the docs:
> The --no-default-lib option tells the CL65 to not include default libraries into the list of libraries.

groessler:
gcc uses "-nostdlib" for something similar

regards,
chris

vrubleg:
So, we need to choose between `--no-rtl`, `--no-std-lib`, and `--no-default-lib`. What maintainers of the project would prefer?

greg-king5:
I like `--no-sys-lib`.

vrubleg:
`--no-crt` or `--no-crt-lib`. We need more options to choose from =)

oliverschmidt:
> [...] without the runtime library nothing works (you probably always want to link it).

According to my understanding going without the runtime library is the very goal of this pull request. For sure no C code works but the pull request comment says "It could be useful for pure assembly projects, for example.".

vrubleg:
Actually, I like `--no-crt-lib`. The library which is used by CL65 by default is actually a C runtime library. So, it will be clear from the name of the option that it just won't try to use default C runtime library. Probably, a user wants to provide own version of runtime (a patched one, for example), or it is a pure assembly project... And there could be a useful option for it with a meaningful name.

src/cl65/main.c
```diff
@@ -414,6 +415,12 @@ static void SetTargetFiles (void)
     const char* TargetName = GetTargetName (Target);
     unsigned    TargetNameLen = strlen (TargetName);
 
+    if (NoRTL)
+    {
```
greg-king5: That option should be implemented in the `Link()` function.  Example:
```c
if (LinkTargetLib) {
    SetTargetFiles ();
    CmdAddArg (&LD65, TargetLib);
}
```

vrubleg: I already moved this check to the `Link()` function.

oliverschmidt:
> The library which is used by CL65 by default is actually a C runtime library.

I must totally disagree. The library in question can't be a "C runtime library" as there's no "C runtime library" in general.

There are two things:

* A "C standard library". Things like memcpy() come from here.
* A "runtime library". Things private / specific to the compiler come from here. This includes the startup code.

The term "C runtime library" was as far as I know coined in the context of Microsoft C. But the libary in question isn't a C runtime library from Microsoft. It's a runtime library for Microsoft C. Trying to express in a different way: It's not `(Microsoft) C runtime library`. It's `(Microsoft C) runtime library`. From that perspective I dislike the `--no-crt` or `--no-crt-lib` options.

I don't like the `--no-sys-lib` because the notion of a "system library" would be totally new.

Valid options from my POV are `--no-std-lib` and `--no-default-lib`. Given that the library in question contains both the C standard library and the runtime library I think that `--no-default-lib` is better. But if there's a momentum for `--no-std-lib` I'm happy with that too.

vrubleg:
@oliverschmidt, OK, I'll update the PR in a moment. `--no-std-lib` looks nice for me also.

oliverschmidt:
> OK, I'll update the PR in a moment.

Thanks :-)

However, I "need" to do some nitpicking... You mention 'default runtime' in comments and the user help texts but the option is now called --no-std-lib.

* Please name "the thing" 'standard library' where visible to the user for consistency.

* In comments you can either go for 'standard library' too or call it 'system runtime library' given that there was already the comment `/* Add the system runtime library */` before.

vrubleg:
@oliverschmidt, done. I've updated my previous commit.

vrubleg:
Thanks. I'm glad that now we have such an option.

greg-king5:
> I don't like the `--no-sys-lib` because the notion of a "system library" would be totally new.

The idea of a system library has existed for a long time.  The word "system" is used in the cl65 document six times.  Here's the most relevant quote:
> When linking, cl65 will supply the name of the library file for the target _system_ to the linker; so, you don't have to do that.

The names:
- target library
- platform library
- system library

are synonyms.

oliverschmidt:
I don't find the term "system library" in the sentence "When linking, cl65 will supply the name of the library file for the target system to the linker; so, you don't have to do that." ;-))

> The names: target library / platform library / system library are synonyms.

Not from my POV.

BTW: I thought that "target library" (in contrast to "system library") would at least for me personally describe "the thing" best - thus sort of agreeing with you. It is the combination of the C standard library and the runtime library so both terms aren't great. But for sure it's the target library. However I felt that for the average user the term "standard library" is preferable.

mrdudz:
> According to my understanding going without the runtime library is the very goal of this pull request. For sure no C code works but the pull request comment says "It could be useful for pure assembly projects, for example.".

if that is actually the goal here - ie linking nothing at all, no runtime, no crt0, nada - then i would suggest calling it "freestanding" like GCC does. "no standardlib" implies the crt0 and runtime is still being used.

oliverschmidt:
If there's a strong momentum for a different name than the current one I'm not against renaming it. At least by now there should be no backward compatibility issue ;-)

Just two points from my POV:
* I hope we agree that "C runtime library" is no option - as such a thing doesn't exist.
* I strongly prefer 'target library' over 'system library'.

vrubleg:
https://gcc.gnu.org/onlinedocs/gcc-8.2.0/gcc/Link-Options.html

> -nostartfiles
>     Do not use the standard system startup files when linking. The standard system libraries are used normally, unless -nostdlib or -nodefaultlibs is used.
> 
> -nodefaultlibs
>     Do not use the standard system libraries when linking. Only the libraries you specify are passed to the linker, and options specifying linkage of the system libraries, such as -static-libgcc or -shared-libgcc, are ignored. The standard startup files are used normally, unless -nostartfiles is used.
>     The compiler may generate calls to memcmp, memset, memcpy and memmove. These entries are usually resolved by entries in libc. These entry points should be supplied through some other mechanism when this option is specified.
> 
> -nostdlib
>     Do not use the standard system startup files or libraries when linking. No startup files and only the libraries you specify are passed to the linker, and options specifying linkage of the system libraries, such as -static-libgcc or -shared-libgcc, are ignored.
>     The compiler may generate calls to memcmp, memset, memcpy and memmove. These entries are usually resolved by entries in libc. These entry points should be supplied through some other mechanism when this option is specified.
>     One of the standard libraries bypassed by -nostdlib and -nodefaultlibs is libgcc.a, a library of internal subroutines which GCC uses to overcome shortcomings of particular machines, or special needs for some languages. (See Interfacing to GCC Output in GNU Compiler Collection (GCC) Internals, for more discussion of libgcc.a.) In most cases, you need libgcc.a even when you want to avoid other standard libraries. In other words, when you specify -nostdlib or -nodefaultlibs you should usually specify -lgcc as well. This ensures that you have no unresolved references to internal GCC library subroutines. (An example of such an internal subroutine is __main, used to ensure C++ constructors are called; see collect2 in GNU Compiler Collection (GCC) Internals.)

So, GCC's `-nostdlib` does exactly the same as our `--no-std-lib`.

groessler:
> Probably, a user wants to provide own version of runtime (a patched one, for example), or it is a pure assembly project... 

Nitpicking a bit :-), I'm regularily using "cl65" to compile asm programs. Unless the asm program doesn't import a symbol (which isn't provided by the asm program itself) of the runtime/system/howeveryoucallit library, it doesn't hurt.

oliverschmidt:
> I'm regularily using "cl65" to compile asm programs.

I'm almost sure that a lot of users do that (incl. me).

> Unless the asm program doesn't import a symbol (which isn't provided by the asm program itself) of the runtime/system/howeveryoucallit library, it doesn't hurt.

If that we're the case we likely would have the option proposed here for many years.

greg-king5:
> I don't find the term "system library" in the sentence "When linking, cl65 will supply the name of the library file for the target system to the linker; so, you don't have to do that." ;-))

To me, the phrase "library file for the target system" means "library for the system", which means "system library".  The phrase means also "library for the target", which means "target library".  So, I still claim that they're synonyms.

I like `--no-target-lib`, as much as I like `--no-sys-lib`.

groessler:
> now thats interesting... when i did related things with gcc (admitted, that was over 15 years ago) "freestanding" was the keyword :)

That is still supported. Don't know the difference right now btw. -ffreestanding and -nostdlib -nostdfinc.

oliverschmidt:
I think especially for a cross dev tool "system" in itself can't describe something well as it may mean "host system" as well as "target system".

greg-king5:
> I think, especially for a cross-dev. tool, "system" in itself can't describe something well, as it may mean "host system" as well as "target system".

Good point.

silverdr:
> I like --no-target-lib, as much as I like --no-sys-lib.

FWIW and if my vote counted, I would vote for "target" (as in "--no-target-lib") rather than "system". Target is a well-established term in cc65 terminology and hard to be mistaken for something else than... "cc65 target". System OTOH isn't specific enough to me. Can mean a lot of things and be easily mistaken for something else than intended.

mrdudz:
i'd prefer "target" as well...

oliverschmidt:
Okay, I see about everybody commenting so far to like `--no-target-lib`. @greg-king5: Can you imagine to create a commit / pull request?

vrubleg:
@oliverschmidt, https://github.com/cc65/cc65/pull/783

oliverschmidt:
Me too ;-)

Fabrizio-Caruso:
Sorry for the stupid question: what is the difference between using this option and using the none target?

Fabrizio-Caruso:
This new name makes it look like target=none. This is confusing for me.

oliverschmidt:
The none target links the none.lib. It contains all non-target-specific code. This includes:
* The runtime library
* The non-target-specific parts of the C standard library

It allows you to link C programs doing nothing target-specific but e.g. call memcpy() and alike.

In constrast the new option links no library at all. It allows you to link assembler programs.

oliverschmidt:
Me:
> I thought that "target library" [...] would at least for me personally describe "the thing" best [...] However I felt that for the average user the term "standard library" is preferable.

You:
> This new name makes it look like target=none. This is confusing for me.

;-))

PR #791, opened 2018-11-07
Atari target: Support for self explanatory KBCODE values

IrgendwerA8:


include/_pokey.h
```diff
@@ -7,6 +7,7 @@
 /*                                                                           */
 /*                                                                           */
 /* (C) 2000 Freddy Offenga <taf_offenga@yahoo.com>                           */
+/* 06-Nov-2018: Christian Krueger: Added defines for keyboard codes          */
```
oliverschmidt: As far as I know/see none of the cc65 C header files contains a comment like this so far. So I'm wondering you are willing to contribute without it. In general I personally don't see the point in having "inband" change history given the omnipresence of version control systems.

groessler: There is "prior art" like that in _antic.h. But I'd prefer to have it formatted (and _antic.h changed accordingly) like in e.g. apple2.h.

groessler:
I think the assembler defines shouldn't be in atari_pokey.inc, but in atari.inc. Hmm, yes, the same for the C defines in _pokey.h, they should be in atari.h. atari_pokey.inc and _pokey.h are also used by the atari5200 target where they don't apply.

include/_pokey.h
oliverschmidt: Thanks for the hint :-) I fully agree: The 'apple2.h' style is fine - if considered necessary in the first place.

include/atari.h
```diff
@@ -159,6 +159,103 @@
 #define JOY_FIRE_MASK   JOY_BTN_1_MASK
 #define JOY_FIRE(v)     ((v) & JOY_FIRE_MASK)
 
+/* Keyboard values returned by kbcode / CH */
+#define KEY_NONE        (unsigned char) 0xFF
```
oliverschmidt: I'm not sure if it's necessary because of the high precedence of the cast operator but my gut feeling asks for brackets around the term.

However, if "everybody else" says this is just stupid and only shows I haven't understood C then I'm perfectly willing to merge as-is.

groessler: I'd prefer too to have brackets around the whole expression.
But, may I ask, why is the cast operator there? The other "constant defines" in atari.h don't have it.

IrgendwerA8: So that when they are used in void variables/arrays (cc65 extension) only eat up one byte instead of two. (Like the definitions in "_antic.h").

groessler: Can you give an example where this makes a difference? (Don't want to talk about the _antic.h definiitions right now, I was never in favor of them...)

include/_pokey.h
IrgendwerA8: I've removed the info but thought about if removing all personal credits doesn't make more sense?
(A wrong information is more worse than none.)
Wouldn't make something like "(c) The cc65 team and contributors (https://www.cc65.org/doc/CREDITS)" make more sense? (Of course with a link to an updated list here.)

include/atari.h
IrgendwerA8: `void keysToUpHandler = {KEY_W,KEY_UP,KEY_NONE,upHandler};` ?

About "_antic.h": Did you ever used the definitions?

include/_pokey.h
oliverschmidt: I personally rather dislike both those credits in the header files as well as the credit/history section at the very top of many/most asm source files for exactly the reason given: They are wrong/outdated. I'm personally open to removing any of those. **HOWEVER**,  I can imagine that those things are important for some contributor or another so I suggest to not remove anything as they just don't bother me/us enough to risk any irritation of even a single contributor. But at least I personally never add anything anywhere, I just don't care.

IrgendwerA8: As the main maintainer and so being rather "visible" you can talk... ;)
But what about reestablishing the central credits and linking to them? A contributor could add himself to the credits if he feels so.
I have zero problem with not being mentioned, but giving the copyright in the header to a person which isn't fully responsible for the content doesn't feels right.
Any opinions from others?

oliverschmidt: Trying to bring that topic forward...

What about this: Someone (@IrgendwerA8 ?) checks that credits file and adds those who are named in any header file. Then the header files are changed as proposed. I'm willing to merge such a PR without broad consultation.

IrgendwerA8: Good proposal. Being quite thoroughly, maybe @polluks could chime in?

include/atari.h
groessler: > About "_antic.h": Did you ever used the definitions?
No.
I guess it's just a matter of personal preference. To me, with the old definitions, something like `DL_CHR40x8x1 | DL_DLI` is immediately obvious, but when I see `DL_DLI(DL_CHR40x8x1)` I need to run another thinking thread along the lines "what is `DL_DLI()` exactly doing again?"

asminc/atari_pokey.inc
```diff
@@ -41,4 +41,3 @@ POTGO   = POKEY + $0B         ;start potentiometer scan sequence
 SEROUT  = POKEY + $0D         ;serial port output
 IRQEN   = POKEY + $0E         ;IRQ interrupt enable
 SKCTL   = POKEY + $0F         ;serial port and keyboard control
-
```
groessler: spurious whitespace change?

include/_pokey.h
```diff
@@ -73,6 +73,7 @@ struct __pokey_read {
     unsigned char   skstat; /* serial port status */
 };
 
+
```
groessler: spurious whitespace change?

oliverschmidt:
Maybe I'm missing the point but to me it seems that this pull request...
* now is consistent
* improves the current  situation

If `DL_CHR40x8x1 | DL_DLI` isbetter or worse than `DL_DLI(DL_CHR40x8x1)` seems rather unrelated to this pull request, correct?

So if I'm right I'd merge this.

groessler:
> So if I'm right I'd merge this.

Fine with me.
The DL definitions parts was just the discussion running into OT-land.
For the two whitespace changes I don't have a strong optinion. Just wanted to mention them.

oliverschmidt:
> The DL definitions parts was just the discussion running into OT-land.

Thanks for the confirmation. You're of course welcome to discuss this further here on the (by then merged) pr conversation.

> For the two whitespace changes I don't have a strong optinion. Just wanted to mention them.

Both changes were made deliberately and are actually an improvement.

groessler:
> Both changes were made deliberately and are actually an improvement.

Could you elaborate?

oliverschmidt:
* asminc/atari_pokey.inc: It had an empty line at the end. Files edited by Uz often had 6 and more empty lines at the end. I too remove them when I notice them.

* include/_pokey.h: The predominant style are *THREE* empty lines between
  * header comment and multi inclusion guard start
  * multi inclusion guard start and "body"
  * in front of and behind section separation comments
  * "body" and multi inclusion guard end

groessler:
OK

include/_pokey.h
greg-king5: > spurious whitespace change?

It's a style change that brings the file closer to the way that Uz formatted his files.

PR #795, opened 2018-11-10
Add VIC control register names

ops:
Added VIC control registers to vic20.inc. Register names from the COMPUTE!'s book 'Mapping the VIC'.


oliverschmidt:
As nobody complained so far I'm merging now...

PR #803, opened 2018-11-25
CBM: Move kernal entries to cbm_kernal.inc

ops:
- Moved kernal entries to cbm_kernal.inc from platform specific files
- Adjusted libsrc to use cbm_kernal.inc 
- Some whitespace cleanups


libsrc/c64/ser/c64-swlink.s
```diff
@@ -45,15 +46,15 @@
...
-        .word   IRQ
+        .word   SWL_INSTALL
+        .word   SWL_UNINSTALL
+        .word   SWL_OPEN
+        .word   SWL_CLOSE
+        .word   SWL_GET
+        .word   SWL_PUT
+        .word   SWL_STATUS
+        .word   SWL_IOCTL
+        .word   SWL_IRQ
```
oliverschmidt: Although of course not technically necessary these names were so far always identical for all drivers of a certain driver "class" (and to https://github.com/cc65/cc65/blob/master/asminc/ser-kernel.inc#L44-L52). I'd really like to keep it that way. Therefore I ask you to change them in all cc65 SER drivers (and https://github.com/cc65/cc65/blob/master/asminc/ser-kernel.inc#L44-L52). As far as I can see SER_INSTALL etc. should work fine.

ops: Done. I hope I didn't miss any.

oliverschmidt: Thanks for the adjustment :-)

PR #808, opened 2018-12-03
Serial driver for the C64 w/o dedicated hardware

polluks:
Johan's driver is waiting for four years, time for an official commit.

### DIFF
--- a/libsrc/c64/ser/c64-up2400.s
+++ b/libsrc/c64/ser/c64-up2400.s
@@ -1,0 +1,476 @@
+; Serial driver for the C64 using Userport.
+;
+; Johan Van den Brande, (c) 2014
+; 
+; Based on George Hug's 'Towards 2400' article in the Transactor Magazine volume 9 issue 3.
+; (https://archive.org/details/transactor-magazines-v9-i03) 
+;
+        .include        "cbm_kernal.inc"
+        .include        "zeropage.inc"
+        .include        "ser-kernel.inc"
+        .include        "ser-error.inc"
+        .include        "c64.inc"
+
+        .macpack        module
+
+; ------------------------------------------------------------------------
+; Constants
+ribuf = $f7
+robuf = $f9
+baudof = $0299
+ridbe = $029b
+ridbs = $029c
+rodbs = $029d
+rodbe = $029e
+enabl = $02a1
+rstkey = $fe56
+norest = $fe72
+return = $febc
+oldout = $f1ca
+oldchk = $f21b
+findfn = $f30f
+devnum = $f31f
+nofile = $f701
+
+; ------------------------------------------------------------------------
+; Header. Includes jump table
+
+        module_header   _c64_up2400_ser
+
+; Driver signature
+
+        .byte   $73, $65, $72           ; "ser"
+        .byte   SER_API_VERSION         ; Serial API version number
+
+; Library reference
+
+        .addr   $0000
+
+; Jump table
+
+        .word   SER_INSTALL
+        .word   SER_UNINSTALL
+        .word   SER_OPEN
+        .word   SER_CLOSE
+        .word   SER_GET
+        .word   SER_PUT
+        .word   SER_STATUS
+        .word   SER_IOCTL
+        .word   SER_IRQ
+
+.data
+
+strt24: .word $01cb     ; 459   start bit times
+strt12: .word $0442     ; 1090
+strt03: .word $1333     ; 4915
+full24: .word $01a5     ; 421   full bit times
+full12: .word $034d     ; 845
+full03: .word $0d52     ; 3410
+
+.rodata
+
+baudrate = 10
+databits = 0
+stopbit = 0
+
+wire = 1
+duplex = 0
+parity = 0
+
+serial_config:
+  .byte baudrate + databits + stopbit
+  .byte wire + duplex + parity
+
+.code
+
+;----------------------------------------------------------------------------
+; INSTALL routine. Is called after the driver is loaded into memory. If
+; possible, check if the hardware is present.
+; Must return an SER_ERR_xx code in a/x.
+
+SER_INSTALL:
+;;        lda     #<SER_ERR_OK
+;;        tax     
+;;        rts
+
+;----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


libsrc/c64/ser/c64-up2400.s
```diff
@@ -0,0 +1,476 @@
...
+rodbe = $029e
+enabl = $02a1
+rstkey = $fe56
+norest = $fe72
+return = $febc
+oldout = $f1ca
+oldchk = $f21b
+findfn = $f30f
+devnum = $f31f
+nofile = $f701
```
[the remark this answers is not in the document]
groessler: Maybe consider adding the addresses to `c64.h` in case they could be useful for other programs.

oliverschmidt:
Nice! Seems to be the right time with all those ESP8266 asking to be connected ;-)

I'm looking forward to remove the...

"Implement a Userport SER driver based on George Hug's 'Toward 2400' source code -- http://archive.org/details/transactor-magazines-v9-i03."

...from https://github.com/cc65/wiki/wiki/Reminders#CBM_library.

I presume a brief addition to https://cc65.github.io/doc/c64.html#ss7.5 is about to come as well...

mrdudz:
While i dont mind self modifying code myself - is it really a good idea to use it in cc65 library code? it will certainly lead to unexpected results in some cases (ROM)

also, i am missing the table of CIA timer values for both PAL and NTSC... the tables in the kernal had various issues (at least 2400 is broken in PAL, on R1 kernal its completely broken, and probably more i dont remember)

libsrc/c64/ser/c64-up2400.s
```diff
@@ -0,0 +1,476 @@
...
+;;        tax
+;;        rts
+
+;----------------------------------------------------------------------------
+; STATUS: Return the status in the variable pointed to by ptr1.
+; Must return an error code in a/x.
+;
+
+SER_STATUS:
+        lda     #<SER_ERR_OK
```
groessler: Since it doesn't really do anything it shouldn't return an "OK" status.

libsrc/c64/ser/c64-up2400.s
```diff
@@ -0,0 +1,476 @@
...
+oldout = $f1ca
+oldchk = $f21b
+findfn = $f30f
+devnum = $f31f
+nofile = $f701
+
+; ------------------------------------------------------------------------
+; Header. Includes jump table
+
+        module_header   _c64_up2400_ser
```
groessler: No other serial driver has the max baud speed in its name. I vote for calling it `_c64_userport_ser` or `_c64_up_ser`.

mrdudz: it makes sense - to seperate it from a driver for daniel dallmanns 9600 baud hack (which requires some extra connections), which is commonly referred to as "up9600"

perhaps drivers that dont require extra hardware should generally have "std" in their names?

groessler: Kinda off-topic: where can I get an userport interface with or w/o "daniel dallmann's hack for 9600"?

mrdudz: if you arent afraid of some (very easy) soldering, you can make it yourself. easiest is to get a TTL->rs232 dongle from ebay and just add a userport connector. i recommend something like this: https://gglabs.us/node/530

groessler: > i recommend something like this: https://gglabs.us/node/530

I think I could build that from the schematics. But browsing their site I found that: https://gglabs.us/node/2044
Shouldn't that be the same? (When I have the choice between building myself or buying a prebuilt part, I typically choose the latter. :-) )

groessler: And, another, sorry for the now really off-topic, question: What kind of adapter would I need to try the Plus/4 implemetation on hardware?

polluks: @groessler http://cbm-hackers.2304266.n4.nabble.com/userport-C64-Plus4-others-tp4447868p4450398.html

libsrc/c64/ser/c64-up2400.s
```diff
@@ -0,0 +1,479 @@
...
+        .include        "zeropage.inc"
+        .include        "ser-kernel.inc"
+        .include        "ser-error.inc"
+        .include        "c64.inc"
+
+        .macpack        module
+
+; ------------------------------------------------------------------------
+; Constants
+RecvHead = $b5          ; Head of receive buffer
```
oliverschmidt: Thanks for starting to add (more) constants. It would be nice if the names would adhere to the same style. Either CamelCase for all or for none (`RecvHead` vs. `ribuf`).

polluks: I was using the label of c64-swlink.s...

greg-king5:
I must close this PR because we cannot add Johan's driver to our project.  He wrote it for his private project,  _not_ cc65.  It has too many flaws.

In addition to the cosmetic ones that Oliver noticed, there are problems in the code.  Some functions are overly complicated; and, an important function is missing.  `SER_CLOSE:` does **nothing** (Johan didn't need it).  That's why this driver causes trouble for PlatoTerm64 (@tschak909, your observations are absolutely correct).

I've been working, on and off, on a fixed version of his driver.  It worked well, in the past; but, I might have broken it recently.  I have trouble debugging the thing because my connections between the VICE emulator, my computers, the Internet, and Irata.Online are unreliable.

greg-king5:
By the way, George Hug's code has a timing table for only NTSC because he needed to find good values through trial-and-error -- on his NTSC C64.  After we have a working driver, someone with a PAL machine will need to do the same trials; and, publish the results.

tschak909:
@greg-king5 anything I can do to help with your connection issues?

mrdudz:
also i can only repeat once more: use real hardware to debug this rs232 stuff. VICE is not good at it.

oliverschmidt:
So I'm left to say sorry for initiating effort to improve the cosmetic issues on code which was doomed in the first place.

polluks:
@greg-king5 At least SER_CLOSE should call the SER_IOCTL...

PR #824, opened 2018-12-27
add support for detecting 45GS02

gardners:
Added support for detecting for MEGA65 45GS02 CPU.
Also renamed a bunch of labels to make it easier to follow the flow of getcpu().

PR #826, opened 2018-12-27
add support for detecting 45GS02, give labels in getcpu.s useful names

gardners:
This is the pull request to replace the previous one, with all requested changes made.

### DIFF
--- a/libsrc/common/getcpu.s
+++ b/libsrc/common/getcpu.s
@@ -3,7 +3,7 @@
 ;
 ; unsigned char getcpu (void);
 ;
-
+	.include	"zeropage.inc"
         .export         _getcpu
 
 ; ---------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


libsrc/common/getcpu.s
```diff
@@ -26,52 +27,119 @@
...
+        ldx     #4
+@L10:   lda     $fb,x
+        sta     @GetCPUTemp,x
+        dex
+        bpl     @L10
+
+        ; Setup 32-bit pointer to $000200FF
+        lda     #$ff
+        sta     $fb
+        lda     #$00
```
oliverschmidt: Rather #0 ?

libsrc/common/getcpu.s
```diff
@@ -26,52 +27,119 @@
...
+        sta     $fc
+        sta     $fe
+        lda     #$02
```
oliverschmidt: Rather #2 ?

libsrc/common/getcpu.s
```diff
@@ -26,52 +27,119 @@
...
+        ; Save the 32-bit ZP pointer and data byte
+        ldx     #4
+@L11:   lda     @GetCPUTemp,x
+        sta     $fb,x
+        dex
+        bpl     @L11
+        rts
+
+        ; Temporary storage for the ZP locations we modify above
+@GetCPUTemp:    .byte   0,0,0,0,0
```
oliverschmidt: Please take care of https://github.com/cc65/cc65/pull/824#discussion_r244158099.

libsrc/common/getcpu.s
gardners: Here I used #$00 to indicate that it is part of the 32-bit address indicated in the comment on line 66, which would not be as obvious if #0 were used instead. But happy to go with your preference.

libsrc/common/getcpu.s
gardners: Here I used #$02 to indicate that it is part of the 32-bit address indicated in the comment on line 66, which would not be as obvious if #0 were used instead. But happy to go with your preference.

gardners: Actually, for these, would it be better to use LDA #<$0002 / LDA #>$0002 , LDA #<$00FF / LDA #$>00FF to make it clearer to the reader that the constants are parts of addresses?

greg-king5: > Actually, for these, would it be better to use `LDA #<$0002` / `LDA #>$0002` , `LDA #<$00FF` / `LDA #>$00FF` to make it clearer to the reader that the constants are parts of addresses?

Definitely yes!  And, don't forget about the `^` operator which can pull the third (bank) byte out of a 24-bit number.

libsrc/common/getcpu.s
```diff
@@ -26,52 +27,119 @@
...
+
+        ; 45GS02 supports 32-bit ZP indirect, so use that to check CPU type
+        ; without requiring a functioning MEGA65 hypervisor.
+        ; We setup a read of $200FF, then store a different value in $00FF
+        ; and then re-read $200FF to see if it is unchanged.
+
+        ; Save the 32-bit ZP pointer and data byte
+        ldx     #4
+@L10:   lda     $fb,x
+        sta     @GetCPUTemp,x
```
greg-king5: You don't need to "steal" any zero-page locations.  The run-time code uses many locations that are available.  Most of them aren't being used while `getcpu()` is running; and, their contents don't need to be saved.

Look at [zeropage.s](https://github.com/cc65/cc65/blob/master/libsrc/runtime/zeropage.s).  You must `.include "zeropage.inc"`.  Then, you can use any group of four bytes except `sp:` and `regbank:`.

libsrc/common/getcpu.s
oliverschmidt: Full ACK. I just didn't notice that these is about addr bytes.

libsrc/common/getcpu.s
oliverschmidt: There's even documentation: https://github.com/cc65/wiki/wiki/Using-runtime-zeropage-locations-in-assembly-language

libsrc/common/getcpu.s
```diff
@@ -26,52 +27,119 @@
...
+        ; (We assume Z=$00, which will be the normal case)
+        nop                     ; prefix to tell next instruction to be 32-bit ZP
+        .byte   $b2,$fb         ; LDA ($nn),Z
+        eor     #$ff            ; change the value
+        sta     $ff             ; store in $FF
+        ; now try again to load it: If the same, then 45GS02, as $200FF is unchanged
+        nop                     ; prefix to tell next instruction to be 32-bit ZP
+        .byte   $b2,$fb         ; LDA ($nn),Z
+        cmp     $ff             ; does the loaded value match what is in $FF?
+        beq     @Is4510         ; matches, so must be a 4510 = C65
```
greg-king5: That line isn't needed.  Its target is the next line after the `bne`.

libsrc/common/getcpu.s
gardners: Addressed in commits below.

libsrc/common/getcpu.s
gardners: Now uses <, >, and ^ to make it all more obvious.

libsrc/common/getcpu.s
gardners: Now uses zeropage.inc to use existing ZP storage.

libsrc/common/getcpu.s
gardners: Now uses zeropage.inc to use existing ZP storage.

libsrc/common/getcpu.s
gardners: Removed.

gardners:
I think I have addressed everything raised above now. Let me know if you spot anything else that needs my attention.

libsrc/common/getcpu.s
```diff
@@ -53,64 +53,40 @@ _getcpu:
...
-        ; now try again to load it: If the same, then 45GS02, as $200FF is unchanged
+        sta     tmp1            ; store in $xx
+        ; now try again to load it: If the same, then 45GS02, as $200xx is unchanged
         nop                     ; prefix to tell next instruction to be 32-bit ZP
-        .byte   $b2,$fb         ; LDA ($nn),Z
-        cmp     $ff             ; does the loaded value match what is in $FF?
-        beq     @Is4510         ; matches, so must be a 4510 = C65
+        .byte   $b2,regsave     ; LDA (regsave),Z
+        cmp     tmp1            ; does the loaded value match what is in $FF?
         bne     @Is45GS02       ; $200FF and $FF have different values, so must be a MEGA65 45GS02
```
greg-king5: `$xx` instead of `$FF`.

gardners: Well spotted.  Fix pushed.

oliverschmidt:
@greg-king5: You have by now a much better insight than I have. Please feel free to merge as you see fit.

oliverschmidt:
Thanks :-)

gardners:
Thanks both for your assistance,
Paul.

PR #829, opened 2019-01-05
ca65: Add string_escapes feature. Resolves #535

### LINKED ISSUE #535: ca65: Feature Request: Support escape sequences in strings
It would be handy to be able to write:
```asm
    .byte   "Give us a hint, by ringing a bell!\x07"
```
Although it can be written as `"abc",$07,"def"` this makes string-processing macros far more complex, and impacts readability.

This would presumably require a new directive to enable escapes (`.STRINGESCAPES +` or some such) for backwards compatibility, and making it play nicely with `.LINECONT` would be important - I want both!

* The most important escape to include would be the C-style `\xNN` form (IMHO)
* `\\` for a literal backslash, of course.
* `\"` for embedded quotes
* `\t` ($09) and `\r` ($0D) are probably safe
* `\n` gets tricky. Is that platform specific ($0D on apple2) or generic ($0A) ?
* I've never been a fan of octal `\000` but presumably some folks are.

C has grown [others](https://en.wikipedia.org/wiki/Escape_sequences_in_C#Table_of_escape_sequences). No objections to including more, but not advocating for them either.

inexorabletash:
I don't have a strong need for this feature, but since I suggested it and there was positive feedback I thought I'd use it as a way to get acquainted with cc65 contributions. If there is reluctance to include this functionality, please just close out.

The PR lacks tests; happy to add them if someone can point me at examples.

### DIFF
--- a/doc/ca65.sgml
+++ b/doc/ca65.sgml
@@ -2831,6 +2831,24 @@ Here's a list of all control commands and a description, what they do:
     removing the lines with the assignments may also be an option when porting
     code written for older assemblers).
 
+  <tag><tt>string_escapes</tt><label id="string_escapes"></tag>
+
+    Allow C-style backslash escapes within string constants to embed
+    special characters. The following escapes are accepted:
+
+<itemize>
+<item><tt>\\</tt> backslash (<tt>$5C</tt>)
+<item><tt>\'</tt> single quote (<tt>$27</tt>)
+<item><tt>\&quot;</tt> double quote (<tt>$22</tt>)
+<item><tt>\t</tt> tab (<tt>$09</tt>)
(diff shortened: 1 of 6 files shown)


oliverschmidt:
First of all let me apologize for not reacting on this PR for two weeks! Everybody following this repo knows that this isn't my usual conduct. I currently just can't spend the appropriate amount of time.

The second unusual aspect about this PR is that I seem to be not able to have something to gripe about. And again - everybody following this repo will consent that *__THIS__* is indeed unusual ;-))

So given that nobody else took the opportunity to object in any way I'm going to merge this PR now.

Thanks - and again sorry for the delay!

PR #832, opened 2019-01-22
Optimize the conversions from Atari ASCII to screen codes.

PiotrFusik:
I made it shorter and faster.

### DIFF
--- a/libsrc/atari/cputc.s
+++ b/libsrc/atari/cputc.s
@@ -1,5 +1,5 @@
 ;
-; Mark Keates, Christian Groessler
+; Mark Keates, Christian Groessler, Piotr Fusik
 ;
 ; void cputcxy (unsigned char x, unsigned char y, char c);
 ; void cputc (char c);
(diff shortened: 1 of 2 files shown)


oliverschmidt:
@groessler: Please merge if you are fine with this PR.

libsrc/atari/cputc.s
```diff
@@ -30,16 +30,13 @@ L4:     cmp     #$0A            ; LF
...
-        tax
-        tya
-        and     #$9f
-        ora     ataint,x
+        asl     a
+        adc     #$c0
+        bpl     intok
+        eor     #$40
+intok:  lsr
+        bcc     cputdirect
```
greg-king5: The new code shifts a zero into the least significant bit; then, it changes the upper nybble.  Then, it shifts the least significant bit into the carry flag, and tests it.  But, that bit always is zero (because of the `asl`).

greg-king5: Ah, the original bit 7 is tested -- clever.

The shifted accumulator is manipulated, then deshifted.  Would it make sense to write those immediate operands as
```asm
adc #$60 << 1
eor #$20 << 1
```

groessler: I'd prefer that, too. Either make it `<<1` as @greg-king5 suggested, or `*2`.

pfusik: While I agree that `<<1` could be helpful, the first `adc` translates ATASCII codes $20..$5f to screen codes $00..$3f, so it would be `adc #$100-($20<<1)`, which I consider harder to follow, especially that the next thing is the sign test to correct the other codes.

I decided to add comments instead. What do you think?

groessler:
Comments are good as well.
Thanks for your contribution.

oliverschmidt:
> Thanks for your contribution.

Me too :-)

PR #833, opened 2019-01-23
Add translation from PETSCII to screen codes.

skeetor:
This PR adds include files which translate from PETSCII to screen codes.


mrdudz:
mmmh what problem does this solve?

skeetor:
I was writing a small hellow world in asm by writing directly to the screen. The characters got screwed up, because they were in PETSCII. So I tried to use the '.macpack cbm' with '.scrcode' but it gave me the same output.
There is some other advantage as well. When I plan to write only to screen anyway, I always needs screencodes. Even if scrcode works, you would have to annotate every single character or string with it. 
And something like this isn't even possible:

lda #.scrcode 'A'

So with the include, all characters automatically get converted properly. I have seen that for atari there is a similar file included already, so I thought I add this as well.

oliverschmidt:
> So I tried to use the '.macpack cbm' with '.scrcode' but it gave me the same output.

Shouldn't "someone" follow up on that statement and find out if something about .scrcode is broken or if it wasn't used correctly?

> I have seen that for atari there is a similar file included already

* This isn't true for c64_screenmap.inc.
* What about the `#pragma warn (remap-zero, push, off)` in the Atari file?
* What about the `/* No include guard here! Multiple use in one file may be intentional. */` in the Atari file?

include/c64_screenmap.h
```diff
@@ -0,0 +1,275 @@
...
+/* If this include is used, no additional macros are needed                  */
+/*                                                                           */
+/*****************************************************************************/
+
+#ifndef _C64_SCREENMAP_H
+#define _C64_SCREENMAP_H
+
+#pragma charmap ($02, $82)
+#pragma charmap ($01, $81)
+#pragma charmap ($00, $80)
```
oliverschmidt: Why is the order $02, $01, $00 ?

greg-king5:
> > So, I tried to use the '.macpack cbm' with '.scrcode'; but, it gave me the same output.
> 
> Shouldn't "someone" follow up on that statement; and, find out if something about .scrcode is broken, or if it wasn't used correctly?

I am curious about that.  `scrcode` converts PetSCII into something else.  The only way that it could have produced "the same output" is that the input was screen codes.

include/c64_screenmap.h
skeetor: I think this probalby happend when I copy&pasted it. No particular reason. I can change it if required.

skeetor:
I have updated the include guard and also reorded it. Will also attach my sample source code.

[hello_world_macpack.s.txt](https://github.com/cc65/cc65/files/2790813/hello_world_macpack.s.txt)
[hello_world_screenmap.s.txt](https://github.com/cc65/cc65/files/2790814/hello_world_screenmap.s.txt)

oliverschmidt:
@skeetor: Thanks for the adjustments :-)

The current status from my pov:

* The C header file is an obvious orthogonal extension and ready to be merged.
* The benefit of the asm inc file isn't fully clarified. If it turns out to be significant then the obvious question is if it's desirable for CBM too.

skeetor:
>     * The benefit of the asm inc file isn't fully clarified. If it turns out to be significant then the obvious question is if it's desirable for CBM too.

Not sure how many systems this affects, I only know the C64. I dabbled with C128 too, and am pretty confident that it should be the same. Can't say for others like VC20, Plus/4, etc.. Since my hello world doesn't have any system dependencies, other then the screen address and the colour (which could even be removed) I can check if the machines, supported by vice would show the same behaviour (which I assume).

skeetor:
I have the sourcecode here:
https://github.com/skeetor/c64-src/tree/master/hello_world
and here:
https://github.com/skeetor/c64-src/tree/master/hello_world_macpack

When I compile and run it on VICE I get this output.

c64_macpack
c64_scrcode

MonteCarlos:
Thought, that .charmap was for translation of the pc ASCII to petscii 
and not for translation of petscii to scrcode?

The scrcode macro on the other hand translates pc ASCII to cbm scrcode.

Now we use .charmap to map pc ASCII to c64 scrcode which breaks the 
original intention.

How you tried


lda #<text

ldy #>text

jsr $ab1e


after inclusion of that header?

MonteCarlos:
Greg-King:

Yes, the first example with .byte is definitely wrong.
Use lda.., ldy..., jsr $ab1e with .byte but lda ..., sta $0400,x with scrcode. Not vice versa.
On the C level we have no scrcode macro, so that #pragma charmap (.,.) could serve as a workaround to be able to write directly to the screen from C code. But one must be aware that it breaks stdio code.

skeetor:
> Thought, that .charmap was for translation of the pc ASCII to petscii and not for translation of petscii to scrcode? The scrcode macro on the other hand translates pc ASCII to cbm scrcode. Now we use .charmap to map pc ASCII to c64 scrcode which breaks the original intention.

I assumed .charmap to translate from whatever encoding it currently has activated. Since I use C64 as system, it uses PETSCII and this is what the charmap converts from.

> How you tried
> lda #<text
> ldy #>text
> jsr $ab1e
> after inclusion of that header?

When I do this, I get a screen which is screwed up when "GREETINGS" should appear. It interprets the bytes as some control bytes to switch to lower case, change colour etc.

MonteCarlos:
Sorry, you got the usecase of .charmap wrong, i think. 
I should have mentioned, that $ab1e needs termination 0 byte. 
Use .asciiz instead of .byte
Sorry for that.

skeetor:
> Yes, the first example with .byte is definitely wrong.

Why? It works exactly as I want it to, so how can it be "wrong"?

> Use lda.., ldy..., jsr $ab1e with .byte but lda ..., sta $0400,x with scrcode.

Even if that were the case, you still can not do

> lda #'A' 

and get the expected conversion, namely screencode.

> On the C level we have no scrcode macro, so that #pragma charmap (.,.) could serve as a workaround to be able to write directly to the screen from C code. But one must be aware that it breaks stdio code.

Well, obviously the include is for assembly anyway, and a programmer should be aware of how he uses the features. You can not use stdio with screencodes because it does something other then what might be expected.

skeetor:
> Sorry, you got the usecase of .charmap wrong, i think.
> I should have mentioned, that $ab1e needs termination 0 byte.

I was aware of that and added the 0-byte anyway.

> Use .asciiz instead of .byte

This would create an ASCII string or a PETSCII?

skeetor:
> From the screenshots i see, that your first example outputs grafic characters to the screen instead of the "hello world" text. I don't really understand your answer: "Why? It works exactly as I want it to, so how can it be "wrong"?" under these circumstances.
> The second example uses the scrcode macro and produces correct output.
> So what's the purpose of that newly introduced header?

The first example with the graphics character is the one using .macpack/screcode instead of the include. That's why the characters are wrong, which is what I'm saying all along. In the second screenshot it uses my include, so the output is as expected, being the intended text.

greg-king5:
@skeetor, you're showing the wrong font.  You're showing the UPPER-case/graphics font.  You must tap the SHIFT and Commodore keys together to switch to the lower-case/UPPER-case font.  Then, you'll see that the `scrcode` program makes the correct upper-case "GREETINGS"; and, the `.charmap` program makes an incorrect lower-case "greetings".

MonteCarlos is correct.  (I completely forgot about it.)  The "charmap" feature converts the source file's encoding (ASCII-based) into some other encoding.  The built-in charmaps convert into each target machine's encoding (which is PetSCII when `-t c64`).

`#pragma charmap()` and `.charmap` replace the built-in maps with their own translations.  Therefore, your files must translate from ASCII into screen-codes (not PetSCII into screen-codes).

skeetor:
> @skeetor, you're showing the wrong font. You're showing the UPPER-case/graphics font. You must tap the SHIFT and Commodore keys together to switch to the lower-case/UPPER-case font. Then, you'll see that the `scrcode` program makes the correct upper-case "GREETINGS"; and, the `.charmap` program makes an incorrect lower-case "greetings".

I don't know but maybe we are talking about different things.

In the attached screenshot I wrote a little sample which puts all characters on the screen ($c000).
In the memory dump you can see that $01 = 'A', $02 = 'B', etc. when writing directly to the screen.
When I write in the source code:
.byte 'ABC'
my expectation, when I copy this to the screen, is that I get exactly that. Everything else would be unintuitive, because I would have to write something totally different in the sourcecode than I can see on the screen.
And this is what the include file achieves. Maybe the naming is wrong and should rename it to 'petscci_to_screencode.inc'  to make it clearer for the user?
I'm aware that there are different encodings, so when I would use a stdout function, this gives a different result, but then, you wouldn't include this file anyway.

screencode

skeetor:
> You must tap the SHIFT and Commodore keys together to switch to the lower-case/UPPER-case font. Then, you'll see that the `scrcode` program makes the correct upper-case "GREETINGS"; and, the `.charmap` program makes an incorrect lower-case "greetings".

If this is true, then the name '.scrcode' is IMO wrong, because from the name, I would expect that scrcode emits screen code bytes, which should be suitable for being copied directly to the screen. If I write an upper case character in the source, but only get the upper case character when I switch to lower case display, then this sounds as if the macro does not what its name says.

skeetor:
> In which font? Upper/graphics or lower/upper? How should the assembler know which font will be enabled at runtime?

When you write directly to the screen, then it is the programmers responsibillity. The assembler shouldn't need to know this. When writing low level, you obviously have to take on the responsibillity that comes with that kind of coding.

.byte 'ABC' is simply a shortcut for writing .byte 1, 2, 3 which wouldn't be really maintainable.

> Especially that it seems to me that you got somewhat confused about what you "should see on the screen".

I'm not confused, because I know what I want to see on screen, when I write directly to it.
 
>This propagates to the submitted PR, and potentially to other users who may want to use your submission at some point. Dealing with screencodes is IMHO an edge-case, which – yes, has to be handled sometimes, but – for such cases there is either the macro, which can be used along with regular ASCII/PETSCII in the same file 

But that's my point. The macro can not be used for all cases, while the charmap can.

When I write:

lda #'G'
sta $0400

I would expect the same result as if I write

include screencode.inc
lda Text
sta $0400
Text: .byte 'G'

So using the macro I would write

lda Text
sta $0400
Text: scrcode 'G'

And suddenly it shows something different, because it actually shows me a '|' character instead.

And furthermore writing

lda #scrcode 'G'

isn't even possible because it gives an error, not mentioning that it is rather ugly for such a simple thing.

hello_world.s(57): Error: Unexpected trailing garbage characters

> or you can spawn your own conversion in close to no time.

Yes, of course I can, and I did. :) But AFAIK a compiler should also include stuff which is generically usefull. which this is IMO.

skeetor:
> It does exactly what it should do - translates what you type into your assembler sources: SHIFT-A, SHIFT-B, ... to what you would see on the screen of your 64 when typing the same sequences directly on it. And regardless of which font/mode you have enabled on your 64 at _any_ given moment.

Typing is very differnt from writing to screen, as it uses the standard BASIC methods. If yoou want to do this, you wouldn't include this file, as I already mentioned above.

silverdr:
>Typing is very differnt from writing to screen, as it uses the standard BASIC methods. [...]

... to place appropriate screencodes in screen memory. The very same thing you want to do yourself, bypassing ROM routines, and what the said macro helps you to achieve. The place where you confuse things is that you "expect" uppercase 'A' from your sources to be uppercase 'A' on your screen. But seem to forget that you can't have uppercase 'A' in _both_ character sets when using only _one_ screencode. Your uppercase 'A' will _always_ be visible as uppercase 'A' only in one of the possible character sets and will _always_ differ from uppercase 'A' when the other one is enabled.

>lda Text
sta $0400
Text: scrcode 'G'
And suddenly it shows something different, because it actually shows me a '|' character instead.

Which is what it should show, given a SHIFT-G character.

>And furthermore writing
>lda #scrcode 'G'
>isn't even possible because it gives an error,

Sure thing. This doesn't prevent you from building a macro that will give you the one-liner you want. Something like

    lda_scr 'G'

or whatever you find more attractive visually.

mrdudz:
just to asset it: when i write .scrcode "aB" i'd expect the result to be the same as .byte 1,$41 - and same behaviour for anything else that converts into screencode.

skeetor:
Since there doesn't seem to be any progress, I will withdraw the PR.

oliverschmidt:
Hm, I felt that there's was no prob/discussion about the C header - or did I miss the point?

I can imagine that you're probably not in the mood to do so after the discussion but in case you are you could omit the asm inc and have the C header merged.

greg-king5:
I wrote a BASIC program that does EXACTLY what your Assembly program does; it puts all screen-codes on the screen.  But, I use the lower-case/UPPER-CASE font to display the result.  It's the SAME screen-codes; but, they show different pictures.  You can see that, when that font is active, $01 = 'a', $02 = 'b', $03 = 'c', etc. (lower-case letters).  And, $41 = 'A', $42 = 'B', $43 = 'C', etc. (upper-case letters).  The `scrcode` macro is designed to handle lower-case, as well as upper-case.  Therefore, it expects that a user will use the lower-case/UPPER-CASE font instead of the UPPER-CASE/graphics font (the graphics font doesn't have any lower-case letters).

As I wrote earlier, the charmap files must translate from ASCII, not from PetSCII.  And, they must handle lower-case also (the current files can't do it because they work with a font that can't show lower-case).

screenshot

skeetor:
> 
> 
> Hm, I felt that there's was no prob/discussion about the C header - or did I miss the point?
> 
> I can imagine that you're probably not in the mood to do so after the discussion but in case you are you could omit the asm inc and have the C header merged.

I removed the asm file now and only added the include file. Seems I understand the problem now a bit better.

greg-king5:
I attached a C program that puts all 256 -- converted -- ASCII (plus extra) characters on the screen.  You can use it to debug the C header.  (cc65's system library automatically enables the lower-case/UPPER-CASE font.  So, you will see the codes as we are meant to see them.)
[charmap-demo.c.gz](https://github.com/cc65/cc65/files/2806873/charmap-demo.c.gz)

You should test your files before you submit a PR.  The current header cannot be compiled because it has Assembly hex numbers instead of C hex numbers (`$AB` instead of `0xAB`).

Here is a screen-shot.  You can see that the alphabets have swapped cases.

ascii2screen

PR #838, opened 2019-01-26
Add asm config for VIC-20 base version

skeetor:
This PR adds an asm config for the VIC-20 base version, analogous to the c64-asm.cfg.


oliverschmidt:
* Why isn't there a line like https://github.com/cc65/cc65/blob/master/cfg/c64-asm.cfg#L15
* As far as I understand it's desirable to have `DATA` and `RODATA` defined with `optional = yes` in asm configs.

oliverschmidt:
No feedback whatsoever after more than a week. Likely lost interest in contributing to cc65. Closing...

PR #847, opened 2019-02-16
Move Atari-specific PIA reg vals to atari.h

billkendrick:
PIA is also used in the Commodore PET, and for different
purposes (see http://www.6502.org/users/andre/petindex/progmod.html#pia1),
so extracted Atari-specific register #defines and placed them in atari.h.


groessler:
Merged. Thanks.

PR #848, opened 2019-02-18
Adds linker configuration files for Atari Binary format of C programs.

dmsc:
A followup from #762 , this adds:
- Support for INITAD to memory areas in XEX files,
- Splits headers from system-check chunk,
- Uses the above to add new configuration files for C programs using Atari binary output.


dmsc:
Fixed using "`-Wl -D__SYSTEM_CHECK__=1`" as a means to remove the system check chunk. Tested and works in both original linker config and xex linker configs.

testcode/lib/atari/multi-xex.s
```diff
@@ -14,8 +14,25 @@
...
+        sta     ICBAH,x
+        lda     #$FF
+        sta     ICBLL,x         ; length
+        sta     ICBLH,x
+        lda     #PUTREC
+        sta     ICCOM,x
+        jmp     CIOV
+
+msg_load:
+        .byte   "Loading....", $9B
```
groessler: You could use ATEOL here instead of $9B.

groessler:
Looks good in general.
One thing: To be more in order with other linker config names, could you rename the XEX config files which are for "C" to not have "-c-" in their names? So, rename `atari-c-xex.cfg` to `atari-xex.cfg`, `atarixl-c-xex.cfg` to `atarixl-xex.cfg`, and the existing `atari-xex.cfg` to `atari-asm-xex.cfg`.

testcode/lib/atari/multi-xex.s
dmsc: Ok, changed.

dmsc:
> Looks good in general.
> One thing: To be more in order with other linker config names, could you rename the XEX config files which are for "C" to not have "-c-" in their names? So, rename `atari-c-xex.cfg` to `atari-xex.cfg`, `atarixl-c-xex.cfg` to `atarixl-xex.cfg`, and the existing `atari-xex.cfg` to `atari-asm-xex.cfg`.

Renamed and added more documentation in a new commit.

groessler:
Thanks.

dmsc:
> Thanks.

Thanks to you, for the review and merge.

Issue #696, opened 2018-07-05
kbhit and mouse driver - c128 target


xlar54:
Seems the "1" and "2" keys are not detected when the kbhit is called and the mouse driver is loaded.

xlar54:
As a temp fix, Im PEEKing location 212 to get the pressed keycode for keys 1 and 2.  But within a loop, this can register multiple times.

oliverschmidt:
cc65 is a multi target tool chain. Therefore it might make sense to provide the machine you are experiencing the issue on.

xlar54:
my apologies.. This issue was for the c128 target.  I havent tested the 64 target.

oliverschmidt:
Hm, I was under the impression that these type of issues were solved with https://github.com/cc65/cc65/pull/114 ...

polluks:
@xlar54 "3" works? "1" and "2" are some kind of special keys in the matrix.

mrdudz:
also which mouse driver would be interesting... does this happen also with the regular 1531 driver?

i wouldnt call 1 and 2 "special" :) they are just used in parallel with joystick port 1 - some interferences are expected. however when the mouse is not being moved, nor mouse buttons are pressed, then those keys should work.

xlar54:
Yeah, the normal std 1351 driver.  I have a work around, scanning for the keys via PEEK to 212.  If this isnt something that cam be fixed, then my solution will get me through.  I realize that some of these things are just the way they are as far as the hardware is concerned (Id have to do the same in assembly or basic, etc).  Thank you all for being responsive.  Im writing a web browser for the 128 and its coming along very nicely thanks to your work on cc65.    https://www.reddit.com/r/Commodore/comments/8w02m0/rhml_browser_is_coming_along/?st=jj6v0n5k&sh=58a33a17

mrdudz:
as said, if it also happens when you dont touch the mouse (press buttons) then its clearly a bug. i think this "exorcism code" does too much, it disables some CIA line completely - which probably removes the phantom keys, by disabling those keys alltogether :)

polluks:
@greg so we need the exorcist again :wink:

mrdudz:
the proper solution is, btw, to a) toggle port A DDR between in/out (in for joystick,out for keyboard) and b) skip keyboard input when joystick input was detected while port A DDR was input.

xlar54:
[...]

PR #853, opened 2019-03-07
C128 Mouse-keyboard fix

xlar54:
This appears to fix bug [696](https://github.com/cc65/cc65/issues/696).

reverts code back to c64's version of the 1351 mouse driver.  Didnt experience any issues in my RHML application, all keys worked fine.  test code worked fine as well (see 696 for test example)


xlar54:
guys not sure why 814 was included in this pull request and not sure how to fix.

libsrc/c128/mou/c128-1351.s
```diff
@@ -5,6 +5,24 @@
...
+													  
+ 
+																	   
+																	 
+																	   
+																		
+																		
+																	 
+																 
+
```
polluks: ???

xlar54: I blame the IDE.  lol.. but can fix.  Thanks polluks

libsrc/c128/mou/c128-1351.s
```diff
@@ -77,40 +95,33 @@ Vars:
...
-old_irq:        .res    2
-
 .rodata
 
 ; Default values for above variables
 ; (We use ".proc" because we want to define both a label and a scope.)
 
 .proc   DefVars
-        .byte   0, 0                    ; OldPotX/OldPotY
+        .byte   0, 0                    ; OldPotX/OldPotY
```
spiro-trikaliotis: Adding whitespace before a NL?

libsrc/c128/mou/c128-1351.s
```diff
@@ -202,20 +163,9 @@ INSTALL:
 
 ;----------------------------------------------------------------------------
 ; UNINSTALL routine. Is called before the driver is removed from memory.
-; No return code required (the driver is removed from memory on return).
+; No return code required (the driver is removed from memory on return).
```
spiro-trikaliotis: Again adding whitespace at end of line.
If your IDE does this, it is clearly broken. That's why I hated Eclipse years ago when I had to use it.

libsrc/c128/mou/c128-1351.s
```diff
@@ -270,14 +220,14 @@ GETBOX: sta     ptr1
...
-        sei
 
+		
 @L1:    lda     XMin,y
         sta     (ptr1),y
         dey
         bpl     @L1
 
-        cli
+
```
spiro-trikaliotis: Blank line with whitespace

groessler: emacs 'delete-trailing-whitespace'

libsrc/c128/mou/c128-1351.s
xlar54: Notepadd++... used a compare plugin that apparently went stupid.  Sorry for the issue

oliverschmidt:
@greg-king5 @groessler - I need help to assess this PR !

I understand that there's "crosstalk" between the keyboard and the mouse.

Back in 2010 @greg-king5 solved this issue for the C64 with https://github.com/cc65/cc65/commit/5f18ddbe (I added his explanation with https://github.com/cc65/cc65/commit/9ef9396b).

In fact I don't understand until today if this approach would be applicable to the C128 as well.

Then in 2014 @groessler "workarounded" this issue for the C128 with https://github.com/cc65/cc65/commit/608dbc2cec5e9cc7e9e9b96c139d838837c7df3b, https://github.com/cc65/cc65/commit/4406307c2fee0e3e2369d03e38bf8c118c2393ac, https://github.com/cc65/cc65/commit/54be6de9bc0afa61b4f15e1f651832a614176852 and https://github.com/cc65/cc65/commit/6307815ada83e09589864fe218baab2c907e1641.

@greg-king5 then applied that workaround (incl. some optimizations) to more C128 drivers with https://github.com/cc65/cc65/commit/667f8229c02ceaaf94c2c2b169079e0eddea3ef0 and added his optimizations to the drivers workarounded by @groessler before with https://github.com/cc65/cc65/commit/7d512ca238e14cbbeb2e8e2879271f2488461ea4.

Now I'm confronted with a PR that asks to revert code back to some state before for seemingly the same reasons the "workaround" was introduced for in the first place.

Sorry guys, but I'm totally lost :-(

mrdudz:
first, the keyboard and joystick works exactly the same on c64 and c128 (exception is the numblock, which uses some extra matrix lines connected to d02f in the VICII) that means all drivers touching the keyboard lines or joystick lines can (and should, imho) be shared between the two. no need for code duplication there, a few ifdefs here and there should be enough.

then, its important to understand the root problem - which is the joystick lines being connected in parallel with the keyboard matrix lines. there are no diodes or anything preventing one messing up the other. there is no perfect fix for this, only workarounds that work somewhat, sometimes more, sometimes less, depending on the exact usecase.

on top of that it should be noted that the exact behaviour is NOT emulated. DONT rely on testing in emulators for this. AT ALL. also make sure to understand that switching to 2Mhz will change the behaviour slightly (again: probably not in the emulator) :)

my proposal would be: make a test program first, which uses the conio functions to read the keyboard and print whats typed, and which uses the mouse driver at the same time and also shows the button state of the mouse. this should demonstrate the problem. now test it - on a real c128 - and document the exact problem. now a workaround could be made that removes (part of) the problem. the same workaround should then be applied to both, the c64 and c128 version (which ideally is the same thing).

that said, both the old and new code look like valid attempts for a workaround to me.

groessler:
> my proposal would be: make a test program first, which uses the conio functions to read the keyboard and print whats typed

This is more or less already available with ``samples/mousedemo.c`` and ``testcode/lib/mouse-test.c``.

> on top of that it should be noted that the exact behaviour is NOT emulated. DONT rely on testing in emulators for this.

I had tested on a real C64 (not really, but a C128 in C64 mode)

Take the rest with a grain of salt. This is from memory and my memory has a few sticky bits.

I had noticed in "mouse-test" or "mousedemo" for C64 that I'm getting many "phantom key presses" when I move the mouse. Therefore I've implemented a fix for C64. In C128 mode the problem was not that obvious, less "phantom keys", but still not "none". Therefore, and because I know more about the C64 than the C128, I left the adaptation of the C128 driver to somebody else.
But I think I also tested the C128 driver then on real hardware.

groessler:
> This is from memory and my memory has a few sticky bits.

Looking at the refs @oliverschmidt posted, my memory seems to have quite some more than "few" sticky bits... :-(

Sorry, I need to dig deeper into git history.

groessler:
Ok, sorry. How memory can play tricks.

I had implemented the C128 workaround for 1351, and let others do the change of the other drivers.

So fine with me if the new version works on real hardware. I hadn't gotten the similar-to-C64 workaound work for me, but this could just be because - as I said - I'm not so deep into C128.

mrdudz:
i admit i havent looked at the examples for a while :) if they use the mouse and read keyboard at the same time, its sufficient, i guess.

however, *moving* the mouse should not clash with the keyboard *at all* - that uses the POTX/Y lines, which are completely unrelated to keyboard. if that happens, something is fundamentally wrong. you should only ever get problems with keyboard when mouse buttons are pressed. i'd comment out everything dealing with CIA in the driver and make sure that is the case first. then the button check can be added.

that said, the driver must be properly chained with the system irq (which does the keyboard scanning). it should probably not modify the DDR, ie leave it as it was before entering the driver. when system irq can trigger during execution of the mouse driver, all sorts of weird things are expected (and there is no way to change that, imho).

groessler:
> however, moving the mouse should not clash with the keyboard at all 

maybe it were just mouse clicks. Don't remember exactly and would need to test again on hw.

xlar54:
I can only say that in my work, the 1 and 2 keys (and others on that line like spacebar) would not work whenever the mouse driver was loaded and running.  The mouse didn’t need to be moving nor clicks.

Sent from Mail for Windows 10

silverdr:
FWIW:
* On a C64 keyboard and joystick share the same hardware lines
* "true" aka "proportional" mouse (like 1351 and compatibles) shares a subset of hardware lines with joystick (buttons) hence with keyboard too
* there is a class of "non-true" aka "non-proportional" mice, which emulate joystick directions upon movement of the mouse and therefore utilise all the same lines as joysticks do and therefore interfere with keyboard scanning the same way as joysticks

Software, which intends to use joystick or mouse (either "true" or "non-true") must not rely on standard keyboard scanning while reading mouse/joystick (buttons for "true" mouse, buttons and position for "non-true" mouse, buttons and directions for joystick) connected to CONTROL PORT 1. When not observing the above rule, each of the mentioned input devices will generate fake "keypresses", with "true" mouse generating a subset of those, which "non-true"/joystick does. The only way around the issue is a non-standard key scanning routine, which excludes affected lines' state combinations from being read as keypresses.

silverdr:
@mrdudz this is a simple approach to avoid fake keypresses being registered, but - as you noticed - adds a side-effect of the whole keyboard not being usable during buttons/movements, while not solving the problem of non-working real keypresses (affected keys will be read as joystick/mouse inputs and not passed further down). In any case, there is no full, known to me, solution to this hardware design related issue. Some keys are either disabled or "fake pressed" by input device event.

xlar54:
Could a hybrid solution be done where the line that isnt read from the keyboard is checked during the mouse routine?  Instead of creating a completely new keyboard scanning routine?  Or would you guys consider this just a poor hack?

mrdudz:
as silverdr said, there is no way to fix this 100%. you'll have to make the right compromise for your specific usecase

silverdr:
@xlar54 for my own software (I wrote quite a few mouse-enabled ones back then) I always scanned keyboard myself when using mouse/joystick in CP1. For a generic/library use, I surely wouldn't like to add this kind of overhead. What @mrdudz proposed is probably the most lightweight approach in such case.

groessler:
> what could be done is...

IIRC, that's more or less the same what the current implementation does. But it doesn't "skip the keyboard" scanner, but runs it, and removed the (maybe) detected keypress afterwards. Appears to be similar semantics to me.

groessler:
Any news?

groessler:
Any news question #2?

greg-king5:
This thread is dormant until we hear from @xlar54 about the much simpler fix that I posted in https://github.com/cc65/cc65/issues/696#issuecomment-477160153.

xlar54:
Ill try it, thanks Greg

xlar54:
Greg was right.  Simple fix, solved the issue.  If all else is ok, Id like to get these commits into the next release.

xlar54:
Not sure what the travis check was - Im new to that.

greg-king5:
> Not sure what the Travis check means -- I'm new to that.

You need to remove **all** of the old changes that you put into this PR.  Those lines must be fixed in the _**original**_ driver code.  ("Original" means "before this pull request".)

Perhaps, you should start over from the beginning:
1. Close this PR.
1. Delete your "mouse-keyboard-fix" branch on Github.
1. Then, in the repository on your local machine, go to the master branch.
1. Delete the "mouse-keyboard-fix" branch.  (You should use the upper-case `-D` option.)
   1. Delete its remote-tracking branch, too.
1. Make a new "mouse-keyboard-fix" branch -- but, make it from "upstream/master" (your "master" branch has extra commits that we don't want in the PR).
1. Fix that line in _all_ of the files in "libsrc/c128/mou/", in the new branch.  Commit it.
1. Push it to Github.
1. Make a new pull request.

oliverschmidt:
https://sourceforge.net/p/cc65/mailman/message/37056962/

greg-king5:
I'll finish this fix.

oliverschmidt:
Thanks in advance :-)

greg-king5:
@xlar54, you can delete your "mouse-keyboard-fix" branches now.

xlar54:
sorry, dropped the ball on this one.  thank you greg-king5

PR #854, opened 2019-03-08
Atari OS: Reestablished DCB structure, fixed union span and timer2.

IrgendwerA8:


include/_atarios.h
```diff
@@ -540,7 +558,7 @@ struct __os {
     // --- Page 3 ---
 
     union {     
-        unsigned char dcb[0x40];            // = $0300-$03XX    DEVICE CONTROL BLOCK
+        DCB dcb;                            // = $0300-$030B    DEVICE CONTROL BLOCK
         struct {        
             unsigned char ddevic;           // = $0300          PERIPHERAL UNIT 1 BUS I.D. NUMBER
```
greg-king5: You shouldn't declare two structures that define the same objects.  "`DCB dcb;`" is enough.  If you want to give direct access to the high/low bytes of those two words, then you should move the unions into `struct __dcb`.  (Then, the high byte of the address could be named "`OS.dcb.buffer_hi`".)

IrgendwerA8: The problem is that the names in the struct aren't official and I like to offer the official ones. Therefore I removed the __dcb struct.
But since @groessler requested (due to compatibility reasons?) the old definition, I have no idea to resolve this conflict in any other way...

groessler: > The problem is that the names in the struct aren't official and I like to offer the official ones.

The struct had been defined back then in an ad-hoc way. So I agree that they could be changed to the official names. 

But I don't see the point of having unions for the 16-bit fields. Can you point me to an use case where the byte-variants are needed in C code?

> But since @groessler requested (due to compatibility reasons?) the old definition

I want to have the structure itself as its own entity, at least. I can live with renaming the fields to their official names.

IrgendwerA8: >But I don't see the point of having unions for the 16-bit fields. Can you point me to an use case where the byte-variants are needed in C code?

I guess the 16 bit to 8 bit use case is maybe theoretical and the byte access is not a huge gain (e.g. modify the high byte to target a buffer on another page), but examine the aux1/2 definition: Setting up the sector for a disk read by using the 16 bit definition "daux" seems more comfortable to me.

>I want to have the structure itself as its own entity, at least. I can live with renaming the fields to their official names.

The names in the structure are not bad. Changing them to the OS names and removing the explicit memory locations from the OS structure would mean that a direct access would need the dcb-"indirection": `OS.dcb.ddevic` instead of `OS.ddevic`.
I'm not sure if this is really better compared to the current state of the PR...
(In case of the `IOCB` things are clearer because of the multiple instances in address space.)

groessler: > but examine the aux1/2 definition: Setting up the sector for a disk read by using the 16 bit definition "daux" seems more comfortable to me.

Why did you define `daux` as `int` in the first place? Where is it used as 16-bit value?

Regarding `struct __dcb`, I want to be able to use it like this (in existing local code):
```
    memset(&DCB,0,sizeof(DCB));
    ...
    DCB.command = 0x4e;
    ...
```

Also, one could have multiple DCBs in memory, and just memcpy the one to use to the OS location. To me, the 'DCB'  identifier appears more informative, than the 'OS' identifier. 'OS' is too general there.

groessler: > Why did you define daux as int in the first place? Where is it used as 16-bit value?

Re-read your original message. Got it.

IrgendwerA8: >Regarding struct __dcb, I want to be able to use it like this (in existing local code):
>Also, one could have multiple DCBs in memory, and just memcpy the one to use to the OS 

Fully understood and valid points... and that's why I find the current PR appropriate:  You have the best of both worlds. Compatibility, flexibility and the familiar OS locations.

groessler: > I can live with renaming the fields to their official names.

I guess I wasn't expicit enough. I agree with Greg that we shouldn't have two different structs which describe the same thing (a DCB) just using different member names.

So please get rid of the anonymous struct in OS and change 'struct __dcb' to use the official names. And in turn use this to define the DCB in the OS struct.

Also, I'm seeing another loss in your PR. Previously there were
```
#define DCB (*(struct __dcb *)0x300)
#define ZIOCB (*(struct __iocb *)0x20)  /* zero page IOCB */
#define IOCB (*(struct __iocb *)0x340)  /* system IOCB buffers */
```

IOCB and DCB are now typedef names, which don't refer to specific address (AFAICT). Maybe rename the typedef names to something with "_t" at the end, like "IOCB_t". Or all lowercase (iocb_t). For example, in "size_t" the "size" part isn't uppercased as well.

And then reinstate definitiions of IOCB, ZIOCB, DCB to point to their OS locations.

include/_atarios.h
```diff
@@ -569,39 +587,40 @@ struct __os {
...
-            unsigned char pupbt3;           // = $033F          ##1200xl## 1-byte power-up validation byte 3
-#endif
         };
     };
+    unsigned int timer1;                    // = $030C/$030D    INITIAL TIMER VALUE
+#ifdef OSA                      
+    unsigned char addcor;                   // = $030E          ##old## ADDITION CORRECTION
+#else                       
+    unsigned char jmpers;                   // = $030E          ##1200xl## 1-byte jumper options
+#endif
```
groessler: Regarding the `#ifdef OSA`: I understand that you added it for completeness. And that's fine.

But how would one use these definitions? I guess nowadays no one in his right mind would write a program which only works on the OSA ROM version. So the program would have to to adapt at runtime to the OS version.

Wouldn't it be better to make unions of the OSA and later version definitions? In order that both the OSA and later definitions are readily at hand?

include/_atarios.h
greg-king5: > Regarding `struct __dcb`, I want to be able to use it like this (in existing local code):
> ```c
> memset(&DCB,0,sizeof(DCB));
> ...
> DCB.command = 0x4e;
> ...
> ```

That code would look like this:
```c
memset(&OS.dcb, 0, sizeof OS.dcb);
...
OS.dcb.dcomnd = 0x4e;
...
```

> Also, one could have multiple DCBs in memory, and just memcpy() the one to use to the OS location. To me, the 'DCB' identifier appears more informative than the 'OS' identifier. 'OS' is too general there.

That would look like this:
```c
DCB my_dcbs[4];
...
/* Use one of my Device Control Blocks. */
OS.dcb = my_dcbs[n];
...
```
Those expressions look very "readable" and informative to me.

IrgendwerA8: >Maybe rename the typedef names to something with "_t" at the end, like "IOCB_t". Or all lowercase (iocb_t). For example, in "size_t" the "size" part isn't uppercased as well.

I could do that (`iocb_t, dcb_t, devhdl_t` etc.) - even I'm not a big fan of it.

>And then reinstate definitiions of IOCB, ZIOCB, DCB to point to their OS locations.

I wouldn't do that. Zero page IOCB is now `OS.ziocb` which I find perfectly fine and I don't see any good reason to introduce an second symbol for that. Then someone could also argument to revive the shadow register define names - making the whole OS structure approach superfluous.

To progress with this PR (as the current master isn't even correct) here a summary what I would change as outcome of the discussion:
* making the uppercase type names "lowercase_t"
* change the DCB structure to use official names and introduce a union for aux1/2
* let the OS structure only contain the DCB as substructure and not reflect the singular locations (access to DCB values is than always `OS.dcb`)

Can we agree on this proposal?

include/_atarios.h
IrgendwerA8: > I understand that you added it for completeness.
Yes, that was exactly my motivation.

>Wouldn't it be better to make unions of the OSA and later version definitions?
I don't think so. Currently you actively have to decide to service OSA. Offering a union would allow  usage by accident (e.g. wrong documentation or false understanding of the OS model.)

include/_atarios.h
groessler: > Can we agree on this proposal?

Fine with me.

IrgendwerA8: Done. I guess @greg-king5 is fine with the current state too.

greg-king5: :+1:

groessler:
@IrgendwerA8: I think it should be this way:

```
--- a/include/_atarios.h
+++ b/include/_atarios.h
@@ -150,12 +150,12 @@ struct __fpreg {
 
 typedef struct __fpreg fpreg_t;
 
-enum {                          /* enum for access of floating point registers */
+enum FPIDX {                    /* enum for access of floating point registers */
     R0 = 0,                     /* (to use as index) */
     RE = 1,
     R1 = 2,
     R2 = 3
-} FPIDX;
+};
 
 
 /* Define a structure with atari os register offsets */
```

Otherwise a variable FPIDX will be created in all files including `atari.h`.
Do you agree?

IrgendwerA8:
> Do you agree?

Of course you are right. Fixed in the new PR.

PR #855, opened 2019-03-13
AtariOS: Fix for FPIDX enum

IrgendwerA8:


include/_atarios.h
```diff
@@ -150,12 +150,12 @@ struct __fpreg {
 
 typedef struct __fpreg fpreg_t;
 
-enum {                          /* enum for access of floating point registers */
+enum FPIDX {                    /* enum for access of floating point registers */
```
greg-king5: Shouldn't it be `fpidx_t` or `fpidx_e`?

groessler: The only other include file in cc65 which uses `enum`s is `geos/gdlgbox.h`. And it uses an anonymous `enum`.
So I suggest to make this `enum` anonymous as well, or use `#define`s.

I generally anyway don't see the benefit of `enum`s in C code (C++ is a different story). One might have better debugging information with `enum`, but with `#define`s their presence can be detected with `#ifdef`. The latter probably is not relevant here, and for the former I don't know if cc65 debugging information does contain `enum` definitions.

Regardless what you do, please rename the values to something like `FPIDX_Rn` or similar, in order not to pollute the name space. Something like `Rn` could very well be used by an user program for something different.

IrgendwerA8: Done.

groessler:
Thanks!

PR #857, opened 2019-03-14
Add an optimization for an indexed load from a pointer literal.

### LINKED ISSUE #856: Load from array via constant pointer not optimized
Declare an array at an absolute address (via macro, but doesn't have to be, also could be a char*):
```c
#define updbuf ((unsigned char[256])0x100)
```
Then, do a load/store:
```c
unsigned char i,y;
void main(void) {
    y = updbuf[i];
    updbuf[i] = y;
}
```
The load could be optimized to "lda $100,y"; but, it is not:
```asm
        ldx     #$01
        lda     #$00
        ldy     _i
        sta     ptr1
        stx     ptr1+1
        lda     (ptr1),y
        sta     _y
        ldy     _i
```
But, the store is optimized:
```asm
        lda     _y
        sta     $0100,y
```
If `updbuf[]` is declared as a regular array, both cases are optimized.

sehugg:
Closes #856. Optimizes expressions of the form: 
```c
unsigned char i, y;
y = ((unsigned char*)0x100)[i];
```
Also optimizes 16-bit masked array look-ups of the form:
```c
arr16[i & mask];  // where mask < 128
```

### DIFF
--- a/src/cc65/coptptrload.c
+++ b/src/cc65/coptptrload.c
@@ -1554,3 +1554,116 @@ unsigned OptPtrLoad18 (CodeSeg* S)
 
 
 
+unsigned OptPtrLoad19 (CodeSeg* S)
+/* Search for the sequence:
+**
+**	and	#mask		(any value < 128)
+**      jsr     aslax1/shlax1
+**      clc
+**      adc     #<(label+0)
+**      tay
+**      txa
+**      adc     #>(label+0)
+**      tax
+**      tya
+**      ldy     #$01
+**      jsr     ldaxidx
+**
+** and replace it by:
+**
+**	and	#mask		(remove if == 127)
+**	asl
+**	tay
+**	lda	label,y
+**	ldx	label+1,y
+*/
+{
+    unsigned Changes = 0;
+    unsigned I;
+
+    /* Walk over the entries */
+    I = 0;
+    while (I < CS_GetEntryCount (S)) {
+
+        CodeEntry* L[11];
+
+        /* Get next entry */
+        L[0] = CS_GetEntry (S, I);
+
+        /* Check for the sequence */
+        if (L[0]->OPC == OP65_AND                               &&
+            L[0]->AM == AM65_IMM                                &&
+            CE_HasNumArg (L[0]) && L[0]->Num <= 127             &&
+            CS_GetEntries (S, L+1, I+1, 10)                     &&
+            L[1]->OPC == OP65_JSR                               &&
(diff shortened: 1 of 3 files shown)


oliverschmidt:
Any comments?

mrdudz:
i find this quite interesting, but i'm too busy with other stuff right now to actually look at it (sorry)

greg-king5:
Please expand the TAB characters in your patches.

oliverschmidt:
@greg-king5: Thanks for the detailed review!

src/cc65/coptptrload.c
```diff
@@ -1457,3 +1457,219 @@ unsigned OptPtrLoad17 (CodeSeg* S)
...
+            /* allocate Label memory */
+            Label = xmalloc (Len);
+            /* lda label,y */
+            Label = memcpy (Label, L[4]->Arg+2, Len-3);
+            Label[Len-3] = '\0';
+            X = NewCodeEntry (OP65_LDA, AM65_ABSY, Label, 0, L[10]->LI);
+            CS_InsertEntry (S, X, IP++);
+
+            /* ldx label+1,y */
+            Label = memcpy (Label, L[4]->Arg+2, Len-3);
```
greg-king5: That second `memcpy()` isn't needed anymore.  The `Label` still exists exactly as the next line needs it to be.

src/cc65/coptptrload.c
```diff
@@ -1457,3 +1457,219 @@ unsigned OptPtrLoad17 (CodeSeg* S)
...
+            CS_InsertEntry (S, X, IP++);
+
+            /* tay */
+            X = NewCodeEntry (OP65_TAY, AM65_IMP, 0, 0, L[2]->LI);
+            CS_InsertEntry (S, X, IP++);
+
+            /* allocate Label memory */
+            Label = xmalloc (Len);
+            /* lda label,y */
+            Label = memcpy (Label, L[4]->Arg+2, Len-3);
```
greg-king5: I like the style that you had here before.  `xmalloc()` was the first argument of `memcpy()`; and, `Label` was assigned from the return value of `memcpy()`.  It's compact.
[Also, it's the way that `OptPtrLoad3()` and `OptPtrLoad4()` use `xmalloc()`.]

oliverschmidt:
@sehugg: @greg-king5 asked for two more adjustments two weeks ago. Does it makes sense to presume that you'll take care of them anytime soon?

sehugg:
Got 'em in there, pending CI checks.

greg-king5:
Thanks for your first patch contribution to the project.  Merging now.

PR #861, opened 2019-03-20
First proposal to get header for contributors and license info straight.

IrgendwerA8:
You may remember the discussion regarding the file headers?
If we can agree on a header I would browse through the files and change them.

### DIFF
--- a/CREDITS
+++ b/CREDITS
@@ -1,0 +1,51 @@
+
+This file lists the names of contributors.
+Without their work, cc65 wouldn't be what it currently is.
+Feel free to add your name here if you contributed to cc65, too.
+
+
+Many special thanks go to the guy who started it all:
+
+            John R. Dunning
+
+...and the person who advanced cc65 to a new level:
+
+            Ullrich von Bassewitz
+
+...and the current main maintainer
+
+            Oliver Schmidt <ol.sc@web.de>
+
+
+...and the other important persons (in alphabetical order):
+
+
+Eric Au                         <eric_au@hotmail.com>
+Craig Bruce
+Sidney Cadot                    <sidney@ch.twi.tudelft.nl>
+Piotr Fusik                     <P.Fusik@elka.pw.edu.pl>
+Keith W. Gerdes                 <kwg@netzero.net>
+groepaz                         <groepaz@gmx.net>
+Christian Groessler             <cpg@aladdin.de>
+Stefan A. Haubenthal            <polluks@web.de>
+Shawn Jefferson                 <jefferson_shawn_a8bit@yahoo.com>
+Stephen L. Judd
+Karri Kaksonen                  <karri@sipo.fi>
+Mark Keates                     <Mark.Keates@dendrite.com> 
+Greg King                       <gngking@erols.com>
+Michael Klein                   <michael.klein@puffin.lb.shuttle.de>
+Christian 'Irgendwer' Krueger
+David Lloyd                     <dmlloyd@atari-central.com>
+MagerValp                       <magervalp@cling.gu.se>
+Mirco Miranda                   <mircomir@libero.it>
+Freddy Offenga                  <taf_offenga@yahoo.com>
+Marc 'BlackJack' Rintsch        <blackjack@civitas64.de>
+Kevin Ruland                    <kevin@rodin.wustl.edu>
+Bastian Schick
+Steve Schmidtke                 <steve_schmidtke@hotmail.com>
+Josef Soucek                    <josef.soucek@ct.cz>
+Peter Trauner                   <peter.trauner@utanet.at>
+Maciej Witkowiak                <ytm@friko.onet.pl>
+
+
+Please see also https://github.com/cc65/cc65/graphs/contributors
--- a/src/sim65/6502.c
+++ b/src/sim65/6502.c
@@ -4,31 +4,13 @@
 /*                                                                           */
 /*                           CPU core for the 6502                           */
 /*                                                                           */
+/* This file is part of                                                      */
+/* cc65 - a freeware C compiler for 6502 based systems                       */
 /*                                                                           */
+/* https://github.com/cc65/cc65                                              */
 /*                                                                           */
-/* (C) 2003-2012, Ullrich von Bassewitz                                      */
-/*                Roemerstrasse 52                                           */
-/*                D-70794 Filderstadt                                        */
-/* EMail:         uz@cc65.org                                                */
-/*                                                                           */
-/* Mar-2017, Christian Krueger, added support for 65SC02                     */
-/*                                                                           */
-/* This software is provided 'as-is', without any expressed or implied       */
-/* warranty.  In no event will the authors be held liable for any damages    */
-/* arising from the use of this software.                                    */
-/*                                                                           */
-/* Permission is granted to anyone to use this software for any purpose,     */
-/* including commercial applications, and to alter it and redistribute it    */
-/* freely, subject to the following restrictions:                            */
-/*                                                                           */
-/* 1. The origin of this software must not be misrepresented; you must not   */
-/*    claim that you wrote the original software. If you use this software   */
-/*    in a product, an acknowledgment in the product documentation would be  */
-/*    appreciated but is not required.                                       */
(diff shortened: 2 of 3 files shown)


mrdudz:
I like this approach :)

src/sim65/6502.c
```diff
@@ -4,31 +4,13 @@
...
-/* including commercial applications, and to alter it and redistribute it    */
-/* freely, subject to the following restrictions:                            */
-/*                                                                           */
-/* 1. The origin of this software must not be misrepresented; you must not   */
-/*    claim that you wrote the original software. If you use this software   */
-/*    in a product, an acknowledgment in the product documentation would be  */
-/*    appreciated but is not required.                                       */
-/* 2. Altered source versions must be plainly marked as such, and must not   */
-/*    be misrepresented as being the original software.                      */
-/* 3. This notice may not be removed or altered from any source              */
```
spiro-trikaliotis: IANAL, but does this change really qualify as "This notice may not be removed or altered from the source distribution"?

Yes, I see that the LICENSE file is there. But, from what I understood from Uz' intentions, he wanted to make sure that this notice is retained even if someone takes only one source file (and forgets to create a "whole distribution")

Unless someone more knowledgeable in the legal aspects of licenses confirms otherwise, I would not remove these notices from the individual files.

But YMMV.

IrgendwerA8: Good point. I've contacted Uz how he sees this issue.

CREDITS
```diff
@@ -0,0 +1,51 @@
...
+
+Many special thanks go to the guy who started it all:
+
+            John R. Dunning
+
+...and the person who advanced cc65 to a new level:
+
+            Ullrich von Bassewitz
+
+...and the current main maintainer
```
oliverschmidt: Missing colon.

CREDITS
```diff
@@ -0,0 +1,51 @@
...
+Marc 'BlackJack' Rintsch        <blackjack@civitas64.de>
+Kevin Ruland                    <kevin@rodin.wustl.edu>
+Bastian Schick
+Steve Schmidtke                 <steve_schmidtke@hotmail.com>
+Josef Soucek                    <josef.soucek@ct.cz>
+Peter Trauner                   <peter.trauner@utanet.at>
+Maciej Witkowiak                <ytm@friko.onet.pl>
+
+
+Please see also https://github.com/cc65/cc65/graphs/contributors
```
oliverschmidt: Missing linefeed.

IrgendwerA8:
The discussion was part of this PR:
https://github.com/cc65/cc65/pull/791

I'll wait with the changes until I have feed-back from Uz.

IrgendwerA8:
Got a first answer from Uz. Doesn't look good.
Sent him some additional questions back - will report here.
I even find point 2 now interesting:

`/* 2. Altered source versions must be plainly marked as such, and must not   */
/*    be misrepresented as being the original software.  `

So are "we" now the original software (why?) or is a "mark" missing?

...and to play the fool:

`3. This notice may not be removed or altered from any source`

We just could preserve that sentence to fulfill it... ;)

oliverschmidt:
I must admit that you already totally lost me at this point of the process :-(

- I see that there's a certain text present in many files.
- The very same text is present in https://github.com/cc65/cc65/blob/master/LICENSE.
- This pull request wants to replace the text in the many files with a reference to https://github.com/cc65/cc65/blob/master/LICENSE.

So my stupid(?) question is why this pull request ends up in a discussion about that text.
- I don't see any momentum whatsoever to change that text.
- If there would be a momentum to change that text the discussion should certainly go elsewhere.

Or am I totally missing the point of this pull request?

IrgendwerA8:
As the PR removes the enumeration from the header and "links" (only) the LICENSE, point 3 is conflicted as it states `3. This notice may not be removed or altered from any source`.
If you see the "source" as it was most likely meant as "source-file", it permits the change I was trying to introduce.
Further I'm thinking even if point 2. is violated as changes to file(s) as they happened are not marked in the files (what is original). I tend to close this PR, Uz doesn't seem to be interested in answering my additional questions - which I find understandable after his first answer with my request to (only) "link" the notice:
```
UvB:
Also ich denke, der Text ist doch recht klar: Es darf keine Änderungen am Kopf
geben. Ich kann da auch nicht viel machen, weil ja nicht nur Code von mir
drinsteckt. 
```
(my translation: `I'm thinking the text is quite clear: Changes to the header are not permitted. I cannot do a thing about it - it's not only affecting my code.`)

oliverschmidt:
> [...] point 3 is conflicted as it states [...]

Thanks for the detailed explanation and soryy for being slow on takeup.

> I tend to close this PR [...]

I seem to understand that this PR was originally motivated by the replacing content with a link to the CREDITS file, not by replacing content with a link to the LICENSE file. Isn't the former not still valid?

IrgendwerA8:
>Isn't the former not still valid?

That was one of the additional questions I asked Uz. If the "do not change the header"-point includes the creator/origin/copyright-message of header or the part beginning with `/* This software is provided 'as-is'...` from his point of view.
As I haven't got an answer (yet), I can only guess. Currently I have the impression that he may like to have that preserved as well.
So what's left?
I think we are at least able to introduce the new header for the assembler files, as there is no disclaimer yet (at least I'm not aware of one) - and the CREDITS file at project home wouldn't be of any harm neither.
So new C-sources could also stick to the scheme...
(But I'm still asking myself how to handle point 2. with requesting 'a mark' for non-original files - does  that mean exactly to put a revision history with the editor into the header?)

IrgendwerA8:
I've got an answer from Uz regarding my additional questions! I have to sort it out and will update this subject soon.

oliverschmidt:
> If the "do not change the header"-point includes the creator/origin/copyright-message of the header or only the part beginning with /* This software is provided 'as-is'... from his point of view.

That seems indeed to be THE very question.

oliverschmidt:
> I've got an answer from Uz regarding my additional questions! I have to sort it out and will update this subject soon.

Thanks for the update !

IrgendwerA8:
>That seems indeed to be THE very question.

To give a short summary of his quite comprehensive answer:
While he sees not really a strong claim to preserve the creator information in the header, he not only prefers that this information is preserved (and I would respect that), but also states that point 1 of the disclaimer relates to the "origin" (the creator). From his point of view a global CREDITS file wouldn't be sufficient - if we like to change that, we have to ensure at least a "link" to the history of authors, including the creator/origin. I'm unsure if this can be accomplished with Github, esp. while the creator/source has not committed the initial version.

Extrapolating on his thoughts, I would even tend to think it would be "more clean" if the history of authors **is** mentioned in the header (as "origin" and "mark" information) - but your mileage may vary...

(I'm even unsure if we should touch the assembler files to "indirect" to a global contributors file.)

oliverschmidt:
First of all thanks for the effort you put sof far into this topic.

Looking at https://github.com/cc65/cc65/pull/791 I think our original "issue" was the question if someone working on an exsisting source file should be named in the header and what amount of work entitles to be named there.

So if we return back to that point we can maybe say that we leave the original author stuff alone. But instead of adding additional authors we add some type of generic link to the CREDITS file everywhere.

I'm aware that this contradicts your "Extrapolating on his thoughts [...]" section.

After a very brief scan over the files in question it seems that there are rather few files naming more than one author in the header so far. I imagine that in most/all cases we could get in touch with the additional authors and ask them if it's okay with them to be moved to the CREDITS file. So I'm confident that we could implement the "original author + CREDITS file" approach in an (almost) consistent way.

While not being the cleanest thing imaginable it would for sure solve the issue that started it all: How to handle additional authors working on existing files.

I'm personally willing to go down that road even if there should be some lawyer somewhere claiming that it isn't 100% correct.

Any thoughts?

oliverschmidt:
No opinions after one week? Do I have to consider this pull request "dead"?

IrgendwerA8:
I guess so. The now minor change and tracking down original authors isn't worth the hassle. For new files I will stick to the header I already used in the mulxx case (just "linking" to the LICENSE).
BTW: Uz stated that adding an additional author in the header (like I did in #791) matches quite well to the "rules" in it.

oliverschmidt:
Thanks for the explanation.

PR #862, opened 2019-03-21
Optimized mul20 & mul40 and extracted to new library.

IrgendwerA8:


mrdudz:
wouldnt it be a better idea to just name those functions the same as the old ones, so they can be linked and "overload" the existing ones when needed, and not make cputc require them?

oliverschmidt:
@mrdudz: Even after reading your comment several times I don't understand it. The new functions seem to consist of exactly the same code as the old ones. So why should that code be duplicated? My understanding is that cputc() is just as big/fast/... as before but the functions are additionally made available to the end user.

groessler:
Two points:

- can't these functions live in `libsrc/common` instead of a new directory
- maybe use `__mulXX` (in ASM, C would be `_mulXX`), in order not to pollute the name space. Just thinking...

mrdudz:
oliver: then i totally dont understand why they must be moved into a seperate library - ie what groessler said :)

oliverschmidt:
@groessler: I agree with you.

@mrdudz: Your original comment was about the renaming. That was/is necessary to allow to reuse them from C.

IrgendwerA8:
`libsrc/common ` is not an example for facility of inspection (it's already not fun to relate headers and code there). Could you live with a sub-folder "specialmath" below common?
(Maybe it's even a good idea to add further folders, named like the #include filename and put the matching code in there...?)

groessler:
>`libsrc/common` is not an example for facility of inspection

Do I understand you correctly that you have problems matching source files and headers there? `grep` is your friend.

> Maybe it's even a good idea to add further folders,

I guess that's a call for Oliver. I personally don't see any need for that. And, personally, dislike it.

oliverschmidt:
> I guess that's a call for Oliver.

Some called me? ;-)

Given that there are already those eight cc65_??? match files in that directory I don't see an issue with adding two more. I don't disagree that at some point in time this setup might ask for some elaboration. But I don't see it now - and I don't see it fro two files. Therefore I ask to put the two files there.

IrgendwerA8:
>I don't disagree that at some point in time this setup might ask for some elaboration.

So in the nearer future a directory "ctype", "string" and "stdio" under common would be ok?

groessler:
> So in the nearer future a directory "ctype", "string" and "stdio" under common would be ok?

Come on, that's kind of overengineering...

IrgendwerA8:
>Come on, that's kind of overengineering...

How many files did you changed in that directory?
I think I know "grep" quite well, but I guess you also don't put logic groups of files into your home folder... ;)

My request is serious, I would even add a folder "stdlib". I mean there is even a directory "libsrc/conio" containing fewer files than a "string"-folder would have...

groessler:
> How many files did you changed in that directory?

?

> I guess you also don't put logic groups of files into your home folder.

Sure. "tmp", "work" and "atari" (which contains the cc65 sandbox) :-) . If file name is unique, why do you need different subdirs?

oliverschmidt:
First of all I'd like to state that I personally dislike the communication style in the recent comments of this pull request.

As I said I don't see the two files in question as a compulsory trigger to rearrange things. However, this project lives from its contributors and if this topic is important for a significant contributor I'm of course willing to give it some thoughts.

- I don't think that there are already so many directories in `/libsrc` that we need a sub-directory structure. If we wanted to have a structure I'd see three directories under `/libsrc`: `target`, `common`, `drivers`.

- The code coming with a C toolchain is usally broken up into two parts: 'runtime' and 'C library'. The call sites to code in the former are generated by the C compiler implicitly while the call sites to code in the latter are generated from explicit C function calls.

- We already have a `runtime` directory. So that fine.

- With 'C library' people usually associate the 'C standard library'. Things outside of the C standard library tend to go into separate libraries - and their source code for sure goes into separate directories.

- However with cc65 we went so far for a single monolithic library. I know that there are reasons to question that but I don't want to change that right here right now.

- Anyhow, I always felt that it would be nice to have a directory that contains the code necessary for the C standard library.

So my proposal ist to not have any big bang change but rather "just" split the C standard library stuff in `common` from the non C standard library stuff. This is deliberately a down-to-earth approach. It's certainly not orthogonal as it only targets the files shared between machines and not the machine-specific files.

Given that cc65 is only about C and therefore no ambiguity issue I suggest to have a new directory named `/libsrc/standard` and move all C standard library stuff there from `common`.

This should ease the pressure @IrgendwerA8 feels regarding the `common` directory significantly and is for sure a step we won't regret whatever a potential momentum in the future towards a library modularization might bring.

Just my two cents,
Oliver

groessler:
So, how to continue here?

We had this discussion before (splitting the single lib of each target into multiple libs). I don't remember the details, but we ditched the idea, or postponed it at least. (Could be that now is the time to reconsider?)

Right now I'm fearing that this won't work in general. A single "C-library" lib would be fine. But I think that it might not work (at least) because of the different character translations in each target. Maybe it will work. maybe not, but I don't have enough motivation to check it out.

For this PR, @IrgendwerA8, I'm wondering why you came up with a new include file (`specialmath.h`). `cc65.h` already contains some math functions, why didn't you add the functions there?

greg-king5:
> For this PR, @IrgendwerA8, I'm wondering why you came up with a new include file (`specialmath.h`). `cc65.h` already contains some math functions, why didn't you add the functions there?

I think that we should go one step further:  their function names should be prefixed with "cc65_".

oliverschmidt:
@groessler: Sorry that my comment was obviously too long. Trying to be terse: I did *NOT* propose to add a new library. I only proposed to add a new source file directory!

@greg-king5: I fully agree. The function in question should be treated in any respect just like "the other" math functions. I just wanted to first have the "too many files in common" discussion sorted out before moving over to the "name game" discussion.

greg-king5:
> Given that cc65 is only about C, and therefore no ambiguity issue, I suggest to have a new directory named `/libsrc/standard`, and move all C standard library stuff there from `common`.

We have C standard functions and POSIX standard functions.  @oliverschmidt, do you propose to put both C and POSIX functions in `libsrc/standard/`, or only C standard functions?

oliverschmidt:
I personally would put both C and POSIX functions there. I'm however open to different proposals.

groessler:
> I think that we should go one step further: their function names should be prefixed with "cc65_".

Agree.

But then, shouldn't `doesclrscrafterexit` also have the `cc65_` prepended?

oliverschmidt:
Why exactly `doesclrscrafterexit`? We seem to have quite some cc65 specific functions without that prefix. I wouldnÄt like to raname them all. But we started using that prefix for math stuff, therefore the idea to add it to the new two functions.

groessler:
I was just looking at the `cc65.h` include file, and `doesclrscrafterexit` is the only one without `cc65_` prefix.

Why should `cc65_` just be used for math stuff? It's not clear to me, and maybe I forgot a former discussion about it....

oliverschmidt:
I see, If someone adds the prefix all over the cc65 repo, I'll take care of my usages outside this repo. I presume it's not that popular - and usually backward compatibility isn't such a strong argument for me anyway ;-)

groessler:
My message above was while looking at `cc65.h` in "pattern-matching" mode, not in "thinking" mode.

I do not propose to add `cc65_` to every cc65-special function. In fact, I rather question the fact that those math functions have a `cc65_` prefix.

oliverschmidt:
> I rather question the fact that those math functions have a cc65_ prefix.

I see. That decision was made by Uz. I don't know his reasoning. I fact I was a bit surprised.

IrgendwerA8:
>For this PR, @IrgendwerA8, I'm wondering why you came up with a new include file (specialmath.h). cc65.h already contains some math functions, why didn't you add the functions there?

Taking the time into account you need to give this statement I think I have the same reason like you: I wasn't aware of this header.
From my side I regard this PR as closed. I renamed the functions already. Currently the status is like the discussion open: With "cc65_ " or without - nobody knows.

TBH I find this boring. The value added by selecting a specific location or name is IMHO much lesser than the intrinsic improvement of smaller code and less cycles.
So please name the function whatever you like and put it wherever you want.

oliverschmidt:
@IrgendwerA8: I think ...

> How many files did you changed in that directory? [...] My request is serious, I would even add a folder "stdlib". [...]

... sort of contradicts ...

> TBH I find this boring.

..., doesn't it? ;-) I was adding thoughts to this because it seems to be the opposite of boring to YOU.

> From my side I regard this PR as closed. [...] 

I must admit that I personally don't consider this reaction exactly appropriate...

IrgendwerA8:
>... sort of contradicts ...

It depends.
I would like to have the sub-structure - and thanks BTW that you felt my "urge" and also disliked the style of communication - me too.
But I'm only "fighting for a change" in a very limited way as it takes away to much precious time and efforts - for things that are not that much of importance - esp. in discussions where you can perfectly play the "blame game":

a) You put it there - no, here is a much better place.
b) Why you created it here, there is a much more appropriate place.

a) This is so special, you need to handle in a separate file.
b) Comm 'on you're wasting an additional file, this needs to be included.

I don't like to play this game, it adds no real value if the request is not justified very well and is too often used as "I know better than you".

>I must admit that I personally don't consider this reaction exactly appropriate...

What should I do? Rename the functions to "cc65_mulxx" move them to cc65.h  or renaming the other functions in cc65 to "_...."  or "..." or offer all possible permutations  to make everyone happy?

I revised the names already once like wished, if still not suitable please change the name like you want - or how many times I have to go through this process?
Or for the future: If I form a PR how long should I wait before revising the code to reflect all change requests?

If you don't like to integrate my PR I'm fine with that too. Really! No bad feelings.

oliverschmidt:
I saw two phases:

1. Decide if we want to change the given structure and/or names in any way.

2. Decide where the new content fits into the (potentially changed) structure/naming.

From my perspective we were/are in phase 1. So changing the new content was indeed (too) eagerly.

oliverschmidt:
There was so far objection against my proposal to have a `standard` directory along side `common`, `conio` (and all the others).

@IrgendwerA8: If you think that the `standard` sibling directory is an improvement worth realizing I'd ask you to realize it. Otherwise I think we should keep the structure as-is for the foreseeable future.

IrgendwerA8:
>If you think that the standard sibling directory is an improvement worth realizing I'd ask you to realize it. Otherwise I think we should keep the structure as-is for the foreseeable future.

Then let's keep the structure like it is.
I'm still convinced that the more valuable change would be the introduction of header filename sub-directories. If this isn't about to be welcomed I don't care about a standard or common folder.
But thank you nevertheless for your thoughts and efforts.

oliverschmidt:
I see quite some header declaring cc65 specific stuff without any prefix. However, most of them are target-specific.

I personally see two viable options:

1. Say that C code to be used with cc65 / written for cc65 just isn't the usual C code and that it doesn't really hurt anybody if we pollute the global name space.

2. Say that we in general like the approach to add a `_` prefix to cc65 specific stuff. But likely we consider the incompatibility to existing code too significant to add a `_` prefix all over the place - apart from the question, who's willing to do the change. The result of this perspective would be to keep about everything as-is but add a `_` prefix to new cc65 specific stuff.

Independently from the two options above I consider the `cc65_` prefix for the math stuff in https://github.com/cc65/cc65/blob/master/include/cc65.h as strange and want to get rid of it. Depending on the two options above the `cc65_` prefix should be omitted altogether or replaced by a `_` prefix.

After this has been decided on and realized we're ready to add the new functions this PR was originally all about. They should be added to https://github.com/cc65/cc65/blob/master/include/cc65.h and named in line with the other math functions.

I'm open for opinions on 1.) vs. 2.) ...

mrdudz:
i dont like _ prefix... to me it implies a lowlevel function that a user program shouldnt be using normally. i dont really like cc65_ prefix either, because indeed we have a lot of "cc65 specific" stuff that would have to get it then (large parts of the libraries actually). we'd even end up with some parts of the libraries mixing these (eg conio, we have a bunch of "standard" things in it, but also cc65 specific functions) - and that would be really horrible IMHO. my vote would be on 1) :)

silverdr:
None of the two options are perfect but if I had to choose then everything that makes cc65 wander away from the actual "C" lang norms is a "no-go please". That means I'd have less of an issue with [2.] than I'd have with [1.]. The problem with [2.] is how I see it that we can't say when/if the rest will be adapted. OTOH, if we begin with the new stuff conforming to the rule, then this will be some form of a start of the potentially long process.

IrgendwerA8:
Just in case you not already guessed that, I more tend to option 1 (see my original PR). (And the double underscores in the assembly files are really ugly.)

>Independently from the two options above I consider the cc65_ prefix for the math stuff in https://github.com/cc65/cc65/blob/master/include/cc65.h as strange and want to get rid of it.

I think so too, as they mean not really something. But what about the already existing functions starting with e.g. "`cbm_`", "`tgi_`", "`joy_`" or "`em_`". What about a new "`math_`" ?

What makes me still a little bit wonder is that this subject is now that important. Was this ever an issue when @oliverschmidt created in 2012 the "`device.h`" with it's global namespace "polluting" function names?

oliverschmidt:
Adding my own preference I see a clear momentum towards option 1.).

Therefore I ask @IrgendwerA8 to name the functions `mul20()` resp. `mul40()` and add their declarations to `cc65.h`. The source files should go into `common` and should be named `mul20.s` and `mul40.s`. Thanks in advance for your patience and cooperation.

oliverschmidt:
@IrgendwerA8: May I ask you to take care of the existing `cc65_???` stuff? That would mean to

- Rename the 8 files in question to omit the `cc65_` prefix.
- Rename the the 9 functions in question to omit the `cc65_` prefix.

oliverschmidt:
> What makes me still a little bit wonder is that this subject is now that important. Was this ever an issue when @oliverschmidt created in 2012 the "device.h" with it's global namespace "polluting" function names?

I see actually two aspects here:

- Why was Uz fine with my non-prefixed names back here but saw the need for the cc65_ prefix elsewhere? I guess he knows just as little as we do ;-))

- Why is the subject now important? It's not unusual for the cc65 project to have discussions to pop up at unexpected times for unexpected reasons. I see my role in not trying to suppress that happening but just trying to moderate it. I'm not trying to generate momentums (anymore ;-) but just trying to give upcoming momentums a direction that doesn't hurt the project.

groessler:
> Adding my own preference I see a clear momentum towards option 1.)

I haven't had a chance to vote :-)

I'm for option 2) (as said there, leave existing functions as they are, but add `_` to new ones).

In longer words: The existing functions in `cc65.h` (even with the `cc65_` prefix removed) appear to me to have litte chance to clash with users' functions. In the case of `mul20`/`mul40` I'm not so sure.

IrgendwerA8:
...and now? (I didn't changed the code yet.)

oliverschmidt:
I asked you to do "it" in a certain way. If you follow my wishes I will merge your changes.

groessler:
> I asked you to do "it" in a certain way. If you follow my wishes I will merge your changes.

I don't have a strong enough opinion about the function naming, so, @IrgendwerA8, go forward...

oliverschmidt:
@groessler: Thanks for the clarification.

include/cc65.h
```diff
@@ -44,43 +44,53 @@
...
 /* Multiplicate an unsigned 16 bit by an unsigned 8 bit number yielding a 24
 ** bit unsigned result that is extended to 32 bits for easier handling from C.
 */
 
-unsigned long __fastcall__ cc65_umul16x16r32 (unsigned lhs, unsigned rhs);
+unsigned long __fastcall__ umul16x16r32 (unsigned lhs, unsigned rhs);
 /* Multiplicate two unsigned 16 bit to yield an unsigned 32 bit result */
 
-int __fastcall__ cc65_sin (unsigned x);
+unsigned int __fastcall__ mul20(unsigned char value);
```
oliverschmidt: Please add the space before the '('.

include/cc65.h
```diff
@@ -44,43 +44,53 @@
...
+/* Multiply an 8 bit unsigned value by 20 and return the 16 bit unsigned
+** result
+*/
+
+unsigned int __fastcall__ mul40(unsigned char value);
```
oliverschmidt: Please add the space before the '('.

doc/funcref.sgml
```diff
@@ -5558,6 +5559,29 @@ memory allocated for the driver.
...
+</descrip>
+</quote>
+
+<sect1>mul40<label id="mul40"><p>
+
+<quote>
+<descrip>
+<tag/Function/Multiplies argument by 40.
+<tag/Header/<tt/<ref id="cc65.h" name="cc65.h">/
+<tag/Declaration/<tt/unsigned int __fastcall__ mul40(unsigned char value);/
```
oliverschmidt: Please add the space before the '('.

doc/funcref.sgml
```diff
@@ -5558,6 +5559,29 @@ memory allocated for the driver.
 </descrip>
 </quote>
 
+<sect1>mul20<label id="mul20"><p>
+
+<quote>
+<descrip>
+<tag/Function/Multiplies argument by 20.
+<tag/Header/<tt/<ref id="cc65.h" name="cc65.h">/
+<tag/Declaration/<tt/unsigned int __fastcall__ mul20(unsigned char value);/
```
oliverschmidt: Please add the space before the '('.

oliverschmidt:
Easy one, wasn't it ;-))

IrgendwerA8:
It happens not often that I'm speechless - I am.

clbr:
Please restore the cc65_ prefix to the sin and cos functions. Those will clash with user code, which may work in other coordinate spaces.

mrdudz:
i'd prefer names that show they work in fixedpoint, and in what range

but using names of standard functions should really be a nono, indeed.

oliverschmidt:
@mrdudz: Can you provide a specific proposal for names (maybe matching the naming style of `udiv32by16r16` and friends) ?

mrdudz:
perhaps something like ssin360r8p8 ? or ssin360r16 ? its a bit tricky because the function is so limited =P

now that i look at the functions, i think they are almost unuseable in practise, because IF you are using fixedpoint already, you usually want to use 0...255 as input (or at least a reasonable power of 2, if only to handle values of any range easily). and likely a 8bit result is ok too :)

groessler:
My stance on this issue is still to have an underscore at the front in the name of these functions. For the math functions at least. For `doesclrscrafterexit` I'm not sure. Here with or without underscore seems fine to me.

oliverschmidt:
> perhaps something like ssin360r8p8 ? or ssin360r16 ? its a bit tricky [...]

@mrdudz: I see. Nevertheless, thanks for your effort.

Do we need to vote here (again)? Or is an underscore prefix "good enough" for everybody?

oliverschmidt:
Here you go: https://github.com/cc65/cc65/commit/ac0b452834265d7f4287e9d6e10f2631fbd66dd0

clbr:
Thanks!

PR #865, opened 2019-03-23
Adds address after JVB instruction in the display list example.

dmsc:
The correct address is needed as argument to the JVB instruction.


IrgendwerA8:
Yes, that's better. Thanks!

PR #867, opened 2019-03-30
Fix gcc compile warning

ops:
```c
ca65/scanner.c
ca65/scanner.c: In function ‘ReadStringConst’:
ca65/scanner.c:819:24: warning: this statement may fall through [-Wimplicit-fallthrough=]
                     if (IsXDigit (C)) {
                        ^
ca65/scanner.c:828:17: note: here
                 default:
                 ^~~~~~~
```
Fixed warning by adjusting the comment. It is now recognized by gcc.
For more information see:

https://developers.redhat.com/blog/2017/03/10/wimplicit-fallthrough-in-gcc-7/

### DIFF
--- a/src/ca65/scanner.c
+++ b/src/ca65/scanner.c
@@ -824,7 +824,7 @@ static void ReadStringConst (int StringTerm)
                             break;
                         }
                     }
-                    /* FALLTHRU */
+                    /* FALLTHROUGH */
                 default:
                     Error ("Unsupported escape sequence in string constant");
                     break;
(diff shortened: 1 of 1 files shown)


src/ca65/scanner.c
```diff
@@ -824,7 +824,7 @@ static void ReadStringConst (int StringTerm)
                             break;
                         }
                     }
-                    /* otherwise, fall through */
+                    /* FALLTHRU */
```
greg-king5: This project prefers the spelling:
```c
/* FALLTHROUGH */
```

ops: Fixed

greg-king5:
That warning is an insignificant issue -- except that we must read it every time in order to see that it's insignificant.  So, thanks for putting it out of our way.

Issue #866, opened 2019-03-27
hello.c for Atari 5200


polluks:
We have conio support but hello does not compile.
By the way how about adding a .CAR header?

groessler:
which "hello" does not compile? Please post source code.

polluks:
samples/hello.c our reference.

groessler:
Yes, you're right. I'm getting this (why didn't you post that as well?):

```
$ cl65 -tatari5200 -o /tmp/x.bin samples/hello.c 
samples/hello.c(51): Error: Undefined symbol: 'CH_ULCORNER'
samples/hello.c(53): Error: Undefined symbol: 'CH_URCORNER'
samples/hello.c(59): Error: Undefined symbol: 'CH_LLCORNER'
samples/hello.c(61): Error: Undefined symbol: 'CH_LRCORNER'
$
```

After adding these defines, I'm getting:

```
$ cl65 -tatari5200 -o /tmp/x.bin samples/hello.c 
Unresolved external '_bgcolor' referenced in:
  samples/hello.s(51)
Unresolved external '_bordercolor' referenced in:
  samples/hello.s(49)
Unresolved external '_cgetc' referenced in:
  samples/hello.s(128)
Unresolved external '_textcolor' referenced in:
  samples/hello.s(47)
Unresolved external 'screensize' referenced in:
  conio/scrsize.s(20)
ld65: Error: 5 unresolved external(s) found - cannot create output file
$
```

The `cgetc` error should be fixed in `hello.c` I guess. Extend the `#if defined(__NES__) || defined(__PCE__) || defined(__GAMATE__)` line.

For the rest, you're invited to provide implementations. :-)

polluks:
C'mon, it's your baby! :-P

groessler:
Which does't mean that I'm the only one to make changes :-)

groessler:
How do you think that the fact that there is no difference between "border color" and "background color" be handled? Atari5200 uses Antic mode #6 (BASIC GRAPHICS mode #1) for its conio.

oliverschmidt:
I presume this is an "open" question to everybody, right?

I'm not sure I understand you right. I presume you want to express that there's no way to have the background and the border to have different colors...

In that case I'm in fact pretty sure that the "right" way to handle this is that bordercolor() does nothing while bgcolor() sets the color in question. Reasons:

[...]

PR #870, opened 2019-04-02
atari5200 conio fixes

groessler:
Documentation missing. I want to propose that in response to issue #866.

### DIFF
--- a/asminc/atari5200.inc
+++ b/asminc/atari5200.inc
@@ -10,6 +10,16 @@
  
 ATEOL   = $9B      ; END-OF-LINE, used by CONIO
 
+;-------------------------------------------------------------------------
+; CONIO CHARACTER DEFS
+;-------------------------------------------------------------------------
+
+CH_ULCORNER = $0B  ; '+' sign
+CH_URCORNER = $0B
+CH_LLCORNER = $0B
+CH_LRCORNER = $0B
+CH_HLINE    = $0D  ; dash
+CH_VLINE    = $01  ; exclamation mark
 
 ;-------------------------------------------------------------------------
 ; Zero Page
--- a/asminc/atari5200.inc
+++ b/asminc/atari5200.inc
@@ -115,10 +115,10 @@ POKEY     = $E800  ; POKEY area
 ; conio color defines
 ;-------------------------------------------------------------------------
 
-COLOR_WHITE      = 0
-COLOR_LIGHTRED   = 1
-COLOR_LIGHTGREEN = 2
-COLOR_BLACK      = 3
+COLOR_WHITE = 0
+COLOR_RED   = 1
+COLOR_GREEN = 2
+COLOR_BLACK = 3
 
 ;-------------------------------------------------------------------------
 ; Cartridge Parameters
--- a/asminc/atari5200.inc
+++ b/asminc/atari5200.inc
@@ -111,6 +111,14 @@ ANTIC     = $D400  ; ANTIC area
 POKEY     = $E800  ; POKEY area
 .include  "atari_pokey.inc"
 
+;-------------------------------------------------------------------------
+; conio color defines
+;-------------------------------------------------------------------------
+
+COLOR_WHITE = 0
+COLOR_RED   = 1
+COLOR_GREEN = 2
+COLOR_BLACK = 3
 
 ;-------------------------------------------------------------------------
 ; Cartridge Parameters
--- a/asminc/atari5200.inc
+++ b/asminc/atari5200.inc
@@ -115,10 +115,10 @@ POKEY     = $E800  ; POKEY area
 ; conio color defines
 ;-------------------------------------------------------------------------
 
-COLOR_WHITE = 0
-COLOR_RED   = 1
-COLOR_GREEN = 2
-COLOR_BLACK = 3
+COLOR_WHITE      = 0
+COLOR_LIGHTRED   = 1
+COLOR_LIGHTGREEN = 2
+COLOR_BLACK      = 3
 
 ;-------------------------------------------------------------------------
 ; Cartridge Parameters
--- a/libsrc/atari5200/cputc.s
+++ b/libsrc/atari5200/cputc.s
@@ -15,9 +15,9 @@
         .importzp       screen_width, screen_height
         .importzp       ptr4
 
-        .import         screen_setup
-        .constructor    screen_setup_constructor, 26
-screen_setup_constructor = screen_setup
+        .import         initconio
+        .constructor    initconio_constructor
+initconio_constructor = initconio
 
 _cputcxy:
         pha                     ; Save C
--- a/libsrc/atari5200/cputc.s
+++ b/libsrc/atari5200/cputc.s
@@ -82,8 +82,8 @@ putchar:
         sta     ptr4+1
         pla                     ; get char again
 
-;       and     #$C0            ; without this we are compatible with the old version. user must not try to output a char >= $3F
-        ora     conio_color
+        and     #$3F            ; clear palette index bits
+        ora     conio_color     ; use currently selected palette
 
         ldy     COLCRS_5200
         sta     (ptr4),y
--- a/libsrc/atari5200/cputc.s
+++ b/libsrc/atari5200/cputc.s
@@ -15,9 +15,9 @@
         .importzp       screen_width, screen_height
         .importzp       ptr4
 
-        .import         initconio
-        .constructor    initconio_constructor
-initconio_constructor = initconio
+        .import         screen_setup
+        .constructor    initconio
+initconio               =       screen_setup
 
 _cputcxy:
         pha                     ; Save C
--- a/libsrc/atari5200/cputc.s
+++ b/libsrc/atari5200/cputc.s
@@ -15,10 +15,9 @@
         .importzp       screen_width, screen_height
         .importzp       ptr4
 
-        .constructor    screen_setup, 26
-        .import         screen_setup_20x24
-screen_setup    = screen_setup_20x24
-
+        .import         screen_setup
+        .constructor    screen_setup_constructor, 26
+screen_setup_constructor = screen_setup
 
 _cputcxy:
         pha                     ; Save C
--- a/libsrc/atari5200/cputc.s
+++ b/libsrc/atari5200/cputc.s
@@ -11,8 +11,9 @@
         .export         _cputcxy, _cputc
         .export         plot, cputdirect, putchar
         .import         gotoxy, _mul20
+        .import         conio_color
+        .importzp       screen_width, screen_height
         .importzp       ptr4
-        .import         setcursor
 
         .constructor    screen_setup, 26
         .import         screen_setup_20x24
@@ -44,7 +45,7 @@ L4:     cmp     #$0A            ; LF
         and     #3
         tax
         tya
-        and     #$9f
+        and     #$9F
         ora     ataint,x
 
 cputdirect:                     ; accepts screen code
@@ -53,7 +54,7 @@ cputdirect:                     ; accepts screen code
 ; advance cursor
         inc     COLCRS_5200
         lda     COLCRS_5200
-        cmp     #20
+        cmp     #screen_width
         bcc     plot
         lda     #0
         sta     COLCRS_5200
@@ -62,12 +63,11 @@ cputdirect:                     ; accepts screen code
 newline:
         inc     ROWCRS_5200
         lda     ROWCRS_5200
-        cmp     #24
+        cmp     #screen_height
         bne     plot
         lda     #0
         sta     ROWCRS_5200
-plot:   jsr     setcursor
-        ldy     COLCRS_5200
+plot:   ldy     COLCRS_5200
         ldx     ROWCRS_5200
         rts
 
@@ -83,10 +83,12 @@ putchar:
         sta     ptr4+1
         pla                     ; get char again
 
+;       and     #$C0            ; without this we are compatible with the old version. user must not try to output a char >= $3F
+        ora     conio_color
+
         ldy     COLCRS_5200
         sta     (ptr4),y
-        jmp     setcursor
+        rts
 
         .rodata
 ataint: .byte   64,0,32,96
-
--- a/asminc/atari_gtia.inc
+++ b/asminc/atari_gtia.inc
@@ -79,3 +79,41 @@ VDELAY  = GTIA + $1C         ;vertical delay
 GRACTL  = GTIA + $1D         ;graphic control
 HITCLR  = GTIA + $1E         ;collision clear
 
+
+; Hue values
+
+HUE_GREY        = 0
+HUE_GOLD        = 1
+HUE_GOLDORANGE  = 2
+HUE_REDORANGE   = 3
+HUE_ORANGE      = 4
+HUE_MAGENTA     = 5
+HUE_PURPLE      = 6
+HUE_BLUE        = 7
+HUE_BLUE2       = 8
+HUE_CYAN        = 9
+HUE_BLUEGREEN   = 10
+HUE_BLUEGREEN2  = 11
+HUE_GREEN       = 12
+HUE_YELLOWGREEN = 13
+HUE_YELLOW      = 14
+HUE_YELLOWRED   = 15
+
+; Color defines, similar to c64 colors (untested)
+
+GTIA_COLOR_BLACK      = (HUE_GREY << 4)
+GTIA_COLOR_WHITE      = (HUE_GREY << 4 | 7 << 1)
+GTIA_COLOR_RED        = (HUE_REDORANGE << 4 | 1 << 1)
+GTIA_COLOR_CYAN       = (HUE_CYAN << 4 | 3 << 1)
+GTIA_COLOR_VIOLET     = (HUE_PURPLE << 4 | 4 << 1)
+GTIA_COLOR_GREEN      = (HUE_GREEN << 4 | 2 << 1)
+GTIA_COLOR_BLUE       = (HUE_BLUE << 4 | 2 << 1)
+GTIA_COLOR_YELLOW     = (HUE_YELLOW << 4 | 7 << 1)
+GTIA_COLOR_ORANGE     = (HUE_ORANGE << 4 | 5 << 1)
+GTIA_COLOR_BROWN      = (HUE_YELLOW << 4 | 2 << 1)
+GTIA_COLOR_LIGHTRED   = (HUE_REDORANGE << 4 | 6 << 1)
+GTIA_COLOR_GRAY1      = (HUE_GREY << 4 | 2 << 1)
+GTIA_COLOR_GRAY2      = (HUE_GREY << 4 | 3 << 1)
+GTIA_COLOR_LIGHTGREEN = (HUE_GREEN << 4 | 6 << 1)
+GTIA_COLOR_LIGHTBLUE  = (HUE_BLUE << 4 | 6 << 1)
+GTIA_COLOR_GRAY3      = (HUE_GREY << 4 | 5 << 1)
--- a/include/_gtia.h
+++ b/include/_gtia.h
@@ -131,44 +131,23 @@ struct __gtia_write {
 #define HUE_YELLOWRED   15
 
 /* Color defines, similar to c64 colors (untested) */
-/* Note that the conio color implementation is monochrome
-** (bgcolor and textcolor are only placeholders)
-*/
-/* Use the defines with the setcolor() or _atari_xxxcolor() functions */
-#define COLOR_BLACK             _gtia_mkcolor(HUE_GREY,0)
-#define COLOR_WHITE             _gtia_mkcolor(HUE_GREY,7)
-#define COLOR_RED               _gtia_mkcolor(HUE_REDORANGE,1)
-#define COLOR_CYAN              _gtia_mkcolor(HUE_CYAN,3)
-#define COLOR_VIOLET            _gtia_mkcolor(HUE_PURPLE,4)
-#define COLOR_GREEN             _gtia_mkcolor(HUE_GREEN,2)
-#define COLOR_BLUE              _gtia_mkcolor(HUE_BLUE,2)
-#define COLOR_YELLOW            _gtia_mkcolor(HUE_YELLOW,7)
-#define COLOR_ORANGE            _gtia_mkcolor(HUE_ORANGE,5)
-#define COLOR_BROWN             _gtia_mkcolor(HUE_YELLOW,2)
-#define COLOR_LIGHTRED          _gtia_mkcolor(HUE_REDORANGE,6)
-#define COLOR_GRAY1             _gtia_mkcolor(HUE_GREY,2)
-#define COLOR_GRAY2             _gtia_mkcolor(HUE_GREY,3)
-#define COLOR_LIGHTGREEN        _gtia_mkcolor(HUE_GREEN,6)
-#define COLOR_LIGHTBLUE         _gtia_mkcolor(HUE_BLUE,6)
-#define COLOR_GRAY3             _gtia_mkcolor(HUE_GREY,5)
-
-/* TGI color defines */
-#define TGI_COLOR_BLACK         COLOR_BLACK
-#define TGI_COLOR_WHITE         COLOR_WHITE
-#define TGI_COLOR_RED           COLOR_RED
-#define TGI_COLOR_CYAN          COLOR_CYAN
-#define TGI_COLOR_VIOLET        COLOR_VIOLET
-#define TGI_COLOR_GREEN         COLOR_GREEN
-#define TGI_COLOR_BLUE          COLOR_BLUE
-#define TGI_COLOR_YELLOW        COLOR_YELLOW
-#define TGI_COLOR_ORANGE        COLOR_ORANGE
-#define TGI_COLOR_BROWN         COLOR_BROWN
-#define TGI_COLOR_LIGHTRED      COLOR_LIGHTRED
-#define TGI_COLOR_GRAY1         COLOR_GRAY1
-#define TGI_COLOR_GRAY2         COLOR_GRAY2
-#define TGI_COLOR_LIGHTGREEN    COLOR_LIGHTGREEN
-#define TGI_COLOR_LIGHTBLUE     COLOR_LIGHTBLUE
-#define TGI_COLOR_GRAY3         COLOR_GRAY3
+/* Hardware palette values (for GTIA colxxx registers) */
+#define GTIA_COLOR_BLACK             _gtia_mkcolor(HUE_GREY,0)
+#define GTIA_COLOR_WHITE             _gtia_mkcolor(HUE_GREY,7)
+#define GTIA_COLOR_RED               _gtia_mkcolor(HUE_REDORANGE,1)
+#define GTIA_COLOR_CYAN              _gtia_mkcolor(HUE_CYAN,3)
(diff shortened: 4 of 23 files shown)


groessler:
Hmm, checks failed. Don't seem to come from my commit, looks like a certificate issue...

oliverschmidt:
I just restarted the checks. Have you tried to do that yourself (on travis-ci.org)?

groessler:
I haven't tried to do that myself, I never visited travis-ci.org (didn't know it).

groessler:
Continuing the discussion from #866 here, I guess you've noticed that I didn't change any COLOR_ defines.

In `include/_gtia.h` we have COLOR_ defines which emulate the C64 colors. These values are good to write into the palette registers.

But the `textcolor()` values have nothing to do with these values, they are just an index for which palette register to use. So here the `textcolor()` values are totally independent of the COLOR_ values. This Atari text mode just doesn't fit into the (C64 derived?) conio COLOR_ definitions.

One should use one of 0, 1, 2, 3 as parameters for `textcolor()`. I could preset the palette registers with "nice" colors, as Oliver suggested, and define COLOR_ values for them (in the range 0..3). But this would destroy the existing COLOR_ defines which I think are handy. So I didn't create artificial COLOR_ defines for the "nice" colors and left the palette registers as they were, with the same colors as on Atari computers. People will recognize them if they used Atari computers' BASIC "GRAPHICS 1" mode and wrote text to the screen.

oliverschmidt:
Hm, my understanding is that the COLOR_ macros are meant to be used with textcolor(), bgcolor() and bordercolor() - and nothing else. If they happen to be useful for something else then that may be fine, but I certainly don't see the COLOR_ macros to be "handy" but not useful for the very thing they are meant to be used.

If you think that "some" macros for setting the palette to "well known" values are desirable then please define "some" macros for that purpose - but not the COLOR_ macros. How is somebody supposed to write cross target color CONIO code if he can't write `textcolor(COLOR_RED);` yielding red text?

I personally consider everything else as a bug that should be fixed - and therefore backward compatibility doesn't count here for me.

mrdudz:
i am with oliver there... common macros should be used for the common purpose (writing portable programs). anything else is not expected behaviour, and thus a bug that should be fixed.

groessler:
@oliverschmidt: I read your message multiple times, but don't exactly understand what you're aiming at.

> my understanding is that the COLOR_ macros are meant to be used with textcolor(), bgcolor() and bordercolor()

Exactly that does not work. There can be COLOR_ defines which either work for `bgcolor()` or `textcolor()`, but not for both at the same time.
`bgcolor()` sets the palette register of the background (globally for the whole screen, not like on C64 for each character cell). The parameter is a 7bit value indicating the color. The current COLOR_ defines define values which loosely match the C64 colors.
`textcolor()` sets the 2bit value to be used for the written text. It's an index of which palette register to use.

>  and nothing else. If they happen to be useful for something else then that may be fine

I never talked about "something else".

> I certainly don't see the COLOR_ macros to be "handy" but not useful for the very thing they are meant to be used.

With "handy" I meant compatible color values with the C64 colors.

> How is somebody supposed to write cross target color CONIO code if he can't write textcolor(COLOR_RED); yielding red text?

This just won't work. I can either define COLOR_RED as the palette value, or as the index (after initializing the palette register accordingly). And then it either works with `bgcolor()` or `textcolor()`.
Anyway, we just have 4 foreground colors here. How would cross-platform work if a program is using one of the non-existent colors?

mrdudz:
bgcolor also on the c64 sets the background color of the entire screen. (bgcolor sets $d021, bordercolor sets $d020, and textcolor is the character color that can be different for each character)

if the same values cant be used directly for registers on the atari, then the functions must map them to whatever is needed internally. even if that means that the atari conio can use only 4 colors.

groessler:
> bgcolor also on the c64 sets the background color of the entire screen. (bgcolor sets $d021, bordercolor sets $d020, and textcolor is the character color that can be different for each character)

You're right. I've mixed that up with CGA/EGA/VGA, where each char cell can have a different bg color.

> if the same values cant be used directly for registers on the atari, then the functions must map them to whatever is needed internally

Pah. I could of course search if the color value given to `textcolor()` happens to be in one of the palette register. But I'm considering that kind of overengineered. And what should be done if the value isn't in any palette register? There is no "error" return code for `textcolor()`.

> even if that means that the atari conio can use only 4 colors.

It can only use 4 of the possible 128 colors simultaneously. If a palette register is changed, chars previously written in another color will change their color.

mrdudz:
> Pah. I could of course search if the color value given to textcolor() happens to be in one of the palette register. But I'm considering that kind of overengineered. And what should be done if the value isn't in any palette register? There is no "error" return code for textcolor().

no, the right way is to define 4 colors that work with all of these functions.

groessler:
> no, the right way is to define 4 colors that work with all of these functions.

What in "This just won't work" didn't you understand?

mrdudz:
I don't see how it "just won't work". I am pretty sure there are colors that can be used for both background and characters. It might require a lookup table inside the respective functions to map them to whatever values go to registers, but that doesnt mean "it doesnt work".

oliverschmidt:
Again, I see the general communication style change. So far the cc65 community was pretty much spared of this change. I'd *REALLY* appreciate if we could keep it that way - at least as long as possible!

mrdudz:
I don't see the problem myself. Apparently conio can only be implemented for 4 (fixed!) colors without changing the semantics of the functions (which is a big nono - IMHO). Any further functionality must go into atari specific functions, which can then use whatever they need, and support specific atari features.

groessler:
> Again, I see the general communication style change.

Ok. And it was me again. Sorry, but I'm a bit annoyed. I didn't want to offend @mrdudz .

> I don't see how it "just won't work". I am pretty sure there are colors that can be used for both background and characters. It might require a lookup table inside the respective functions to map them to whatever values go to registers, but that doesnt mean "it doesnt work".

To me that's more or less the same as I wrote:

>  I could of course search if the color value given to textcolor() happens to be in one of the palette register. But I'm considering that kind of overengineered. And what should be done if the value isn't in any palette register? There is no "error" return code for textcolor().

I want to keep the current COLOR_ defines, since they are good for palette values to get similar colors as on the other targets.

> I don't see the problem myself. Apparently conio can only be implemented for 4 (fixed!) colors without changing the semantics of the functions 

You are suggesting to initialize the palette values to some predefined colors (e.g. white, red, green, blue), and determine in the `textcolor()` function which of those was selected to find out which palette register to use? What should be done if a color not out of those four was passed to the function?

This could work. But in requires an additional `textcolor5200()` function or so, to make it usable for 5200 specific programs.

I think that's kind of overengineering for the sake of cross-platform compatibilty which anyway cannot be accomplished in a 100% correct way since 11 colors are missing.

mrdudz:
> You are suggesting to initialize the palette values to some predefined colors (e.g. white, red, green, blue), and determine in the textcolor() function which of those was selected to find out which palette register to use?

yes. thats how other targets do it too. i vaguely remember a discussion about what colors are reasonable to use for one of them too (but i forgot which target it was, NES perhaps?) (i would suggest to include black, since that is often the background color)

> What should be done if a color not out of those four was passed to the function?

that can not happen, because no COLOR_ defines for those non working colors should exist. ie you will get an error at compile time. the differences between targets must be handled in ifdefs.

> This could work. But in requires an additional textcolor5200() function or so, to make it usable for 5200 specific programs.
I think that's kind of overengineering for the sake of cross-platform compatibilty which anyway cannot be accomplished in a 100% correct way since 11 colors are missing.

it's the only sane way to do it - imho. the portable functions remain portable (within the limits of the target) - and whatever else is done in target specific functions.

groessler:
> yes. thats how other targets do it too. i vaguely remember a discussion about what colors are reasonable to use for one of them too 

I'm almost with you...

> i would suggest to include black, since that is often the background color

Yeah, that's another problem. In this text mode we are using 5 palette registers / colors. One for the background, and 4 for the possible text colors.
The background color palette register is set to black. And the other four could be set to the default colors. Written text can only appear in the colors of the four other palette registers.

Means, that if someone sets the bg color to white, he cannot (without messing platform-specific with the palette registers) set a fg color of black. which in turn won't work with the `textcolor()` function as it checks for "well known" color values. But a `textcolor5200()` would work.

> that can not happen, because no COLOR_ defines for those non working colors should exist. ie you will get an error at compile time. the differences between targets must be handled in ifdefs.

Ok. I'd like to keep the existing COLOR_ defines. Could rename them to GTIA_COLOR_ or so.

The API really doesn't fit well to the way the hardware works here :-(

oliverschmidt:
Hi Chris,

> I'd like to keep the existing COLOR_ defines. Could rename them to GTIA_COLOR_ or so.

For sure. That's what I already proposed before.

> The API really doesn't fit well to the way the hardware works here :-(

I'd really like to understand the issue, but so far I don't. Maybe you can elaborate a bit more...

> In this text mode we are using 5 palette registers / colors. One for the background, and 4 for the possible text colors.

So far so good.

> The background color palette register is set to black. And the other four could be set to the default colors.

That would be a reasonable startup scenario.

> Written text can only appear in the colors of the four other palette registers.

Okay.

> Means, that if someone sets the bg color to white, he cannot (without messing platform-specific with the palette registers) set a fg color of black.

Why? I suggested the four "nice" colors:

* Black
* White
* Red
* Green

So one of the four palette registers for the text colors contains black. Where's the problem in using it as-is?

> which in turn won't work with the textcolor() function as it checks for "well known" color values.

Now you completely lost me. I don't know if this is the misunderstanding but it was mentioned before that for sure a single COLOR macros does not need to be able to be both written in the background color palette register _AND_ serve as text color palette register index value. Either one or the other needs to be done through a lookup table.

Likely one will define the four COLOR macros as serving 1:1 as text color palette register index values. I presume this means the are define as 0 - 3. Then there needs to be a 4 item table inside the bgcolor() function that is able to translate COLOR macro values to the value to be written in the background color palette register. Additionally the parameter (aka macro value) itself is saved in order to be returned as old background color value on the next bgcolor() call. This way there's never a "search for colors". If someone modifies any of the five color palette entries with other means than the portable CONIO code then the program still behaves correct and still gets the expected return values. Just the visual appearance is "broken".

Maybe I'm getting all this wrong because of missing information. Then I ask you to provide it. I'm just trying to push this thread forward...

Regards,
Oliver

polluks:
I like the new 20×12 mode 👍

groessler:
> Maybe you can elaborate a bit more...

I don't know. I've written it a few times already.

> So one of the four palette registers for the text colors contains black. Where's the problem in using it as-is?

My problem here is that we then are losing one possible foreground color.
I could set bg color and one of the fg color palette registers to black, but then there are only 3 possible fg colors, while the hw supports 4 fg colors.

groessler:
> I like the new 20×12 mode

Thanks :-)

But, @polluks, you've opened a can of worms with posting this isse :-)

oliverschmidt:
I must that I feel a little pranked now:

* I tell you that I suggest 4 colors.
* You say repeatedly "It doesn't work!".
* There are quite some useless postings.
* I finally explain in detail how I think it can work.
* You say "But it doesn't work with 5 colors."

Yes, it doesn't work with 5 colors. Nobody said it would...

Maybe it helps when I say: "I prefer a portable solution with 4 colors over a proprietary solution with 5 colors.".

groessler:
@oliverschmidt, I didn't want to prank you (can you say so in English?).

> I tell you that I suggest 4 colors.

I was (probably erroneously) under the impression that you don't know that 5 colors are available.
I wanted to make everything the hw can do available, and not cripple it.

> You say repeatedly "It doesn't work!"

Hmm. Yes. I've thought that everything (within sense) the hw can do should be made available to the programmer. With just having 4 colors you're right, this could be implemented, just 4 colors in total.

> There are quite some useless postings.

I'm considering this statement a little bit of offensive. We've discussed to find a basic understanding/agreement. Doing this by email/posts is less efficient as could be done by talking directly (on phone or meeting in person). I guess, talking directly would have taken 15 to 30min to find a common agreement/understanding.

> I finally explain in detail how I think it can work.

I don't find that explanation right now. So just 4 colors, the 5 palette registers initialized with 4 different color values?

> You say "But it doesn't work with 5 colors."

Still true, doesn't work. Leave out the 5th color, then it's good.

> I prefer a portable solution with 4 colors over a proprietary solution with 5 colors.

Ok. Not nice, but can be done.

oliverschmidt:
>> There are quite some useless postings.

> I'm considering this statement a little bit of offensive.

Sorry, I didn't want to express that some participant wasn't willing to reach an understanding. I wanted to express that now looking back they didn't actually help to reach an understanding.

>> I prefer a portable solution with 4 colors over a proprietary solution with 5 colors.

> Ok. Not nice, but can be done.

That's great :-) So we now have a common base line that is either improved or can be "just" implemented as-is.

oliverschmidt:
I already wrote back in https://github.com/cc65/cc65/issues/866#issuecomment-479036893:

_If I got it right that the background color can be changed to e.g. white I'd propose these four colors [...]_

So I was already implying that black needs to be present both as (initial) value in the background color register as well as in one of text color registers. Five registers with four colors...

groessler:
> So we now have a common base line that is either improved or can be "just" implemented as-is.

For the sake of cross-platform compatibility... (Don't get me wrong, I'm pro cross-platform, but in this case I'm not really convinced so that I can stand 100% behind the solution.)

There are still some loose ends which I'm going to address in an ad-hoc way. Feel free to point out bad decisions you find in my upcoming commits.

For example, this means that `bgcolor` will only be able to set one of the predefined colors. While it can set any of on the hw available colors currently.
Also, the return value of the functions forces me to keep track of what color was set and return that. Which might not be the actual color if the user set the palette registers directly. (But one could argue that conio and hw dependent functions shouldn't be mixed. OTOH, it works quite well currently.)

To me this is a nice solution for a cross-platform program where the user just wants to compile it for 5200, but doesn't know much of the 5200 system. For an advanced user of the system, this solution will get in the way of what the user wants to accomplish (just my guts feeling, cannot back that up with anything).

mrdudz:
> Which might not be the actual color if the user set the palette registers directly. (But one could argue that conio and hw dependent functions shouldn't be mixed. OTOH, it works quite well currently.)

its fine imho - that is platform specific behaviour then. for someone writing a 5200 specific program, you can (and should, imho) provide extra functions and/or macros (i dont know the platform, so no idea what is doable) that can be used instead of the portable ones.

groessler:
I've updated the PR. Please take a look....

greg-king5:
I suggest that you use the concept of "default" colors instead of "reference" colors.  Those defaults are used only once, by the constructor that runs if "cputc.o" is linked into a program (it initiates the palette registers).

`bgcolor()` should work exactly the same way that `textcolor()` works.  Its argument is a two-bit register number; its return value is the "previous" two-bit register number.  The function copies the palette value from the named text register into the background register.  That implementation allows cross-platform conio programs to work with any palette that a programmer wants to load into the text color registers.

The `COLOR_` macroes in <atari5200.h> should have the names of the default colors.  If a programmer changes the palette while a program is running, then those names become "nicknames" for the new colors.

groessler:
> I suggest that you use the concept of "default" colors instead of "reference" colors. Those defaults are used only once, by the constructor that runs if "cputc.o" is linked into a program (it initiates the palette registers).

Hmm... (see later)

> bgcolor() should work exactly the same way that textcolor() works. Its argument is a two-bit register number; its return value is the "previous" two-bit register number. The function copies the palette value from the named text register into the background register. That implementation allows cross-platform conio programs to work with any palette that a programmer wants to load into the text color registers.

If I understand you correctly, I should get rid of the `conio_color` color table, and just initialize the color registers in the contructor of `cputc` (`conioscreen.s`) to exactly these values? And then only use the system `COLORn` registers instead of the color table?

Sounds like a good idea.

> The COLOR_ macroes in <atari5200.h> should have the names of the default colors.

Don't they have them already? (No, not really, COLOR_BLUE and COLOR_GREEN should be called COLOR_LIGHTBLUE and COLOR_LIGHTGREEN. I had already noticed that.)

groessler:
Erroneously spoke about COLOR_BLUE instead of COLOR_RED....

groessler:
I'm considering this PR as "finished" from my side. Please merge if appropriate...

asminc/atari5200.inc
```diff
@@ -101,6 +111,14 @@ ANTIC     = $D400  ; ANTIC area
 POKEY     = $E800  ; POKEY area
 .include  "atari_pokey.inc"
 
+;-------------------------------------------------------------------------
+; conio color defines
+;-------------------------------------------------------------------------
+
+COLOR_WHITE      = 0
+COLOR_LIGHTRED   = 1
+COLOR_LIGHTGREEN = 2
```
oliverschmidt: I asked for
* COLOR_BLACK
* COLOR_WHITE
* COLOR_RED
* COLOR_GREEN

Having COLOR_LIGHTRED and COLOR_LIGHTGREEN doesn't help in writing cross target code, which this was all about.

I presume you did it that way because the colors you use are named that way in _gtia.h. But that isn't a good reason from my pov.

Rather I question the names in _gtia.h. The color that "makes most sense to be used" needs to have the "primary" name. And the secondary color needs to have the secondary name. So if the brighter red and green are to be considered the major colors they need to be named COLOR_RED and COLOR_GREEN - and the other two can e.g. be named COLOR_DARKRED and COLOR_DARKGREEN (or any other name you like).

Anyhow, whatever approach you choose (different names, different colors, name aliases, ...) the result should have COLOR_RED and COLOR_GREEN defined.

libsrc/atari5200/cputc.s
```diff
@@ -11,13 +11,13 @@
...
+        .importzp       screen_width, screen_height
         .importzp       ptr4
-        .import         setcursor
-
-        .constructor    screen_setup, 26
-        .import         screen_setup_20x24
-screen_setup    = screen_setup_20x24
 
+        .import         screen_setup
+        .constructor    screen_setup_constructor, 26
```
oliverschmidt: Please remember to adjust https://github.com/cc65/wiki/wiki/Priority-of-constructors%2C-destructors%2C-and-interruptors accordingly.

Maybe just `initconio` (like on several other targets) instead of `screen_setup_constructor`.

BTW: Do you know if the prio is actually necessary? If yes, it would perhaps be nice if it were explained briefly in a comment.

oliverschmidt:
Thanks for the comprehensive contribution :-)

groessler:
> Having COLOR_LIGHTRED and COLOR_LIGHTGREEN doesn't help in writing cross target code, which this was all about.

Please elaborate. Why is having (e.g.) COLOR_LIGHTRED instead of COLOR_RED less cross-platform friendly? We have four colors here (out of 16 as defined for C64). Which four colors are available shouldn't matter much, do they?
I renamed them with the "light" name to have a 1:1 matching to the GTIA_COLOR defines which in turn derive from the C64 defines.

> Rather I question the names in _gtia.h. The color that "makes most sense to be used" needs to have the "primary" name. And the secondary color needs to have the secondary name. So if the brighter red and green are to be considered the major colors they need to be named COLOR_RED and COLOR_GREEN - and the other two can e.g. be named COLOR_DARKRED and COLOR_DARKGREEN (or any other name you like).

The names are the same as for C64. I chose the light variants for red and green as default colors since on dark background (which was suggested before, and I agree with that) are not that nice.
COLOR_DARKRED would be a new invention which is not being used by any other target.

groessler:
> Please remember to adjust https://github.com/cc65/wiki/wiki/Priority-of-constructors%2C-destructors%2C-and-interruptors accordingly.

ok

> Maybe just initconio (like on several other targets) instead of screen_setup_constructor

ok

> BTW: Do you know if the prio is actually necessary?

I don't think it's necessary. That is a copy-n-paste artifact.

oliverschmidt:
> Please elaborate. Why is having (e.g.) COLOR_LIGHTRED instead of COLOR_RED less cross-platform friendly?

Because the Oric machines and the Lynx have COLOR_RED / COLOR_GREEN but don't have COLOR_LIGHTRED / COLOR_LIGHTGREEN.

Apart from that: Don't you agree that it's just more intuitive to have plain names like COLOR_RED / COLOR_GREEN?

> I renamed them with the "light" name to have a 1:1 matching to the GTIA_COLOR defines which in turn derive from the C64 defines.

I know - and I wrote that I know. But GTIA and C64 have COLOR_RED / COLOR_GREEN too - while they are missing here.

> I chose the light variants for red and green as default colors since on dark background (which was suggested before, and I agree with that) are not that nice.

I wrote that you have several choices. If you think that the colors you have choosen are _THE_ colors to use on black background, then these very colors have be addressable under the names  COLOR_RED / COLOR_GREEN.

* It doesn't matter if the very same color has a different name on the C64.
* If someone plays with the palette and GTIA stuff then the four color names become "just identifies" (aka nick names as Greg wrote) so it doesn't matter if they one name or another.

groessler:
> Because the Oric machines and the Lynx have COLOR_RED / COLOR_GREEN but don't have COLOR_LIGHTRED / COLOR_LIGHTGREEN.

Hmm, ok.

> Apart from that: Don't you agree that it's just more intuitive to have plain names like COLOR_RED / COLOR_GREEN?

No question.

> But GTIA and C64 have COLOR_RED / COLOR_GREEN too - while they are missing here.

Sure. Since they are not "red" or "green" but "light red" and "light green".

My stance is, that the GTIA_COLOR and COLOR (also from C64) defines should match exactly. Means, that COLOR_LIGHTRED is exactly GTIA_COLOR_LIGHTRED. So mapping for the 5200 the COLOR_RED to GTIA_COLOR_LIGHTRED is considered undesirable by me.

I don't know of a good solution to fix these conflicting requirements.

And (if not told before) I am color blind. So my choice of colors is definitely nothing which could be considered definitive. I'm open for other suggestions.

> It doesn't matter if the very same color has a different name on the C64.

I think it does matter. If color names mean different colors on the screen over different targets, it's not really compatible over platforms.

> If someone plays with the palette and GTIA stuff then the four color names become "just identifies"

Sure, yes. But that is not relevant to the discussion here.

regards,
chris

oliverschmidt:
>> But GTIA and C64 have COLOR_RED / COLOR_GREEN too - while they are missing here.

> Sure. Since they are not "red" or "green" but "light red" and "light green".

"light red" means "lighter than the normal red". So without a normal red the term "light red" just doesn't make any sense. So the color in question can by definition not be "light red".

> My stance is, that the GTIA_COLOR and COLOR (also from C64) defines should match exactly.

Nope, not at all. In the very contrast: If there's only one red available for the 5200 CONIO it should likely be neither some C64 red nor another. It should be the red that makes most sense if there's only one red to have. It just doesn't matter at all what reds other targets have and how they call them.

greg-king5:
A semantic of the color functions is that they return their old argument.  The argument of the Atari 5200 `bgcolor()` is a two-bit register index.  The next time that it's called, it should return that same two-bit index.  Here is what I have in mind:
```asm
        .export         _bgcolor

        .include        "atari5200.inc"

.data

old_bg_index:
        .byte   COLOR_BLACK     ; see conioscreen.s for default palette

.code

_bgcolor:
        and     #$03
        tax
        ldy     COLOR0,x
        lda     old_bg_index
        sty     COLOR4          ; set new value
        stx     old_bg_index
        ldx     #0              ; fix high byte
        rts

.end
```

oliverschmidt:
I already wrote back in https://github.com/cc65/cc65/pull/870#issuecomment-479948812:

"Additionally the parameter (aka macro value) itself is saved in order to be returned as old background color value on the next bgcolor() call."

groessler:
@greg-king5, thanks for noticing. I had a brain fart when implementing `bgcolor()`.

groessler:
> "light red" means "lighter than the normal red". So without a normal red the term "light red" just doesn't make any sense. So the color in question can by definition not be "light red".

Strictly speaking, yes.

> If there's only one red available for the 5200 CONIO it should likely be neither some C64 red nor another. It should be the red that makes most sense if there's only one red to have. It just doesn't matter at all what reds other targets have and how they call them.

The reason I wanted to have COLOR_LIGHTRED instead of COLOR_RED is that if someone writes a program both for Atari computers and 5200, and for example sets the background color to COLOR_RED, he will get different reds. This might be counter intuitive, since the reds could be exactly the same. And with "exactly" I mean "exactly exaclty" :-) Same hw chips are used in both machine.
OTOH, if 5200 doesn't have COLOR_RED he will get a compile time error and can notice ahead that there are differences.

But I also understand your argument.

mrdudz:
rename that red that is available on both to COLOR_RED, and the darker one on other ataris to COLOR_DARKRED ? or how many reds are we talking about? :)

groessler:
There are 8 shades/luminances of red supported by the hw.

The GTIA_COLOR (from this PR) and the Atari computers' colors mimic the colors available on the C64.
I would like to keep it that way. Otherwise, for a cross-platform program btw. Atari and C64, COLOR_RED would give again different reds.

oliverschmidt:
> There are 8 shades/luminances of red supported by the hw.

Then the 5200 CONIO should use a medium one. I've written many times by now that it isn't relevant if the red is the same as on any other machine. It's even not relevant if a 5200 TGI has the same red. Because if the TGI has 16 colors then it's fine to have a lighter red and a darker red. But if there's only one red for the 5200 CONIO then that red has to be a "good" red and it needs to be named COLOR_RED. I really don't see what else I can say about this.

oliverschmidt:
Okay, maybe a final try:

It's all about "conformity with expectations" (Erwartungskonformität)...
- Does a user have the right to expect that COLOR_RED looks the same on two targets? No.
- Does a user have the right to expect that red has the same name on all targets? Yes.

Conclusion: Easy.

groessler:
> It's all about "conformity with expectations" (Erwartungskonformität)...

exactly

> Does a user have the right to expect that COLOR_RED looks the same on two targets? No.

In general I agree, but not in the case of Atari computers <--> 5200. I think people expect the colors to be the same.
I also don't want to get rid of the similarity of COLOR defines between Atari computers and C64. I think that's also a quite probable use case.

> Does a user have the right to expect that red has the same name on all targets? Yes.

There is (currently) no red for Atari 5200. As it isn't for e.g. Apple2 AFAIK. So, if there is no red, but only a light red, why not name it that way?
What about e.g. COLOR_YELLOW or COLOR_BROWN? They aren't available. Here it's just the case that COLOR_RED is not available.

I don't know if I was able to get my point over, but I'm a bit tired of arguing. I already wrote
> But I also understand your argument.

So I'd be willing to change the 5200 COLOR names from "LIGHTxxx" to "xxx". Issues posted by users will be directed to you :-)  (But probably I've forgotten this by that time and might answer anyway... :-) )

groessler:
I believe I've addressed all issues. Please review...

libsrc/atari5200/cputc.s
```diff
@@ -11,13 +11,13 @@
...
+        .importzp       screen_width, screen_height
         .importzp       ptr4
-        .import         setcursor
-
-        .constructor    screen_setup, 26
-        .import         screen_setup_20x24
-screen_setup    = screen_setup_20x24
 
+        .import         initconio
+        .constructor    initconio_constructor
```
oliverschmidt: My propsoal was that the "internal" functions have "some" name like e.g. screen_setup and whatever and that the constructor (that goes into the Wiki) has the name "initconio" (not initconio_constructor).

groessler: you mean like just committed?

oliverschmidt: Yes, exactly :-)

PR #873, opened 2019-04-10
Fix optimization bug when an external jmp is directly generated

clbr:
When the optimizers turn a jsr to a jmp, it inherits the function's register info. When there's a direct jmp however, the function's register info was never gathered, and needed loads were removed.


greg-king5:
Closing this PR because it was replaced by PR #874.

PR #875, opened 2019-04-12
atari5200: update docs for recent conio changes

groessler:
also renames libsrc/atari5200/extra/conioscreen_20x12.s to
libsrc/atari5200/extra/conioscreen-20x12.s to be in line with
other optional link modules


doc/atari5200.sgml
```diff
@@ -151,6 +151,43 @@ No serial drivers are available for the Atari 5200.
...
+
+The default console screen has a layout of 20x24 characters. An
+alternative layout, 20x12, Antic mode 7, BASIC mode 2, is provided in
+the file <tt/atari5200-conioscreen-20x12.o/.
+
+Using <tt/atari5200-conioscreen-20x12.o/ is as simple as placing it on
+the linker command line like this:
+
+<tscreen><verb>
+cl65 -t atari5200 myprog.c lib/atari5200-conioscreen-20x12.o
```
greg-king5: ld65 knows where to find standard libraries and extra object files.  We don't need to give the path (just the filename).

groessler: Right. I've changed it. Thanks.

Issue #830, opened 2019-01-06
Optimizer pass "OptStackOps" generates incorrect code for a ternary expression.


holmak:
Minimal repro program:
```C
extern char *dst;

char repro()
{
    char src = 0;
    *dst = (src == 0) ? 0xAA : src;
    return *dst;
}
```
The line with the ternary expression is optimized into incorrect code. Specifically, the code for the "true" case is optimized out. The problem is introduced in the optimizer pass "OptStackOps". I used `--debug-opt-output` to see the effects of each optimizer pass. Here is the before and after of "OptStackOps":

Before:
```asm
;
; char src = 0;
;
	lda     #$00
	jsr     pusha
;
; *dst = (src == 0) ? 0xAA : src;
;
	lda     _dst
	ldx     _dst+1
	jsr     pushax
	ldy     #$02
	lda     (sp),y
	jne     L000B
	lda     #$AA
	jmp     L000C
L000B:	lda     (sp),y
L000C:	ldy     #$00
	jsr     staspidx
;
; }
;
	jmp     incsp1
```
After:
```asm
;
; char src = 0;
;
	lda     #$00
	jsr     pusha
;
; *dst = (src == 0) ? 0xAA : src;
;
	lda     _dst
	ldx     _dst+1
	sta     ptr1
	stx     ptr1+1
	ldy     #$00
	lda     (sp),y		; (A)
	ldy     #$02		; (B)
	jne     L000B		; (C)
	lda     #$AA
	jmp     L000C
	ldy     #$00
L000B:	lda     (sp),y
L000C:	ldy     #$00
	sta     (ptr1),y
;
; }
;
	jmp     incsp1
```
Before the optimization, line (A) loads the value of the ternary expression's "condition" sub-expression. Then, the following `jne` instruction on line (C) uses the zero-flag generated by line (A) to determine which of the ternary expression's branches are taken.

The optimization pass changes and adds several lines. The important change is that it inserts a `ldy` instruction on line (B), which overwrites the zero-flag from (A) that was intended to be used on line (C). Since `ldy #$02` always clears the zero-flag, only the false branch of the ternary expression will ever be used. Subsequent optimization passes remove all the code for the "true" branch of the ternary expression, since it is now dead code.

Please let me know if anything in this description is unclear.

holmak:
It looks like issue #167 might be related; it also involves the "OptStackOps" pass.

oliverschmidt:
Thanks for the qualified bug report !

[...]

PR #877, opened 2019-04-22
Fix for #830 supplied by UvB

IrgendwerA8:
I've added the test, too.
Uz suggested to check several sources if the patch has any drawbacks (which I did with my own ones and the ones in the test folder) -- I found no issues so far (besides that the generated code isn't convincing with it's 'double' `lda     (sp),y` -- but that's independent from the fix).

If you like to check also your sources, here's the shell-script Uz supplied:
```sh
#!/bin/bash
mkdir old new
for SRC in *.c; do
    ASM=$(basename $SRC .c).s
    /path/to/old/cc65 -t <system> -Oirs $SRC
    mv $ASM old
    /path/to/new/cc65 -t <system> -Oirs $SRC
    mv $ASM new
done
diff -r old new
```
Since it works on a single folder with a bunch of C-files, you may have to flatten your sources for this test, and resolve naming-conflicts/include-path locations.

### DIFF
--- a/src/cc65/coptstop.c
+++ b/src/cc65/coptstop.c
@@ -437,51 +437,78 @@ static void AdjustStackOffset (StackOpData* D, unsigned Offs)
 
         CodeEntry* E = CS_GetEntry (D->Code, I);
 
-        int NeedCorrection = 0;
+        /* Check if this entry does a stack access, and if so, if it's a plain
+        ** load from stack, since this is needed later.
+        */
+        int Correction = 0;
         if ((E->Use & REG_SP) != 0) {
 
             /* Check for some things that should not happen */
             CHECK (E->AM == AM65_ZP_INDY || E->RI->In.RegY >= (short) Offs);
             CHECK (strcmp (E->Arg, "sp") == 0);
-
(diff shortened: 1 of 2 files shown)


src/cc65/coptstop.c
```diff
@@ -437,51 +437,78 @@ static void AdjustStackOffset (StackOpData* D, unsigned Offs)
...
+        int Correction = 0;
         if ((E->Use & REG_SP) != 0) {
 
             /* Check for some things that should not happen */
             CHECK (E->AM == AM65_ZP_INDY || E->RI->In.RegY >= (short) Offs);
             CHECK (strcmp (E->Arg, "sp") == 0);
-
             /* We need to correct this one */
-            NeedCorrection = 1;
+            Correction = (E->OPC == OP65_LDA)? 2 : 1;
```
oliverschmidt: If the code would come from another source I'd ask to replace the `1` and `2` magic constants with some enum or macros.  But in this case I think it's supposed to be used as-is ;-)

oliverschmidt:
As there are no comments so far I'd say this asks for being merged...

PR #880, opened 2019-04-26
ld65: give an error for multiply defined global symbols.

### LINKED ISSUE #878: Linker doesn't throw error for multiply defined symbols
Is there a reason that only a warning is generated? Other compilers (linkers) throw an error.

main.c:
```c
#include <stdio.h>
#ifdef __CC65__
#include <conio.h>
#endif

extern int testfun(void);

int main(void)
{
        printf("testfun returned %d\n", testfun());
#ifdef __CC65__
        cgetc();
#endif
        return 0;
}
```
t1.c:
```c
int testfun(void)
{
        return 1;
}
```
t2.c:
```c
int testfun(void)
{
        return 2;
}
```
compile:
```sh
$ cl65 -tatari -o t.com main.c t2.c t1.c 
ld65: Warning: Duplicate external identifier: '_testfun'
$ ls -l t.com
-rw-r--r--. 1 chris chris 3129 Apr 25 13:17 t.com
$ 
```

groessler:
Closes #878.

### DIFF
--- a/doc/ld65.sgml
+++ b/doc/ld65.sgml
@@ -71,23 +71,24 @@ Short options:
   -vm                   Verbose map file
 
 Long options:
-  --cfg-path path       Specify a config file search path
-  --config name         Use linker config file
-  --dbgfile name        Generate debug information
-  --define sym=val      Define a symbol
-  --end-group           End a library group
-  --force-import sym    Force an import of symbol 'sym'
-  --help                Help (this text)
-  --lib file            Link this library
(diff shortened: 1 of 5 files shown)


oliverschmidt:
I personally like it this way. However, I'll give others some time to veto...

greg-king5:
"ld65.sgml" needs to know about the new usage list.

groessler:
Right. I've updated the PR.

greg-king5:
:+1:

Issue #879, opened 2019-04-25
65816 JSR and JMP fail to produce error for truncated range


bbbradsmith:
A JSR or JMP from one bank to another can only work if it gets promoted to a JSL/JML. If the assembler failed to produce a 3-byte operand, the linker should throw an error when it's trying to fill in those values with a truncated 2-byte operand, unless both source and destination bank are the same.

In its current form, JSR and JMP are very prone to errors of this nature. It's very easy to forget that a label is a far target and accidentally use JSR instead of JSL. If you do this, you get non-functional code generated with no warning or error.

The ability of JSR /JMP to promote to JSL/JML increases the possibility of confusion here, especially with the way this assembler works in one pass, it's very easy to fail to annotate a segment with :far or declare it below rather than above in the file. Generating a range error when the linker later discovers it needed a 3-byte operand would be a lot better than just quietly generating a JSR that will never work.

A simple code example demonstrating the problem:
[65816_jsr_range_fail.zip](https://github.com/cc65/cc65/files/3118359/65816_jsr_range_fail.zip)

In this example we have two impossible JSR instructions that will generate invalid code with no error.

As an aside I would ask if there's a way to turn off promotion of JSR/JMP to JSL/JML as well? It doesn't affect the error directly, but requiring it to be explicit as a habit would give the programmer a visual reminder that a label is far and will require an RTL. Positive reinforcement for remembering the type of procedure by not allowing its calls to "hide" as a JSR. (Though I would probably suggest that this is an appropriate promotion in .smart mode, but not with .smart off? Currently it just does it always.)

bbbradsmith:
As another aside, I can't think of a way to forward-declare a label as :far aside from .global, but that has the side effect of making it an exported symbol. Is there any way to do that without?

oliverschmidt:
If I get this right then I'd say that not issuing at least a warning has to be considered as a bug.

bbbradsmith:
I think it should be an error. I can't really think of a use for the 16-bit truncation, but .loword could easily be used to suppress the error if that's what someone really wanted?

.

There is a related truncation to 16-bits that happens with data instructions as well i.e. you can use a label with LDA and then later in the file define that label in a :far segment, and the linker won't give any indication that the earlier use was truncated. It's sort of like how 6502 silently produces abs addressing for zp references if you forget to declare them ahead of time, but in this case produces broken code because it's truncating rather than expanding.

[...]

PR #885, opened 2019-05-01
Range checking for 65816 mode when abs addresses are assumed rather than known

bbbradsmith:
This is a possible resolution to issue: #879 

It looks like a satisfactory resolution to me, but I could use code review and/or assistance in testing before I would be confident that this should be merged.

.

What I found was that when the address size of an operand was not sufficiently known in 65816 mode, it would use `GenWordExpr` which creates an `EXPR_WORD0` type expression that automatically truncates to 16-bit. Otherwise if it is known at that line, the assembler will just generate an instruction of known size.

See: https://github.com/cc65/cc65/blob/392e6e10fc40a046d4e7dbc083a1e0bfd2e07d69/src/ca65/instr.c#L1220

I believe the idea is that by truncating it, the address was prepared in advance for the linker, which in 65816 correctly has to truncate near addresses to remove their bank byte. However, this disabled the ability to check if this was a valid truncation.

My replacement creates a new expression type `EXPR_NEARADDR`, which the linker will treat the same as `EXPR_WORD0` as a 16-bit truncation, but now the assembler can range check it in `SegDone` (unless the relaxed checking mode is used).

This change seems to correctly catch the error of accidentally making a reference to a symbol from a `far` segment before that segment is declared. If the instruction was assumed near, but the symbol was later declared far, this now gets caught during `SegDone`.

One caveat: a later (or prior) export via `.global` + `: far` does not override the address modifier of the symbol's containing segment. I think this is correct behaviour, so that near addresses used within the file can be exported globally, but the consequence might be unintuitive: a `.global : far` will cause a symbol to assume far addressing until its containing segment is reached, which overrides it. (This feels correct to me, but still weird to think through.)

.

Also, was it appropriate to add the EXPR_NEARADDR enum in the middle of the list, or should it have been appended to the end? (Would that invalidate existing object files?)

### DIFF
--- a/src/ca65/expr.c
+++ b/src/ca65/expr.c
@@ -1865,6 +1865,28 @@ ExprNode* GenWordExpr (ExprNode* Expr)
 
 
 
+ExprNode* GenNearAddrExpr (ExprNode* Expr)
+/* A word sized expression that will error if given a far expression at assemble time. */
+{
+    long      Val;
+    /* Special handling for const expressions */
+    if (IsEasyConst (Expr, &Val)) {
+        FreeExpr (Expr);
+        Expr = GenLiteralExpr (Val & 0xFFFF);
+        if (Val > 0xFFFF)
+        {
+            Error("Range error: constant too large for assumed near address.");
+        }
+    } else {
+        ExprNode* Operand = Expr;
+        Expr = NewExprNode (EXPR_NEARADDR);
+        Expr->Left = Operand;
+    }
+    return Expr;
+}
+
+
+
 ExprNode* GenFarAddrExpr (ExprNode* Expr)
 /* Force the given expression into a far address and return the result. */
 {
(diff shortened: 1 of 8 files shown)


oliverschmidt:
Unfortunately I can't add any value to this topic - I have no idea about the 65816 at all.

The only thing I can say is that the commits provided so far don't look so large and/or complex that I'd have to reject them for further-code-base-maintenance reasons. However, I'd like to see others comment on the question if the proposed change is correct/helpful/desirable/<you name it>.

oliverschmidt:
Last call: If anybody has reasons against merging this pull request, please speak up and provide an explanation. Otherwise I'll merge it in the next days even if I don't understand it - simply because I don't see a reason not to do so...

oliverschmidt:
Is https://github.com/cc65/cc65/issues/879 supposed to be closed?

bbbradsmith:
Yes, this fixed that issue. I will close it. Thanks!

oliverschmidt:
It's me who has to thank for the fix!

PR #896, opened 2019-05-14
Document using inline asm with SoA, structs with array members

clbr:


doc/cc65.sgml
```diff
@@ -1448,10 +1448,16 @@ Or, to access a struct member of a static variable:
             unsigned char x;
             unsigned char y;
             unsigned char color;
+            unsigned char aux[32];
```
oliverschmidt: Sorry for nitpicking as always but I'd like to have this not named `aux` but `data` or `buffer`.

PR #898, opened 2019-05-14
Document --debug-opt-output and --debug-opt

bbbradsmith:


doc/cc65.sgml
```diff
@@ -193,8 +194,26 @@ Here is a description of all the command line options:
 
   <tag><tt>-d, --debug</tt></tag>
 
-  Enables debug mode, something that should not be needed for mere
-  mortals:-)
+  Enables debug mode, for debugging the behavior of CC65.
```
oliverschmidt: According to my understanding both the project and actual compiler binary are named cc65 (lower case).

Issue #895, opened 2019-05-14
16-bit optimization error uses uninitialized carry?


bbbradsmith:
The following produces invalid code when -O is used:
```c
unsigned int a;
unsigned int b;

void foo(void)
{
	a = 259;
	if (a < 7)
	{
		b = 473;
	}
}
```
Resulting assembly:
```asm
.proc	_foo: near

.segment	"CODE"

	ldx     #$01
	lda     #$03
	sta     _a
	stx     _a+1
	bcs     L0004
	lda     #$D9
	sta     _b
	stx     _b+1
L0004:	rts
```
Note that the comparison has entirely disappeared. With an `unsigned char` for `a`, the optimizer seems to be able to determine if the condition is always true or false, and can even remove the whole assignment to `b`. But, with `unsigned int`, it bungles the optimization, and simultaneously leaves the assignment in there, and branches over it (or fails to branch) with an uninitialized carry? Very odd.

Source: https://forums.nesdev.com/viewtopic.php?f=2&t=18843&p=238785

bbbradsmith:
Have narrowed down the culprit to: OptStore4

bbbradsmith:
Now that I discovered --debug-opt-output I can see where it goes wrong:
```
=========================================================================
Initial code for function 'foo':
.segment	"CODE"

	ldx     #$01
	lda     #$03
	sta     _a
	stx     _a+1
	lda     _a
	ldx     _a+1
	cpx     #$00
	bne     L0006
	cmp     #$07
L0006:	jsr     boolult
	jeq     L0004
	ldx     #$01
	lda     #$D9
	sta     _b
	stx     _b+1
L0004:	rts
=========================================================================
Code after applying 'OptStore4':
.segment	"CODE"

	ldx     #$01
	lda     #$03
	sta     _a
	stx     _a+1
	cpx     #$00
	bne     L0006
	cmp     #$07
L0006:	jsr     boolult
	jeq     L0004
	ldx     #$01
	lda     #$D9
	sta     _b
	stx     _b+1
L0004:	rts
=========================================================================
Code after applying 'OptBoolTrans':
.segment	"CODE"

	ldx     #$01
	lda     #$03
	sta     _a
	stx     _a+1
	cpx     #$00
	bne     L0006
	cmp     #$07
L0006:	jcs     L0004
	ldx     #$01
	lda     #$D9
	sta     _b
	stx     _b+1
L0004:	rts
=========================================================================
Code after applying 'OptCmp8':
[...]

PR #899, opened 2019-05-15
Disable too-aggressive optimization in OptCmp8

bbbradsmith:
This optimization generates incorrect code for some 16-bit cases (possibly others). See: #895

I'm not sure if it would be better to re-evaluate how the compiler generates a shared branch so that this optimization could be kept, but this change is simple and will prevent it from being broken by the attempted optimization.


clbr:
Please add a test case of the bug to test/val, to increase coverage and make sure it stays fixed.

bbbradsmith:
Thank you @clbr for that suggestion. I have added a unit test.

PR #902, opened 2019-05-21
gr65: Fix realloc bug

rhalkyard:
The pointer to the input buffer was not being updated after a call to `realloc()`, causing the program to crash if `realloc()` moved the buffer (which on my system, seems to be always).

This is just a quick fix, tested as far as "works on my machine, and valgrind doesn't complain about invalid accesses anymore."  Let me know if this needs any changes or further testing.


oliverschmidt:
First of all thanks for contributing to cc65 :-)

I don't think that this change asks for a regression test. Merging...

PR #903, opened 2019-05-25
some documentation for sim65

bbbradsmith:
When working with the tests I noticed sim65 is a standard part of the toolchain, and realized it could be very useful for unit testing on many 6502 projects that use cc65, not just internally. Added a little bit of documentation to help with this.

### DIFF
--- a/doc/sim65.sgml
+++ b/doc/sim65.sgml
@@ -106,14 +106,18 @@ PVExit ($01)
 
 <sect>Creating a Test in C<p>
 
-For a C test compiled and linked with <tt/--target sim65/ the
+For a C test compiled and linked with <tt/--target sim6502/ the
 command line arguments to <tt/sim65/ will be passed to <tt/main/,
 and the return value from <tt/main/ will become sim65's exit code.
 The <tt/exit/ function may also be used to terminate with an exit code.
  
 Exit codes are limited to 8 bits.
 
-In addition to this, the simulator provides a set of built-in functions
+The standard C library file input and output is functional,
+with <tt/STDIN_FILENO/, <tt/STDOUT_FILENO/ and <tt/STDERR_FILENO/ mapped to sim65's
+STDIN_FILENO, STDOUT_FILENO and STDERR_FILENO.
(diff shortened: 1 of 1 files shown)


doc/sim65.sgml
```diff
@@ -104,6 +104,60 @@ PVExit ($01)
...
+Exit codes are limited to 8 bits.
+
+In addition to this, the simulator provides a set of built-in functions
+for simple file input and output:
+
+<tscreen><verb>
+  int open (const char* name, int flags, ...);
+  int __fastcall__ close (int fd);
+  int __fastcall__ read (int fd, void* buf, unsigned count);
+  int __fastcall__ write (int fd, const void* buf, unsigned count);
```
oliverschmidt: It seems that it would be worth mentioning here that:
- STDIN_FILENO, STDOUT_FILENO and STDERR_FILENO are mapped to sim65's STDIN_FILENO, STDOUT_FILENO and STDERR_FILENO
- these functions allow to use the FILE*-based stdio as well (incl. stdin, stdout and stderr being mapped to sim65's stdin, stdout and stderr)

oliverschmidt:
Improvements of the documentation are always particularly welcome :-)

doc/sim65.sgml
bbbradsmith: Added.

polluks:
You may add your name :-)

doc/sim65.sgml
oliverschmidt: I think we have a misunderstanding here. I see two separate things (therefore the two bullets in my last comment):

1. open() and be used with STDIN_FILENO, STDOUT_FILENO and STDERR_FILENO (incl. short comment what it does).
2. fopen() can be used with stdin, stdout and stderr (incl. short comment what it does).

Your term "The standard C library file input and output" doesn't make it clear if you refer to 1.) or 2.). According to my understanding there's no single common terminology. I suggest something like "high-level FILE based functions" vs. "low-level file descriptor based functions".

PR #905, opened 2019-05-28
sim6502 protection assert for hardcoded sp address

bbbradsmith:
sim65 hard codes the parameter stack pointer `sp` at $0000 for its paravirtualization functions that need it (`args`, `open`, `read`, `write`). If you're trying to use it with a custom .CFG, or basically try to add anything to the `ZEROPAGE` segment it's easy to break these by accident. Since it breaks `printf` too, it's really hard to diagnose the problem.

Added an assert to `paravirt.s` to make it clear when this assumption is being broken.

Shifted the definition of `exit` to `crt0.s` so that the assert is not triggered if `main` is used without args, or the other affects functions are never linked in from the library.

### DIFF
--- a/libsrc/sim6502/paravirt.s
+++ b/libsrc/sim6502/paravirt.s
@@ -8,10 +8,10 @@
 ;
 
         .importzp       sp
-        .export         args, exit, _open, _close, _read, _write
+        .export         args, _open, _close, _read, _write
 
 args            := $FFF0
-exit            := $FFF1
+;exit           := $FFF1 ; crt0.s (not affected by assert below)
 _open           := $FFF2
 _close          := $FFF3
 _read           := $FFF4
--- a/libsrc/sim6502/paravirt.s
+++ b/libsrc/sim6502/paravirt.s
@@ -7,6 +7,7 @@
 ; int __fastcall__ write (int fd, const void* buf, unsigned count);
 ;
 
+        .importzp       sp
         .export         args, exit, _open, _close, _read, _write
 
 args            := $FFF0
@@ -15,3 +16,5 @@ _open           := $FFF2
 _close          := $FFF3
 _read           := $FFF4
 _write          := $FFF5
+
+.assert sp=0, error, "sim6502 paravirtualization functions require parameter stack pointer (sp) at $00"
--- a/libsrc/sim6502/paravirt.s
+++ b/libsrc/sim6502/paravirt.s
@@ -11,7 +11,7 @@
         .export         args, _open, _close, _read, _write
 
 args            := $FFF0
-;exit           := $FFF1 ; crt0.s (not affected by assert below)
+;exit           := $FFF1 ; in crt0.s, doesn't need .assert below
 _open           := $FFF2
 _close          := $FFF3
 _read           := $FFF4
--- a/libsrc/sim6502/crt0.s
+++ b/libsrc/sim6502/crt0.s
@@ -5,14 +5,16 @@
 ;
 
         .export         _exit
+        .export         exit
         .export         __STARTUP__ : absolute = 1      ; Mark as startup
         .import         zerobss, callmain
         .import         initlib, donelib
-        .import         exit
         .import         __MAIN_START__, __MAIN_SIZE__   ; Linker generated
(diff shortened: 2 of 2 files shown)


libsrc/sim6502/paravirt.s
```diff
@@ -7,11 +7,14 @@
 ; int __fastcall__ write (int fd, const void* buf, unsigned count);
 ;
 
-        .export         args, exit, _open, _close, _read, _write
+        .importzp       sp
+        .export         args, _open, _close, _read, _write
 
 args            := $FFF0
-exit            := $FFF1
+;exit           := $FFF1 ; crt0.s (not affected by assert below)
```
greg-king5: I think that the comment should be a little different: "; in crt0.s, doesn't need .assert below".

oliverschmidt:
Some thoughts of mine:

* CFGs need in general to match the hardware. If one e.g. uses $0000 and $0001 on a 6510 it just won't work. sim65 is a virtual hardware, so there in general nothing wrong with it not adapting to different CFGs.

* In contrast to real hardware the sim65 virtual hardware can complain about a wrong CFG. This is what this PR is about.

* I don't see that the question of a wrong CFG is limted to the paravirt functions. To me the question of a correct or wrong CFG is property of the hardware. So I don't see why the .assert shouldn't be placed in the crt0.s (and of course I don't see that exit is handled in a special way).

* Maybe it isn't that hard to actually make sim65 adapt to the CFG used. Meaning that https://github.com/cc65/cc65/blob/master/src/sim65/paravirt.c#L123 and friends are given the right address.

bbbradsmith:
CFG absolutely needs to match the hardware, but 6510 is well known and well documented, and sim65 is a completely made up machine. This PR is entirely about giving the user feedback when they've made a mistake. I added this assert because I think it's error prone. None of the other target platforms rely on a hard coded `sp`, which is a unique and non-intuitive constraint here.

More specifically, I ran into this because I wanted to make a test of some assembly code that requires a ZP allocation, with a C program to run the actual test and give feedback via printf. On any other cc65 target platform, you can generally just increase the size of or move ZP/ZEROPAGE without much problem, but not with sim65. It's not really about the size of that segment itself, but a more subtle and very complicated issue about a CRT internal variables, calling convention, link order, etc.


Instead, the assert identifies exactly what goes wrong: if you use one of the affected `paravirt.s` functions, it will break if `sp` isn't at 0. That's easy to check at link time.

The `exit` function does not use `sp`, so it's in crt0.s to _avoid_ the assert. The reason for this is that it's safe to make a sim65-compatible binary that doesn't use the affected functions, and it's also valid to do so if you don't need file i/o (e.g. unit tests on exit code only). In fact there are some tests that rely on this like test/val/bss-name.c which is what made me realize that it should be exempt.


If you wanted to adapt sim65 instead, you would need to add it to its binary format. 2 more bytes in the (currently 1 byte) header could tell it where `sp` is. Can be done, but I think the assert for a link time error can do the job well, and is backward compatible.


In general I think the made up constraints of sim65 are pretty easy to remember, _except_ for the paravirtualization functions, which depend on a lot of complicated things like cc65's calling conventions, knowing about the parameter stack, ordering of things that get linked into zero page, etc. I don't think that would be useful to try and document, because the correct approach it just to use them from C. This added assert is intended as a guiding hand to keep the user from setting up the C target incorrectly.

libsrc/sim6502/paravirt.s
bbbradsmith: Ok. Changed.

bbbradsmith:
For comparison, I created the alternative in #906 which instead adds the location of `sp` to the sim65 header.

That solution is probably more intuitive for users, but it will invalidate any existing binaries or custom CFG currently in use.

oliverschmidt:
> For comparison, I created the alternative in https://github.com/cc65/cc65/pull/906 which instead adds the location of sp to the sim65 header.

Thanks. Let's follow up on that approach...

> That solution is probably more intuitive for users, 

Ack.

> but it will invalidate any existing binaries or custom CFG currently in use.

I don't consider this an issue for sim65.

PR #906, opened 2019-05-28
sim65 movable parameter stack pointer

bbbradsmith:
As an alternative to #905 this adds the location of `sp` to the sim65 binary header.

This is probably more intuitive than the assert approach, but it is not backwards compatible. PR includes a version bump for this reason.

### DIFF
--- a/libsrc/sim6502/exehdr.s
+++ b/libsrc/sim6502/exehdr.s
@@ -1,8 +1,7 @@
 ;
 ; Oliver Schmidt, 2013-05-16
 ;
-; This module supplies a 2 byte header identifying the simulator type,
-; and parameter stack pointer sp.
+; This module supplies a header used by sim65.
 ;
 
         .export         __EXEHDR__ : absolute = 1       ; Linker referenced
--- a/libsrc/sim6502/exehdr.s
+++ b/libsrc/sim6502/exehdr.s
@@ -7,8 +7,14 @@
 
         .export         __EXEHDR__ : absolute = 1       ; Linker referenced
         .importzp       sp
+        .import         __MAIN_START__
+        .import         sim65start
 
         .segment        "EXEHDR"
 
-        .byte   .defined(__SIM65C02__)
-        .byte   sp
+        .byte   $73, $69, $6D, $36, $35        ; 'sim65'
+        .byte   2                              ; header version
+        .byte   .defined(__SIM65C02__)         ; CPU type
+        .byte   sp                             ; sp address
+        .addr   __MAIN_START__                 ; load address
+        .addr   sim65start                     ; reset address
--- a/libsrc/sim6502/exehdr.s
+++ b/libsrc/sim6502/exehdr.s
@@ -1,11 +1,14 @@
 ;
 ; Oliver Schmidt, 2013-05-16
 ;
-; This module supplies a 1 byte header identifying the simulator type
+; This module supplies a 2 byte header identifying the simulator type,
+; and parameter stack pointer sp.
 ;
 
         .export         __EXEHDR__ : absolute = 1       ; Linker referenced
+        .importzp       sp
 
         .segment        "EXEHDR"
 
         .byte   .defined(__SIM65C02__)
+        .byte   sp
--- a/src/sim65/main.c
+++ b/src/sim65/main.c
@@ -137,7 +137,7 @@ static unsigned char ReadProgramFile (void)
 {
     int Val;
     unsigned Addr = 0x0200;
-    unsigned char SPAddr = 0x0000;
+    unsigned char SPAddr = 0x00;
 
     /* Open the file */
     FILE* F = fopen (ProgramFile, "rb");
--- a/src/sim65/main.c
+++ b/src/sim65/main.c
@@ -132,11 +132,12 @@ static void OptQuitXIns (const char* Opt attribute ((unused)),
     MaxCycles = strtoul(Arg, NULL, 0);
 }
 
-static void ReadProgramFile (void)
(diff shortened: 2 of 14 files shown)


oliverschmidt:
I like this Approach. However, I have several ideas:

- The program files could/should have a somewhat more sophisticated header to allow for future extensions. This doesn't necessarily imply to build code to load old files but at least to be able to bail out on old files. I suggest to follow the approach used for loadable drivers: Three byte signature + one byte version - e.g.: https://github.com/cc65/cc65/blob/master/libsrc/c64/tgi/c64-hi.s#L27-L28

- When the addr of the zeropage becomes a parameter it makes sense to make the startaddr a parameter as well.

- When we add additional paravirt functions in the future they'll end up at lower addresses. So we could/should maybe reverse the order of the existing ones now that we become incompatible anyhow. That way you wouldn't have to mention the ones for the C library in your new sim65 doc at all ;-)
```
_open           := $FFF0
_close          := $FFF1
_read           := $FFF2
_write          := $FFF3
args            := $FFF4
exit            := $FFF5
```

bbbradsmith:
I agree that if we're going to start adding to the header, we might as well make it versionable.

I added a signature, version byte, and 2 words for load address and reset address, so that code can both be placed and run from different locations than $0200.

I'm not sure I understand the motivation to reorder the paravirtualization functions? I am hesitant to make any changes there, as the rest of this pull does nothing to invalidate them, the only new thing is the header.

The only function whose explicit address I documented was `exit`, because it's the only one that's reasonable to use from assembly. The rest really need the C target anyway. Theoretically new ones could be added in the future, but I don't think reordering the existing ones at this point would help with that?

oliverschmidt:
> I'm not sure I understand the motivation to reorder the paravirtualization functions? I am hesitant to make any changes there, as the rest of this pull does nothing to invalidate them, the only new thing is the header.

There's no strong reason for that. I just like to think of general things like exit/args being "old"/basic and other things newer/more specific. And as we use locations just below the 6502 vectors I see those things "grow" downwards. And as we become totally incompatible anyhow because of the file header change we can as well "beautify" the order of the paravirt functions, okay?

libsrc/sim6502/exehdr.s
```diff
@@ -1,11 +1,20 @@
 ;
 ; Oliver Schmidt, 2013-05-16
 ;
-; This module supplies a 1 byte header identifying the simulator type
+; This module supplies a 2 byte header identifying the simulator type,
```
oliverschmidt: More than 2 bytes by now...

bbbradsmith: OK fixed.

src/sim65/main.c
```diff
@@ -132,18 +142,34 @@ static void OptQuitXIns (const char* Opt attribute ((unused)),
...
+    unsigned char SPAddr = 0x00;
 
     /* Open the file */
     FILE* F = fopen (ProgramFile, "rb");
     if (F == 0) {
         Error ("Cannot open '%s': %s", ProgramFile, strerror (errno));
     }
 
+    /* Verify the header signature */
+    for (I=0; I<HEADER_SIGNATURE_LENGTH; ++I) {
```
oliverschmidt: Our style wants more spaces: `I = 0; I < HEADER_SIGNATURE_LENGTH`

src/sim65/memory.c
```diff
@@ -64,6 +64,15 @@ void MemWriteByte (unsigned Addr, unsigned char Val)
 
 
 
+void MemWriteWord (unsigned Addr, unsigned Val)
```
oliverschmidt: I had deliberately not put this here as a 65xx never writes a word meaning that this is conceptually always limited to paravirt stuff. But I'm as well okay with it going here.

src/sim65/main.c
bbbradsmith: OK.

bbbradsmith:
Well, if you do want to grow the table, I would suggest using up the existing remaining space first ($FFF6-FFF9), otherwise it would be wasting even more usable space for no reason.

Otherwise, if it's not supposed to be a functional change, I'm not comfortable reordering internal details like this without more justification.

oliverschmidt:
> I'm not comfortable reordering internal details like this without more justification.

If it's such an issue for you I'll just do it on my own ;-)

oliverschmidt:
Thanks for the contribution :-)

bbbradsmith:
Well, if you want it done, it should be done now with the version change.

You didn't give me an actionable list, though, you said the list will need to grow "down", but didn't respond when I pointed out there's still space at the end.

bbbradsmith:
Also, please tag 2.18 when ready (but do the reorder first if you're going to).

oliverschmidt:
> Well, if you want it done, it should be done now with the version change.

That's why I asked you to include it in the pull request...

bbbradsmith:
??? I can't put it in the pull request, you had already merged it.

Are you asking me to make another pull request to implement a reordering? If so, what is the actual order you want to use? You never responded about there already being unused space here.

oliverschmidt:
Easier than solving the misunderstandings: https://github.com/cc65/cc65/commit/6efb71bea773837c6fc4aae0c8270b2d70e3e69f

> Also, please tag 2.18 when ready (but do the reorder first if you're going to).

Thanks for the hint - done.

bbbradsmith:
Thanks for following up with the rest.

spiro-trikaliotis:
I also updated the OBS Debian packages to v2.18.

oliverschmidt:
Thanks :-)

PR #911, opened 2019-06-08
Replace GIT_SHA with a more versatile BUILD_ID definition.

besser82:
When compiling cc65, it will by default place the git hash (if available) of the checked out commit in the version string.  This isn't useful when building a package for a Linux distribution, since there either won't be an upstream git hash if there is one at all.

Thus we replace GIT_SHA with a more versatile BUILD_ID, which can be defined to any arbitrary string.  When building, its contents will be appended to the version string instead of the git hash.

If BUILD_ID is not defined by the user the behaviour will be exactly the same as before.  That means BUILD_ID gets automatically defined to Git <GIT_SHA>, if it can be determined from a checkout.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -52,13 +52,13 @@ ifdef USER_CFLAGS
   $(info USER_CFLAGS: $(USER_CFLAGS))
 endif
 
-ifdef GIT_SHA
-  $(info GIT_SHA: $(GIT_SHA))
+ifdef BUILD_ID
+  $(info BUILD_ID: $(BUILD_ID))
 else
-  GIT_SHA := $(shell git rev-parse --short HEAD 2>$(NULLDEV) || svnversion 2>$(NULLDEV))
-  ifneq ($(words $(GIT_SHA)),1)
-    GIT_SHA := N/A
-    $(info GIT_SHA: N/A)
+  BUILD_ID := Git $(shell git rev-parse --short HEAD 2>$(NULLDEV) || svnversion 2>$(NULLDEV))
+  ifneq ($(words $(BUILD_ID)),2)
+    BUILD_ID := N/A
+    $(info BUILD_ID: N/A)
   endif
 endif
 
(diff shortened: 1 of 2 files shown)


spiro-trikaliotis:
The GIT_SHA mechanism is for exactly that purpose! Just call make with GIT_SHA=... and you get what you want.
No need for a second mechanism.

That's what I use for the OBS build packages for Debian myself (cf. https://build.opensuse.org/package/show/home:strik/cc65)

oliverschmidt:
Maybe a misunderstanding and/or an unclear PR description (at least to me).

The GIT_SHA variable is there to to provide the Git hash explicitly when it can't be implizitly determined by calling `git rev-parse --short HEAD`.

Here it seems that the PR author asks for something different than a Git hash.

I personally don't see the point in doing so. There's a version major and minor. That is supposed to be enough for any "externally built" delivery in the first place. I don't see anybody wanting to deliver cc65 in a higher frequency than the frequency I change the version.

The `Git <...>` has the purpose to allow to distinguish all those builds created automatically fro every commit. If an external build isn't able to provide the right Git hash via GIT_SHA it can be left at the `N/A` default.

I personally don't know what a "package version" is and I'm not interested in learning it nor adding explicit support for it to this repo.

However, I'm in general willing to merge a PR that replaces the `xsnprintf (Buf, sizeof (Buf), "%u.%u - Git %s"` with a xsnprintf (Buf, sizeof (Buf), "%u.%u - %s"`` and changes the Makefile to add the `Git ` prefix for the case when GIT_SHA is _not_ provided. But @spiro-trikaliotis would have to change his "build file" to add the `Git ` prefix as part of the GIT_SHA variable. So he needs to agree.

And when we're at changing things we could give GIT_SHA a more general name, maybe `BUILD_ID` or alike.

Then you could "abuse" that BUILD_ID for whatever you see fit.

besser82:
I've addressed your comments in the rebased commit.

spiro-trikaliotis:
The change is ok for me, I just added BUILD_ID alongside with GIT_SHA. Once the transition is there, I will remove GIT_SHA.

src/Makefile
```diff
@@ -52,21 +52,21 @@ ifdef USER_CFLAGS
...
-  $(info GIT_SHA: $(GIT_SHA))
+ifdef BUILD_ID
+  $(info BUILD_ID: $(BUILD_ID))
 else
-  GIT_SHA := $(shell git rev-parse --short HEAD 2>$(NULLDEV) || svnversion 2>$(NULLDEV))
-  ifneq ($(words $(GIT_SHA)),1)
-    GIT_SHA := N/A
-    $(info GIT_SHA: N/A)
+  BUILD_ID := Git $(shell git rev-parse --short HEAD 2>$(NULLDEV) || svnversion 2>$(NULLDEV))
+  ifneq ($(words $(BUILD_ID)),1)
```
greg-king5: There now are two words.

besser82: Fixed in rebased commit.

oliverschmidt:
@besser82: Thanks for the quick adjustment.
@spiro-trikaliotis : Thanks for you willingness to adapt your code.
@greg-king5: Thanks for your - as always - valuable review.

PR #913, opened 2019-06-09
util/zlib/deflater: Fix several compiler warnings.

besser82:
As the title says, just some small fixes.

### DIFF
--- a/util/zlib/deflater.c
+++ b/util/zlib/deflater.c
@@ -16,8 +16,8 @@
 int main(int argc, char* argv[])
 {
         FILE* fp;
-        char* inbuf;
-        char* outbuf;
+        unsigned char* inbuf;
+        unsigned char* outbuf;
         size_t inlen;
         size_t outlen;
         z_stream stream;
@@ -84,7 +84,7 @@ int main(int argc, char* argv[])
         }
 
         /* display summary */
-        printf("Compressed %s (%d bytes) to %s (%d bytes)\n",
+        printf("Compressed %s (%zu bytes) to %s (%zu bytes)\n",
                 argv[1], inlen, argv[2], outlen);
         return 0;
 }
--- a/include/zlib.h
+++ b/include/zlib.h
@@ -48,7 +48,8 @@
 #define Z_NULL       0
 
 
-unsigned __fastcall__ inflatemem (char* dest, const char* source);
+unsigned __fastcall__ inflatemem (unsigned char* dest,
+                                  const unsigned char* source);
 /*
      Decompresses the source buffer into the destination buffer.
    Returns the size of the uncompressed data (number of bytes written starting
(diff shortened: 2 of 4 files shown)


util/zlib/deflater.c
```diff
@@ -84,7 +84,8 @@ int main(int argc, char* argv[])
         }
 
         /* display summary */
-        printf("Compressed %s (%d bytes) to %s (%d bytes)\n",
-                argv[1], inlen, argv[2], outlen);
+        printf("Compressed %s (%lu bytes) to %s (%lu bytes)\n",
+                argv[1], (unsigned long) inlen,
+                argv[2], (unsigned long) outlen);
```
greg-king5: cc65's `printf()` family of functions know the `h, j, t,` and `z` size modifiers.  Those functions don't know the `hh` modifier.

besser82: Fixed in rebased commit.

oliverschmidt:
Sorry but could one of you both elaborate a bit?

- According to my understanding chars are by default unsigned with cc65.
- https://github.com/cc65/cc65/blob/master/include/zlib.h is about chars, not unsigned chars.

besser82:
From my understanding this program is considered to be compiled and run on a system, that has a 'real' zlib available to compress data, that can later be decompressed using cc65's zlib `inflatemem()`.  Those systems are not guaranteed to have `unsigned char` as default, nor `size_t` to be defined as an `unsigned int`.

groessler:
It's never bad to explicitly put the signedness of a char variable in its definition ("signed char" or "unsigned char").
Don't know if the mixed chars in `zlib.h` are on purpose or accidently there.

oliverschmidt:
>From my understanding this program is considered to be compiled and run on a system, that has a 'real' zlib available to compress data, that can later be decompressed using cc65's zlib inflatemem().

Thanks for the clarification.

> Those systems are not guaranteed to have unsigned char as default, nor size_t to be defined as an unsigned int.

This isn't guaranteed with cc65 either. There's the `--signed-char` cmdline option. From that perspective it becomes clear why there are so many `unsigned char` declaration in cc65 header files.

So if you think the variables in question have to be unsigned then the same holds for the function declarations in https://github.com/cc65/cc65/blob/master/include/zlib.h, the code in https://github.com/cc65/cc65/blob/master/libsrc/zlib/uncompress.c (and the comment in https://github.com/cc65/cc65/blob/master/libsrc/zlib/inflatemem.s).

Can you imagine to adjust those locations too?

besser82:
> Can you imagine to adjust those locations too?

Done in https://github.com/cc65/cc65/pull/913/commits/364b3a575b0a391918b918880d875e13c4ae72b3.

oliverschmidt:
Thanks for taking care of my request and for your patience. Merging now...

PR #914, opened 2019-06-09
common/version: Increase static buffer to 100 chars.

besser82:
In a previous commit we introduced BUILD_ID to replace GIT_HASH.  As this allows for strings of an arbitrary length to be appended to this buffer, we should increase its size a bit so longer strings will not get truncated.

### DIFF
--- a/src/common/version.c
+++ b/src/common/version.c
@@ -60,7 +60,7 @@
 const char* GetVersionAsString (void)
 /* Returns the version number as a string in a static buffer */
 {
-    static char Buf[60];
+    static char Buf[100];
 #if defined(BUILD_ID)
     xsnprintf (Buf, sizeof (Buf), "%u.%u - %s", VER_MAJOR, VER_MINOR, STRINGIZE (BUILD_ID));
 #else


spiro-trikaliotis:
What do you intend to put into it? Please do not write your homework into the version string. ;)

besser82:
No, not my whole homework.  :P

But as I'm planning to package cc65 into the official Fedora (and Fedora EPEL) repos, I intend to define `BUILD_ID` to something like: `%{release} - %{dist_string} - %{_target_platform}`, which will expand to `0.1.el7 - Fedora EPEL 7 - aarch64-redhat-linux-gnu` during build.  60 chars can be slightly to few in such cases.

The actual reasoning for such a string is: Make bug reporting easy for users.

```
$ cc65 --version
cc65 V2.18 - 0.1.el7 - Fedora EPEL 7 - aarch64-redhat-linux-gnu
```
will carry all basic information for initial work on reported bugs.

besser82:
Closing here, as I ended up using a different `BUILD_ID`.

PR #922, opened 2019-07-16
cc65: Add support for binary literals

clbr:
Binary literals, 0b001, are a GCC extension in C and a C++14 feature.


mrdudz:
i like this :)

the test should perhaps also test literals with leading zeros (0b00010001)

greg-king5:
It should be documented in https://cc65.github.io/doc/cc65.html#s5.

oliverschmidt:
@clbr: Thanks for the contribution (incl. test). The code base seemed to have actually asked for it ;-)

@mrdudz: Thanks for your thoughts regarding the test.

@greg-king5: Thanks for the hint regarding the documentation.

As far as I understand this PR is ready to be merged...

PR #924, opened 2019-07-26
Add cclear and cclearxy for Telestrat target

jedeoric:
Here is the cclear and cclearxy for Telestrat target


oliverschmidt:
As there was no feedback so far I just presume this one is ready for prime time...

PR #929, opened 2019-08-24
doc: clarify need for .IMPORT on some special symbols

koitsu:


doc/ld65.sgml
```diff
@@ -562,7 +562,8 @@ segment, where this attribute is true, the linker will export three symbols.
 </verb></tscreen>
 
 Replace <tt/NAME/ by the name of the segment, in the example above, this would
-be <tt/BSS/. These symbols may be accessed by your code.
+be <tt/BSS/. These symbols may be accessed by your code once the symbol has
+been imported using the <tt>.IMPORT</tt> directive.
```
greg-king5: "Symbols" is plural.  Therefore, it should say, "`... once they have been imported ...`".

Also, it might be a good idea to say, "`... using ca65's <tt>.IMPORT</tt> directive.`"

doc/ld65.sgml
```diff
@@ -766,7 +767,8 @@ useful for things like a software stack, or an I/O area.
         }
 </verb></tscreen>
 
-This will define some external symbols that may be used in your code:
+This will define some external symbols that may be used in your code once
+the symbol has imported using the <tt>.IMPORT</tt> directive:
```
greg-king5: Again, "symbols" is plural, so "`... once they have been imported ...`"

koitsu:
Thanks; good catch on plural vs. singular. I've changed the grammar slightly, otherwise these begin to sound like run-on sentences.

oliverschmidt:
@greg-king5: Thanks for taking care of this PR. Please merge as you see fit.

PR #931, opened 2019-08-27
Assignment 1

sap9433:
Hi professor,

I see in the course website submission date was 24th August. But I thought you said you would provide a vagrant environment first, also I'm not sure which repo we should commit our assignments to, so sending a PR here.

Anyways I tried to do it on my OSX. Fibonacci works in regular c , but when I compile it for 6502, it gives me the wrong result, maybe because it has a smaller size for int, not sure


oliverschmidt:
> [...] also I'm not sure which repo we should commit our assignments to [...]

They absolutely certainly don't belong here.

PR #942, opened 2019-10-02
[Gamate] Remove "eor RVS" in cputc.s + [Supervision] hello world example

Fabrizio-Caruso:
1. [Gamate] Fix the wrong display of reversed characters in REAL hardware (this is not seen in Mame).
2. [Supervision] hello world example (badly needed because getting a working binary is not obvious from the doc and there is no example).

image

Tested by Plamen Vaysilov @PlamenVaysilov (https://wwwwww.facebook.com/vaysilov) on real hardware.

### DIFF
--- a/libsrc/gamate/cputc.s
+++ b/libsrc/gamate/cputc.s
@@ -101,7 +101,6 @@ putchar:
         ldy     #$f8
 @copylp1:
         lda     (ptr3),y
-        eor     RVS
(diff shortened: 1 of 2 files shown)


Fabrizio-Caruso:
I am going to do two separate pull requests

oliverschmidt:
> I am going to do two separate pull requests

Thanks :-)

PR #944, opened 2019-10-03
[Supervision][fix] Screen init code fix + Hello world example for the Supervision

Fabrizio-Caruso:
This is a simple helloworld example for the Supervision.
I have managed to move screen init code into crt0.s
This has required changing the linker cfg by 6 bytes.
The helloworld example is a unit-test for the init code.

Remark: Do NOT test this on Potator 0.6/0.7. Either use real hardware or at least Wataro or Mame.

It has been tested with the accurate emulator Wataro and on REAL hardware by Plamen @PlamenVaysilov.

@greg-king5, @oliverschmidt could you please take a look at this fix?

### DIFF
--- a/cfg/supervision.cfg
+++ b/cfg/supervision.cfg
@@ -19,7 +19,7 @@ SEGMENTS {
     CODE:     load = ROM,            type = ro,  define   = yes;
     RODATA:   load = ROM,            type = ro,  define   = yes;
     DATA:     load = ROM, run = RAM, type = rw,  define   = yes;
-    FFF0:     load = ROM,            type = ro,  offset   = $7FEA;
+    FFEA:     load = ROM,            type = ro,  offset   = $7FEA;
     VECTOR:   load = ROM,            type = ro,  offset   = $7FFA;
     BSS:      load = RAM,            type = bss, define   = yes;
 }
--- a/cfg/supervision.cfg
+++ b/cfg/supervision.cfg
@@ -19,7 +19,7 @@ SEGMENTS {
     CODE:     load = ROM,            type = ro,  define   = yes;
     RODATA:   load = ROM,            type = ro,  define   = yes;
     DATA:     load = ROM, run = RAM, type = rw,  define   = yes;
-    FFF0:     load = ROM,            type = ro,  offset   = $7FF0;
+    FFF0:     load = ROM,            type = ro,  offset   = $7FEA;
     VECTOR:   load = ROM,            type = ro,  offset   = $7FFA;
     BSS:      load = RAM,            type = bss, define   = yes;
 }
--- a/cfg/supervision.cfg
+++ b/cfg/supervision.cfg
@@ -19,7 +19,7 @@ SEGMENTS {
     CODE:     load = ROM,            type = ro,  define   = yes;
     RODATA:   load = ROM,            type = ro,  define   = yes;
     DATA:     load = ROM, run = RAM, type = rw,  define   = yes;
-    FFEA:     load = ROM,            type = ro,  offset   = $7FEA;
+    FFF0:     load = ROM,            type = ro,  offset   = $7FF0;
     VECTOR:   load = ROM,            type = ro,  offset   = $7FFA;
     BSS:      load = RAM,            type = bss, define   = yes;
 }
--- a/libsrc/supervision/crt0.s
+++ b/libsrc/supervision/crt0.s
@@ -67,7 +67,7 @@ not_dma:
 ; Removing this segment gives only a warning.
         .segment "FFEA"
 .proc reset32kcode
-        lda     #$A0
+        lda     #160
         sta     lcd_width
         sta     lcd_height
         lda     #(6<<5) | SV_LCD_ON | SV_NMI_ENABLE_ON 
--- a/libsrc/supervision/crt0.s
+++ b/libsrc/supervision/crt0.s
@@ -65,11 +65,8 @@ not_dma:
 .endproc
 
 ; Removing this segment gives only a warning.
-        .segment "FFEA"
+        .segment "FFF0"
 .proc reset32kcode
-        lda     #160
-        sta     lcd_width
-        sta     lcd_height
         lda     #(6<<5) | SV_LCD_ON | SV_NMI_ENABLE_ON 
         sta     sv_bank
 ; Now, the 32Kbyte image can reside in the top of 64Kbyte and 128Kbyte ROMs.
--- a/libsrc/supervision/crt0.s
+++ b/libsrc/supervision/crt0.s
@@ -65,12 +65,12 @@ not_dma:
 .endproc
 
 ; Removing this segment gives only a warning.
-        .segment "FFF0"
+        .segment "FFEA"
 .proc reset32kcode
         lda     #$A0
         sta     lcd_width
         sta     lcd_height
-        lda     #$C9
+        lda     #(6<<5) | SV_LCD_ON | SV_NMI_ENABLE_ON 
         sta     sv_bank
 ; Now, the 32Kbyte image can reside in the top of 64Kbyte and 128Kbyte ROMs.
         jmp     reset
--- a/libsrc/supervision/crt0.s
+++ b/libsrc/supervision/crt0.s
@@ -67,7 +67,10 @@ not_dma:
 ; Removing this segment gives only a warning.
         .segment "FFF0"
 .proc reset32kcode
-        lda     #(6<<5)
+        lda     #$A0
+        sta     lcd_width
+        sta     lcd_height
+        lda     #$C9
         sta     sv_bank
 ; Now, the 32Kbyte image can reside in the top of 64Kbyte and 128Kbyte ROMs.
         jmp     reset
--- a/samples/Makefile
+++ b/samples/Makefile
@@ -198,6 +198,9 @@ EXELIST_atarixl = $(EXELIST_atari)
 
 EXELIST_atari2600 = \
         atari2600hello
+        
+EXELIST_supervision = \
+        supervisionhello
 
 # Unlisted targets will try to build everything.
 # That lets us learn what they cannot build, and what settings
--- a/samples/supervisionhello.c
+++ b/samples/supervisionhello.c
@@ -1,17 +1,15 @@
 #include <supervision.h>
 #include <peekpoke.h>
 
-#define SV_OFFSET 0x0A
-#define SV_INIT_POSITION (SV_VIDEO+0x0A*48+0x0A)
+#define BYTES_PER_LINE 48
 
-unsigned char h_char[] = {0x66,0x66,0x66,0x7e,0x66,0x66,0x66,0x00};
-unsigned char e_char[] = {0x7e,0x60,0x60,0x78,0x60,0x60,0x7e,0x00};
-unsigned char l_char[] = {0x60,0x60,0x60,0x60,0x60,0x60,0x7e,0x00};
-unsigned char o_char[] = {0x3c,0x66,0x66,0x66,0x66,0x66,0x3c,0x00};
-unsigned char w_char[] = {0x63,0x63,0x63,0x6b,0x7f,0x77,0x63,0x00};
-unsigned char r_char[] = {0x7c,0x66,0x66,0x7c,0x78,0x6c,0x66,0x00};
-unsigned char d_char[] = {0x78,0x6c,0x66,0x66,0x66,0x6c,0x78,0x00};
-
+const unsigned char h_char[] = {0x66,0x66,0x66,0x7E,0x66,0x66,0x66,0x00};
+const unsigned char e_char[] = {0x7E,0x60,0x60,0x78,0x60,0x60,0x7E,0x00};
+const unsigned char l_char[] = {0x60,0x60,0x60,0x60,0x60,0x60,0x7E,0x00};
+const unsigned char o_char[] = {0x3C,0x66,0x66,0x66,0x66,0x66,0x3C,0x00};
+const unsigned char w_char[] = {0x63,0x63,0x63,0x6B,0x7F,0x77,0x63,0x00};
+const unsigned char r_char[] = {0x7C,0x66,0x66,0x7C,0x78,0x6C,0x66,0x00};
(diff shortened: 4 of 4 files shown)


Fabrizio-Caruso:
I have now managed to move screen init code into crt0.s
This has required changing the linker cfg by 6 bytes.

oliverschmidt:
Please don't combine fixes for the C library (of any target) with the contribution of a new program in a single pull request.

libsrc/supervision/crt0.s
```diff
@@ -67,7 +67,10 @@ not_dma:
 ; Removing this segment gives only a warning.
         .segment "FFF0"
 .proc reset32kcode
-        lda     #(6<<5)
+        lda     #$A0
+        sta     lcd_width
+        sta     lcd_height
+        lda     #$C9
```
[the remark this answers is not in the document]
oliverschmidt: Why `%11000000` instead of the `(6<<5)` of the original code?

cfg/supervision.cfg
```diff
@@ -19,7 +19,7 @@ SEGMENTS {
     CODE:     load = ROM,            type = ro,  define   = yes;
     RODATA:   load = ROM,            type = ro,  define   = yes;
     DATA:     load = ROM, run = RAM, type = rw,  define   = yes;
-    FFF0:     load = ROM,            type = ro,  offset   = $7FF0;
+    FFF0:     load = ROM,            type = ro,  offset   = $7FEA;
```
oliverschmidt: Sure that the segment should still be named `FFF0` if its offset is now `$7FEA`? Is it really intended to have the same offset twice (see segment `VECTOR`)?

libsrc/supervision/crt0.s
```diff
@@ -67,7 +67,10 @@ not_dma:
 ; Removing this segment gives only a warning.
         .segment "FFF0"
 .proc reset32kcode
-        lda     #(6<<5)
+        lda     #$A0
```
oliverschmidt: I don't see why initializing the screen should happen in a function called reset32kcode. This somehow looks like a hack rather than a fix.

In general I don't see why this code should be placed in crt0.s. This target doesn't have any CONIO support so why should it initialize the screen in the first place. Without CONIO support it's _all_ up to the user. And if there's CONIO support at some point then the screen initialization belongs to a CONIO constructor.

Fabrizio-Caruso: Ok, but I am not sure about your point: what is the of not having the screen on and the resolution set up to some useful default value? This is not a conio init. This is just a very generic initialization that has nothing to do with conio, cursors, text, etc.
Let us at least have one helloworld example for this under-documented target 
I can move the screen init part into the hello world example and I improve the example if necessary.

libsrc/supervision/crt0.s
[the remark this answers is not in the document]
Fabrizio-Caruso: Why (6<<5) does not work as verified on Wataroo, Mame and REAL hardware by Plamen @PlamenVaysilov. The fix has been equally verified on real real hardware by Plamen.

Remark: Potator 0.6/0.7 does not show the problem.

cfg/supervision.cfg
Fabrizio-Caruso: You are perfectly right on the name.

oliverschmidt:
@Fabrizio-Caruso:

First of all let me clarify that I personally appreciate your approach to push the "cross target" idea to the limit.

Additionally I's like to mention that I appreciate that you (try to) use aspects of the cc65 C libraries not used before as this discovers quite some issues unknown before.

And finally I'm happy to see that you're into "giving back" by creating pull requests :-)

Okay, now to the Supervision...

The target certainly deserves way better support by cc65. The current support was added back in 2003 (https://github.com/cc65/cc65/commit/6afee6fb0578518daf7bde5b31f04d5ed2b1fc73) and never really improved since then.

Given that I think it's a good thing to have a C program available that shows how to get anything out of the Supervision. However, I need to ask you to:

* Adjust the source code style to the predominant style as described in https://github.com/cc65/wiki/wiki/C-and-Assembly-sources-style-guide

* Add the necessary support to the Makefile

* Optionally add a section to the README

Fabrizio-Caruso:
> 
> 
@oliverschmidt 
> Please don't combine fixes for the C library (of any target) with the contribution of a new program in a single pull request.

The new program is a unit-test for the fix. If you see nothing on the screen the fix is wrong. If you see "hello world" the fix is working. Usually unit-tests are coding together with the code they are supposed to test. I can separate the unit-test but you will have to write something to test the fix anyway.

libsrc/supervision/crt0.s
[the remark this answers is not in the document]
oliverschmidt: Okay, trying to be more explicit:

* The original code is `#(6<<5)`

* You suggest `#$C9`

* @polluks suggests `#%11000000 | SV_LCD_ON | SV_NMI_ENABLE_ON` instead

* I ask why not `#(6<<5) | SV_LCD_ON | SV_NMI_ENABLE_ON`

oliverschmidt:
> The new program is a unit-test for the fix.

Unit tests don't go into the samples directory.

libsrc/supervision/crt0.s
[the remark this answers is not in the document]
Fabrizio-Caruso: Sorry, #$C9 was suggested by me by @PlamenVaysilov who is a real expert and who has a multi-cart for the Supervision and has tested this on real hardware positively as well tested the bug.
I can try #%11000000 but I would like to have it tested on real hardware if Plamen is available. Let me see...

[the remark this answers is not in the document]
oliverschmidt: Okay, trying to be _EVEN_ more explicit:

`#(6<<5)` equals `#$C0`. You have benn told that `#$C9` is better than `#$C0`. Nobody disagrees so far.

But instead of writing `#$C9` you can as well write `#%11000000 | SV_LCD_ON | SV_NMI_ENABLE_ON`. It's the same value, it's just easier to understand (no "magic constant").

And instead of `#%11000000 | SV_LCD_ON | SV_NMI_ENABLE_ON` you can as well write `#(6<<5) | SV_LCD_ON | SV_NMI_ENABLE_ON`. It's still the same value, but maybe even more easy to understand.

Fabrizio-Caruso:
@oliverschmidt, correct. It is not a unit-test nor an automatic test. It is a visual test for a fix. I thought it was necessary and part of the fix. If you want I separate it from the PR.

I will try to comply with the code guidelines. 

I don't understand what I should do with Makefile. Should the hello world example be compiled?

P.S.: I have already "given back" something with other (rather minimal) pull-requests that fixed joystick issues on the Supervision and Gamate (both had inverted bit masks). This was one or two years ago. 
I have minimal knowledge of Assembly but I am a half-decent C coder so I am more of a tester than a code contributor. I consider testing CC65 and opening issues a significant contribution.

libsrc/supervision/crt0.s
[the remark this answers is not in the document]
Fabrizio-Caruso: I agree. Sorry for not understanding you... I am a bit tired.

oliverschmidt:
> I don't understand what I should do with Makefile. Should the hello world example be compiled?

Yes. The name you choose seems to suggest that you saw atari2600hello.c. If you consider atari2600hello.c as blueprint do the same regarding treatment in the Makefile.

> P.S.: I have already "given back" something with other (rather minimal) pull-requests that fixed joystick issues on the Supervision and Gamate (both had inverted bit masks). This was one or two years ago.
I have minimal knowledge of Assembly but I am a half-decent C coder so I am more of a tester than a code contributor. I consider testing CC65 and opening issues a significant contribution.

Hm, I just tried to behave friendly towards you. I didn't expect my statements to have the potential to be misunderstood. However, that can be easily fixed by not showing any such behavior from my side in the future ;-)

polluks:
This is a hobby project, nevertheless Oliver is a strict maintainer (needed for high quality).
Fixing a critical 16 years old bug because of a buggy emulator reference is also well done.

Fabrizio-Caruso:
@oliverschmidt  no problems on my side.
I will follow your sugestions.

libsrc/supervision/crt0.s
Fabrizio-Caruso: @oliverschmidt So, is it ok to have the screen init code in crt0.s or should I just mobe it to the helloworld example in a separate PR? I do as you prefer.

greg-king5: The bank setting is needed, in "crt0.s", in order to make cartridges work in a predictable way.  I think that the other parts of that register may as well be set at the same time.  But, the screen size settings should be with code that actually uses the screen.  At this time, it's the "Hello world" program.  (Later on, we might have an implementation of conio.  Then, it would set those sizes.)

greg-king5:
Tests for library code should be put into "testcode/lib/".

(But, as Oliver mentioned, your "hello world" example is similar to "atari2600hello.c".  So, keep it in "samples/".)

libsrc/supervision/crt0.s
oliverschmidt: @greg-king5: That's exactly my perspective too. Thanks for elaborating.

Fabrizio-Caruso: Ok! So I will revert my changes about lcd size in the cfg and crt0.s.

Fabrizio-Caruso: @oliverschmidt  should I cancel this PR and make two PRs? One for crt and one for the example?
Should I keep this one for the crt?

Fabrizio-Caruso: @oliverschmidt @greg-king5  the reason why I put the 160x160 setting is that this tells the controller it is on a 160x160 which is the Watara Supervision one. 160x160 is going to be always used for the Watara Supervision target. The video control can be told to use different screens but this is never the case on this handheld. I cannot think of a real use case with different values. Not having this setting makes the crt a generic non-Supervision one. Is this what we really want?

Fabrizio-Caruso: Supervision display is always 160x160 and the crt0.s for the Supervision can assume it is running on a Supervision which only has a fixed size screen. No other values will ever be used.
What is the use-case you guys are thinking of?

Fabrizio-Caruso:
I can put supervisionhello in samples/Makefile but how do I build it?
It seems atari2600hello cannot be built either:
$ make atari2600hello
../bin/cc65  -Ors --codesize 500 -T -g -t c64 atari2600hello.c
/home/fabri/CC65_FIX/cc65/include/atari2600.h(20): Error: #error: This module may only be used when compiling for the Atari 2600!
make: *** [Makefile:138: atari2600hello.o] Error 1

libsrc/supervision/crt0.s
polluks: I agree, Fabri. By the way isn't it natural to use `#160` if you say 160 pixel?

Fabrizio-Caruso: @polluks $A is 10 (just after 9) and second hex position is multiples of 16... So $A0 = 160 is quite obvious to me.

Fabrizio-Caruso: @oliverschmidt do you really want me to remove the lcd size setting from the crt even though this is the only meaningful value for the Supervision taget in all (reasonable) use-cases?
Other lcd size values make sense for other screen in other consoles with the same video controller.
I am just asking you to be sure. I will not insist on this.

greg-king5: > @polluks, $A is 10 (just after 9), and the second hex position is multiples of 16... So, $A0 = 160 is quite obvious to me.

It isn't obvious to me -- I need to do the arithmetic in my head every time that I see it.

Fabrizio-Caruso:
@oliverschmidt Could you please give me your final decision on following points?
1. Should I cancel this PR? Should I use new two separate PR? Should I instead keep this PR and do an extra one just for the hello world example?
2. Should I really remove the lcd size init? even though it is the only possible one and does not depend on any api (conio, tgi, etc.) or code. Any code for the Supervision will have those values. Other values are only usable for other targets with the same video controller.
3. Is the helloworld example OK for you now?

Current situation is.
- Screen init is in crt0.s and default cfg for Supervision has been changed to make room for the extra initialization in crt0.s. I have used the suggested format for the setting of SV_BANK.

- Hello world for Supervision
I have improved the hello world example supervisionhello.c , which in in samples and can be built similarly to atari2600hello.c by running `make SYS=supervision supervisionhello` inside `samples` directory.
I have tried to follow the guidelines for the C code in CC65 (no tabs, no small letters in hex, etc.).

libsrc/supervision/crt0.s
Fabrizio-Caruso: @greg-king5 16x10 is just adding a zero but OK, I am going to make it even clearer.

Fabrizio-Caruso: @greg-king5 I have change it to #160 but the real problem is whether it should be done or not... I am doing it in crt for the reasons that I have given. LCD on Supervision is fixed to 160x160 and other values have no (real) use on the Supervision no matter what you use (conio, low level, tgi, stdio, etc.).
I will remove it from there if @oliverschmidt decides so.

oliverschmidt:
> [...] I will just apply your corrections and follow your decisions.

Thanks :-)

Please revert your changes in crt0.s (and the linker config) so that the only change is the addtion of ` | SV_LCD_ON | SV_NMI_ENABLE_ON` and put the LCD initialization back into main(). Then I'm willing to merge this PR:

oliverschmidt:
> I have improved the hello world example supervisionhello.c , which in in samples and can be built similarly to atari2600hello.c by running make SYS=supervision supervisionhello inside samples directory.

It is supposed to work by running just `make SYS=supervision`.

Fabrizio-Caruso:
> 
> It is supposed to work by running just `make SYS=supervision`.
@oliverschmidt 
It does! :-)
Please let me know if you can tolerate it in the same PR or I should remove these changes in this PR.

Fabrizio-Caruso:
@oliverschmidt I have removed the lcd init from crt and moved it into the helloworld. Could you review it and merge it if OK?

oliverschmidt:
Thanks :-)

PR #951, opened 2019-10-08
Significantly faster rand() implementation

LRFLEW:
I was experimenting with PRNG's and I found an interesting optimization for the version used for the stdlib.h function `rand()`. It's likely slightly weaker in terms of entropy, but it is still an [LCG](https://en.wikipedia.org/wiki/Linear_congruential_generator) with a period of 2<sup>32</sup>. To make this work, I had to make two changes to the LCG.

Firstly, the order of operations is changed. Instead of calculating `ax + c (mod m)` each step, it calculates `a(x + c) (mod m)` instead. This is still an LCG despite this change, as it's just pulling values at a different point in the LCG generation sequence. If we consider an LCG as being the sequence `x_1 = a*x_0, x_2 = x_1 + c, x_3 = a*x_2, x_4 = x_3 + c, ... (mod m)`, then this is just sampling the odd `x`'s instead of the even ones. Since `x (mod m) → x + c (mod m)` is a bijection, there is no loss in entropy making this change.

Secondly, I changed the `c` value for the LCG. Specifically, to make this optimization work, I had to get the `c` value to be only 8 bits. This is an easy change to make, as the only requirement to maintain the maximum period is that `c` must be an odd number. Making this change may result in a lower apparent entropy, but from a visual check it didn't appear to be noticeably worse (and LCG's aren't known for great spectral properties anyways).

Now for the optimization. First, the algorithm performs the addition `(rand) + c` on the least significant byte `rand+0`. Since `c` is only 8 bits, the other bytes of the state don't need any value specifically added to them. However, the addition sets the carry flag `C` that needs to be accounted for in the other bytes. For the next significant byte `rand+1`, we can describe this requirement as `(rand+1) + $00 + C → (rand+1)`. However, the multiplication `rand*a` also requires an addition in the form of `(rand+0) + (rand+1) + 0 → (rand+1)` (i.e. with the `C` flag cleared). With the 0's in both additions, we can combine them into a single addition `(rand+0) + (rand+1) + C → (rand+1)`. The assembly for this addition looks like this: 
```
; (rand+0) is in A due to the last addition
adc rand+1
sta rand+1
```
Now `C` has a carry value from the combination of both the addition and multiplication steps. This means that continuing the multiplication on to the third and most significant bytes is the same as just the multiplication alone.

The key to this optimization is that now we only need to read, modify, and write each byte of the state once in a single pass, opposed to the two separate passes in the existing version. I calculated the clock cycles required for both versions (excluding the `rts` instruction, as I'm unclear on how to time it based on the WDC 65C02S documentation). Assuming that the `rand` value is not in the zero page, I got that this version takes 44 clock cycles, while the original version takes 76 clock cycles (1.73x improvement). Making this change produced a very noticeable speed improvement in my `rand()` heavy test program (producing white noise bitmaps), with it running 1.42x faster.

### DIFF
--- a/libsrc/common/rand.s
+++ b/libsrc/common/rand.s
@@ -35,27 +35,17 @@ rand:   .dword   1
 .code
 
 _rand:  clc
-        lda     rand+0          ; SEED *= $01010101
-        adc     rand+1
+        lda     rand+0          ; SEED += $B3
+        adc     #$B3
+        sta     rand+0
+        adc     rand+1          ; SEED *= $01010101
         sta     rand+1
         adc     rand+2
         sta     rand+2
-        adc     rand+3
-        sta     rand+3
-        clc
-        lda     rand+0          ; SEED += $31415927
-        adc     #$27
-        sta     rand+0
-        lda     rand+1
-        adc     #$59
-        sta     rand+1
-        lda     rand+2
-        adc     #$41
-        sta     rand+2
         and     #$7f            ; Suppress sign bit (make it positive)
         tax
-        lda     rand+3
-        adc     #$31
+        lda     rand+2
+        adc     rand+3
         sta     rand+3
         rts                     ; return bit (16-22,24-31) in (X,A)
 
--- a/libsrc/common/rand.s
+++ b/libsrc/common/rand.s
@@ -3,6 +3,7 @@
 ;
 ; Written and donated by Sidney Cadot - sidney@ch.twi.tudelft.nl
 ; 2016-11-07, modified by Brad Smith
+; 2019-10-07, modified by Lewis "LRFLEW" Fox
 ;
 ; May be distributed with the cc65 runtime using the same license.
 ;
@@ -23,6 +24,17 @@
 ;  low byte A to provide the best entropy in the
 ;  most commonly used part of the return value.
 ;
+;  Uses the following LCG values for ax + c (mod m)
+;  a = $01010101
+;  c = $B3B3B3B3
+;  m = $100000000 (32-bit truncation)
+;
+;  The multiplier was carefully chosen such that it can
+;  be computed with 3 adc instructions, and the increment
+;  was chosen to have the same value in each byte to allow
+;  the addition to be performed in conjunction with the
+;  multiplication, adding only 1 additional adc instruction.
+;
 
         .export         _rand, _srand
 
@@ -35,10 +47,10 @@ rand:   .dword   1
 .code
 
 _rand:  clc
-        lda     rand+0          ; SEED += $B3
+        lda     rand+0
         adc     #$B3
         sta     rand+0
-        adc     rand+1          ; SEED *= $01010101
+        adc     rand+1
         sta     rand+1
         adc     rand+2
         sta     rand+2


oliverschmidt:
Thanks for the proposal - and the very substantial explanation :-)

Unfortunately my mathematical background is so weak that I don't see myself able to assess it. So I'm hoping for feedback from other members of the cc65 community.

Merovius:
Drive-by contribution (friend pointed me to this thread): To me, this seems to check out.

AIUI the order of operations indeed doesn't change the property of being an LCG: `a(x+c) = ax+ac`, so you get an LCG with the same multiplier but a different offset. And the offset is changed anyway. So semantically this just changes the constant from `0x31415927` to `0xb3b3b3b3` (based on the multiplier `0x01010101`). It does look like a "less random" offset, but AIUI the offset doesn't matter for the quality of the PRNG. And the optimization also seems to check out. It relies on the byte-pattern of `a` as just 1s, but it did before as well. So, it's probably not worth a lot, as none of you know me :) But mathematically, this looks fine (and maybe my explanation helps).

I assume, though, that the previous constant was chosen to look random like that for a reason - from a theoretical POV `1` would serve just as well and yet the code used to add several instructions worth of overhead for a more random looking constant. As mentioned though, this only changes how the bit-patterns look, not how high the entropy is or anything, so it might not matter?

LRFLEW:
I think @Merovius had the better explanation for this change to the LCG. This is an LCG with a=0x01010101, c=0xB3B3B3B3, and m=2<sup>32</sup>. Wikipedia has [great rules for making a maximum period LCG](https://en.wikipedia.org/wiki/Linear_congruential_generator#c_%E2%89%A0_0), which this satisfies. (The rules listed in the ASM comments can be derived from these rules by assuming c≠0 and m=2<sup>n</sup>).

The only thing I worried about was whether the carry flag was actually doing what is thought it was. Despite proving to myself that it was correct, I did keep worrying about it, so I wrote a brute force test for it. You can see my code here: https://gist.github.com/LRFLEW/4ace787ecb3190f8896bf218adde3762. It calculates the LCG value directly, then evaluates the steps that the 6502 assembly would take for that call (ignoring the bit manipulations performed on the returned value) and compares the result. Running this test for 1 billion repetitions didn't return any errors, so I'm confident that my proof of accuracy is correct.

LRFLEW:
I noticed the attributions at the top of the assembly file, and decided to add my name to it. Let me know if I shouldn't have done that or need to do it differently.

I added a block comment after the comments at the top of the file to explain the change better. I put the LCG values in that comment and removed them from the inline comments, as its more challenging to accurately label them now. I feel better about how the change is documented now.

oliverschmidt:
> I feel better about how the change is documented now.

Me too :-)

> Let me know if I shouldn't have done that or need to do it differently.

Everything is just fine.

I'm going to merge this PR now as there was no dissent by now.

bbbradsmith:
I missed this when it happened, but just wanna confirm that it looks good by my tests. Did verify that the seed has a complete cycle through all values. Ran that test program for 2^32 rather than just a billion values, checked for even distribution of values. Simple spectral test looks normal. I couldn't see any way that I think this is deficient compared to the previous LCG.

Also just to check the implementation I wrote a cc65 C program to verify against what it's supposed to be with sim65:
```
#include <stdio.h>
#include <stdlib.h>
#include <stdint.h>

#define COUNT 10000000

uint32_t i = 0;
uint32_t s = 1;
int16_t r0, r1;

int main()
{
	srand(s);
	for (i=COUNT; i; --i)
	{
		s = s * 0x01010101 + 0xB3B3B3B3;
		r1 = (((s >> 8) & 0x7F00) | (s >> 24));
		r0 = rand();
		if (r0 != r1)
		{
			printf("rand()=%d != %d\n",r0,r1);
			return -1;
		}
	}
	printf("success! %lu cases checked.\n",(i=COUNT));
	return 0;
}
```
Run with:
```
cl65 -O --target sim6502 -o test.bin test.c
sim65 test.bin
```

The first 10 million cases run in about a minute, I guess I could run it for half a day and cover the entire period, but I just wanna say this looks good to me.

oliverschmidt:
Thanks for your feedback :-)

bbbradsmith:
Someone I was talking to came across this minor flaw, a simple 8bit vs. 8bit pair plot test of 100,000 points shows a diagonal line, both in the original LCG and the new one:

Original:
cc65rand_pi

New:
cc65rand_b3

In other words, looking at successive pairs of values, the new LCG **will never hit the same value twice in a row**. The old one suffers from the same problem, but not quite as badly. It seems it will hit a repeat only about 10% as often as it should... i.e. the problem is about 90% as bad as with the new one.

This only affects back to back values. Once you add at least one extra rand() in between the distribution becomes flat again.

Anyway, that's one way that it seems the PRNG is slightly weakened by the change, though I'm not sure whether I think it's significant. If it's a real problem, then the previous implementation still had almost as much of this problem.

How to correct it, I'm not sure, as this is probably inherent in using such a "convenient" multiplier. I notice for the new version the displaced pairs seem to transfer 1 case each of X->X to X->(X+$B3). Spending some time working out why X->X is impossible might yield some low-impact ideas for fixing it.

bbbradsmith:
Tentatively, adding this just before the RTS seems to fix this particular pattern (at the expense of 3 bytes / 4 cycles). Need to spend a bit more time testing it though.
```
eor seed+1
```

jmr:
Four adds does seem a bit too good to be true. Has anyone compared these parameters to some known good/recommended ones with PractRand/TestU01/etc?

bbbradsmith:
I don't think PractRand and TestU01 are really suitable for this kind of RNG. I have tried them in the past, bit they just aren't the right thing to use here. They are designed for evaluating much more powerful/secure kinds of RNG, and the kind of results they give aren't really very applicable. (I don't think they even accept 16-bit random numbers as input, for starters.)

In a case like this, I think it is better to come up with a checklist of desirable and undesirable properties and make sure they're met.

1. The 32-bit seed state has the maximal sequence of 2^32 unique states.
2. Every possible output number is delivered in equal distribution across the full sequence.
3. Across shorter portions of the sequence, the numbers cover the output space evenly but nicely chaotically.
4. Successive pairs from the sequence also have even distribution (i.e. any number can follow any number)
5. Condition 4 should apply for numbers any distance apart, but with decreasing importance.
6. Conditions above can be evaluated on groups of bits (e.g. low 8 bits, low 4 bits, top 7 bits, etc.) as well.

RNG test suites tend to have a bunch of other goals, mostly to do with cryptographic security, and I don't think that's important at all here. If someone actually needs that, they should never expect a C library rand() to be appropriate.

The last few days doing various tests on the conditions mentioned, there seems to be a minor weakness in condition 4 (and 6 in a related way), though as I mentioned in my post above, the same weakness existed to a lesser extent in the previous version.

A proposed remedy:
```
eor seed+0 ; inserted before: and #$75
eor seed+1 ; inserted before: rts
```

The first addition improves the quality of the top 7 bits, and the second addition improves the quality of the lower 8 bits. This claws back the optimization by 8 cycles. With this additional XOR, the distribution of pairs was improved, and none of the other properties I was evaluating became worse. The underlying seed sequence is unchanged.

There is still a suspicion that 0x01010101 is a poor multiplier for an LCG, but I can't yet find a test I think is worthwhile that can demonstrate a relevant weakness, so I can't really offer a reason not to do it. It seems to be working well enough to me.

mrdudz:
One thing that is also very important: any argument to srand() will result in a properly working PRNG (some of them will get stuck at certain seeds and produce really undesireable output sequences)

bbbradsmith:
That is not a problem here. The maximal length sequence property (1) guarantees it doesn't get stuck, and that it has no disconnected subsequences like that.

LRFLEW:
I'm largely of the opinion that `rand()` doesn't need to be particularly good because anybody who needs high-quality PRNG will usually roll their own. For an implementation like this where not everybody is going to use it, and those that do will have wildly different performance requirements, it's going to be a trade-off between quality, speed, and memory space. I remember someone on the X16 forum had a much stronger PRNG that was higher quality with the trade-off of more space required and more complexity. Looks like the forum has been archived, and some of the thread didn't get archived, but you can see what's left here: https://web.archive.org/web/20200530202810/http://temp.murray2.com/threads/how-to-get-a-random-number-in-asm-on-the-commander-x16.534/page-2

I think one easy, quick improvement that could still be made is to improve `srand()`. Right now, it zero-extends the seed value, which causes some amount of regularity in the initial output. Perhaps making the state `seed | (seed << 16)` would help things somewhat. It would even be one instruction less:

```
_srand: sta     rand+0          ; Store the seed
        stx     rand+1
        sta     rand+2          ; Put the seed in the upper word as well
        stx     rand+3
        rts
```

bbbradsmith:
I think rand() needs to perform reasonably well, and generate reasonably good output. It doesn't need crypto level security, but it should at least have even distribution as much as it is practical. Obviously there's no one-size fits all... and I think the current rand() is better than several I've seen in compilers of days past.

The main question I wanna ask is if it's worth trading a few cycles to address the issue I pointed out (i.e. this RNG won't return the same value twice in a row), and in my opinion it is, so now that I've tested it to my own satisfaction I created #1107

If one wants more speed, there's a bunch of easy steps that a user that needs that could take by copying this routine and tweaking it in assembly, e.g. moving seed to ZP, or reducing output to just 8-bits (the fact that rand() is 15-bit by definition is kind of impractical just by itself)... and if want to trade quality you can omit the EORs, or you can cut down the seed state to 24 bits, etc. there's lots of options, but within the context of rand() you can call from C code I think it's already kinda fabulously lightweight for what it provides.

.

I really like that this compiler seems to have a novel approach to creating a PRNG. Its initial form that it had for many years actually had very poor output, but the novel idea of using 0x01010101 as a multiplier makes it interesting. Even though it looks scary for an LCG multiplier, the test results seems surprisingly good despite this, once we started using the high bits properly.

There are some other existing 6502 PRNG implementations that could be imported, but I think it's great to have a unique one like this, and I know it meets my own (very subjective) quality goal at this point. Several months ago I did a survey of many 6502 RNG implementations I found lying around, and was disappointed by how poor the quality of the majority of them were, and how almost none of them were documented any better than "here's a random number generator". There's certainly good quality ones out there, but finding some chain of trust to put faith in their quality can be difficult. 

.

The other suggestion, for this particular RNG I don't think the high bytes of the initial seed being 0 don't really negatively affect anything? The first call of rand() propagates the low bytes to the high bytes (with obfuscation) already. Removing the lda #0 saves 2 bytes of code at least, but I don't think it affects the quality of the output, so... fine to do?

oliverschmidt:
@bbbradsmith: Thanks for your comprehensive input.

@LRFLEW:
* I'd like to get your opinion on https://github.com/cc65/cc65/pull/1107.

* In https://github.com/cc65/cc65/pull/951#issuecomment-660713967 you propose an optimized _srand. Please provide a PR.

Issue #948, opened 2019-10-05
[creativision][bug?] Lack of COLOR_BLACK and COLOR_WHITE (samples/hello.c won't compile)


Fabrizio-Caruso:
If I try to compile samples/hello.c for the Creativition target, I get an error about missing COLOR_BLACK and COLOR_WHITE definitions.

polluks:
Well, no conio, therefore no colours.

oliverschmidt:
Hm, looking at https://github.com/cc65/cc65/tree/master/libsrc/creativision there seems to be CONIO stuff...

greg-king5:
The macro names are prefixed with `C_` instead of `COLOR_`.

Fabrizio-Caruso:
@pollucks I am using conio with the Creativision

PR #952, opened 2019-10-09
Fix colors for #948

polluks:


Fabrizio-Caruso:
@polluks  Current conio implementation uses mode1 (32x24 text mode) which makes text monochromatic in practice.
I think it is good to have definitions for all colors but I am not sure this is consistent with the other conio-enabled targets.
A better conio implementation may use mode 2 (sort of bitmapped mode) where tiles/characters can have any color.

oliverschmidt:
Thanks for the contribution and the merging :-)

Issue #946, opened 2019-10-04
[VIC-20] CRAM_PTR incorrect at startup


0cjs:
As described in detail in [this Retrocomputing StackExchange answer][rc], at the start of `main()` on the VIC-20, when loading/running the PRG with VICE's `-autostartprgmode 1` option, `CRAM_PTR` is incorrectly set and any printing using `cputc()` will not be visible (because the colour data is being set in the wrong place) until something is done that would reset the cursor position, such as clearing the screen, doing a `gotoxy()` or printing a newline. My [vic20cc65] repo contains code and scripts to easily demonstrate this.

(I've reproduced this using VICE 3.0.0 on Debian and ROM images from a source I can't remember now; as you can see from the post another user has reproduced this too, most likely with a different system.)

I'm not sure exactly what the true source of the problem is. If the bug is in VICE it should obviously be fixed there, but that seems less likely to me than a bug in the VIC ROM or a misunderstanding on the part of vic20.lib on how `CRAM_PTR` is supposed to work. (That said, I'm no expert on this, so I could well be wrong.)

I've looked through `libsrc/vic20/crt0.s` and I don't see anything obviously wrong there, unless it's actually supposed to be checking or resetting `CRAM_PTR`; right now it doesn't appear to touch it at all, so when `main()` is run it should be exactly what it was set to by the VIC ROMs.

So, does this need a fix in vic20.lib? If so, what should the nature of the fix be?

I'd be very pleased to code up the fix and submit a PR if the fix should be done here and someone can give me a quick summary of what approach to take. (I need the coding practice and, hey, it's Hacktoberfest. :-))

[rc]: https://retrocomputing.stackexchange.com/a/12494/7208
[vic20cc65]: https://github.com/0cjs/vic20cc65

oliverschmidt:
First of all thanks for the detailed problem description :-)

For now I'll classify this issue as question. I hope the Subject Matter Experts will pick it up and there will be a conclusion on how to proceed...

0cjs:
Ah! Looking through this further, I see we have a `libsrc/vic20/kplot.s` that replaces the Kernal PLOT function at $FFF0, because "[t]he kernal function does not set the pointer to the color RAM correctly."
(In `asminc/cbm_kernal.inc`, however, `PLOT` is still set to `$FFF0` if `__VIC20__` is defined. I'm not sure exactly how the replacement overrides this, if indeed it does.)

My guess at this point about what's going on is:
[...]

PR #965, opened 2019-10-24
vic20/cputc: Fix incorrect CRAM_PTR at startup when using conio

0cjs:
This fixes the problem discussed in issue #946 and is an improved solution over that in PR #953.

The first commit is a refactoring to replace hardcoded addresses in VIC-20, C64 and Plus/4 `kplot.s` with symbols from `cbm_kernal.inc`. The second commit fixes the VIC-20 bug by adding a constructor call to `UPDCRAMPTR` so that `CRAM_PTR` is set correctly when the program starts. (This adds two additional bytes to programs using `cputc()` or other routines that call it. These are in theory recoverable, but the VIC-20 runtime does not yet free space used by constructors after the constructors have been called.)

These changes have been tested on the VICE emulators for all three platforms. You can replicate this using my framework in [`0cjs/vic20cc65`](https://github.com/0cjs/vic20cc65.git); just clone that repo and run `./Test`, editing the variables in that file to choose the platform and test to run. (`README.md` there has more information, including how to demonstrate the original problem.)

Thanks to @greg-king5 who in [this comment](https://github.com/cc65/cc65/issues/946#issuecomment-538502820) dug into the details of the VIC-20 ROM to clarify the problem and proposed the solution used here. (And also for the comments on the previous PR that led to the refactoring done before this fix.)

If you wish to  examine the ROM code yourself, you may find it convenient to refer to copies of the original ROM source and various disassemblies in [`0cjs/vic20-stuff/rom`](https://github.com/0cjs/vic20-stuff/tree/master/rom).

### DIFF
--- a/asminc/cbm_kernal.inc
+++ b/asminc/cbm_kernal.inc
@@ -100,14 +100,26 @@ UDTIM          := $FFEA
 
 ; ---------------------------------------------------------------------------
 ; Kernal routines, direct entries
+;
+; Unlike the above, these are not standard functions with entries in the jump
+; table. They do not exist in all Kernals, and where they do the entry point is
+; specific to that particular machine and possibly even Kernal version.
+;
+; This list is not comprehensive: missing items for particular machines
+; should be added as needed.
+;
+; UPDCRAMPTR: Updates the color RAM pointer to match the screen RAM pointer.
+;
 
 .if .def(__VIC20__)
   CLRSCR       := $E55F
   KBDREAD      := $E5CF
+  UPDCRAMPTR   := $EAB2
 .elseif .def(__C64__)
   CLRSCR       := $E544
   KBDREAD      := $E5B4
   NMIEXIT      := $FEBC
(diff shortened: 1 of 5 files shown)


oliverschmidt:
Thanks for contributing to cc65 - and the comprehensive description.

asminc/cbm_kernal.inc
```diff
@@ -100,14 +100,26 @@ UDTIM          := $FFEA
 
 ; ---------------------------------------------------------------------------
 ; Kernal routines, direct entries
+;
+; Unlike the above, these are not standard functions with entries in the jump
+; table. They do not exist in all Kernals, and where they do the entry point is
+; specific to that particular machine and possibly even Kernal version.
+;
+; This is list is not comprehensive: missing items for particular machines
```
greg-king5: The first "is" is superfluous.

0cjs: Thanks. I've fixed this and force-pushed the new commits.

PR #966, opened 2019-10-28
Second tentative fix

Fabrizio-Caruso:
@greg-king5 could you please give your advice on this?
Forgive me for the poor code. I am not a good Assembly coder.

### DIFF
--- a/libsrc/lynx/cgetc.s
+++ b/libsrc/lynx/cgetc.s
@@ -19,14 +19,10 @@
 ; and Opt1 + Opt2 pressed '3'.
 ; So the keyboard returns '1', '2', '3', 'P', 'R', 'F' or '?'.
 
-_cgetc:
-        lda     KBEDG
-        bne     _start     
+_cgetc:   
         jsr     _kbhit          ; Check for char available
         beq     _cgetc
-_start:
-        lda     KBSTL
-        ora     KBEDG
+        ora     KBSTL
         ldx     #0
         and     #1
         beq     @L5
@@ -55,7 +51,6 @@ _start:
         bra     reset_and_exit
 @L5:
         lda     KBEDG           ; No Pause pressed
-        stz     KBEDG
         ora     KBSTL
         bit     #$08
         beq     @L7
--- a/libsrc/lynx/cgetc.s
+++ b/libsrc/lynx/cgetc.s
@@ -20,8 +20,10 @@
 ; So the keyboard returns '1', '2', '3', 'P', 'R', 'F' or '?'.
 
 _cgetc:
+        bne     _start     
         jsr     _kbhit          ; Check for char available
         beq     _cgetc
+_start:
         lda     KBSTL
         ora     KBEDG
         ldx     #0
@@ -52,6 +54,8 @@ _cgetc:
         rts
 @L5:
         lda     KBEDG           ; No Pause pressed
+        ldx     #$00
+        stx     KBEDG
         ora     KBSTL
         bit     #$08
         beq     @L7
--- a/libsrc/lynx/cgetc.s
+++ b/libsrc/lynx/cgetc.s
@@ -54,8 +54,7 @@ _start:
         rts
 @L5:
         lda     KBEDG           ; No Pause pressed
-        ldx     #$00
-        stx     KBEDG
+        stz     KBEDG
         ora     KBSTL
         bit     #$08
         beq     @L7
--- a/libsrc/lynx/cgetc.s
+++ b/libsrc/lynx/cgetc.s
@@ -20,6 +20,7 @@
 ; So the keyboard returns '1', '2', '3', 'P', 'R', 'F' or '?'.
 
 _cgetc:
+        lda     KBEDG
         bne     _start     
         jsr     _kbhit          ; Check for char available
         beq     _cgetc
@@ -39,19 +40,19 @@ _start:
         bit     #$08
         beq     @L4                     ; Pause + Opt 2 = Flip
         lda     #'?'                    ; All buttons pressed
-        rts
+        bra     reset_and_exit
 @L2:
         lda     KBSTL           ; Pause alone was the last placed button
         and     #$0c
         bne     @L1
         lda     #'P'                    ; Pause pressed
-        rts
+        bra     reset_and_exit
 @L3:
         lda     #'R'                    ; Reset pressed
-        rts
+        bra     reset_and_exit
 @L4:
         lda     #'F'                    ; Flip pressed
-        rts
+        bra     reset_and_exit
 @L5:
         lda     KBEDG           ; No Pause pressed
         stz     KBEDG
@@ -61,10 +62,13 @@ _start:
         bit     #$04
         beq     @L6
         lda     #'3'                    ; opt 1 + opt 2 pressed
-        rts
+        bra     reset_and_exit
 @L6:
         lda     #'1'                    ; opt 1 pressed
-        rts
+        bra reset_and_exit
 @L7:
         lda     #'2'                    ; opt 2 pressed
+
+reset_and_exit:
+        stz     KBEDG
         rts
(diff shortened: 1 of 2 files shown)


libsrc/lynx/cgetc.s
```diff
@@ -52,6 +54,8 @@ _cgetc:
         rts
 @L5:
         lda     KBEDG           ; No Pause pressed
+        ldx     #$00
```
BrianPeek: Can these 2 lines be replaced with "stz KBEDG" ?

Fabrizio-Caruso: I think so. Thanks! I am poor in Assembly and this is an extra opcode not available on standard 6502.

@BrianPeek Could you please test this fix?

Fabrizio-Caruso: @BrianPeek I have pushed the the fix with STZ

BrianPeek: Did a very quick compile of the lynx lib in your branch/fork and it seems to be working correctly.  I will verify again when the PR is accepted into the main branch and the Windows snapshot is available.  But, overall, this seems good.  Thanks for the fix!

greg-king5:
`kbhit()` must start with:
```asm
_kbhit:
    lda KBEDG
    bne @L1
    lda $FCB0
```
Then, `cgetc()` must start with:
```asm
_cgetc:
    jsr _kbhit
    beq _cgetc
    ora KBSTL
    ldx #0
```
Notice that `cgetc()` doesn't need to load `KBEDG` the first time because `kbhit()` already did it.

Fabrizio-Caruso:
@greg-king5 this prevents `kbhit()` from polling the status of the buttons.
Your code cannot work (I have also tested it) because once `kbhit()` sets `KBEDG` to zero, it will never be able to set it to any other value.

Fabrizio-Caruso:
@greg-king5 @BrianPeek  I have tested `cgetc()` and `kbiht()` with 
https://github.com/Fabrizio-Caruso/CROSS-CHASE/blob/master/src/test/tgi_lynx_cgetc_test.c

This PR is ready on my side if no other issue is found.

greg-king5:
> This prevents `kbhit()` from polling the status of the buttons.
> Your code cannot work (I have also tested it) because, once `kbhit()` sets `KBEDG` to zero, it will never be able to set it to any other value.

My code tests for _not_ zero.  The idea is that, after `kbhit()` detects a pressed button, it continues to say so until `cgetc()` handles _that_ button (it prevents skipped buttons).

The real problem is that `cgetc()` has several exit paths -- _every one_ of them must clear `KBEDG`!
Therefore, you must replace all but one of the `rts` instructions with `bra` instructions.
They must branch to a `stz KBEDG` instruction that you put just above the remaining `rts`.

Fabrizio-Caruso:
@greg-king5 I don't understand what is wrong functionally with the current code.
It passes both my and @BrianPeek 's tests.
My understanding is that `kbhit()` polls the current keyboard's status. So if it is run twice, it should re-poll the keyboard.
@oliverschmidt what is your opinion on this?

Fabrizio-Caruso:
@greg-king5 your interpretation of kbhit() is different from the one used on all other tsrgets. In my (universal) code to wait for a key press I successfully use:
while(kbhit())
                cgetc();
            while(!kbhit())
            { 
            }; 
            cgetc();

This would be broken with your code.

greg-king5:
> My understanding is that `kbhit()` polls the current keyboard's status. So, if it is run twice, it should re-poll the keyboard.

What happens if you press a button before the first poll, then release that button before the second poll?  Will your `kbhit()` remember the results of the first poll?  Won't `cgetc()` be told that no button is pressed?  Won't it then _wait_ for another press?  When it sees the second press, the first one will have been lost!  That causes the skipped buttons that Brian reported.

If you haven't seen that behaviour in your tests, then you were "unlucky".

> I have tested `cgetc()` and `kbhit()` with https://github.com/Fabrizio-Caruso/CROSS-CHASE/blob/master/src/test/tgi_lynx_cgetc_test.c

That test isn't a good example of an arcade game's real-time action.  The `for` loop inside the `while` loop slows the test.  That forces you to hold down a button until you see an indication -- therefore, you never release at the dangerous moment.

Here is better code:
```c
    i = 0;
    while(1)
    {
        if(kbhit())
        {
            str[0] = cgetc();
            tgi_outtextxy(8*8,8*4,str);
            i = 0;
        }
        else
        {
            if (++i == 10000)
                tgi_outtextxy(8*8,8*6,"*");
            if (i == 20000) {
                tgi_outtextxy(8*8,8*6,"+");
                i = 0;
                tgi_outtextxy(8*8,8*4," ");
            }
        }
    }
```
You'll see the button immediately.  The display will last for a short while.

Fabrizio-Caruso:
@BrianPeek  @greg-king5. I undestood Brian has tested the new code and it works for him.

Fabrizio-Caruso:
@greg-king5  so kbit() should remember key-press but if it had previously detected no key it should poll again and not the other way round. I can try to implement this check.
 
 .

Fabrizio-Caruso:
Hi! I have just made a new PR:
1. kbhit checks KBEDG and it is is different from zero it returns without pooling
2. cgetc now exists always by resetting KBEDG

libsrc/lynx/cgetc.s
```diff
@@ -37,31 +40,35 @@ _cgetc:
...
         lda     #'R'                    ; Reset pressed
-        rts
+        bra     reset_and_exit
 @L4:
         lda     #'F'                    ; Flip pressed
-        rts
+        bra     reset_and_exit
 @L5:
         lda     KBEDG           ; No Pause pressed
+        stz     KBEDG
```
greg-king5: That line isn't needed anymore (because of `reset_and_exit`).

libsrc/lynx/cgetc.s
```diff
@@ -20,8 +20,11 @@
 ; So the keyboard returns '1', '2', '3', 'P', 'R', 'F' or '?'.
 
 _cgetc:
+        lda     KBEDG
+        bne     _start     
         jsr     _kbhit          ; Check for char available
         beq     _cgetc
```
greg-king5: (That `beq` instruction is a part of why the first two lines in this function aren't needed.  And, it's why the `tax` instruction at the end of the `_kbhit` function _is_ needed.)

libsrc/lynx/kbhit.s
```diff
@@ -49,6 +51,6 @@ _kbhit:
         sta     KBNPR           ; inverted previous ones pressed
         stx     KBPRV
         lda     KBEDG
-@L1:    tax                     ; No new keys hit
+L1:     ldx     #$00
```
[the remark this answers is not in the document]
Fabrizio-Caruso: This does not work if I keep  `ldx #$00`

Fabrizio-Caruso:
@greg-king5 
This sort of works but how do I set X to zero in `kbiht()` efficiently?
Aren't correct return values defined as:
FALSE: A=0,X=0, and 
TRUE: non-zero A<>0, X anything

If I want your code to work with my tests, I need to remove `ldx #$00` in 
```
...
L1:    tax
        ldx     #$00                     
        rts
```

which passes my limited tests but what if A=0 and X<>0? Is this always interpreted in both Assembly and C as FALSE/zero?

Fabrizio-Caruso:
@greg-king5 
Travis seems blocked to me.

I have made a PR to apply your suggestions.
Could you please take a look at the PR and at the continuous integration, which seems discontinued.

Fabrizio-Caruso:
@greg-king5  @oliverschmidt  travis is stuck in pending status. Could someone please check this?

The PR is ready for me.

greg-king5:
This PR looks great to me.  Therefore, merging...

Fabrizio-Caruso:
Thanks a lot @greg-king5 for you help and advice!

Issue #822, opened 2018-12-21
plus4 TGI port from c64 TGI.


tschak909:
I've managed to port the C64 TGI driver to the PLUS4.

I still need to get the color ram setting right, but it already works well enough that PLATOTerm came up successfully (serial I/O works too, but xplus4's RS232 is currently too broken to be useful):

![successful_plus4_bringup](https://user-images.githubusercontent.com/2759782/50357232-dc9bd700-051a-11e9-984e-5a1af0ef1e5a.gif)

code is over in my fork:
https://github.com/tschak909/cc65/commit/f1e0fd23c999faf7a52e5490107ca02207f0d544

Am I doing anything ridiculously stupid? (The banking bits of the C64 don't really apply to TED, so they've been commented out, and I don't think the NMI trampoline that @greg-king5 provided for PLATOTerm is needed on the plus4.)

-Thom

rhalkyard:
Ha, I just did the same thing, motivated by "wouldn't it be neat to port PLATOTerm to the Plus/4," without realising that you'd already done this! I *think* I've managed to get the color setting right (at least, the `tgidemo` sample gives the right results). I've also added a linker config that puts the stack (and a second BSS segment) in the `$DF40`-`$FCFF` range that would otherwise go unused.

My work-in-progress is in the [feature/plus4-tgi-driver](https://github.com/cc65/cc65/compare/master...rhalkyard:feature/plus4-tgi-driver) branch in my fork.

I recently got my hands on a real Plus/4 that I can test with, so hopefully I can get to the bottom of the RS232 issues too.

tschak909:
This is 10,000 shades of awesome! :)

-Thom

mrdudz:
now in master

PR #967, opened 2019-10-30
TGI driver and IRQ-handler fix for plus4

rhalkyard:
As promised in #822, here's a TGI driver for the plus4, and a couple of other fixes

* The TGI driver is based heavily on the `c64-hi` driver. Because we can't bank out IO on the plus4, the bitmap can't go at `$E000`-`$FF39` like on the C64. Instead, we put it at `$C000`-`$DF39`. As such, programs that use the TGI driver must be linked either with `-D __HIMEM__=0xC000`, or using the new `plus4-hires.cfg` linker config, that allows the remaining ram at `$DF40`-`$FCFF` to still be used for the stack and an additional BSS segment.

* To support [PLATOTerm](https://github.com/tschak909/platoterm64), I modified the `plus4-stdser` serial driver to use the Kernal variables at `$07D1`-`$07D3` for its receive buffer pointers and counts (like the C64 and C128 SwiftLink drivers do), so that we can examine buffer usage.

* The plus4 runtime had a bug in the IRQ handler, where if an interrupt occurred while ROM was paged in, `callirq` did not get called. This could cause us to get stuck in the ROM interrupt handler (for example, using the serial driver, if data came in while we were in the middle of a Kernal call, the ACIA's interrupt would never get serviced). I've fixed this by adding a `$0314` IRQ stub that enables RAM, calls `callirq`, then enables ROM again (both the Kernal and our RAM IRQ handler jump to `$0314` with ROM banked in). We waste a few cycles doing so, but this way our handlers always run.


rhalkyard:
Note that VICE's ACIA emulation is currently broken on the plus4 (IRQs don't get generated - see [bug 1084](https://sourceforge.net/p/vice-emu/bugs/1084/) and [this patch](https://gist.github.com/rhalkyard/875f8245621103b23f4f9425e1a279ed)).

You'll need this patch to reproduce the IRQ-handling issue that I was seeing, and for serial to work reliably anyway.

samples/tgidemo.c
```diff
@@ -12,8 +12,8 @@
 #  define DYN_DRV       1
 #endif
 
-#define COLOR_BACK      TGI_COLOR_BLACK
-#define COLOR_FORE      TGI_COLOR_WHITE
+#define COLOR_BACK      0
```
polluks: Why?

samples/Makefile
```diff
@@ -228,7 +242,15 @@ multdemo: multidemo.o
...
-OVERLAYLIST := $(foreach I,1 2 3,multdemo.$I ovrldemo.$I)
+OVERLAYLIST := 
+
+ifneq ($(filter ovrldemo,$(EXELIST_$(SYS))),)
+OVERLAYLIST += $(foreach I,1 2 3,ovrldemo.$I)
+endif
+
+ifneq ($(filter multdemo,$(EXELIST_$(SYS))),)
+OVERLAYLIST += $(foreach I,1 2 3,multdemo.$I)
+endif
```
oliverschmidt: Please elaborate on this change. Especially as neither ovrldemo nor multdemo seem to get built for the Plus/4.

libsrc/plus4/crt0.s
```diff
@@ -183,6 +193,20 @@ nohandler:
...
+; __INTERRUPTOR_COUNT__ directly, since the condes function is not reentrant.
+; The irqcount flag will be set/reset from the main code, to avoid races.
+IRQStub:
+        cld                             ; Just to be sure
+        sta     ENABLE_RAM
+        ldy     irqcount
+        beq     @L1
+        jsr     callirq_y               ; Call the IRQ functions
+@L1:    sta     ENABLE_ROM
+        jmp     (IRQInd+1)              ; Jump to the saved IRQ vector
```
oliverschmidt: Please align the comment column with the pre-existent code.

libsrc/plus4/crt0.s
```diff
@@ -65,6 +64,15 @@ L1:     lda     sp,x
...
         stx     $FFFE           ; Install interrupt handler
         sty     $FFFF
+        ldx     IRQVec
+        ldy     IRQVec+1
+        stx     IRQInd+1
+        sty     IRQInd+2
+        ldx     #<IRQStub
+        ldy     #>IRQStub
+        stx     IRQVec
+        sty     IRQVec+1
```
oliverschmidt: The general cc65 style is to use A/X for addresses. A doesn't need to be preserved here anymore.

libsrc/plus4/ser/plus4-stdser.s
```diff
@@ -64,15 +64,15 @@ ACIA_STATUS     := ACIA+1       ; Status register
 ACIA_CMD        := ACIA+2       ; Command register
 ACIA_CTRL       := ACIA+3       ; Control register
 
+RecvHead        := $07D1       ; Head of receive buffer
+RecvTail        := $07D2       ; Tail of receive buffer
+RecvFreeCnt     := $07D3       ; Number of bytes in receive buffer
```
oliverschmidt: Please align the comment column with the pre-existent code.

libsrc/plus4/ser/plus4-stdser.s
```diff
@@ -353,26 +353,27 @@ SER_IOCTL:
...
 
 ; Assert flow control if buffer space too low
 
-@L1:    lda     RtsOff
-        sta     ACIA_CMD
-        sta     Stopped
-        sec                     ; Interrupt handled
+@L2:    lda     RtsOff          ;(3) assert flow control if buffer space too low
+        sta     ACIA_CMD        ;(4) command
+        sta     Stopped         ;(3)
```
oliverschmidt: Please always have a blank between the `;` and the first char of the comment. Please start comments with uppercase.

libsrc/plus4/tgi/ted-hi.s
```diff
@@ -0,0 +1,883 @@
...
+; ------------------------------------------------------------------------
+; INSTALL routine. Is called after the driver is loaded into memory. May
+; initialize anything that has to be done just once. Is probably empty
+; most of the time.
+;
+; Must set an error code: NO
+;
+
+INSTALL:
+;       rts                     ; fall through
```
oliverschmidt: Please start comments with uppercase.

libsrc/plus4/tgi/ted-hi.s
```diff
@@ -0,0 +1,883 @@
...
+; ------------------------------------------------------------------------
+; DONE: Will be called to switch the graphics device back into text mode.
+; The graphics kernel will never call DONE when no graphics mode is active,
+; so there is no need to protect against that.
+;
+; Must set an error code: NO
+;
+
+DONE:   lda     $FF12
+        ora     #%00000100      ; fetch from ROM
```
oliverschmidt: Please start comments with uppercase.

libsrc/plus4/tgi/ted-hi.s
```diff
@@ -0,0 +1,883 @@
...
+        lda     #>CHRBASE       ; Reset character/color matrix address
+        sta     $FF14
+.else
+        sta     ENABLE_ROM      ; Clear text display since we clobbered it
+        jsr     CLRSCR
+        sta     ENABLE_RAM
+.endif
+
+        lda     $FF06
+        and     #%11011111      ; exit bitmap mode
```
oliverschmidt: Please start comments with uppercase.

libsrc/plus4/tgi/ted-hi.s
```diff
@@ -0,0 +1,883 @@
...
+
+; ------------------------------------------------------------------------
+; SETVIEWPAGE: Set the visible page. Called with the new page in A (0..n).
+; The page number is already checked to be valid by the graphics kernel.
+;
+; Must set an error code: NO (will only be called if page ok)
+;
+
+SETVIEWPAGE:
+;       rts                     ; fall through
```
oliverschmidt: Please start comments with uppercase.

libsrc/plus4/tgi/ted-hi.s
```diff
@@ -0,0 +1,883 @@
...
+;
+; X1,X2 etc. are set up above (x2=LINNUM in particular)
+; Format is LINE x2,y2,x1,y1
+;
+; Must set an error code: NO
+;
+
+LINE:
+
+@CHECK: lda     X2           ;Make sure x1<x2
```
oliverschmidt: Please always have a blank between the ; and the first char of the comment.

libsrc/plus4/tgi/ted-hi.s
```diff
@@ -0,0 +1,883 @@
...
+        sec
+        sbc     X1
+        tax
+        lda     X2+1
+        sbc     X1+1
+        bpl     @CONT
+        lda     Y2           ;If not, swap P1 and P2
```
oliverschmidt: Please always have a blank between the ; and the first char of the comment.

libsrc/plus4/tgi/ted-hi.s
```diff
@@ -0,0 +1,883 @@
...
+        sbc     Y1
+        tay
+        lda     Y2+1
+        sbc     Y1+1
+        bpl     @DYPOS       ;Is y2>=y1?
+        lda     Y1           ;Otherwise dy=y1-y2
+        sec
+        sbc     Y2
+        tay
+        ldx     #$88         ;DEY
```
oliverschmidt: Please always have a blank between the ; and the first char of the comment.

libsrc/plus4/tgi/ted-hi.s
```diff
@@ -0,0 +1,883 @@
...
+        stx     XINCDEC
+
+        jsr     CALC            ; Set up .X, .Y, and POINT
+        lda     BITCHUNK,X
+        sta     OLDCHUNK
+        sta     CHUNK
+
+        ldx     DY
+        cpx     DX           ;Who's bigger: dy or dx?
+        bcc     STEPINX      ;If dx, then...
```
oliverschmidt: Please align the comment column. Please always have a blank between the `;` and the first char of the comment.

Note: Please apply the adjustments to the rest of this file...

cfg/plus4-hires.cfg
```diff
@@ -0,0 +1,57 @@
...
+
+FEATURES {
+    STARTADDRESS: default = $1001;
+}
+SYMBOLS {
+    __LOADADDR__:  type = import;
+    __EXEHDR__:    type = import;
+    __STACKSIZE__: type = weak, value = $0800; # 2k stack
+    __HIMEM__:     type = weak, value = $FD00;
+    __RESERVED_C000__: type = weak, value = 8000; # Reserve 8000 bytes for hi-res bitmap
```
oliverschmidt: Please elaborate why this is a symbol. Is there a valid use case where its value isn't 8000?

cfg/plus4-hires.cfg
```diff
@@ -0,0 +1,57 @@
...
+}
+MEMORY {
+    # Memory reserved for bitmap
+    RESERVED: file = "", define = yes, start = $C000,           size = __RESERVED_C000__;
+
+    ZP:       file = "", define = yes, start = $0002,           size = $001A;
+    LOADADDR: file = %O,               start = %S - 2,          size = $0002;
+    HEADER:   file = %O, define = yes, start = %S,              size = $000D;
+    MAIN:     file = %O, define = yes, start = __HEADER_LAST__, size = __RESERVED_START__ - __MAIN_START__;
```
oliverschmidt: Please have the `size =` re-aligned with your additional line below.

cfg/plus4-hires.cfg
```diff
@@ -0,0 +1,57 @@
...
+    CODE:     load = MAIN,     type = ro;
+    ONCE:     load = MAIN,     type = ro,  optional = yes;
+    RODATA:   load = MAIN,     type = ro;
+    DATA:     load = MAIN,     type = rw;
+    INIT:     load = MAIN,     type = bss;
+    BSS:      load = MAIN,     type = bss, define   = yes;
+
+    # Allow data between bitmap and top of memory to be used as a second BSS
+    # space. Define symbols for it so that it can be supplied to _heapadd().
+    HIBSS:    load = HIRAM,    type = bss, optional = yes, define   = yes;
```
oliverschmidt: The additional blanks between `define` and `= yes` aren't valid here as there's nothing to align to.

samples/tgidemo.c
rhalkyard: Oops, I see what you mean now; I misread the docs. Sorry!

cfg/plus4-hires.cfg
rhalkyard: The only reason is because I wanted to make it as general as possible, following the example of `__RESERVED_MEMORY__` in atari.cfg. Happy to remove it if it's not wanted.

libsrc/plus4/tgi/ted-hi.s
rhalkyard: Noted (as with all of the above)

Most of that file's code was taken directly from the c64-hi TGI driver. I might as well fix the formatting there while I'm at it (in a new PR).

cfg/plus4-hires.cfg
oliverschmidt: Afaik there are quite some graphics modes and setups to be supported by the Atari cfg. I'd suggest to just go with plain constants here - at least as long as there's only one TGI driver to support.

samples/Makefile
rhalkyard: My mistake; I didn't mean to stage that. I've written an overlay config for the Plus/4 as well, but I wasn't going to put that in this PR.

libsrc/plus4/tgi/ted-hi.s
oliverschmidt: > I might as well fix the formatting there while I'm at it (in a new PR).

That would be appreciated :-)

samples/Makefile
rhalkyard: Correction (and let this remind me that I shouldn't work on these things before I've had my coffee...):

Without overlays, this is necessary for the plus/4 use the same D64 recipe as the C64 and C128.

rhalkyard:
Sorry about the commit-spam; let me know if you'd prefer them to be squashed.

oliverschmidt:
> Sorry about the commit-spam; let me know if you'd prefer them to be squashed.

That's no issue. A nice commit history hasn't been a high prio topic for this repo so far.

samples/tgidemo.c
oliverschmidt: If I don't misunderstand the issue here then it's a variant of the issue we "always" have:

TGI_COLOR_... macros don't describe color values to be set into the palette. They describe colors in the default palette. There may be targets where those macros can be used for both purposes, but if they can't, then describing colors in the default palette takes precedence. The reasoning behind that is that TGI is meant to be a cross target API. And there are targets without palette in the first place so setting entries in the palette is already to be considered target specific code.

The user docs describe this pretty clearly: _The actual color is an index to a palette. During tgi_init you will get a default palette. The number of colors depend on the platform. All platforms recognize at least TGI_COLOR_BLACK and TGI_COLOR_WHITE. But some platforms have many more predefined colors. If you paint using TGI_COLOR_GREEN and then you change the green of the palette to blue using tgi_setpalette then after this painting in TGI_COLOR_GREEN will actually be blue._

I understand very well that it would be desirable to have a meaningful abstraction of the values to be set in the palette. This - and quite some other issues with the TGI API - have been discussed several times in the past decades. But there were no solutions because:
* there's no momentum to implement changes on all (or just the majority of) targets
* there are diametral opinions and the last significant discussion led to one member leaving the cc65 community

If TGI_COLOR_BLACK and TGI_COLOR_WHITE aren't defined as 0 and 1 then you need something like https://github.com/cc65/cc65/blob/master/libsrc/lynx/tgi_colors.s to "override" the default https://github.com/cc65/cc65/blob/master/libsrc/common/tgi_colors.s

Maybe somewhat interesting to read: https://github.com/cc65/cc65/pull/870

mrdudz:
what is the status of this PR?

mrdudz:
> what is the status of this PR?

Yeah? Please someone fix the conflicts and confirm this is a working fix we can merge

mrdudz:
It would be great if someone could pick this up and fix it... i am not very much into Plus4 myself - and other things already take too much time :)

rhalkyard:
My apologies, I suppose it's been a while! Not quite sure why I stopped responding. 2019 was a rough year for me and I let a lot of things get away from me. I suppose this PR was one of them.

Unfortunately in the intervening years my Plus/4 has had a hardware (RAM?) failure and these days is sitting uselessly on a shelf in my workshop waiting for me to get around to repairing it. This has, unfortunately, reduced my interest in working on Plus/4 graphics for the time being :(. While I'd like to revisit it one day, realistically I don't see myself having the time or energy for it any time soon.

If anyone wants to pick up my work-in-progress and get it merged, I'd strongly support that and offer whatever guidance that I can recall, but otherwise it might be a while before I'm in a position to work on it again.

mrdudz:
```
ld65: Warning: /home/runner/work/cc65/cc65/cfg/apple2.cfg:18: Segment 'BSS' overflows memory area 'BSS' by 255 bytes
../bin/cl65 --start-addr 0x4000  -o mandelbrot -t apple2 -m mandelbrot.map mandelbrot.o apple2.lib
ld65: Error: Cannot generate most of the files due to memory area overflow 
```

i tried to solve the merge conflicts... and now this? why would this PR affect anything apple2? @oliverschmidt any idea?

oliverschmidt:
As you surely noticed, there have been some changes in the Apple II area very lately. But if we can be _really_ sure that the above still works with those changes applied, then I have no idea why it stops working with this PR.

mrdudz:
> But if we can be really sure that the above still works with those changes applied, then I have no idea why it stops working with this PR.

I fail to see how anything apple2 would be relevant or clash with this... the only changes that *might* affect apple2 at all, are those in samples/Makefile and samples/tgidemo.c - but even there, i fail to see how. *shrug*

(I didnt actually do any local testing yet, just resolved the merge conflicts in the web interface)

mrdudz:
Checked a bit more... 

First of all: it was the CL vs LD change that you also wondered about. I removed this (and its now linking all plus4 samples with that "hires" linker config... because why not). That made them work again.

ps: @oliverschmidt there seems to be a small glitch in the Makefile in samples/apple2 - it builds dhgrshow.c
 for the apple2enh target, even when SYS=apple2 ... that might be misleading

So far so good.

I noticed some glitchy behaviour with the samples though (tested in xplus4, which might not be the best idea, admitted) - i don't know (yet) if this is caused by this PR though (eg the "ascii" program moves the cursor down after each printed value... the tgidemo works, but crashes when its done - which may or may not be expected)

Edit: "ascii" works ok before this PR - i cant tell if the change to cgetc is the problem, or if VICE is broken :)

mrdudz:
> Edit: "ascii" works ok before this PR - i cant tell if the change to cgetc is the problem, or if VICE is broken :)

I must have been doing something wrong last night, it works :)

I also briefly tested the RS232 stuff (in VICE) - and it seems to work

@oliverschmidt  I'll try to address your last two requests/comments. Would you try to recap what the problem is with those color defines and what has to be done here?

oliverschmidt:
> @oliverschmidt I'll try to address your last two requests/comments. Would you try to recap what the problem is with those color defines and what has to be done here?

I've reread what I wrote 6 years ago (incl. the references) several times by now. But to be honest, it's just too long ago. I don't really understand those things anymore. I'm sorry 😒

mrdudz:
> I've reread what I wrote 6 years ago (incl. the references) several times by now. But to be honest, it's just too long ago. I don't really understand those things anymore.

darn, exactly how i felt :) I remember we agreed on something at the end... but then momentum was lost with gregg.

Oh well. Might just merge as is then - it works for the most part afterall

mrdudz:
I think i remember the issue now... and fixed it :) Also updated the tgidemo a bit with some comments and less ambigious use of the TGI color defines.

mrdudz:
@oliverschmidt would you check if tgidemo still works that way on apple target(s)?

mrdudz:
@oliverschmidt ping :) would you do a quick check? i dont have apple2 emu etc handy :)

oliverschmidt:
I'm right now on the road - sorry.

samples/tgidemo.c
```diff
@@ -216,10 +240,12 @@ int main (void)
...
     /* Do graphics stuff */
+
+    /* first uses the default palette */
     DoCircles ();
-    DoCheckerboard ();
-    DoDiagram ();
-    DoLines ();
+    DoPalette (0); DoCheckerboard ();
+    DoPalette (1); DoDiagram ();
+    DoPalette (2); DoLines ();
```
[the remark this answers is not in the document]
oliverschmidt: Sorry, but "comment above" isn't specific enough for me to be able to follow you. Addionally I don't understand why the first dimension of `Palette` has a size of 4 if the code in question only uses indexes 0-2.

oliverschmidt:
I believe that the program wasn't correct - and still isn't correct. The second dimention of `Palette` has a value of 2. This implies to me that only the indexes 0 and 1 are to be used. But there's
```
#define COLOR_BACK      TGI_COLOR_BLACK
#define COLOR_FORE      TGI_COLOR_WHITE
```
which seems to imply that TGI_COLOR_BLACK is 0 and TGI_COLOR_WHITE is 1. But exactly that isn't true for all targets.

mrdudz:
> which seems to imply that TGI_COLOR_BLACK is 0 and TGI_COLOR_WHITE is 1. But exactly that isn't true for all targets.

All it does is that it expects the TGI_COLOR_ values being valid indices into the default palette - as the documentation says. (And i'm not going into another round of wrestling about it - the documentation says how it is supposed to be)

What really was broken beyond repair before in this program was that it used TGI_COLOR_ values to refernce to colors in the palette as well. how THAT ever passed anyones review is just... wow - it doesn't make sense in any way :)

oliverschmidt:
Hm, if that is your perspective that there's no point in this discussion. Do whatever you see fit and live with the result. Bye...

mrdudz:
> Hm, if that is your perspective that there's no point in this discussion. Do whatever you see fit and live with the result.

Sorry, did i hit a nerve there? What else gives the authorative answer to this, if not the documentation? Do you expect an example program to not follow documentation, because someone implemented something in the library not according to the documentation? I am baffled - i didn't expect this to be a problem. At all.

What do you suggest instead? As i see it, the program does now match the documentation - which it really should, don't you agree?

If what you say is true, then someone (someone who thinks the current way is not what it should be - not me) must redefine how TGI_COLORS_ work and what they are. Actually if we can not use them like this, there is no use for those values *at all*. Or perhaps they then can *only* be used for the palette colors themselves, but NOT AND NEVER to index into the palette.

PS: after rereading that wall of text from the past - i believe that was pretty much what you were proposing too

mrdudz:
@oliverschmidt oh damn, i have to apologize - i totally misunderstood what was written before and what i did there was wrong. So i reverted the change in tgidemo.c and defined the Plus 4 colors accordingly.

I will merge this now, and come up with another PR to fix the driver further (and add the default colors the right way, ie use tgi_colors.s?, plus update the documentation and make the difference between the color indices and TGI_COLOR values more clear.

mrdudz:
follow up in #2711

PR #968, opened 2019-10-31
C64 TGI driver: reformat comments to style guide rules

rhalkyard:
As per discussion in #967 (where I borrowed heavily from it to make a TGI driver for the plus4), an attempt to make the c64-hi TGI driver source better adhere to the style guide.


oliverschmidt:
Thanks for the housekeeping :-)

PR #977, opened 2019-11-12
Named exit constants

MonteCarlos:
The replacement of hardcoded exit codes values for exit(0) and exit(1) was straightforward in most cases. However, there was an enum introduced for the exit codes used in abort.c and _afailed.c defining the exit codes 2 and 3. 
To prevent accidential duplicated assignment of exit codes 0 and 1 with EXIT_SUCCESS and EXIT_FAILURE from stdlib.h the enum also reserves the values 0 and 1 using the constants CC65_EXIT_SUCCESS and CC65_EXIT_FAILURE. This way CC65_EXIT_XXX constants can be used throughout in the cc65 library code.
I left the stdlib.h constants in the testcode and demo code which is not part of the cc65 library itself because no one would expect that this kind of code returns library error codes.
There was a case in dbg.c in which is was intended to return an error code upon normal quit due to the fact that DbgEntry itself indicates an error condition which is just promoted. A comment was added to explicitly point this out. 

Although being a tedious procedure it resulted in a bug fixture in yacc.c where accidently success was returned when a failure should have to be returned. Thanks to Polluks for fixing this.


oliverschmidt:
In know that @polluks wrote...

> Enums are welcome :-)

...but looking at this PR I must admit that I don't see the benefit. https://github.com/cc65/cc65/blob/master/include/stdlib.h#L47-L49 are macros, so why introduce enums for doing the same thing? Especially since the enums - as written in the original comment of this PR - might interfere with the macros and therefore "require" a duplication for SUCCESS and FAILURE. All this seems to me like the solution for a problem that nobody had in the first place.

If we want EXIT_AFAILED and EXIT_ABORT then I suggest to just have two macros simply "extending" what is in https://github.com/cc65/cc65/blob/master/include/stdlib.h#L47-L49. I don't know if _EXIT_AFAILED or CC65_EXIT_AFAILED is more appropriate.

MonteCarlos:
i will have a look at it again. did not want to rise enum vs define debate. i'll remove the definition of cc65_success and cc65_failure constants and use define, instead of the enum. however, i would like to leave a comment that these cc65 exit constants should not redefine the exit codes 0 and 1 but are only defined in addition to the stdlib constants. probably someone will add more exit constants in future and he should know that 0 and 1 is already reserved.

oliverschmidt:
I fully agree with what you write (incl. the comment). Thanks for your patience.

greg-king5:
I don't like the `CC65_` prefix.  I don't like them being put in `<cc65.h>`.  I prefer simply adding them to `<stdlib.h>` [because that's where the `exit()` prototype is]:
```C
/* Standard exit codes */
#define EXIT_SUCCESS    0
#define EXIT_FAILURE    1

/* cc65 exit codes */
#if __CC65_STD__ == __CC65_STD_CC65__
#define EXIT_ASSERT     2
#define EXIT_ABORT      3
#endif
```
Note that `_afailed()` is an opaque implementation name.  We use `assert()`.  Therefore, the exit code must be named `"..._ASSERT"`.

MonteCarlos:
Hhmm, the stdlib.h header is a standard header and for portability reasons i don't think it's a good idea to put own additions in there although they're guarded with #if ...#endif.
That's why i would prefer cc65.h. However, if that's the official opinion i would move the definitions to stdlib.h, of course.
For the both renamings i'm passionless.
Sorry, i just pushed before reading the latest comment and writing this text.

oliverschmidt:
Searching the cc65 standard headers for `#if __CC65_STD__ == __CC65_STD_CC65__` shows that things I consider very similiar have been done that way before. Therefore I suggest to follow @greg-king5 proposal for consistency reasons.

@MonteCarlos: Once more thanks for your patience with the fact that the "official opinion" seems to be a pretty moving target here ;-)

MonteCarlos:
Patience is one thing i have with which i can compensate other things (◠‿◕)

libsrc/geos-apple/targetutil/convert.c
```diff
@@ -5,6 +5,7 @@
 #include <dirent.h>
 #include <device.h>
 #include <dio.h>
+#include <cc65.h>
```
oliverschmidt: That seems to be a leftover from the previous approach, isn't it?

include/stdlib.h
```diff
@@ -48,6 +48,17 @@ typedef unsigned size_t;
 #define EXIT_SUCCESS    0
 #define EXIT_FAILURE    1
 
+#if __CC65_STD__ == __CC65_STD_CC65__
+
+/* Those non-standard cc65 exit constants definitions are in addition
+   to the EXIT_SUCCESS and EXIT_FAILURE constants, which should not be
+   redefined */
```
oliverschmidt: Please adjust the multi-line comment style to match the predominant style.

libsrc/dbg/dbg.c
```diff
@@ -1580,8 +1579,11 @@ void DbgEntry (void)
             case 'q':
                 /* Quit program */
                 clrscr ();
-                exit (1);
-
+                
+                /* Exit intentionally with error because one may
+                   say that DbgEntry is always abnormal. */
```
oliverschmidt: Please adjust the multi-line comment style to match the predominant style.

MonteCarlos:
probably, thought i'd catched them all. eventually i can fix those remaining issues this evening.

PR #980, opened 2019-11-17
Fix warnings from linker-generated symbols causing a crash

### LINKED ISSUE #979: .importzp with ld65 -D causes silent failure
If I use ld65 to define a symbol with -D and then use .importzp to import that symbol, when trying to link it appears to fail to produce any output file, but also fails to produce an error return result?

When debugging there seems to be a null pointer exception in src/ld65/exports.c where in CheckSymType the string pointer ExportLI is NULL when it tries to print something about it.

Not sure exactly how to correct this at the moment, but making a note of it to follow up on later.

bbbradsmith:
fixes #979


oliverschmidt:
Thanks :-)

Issue #969, opened 2019-10-31
emd/c128-vdc.s is broken


mrdudz:
relaying from this VICE bugreport: https://sourceforge.net/p/vice-emu/bugs/1157/

apparently the driver does not handle the VDC status flag correctly, causing transfers to fail. (it works in VICE because it doesnt emulate the VDC correctly).

so, someone who has a real C128 might want to fix it :)

MonteCarlos:
I suspect this code:

VDCReadByte:
        ldx     #VDC_DATA
VDCReadReg:
        stx     VDC_ADDR_REG
@L0:    bit     VDC_ADDR_REG
        bpl     @L0
        lda     VDC_DATA_REG
        rts

VDCWriteByte:
        ldx     #VDC_DATA
VDCWriteReg:
        stx     VDC_ADDR_REG
@L0:    bit     VDC_ADDR_REG
        bpl     @L0
        sta     VDC_DATA_REG
        rts

I would change the bpl's by bmi's. If it's not working in this way it should work in the other way.
The documentation of this reg is a pity anyway and it was causing major performance issues with c128 software, because everyone put in wait states on those register writes.
As far as i remember this is necessary only for memory copying functions but not for single register reads and writes.

MonteCarlos:
I offer to try this on my c128.

polluks:
@MonteCarlos Please take care of reg 36, the RAM refresh rate

MonteCarlos:
Here is a trial i made using vdc registers without waitstates and it works pretty well on my C128 D:

https://my.pcloud.com/publink/show?code=VZ6J9QkZoLqRBLqOgqkA0pkP1aroSYzn5VDk

MonteCarlos:
Maybe you need to start the file with sys. i don't remember actually.

MonteCarlos:
I see that i used "ldx $d600" "bpl ..", too at one place in my code. So probably i have to correct my statement that the "bpl" should be exchanged with "bmi"  in the c128 vdc code.

MonteCarlos:
I just ran the tgidemo on my c128D. It loaded and ran perfectly in 80 column mode.
so, what's actually broken?
Can you specify more detailed how to reproduce the problem?

mrdudz:
check the original bug report linked in the first post, please. perhaps the bug is in em-test.c ... or perhaps even in the compiler used? that would be odd though :)

MonteCarlos:
[...]

PR #255, opened 2016-01-05
Some missing docu.

polluks:

PR #987, opened 2019-11-29
Changes in INSTALL routine from emd/c128-vdc.s.

MonteCarlos:
tmp1 was used at two places resulting in the bug that VDC_CSET was set to garbage on 16k VDC.
pagecount and curpage were not reset on INSTALL resulting in non-reentrant code on static linkage of emd driver.

I used the "both-c128-vdc.s" from greg-king as a base for this pr.

### DIFF
--- a/libsrc/c128/emd/c128-vdc.s
+++ b/libsrc/c128/emd/c128-vdc.s
@@ -52,13 +52,11 @@ VDC_DATA          = 31
 ; ------------------------------------------------------------------------
 ; Data.
 
-.data
-
-pagecount:      .word  64                  ; $0000-$3fff as 16k default
-curpage:        .word  $ffff               ; currently mapped-in page (invalid)
-
 .bss
 
+pagecount:      .res    1                  ; $0000-$3fff as 16k default
+curpage:        .res    2                  ; currently mapped-in page (invalid)
+vdc_cset_save:  .res    1
 window:         .res    256                ; memory window
 
 .code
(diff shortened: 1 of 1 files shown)


oliverschmidt:
To me this change makes sense but there are other who have been involved much more in the topic in the recent days so I ask form them to assess (and merge if appropriate) this PR...

MonteCarlos:
Googling around i stumbled upon this:
https://c-128.freeforums.net/thread/13/vdc-ram-size-detection
Several versions for vdc ram size detection are presented there.
According to the post from Jun 2, 2014 at 9:14pm,
i seems that it is most stable to use an offset of $8000 instead of $4000 for the ram size test.
This should be easy to test with the testprogram version from greg-king.

MonteCarlos:
Hopefully this works now as mentioned in https://github.com/cc65/cc65/issues/969.
I refrained from doing optimizations in order to have a working version at first.
After thinking over i moved the VDC_CSET reset code to the 16k case, only. For the 64k case it is intented to leave that 64k flag set, of course.
If that's working now one could try to implement the 16/64k ram detection code like in the following (see like above) which would approx half the size of the detection code:

;test RAM size and set flag
 lda #$00    ;point to $0000
 jsr VadrSet
 jsr ReadVram
 tay         ;save original
 lda #$80    ;point to $8000
 jsr VadrSet
 tya
 eor #$ff ;toggle all bits
 jsr WriteVram ;we just wrote to $8000, but $0000 should be unchanged...
 lda #$00     ;point to $0000
 jsr VadrSet
 jsr ReadVram
 sta index2  ;save value read
 lda #$00    ;point to $0000 again because of auto-increment
 jsr VadrSet
 tya         ;original value
 jsr WriteVram ;restore original (if it was changed)
 eor index2  ;test original (.A) with value read back from $0000
 sta only16k ;save result as flag (0=64k, not 0[$ff]=16k)
 beq DoFont  ;user has 64K

;user has 16K, reset VDC RAM size 
 ldx #vdcFont
 jsr ReadReg
 and #255-16 ;bit 4 clear to enable 4416 RAM chip
 jsr WriteReg
 bmi Done ;always

greg-king5:
I should merge this before another year goes by. 😏

oliverschmidt:
> I should merge this before another year goes by ;-)

;-)

Issue #337, opened 2016-08-19
Optimization chance: needless X negation


clbr:
``` c
#include <stdint.h>

extern uint8_t a, b;

uint8_t inversion() {
        return !!(a & b);
}
```

``` asm
.proc   _inversion: near

.segment        "CODE"

        lda     _b
        and     _a
        jsr     bnega
        jsr     bnegax
        ldx     #$00
        rts
```

It's clearly pointless to negate X when the result is 8-bit.

mrdudz:
``
return !!(a & b);

isnt that equivalent to not negating at all? or am i having a latenight brainfart here? :)

groessler:
It returns 1 and not the bits which are still set from (a & b) when (a & b) is not zero. Also don't know a use case right now but I've seen such constructs before. Probably some modern oo paradigm :-)

greg-king5:
It's a fancy way to convert from a bitfield-like value into a _pure_ boolean value.  It's equivalent to
```C
return (a & b) != 0;
```

prandeamus:
I've been looking at this, and have a side question to ask. It may be a bug. Compiling the following with -Oi

```c
uint8_t a;
uint8_t b;
uint8_t test2(void) {
    return (a & b) != 0;
}
```
generates

```asm
lda     _b
and     _a
lda     #$00
jmp     boolne
```
Given that the first instruction of boolne is a condtional branch (as for all the corresponding functions in makebool.s) I'm not sure if this correct. Surely the lda #0 will guarantee the Z flag is set on entry to boolne?

```c
int main() {
    uint8_t c,d;
    
    a=0xff;
    b=0xff;
    c=test2();
    a=0x00;
    b=0x00;
    d=test2();

    printf("%d,%d\n", c,d);

    return EXIT_SUCCESS;
}
```
Compile with
```cfg 
cl65 -O --target sim6502 -l hello.lst -T hello.c
sim65 hello
```
Results "0,0" when I would have expected c to be non-zero. 

Take out the -Oi flag, and the result is "1,0".  It's much slower and bigger code, but it extends a to int, pushes to stack, loads primary register with b, then ands the TOS with primary register, does a 16 bit comparison and finally calls boolne. Bloated but probably correct.

```asm
ldx     #$00
lda     _a
jsr     pushax
ldx     #$00
lda     _b
jsr     tosandax
cpx     #$00
bne     L0007
cmp     #$00
jsr     boolne
ldx     #$00
jmp     L0004
L0004:	rts
[...]

Issue #992, opened 2019-12-17
Wrong code optimization on loop (char) condition?


acqn:
```c
int wrong(int);
char i;

int main(void)
{
	i = 10;
	while (--i) ;
	return wrong(0);
}
```
Compiled with any "-O" option, the generated code becomes something like (only the relevant part quoted here) :
```asm
.segment	"CODE"

	lda     #$0A
	sta     _i
L0004:	dec     _i
	ldx     _i
	bne     L0004
	jmp     _wrong
```
CC65 with optimization on seems to errorneously assume that register A must be 0 after the loop that checks against 0, and will pass it to any function that takes 0 as the parameter right after the loop.

This issue goes away if optimization is turned off, as expected.

acqn:
Could the culprit be Line 1474 in cc65\coptind.c?
It is supposed to check whether a REG is not being used (that is not marked with certain bit-flags), but in a seemingly wrong way:
`GetRegInfo (S, I+2, E->Chg) != E->Chg`
I think the correct way to check the flags should be:
`(GetRegInfo (S, I+2, E->Chg) & E->Chg) == 0`

oliverschmidt:
Fixed with https://github.com/cc65/cc65/pull/996.

Issue #971, opened 2019-11-07
Possible bug in code optimization?


henriko-se:
I got into a problem using some portable third party library on the C64. It works if I don't use any optimization. But, it behaves very strange when having optimization enabled.

Managed to isolate the problem into the small test program below. Perhaps I would not write code like this myself, but I don't think there is anything in the code that is illegal, dangerous, or undefined.
```C
/*
 * test.c
 *
 * Small test program for demonstrating
 * a possible bug in the code optimization.
 *
 * When compiled using
 * > cl65 -t c64 -o test.d64 test.c
 * the result is the expected 00
 *
 * When compiled using
 * > cl65 -t c64 -O -o test.d64 test.c
 * the result is the unexpected EE
 *
 * > cl65 --version
 * cc65 V2.18 - Git 1d39863a
 * > cc65 --version
 * cc65 V2.18 - Git 1d39863a
 *
 * Used VICE to test, as this
 * > vice-jz.x64 -autostart test.d64
 */

#include <stdio.h>
#include <stdint.h>

uint16_t global;

static uint8_t unpredictable() {
    uint8_t result;
    uint16_t leftside = 0x0020u;
    uint16_t rightside = 0x0020u;

    global = leftside & rightside;

    if (global & (uint16_t)0xFFu)
    {
      result = 0x00u;
    }
    else
    {
      result = 0xEE;
    }

    return result;
}

void main(void)
{
  uint8_t zeroplease;
  zeroplease = unpredictable();

  printf("test of unpredictable results\n");
  printf("zero please: %02x\n", zeroplease);
}
```

mrdudz:
interesting... does it still break when using "static uint8_t unpredictable(void)" ?

henriko-se:
Good point. I missed miss that. But adding void in the function definition didn't dodge the problem. Still get the unexpected result when optimization is enabled.

oliverschmidt:
As expected this isn't related in any way to the C64:
```
>cl65 -t sim6502 test.c
>sim65 test
test of unpredictable results
zero please: 00
>cl65 -t sim6502 -O test.c
>sim65 test
test of unpredictable results
zero please: ee
```

acqn:
I think the issue is caused by a mistake in OptPushPop. It optimizes out PHA/PLA pairs followed by a STA but ignores later usage of N/Z changes caused by the PLA. A quick fix could be just to insert a LDA to reload the N/Z states from the STA'ed value in place of the removed PLA.

[...]

Issue #784, opened 2018-10-20
Regression: "0 < i--" is optimized to "0 < --i".


vrubleg:
I have tried to build “[Sir Ababol](http://www.mojontwins.com/juegos_mojonos/sir-ababol-nes/)” using the latest version of cc65. For some reason, it had a very glitchy title screen. I have investigated it a bit, and it turned out that the issue is in this function:
```c
void __fastcall__ un_rle_screen (unsigned char *rle) {
	wx = 0; wy = 0; i = 0;
	while (i < 240) {
		t = (*rle);
		rle ++;
		if (t & 128) {
			top = t & 127;
			t = (*rle);
			rle ++;
			while (0 < top--) {
				draw_tile (wx, wy, t);
				wx = (wx + 2) & 31; if (!wx) wy += 2;
				i ++;
			}
		} else {
			draw_tile (wx, wy, t);
			wx = (wx + 2) & 31; if (!wx) wy +=2;
			i ++;
		}
	}
	// 64 attributes
	i = 64; gp_addr = 0x23c0;
	while (0 < i--) {
		vram_adr (gp_addr++);
		vram_put (*rle++);
	}
}
```
Just look at this part of generated assembly:
```Assembly
;
; while (0 < top--) {
;
L330D:	lda     #$00
	dec     _top
	cmp     _top
	bcc     L330B
;
; } else {
;
	jmp     L3B6B
```
It decrements `_top`; and only then, it does the comparison. But, it should do comparison first; and only then, it should decrement `_top`. I have replaced it to this code:
```Assembly
L330D:
	lda _top
	dec _top
	cmp #$00
	beq L3B68
	bcs L330B
	jmp L3B68
```
And now, it works as expected.

Old cc65 had generated this monstrous code:
```Assembly
;
; while (0 < top--) {
;
L330D:	jsr     push0
	lda     _top
	pha
	sec
	sbc     #$01
	sta     _top
	pla
	jsr     tosicmp0
	bcc     L330B
;
; } else {
;
	jmp     L3B6B
```
It seems that someone had tried to add a great optimization, and forgot to take into account the difference between `--i` and `i--`. It would be really nice to preserve the optimization, but to do it in a right way.

oliverschmidt:
First of all thanks for the analysis already done so far :-)

Could you additionally try to find the (close to) minimal code that triggers the bug? Usually that helps a lot when trying to find the issue.

vrubleg:
[...]

PR #996, opened 2019-12-31
[cc65] Fixes for some opt bugs

### LINKED ISSUE #167: equality comparison is evaluated wrong in some cases
Here is an example that demonstrates the behaviour:

``` c
    #include <stdio.h>
    #include <stdint.h>

    int main ( void)
    {
      uint8_t  joy_state = 0x7e;
      printf("bad: %u\n", 0 == (joy_state & 1) );
      printf("good: %u\n", (joy_state & 1) == 0 );

      return 0;
    }
```

This results in:

```
bad: 0
good: 1
```

http://www.cc65.org/mailarchive/2014-10/11680.html

acqn:
Fix <s>#167 (dropped),</s> #971 and #992.

Edit: Also fixed the bug in #337.

### DIFF
--- a/src/cc65/coptind.c
+++ b/src/cc65/coptind.c
@@ -1948,17 +1948,22 @@ unsigned OptPushPop (CodeSeg* S)
                     !MemAccess (S, Push+1, Pop-1, E)) {
 
                     /* Insert a STA after the PHA */
-                    X = NewCodeEntry (E->OPC, E->AM, E->Arg, E->JumpTo, E->LI);
+                    X = NewCodeEntry (OP65_STA, E->AM, E->Arg, E->JumpTo, E->LI);
                     CS_InsertEntry (S, X, Push+1);
 
                     /* Remove the PHA instead */
                     CS_DelEntry (S, Push);
 
+                    /* Insert a LDA after the PLA */
+                    X = NewCodeEntry (OP65_LDA, E->AM, E->Arg, E->JumpTo, CS_GetEntry (S, Pop)->LI);
+                    CS_InsertEntry (S, X, Pop+1);
+
                     /* Remove the PLA/STA sequence */
-                    CS_DelEntries (S, Pop, 2);
+                    CS_DelEntry (S, Pop);
+                    CS_DelEntry (S, I);
 
                     /* Correct I so we continue with the next insn */
-                    I -= 2;
+                    --I;
 
                     /* Remember we had changes */
                     ++Changes;
--- a/src/cc65/coptind.c
+++ b/src/cc65/coptind.c
@@ -1968,7 +1968,7 @@ unsigned OptPushPop (CodeSeg* S)
                     /* Remember we had changes */
                     ++Changes;
 
-                } else if ((E->Info & OF_CBRA) == 0     &&
+                } else if (!CE_UseLoadFlags (E) &&
                            (!RegAUsed (S, I) || !ChgA)) {
 
                     /* We can remove the PHA and PLA instructions */
--- a/src/cc65/coptind.c
+++ b/src/cc65/coptind.c
@@ -1471,7 +1471,7 @@ unsigned OptTransfers2 (CodeSeg* S)
             (N = CS_GetNextEntry (S, I)) != 0       &&
             !CE_HasLabel (N)                        &&
             (N->Info & OF_XFR) != 0                 &&
-            GetRegInfo (S, I+2, E->Chg) != E->Chg) {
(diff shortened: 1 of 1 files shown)


acqn:
So it's the yaccdbg.c test that failed. I'll see what's wrong with it.

acqn:
For whatever reasons I couldn't make the fixed commit for #167 into this PR (I am using Mercurial)... Maybe I'd just drop that in this PR for now.

oliverschmidt:
@acqn: First of all thanks for all the effort you put in cc65 so far. Fixing bugs is not the most funny thing to do ;-)

> Edit: I've made the fix for #167 and #784 (and maybe #781 as a side effect). Should I append that to this PR or wait for a new one?

I'm not sure what you refer to with "wait". I presume you are aware that you can send several pull requests simultaneously. You just need to make sure that each pull request refers its own branch in your fork.

oliverschmidt:
I must admit that I'm having a pretty hard time to review this PR.

It's always the same with bugs in the optimizer:

Often proposed changes make the optimizer "weaker" by not applying a certain optimization in a certain scenario. Is it appropriate to consider those changes as fixes or are they rather workarounds? I mean, completely deactivating the optimizer would for sure make all bugs related to it go away ;-) So the question is if it was plainly wrong to try a certain optimization in a certain scenario in the first place (then not doing it would be a true fix) or if the optimization is rather conceptually correct but was "just" done in a wrong way (then not doing it would only be a workaround).

I see only two ways to approach this right:

- Understand enough of the cc65 optimizer to assess the fix vs. workaround aspect.
- Test cc65 in several complex scenarios not only for functionality but also for speed and size of the generated code in order to be able to assess the speed loss and/or code size increase in real world scenarios.

Unfortunately I personally am certainly unable to invest the time necessary for one way or the other :-(

Therefore I'm asking for advice on how to proceed with changes like the ones proposed here!

In case nobody comes up with help I'd tend to ask @acqn to mark his changes in the optimizer with hints to the "fixed" issue thus at least potentially helping someone to re-assess the changes later like this:

```
/* Fix for GitHub issue <issue URL> */
<new code>
/* Was: <old code> */
```

Or am I totally missing the point here(?)

Oliver

polluks:
Indeed, the optimizer acts like like a mobile. Maybe we need more unit testing?

mrdudz:
> Maybe we need more unit testing?

is that a question? :)

oliverschmidt:
Okay, today I found some time to play with this PR:

* I could not find any negative side effect on the effectiveness of the optimizer.
* I could confirm that the issues https://github.com/cc65/cc65/issues/337, https://github.com/cc65/cc65/issues/971 and https://github.com/cc65/cc65/issues/992 are fixed.

From that perspective it seems appropriate to consider this PR as true fix and therefore merge it as-is.

@acqn: Thanks for providing the PR and sorry for the delay in processing!

polluks:
@oliverschmidt How about closing these issues including #167?

oliverschmidt:
First of all thanks for reminding me on the necessary cleanup.

I have closed https://github.com/cc65/cc65/issues/337, https://github.com/cc65/cc65/issues/971 and https://github.com/cc65/cc65/issues/992 as I had confirmed their fix back in Februar.

However, I didn't check https://github.com/cc65/cc65/issues/167 as the original message said "Fix <s>#167 (dropped),</s>". Therefoe I personally can't confirm it being fixed. Do you have a clear understanding of the current state of https://github.com/cc65/cc65/issues/167? If yes, please elaborate.

polluks:
The fork https://github.com/acqn/cc65/commit/b6104e447644861811f3655a8bd005ccf69a49f1 seems to contain a 167 fix, but there's no PR. However it's @mrdudz issue...

mrdudz:
? what's "my issue" there?

oliverschmidt:
@polluks: Now you lost me...

First you write:

> How about closing these issues including #167?

This seems to imply that #167 should be closed now. But then you write:

> The fork acqn@b6104e4 seems to contain a 167 fix, but there's no PR.

This pretty much implies that #167 should not be closed now. Can please try to provide a conclusive statement on your perspective?

polluks:
@oliverschmidt I didn't notice the patch is not part of the trunk and @acqn seems be to inactive :-(
@mrdudz #167 is your baby ;-)

mrdudz:
aaah i just clicked on https://github.com/acqn/cc65/commit/b6104e4 and wondered wth its about :)

oliverschmidt:
@polluks: Thanks for the clarification. So from my understanding all relevant issues are now in the state that correctly reflects the current code in this repo.

PR #997, opened 2020-01-02
ctype size optimization

IrgendwerA8:
Like discussed, here my changes for the ctype handling.
Some remarks:
Saving is about 100 bytes in typical use case. When ctype functions are used extensively (like in the unit test) more than 170 bytes are saved.
Currently I'm not content with the stricmp, strnicmp and atoi functions I had to adapt. I will try improving them in the future (whenever that may be) - but I think the changes to ctype are already now of "some use".


asminc/ctypetable.inc
```diff
@@ -0,0 +1,48 @@
...
+;
+; See "LICENSE" file for legal information.
+;
+; Data covering all possible combinations of character flags for target specific definition
+;
+
+.include	"ctype.inc"
+.export		__ctype
+
+; This data is a table of possible ctype combinations, possible during
```
dmsc: possible during what?

asminc/ctypetable.inc
```diff
@@ -0,0 +1,48 @@
...
+ct_lower:				.byte CT_LOWER
+ct_upper:				.byte CT_UPPER
+ct_digit_xdigit:		.byte CT_DIGIT | CT_XDIGIT
+ct_lower_xdigit:		.byte CT_LOWER | CT_XDIGIT
+ct_upper_xdigit:		.byte CT_UPPER | CT_XDIGIT
+ct_ctrl:				.byte CT_CTRL
+ct_ws:					.byte CT_OTHER_WS
+ct_ctrl_ws:				.byte CT_CTRL | CT_OTHER_WS
+ct_space_spacetab:		.byte CT_SPACE | CT_SPACE_TAB
+ct_ctrl_ws_spacetab:	.byte CT_CTRL | CT_OTHER_WS | CT_SPACE_TAB
```
dmsc: Looking at the isspace / isblank code, do we need CT_OTHER_WS?
As the only place used is on "isspace", and there is used as (CT_SPACE | CT_OTHER_WS), imho, one flag should be enough.

libsrc/common/atoi.s
```diff
@@ -54,9 +54,9 @@ L3:     iny
 L5:     stx     tmp1            ; remember sign flag
 
 L6:     lda     (ptr1),y        ; get next char
-        tax
-        lda     __ctype,x       ; get character classification
-        and     #$04            ; digit?
+                                ; get character classification
+        jsr     ctype_preprocessor_no_check
+        and     #CT_DIGIT       ; digit?
```
dmsc: This code uses "ctype" here, but bellow it assumes a digit is '0' to '9' (see the sbc #'0' at line 94), so it would be smaller (and faster) to simply " sbc #'0' / cmp #'9' "

libsrc/common/isascii.s
```diff
@@ -0,0 +1,27 @@
...
+; int isascii (int c);
+;
+
+        .export         _isascii
+
+_isascii:
+        cpx     #$00            ; Char range ok?
+        bne     @L1             ; Jump if no
+
+        tay
```
dmsc: This is smaller and should be equivalent:
`  ASL
   TXA
   BNE @L1
   ADC #$FF
   RTS
  `

IrgendwerA8: Saved another byte:

`
_isascii:
        asl		a				; high-bit to carry
        txa 					        ; check range of input param
        beq		@L1			; out-of bounds?
        ldx		#$00			; yes, prepare returning of 'false'
        sec						; (will result in '0' due to next adc)
@L1:    adc 	#$FF			; calculate return value based on carry
        rts
`

libsrc/common/strupper.s
```diff
@@ -25,11 +24,10 @@ _strupr:
...
 loop:   lda     (ptr1),y        ; get character
         beq     L9              ; jump if done
-        tax
-        lda     __ctype,x       ; get character classification
+        jsr     ctype_preprocessor_no_check 
         and     #CT_LOWER       ; lower case char?
         beq     L1              ; jump if no
-        txa                     ; get character back into accu
+        lda     (ptr1),y        ; fetch character again
         clc
```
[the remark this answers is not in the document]
dmsc: Note that there is also the special case when `('A'-'a') = $20` (most ASCII based charsets), there you could set CT_LOWER to $20 and simply do:

```
    jsr ctype_preprocessor_no_check
    and #CT_LOWER
    eor (ptr1), y
    sta (ptr1), y 

```
This is 5 bytes smaller, but I don't know if it is worth.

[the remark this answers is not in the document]
IrgendwerA8: On the CBM target, 'A' is 193, while 'a' is 65 (at least looking at the CBM ctype map, I'm not that familiar with that target).
While I like your solution, I think making this code target dependent isn't worth the hassle...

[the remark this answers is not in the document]
greg-king5: > On the CBM target, 'A' is 193, while 'a' is 65 (at least looking at the CBM ctype map, I'm not that familiar with that target).

Yes.  `lda #<('A'-'a')` is the same as `lda #$80`.  The idea is that the SHIFT-key sets the high bit in the character code.

asminc/ctype.inc
```diff
@@ -1,12 +1,14 @@
+; ctype.inc
 ;
-; Definitions for the character type tables
+; This file is part of
+; cc65 - a freeware C compiler for 6502 based systems
 ;
-; Ullrich von Bassewitz, 08.09.2001
```
spiro-trikaliotis: Is this file a complete replacement for the original one?
Personally, I always have problems removing a person from the Copyright notice in a file. Even if I change the header, I leave the original name there, to avoid any problems.

libsrc/common/atoi.s
IrgendwerA8: Not sure if I understand this.
You propose to change:
```
         jsr     ctype_preprocessor_no_check
        and     #CT_DIGIT       ; digit?
        beq     L8              ; done
```
to
```
        sbc     #'0'
        cmp     #'9'
```
I guess a 'clc' before 'sbc' is needed, so that a 'bcs L8' after the cmp would work?

asminc/ctypetable.inc
IrgendwerA8: Seems that you are right.
I would change `CT_OTHER_WS` to 'CT_WHITESPACE' and remove the 'CT_SPACE' definition.
Comments?

libsrc/common/atoi.s
greg-king5: @dmsc is saying that `atoi()` didn't need to use the `isdigit()` function, at all!  There's a much simpler way to validate the input string:  get the next character, convert it to binary, then check the binary range.  If the value is 10 or greater, then the character wasn't a digit; `atoi()` has finished scanning the string.  If the value isn't 10 or greater, then save it in .X, multiply the running total by ten, add the saved value, and continue.

asminc/ctypetable.inc
IrgendwerA8: Had a second look. As this change wouldn't save a single byte (all entries in the ctype-table would be still in use) and `CT_CTRL_SPACE` and `CT_NOT_PUNCT` definition use `CT_SPACE` (in `isgraph()` and `ispunct()`) but not `CT_OTHER_WS` I won't touch the current state.

libsrc/common/atoi.s
IrgendwerA8: Changed and 4 bytes saved.

asminc/ctype.inc
IrgendwerA8: Changed, yes it's better to preserve the information.

IrgendwerA8:
Anything missing that prevents the integration?

oliverschmidt:
It seems that I haven't followed the discussion closely enough, sorry :-(

Is there anybody appealing against the merge?

IrgendwerA8:
Done.

libsrc/common/ctype.s
```diff
@@ -0,0 +1,92 @@
+; ctype.s
+;
+; This file is part of
+; cc65 - a freeware C compiler for 6502 based systems
+;
+; https://cc65.github.io
+;
+; See "LICENSE" file for legal information.
+;
+; Character specification table, matching serveral consoles.
```
oliverschmidt: as far as I know the 'c' in ctype stands for 'character', not for 'console'. So I'm wondering where the term 'console' comes from. In case you refer to the game consoles supported by cc65 I'd prefer 'machines' or 'targets'.

libsrc/common/isascii.s
```diff
@@ -0,0 +1,24 @@
+; isascii.s
```
oliverschmidt: One more file with tab chars...

libsrc/atari2600/ctype.s
```diff
@@ -1,162 +1,5 @@
-;
-; Ullrich von Bassewitz, 2003-10-10
-;
 ; Character specification table.
 ;
+; uses the "common" definition
```
oliverschmidt: I thought this "include pattern" could be replaced by that "vpath shadowing pattern". Could you please elaborate why it doesn't work here (and for the other files using the same pattern)?

libsrc/common/ctype.s
IrgendwerA8: Contrary to the ctype_common.inc (also machines/targets), this source file fits mainly to game consoles (yes, and the Apple2).

libsrc/atari2600/ctype.s
IrgendwerA8: Because the 'ctype_common.inc' is/was different from the the 'ctype_console.inc'. Those both files match to most targets and I prefer a central definition. I thought you, being an Apple 2 guy, prefer the Apple 2 definition to be the default one... ;-)

libsrc/common/isascii.s
IrgendwerA8: Amended.

libsrc/atari2600/ctype.s
oliverschmidt: Maybe I'm just slow on take up but I still don't get it :-(

In https://github.com/cc65/cc65/blob/master/src/common/target.c#L183-L218 we have a table stating which target uses which char map. I would presume that:

* All targets listed there with `CTNone` do NOT have a ctype.s file in their libsrc directory - as they just all "fall back" to the ASCII ctype.s in libsrc/common.

* All targets which are listed there with some other `CT<...>` else DO have a ctype.s file in their libsrc directory (or in the libsrc/cbm directory).

So the only scenario where there's a need for including a file would be "secondary" targets like the telestrat based on the atmos - and there one includes the "primary" ctype.s, but not an .inc file.

IrgendwerA8: There is a difference between translation and classification.
The ´CTNone´ applies to all targets with the same translation.
These targets however, differ in classification of the characters. Some of them support only 128 characters, leaving the rest 'unclassified' (ctype_common.inc) while others having inverted characters when the high bit is set (former ctype_console.inc now ctype.s). To those characters the classification is independent from the most significant bit.
For the first group of machines I like to share the definition rather than copy it to each target (which was the state before my PR), this is done be including the common definition.

oliverschmidt: Thanks for the elaborate explanation :-)

PR #1000, opened 2020-01-24
[Telestrat] Fix bug in tgi_line : HRS(X) parameters are 16 bits.

jedeoric:
HRS1,HRS2,HRS3,HRS4 are 16 bits (signed).

This PR fix this bug : impossible to draw a line when X>127 or Y>127

### DIFF
--- a/libsrc/telestrat/tgi/telestrat-228-200-3.s
+++ b/libsrc/telestrat/tgi/telestrat-228-200-3.s
@@ -297,7 +297,13 @@ LINE:
         lda   Y2
         sta   HRS4
         
-        lda   #$ff
+        lda   #$00
+        sta   HRS1+1
+        sta   HRS2+1
+        sta   HRS3+1
+        sta   HRS4+1        
+
+        lda   #$FF
         sta   HRSPAT
 
         BRK_TELEMON(XDRAWA)
--- a/libsrc/telestrat/tgi/telestrat-228-200-3.s
+++ b/libsrc/telestrat/tgi/telestrat-228-200-3.s
@@ -296,12 +296,18 @@ LINE:
         sta   HRS3
         lda   Y2
         sta   HRS4
-        
-        lda   #$00
+
+        lda   X1+1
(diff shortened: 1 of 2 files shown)


libsrc/telestrat/tgi/telestrat-228-200-3.s
```diff
@@ -297,7 +297,13 @@ LINE:
         lda   Y2
         sta   HRS4
         
-        lda   #$ff
+        lda   #$00
+        sta   HRS1+1
+        sta   HRS2+1
+        sta   HRS3+1
+        sta   HRS4+1
```
dmsc: Don't know anything about the Telestrat OS, but if this is needed here shouldn't also be needed at SETPIXEL?

jedeoric: I tested SETPIXEL, and the problem only appears with LINE

jedeoric:
I think that tgi line accept clipping because parameters are signed int. It means that my modification broke clipping. Or is it 'int' because others target are resolution bigger than 240X200 (Oric resolution)

oliverschmidt:
Line clipping is done in the shared TGI "kernel" code.

greg-king5:
> I think that TGI line accept clipping because parameters are signed int. It means that my modification broke clipping. Or, is it 'int' because other targets are resolution bigger than 240X200 (Oric resolution)?

Yes, some platforms need nine bits of precision.  Also, when calculating the slant of the line, the co-ordinates are subtracted from each other.  Those differences must be signed (in order to handle all angles).  Therefore, at least 10 bits are needed.

The TGI kernel gives 16-bit co-ordinates to a driver.  Your driver should copy both bytes of each co-ordinate to the `HRS` arguments of the firmware.

jedeoric:
I will commit modification soon

jedeoric:
Modifications pushed, let me know if it's OK

oliverschmidt:
@greg-king5: Please merge if you agree.

PR #1007, opened 2020-03-06
Update vpeek.s to enable Bank 1 of Vera to be inspected.

mattuna15:
The ldx line prevents vpeek from working in bank 1 of vera. Not sure why this was here and commenting this line out makes it work. Not sure why it is there? Thanks.


greg-king5:
That line has nothing to do with VERA.  It's the normal way to "promote" 8-bit function return values into 16-bit return values.  See https://cc65.github.io/doc/cc65-intern.html#ss1.3

PR #1010, opened 2020-03-12
+cc65 inline asm stp mnemonic support

mlauke:
hi,

want to support stp mnemonic from within inline asm of a c program, to break into my emulator debugger.

cheers,

https://github.com/twoinke/steckschwein-emulator

### DIFF
--- a/src/cc65/codeent.c
+++ b/src/cc65/codeent.c
@@ -615,6 +615,9 @@ void CE_GenRegInfo (CodeEntry* E, RegContents* InputRegs)
         case OP65_BRK:
             break;
 
+        case OP65_STP:
+            break;
+
         case OP65_BVC:
             break;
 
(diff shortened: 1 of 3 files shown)


src/cc65/codeent.c
```diff
@@ -615,6 +615,9 @@ void CE_GenRegInfo (CodeEntry* E, RegContents* InputRegs)
         case OP65_BRK:
             break;
 
+        case OP65_STP:
+            break;
+
```
greg-king5: Please move that case to a position in alphabetical order.  That is, place it just above `case OP65_STX:`.

src/cc65/opcodes.h
```diff
@@ -114,6 +114,7 @@ typedef enum {
     OP65_SED,
     OP65_SEI,
     OP65_STA,
+    OP65_STP, /* 65c02, 65816 stop */
```
greg-king5: Please move the comment over to character column 33 (that is, align it with the other comments).

polluks: Now jump on one leg ;-)

oliverschmidt:
@mlauke: Thanks for contributing to cc65.

@greg-king5: Please merge as you see fit.

mlauke:
jumped on one leg...

Issue #918, opened 2019-06-28
[ld65][bug] Cannot set segment offset to $00 explicitly


domgetter:
When setting the offset of a segment to $00 (or anything that calculates to $00), the linker outputs a "Range error".  Expected behavior is that it behaves as if an offset was not specified.

Example:
```ini
SEGMENTS {
    DATA00: load = DATABANKS, type = ro, offset = $2000 * $00;
    DATA01: load = DATABANKS, type = ro, offset = $2000 * $01;
}
```
If that is changed to:
```ini
SEGMENTS {
    DATA00: load = DATABANKS, type = ro;
    DATA01: load = DATABANKS, type = ro, offset = $2000 * $01;
}
```
ld65 correctly links the code, and produces the expected output file.

If the `$2000 * $00` is changed to `$00`, it produces the same error:
```
ld65: Error: ld_cfg_file.cfg(53): Range error
```

groessler:
Could you post complete cfg file, please?

domgetter:
Here is a minimum example:

program.asm
```
.segment "DATA00"
.segment "CODE"
Main:
  JMP Main
```
config.cfg
```
MEMORY {
    DATABANKS: file = %O, start = $0000, size = $1000, fill = yes, define = yes;
}
SEGMENTS {
    DATA00: load = DATABANKS, offset = $0000, type = ro;
    CODE: load = DATABANKS, offset = $0400, type = ro;
}
```
build.bat
```
ca65 program.asm -o program.o
ld65 program.o -C config.cfg -o program.bin
```


And if config.cfg is changed to this:
```
MEMORY {
    DATABANKS: file = %O, start = $0000, size = $1000, fill = yes, define = yes;
}
SEGMENTS {
    DATA00: load = DATABANKS, type = ro;
    CODE: load = DATABANKS, offset = $0400, type = ro;
}
```

then it links without error.

greg-king5:
Fixed by commit 28584b31f1b3560173c16487f45aad23dc6f9fb5.

oliverschmidt:
Thanks :-)

PR #1011, opened 2020-03-21
SEGMENT start of 0 should be valid?

bbbradsmith:
I don't understand why a segment with start address of 0 generates a range error. Is there some reason this is disallowed? The change to enable it is very small.

### DIFF
--- a/src/ld65/config.c
+++ b/src/ld65/config.c
@@ -753,7 +753,7 @@ static void ParseSegments (void)
 
                 case CFGTOK_START:
                     FlagAttr (&S->Attr, SA_START, "START");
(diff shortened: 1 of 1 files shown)


oliverschmidt:
I agree with @bbbradsmith. Is there something we both aren't aware of?

dmsc:
I checked version 2.3 (oldest that I could find) at the check was still there, so no hint on why it was not allowed. IMHO, i will merge this.

oliverschmidt:
@dmsc: Thanks for your assessment.

Merging now...

greg-king5:
This is a companion to issue #918.  (I wasn't thorough enough. :-/ )

oliverschmidt:
@greg-king5: Thanks for the feedback :-)

PR #1013, opened 2020-03-25
Update get_ostype.s for the Apple //e card

jeremysrand:
This change removes the check of $fbbe when detecting the Apple //e card.  This byte is a version number for the Apple //e card according to misc technote `#7`, and it appears that the last version of the software that I am aware of has a 3 at this location.

Prior to this change, Apple //e cards which were not version 0 would be detected as an Apple //e enhanced.

Note that I do not have access to this HW but one of my projects was tested by someone with this HW and we found this issue.  The tech note can be found here:

http://www.1000bit.it/support/manuali/apple/technotes/misc/tn.misc.07.html

Another alternative would be to create OS types $41, $42, $43 (perhaps more) based on the value of the version byte in the ROM for the Apple //e card which would allow projects to differentiate between these variants.  If an approach like that is preferred, I can make that change but I don't know for sure how many versions there may be in existence.

### DIFF
--- a/libsrc/apple2/get_ostype.s
+++ b/libsrc/apple2/get_ostype.s
@@ -36,7 +36,7 @@ index:  .byte   $B3, $00                ; Apple ][
         .byte   $B3, $1E, $00           ; Apple ][+
         .byte   $B3, $1E, $00           ; Apple /// (emulation)
         .byte   $B3, $C0, $00           ; Apple //e
-        .byte   $B3, $C0, $DD, $BE, $00 ; Apple //e Option Card
+        .byte   $B3, $C0, $DD, $00      ; Apple //e Option Card
         .byte   $B3, $C0, $00           ; Apple //e (enhanced)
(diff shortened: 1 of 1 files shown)


oliverschmidt:
> [...] The tech note can be found here: [...]

As https://github.com/cc65/cc65/pull/1013/files#diff-801f6f1d454c2e1b4eff62b66ab8c9c2L10-L11 shows, the code is actually based on that very tech note.

However, I failed to read/understand this section of the tech note in question:

_Location $FBBE is the version byte for the Apple IIe Card (just as $FBBF is the version byte for the Apple IIc family) and is $00 for the first release of the Apple IIe Card._

So thanks for pointing out :-)

> Another alternative would be to create OS types $41, $42, $43 (perhaps more) based on the value of the version byte in the ROM for the Apple //e card which would allow projects to differentiate between these variants.

As the tech note in question doesn't list actual versions (in contrast to //c versions) I'd opt to rather not do so.

PR #1017, opened 2020-03-30
fix keyboard input to mouse sample program on c64

doj:
the mouse sample program used the tolower() function to read key pressed. But this was not working on c64, probably due to ASCII vs PETSCI. I've changed the code to check for lowercase and uppercase characters in the switch block, and the automatic PETSCI to ASCII conversion of the compiler should to the right thing.

Also fixed the clean: target of the Makefile to remove any disk images.

### DIFF
--- a/doc/coding.sgml
+++ b/doc/coding.sgml
@@ -251,7 +251,7 @@ Register variables may give faster and shorter code, but they do also have an
 overhead. Register variables are actually zero page locations, so using them
 saves roughly one cycle per access. The calling routine may also use register
 variables, so the old values have to be saved on function entry and restored
-on exit. Saving an d restoring has an overhead of about 70 cycles per 2 byte
+on exit. Saving and restoring has an overhead of about 70 cycles per 2 byte
 variable. It is easy to see, that - apart from the additional code that is
 needed to save and restore the values - you need to make heavy use of a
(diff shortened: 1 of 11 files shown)


oliverschmidt:
I agree with your Makefile change. However, I think that tolower() is used in the way it is supposed to be used. If it doesn't work as expected then we shouldn't avoid it but fix it.

doj:
on c64 the tolower(), toupper(), islower() and isupper() are broken. I've written a simple test program see https://gist.github.com/doj/74c11566bbea909bd28e61ea05016efd

But I don't really know on what character set these 4 functions are supposed to work in cc65. Is it ASCII? or the native character set of the machine?

I can create a bug report for the c64 target if you want. I'll submit the Makefile change as a separate pull request.

oliverschmidt:
> I'll submit the Makefile change as a separate pull request.

Thanks - I just merged it.

> But I don't really know on what character set these 4 functions are supposed to work in cc65. Is it ASCII? or the native character set of the machine?

From my POV of course the native character set of the machine. That's what that whole 'target specific character set' support of cc65 is all about.

> I can create a bug report for the c64 target if you want.

Please do so - referring your test program. Thanks, Oliver

PR #1022, opened 2020-04-06
Preserve the accu

polluks:


oliverschmidt:
You lost me here. Where's the benefit in preserving the accu when https://github.com/cc65/wiki/wiki/Parameter-passing-and-calling-conventions states that it isn't preserved and (as you know at least as well as I know) about every existing C library function doesn't preserve it?

mrdudz:
Nicer to use from asm code? :) And it's always good to see BIT used in real code =P

oliverschmidt:
> And it's always good to see BIT used in real code =P

Here you go: https://github.com/cc65/cc65/blob/master/libsrc/apple2/crt0.s#L130-L131 ;-)

In fact I use it always for "hitting" soft switches in the Apple II where the pure access already triggers the desired functionality - just as a means of "documenting" that it's the access what it's all about.

From that perspective I'm fine with merging this change. I just wanted to know if there's something else I'm not aware of...

polluks:
@oliverschmidt Indeed, it's just cosmetic.

PR #1025, opened 2020-04-15
PMG_SIZE_QUAD is 3, not 2.

itaych:
From "Mapping the Atari": "Size of player. POKE with zero or two for normal size (eight color clocks wide), POKE with one to double a player's width (sixteen color clocks wide), and POKE with three for quadruple width (32 color clocks wide). Each player can have its own width set."


dmsc:
A good reference on the register values is the Altirra Hardware Reference Manual, page 286: http://www.virtualdub.org/downloads/Altirra%20Hardware%20Reference%20Manual.pdf

PR #1026, opened 2020-04-22
Better //c support

polluks:


oliverschmidt:
* I must admit that I don't like the idea to always enable the 80 column firmware as it changes the behavior of the machine in quite some ways (e.g. CH/CV vs. OURCH/OURCV). Rather _videomode should iff called if 80 cols are requested.

* Are you positively sure that $C060 never has bit 7 set in a machine with TAPE IN? At least the Apple //e emulators at hand all tend to show $A0 when reading $C060 from the monitor...

oliverschmidt:
> Are you positively sure that $C060 never has bit 7 set in a machine with TAPE IN? At least the Apple //e emulators at hand all tend to show $A0 when reading $C060 from the monitor...

Ooops, I got this wrong: $C060 bit 7 set means 40 cols, right? So on a machine with TAPE IN it's 40 cols (as I'd expect).

polluks:
1. "always enable"? If you are in 80 cols and press the switch you will get 40 cols.
2. The [Technical Reference Manual](https://retrocomputing.stackexchange.com/questions/13977/what-do-the-80-40-and-keyboard-buttons-on-my-apple-iic-do/13989#13989)

We want to support buggy emulators? Come on :-)

oliverschmidt:
> I must admit that I don't like the idea to always enable the 80 column firmware as it changes the behavior of the machine in quite some ways (e.g. CH/CV vs. OURCH/OURCV). Rather _videomode should be called only iff 80 cols are requested.

Ooops, I just checked videomode() and it's does NOT activate the 80 column firmware if both the current mode and the desired mode are 40 cols. Pro: videomode() seems to do exactly what one would desire. Con: I implemented it myself in 2009 and forgot what it does ;-)

So the only issue left is that the size of every apple2enh binary is increased by the size of videomode() (44 bytes if I counted right) no matter if videomode() is called by the user. I'm not sure if the support of the 80/40 switch outweighs that...

oliverschmidt:
> We want to support buggy emulators? Come on :-)

Do we have a misunderstanding? I already corrected myself. The values provided by the emulators are rather correct. They look like the switch being in position '40'.

However, I'm still far from being convinced that the change is beneficial:

So far at least most of my programs stay the mode the display is in when entered from BASIC.SYSTEM. With your change they always use 40 cols on an enhanced //e (without the switch).

I think a program needs to first check if it indeed runs on a //c (and not on an enhanced //e or IIgs). Only after that it can use $C060 to set the video mode.

I think that //c check plus the videomode() code are too much overhead to be added to every program. I'd rather suggest to add "something" to https://github.com/cc65/cc65/blob/master/include/apple2enh.h#L85-L88 that allows the cc65 user to explicitly switch to the setting given by the //c hardware switch - and stay in the current mode if there's no hardware switch present. This also allows to make nice use of the videomode() feature to return to old mode. The program can switch back to whatever mode was present on startup - independently from the setting of the hardware switch.

From the link you provided I understand that BASIC.SYSTEM ignores the switch and might therefore very well be in 40 col mode although the switch is set to '80'. So running a cc65 program calling videomode(VIDEOMODE_AUTOCOL) (or alike) at the start and vieomode(oldmode) at the end would return back to BASIC.SYSTEM being in 40 col mode. And the output of the program would nevertheless still be visible (cropped to 40 cols) because of the "translation" feature of the 80 col firmware. That seems interesting to me...

oliverschmidt:
> "always enable"? If you are in 80 cols and press the switch you will get 40 cols.

Either another misunderstanding or a misconception on your side. When in 80 col mode and swtiching to 40 col mode, videomode() does NOT deactivate the 80 col firmware. It mere switches to 40 col display - which is a pretty different thing!

polluks:
Right, now you know why I squeezed a byte in videomode time ago.
Of course I would promote this feature in retrocomputing.stackexchange.com :-)

polluks:
Hmm, you mean like pressing ESC 4 and 8? I don't know.
At least I could add a //c check, of course no full get_ostype.

polluks:
@oliverschmidt Now //e are save again, ok? Let's meet a lonely switch its destiny!

polluks:
> So the only issue left is that the size of every apple2enh binary is increased by the size of videomode() (44 bytes if I counted right) no matter if videomode() is called by the user. I'm not sure if the support of the 80/40 switch outweighs that...

I know //c are a minority but it's the ONCE segment. Let's imagine...
Wozniak: I've made a nice switch.
Jobs: We ignore it.

oliverschmidt:
* I don't want to ignore the switch. Rather I think it's a great idea to add support for the switch to the apple2enh C library. But I question, if the support for the switch should be "implicit". I still think the support should rather be "explicit" - meaning that the developer wants to have it checked. There are quite some scenarios where the implicit solution doesn't work out.
* The videomode() code isn't in the ONCE segment.

polluks:
1. I know what you mean. Right now explicit is the status quo: zero support.
Please tell me one of your scenarios. In my opinion pressing the switch is like calling PR#0 before, just a shortcut. A program cannot assume anything. We are not talking about stdio which should never call videomode.
2. Okay, do you know if the access of SETALTCHAR and LORES is still necessary?

oliverschmidt:
> Right now explicit is the status quo: zero support.

I don't know what you want to express. Currently there's no support. My suggestion - as already mentioned further above - is to add support to reading the switch to videomode(). Then one could call videomode(VIDEOMODE_AUTOCOL) or videomode(VIDEOMODE_IICSWITCH) or something like this to switch to the mode set by the switch on a //c (and do nothing on all other machines). I consider this significant difference to the current situation.

> Please tell me one of your scenarios.

A cc65 program comes as .SYSTEM file. .SYSTEM files are always started in 40 col mode. The program creates some formatted output that only works in 40 col mode. Currently everything is fine. With your proposed change, the program author needs to add videomode(VIDEOMOD_40COL) to assert 40 col mode.

In general one of the differences to PR#3 / PR#0 is that those are reflected by BASIC.SYSTEM while (at least according to what the article you linked says) this isn't the case regarding the switch. Imagine e.g. BASIC.SYSTEM operating in 40 col mode and the switch being set to '80'. Then a cc65 program is run (not containing any call to videomode()). After its return BASIC.SYSTEM is in 80 col mode. I imagine very easily that this isn't desirable for - let's say enough - users. So I'd consider it obligatory that an "implicit" solution restores on program exit the videomode present on program startup. But if the developer explicitly calls videomode() then that setting needs to be kept after program exit because he might rely on the output being present and readable after exit. If that should not be case the developer can/should express this by calling videomode(oldmode) on program exit. etc. etc.

oliverschmidt:
It seems that we have different perspectives on this topic and it doesn't seem that you are interested in adjusting this PR. So it seems appropriate to close it - or do you want to veto?

polluks:
A little veto: I made an adjustment for non-//c, so 90% of all users should not see any difference at all.

oliverschmidt:
Unfortunately that doesn't touch my objections in https://github.com/cc65/cc65/pull/1026#issuecomment-618334541 at all (apart from the fact that machine detection is already part of the library and doesn't want to be repeated).

polluks:
My point of view is the C128:
- I'm in 40 cols, I `RUN "HELLO"`, it's a 40 cols Hello world
- I'm in 80 cols, I `RUN "HELLO"`, it's an 80 cols Hello world

Back to Apple's world:
- I type `pr#0` and `-hello.system` and except a 40 cols Hello world
- I type `pr#3` and `-hello.system` and except an 80 cols Hello world

Please tell me if I'm completely wrong...

I saw https://github.com/cc65/cc65/commit/7b2e4d0c7f606ce8f5d65eef430c26e6dc95b1e0 you mean, I should use the whole get_ostype?

oliverschmidt:
Now you lost me completely. What makes you believe that a cc65 program built for apple2enh doesn't behave as you describe?

polluks:
Your quote ".SYSTEM files are always started in 40 col mode".

oliverschmidt:
> Your statement ".SYSTEM files are always started in 40 col mode".

That statement meant was (too) short for ".SYSTEM files are always started in 40 col mode by the system (in contrast to by BASIC.SYSTEM)".

The scenario your referred to was covered by my statement "So far at least most of my programs stay the mode the display is in when entered from BASIC.SYSTEM." above.

polluks:
I see.
"BASIC.SYSTEM operating in 40 col mode and the switch being set to '80'. Then a cc65 program is run. After its return BASIC.SYSTEM is in 80 col mode. I imagine very easily that this isn't desirable for - let's say enough - users."
Indeed? I'd say: You want it, you got it!
Solution: Use the switch before or press Ctrl-Reset at the end.
Dear lazy users, in the past you were typing five key strokes `pr#3<CR>`, _this_ was laborious.
But you can always blame me because I was fighting against a meaningless decorative element.

oliverschmidt:
I guess/hop we agree that there's room for quite some different perspectives. Therefore I don't see any automatism, and therefore my proposal to support the switch "only" for programmers who actively want to support the switch (see my lengthy comments above).

So you're explicitly welcome to create a PR following an opt-in pattern in one way or another - and I'll close this one...

Issue #1030, opened 2020-05-08
ld65 --large-alignment


cacciatc:
ca65 supports the ```--large-alignment``` flag to disable warnings; however, ld65 does not, so one still gets warnings if using large alignments when it comes time to link.

cacciatc:
Support recently merged. Closing.

PR #1031, opened 2020-05-08
Added support for --large-alignment in ld65.

cacciatc:
Implemented the same way as in ca65. Addresses #1030

### DIFF
--- a/doc/ld65.sgml
+++ b/doc/ld65.sgml
@@ -303,7 +303,7 @@ Here is a description of all of the command-line options:
   <tag><tt>--large-alignment</tt></tag>
 
   Disable warnings about a large combined alignment. See the discussion of the
-  <tt><ref id=".ALIGN" name=".ALIGN"></tt> directive for futher information.
+  <tt/.ALIGN/ directive in the ca65 Users Guide for futher information.
 
 
   <tag><tt>--lib file</tt></tag>
--- a/doc/ld65.sgml
+++ b/doc/ld65.sgml
@@ -79,6 +79,7 @@ Long options:
   --end-group                   End a library group
   --force-import sym            Force an import of symbol 'sym'
   --help                        Help (this text)
+  --large-alignment             Don't warn about large alignments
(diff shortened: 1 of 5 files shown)


doc/ld65.sgml
```diff
@@ -298,6 +299,12 @@ Here is a description of all of the command-line options:
   information generation is currently being developed, so the format of the
   file and its contents are subject to change without further notice.
 
+  <label id="option--large-alignment">
+  <tag><tt>--large-alignment</tt></tag>
+
+  Disable warnings about a large combined alignment. See the discussion of the
+  <tt><ref id=".ALIGN" name=".ALIGN"></tt> directive for futher information.
```
greg-king5, line 306: That line refers to a label in a different document (`ca65.sgml`).  It can't be linked automatically.  You must replace it with a line such as this:

<tt/.ALIGN/ directive in the ca65 document, for further information.

cacciatc: Good feedback. Will make the change.

cacciatc: Removed link and changed text.

polluks:
Of course you may add `<htmlurl url="ca65.html" name="ca65.html">`.

greg-king5:
> Of course, you may add `<htmlurl url="ca65.html" name="ca65.html">`.

That would be confusing in the "info" and "text" versions of the documents.

polluks:
@greg-king5 Well, this is the calling convention in index.sgml. However we also have <url>...

oliverschmidt:
@greg-king5: I'm really bad at LinuxDoc. So I ask you to please take of this pull request.

greg-king5:
>Well, that is the calling convention in "index.sgml".

Unfortunately, LinuxDoc Tools has no direct support for adjusting file links for the various output formats.  I wanted "index.info" to have a proper selectable menu.  But, the best that I could achieve is more than three times bigger than the current "index.sgml"!  It has two copies of the index, and a lot of preprocessor tests.

PR #1035, opened 2020-05-26
Fix for #928.

### LINKED ISSUE #928: ld65: Multiple Defects with Overwrite
Overwrite mode does not work properly, with two problems.
1. With start=x, the file offset is not computed correctly. In cc65/src/ld65/bin.c line 198, the memory's base address is subtracted from the start address to get the memory offset, but the memory's base file offset is not subsequently added before fseeking.
2. When an overwrite occurs in the non-last memory, ld65 performs the overwrites between writing the relevant memory and the subsequent memory. When performing the overwrite, it seeks to where the overwrite should occur (which is wrong due to `#1`) and writes the overwrite. Then when it moves on to write the following memory it tries to write it immediately after, without ever seeking back to the end of the file, resulting in an internal error because the actual file offset is not the expected file offset (thrown at line 144).

A config file that exposes these bugs:
1. The overwrites occur in the $2000-$4000 file offset range, which is not at all where it should be
2. When it tries to write CHR00 it throws "ld65: Internal Error: Invalid file offset for memory area CHR00: 9862/255638"
[bmq.cfg.txt](https://github.com/cc65/cc65/files/3525056/bmq.cfg.txt)

EDIT: While working on a local kludge to fix the problem, I found the problem runs deeper than just those above. ld65 actually does not even correctly calculate subsequent segment file offsets when there's an overwrite segment in ProcessCfg, before it even gets to BinWriteMem.

laubzega:
Fixes #928.

### DIFF
--- a/src/ld65/bin.c
+++ b/src/ld65/bin.c
@@ -195,7 +195,7 @@ static void BinWriteMem (BinDesc* D, MemoryArea* M)
                 if (DoWrite || (M->Flags & MF_FILL) != 0) {
                     /* Seek in "overwrite" segments */
                     if (S->Flags & SF_OVERWRITE) {
-                        fseek (D->F, NewAddr - M->Start, SEEK_SET);
+                        fseek (D->F, NewAddr - M->Start + M->FileOffs, SEEK_SET);
                     } else {
                         WriteMult (D->F, M->FillVal, NewAddr-Addr);
                         PrintNumVal ("SF_OFFSET", NewAddr - Addr);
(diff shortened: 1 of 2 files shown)


src/ld65/config.c
```diff
@@ -2107,10 +2107,12 @@ unsigned CfgProcess (void)
...
 
             /* If this segment will go out to the file, or its place
-            ** in the file will be filled, then increase the file size.
+            ** in the file will be filled, then increase the file size,
+            ** unless it's an OVERWRITE segment.
             */
             if (S->Load == M &&
-                ((S->Flags & SF_BSS) == 0 || (M->Flags & MF_FILL) != 0)) {
+                ((S->Flags & SF_BSS) == 0 || (M->Flags & MF_FILL) != 0)
+                    && (S->Flags & SF_OVERWRITE) == 0) {
```
oliverschmidt, line 2115: Please format the multi-line expression like e.g. here: https://github.com/cc65/cc65/blob/master/src/ld65/config.c#L2013-L2014

src/ld65/bin.c
```diff
@@ -226,6 +226,12 @@ static void BinWriteMem (BinDesc* D, MemoryArea* M)
             unsigned long P = ftell (D->F);
             SegWrite (D->Filename, D->F, S->Seg, BinWriteExpr, D);
             PrintNumVal ("Wrote", (unsigned long) (ftell (D->F) - P));
+            /* If we have just written an OVERWRITE segement, move position to the
+             * end of file, so that subsequent segments are written in the correct
+             * place. */
```
oliverschmidt, line 231: Please format the multi-line comment like the other multi-line comments.

Issue #1040, opened 2020-05-31
cc65: Wrong arithmetics in non-byte pointer +=/-= byte-size offset?


acqn:
```c
#include <stdint.h>
#include <stdio.h>

typedef uint32_t * ptr_t;
typedef uint8_t idx_t;

ptr_t f(ptr_t p, idx_t i)
{
    return p + i;
}

ptr_t g(ptr_t p, idx_t i)
{
    return p = p + i;
}

ptr_t h(ptr_t p, idx_t i)
{
    return p += i; /* wrong result */
}

ptr_t ptr;

int main(void)
{
    printf("%d\n", f(ptr, 128) - ptr);
    printf("%d\n", g(ptr, 128) - ptr);
    printf("%d\n", h(ptr, 128) - ptr);

    return 0;
}
```

Results in:
```
128
128
0
```
Cause:
The addsubeq() in cc65/expr.c does a type conversion in the wrong way that it blindly clears the X register after scaling up the byte-size pointer offset that could put a non-zero value in the X register.
The fix is trivial and I've made it.

polluks:
Please make a PR.

polluks:
Please make a PR.

oliverschmidt:
Fixed by https://github.com/cc65/cc65/pull/1041.

Issue #1028, opened 2020-04-30
--local-strings is broken for arrays


EtchedPixels:
Simple demo
```c
char *foo[2] = { "one", "two" };
```
```
cc65 --local-strings /tmp/6.c
```
End result is this nonsense:
```asm
.segment	"DATA"

_foo:

.segment	"RODATA"

L0002:
	.byte	$6F,$6E,$65,$00
	.addr	L0002
L0004:
	.byte	$74,$77,$6F,$00
	.addr	L0004
```
If the string output put the segment back, it would help; but, the `const char *foo[2] = ..` case would still be broken.

Oddly, the code version works correctly and does remember to change segment back.

Not a big deal to me: I found it by accident whilst working on a 6800 C compiler based on chunks of CC65.  The fix, I think, is to make `UseLiteral()` ignore the localstring stuff; but instead, call `OutputLiteralPool()` at the end of each function or static/global initializer if localstrings are set.

oliverschmidt:
1. To me it seems that this issue isn't related to the array aspect of the demo. Rather it seems to be related to initializing a character pointer to an unnamed char string.

2. I don't understand what understand what '_the code version_' refers to.

EtchedPixels:
2. If you emit code after the string then the compiler correctly puts it together so 

func("foo");

works properly. Hence it only seems to show up on arrays.

oliverschmidt:
> code [...] func("foo");

I see.

> Hence it only seems to show up on arrays.

As already mentioned I can't see the relation to arrarys. As far as I see

`char *foo = "one";`

fails just as badly as your array example.

EtchedPixels:
Ok I didn't try that example

oliverschmidt:
Fixed by https://github.com/cc65/cc65/pull/1041.

Issue #919, opened 2019-06-28
Foward label reference in C code generates a word


andre-paris:
Take this code as example:
```c
void f(void){
    if (1) goto label;

label:
    while(1) {
    }	    
}
```
compiling it with `cc65 -t nes -Oirs test.c` generates this assembly code:
```asm
.segment	"CODE"

.proc	_f: near

.segment	"RODATA"

L0005:
	.word	$0000

.segment	"CODE"

L000C:	jmp     L000C

.endproc
```
look at the L0005, is it correct?
In several places in my games it's generates the .word into a "CODE" segment making a lot of BRK.
Tested with cc65 V2.18 - Git e34ee329

greg-king5:
Compile it without optimization.  You'll see the code that cc65 generates when a `goto` jumps into a nested code block which has automatic (stack) variables.

I suspect that the bug is a side-effect of cc65's computed-goto code.

oliverschmidt:
Yeah, it's likely a "regression" introduced by fixing the gotos.

MonteCarlos:
It looks like the compiler wants to allocate space for a non existant
variable.
Probably it interpretes the constant '1' as an int variable and then
accidentially emits the

L0005:
	.word	$0000
.

I went back to commit 55ce618bf2e64b2b4331a7f7a41223e3e6933492
and i can tell that the "bug" was already there.

Monte Carlos

Am 28.06.2019 um 20:42 schrieb André Paris de Oliveira:

andre-paris:
> Probably it interpretes the constant '1' as an int variable and then accidentially emits the

I don't this that's the cause, this code
```c
void f(void){
    goto label;

    label:
    	return;	    
}
```
compiles to: (-Oirs)
```asm
; ---------------------------------------------------------------
; void __near__ f (void)
; ---------------------------------------------------------------

.segment	"CODE"

.proc	_f: near

.segment	"RODATA"

L0003:
	.word	$0000

.segment	"CODE"

	rts

.endproc
```
w/o any optimizations:
```asm
; ---------------------------------------------------------------
; void __near__ f (void)
; ---------------------------------------------------------------

.segment	"CODE"

.proc	_f: near

.segment	"RODATA"

L0003:
	.word	$0000

.segment	"CODE"

	pha
	lda     L0003
	clc
	adc     sp
	sta     sp
	lda     L0003+1
	adc     sp+1
	sta     sp+1
	pla
	jmp     L0002
L0002:	jmp     L0001
[...]

Issue #420, opened 2017-04-13
Banked function-local data treated as code


clbr:
```c
#pragma rodata-name (push, "BANK")
#pragma code-name (push, "BANK")

unsigned char func(unsigned char val) {

        static const unsigned char data[] = {
                0, 1, 2
        };

        return data[val];
}

#pragma rodata-name (pop)
#pragma code-name (pop)
```

```asm
;
; File generated by cc65 v 2.16 - Git 9b243ff
;
        .fopt           compiler,"cc65 v 2.16 - Git 9b243ff"
        .setcpu         "6502"
        .smart          on
        .autoimport     on
        .case           on
        .debuginfo      off
        .importzp       sp, sreg, regsave, regbank
        .importzp       tmp1, tmp2, tmp3, tmp4, ptr1, ptr2, ptr3, ptr4
        .macpack        longbranch
        .export         _func

.segment        "RODATA"

.segment        "BANK"
.segment        "RODATA"

; ---------------------------------------------------------------
; unsigned char __near__ func (unsigned char)
; ---------------------------------------------------------------

.segment        "BANK"

.proc   _func: near

.segment        "BANK"

L0004:
        .byte   $00
        .byte   $01
        .byte   $02

.segment        "BANK"

        jsr     pusha
        ldy     #$00
        lda     (sp),y
        tay
        ldx     #$00
        lda     L0004,y
        jmp     incsp1

.endproc
```

It tries to execute the data.

bbbradsmith:
I don't think this is specific to banking. If you set your RODATA name to "CODE" you'd have the same problem, wouldn't you?
Shouldn't you require a separate BANK_RODATA segment and BANK_CODE segment?

clbr:
Using separate segments would be a valid workaround; however, it's
still a compiler bug to output the function label before the data, not
after.

oliverschmidt:
Fixed by https://github.com/cc65/cc65/pull/1041.

PR #1041, opened 2020-05-31
Several code generation fixes.

acqn:
Reported issues: #420, #919 (note below), #1028 and #1040.
Unreported issues: See commits messages.

Note about #919: This only fixes the bug that unused data could be left in the code segments (so more or less the same as #420). The unused data could be removed with future work which I'd consider an "optimization enhancement" rather than a "bug fix".

### DIFF
--- a/src/cc65/codegen.c
+++ b/src/cc65/codegen.c
@@ -1541,16 +1541,17 @@ void g_addlocal (unsigned flags, int offs)
 /* Add a local variable to ax */
 {
     unsigned L;
+    int NewOff;
 
     /* Correct the offset and check it */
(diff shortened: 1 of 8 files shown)


oliverschmidt:
Thanks for fixing quite some open bugs at once :-))

I verified that https://github.com/cc65/cc65/issues/420, https://github.com/cc65/cc65/issues/919, https://github.com/cc65/cc65/issues/1028 and https://github.com/cc65/cc65/issues/1040 are indeed fixed.

CI looks good too (https://travis-ci.org/github/cc65/cc65/builds/693209939) so I'm about to merge...

PR #1042, opened 2020-06-03
Add support for Apple II ProDOS version 2.5+ to Direct I/O (DIO)

bobbimanners:
ProDOS version 2.5 (currently in alpha testing) extends the number of drives that can be supported per slot from 2 to 8.

It does this by extending the device ID from the old `DSSS0000` format to an extended format `DSSS00DD` where the `D` represents drive and `S` represents the slot.  The two *least* significant bits of the device ID are the two *most* significant bytes of the drive number.

Modified cc65's `dio_open` and the `isdevice` support function to support this extended format.

My project [https://github.com/bobbimanners/ProDOS-Utils] needs this change to support drive numbers >2.

### DIFF
--- a/libsrc/apple2/dioopen.s
+++ b/libsrc/apple2/dioopen.s
@@ -9,6 +9,7 @@
 
         .include        "errno.inc"
         .include        "mli.inc"
+        .include        "zeropage.inc"
 
 _dio_open:
         ; Check for ProDOS 8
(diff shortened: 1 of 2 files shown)


oliverschmidt:
Hi,

First of all thanks for your interest in contributing to cc65 :-)

I'm following the ProDOS 2.4.x/2.5 development pretty closely. I'm not ready at this time to accept ProDOS 2.5 specific changes into cc65 upstream. ProDOS 2.5 development seems pretty slow. I could imagine both that

* it'll never be released for one reason or another
* incompatible changes like the 8 drives per slot get reverted before release

Therefore I only see the cc65 Apple C Library adjusted after ProDOS 2.5 got actually released - or I am informed by John that it's close to be released and that he'd like the cc65 Apple C Library to become compatible.

Apart from that

* I'd like to see the necessary adjustments done in little different way then you did
* I think the term "fixed" is misused here. When John deliberately decides to change public ProDOS data structures in an incompatible way then then adjustments to become compatible again are for sure no "fixes" as there never was something broken here.

PR #1055, opened 2020-06-20
Allow overlap of bit-field storage units

jmr:
Previously,
```c
struct s {
    unsigned int x : 10;
    unsigned int y : 10;
};
```
had `sizeof(struct s) == 4`.

With this change, allow the storage units of x and y to overlap, so `sizeof(struct s) == 3`, with y stored immediately after x, with no padding between them.

An int bit-field (the only type currently supported) will still never occupy more than two bytes.

### DIFF
--- a/src/cc65/declare.c
+++ b/src/cc65/declare.c
@@ -2092,7 +2092,11 @@ static unsigned ParseStructInit (Type* T, int AllowFlexibleMembers)
 
             /* Account for the data and output it if we have a full word */
             SI.ValBits += Entry->V.B.BitWidth;
-            /* Make sure unsigned is bit enough to hold the value, 30 bits. */
+            /* Make sure unsigned is big enough to hold the value, 30 bits.
+            ** This is 30 and not 32 bits because a 16-bit bit-field will
+            ** always be byte aligned, so will have padding before it.
+            ** 15 bits twice is the most we can have.
+            */
             CHECK (SI.ValBits <= CHAR_BIT * sizeof(SI.BitVal));
             CHECK (SI.ValBits < 2 * (INT_BITS - 1));
             if (SI.ValBits >= INT_BITS) {
--- a/src/cc65/declare.c
+++ b/src/cc65/declare.c
@@ -833,19 +833,16 @@ static SymEntry* ParseStructDecl (const char* Name)
 
             /* Add a field entry to the table */
             if (FieldWidth > 0) {
-                /* Add full byte from the bit offset to the variable offset.
-                ** This simplifies handling he bit-field as a char type
-                ** in expressions.
+                /* Full bytes have already been added to the StructSize,
+                ** which is passed to the offset of AddBitField.  BitOffs
+                ** is always within a char, which simplifies handling the
+                ** bit-field as a char type in expressions.
                 */
-                unsigned Offs = StructSize + (BitOffs / CHAR_BITS);
-                AddBitField (Decl.Ident, Offs, BitOffs % CHAR_BITS, FieldWidth);
+                CHECK (BitOffs < CHAR_BITS);
+                AddBitField (Decl.Ident, StructSize, BitOffs, FieldWidth);
                 BitOffs += FieldWidth;
                 CHECK (BitOffs <= INT_BITS);
-                /* The code above should either be rewritten to use these
-                ** values, or overlap of bit-field storage units should be
-                ** made into an option or controlled by something like
-                ** __attribute__((__packed__)).
-                */
+                /* Add any full bytes to the struct size. */
                 StructSize += BitOffs / CHAR_BITS;
                 BitOffs %= CHAR_BITS;
             } else {
--- a/src/cc65/declare.c
+++ b/src/cc65/declare.c
@@ -33,6 +33,7 @@
 
 
 
+#include <limits.h>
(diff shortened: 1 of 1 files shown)


src/cc65/declare.c
```diff
@@ -840,10 +841,13 @@ static SymEntry* ParseStructDecl (const char* Name)
...
-                    StructSize += SIZEOF_INT;
-                    BitOffs = 0;
-                }
+                /* The code above should either be rewritten to use these
+                ** values, or overlap of bit-field storage units should be
+                ** made into an option or controlled by something like
+                ** __attribute__((__packed__)).
+                */
+                StructSize += BitOffs / CHAR_BITS;
+                BitOffs %= CHAR_BITS;
```
jmr, line 850: I just wrote it like this to make the diff clearer. It could also be combined with the code above, but I'm wondering if the old and new behavior should be selectable by a command line flag or with `__atttribute__` or similar.

oliverschmidt: The 6502 doesn't benefit from alignment. Consequently cc65 always "packs" structs. In other words: Its struct layout is by default the same as with compilers for "usual" targets when called with `-fpack-struct`. From that perspective I don't see any point in some option here. cc65 should here as well just behave like compilers for "usual" targets when called with `-fpack-struct` - that's by far the most anticipation-compliant behavior. So please write the code in the way that makes most sense with that in mind.

src/cc65/declare.c
```diff
@@ -2079,8 +2092,10 @@ static unsigned ParseStructInit (Type* T, int AllowFlexibleMembers)
 
             /* Account for the data and output it if we have a full word */
             SI.ValBits += Entry->V.B.BitWidth;
-            CHECK (SI.ValBits <= INT_BITS);
-            if (SI.ValBits == INT_BITS) {
+            /* Make sure unsigned is bit enough to hold the value, 30 bits. */
```
greg-king5, line 2095: That comment should say,"... _big_ enough ...".

Why is the limit 30 instead of 32?

src/cc65/declare.c
jmr: Updated.

src/cc65/declare.c
jmr: Fixed and expanded comment.

src/cc65/declare.c
```diff
@@ -2079,8 +2089,14 @@ static unsigned ParseStructInit (Type* T, int AllowFlexibleMembers)
...
             /* Account for the data and output it if we have a full word */
             SI.ValBits += Entry->V.B.BitWidth;
-            CHECK (SI.ValBits <= INT_BITS);
-            if (SI.ValBits == INT_BITS) {
+            /* Make sure unsigned is big enough to hold the value, 30 bits.
+            ** This is 30 and not 32 bits because a 16-bit bit-field will
+            ** always be byte aligned, so will have padding before it.
+            ** 15 bits twice is the most we can have.
+            */
+            CHECK (SI.ValBits <= CHAR_BIT * sizeof(SI.BitVal));
```
jmr, line 2097: Let me know if assuming 32-bit `unsigned`s on the host platform is a problem.

src/cc65/declare.c
```diff
@@ -33,6 +33,7 @@
 
 
 
+#include <limits.h>
```
greg-king5, line 36: That header isn't needed.

src/cc65/declare.c
greg-king5, line 2097: `CHAR_BIT` should be plural: `CHAR_BITS`.

jmr: This `CHECK` is to make sure that we didn't overflow `BitVal` on the host platform. It's the host platform's `sizeof(BitVal)` and `CHAR_BIT` that are relevant here.

src/cc65/declare.c
jmr: It's for `CHAR_BIT` below.

src/cc65/declare.c
greg-king5: > This `CHECK` is to make sure that we didn't overflow `BitVal` on the host platform. It's the host platform's `sizeof(BitVal)` and `CHAR_BIT` that are relevant here.

Ah, I see.

oliverschmidt: It seem safe to consider this resolved...

oliverschmidt: Does that mean at least 32-bit unsigned (in contrast to 16-bit unsigned)?

oliverschmidt:
If my question above is answered positively and if the branch is rebased resolving the conflicts then it seems ready for merging - any objections?

src/cc65/declare.c
jmr: Yes, currently the host platform needs 32-bit `unsigned`. If that's not ok, the next easiest thing would be to use `uint32_fast_t` or `unsigned long` and assume that's 32-bits.

greg-king5:
> ... if the branch is rebased resolving the conflicts, then it seems ready for merging -- any objections?

I think that only the first commit has a conflict.  The second commit fixed it.  Github says that either "merging" or "squashing and merging" this PR will work.

oliverschmidt:
Thanks for pointing out :-)

PR #1058, opened 2020-06-21
Pad bit-fields only to the next byte

### LINKED ISSUE #1054: bit-fields inconsistently packed
This isn't necessarily a bug, but is a bit strange. Bit-fields are padded to 2 bytes when another field comes after them, but only one byte when the last field in a struct.
```c
struct s {
  unsigned int x : 4;
};

struct t {
  unsigned int x : 4;
  unsigned int y;
};

int sizeof_s(void) { return sizeof(struct s); }
int sizeof_t(void) { return sizeof(struct t); }
```
[`sizeof(struct s)` is 1, but `sizeof(struct t)` is 4, not 3.](https://godbolt.org/z/qJQobe)

jmr:
Fixes #1054.

Previously, a bit-field followed by a non-bit field was expanded to two bytes.  A bit-field ending the struct was (and continues to be) expanded to only a single byte.
```c
struct s {
  unsigned int x : 4;
};

struct t {
  unsigned int x : 4;
  unsigned int y;
};
```
Before: `sizeof (struct s) == 1`, `sizeof (struct t) == 4`

After: `sizeof (struct s) == 1`, `sizeof (struct t) == 3`

### DIFF
--- a/src/cc65/declare.c
+++ b/src/cc65/declare.c
@@ -772,7 +772,7 @@ static SymEntry* ParseStructDecl (const char* Name)
             if (BitOffs > 0) {
                 if (FieldWidth <= 0 || (BitOffs + FieldWidth) > INT_BITS) {
                     /* Bits needed to byte-align the next field. */
-                    unsigned PaddingBitWidth = -BitOffs % CHAR_BITS;
+                    unsigned PaddingBits = -BitOffs % CHAR_BITS;
 
                     /* We need an anonymous name */
                     AnonName (Ident, "bit-field");
@@ -780,10 +780,10 @@ static SymEntry* ParseStructDecl (const char* Name)
                     /* Add an anonymous bit-field that aligns to the next
                     ** byte.
                     */
-                    AddBitField (Ident, StructSize, BitOffs, PaddingBitWidth);
+                    AddBitField (Ident, StructSize, BitOffs, PaddingBits);
 
                     /* No bits left */
-                    StructSize += (BitOffs + PaddingBitWidth) / CHAR_BITS;
+                    StructSize += (BitOffs + PaddingBits) / CHAR_BITS;
                     BitOffs = 0;
                 }
             }
--- a/src/cc65/declare.c
+++ b/src/cc65/declare.c
@@ -771,17 +771,19 @@ static SymEntry* ParseStructDecl (const char* Name)
             */
             if (BitOffs > 0) {
                 if (FieldWidth <= 0 || (BitOffs + FieldWidth) > INT_BITS) {
+                    /* Bits needed to byte-align the next field. */
+                    unsigned PaddingBitWidth = -BitOffs % CHAR_BITS;
 
                     /* We need an anonymous name */
                     AnonName (Ident, "bit-field");
 
                     /* Add an anonymous bit-field that aligns to the next
-                    ** storage unit.
+                    ** byte.
                     */
-                    AddBitField (Ident, StructSize, BitOffs, INT_BITS - BitOffs);
+                    AddBitField (Ident, StructSize, BitOffs, PaddingBitWidth);
 
                     /* No bits left */
-                    StructSize += SIZEOF_INT;
+                    StructSize += (BitOffs + PaddingBitWidth) / CHAR_BITS;
                     BitOffs = 0;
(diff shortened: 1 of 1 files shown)


src/cc65/declare.c
```diff
@@ -771,17 +771,19 @@ static SymEntry* ParseStructDecl (const char* Name)
             */
             if (BitOffs > 0) {
                 if (FieldWidth <= 0 || (BitOffs + FieldWidth) > INT_BITS) {
+                    /* Bits needed to byte-align the next field. */
+                    unsigned PaddingBitWidth = -BitOffs % CHAR_BITS;
```
greg-king5, line 775: I think that it should be
```c
unsigned int PaddingBitWidth = CHAR_BITS - (BitOffs % CHAR_BITS);
```

jmr: If `BitOffs` is 8 or 16, we want 0 bits of padding, not 8.

jmr:
I'll also try to add some tests.

greg-king5:
Bit-field initializers still are stored as two bytes.  Do you want to fix that in this PR, or in a separate PR?

jmr:
I'll fix it here. Could you merge #1059 so I can send a follow-up that gets the `fields.c` test working?

jmr:
The tests I added in #1063 cause CHECK failures. I'll let you know when I have a new version with that fixed.

jmr:
> Bit-field initializers still are stored as two bytes. Do you want to fix that in this PR, or in a separate PR?

This is in #1066 instead of this PR.

oliverschmidt:
> > Bit-field initializers still are stored as two bytes. Do you want to fix that in this PR, or in a separate PR?

> This is in #1066 instead of this PR.

Sorry, I'm a bit lost here. I understand the statement above as https://github.com/cc65/cc65/pull/1066 fixing something regarding this PR (https://github.com/cc65/cc65/pull/1058).

But https://github.com/cc65/cc65/pull/1066 is already merged while this PR (https://github.com/cc65/cc65/pull/1058) is not.

@jmr @greg-king5: Can you please shed some light on the status of this PR?

jmr:
> > > Bit-field initializers still are stored as two bytes. Do you want to fix that in this PR, or in a separate PR?
> 
> > This is in #1066 instead of this PR.
> 
> Sorry, I'm a bit lost here. I understand the statement above as #1066 fixing something regarding this PR (#1058).

#1066 was fixing something related to this PR, but it was also safe to do before this one, so I did it separately.

> @jmr @greg-king5: Can you please shed some light on the status of this PR?

This is ready to be merged, but it would be better to merge #1063 first so there's more test coverage. This PR changes some sizes that are checked in #1063 , so I need to change something before the second one is merged.

oliverschmidt:
> it would be better to merge #1063 first so there's more test coverage.

Does that mean that bitfield.c (https://github.com/cc65/cc65/pull/1063) is supposed to fail before this PR (https://github.com/cc65/cc65/pull/1058) is merged?

jmr:
No. Either can be merged first, then the other will need to be updated.

oliverschmidt:
Now that https://github.com/cc65/cc65/issues/1071 is sorted out we can move forward again.

src/cc65/declare.c
acqn: This in MSVC generates [C2220](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-errors-1/compiler-error-c2220?view=vs-2019) + [C4146](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-2-c4146?view=vs-2019) and causes building to fail.

jmr: It's right and intended in this case. 

I can blindly follow https://stackoverflow.com/a/25447795 and do
```
#pragma warning(suppress: 4164)
```
but I can't test it.

Something silly like `(CHAR_BITS - BitOffs) % CHAR_BITS` might fix this warning, but I don't know if it causes another.

jmr: Sent #1079 for fix. Sorry about that.

oliverschmidt: Merged https://github.com/cc65/cc65/pull/1079.

PR #1059, opened 2020-06-23
switch2.c: Remove REFCC, REFCC_UNSIGNED_CHARS, and UNSIGNED_CHARS

jmr:
switch2.c uses these macros.  With them unset, signed chars
are not tested.

### DIFF
--- a/test/ref/Makefile
+++ b/test/ref/Makefile
@@ -54,7 +54,7 @@ $(WORKDIR):
 
 $(WORKDIR)/%.ref: %.c | $(WORKDIR)
 	$(if $(QUIET),echo ref/$*.host)
-	$(CC) $(CFLAGS) -o $(WORKDIR)/$*.host $< $(NULLERR)
+	$(CC) $(CFLAGS) -DREFCC -DREFCC_UNSIGNED_CHARS -o $(WORKDIR)/$*.host $< $(NULLERR)
 	$(WORKDIR)$S$*.host > $@
 
 $(DIFF): ../bdiff.c | $(WORKDIR)
--- a/test/ref/Makefile
+++ b/test/ref/Makefile
@@ -54,7 +54,7 @@ $(WORKDIR):
 
 $(WORKDIR)/%.ref: %.c | $(WORKDIR)
 	$(if $(QUIET),echo ref/$*.host)
-	$(CC) $(CFLAGS) -DREFCC -DREFCC_UNSIGNED_CHARS -o $(WORKDIR)/$*.host $< $(NULLERR)
+	$(CC) $(CFLAGS) -DREFCC -o $(WORKDIR)/$*.host $< $(NULLERR)
 	$(WORKDIR)$S$*.host > $@
 
 $(DIFF): ../bdiff.c | $(WORKDIR)
--- a/test/ref/Makefile
+++ b/test/ref/Makefile
@@ -54,7 +54,7 @@ $(WORKDIR):
 
 $(WORKDIR)/%.ref: %.c | $(WORKDIR)
 	$(if $(QUIET),echo ref/$*.host)
-	$(CC) $(CFLAGS) -DREFCC -o $(WORKDIR)/$*.host $< $(NULLERR)
+	$(CC) $(CFLAGS) -o $(WORKDIR)/$*.host $< $(NULLERR)
 	$(WORKDIR)$S$*.host > $@
 
 $(DIFF): ../bdiff.c | $(WORKDIR)
--- a/test/ref/switch2.c
+++ b/test/ref/switch2.c
@@ -37,20 +37,8 @@ void testlimits(int i) {
 }
 
 void testdefault1(unsigned char i) {
-/* we want a signed char */
-#ifdef REFCC
 
-signed char k;
-        
-#else
-        
-#ifdef UNSIGNED_CHARS
-signed char k;
-#else
-char k;
-#endif
-
-#endif
+        signed char k;
 
         for(;i<254;) {
                 k = i;
@@ -138,20 +126,8 @@ char k;
 }
 
 void testdefault2(unsigned char i) {
-/* we want a unsigned char */
-#ifdef REFCC
 
-unsigned char k;
-        
-#else
-        
-#ifdef UNSIGNED_CHARS
-char k;
-#else
-unsigned char k;
-#endif
-
-#endif
+        unsigned char k;
 
         for(;i<254;) {
                 k = i;
--- a/test/ref/switch2.c
+++ b/test/ref/switch2.c
@@ -40,11 +40,7 @@ void testdefault1(unsigned char i) {
 /* we want a signed char */
 #ifdef REFCC
 
-#ifdef REFCC_UNSIGNED_CHARS
 signed char k;
-#else
-char k;
-#endif
         
 #else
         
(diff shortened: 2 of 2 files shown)


mrdudz:
thats not exactly correct... -DREFCC is needed when compiling the host program that produces the reference output... however REFCC_UNSIGNED_CHARS is only needed when the host compiler uses unsigned chars (which gcc normally does not).

It would probably by better to remove the use of REFCC_UNSIGNED_CHARS in switch2.c alltogether and use explicit "unsigned char" and "signed char" for the host compiler

jmr:
[`-funsigned-chars` is set in `CFLAGS`](https://github.com/cc65/cc65/blob/master/test/ref/Makefile#L41), which are only used for the reference, but deleting `REFCC_UNSIGNED_CHARS` seems to be better.

jmr:
Ok, done.

mrdudz:
> -funsigned-chars is set in CFLAGS, which are only used for the reference

ooops, didnt see that... in that case its fine

> but deleting REFCC_UNSIGNED_CHARS seems to be better.

yes, indeed

mrdudz:
@oliverschmidt as a sidenote, why dont these pullrequests to the tests trigger the CI?

oliverschmidt:
Hi,

I must admit that I don't see the point in this PR. So switch2.c has this code:
```
#ifdef REFCC

#ifdef REFCC_UNSIGNED_CHARS
signed char k;
#else
char k;
#endif
        
#else
        
#ifdef UNSIGNED_CHARS
signed char k;
#else
char k;
#endif
```
What does the author of this code want to archive?

Why is there `REFCC`? Because he wants to handle the reference compiler differently. Why would one want to do that? Because the reference compiler needs different source code to do the same the the compiler under test does. Do we have this problem? No, we can (and actually already do) use `-funsigned-char` to make the reference compiler work the same same the compiler under test works. So there's no need to fiddle with `REFCC` at all.

Why is there `UNSIGNED_CHARS`? Because he wants the test to work with both types of char handling. Actually there are quite some macros of this kind in the test source code base. Therefore there's https://github.com/cc65/cc65/blob/master/test/ref/common.h. It already contains https://github.com/cc65/cc65/blob/master/test/ref/common.h#L21.

So from my POV the only thing to do is to replace https://github.com/cc65/cc65/blob/master/test/ref/switch2.c#L9 with https://github.com/cc65/cc65/blob/master/test/ref/switch.c#L7.

Or do I somehow completely miss the point here?

Oliver

mrdudz:
as i see it, REFCC must be defined for all testprograms when compiling them for the host/reference compiler. if you dont want that, better go through all tests that use it and make sure your host compiler does the right thing (the macro was there for a good reason afterall). REFCC_UNSIGNED_CHARS was originally defined in refconfig.h, i dont know why this was removed - but if the intend was to use compiler options instead, the tests should all be updated so they use the right thing and those macros should no more be used and their usage removed to make clear what happens (i'd just use those macros instead). having the macros in the tests and not use them, and then adjust things to fit by using compiler options is... a bit questionable. confusing in any case - as this ticket demonstrates.

and doh - not including common.h and then use these macros is... broken too. why was this done? :) that kind of thing makes the tests work by accident, not by how they were intended.

oliverschmidt:
Are you aware that REFCC is used in exactly three (!) test programs? It certainly isn't a general concept of the test suite.

> if you dont want that, better go through all tests that use it and make sure your host compiler does the right thing

I did.

> the macro was there for a good reason afterall

Hm, I thought I had covered that aspect with my previous comment ?!?

> REFCC_UNSIGNED_CHARS was originally defined in refconfig.h, i dont know why this was removed

I have no insight into the history of the test suite (anymore). I can only judge on the current state.

> having the macros in the tests and not use them, and then adjust things to fit by using compiler options is... a bit questionable.

I generally agree to that.

> confusing in any case - as this ticket demonstrates.

Definitively.

> that kind of thing makes the tests work by accident, not by how they were intended.

I fully agree. I'd say the most important thing is consistency - to minimize the risk of confusion (see above). I don't see this PR improving consistency. If you want to modify things in either way ("only macros" or "no macros") improving consistency I'd be happy about it.

mrdudz:
> Are you aware that REFCC is used in exactly three (!) test programs? It certainly isn't a general concept of the test suite.

It was once, and it only shows that it wasnt required in many tests. Anyway, it doesnt matter (anymore) - the test suite like it is in the cc65 tree now has little in common with what i started long ago, and some of those things don't really make sense in this context either (like support for different compilers).

I guess the way to go is to remove all of those REFCC, REFERENCE and REFCC_xxx things, anywhere. That means in fields.c, switch2.c, add3a.c, add3.c, cq241.c and then adjust the compiler options for the reference compiler. Currently none of these appear to be used in any makefiles, so i guess careful review of those programs followed by simple deletion of the respective code can be done.

Another thing would be making sure all tests that use macros from common.h does actually include it... not sure how to do this somewhat automatically, manually will be a bit tedious :)

oliverschmidt:
@mrdudz: I fully agree with everything you wrote in https://github.com/cc65/cc65/pull/1059#issuecomment-649110723.

> Another thing would be making sure all tests that use macros from common.h does actually include it... not sure how to do this somewhat automatically, manually will be a bit tedious :)

common.h already includes stdio.h and stdlib.h. Tests likely need (at least one of) them. If they aren't included directly in tests anymore (which can been proven by searching for them) then a missing inclusion of common.h would likely yield in build errors.

mrdudz:
> common.h already includes stdio.h and stdlib.h. Tests likely need (at least one of) them. If they aren't included directly in tests anymore (which can been proven by searching for them) then a missing inclusion of common.h would likely yield in build errors.

i don't understand what you are saying. what we need to find are tests that - like switch2.c - does not include common.h, but still uses the macros it defines. are you saying that generally all tests should include stdio.h and stdlib.h via common.h? not a good idea imho, only the tests that use the macros should be required to do this.

oliverschmidt:
> are you saying that generally all tests should include stdio.h and stdlib.h via common.h?

Yes.

> not a good idea imho, only the tests that use the macros should be required to do this.

If you think so - it was just an idea...

mrdudz:
Looks fine to me now :) Great that you will fix the rest!

oliverschmidt:
> I could move misc/fields.c, which needs REFCC, to ref/fields.c

It would have helped - at least me - to have this pointed out in the first place.

oliverschmidt:
> > I can fix other tests to include common.h where needed in another PR.

> Great that you will fix the rest!

I second that - or to put it in "other" words: I'd appreciate if you would improve the tests in (one of) the two aspects discussed here:

* Remove all of those REFCC, REFERENCE and REFCC_xxx things, anywhere. That means in fields.c, switch2.c, add3a.c, add3.c, cq241.c and then adjust the compiler options for the reference compiler. Currently none of these appear to be used in any makefiles, so a careful review of those programs followed by simple deletion of the respective code can be done.

* Make sure all tests that use macros from common.h does actually include it.

oliverschmidt:
> @oliverschmidt as a sidenote, why dont these pullrequests to the tests trigger the CI?

They do (e.g. https://travis-ci.org/github/cc65/cc65/builds/701926463) but the results don't show up here anymore: 
https://travis-ci.community/t/github-status-not-posted-on-commits-on-repositories-using-legacy-service-integration/7798

PR #1063, opened 2020-06-27
Add test of bit-field packing for #1054 and #1055

jmr:


jmr:
Updated for new behavior after #1058 merge, so this is ready to merge now.

PR #1066, opened 2020-06-28
Output bit-field data as chars instead of ints

jmr:
This prepares for #1058, which will pad bit-fields only to
the next byte, instead of the next `sizeof(int)` (two bytes).

`OutputBitFieldData` now outputs chars instead of ints, and
calls to this function loop until there is less than one byte
to output.  A final partial byte is written out with zero padding
as a final partial int was previously.

### DIFF
--- a/src/cc65/declare.c
+++ b/src/cc65/declare.c
@@ -2137,7 +2137,7 @@ NextMember:
     ConsumeRCurly ();
 
     /* If we have data from a bit-field left, output it now */
-    CHECK(SI.ValBits < CHAR_BITS);
+    CHECK (SI.ValBits < CHAR_BITS);
     OutputBitFieldData (&SI);
 
     /* If there are struct fields left, reserve additional storage */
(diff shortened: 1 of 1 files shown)


src/cc65/declare.c
```diff
@@ -2089,14 +2089,15 @@ static unsigned ParseStructInit (Type* T, int AllowFlexibleMembers)
 
             /* Account for the data and output it if we have a full word */
```
greg-king5, line 2090: The comment should say, "`have full bytes.`".

jmr: Done.

src/cc65/declare.c
```diff
@@ -2137,6 +2137,7 @@ static unsigned ParseStructInit (Type* T, int AllowFlexibleMembers)
     ConsumeRCurly ();
 
     /* If we have data from a bit-field left, output it now */
+    CHECK(SI.ValBits < CHAR_BITS);
```
greg-king5, line 2140: Style fix: a space between the macro-name and the parenthesis.

jmr: Done.

Issue #169, opened 2015-07-03
Inconsistent result on divide with a negative numerator and a constant or variable denominator


mrdudz:
When dividing a negative integer numerator by a constant number, or by a variable containing that same number, the compiler generates different results.

This sample code illustrates, and will erroneously print "Not Equal":

``` C
#include <stdio.h>
void main(void)
{
    int a = -1, b = 8;
    if ((a/8) == (a/b))
        printf("Equal\n");
    else
        printf("Not Equal\n");
}
```

Looking at the generated code, it's easy to see where the error lies. The compiler substituted a shift operation for the divide-by-a-constant. The shift operation doesn't yield the same answer as the divide if the numerator is negative.
For the case above, -1 >> 3 = -1 and -1 / 8 = 0.

``` assembly
;
; if ((a/8) == (a/b))
;
    ldy     #$03
    jsr     ldaxysp
    jsr     asrax3      ; use >> 3 to divide by 8
    jsr     pushax
    ldy     #$05
    jsr     ldaxysp
    jsr     pushax
    ldy     #$05
    jsr     ldaxysp
    jsr     tosdivax    ; use divide code to divide by 8
```

The optimizer switched to using asrax3 for the constant division; and, used tosdivax for the division by the variable, resulting in -1/8 not giving a consistent result, depending on whether the 8 is used as a literal or in a variable.

acqn:
Are there test cases for this? I came up with 3 different ways to fix this and couldn't decide which one should be finally taken.

mrdudz:
probably not... i can create some :) what else other than variants of the above program (using different types) would be useful?

acqn:
> 
> 
> probably not... i can create some :) what else other than variants of the above program (using different types) would be useful?

Perhaps also with various combinations of the data sizes (how big the number is, not just padding with the sign-bit) of the both the numerator and the denominator.

mrdudz:
something like this? http://hitmen.c02.at/temp/bug-div.zip (will add more...)

acqn:
> 
> 
> something like this? http://hitmen.c02.at/temp/bug-div.zip (will add more...)

Thank you! It looks good (and already helps), though I think typedef or the stdint.h stuff would be better for the reference compiler.

mrdudz:
hrm. i knew something is fishy. yeah i should rewrite it in decimal, the extra hex rule is just confusing :)

mrdudz:
[...]

PR #1070, opened 2020-07-03
[cc65] Fix and improvement of signed division with a constant denominator.

acqn:
This fixes Issue #169 and can optimize division with a negative power-of-2 constant denominator as well (as previously positive only) into shift according to the code size factor settings as below:
- positive power-of-2 denominator (> 1): code size factor >= 170.
- negative power-of-2 denominator (< -1): code size factor >= 200.

There are some new optimization steps (that are currently held back due to dependency on other changes) to improve both of the size and speed aspects.

### DIFF
--- a/src/cc65/codegen.c
+++ b/src/cc65/codegen.c
@@ -2628,24 +2628,34 @@ void g_div (unsigned flags, unsigned long val)
 
     unsigned DoShiftLabel, EndLabel;
 
-    /* -Val truncated to the correct size */
+    /* Deal with negative values as well as different sizes */
+    int Negation;
     unsigned long NegatedVal;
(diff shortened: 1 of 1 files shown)


mrdudz:
Perhaps add the tests to this PR?

Awesome work you are doing with the compiler, please don't stop! :)

oliverschmidt:
> Awesome work you are doing with the compiler, please don't stop! :)

Fullest possible ACK !

polluks:
Next release will be the best cc65 ever :-)

oliverschmidt:
Now that https://github.com/cc65/cc65/issues/1071 is sorted out we can move forward again.

Issue #796, opened 2018-11-13
CA65: References inside of a scope aren't known outside of that scope.


vrubleg:
I use ca65 quite often, and there is an inconvenience with using `.PROC`: some labels require to specify the root namespace explicitly, even when there is no ambiguity.

For example, this code compiles as expected:
```asm
JSR testfunc

.IFREF testfunc
testfunc:
	RTS
.ENDIF
```
But, this variant doesn't:
```asm
.PROC caller
	JSR testfunc
.ENDPROC

.IFREF testfunc
testfunc:
	RTS
.ENDIF
```
It tells `Error: Symbol 'testfunc' is undefined`, but when I call the `testfunc` as `JSR ::testfunc`, it works as expected.

Probably, it was designed this way. But, it is inconsistent, sometimes it is not required to specify the root namespace. For example, this code is assembled properly without explicit root namespace:
```asm
.PROC caller
	JSR testfunc
.ENDPROC

testfunc:
	RTS
```

Another variation of the issue. This code is assembled as expected:
```asm
DEBUGMODE = 1

.IF DEBUGMODE
	LDA #0
.ENDIF
```
But, this can't be assembled:
```asm
DEBUGMODE = 1

.PROC caller
	.IF DEBUGMODE
		LDA #0
	.ENDIF
.ENDPROC
```
It tells `Error: Constant expression expected`, but when I use DEBUGMODE as `.IF ::DEBUGMODE`, it starts to work as expected.

vrubleg:
Any news on this matter? Is it hard to fix?

vrubleg:
I use ca65 for my hobby project (for historical reasons: I started from cc65, and then switched to pure assembly). It is nice, and has quite powerful macros, but there are a few things which chould make using ca65 even more enjoyable, all of them are related to each other: #796, #1056, #1057. At this moment, I'm in the middle of other project, and can't switch to learning internals of ca65 and implementation of these enhancements. But if somebody knows internals of the assembler, likes these ideas, and wants to implement them, I would pay 50$ for each of the tickets (or 200$ if of them are implemented by the same person). Not much, but it's just a "thank you" for making ca65 better.

oliverschmidt:
https://sourceforge.net/p/cc65/mailman/message/37044269/

laubzega:
[...]

PR #1072, opened 2020-07-04
Fix for #796

laubzega:


oliverschmidt:
@laubzega: Please explain the status of this PR from your POV.

laubzega:
@oliverschmidt: Experimental code. May have to be scrapped in entirety or in part, depending on the evaluation by @vrubleg. Currently waiting for feedback.

vrubleg:
Sorry that I'm so late. I tested this code on one of my projects. In most cases it works fine, except such situation:

```
.PROC testref
	JSR oam_rescroll
.ENDPROC

.PROC test2
.IFREF oam_rescroll
	.ASCIIZ "teststr1"
.ENDIF
.ENDPROC

.IFREF oam_rescroll
.PROC oam_rescroll
	.ASCIIZ "teststr2"
.ENDPROC
.ENDIF
```

It is expected that `teststr1` is included into output, but there is `teststr2` only, without `teststr1`. Seems like this feature is hard to implement without making the assembler multi-pass.

I would provide FASMG as an example of an assembler where such tricks work without issues. For example, this code shows the same situation, and it works as expected (`testval2` is included into the final binary):

```
namespace testcaller
	dd testval
end namespace

namespace testifref
	if used testval
		testval2 dd 0xEE
	end if
end namespace

if used testval
	testval dd 0xFF
end if
```

mrdudz:
mmh. the given example seems plausible to me - but i cant judge if that fix makes sense or what else it could break

@kugelfuhr comment? :)

kugelfuhr:
I haven't exactly analyzed what the code changes do, but they are surprisingly simple for what they claim to do. And indeed a simple test shows serious problems.

Using this input which contains a forward reference inside a `.proc`:
```
.proc   proc
        lda     foo
        foo = 2
.endproc

.ifref  foo
        lda     #$00
.endif
```
the assembler prints `test.s(2): Warning: Didn't use zeropage addressing for 'foo'` and translates the input as follows:
```000000r 1               .proc   proc
000000r 1  AD 02 00             lda     foo
000003r 1                       foo = 2
000003r 1               .endproc
000003r 1
000003r 1               .ifref  foo
000003r 1                       lda     #$00
000003r 1               .endif
000003r 1
```
But with this patch, we get:
```
000000r 1               .proc   proc
000000r 1  AD 02 00             lda     foo
000003r 1                       foo = 2
000003r 1               .endproc
000003r 1
000003r 1               .ifref  foo
000003r 1  A9 00                lda     #$00
000005r 1               .endif
000005r 1
```
This means that the patch causes the assembler to somehow think that there's a symbol `foo` in the global namespace - but this is not true.

The whole reason for the problems with local scopes comes from the fact that any symbol used within a local scope may be a forward reference to a symbol in this or an embedded scope. Like `proc::foo` in the code above. The assembler cannot just reference a symbol with the same name in the global scope, since a local definition might follow. This is a larger problem buried deep in the design of the assembler and the reason why I said that the patch that claims to fix it is "surprisingly simple".

I don't know if it is possible to fix the PR. My gut feeling says that it needs more than a few lines of code changes.

mrdudz:
Sound like the proposed fix is too simple. So converting this to draft as well

vrubleg:
It requires a significant refactoring to make it working in all cases, it looks like it will require to make assembler multipass (like FASMG).

PR #1073, opened 2020-07-04
In Atari XEX output format, join memory areas if possible.

dmsc:
This makes executables shorter if two memory areas are consecutive.


src/ld65/xex.c
```diff
@@ -415,6 +413,8 @@ void XexWriteTarget (XexDesc* D, struct File* F)
             Write16 (D->F, 0x2E2);
             Write16 (D->F, 0x2E3);
             Write16 (D->F, GetExportVal (I->InitAd->Exp));
+            /* Always write a new segment header after a INITAD segment */
```
polluks, line 416: "an INITAD"

dmsc: Updated commit with this typo fixed 🙂

PR #1078, opened 2020-07-08
Tiny optimization for VIC20 TGI

ops:
Tiny optimization to save 5 bytes and some cycles.

### DIFF
--- a/libsrc/vic20/tgi/vic20-hi.s
+++ b/libsrc/vic20/tgi/vic20-hi.s
@@ -261,16 +261,15 @@ PATTERN_SOLID:
         sta     tmp2+1
         inx                     ; (ldx #$00)
         stx     ERROR           ; Set to TGI_ERR_OK
+        clc
 
 @NEXT_ROW:
         ldy     #$00
         txa
-        clc
         adc     #$10
 
 @NEXT_COLUMN:
         sta     (tmp2),y
-        clc
         adc     #ROWS
         iny
         cpy     #COLS
@@ -282,10 +281,8 @@ PATTERN_SOLID:
         clc
         adc     #COLS
         sta     tmp2
-        bcc     @L1
-        inc     tmp2+1
 
-@L1:    inx
+        inx
         cpx     #ROWS
         bne     @NEXT_ROW
 


oliverschmidt:
Can you please elaborate a bit? It seems to me that you want to say that there's some addition which will never actually overflow into the high byte, right? If yes: What is added and what is the maximum value that can be reached?

dmsc:
Hi!

> Can you please elaborate a bit? It seems to me that you want to say that there's some addition which will never actually overflow into the high byte, right? If yes: What is added and what is the maximum value that can be reached?

Looking at the code, there are only 20 columns and 12 rows, so the addition never overflows, and the CPX at the end clears the carry for the next iteration.

But using that fact, I think you could also remove the "CPY #COLS" just replacing the BNE with a BCC, as the addition will just overflow in the last iteration (starting with $10 to $1B, adding $0C 20 times, it gives $100 to $10B).

Then, you could also start with X = $10 at the top, remove the "ADC #$10" and simply compare at the end with "#ROWS+$10".

dmsc:
Again, this is the code I think would work:

```
.proc INIT

; Initialize variables

        ldy     #$FF
        sty     BITMASK

; Make screen columns.

        lda     #<SBASE
        sta     tmp2
        lda     #>SBASE
        sta     tmp2+1
        iny                     ; (ldy #$00)
        sty     ERROR           ; Set to TGI_ERR_OK
        clc
        ldx     #$10

@NEXT_ROW:
        txa

@NEXT_COLUMN:
        sta     (tmp2),y
        iny
        adc     #ROWS
        bcc     @NEXT_COLUMN

; Step to next row on screen.
        inx
        cpx     #ROWS
        bne     @NEXT_ROW
```

oliverschmidt:
Sorry guys but if you want me to merge such a PR you need to take me with you! Unfortunately I don't have the time budget to figure out things myself :-(

> Looking at the code, there are only 20 columns and 12 rows, so the addition never overflows

What I'd like to have read here is that 20 * 12 = 240 which is less then 256.

> [...] if the lower part of SBASE is less than $10.

What I'd like to have read here is that the SBASE low byte must be less than 16 as otherwise the addition of the above 240 will make the tmp2 pointer low byte overflow.

> But the removal of the "inc tmp2+1" only works if the lower part of SBASE is less than $10.

Okay, so what is known about the low byte of SBASE? Where does that knowledge come from?

Again sorry, but either you have the original commiter of the code (here @greg-king5) review your change or you have to be much more explicit about what's going on.

mrdudz:
SBASE is page aligned, so the lowbyte is always 0

PR #1079, opened 2020-07-08
Fix MSVC build broken by #1058

jmr:
MSVC complains about unary negation of unsigned, but it's
intended.  Suppress the warning.

https://github.com/cc65/cc65/pull/1058#discussion_r451757967

"Tested" with godbolt.org.

### DIFF
--- a/src/cc65/declare.c
+++ b/src/cc65/declare.c
@@ -772,7 +772,11 @@ static SymEntry* ParseStructDecl (const char* Name)
             */
             if (BitOffs > 0) {
                 if (FieldWidth <= 0 || (BitOffs + FieldWidth) > INT_BITS) {
-                    /* Bits needed to byte-align the next field. */
+                    /* Bits needed to byte-align the next field.  MSVC complains
+                    ** about unary negation of unsigned, but it's intended.
+                    ** Disable the warning for the next line only.
+                    */
+                    #pragma warning(disable: 4146)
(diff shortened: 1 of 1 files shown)


jmr:
@acqn

acqn:
I think you can cast BitOffs to int and cast the result of the negation back to unsigned int to avoid the warning without using #pragma.

jmr:
It's a bit surprising that the CI uses different settings than just typing `make`. 

Since this is just an overzealous MSVC warning, the pragma seems to be the most straightforward solution.  If a bunch of different compilers were complaining, I would definitely rewrite it.

If this way is strongly disliked, I could go with `(0 - BitOffs) % CHAR_BITS`.

oliverschmidt:
> It's a bit surprising that the CI uses different settings than just typing make.

The CI wants to make sure that the code base stays warning free with the very gcc used by the CI.

The make wants to generate binaries with as many different gcc variants (and similar compilers) as reasonable.

> If this way is strongly disliked, I could go with (0 - BitOffs) % CHAR_BITS.

I strongly prefer that approach over the pragma.

mrdudz:
> I strongly prefer that approach over the pragma.

+1 ... anything compiler specific should be avoided if possible. perhaps add a comment why it was done like this.

jmr:
Done.

oliverschmidt:
Thanks for adjusting the fix :-)

PR #1082, opened 2020-07-10
Make test/ slightly more make--j-friendly

jmr:
Move `.NOTPARALLEL` from `test/Makefile` to the subdir `Makefile`s that need it.

Add `.$1.$2` suffixes to some rules in `misc`.

There is still a problem with `cl65` intermediate files overwriting each other, which I'm working on separately.

On my eight-year-old computer, `cd test; make -j 5` runs in 4.5 minutes compared to 5.5 for `make`.


jmr:
@acqn

PR #1083, opened 2020-07-10
C64 soft80 conio: save 4 bytes and 6 cycles in `firstinit`

Compyx:
Since the copies of the charset are page-aligned, we can save 4 bytes by using `#ldy 0` to set the LSB of ptr2 and ptr3, and further down we can remove the `ldy #0` from the loop. Skipping the `ldy #0` in the loop allow us to save 6 cycles.
Not a huge improvement, but every little bit helps.

I initially thought the source charset was also page-aligned, but alas, I was wrong, got some 'interesting' results :)

### DIFF
--- a/libsrc/c64/soft80_conio.s
+++ b/libsrc/c64/soft80_conio.s
@@ -67,22 +67,21 @@ firstinit:
 
         inc     soft80_first_init
 
+        ; save 4 bytes due to soft80_lo_charset and soft80_hi_charset being
+        ; page-aligned.
+        ldy     #0
         lda     #<soft80_charset
         ldx     #>soft80_charset
         sta     ptr1
         stx     ptr1+1
-        lda     #<soft80_lo_charset
         ldx     #>soft80_lo_charset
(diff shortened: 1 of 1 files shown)


libsrc/c64/soft80_conio.s
```diff
@@ -67,22 +67,21 @@ firstinit:
 
         inc     soft80_first_init
 
+        ; save 4 bytes due to soft80_lo_charset and soft80_hi_charset being
+        ; page-aligned.
```
oliverschmidt, line 71: Sorry for the nit picking but please shorten this comment to 
; soft80_lo_charset and soft80_hi_charset are page-aligned

oliverschmidt:
Thanks for having a close look at existing code :-)

Compyx:
Portris, the 4-player Tetris clone posted on cc65-devel made me curious about how conio was implemented, and I enjoy optimizing assembly code. But I'm under no illusion I can noticeably speed op conio, at least not without a lot of speedcode :)

PR #1085, opened 2020-07-11
C64 soft80-conio cgetc: save 14 cycles in `invertcursor`

Compyx:
By 'inverting' the loop, we can save 16 cycles by removing the `cpy #8`, saving 16 cycles. But we need an extra `ldy #7` at the start of the loop, so the total cycles saved is 14. Code size doesn't increase due to the addition of the `ldy #7` negating ther removal of the `cpy #xx`.

### DIFF
--- a/libsrc/c64/soft80_cgetc.s
+++ b/libsrc/c64/soft80_cgetc.s
@@ -46,17 +46,16 @@ invertcursor:
         lda     #$34
         sta     $01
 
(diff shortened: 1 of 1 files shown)


oliverschmidt:
It seems to me that 'setcolor' could be inlined...

polluks:
A profiler would be nice *hint* *hint*

PR #1092, opened 2020-07-16
Fix info page building for newer linuxdoc

### LINKED ISSUE #1039: Building info-page for ca65 fails
I tried building the info pages on FreeBSD, using linuxdoc-tools-0.9.80 and texinfo-6.7_3,1. `makeinfo` fails on the output created by `linuxdoc` for `ca65.sgml` in the section 5.5 Available operators with all kinds of error messages about `@item` in a wrong context.

Filling all empty table cells in `ca65.sgml` with an explicit `&nbsp;` solves this problem, and the HTML output sill looks fine. But even then, the resulting info page is unreadable, it looks like this:

image

spiro-trikaliotis:
In ca65.sgml, the following pattern was used in tables in order to
create an empty row:

||~@

That is, the first two columns are empty, the last one has an &nbsp;

Unfortunately, with newer linuxdoc, this fails, as the empty columns
create two @item directly after each other, which is not allowed.

Changing this to

~|~|~@

fixes it by adding an "&nbsp;" into each column.

Furthermore, the last line had a "newrow" (@) separator, which created
an artifact. Removed that one.

This fixes #1039.

### DIFF
--- a/doc/ca65.sgml
+++ b/doc/ca65.sgml
@@ -622,7 +622,7 @@ problem in most cases.
 <tabular ca="clc">
 <bf/Operator/| <bf/Description/| <bf/Precedence/@<hline>
 | Built-in string functions| 0@
-||~@
+~|~|~@
 | Built-in pseudo-variables| 1@
 | Built-in pseudo-functions| 1@
 +| Unary positive| 1@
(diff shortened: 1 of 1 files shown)


spiro-trikaliotis:
BTW: This fix is compatible with both older and newer linuxdoc versions.

oliverschmidt:
Thanks :-)

Issue #1077, opened 2020-07-08
[cc65] Wrong code: a non-array constant address with a non-constant subscript


acqn:
So I find a codegen bug when a non-array constant address is accessed with a non-constant subscript when I was attempting to fix #1075. Example:
```c
#include <stdio.h>

int main(void)
{
    int a = 0;
    (&a)[a] = 42;  
    printf("%d", a);

    return a != 42;
}
```
Current output: out-of-range madness.
Expected output: 42

mrdudz:
Seems to be a Stackpointer misalignment, doing this only prints ONE value before it exits:

~~~
#include <stdio.h>

int test1(void)
{
    int a = 0;
    (&a)[a] = 42;  
    return a;
}
int test2(void)
{
    int a = 0;
    int b = 0;
    (&a)[b] = 42;  
    return a;
}

int main(void)
{
    int res, ret = 0;
    res = test1();
    printf("%d\n", res);
    if (res != 42) ret = 1;
    res = test2();
    printf("%d\n", res);
    if (res != 42) ret = 1;

    return ret;
} 
~~~

acqn:
This can be fixed along with #1075 as they share some common part of code logic. An independent fix only for this is possible but requires more work (-arounds).

oliverschmidt:
Thanks for the intermediate feedback!

oliverschmidt:
There's a test for this issue at https://github.com/cc65/cc65/blob/master/test/todo/bug1077.c

mrdudz:
testcode/strdup-test.c does also fail to compile with an error, this could be related perhaps?

$ cl65 -t sim6502 -o test.prg strdup-test.c 
strdup-test.c(27): Error: Incompatible pointer types at 'Unknown'
strdup-test.c(37): Error: Incompatible pointer types at 'Unknown'

~~~
extern unsigned _hfirst;        /* First free block in list */
...
    unsigned Count = 0;
    unsigned** P = (unsigned**) _hfirst;
    while (P) {
        ++Count;
        P = P[1]; <- this gives the error
    }
~~~

acqn:
It is unrelated to this one. The pointer types are indeed incompatible, although the diagnostic level (error instead of warning, well...) and message ('Unknown') may have issues.

mrdudz:
OK.

oliverschmidt:
Fixed with https://github.com/cc65/cc65/pull/1099.

Issue #1075, opened 2020-07-06
Internal compiler error


gardners:
I've managed to get an internal compiler error using the following sequence:

```
$ git clone https://github.com/gardners/raycaster.git
$ cd raycaster
$ git checkout dff28c33105dcf2848908488d03f0dd4fc35a908
$ make raycaster.prg 
git submodule init
git submodule update
cl65 --config src/tests/vicii.cfg -I src/mega65-libc/cc65/include -O -o raycaster.prg --mapfile raycaster.map RayCaster/RayCasterFixed.c main.c mazegen.c src/mega65-libc/cc65/src/*.c src/mega65-libc/cc65/src/*.s
RayCaster/RayCasterFixed.c(338): Warning: Unreachable code
RayCaster/RayCasterFixed.c(615): Warning: Converting pointer to integer without a cast
RayCaster/RayCasterFixed.c(625): Warning: Converting pointer to integer without a cast
RayCaster/RayCasterFixed.c(633): Warning: Converting pointer to integer without a cast
main.c(171): Warning: Condition is never true
main.c(172): Warning: Condition is never true
src/mega65-libc/cc65/src/conio.c(148): Warning: Control reaches end of non-void function
src/mega65-libc/cc65/src/conio.c(178): Warning: Result of comparison is constant
src/mega65-libc/cc65/src/conio.c(180): Warning: Result of comparison is constant
src/mega65-libc/cc65/src/conio.c(186): Warning: Parameter 'padding' is never used
src/mega65-libc/cc65/src/conio.c(224): Warning: 'g_curAttr' is defined but never used
src/mega65-libc/cc65/src/random.c(41): Internal compiler error:
Code generation messed up: StackPtr is 2, should be 0

Input: (*(unsigned char*)(0xD774))=range>>0;
cl65: Subprocess 'cc65' aborted by signal 6
Makefile:198: recipe for target 'raycaster.prg' failed
make: *** [raycaster.prg] Error 1
paul@F96NG92-L:~/Projects/mega65/raycaster$ cc65 --version
cc65 V2.18 - Git 17c55041
paul@F96NG92-L:~/Projects/mega65/raycaster$ 
```

I'm guessing it is because I am shifting down a 16 or 32 bit value and implicitly selecting the bottom byte to stash.  But this should presumably not confuse the code-generator like this.

acqn:
Minimal:
```c
long rhs;

int main(void)
{
[...]

PR #1099, opened 2020-07-17
[cc65] Fix for Issues #1075 and #1077

acqn:
This PR contains fixes for Issues #1075 and #1077. There are also unpushed fixes for other issues that depend on this.

Brief on what's the matter:

- The compiler didn't have any status flags on whether an expression is to be used dereferenced or directly. It did very carefully everywhere in the code to handle the two cases correctly by separating the code flows and/or looking at some other known conditions, but it still had missed some parts, which resulted in the two issues reported. This PR keeps a track on this to fix the problem.

- There were also some confusions in the code with regards to "lvalue" and "rvalue". It seems the code was originally to use the concepts "lvalue" (as in "**l**ocation **value**", to be used dereferenced) and "rvalue" (as in "**r**egister **value**", to be used directly) to deal with the issue above, but they were overloaded also to mean "mutable" by "lvalue" and "const" by "rvalue". This PR divides by the meanings, uses "indirection"/"reference" and "address"/"immediate numeric" when talking about the former matter, leaves the "lvalue" (as in "**l**eft **value**") and "rvalue" (as in "**r**ight **value**") terms to be used in C's sense and deduces "constant-ness" according to various conditions.

- This PR also adds some tiny facility/utility stuff to be usable for other pending/future changes. They are bundled here since they are closely related to those needed for resolving the issues above. However, some of them might be eventually unused.

EDIT: Tidy up.

### DIFF
--- a/src/cc65/declare.c
+++ b/src/cc65/declare.c
@@ -1780,7 +1780,7 @@ static void DefineData (ExprDesc* Expr)
 
         case E_LOC_ABS:
             /* Absolute numeric address */
-            g_defdata (CF_ABSOLUTE | TypeOf(Expr->Type) | CF_CONST, Expr->IVal, 0);
+            g_defdata (CF_ABSOLUTE | TypeOf (Expr->Type) | CF_CONST, Expr->IVal, 0);
             break;
 
         case E_LOC_GLOBAL:
--- a/src/cc65/declare.c
+++ b/src/cc65/declare.c
@@ -1773,9 +1773,14 @@ static void DefineData (ExprDesc* Expr)
 {
     switch (ED_GetLoc (Expr)) {
 
+        case E_LOC_NONE:
+            /* Immediate numeric value with no storage */
+            g_defdata (CF_IMM | TypeOf (Expr->Type) | CF_CONST, Expr->IVal, 0);
+            break;
+
         case E_LOC_ABS:
-            /* Absolute: numeric address or const */
-            g_defdata (TypeOf (Expr->Type) | CF_CONST, Expr->IVal, 0);
+            /* Absolute numeric address */
+            g_defdata (CF_ABSOLUTE | TypeOf(Expr->Type) | CF_CONST, Expr->IVal, 0);
             break;
 
         case E_LOC_GLOBAL:
--- a/src/cc65/expr.c
+++ b/src/cc65/expr.c
@@ -83,6 +83,7 @@ static unsigned GlobalModeFlags (const ExprDesc* Expr)
 /* Return the addressing mode flags for the given expression */
 {
     switch (ED_GetLoc (Expr)) {
+        case E_LOC_NONE:        return CF_IMM;
         case E_LOC_ABS:         return CF_ABSOLUTE;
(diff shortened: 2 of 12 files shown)


greg-king5:
"lvalue" and "rvalue" mean "left value" and "right value" of an assignment operator:
```c
lvalue += rvalue;
```
The idea is that an lvalue _must be_ an expression that can be used as a location where data can be stored.  An rvalue is any expresion that cannot be used as a location.

src/cc65/declare.c
```diff
@@ -1773,9 +1773,14 @@ static void DefineData (ExprDesc* Expr)
...
+        case E_LOC_NONE:
+            /* Immediate numeric value with no storage */
+            g_defdata (CF_IMM | TypeOf (Expr->Type) | CF_CONST, Expr->IVal, 0);
+            break;
+
         case E_LOC_ABS:
-            /* Absolute: numeric address or const */
-            g_defdata (TypeOf (Expr->Type) | CF_CONST, Expr->IVal, 0);
+            /* Absolute numeric address */
+            g_defdata (CF_ABSOLUTE | TypeOf(Expr->Type) | CF_CONST, Expr->IVal, 0);
```
oliverschmidt, line 1783: Please add a space after `TypeOf`.

src/cc65/expr.c
```diff
@@ -1015,25 +1015,21 @@ static void ArrayRef (ExprDesc* Expr)
...
                 }
-
-                /* Use the offset */
-                Expr->IVal = Subscript.IVal;
             }
 
+            /* Adjust the offset */
+            Expr->IVal += Subscript.IVal;
+
+
```
oliverschmidt, line 1032: Two blank lines?

acqn:
> 
> 
> "lvalue" and "rvalue" mean "left value" and "right value" of an assignment operator:
> 
> ```c
> lvalue += rvalue;
> ```
> 
> The idea is that an lvalue _must be_ an expression that can be used as a location where data can be stored (mutable). An rvalue is any expression that cannot be used as a location.

That's what I mean by "in C's sense" athough there could be a problem when here come "const variables"....

src/cc65/expr.c
acqn: One line removed in d869e3a3b7050be0393dae7d93bcbd5652397c53.

src/cc65/declare.c
acqn: Added in d869e3a3b7050be0393dae7d93bcbd5652397c53.

oliverschmidt:
>> The idea is that an lvalue must be an expression that can be used as a location where data can be stored (mutable). An rvalue is any expression that cannot be used as a location.

> That's what I mean by "in C's sense" athough there could be a problem when here come "const variables"....

So it seems there was no misunderstanding here :-)

mrdudz:
@acqn could you give me a hint on what this PR could have broken in regards to structs? i really have trouble with reproducing the misbehaviour i get in my project. the problem is somehow related to accessing struct members, but apparently the same code works right with some structs and breaks with others. args

PR #1102, opened 2020-07-18
[cc65] Passing structs/unions <= 4 bytes to/from functions

### LINKED ISSUE #264: cc65 fails to warn about a function returning struct
This program gives wrong results when built on cc65, regardless of -O. The function is correct, and it succeeds when built on gcc.

(I was looking for a faster div() alternative. Still need to try a modified stdlib div() that uses unsigned division instead.)

```
#include <nes.h>
#include <conio.h>
#include <stdio.h>
#include <stdlib.h>
#include <stdint.h>

typedef uint16_t u16;
typedef uint8_t u8;

typedef struct {
    u16 quot;
    u16 rem;
} udiv_t;

udiv_t div3(u16 in) {

    udiv_t u;
    u16 q = 0;

    while (in >= 300) {
        in -= 300;
        q += 100;
    }

    while (in >= 30) {
        in -= 30;
        q += 10;
    }

    while (in >= 3) {
        in -= 3;
        ++q;
    }

    u.quot = q;
    u.rem = in;

    return u;
}

int main() {

    u16 i;
    div_t d;
    udiv_t u;

    for (i = 1024; i; i--) {
        d = div(i, 3);
        u = div3(i);

        if (d.quot != u.quot || d.rem != u.rem) {
            cprintf("Mismatch at %u/3, div %u %u, div3 %u %u\n", i,
                d.quot, d.rem, u.quot, u.rem);
            break;
        }
    }

    while (1);

    return 0;
}
```

acqn:
Fixes #264.

PS: It's quite annoying that GitHub ignores the true order in which I push the commits and always re-arranges them in chronological order. To work around this I have to rebase and/or collapse commits a lot to get them in some sensible order, but that's error-prone and it has caught me a few times. Hopefully, this time nothing gets wrong.


src/cc65/assignment.c
```diff
@@ -54,6 +54,69 @@
...
+static int CopyStruct (ExprDesc* LExpr, ExprDesc* RExpr)
+/* Copy the struct represented by RExpr to the one represented by LExpr */
+{
+    /* If the size is that of a basic type (char, int, long), we will copy
+    ** the struct using the primary register, otherwise we use memcpy. In
+    ** the former case, push the address only if really needed.
+    */
+    const Type* ltype  = LExpr->Type;
+    const Type* stype  = GetReplacementType (ltype);
+    int         UseReg = (stype != LExpr->Type);
```
greg-king5, line 66: The source expression `(stype != ltype)` looks cleaner than the above expression.

src/cc65/datatype.c
```diff
@@ -198,6 +225,26 @@ Type* GetImplicitFuncType (void)
 
 
 
+const Type* GetReplacementType (const Type* SType)
```
greg-king5, line 228: I think that the function name should be more specific: `GetStructReplacementType ()`.

acqn: The function is also used for unions. Maybe `GetStructOrUnionReplacementType ()` ?

src/cc65/assignment.c
acqn: Done together with other changes.

src/cc65/datatype.c
acqn: There are many *Struct* functions that also work with unions, too. Maybe too many. I'll modify the comments only.

oliverschmidt:
Thanks - merging...

jmr:
How hard would it be to do structs > 4 bytes?

> PS: It's quite annoying that GitHub ignores the true order in which I push the commits and always re-arranges them in chronological order. 

I'm not really sure what this means, but I noticed your PRs are always on master, not a feature branch.

PR #1103, opened 2020-07-18
Move -Werror from .travis.yml to src/Makefile

jmr:
This prevents warnings from being missed and breaking the CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,11 +4,11 @@ install:
   - sudo apt-get update
   - sudo apt-get install linuxdoc-tools linuxdoc-tools-info binutils-mingw-w64-i686 gcc-mingw-w64-i686 sshpass
 script:
-  - make bin USER_CFLAGS=-Werror
+  - make bin
   - make lib QUIET=1
   - make -C test QUIET=1
   - make -C src clean
-  - make bin USER_CFLAGS=-Werror CROSS_COMPILE=i686-w64-mingw32-
+  - make bin CROSS_COMPILE=i686-w64-mingw32-
   - make doc zip
 after_success:
   - make -f Makefile.travis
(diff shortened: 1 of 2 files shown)


oliverschmidt:
It is done deliberately in the very way it is now: We don't want people building cc65 using other compilers to fail just because of a warning.

jmr:
> It is done deliberately in the very way it is now: We don't want people building cc65 using other compilers to fail just because of a warning.

`-Wextra -Wno-char-subscripts` is already in `CFLAGS` what other compiler supports those flags but will give extra warnings? Are you thinking of other versions of gcc/clang? It seems that someone using a different compiler will already have to change `CFLAGS`.

oliverschmidt:
* I am no compiler guru. My maybe naive assumption is, that other compilers might ignore unknown -W<...> options.

* Nobody complained about -Wextra -Wno-char-subscripts in src/Makefile so far but people DID complain about -Werror in src/Makefile.

mrdudz:
Nah, only compilers that actively try to be "gcc compatible" (like clang) will accept the same kind of options. Try them with watcom =)

Complaints about -Werrror usually come from people who either use a newer GCC than you do, or they use a different architecture where GCC produces warnings that you dont get. Oh and clang also produces different warnings :) It's the one single reason for why we dont use it in VICE :)

oliverschmidt:
@mrdudz: Thanks for the info.

So if there's a momentum to move -Wextra -Wno-char-subscripts from src/Makefile to .travis.yml I'm okay to merge it.

mrdudz:
I'd just leave it there... because in reality, who will NOT build cc65 with either GCC or clang?

polluks:
@mrdudz I was using SAS/C long time ago, but used my own makefiles :-)

PR #1106, opened 2020-07-19
[cc65] Disabled applying 'sizeof' to a bit-field

### LINKED ISSUE #1097: [cc65] Taking the size (with 'sizeof') of a bit-field should be forbidden
I've done that in a branch and it seems to be work right.

There is also 'offsetof()' to think about. IIRC, the standards says calling it on a bit field is undefined but other major compilers accept that with no warnings at all.

acqn:
Fixes #1097, and partly fixed conversions from bit-fields to integers.

156ada3f552bc3e8e16238e3b10eb582a18ba277 is a required fix in order to make `sizeof ((int)s.bitfield)` not an error.

~~Nothing is done about `offsetof (struct S, bitfield)`, as the standard says it's UB and any outcome is literally allowed.~~

Nothing was wrong with  `offsetof (struct S, bitfield)`. It is working as expected by rejecting bit-fields correctly.

### DIFF
--- a/src/cc65/typeconv.c
+++ b/src/cc65/typeconv.c
@@ -169,9 +169,7 @@ ExitPoint:
     ReplaceType (Expr, NewType);
 
     /* Bit-fields are converted to integers */
-    if (ED_IsBitField (Expr)) {
-        ED_DisBitField (Expr);
-    }
+    ED_DisBitField (Expr);
 }
 
 
--- a/src/cc65/typeconv.c
+++ b/src/cc65/typeconv.c
@@ -59,8 +59,8 @@ static void DoConversion (ExprDesc* Expr, const Type* NewType)
 /* Emit code to convert the given expression to a new type. */
 {
     Type*    OldType;
-    unsigned OldSize;
-    unsigned NewSize;
+    unsigned OldBits;
+    unsigned NewBits;
 
 
     /* Remember the old type */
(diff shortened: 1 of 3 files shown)


jmr:
> Nothing is done about `offsetof (struct S, bitfield)`, as the standard says it's UB and any outcome is literally allowed.

`offsetof(..., bitfield)` should be fixed when `&bitfield` is disallowed, right?

https://github.com/cc65/cc65/blob/master/include/stddef.h#L62

src/cc65/typeconv.c
```diff
@@ -167,6 +167,11 @@ static void DoConversion (ExprDesc* Expr, const Type* NewType)
 ExitPoint:
     /* The expression has always the new type */
     ReplaceType (Expr, NewType);
+
+    /* Bit-fields are converted to integers */
+    if (ED_IsBitField (Expr)) {
+        ED_DisBitField (Expr);
```
greg-king5, line 173: The test isn't needed; `ED_DisBitField()` changes nothing when the expression isn't a bit-field.

acqn:
> `offsetof(..., bitfield)` should be fixed when `&bitfield` is disallowed, right?
> 
> https://github.com/cc65/cc65/blob/master/include/stddef.h#L62

That's always some mystery to me. I was able to take address and use `offsetof()` on bit-fields without any compiler errors when I opened #1097, then all of a sudden the address part "becames fixed" after I fixed `sizeof()`, no `offsetof()` too. Perhaps something went wrong with y patch queues that I mixed in some now-obsolete work on the nested structs fix or the alternative approach to fix value catagory that somehow bypassed the check.

src/cc65/typeconv.c
acqn: Removed.

PR #1107, opened 2020-07-19
rand() use XOR to break up unwanted pair correlation

bbbradsmith:
This form of rand() cannot return the same value twice in a row.
Two additonal EOR instructions produce a more even distribution of successive pairs.
see comments on #951


LRFLEW:
I'm not entirely sure about this. I'm not sold either way on this (adding it or not). I'm not entirely clear what the original problem *was* or how much this improves it. I will say that it at least shouldn't be any more *incorrect*. If the internal state has a proper 2^32 period, than adding this XOR shouldn't change the distribution of the output.

This also solves the problem I had with `srand()` where the subsequent call to `rand()` always had the lower and upper bytes have the same low 7 bits. However, I still think that updating `srand()` to not zero-fill would probably be a good idea either way.

bbbradsmith:
The internal state has a 2^32 period, but in the short term output the sequential possibilities have some strong correlations. The XOR breaks up the correlation a bit for a cost of 1 instruction on each output byte.

For example, taking a sample of a random pairing, every time rand() outputs 12345, how many different possibilites does the next number have?

1. current rand(): 383
2. rand() with the proposed XORs: 10758
3. 32-bit LFSR for comparison: 32768

That's probably the easiest way to demonstrate what I was looking to improve.

bbbradsmith:
Hmm, both 0 and using A or X for the other 2 values had undesirable results. With 0 we get the problem LRFLEW noted where both bytes returned by rand() are the same, and with A or X if the number is low we get the same number back from the first rand(), which is also kinda weird. Switched to using EOR #$E5, which flips half the bits in A before storing that in the upper two bytes, which seems to sufficiently obfuscate the srand value.

bbbradsmith:
Giving srand() a single tail call to rand() seems very sufficient.

libsrc/common/rand.s
```diff
@@ -54,18 +54,17 @@ _rand:  clc
...
         stx     rand+1
-        lda     #0
-        sta     rand+2          ; Set MSW to zero
-        sta     rand+3
-        rts
-
-
+        sta     rand+2          ; argument << 16 is convenient fill for MSW
+        stx     rand+3
+        jmp     _rand           ; to sufficiently "shuffle" first rand() result
```
oliverschmidt, line 70: If _srand is placed before _rand then it can fall into _rand saving 3 bytes and cycles.

oliverschmidt:
Sorry guys but I can't really follow your arguments here - what I see is:

@bbbradsmith: "Two additonal EOR instructions produce a more even distribution of successive pairs."

@LRFLEW: "I will say that it at least shouldn't be any more incorrect."

We gain 3 bytes in _srand so spending 6 bytes in _rand seems acceptable. So I'm in general in favor of merging (after my comment on the JMP has been taken care of).

libsrc/common/rand.s
bbbradsmith: Ah, good idea.

Issue #1108, opened 2020-07-20
Problem with static locals?


IrgendwerA8:
Maybe I'm missing something (at least it's Monday), but the following code prints only "0" for "cx" while I think it shouldn't:

```
#include <stdio.h>

unsigned char x=0;

void
PrintVar(void)
{
	unsigned char cx = x+1;
	printf("cx: %d\n", cx);
}


void
main(void)
{
	do
	{
		++x;
		PrintVar();
	}
	while(1);
}
```

Executable created only with the "-Cl" option. Without it, it works ok.

mrdudz:
Did it work in earlier builds?

mrdudz:
@oliverschmidt how would we add this as a test case? it's not practical to add static locals to the options being tested for all tests, right? "misc" and special case rule?

mrdudz:
If you move the assignment into a seperate line, then it works as expected. I guess the compiler somehow gets confused by it not being constant.

IrgendwerA8:
> Did it work in earlier builds?

This is the commit, which breaks the code:

mrdudz:
@acqn HELP! :)

acqn:
I'll take a look.

mrdudz:
It's short for --static-locals ... all local variables turn into static variables (so they dont live in the local stack frame, which speeds up things a lot)

oliverschmidt:
> @oliverschmidt how would we add this as a test case? it's not practical to add static locals to the options being tested for all tests, right? "misc" and special case rule?

How about using #pragma static-locals ([push,] on|off) - https://cc65.github.io/doc/cc65.html#ss7.16 ?

acqn:
It takes more time to figure out how to push a branch to GitHub than to fix this....

oliverschmidt:
Maybe this is all totally obvious to you and/or maybe I'm missing the point, but just in case:

Git branches are super-cheap. So the general idea is to have a separate Git branch for everything intended to become a GitHub pull request. After creating the GitHub pull request further commits pushed to the respective Git branch show up automatically in the pull request. After the pull request is merged the respective Git branch is to be deleted.

acqn:
> 
> 
> > it's not practical to add static locals to the options being tested for all tests, right? "misc" and special case rule?
> 
> How about using #pragma static-locals ([push,] on|off) - https://cc65.github.io/doc/cc65.html#ss7.16 ?

[...]

PR #1109, opened 2020-07-20
[cc65] Fix for "auto" variables made "static" with the "-Cl" options.

acqn:
Fix for Issue #1108.


jmr:
Adding a test case such as a cleaned up version of the one from #1108 would probably be helpful.

oliverschmidt:
That's totally out of question.

Issue #975, opened 2019-11-11
Forward array reference fails to compile


clbr:
```c
#include <nes.h>
#include <conio.h>

static const unsigned char array[];

void main() {
        gotoxy(1, 1);
        cprintf("%u", array[0]);

        while (1);
}

static const unsigned char array[] = { 0, 1, 2 };
```

```
cl65 -t nes -Oisr -o foo.nes foo.c
Unresolved external 'L0000' referenced in:
  foo.s(43)
ld65: Error: 1 unresolved external(s) found - cannot create output file
```

```asm
;
; File generated by cc65 v 2.18 - Git 2f3955d
;
        .fopt           compiler,"cc65 v 2.18 - Git 2f3955d"
        .setcpu         "6502"
        .smart          on
        .autoimport     on
        .case           on
        .debuginfo      off
        .importzp       sp, sreg, regsave, regbank
        .importzp       tmp1, tmp2, tmp3, tmp4, ptr1, ptr2, ptr3, ptr4
        .macpack        longbranch
        .forceimport    __STARTUP__
        .import         _gotoxy
        .import         _cprintf
        .export         _main

.segment        "RODATA"

_array:
        .byte   $00
        .byte   $01
        .byte   $02
L0007:
        .byte   $25,$75,$00

; ---------------------------------------------------------------
; void __near__ main (void)
; ---------------------------------------------------------------

.segment        "CODE"

.proc   _main: near

.segment        "CODE"

        lda     #$01
        jsr     pusha
        jsr     _gotoxy
        lda     #<(L0007)
        ldx     #>(L0007)
        jsr     pushax
        lda     L0000
        jsr     pusha0
        ldy     #$04
        jsr     _cprintf
L0012:  jmp     L0012

.endproc
```

Now what is that L0000, did the compiler intend to replace it with the real reference and forgot?

dmsc:
A smaller test case:

```
static const unsigned char array[];
int main() {
    return array[0];
}
static const unsigned char array[] = { 0, 1, 2 };
```

The code compiles ok if you replace the first line with:
```
static const unsigned char array[3];
```
Problem is in expr.c, line 779:
https://github.com/cc65/cc65/blob/f75657d7e250268c5a4c444ecd8b2ac2ea7815db/src/cc65/expr.c#L779 

The code is:
```
                } else if ((Sym->Flags & SC_STATIC) == SC_STATIC) {
                    /* Static variable */
[...]

PR #1110, opened 2020-07-20
[cc65] Storage classes/types fix and #975

acqn:
The "storage classes flags" were buggy as they were not bitwise-exclusive but often checked simply with bitwise-and. It was pure coincidence that most of them (including `SC_TYPEDEF`) worked as expected. Nevertheless, they have to be fixed up in this PR before they could get in the way sooner or later.

Note: the "SC_**ESU_**" types are now used in a more "value-ish" way rather than as bit flags, so that they needn't be bitwise-exclusive to each other.

---

This PR also fixed Issue #975 by recognizing and tracking the incomplete (unknown-size) array declarations.

Note: not only should the current test case for #975 compile and work, but also the code piece below fail to compile and generate errors like commented:

```c
static const unsigned char array[3];                /* OK */
static const unsigned char array[] = { 0, 1, 2 };   /* OK - complete definition*/
static const unsigned char array[3];                /* OK */
static const unsigned char array[];                 /* OK */
static const unsigned char array[] = { 1, 2, 3 };   /* Error - redefinition */
static const unsigned char array[4];                /* Error - conflicting size */
```

### DIFF
--- a/src/cc65/declare.c
+++ b/src/cc65/declare.c
@@ -490,7 +490,7 @@ static void ParseEnumDecl (void)
         }
 
         /* Add an entry to the symbol table */
-        AddConstSym (Ident, type_int, SC_ENUMERATOR|SC_CONST, EnumVal++);
+        AddConstSym (Ident, type_int, SC_ENUMERATOR | SC_CONST, EnumVal++);
 
         /* Check for end of definition */
         if (CurTok.Tok != TOK_COMMA)
--- a/src/cc65/declare.c
+++ b/src/cc65/declare.c
@@ -490,7 +490,7 @@ static void ParseEnumDecl (void)
         }
 
         /* Add an entry to the symbol table */
-        AddConstSym (Ident, type_int, SC_ENUM, EnumVal++);
+        AddConstSym (Ident, type_int, SC_ENUM|SC_CONST, EnumVal++);
 
         /* Check for end of definition */
         if (CurTok.Tok != TOK_COMMA)
--- a/src/cc65/declare.c
+++ b/src/cc65/declare.c
@@ -490,7 +490,7 @@ static void ParseEnumDecl (void)
         }
 
         /* Add an entry to the symbol table */
-        AddConstSym (Ident, type_int, SC_ENUM|SC_CONST, EnumVal++);
+        AddConstSym (Ident, type_int, SC_ENUMERATOR|SC_CONST, EnumVal++);
 
         /* Check for end of definition */
         if (CurTok.Tok != TOK_COMMA)
--- a/src/cc65/symentry.h
+++ b/src/cc65/symentry.h
@@ -69,39 +69,42 @@ struct CodeEntry;
 
 
 /* Storage classes and flags */
-#define SC_AUTO         0x0001U         /* Auto variable */
-#define SC_REGISTER     0x0002U         /* Register variable */
-#define SC_STATIC       0x0004U         /* Static */
-#define SC_EXTERN       0x0008U         /* Extern linkage */
+#define SC_NONE         0x0000U         /* Nothing */
+#define SC_STRUCT       0x0001U         /* Struct */
+#define SC_UNION        0x0002U         /* Union */
+#define SC_TYPEDEF      0x0004U         /* A typedef */
+#define SC_ESUTYPEMASK  0x0007U         /* Mask for above types */
+#define SC_ENUM         0x0008U         /* An enum */
+#define SC_BITFIELD     0x0010U         /* A bit-field inside a struct or union */
+#define SC_TYPEMASK     0x001FU         /* Mask for above types */
 
-#define SC_ENUM         0x0030U         /* An enum */
 #define SC_CONST        0x0020U         /* A numeric constant with a type */
-#define SC_LABEL        0x0040U         /* A goto label */
+#define SC_LABEL        0x0040U         /* A goto code label */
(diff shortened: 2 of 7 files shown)


src/cc65/symentry.h
```diff
@@ -245,10 +248,10 @@ INLINE int SymIsRegVar (const SymEntry* Sym)
 /* Return true if the given entry is a register variable */
 /* ### HACK! Fix the ugly type flags! */
 {
-    return ((Sym->Flags & (SC_REGISTER|SC_TYPE)) == SC_REGISTER);
+    return ((Sym->Flags & (SC_REGISTER|SC_TYPEMASK)) == SC_REGISTER);
 }
 #else
-#  define SymIsRegVar(Sym)      (((Sym)->Flags & (SC_REGISTER|SC_TYPE)) == SC_REGISTER)
+#  define SymIsRegVar(Sym)  (((Sym)->Flags & (SC_REGISTER|SC_ESUTYPEMASK)) == SC_REGISTER)
```
greg-king5, line 254: The inline function and the macro must do the same test!

acqn: The macro version should use `SC_TYPEMASK` instead of `SC_ESUTYPEMASK`. Fixed.

src/cc65/symtab.c
```diff
@@ -821,7 +821,7 @@ SymEntry* AddLocalSym (const char* Name, const Type* T, unsigned Flags, int Offs
 
         /* Set the symbol attributes */
         Entry->Type = TypeDup (T);
-        if ((Flags & SC_AUTO) == SC_AUTO) {
+        if ((Flags & SC_AUTO) == SC_AUTO || (Flags & SC_TYPEMASK) == SC_TYPEDEF) {
```
greg-king5, line 824: Why should defining a variable with a typedef type mean that it should be put on the C stack?

acqn: It's not put on the stack by this. The code flow here is exactly how things got done prior to 19bee2befa00bde63f470419383c2982f9e2c45d, as `(SC_TYPEDEF & SC_AUTO) == SC_AUTO` was true back then. There's no harm to clear the `V.Offs` field for `typedef`s.

src/cc65/declare.c
```diff
@@ -490,7 +490,7 @@ static void ParseEnumDecl (void)
         }
 
         /* Add an entry to the symbol table */
-        AddConstSym (Ident, type_int, SC_ENUM, EnumVal++);
+        AddConstSym (Ident, type_int, SC_ENUM|SC_CONST, EnumVal++);
```
oliverschmidt, line 493: As far as I can see our style is to have blanks before and after binary operators like `|`. As you usually perfectly follow the given style it seems appropriate to ask for an adjustment here.

src/cc65/symentry.h
```diff
@@ -245,10 +246,10 @@ INLINE int SymIsRegVar (const SymEntry* Sym)
 /* Return true if the given entry is a register variable */
 /* ### HACK! Fix the ugly type flags! */
 {
-    return ((Sym->Flags & (SC_REGISTER|SC_TYPE)) == SC_REGISTER);
+    return ((Sym->Flags & (SC_REGISTER|SC_TYPEMASK)) == SC_REGISTER);
```
oliverschmidt, line 249: It seems there should have been blanks here before, shouldn't they?

src/cc65/symentry.h
```diff
@@ -245,10 +246,10 @@ INLINE int SymIsRegVar (const SymEntry* Sym)
...
 }
 #else
-#  define SymIsRegVar(Sym)      (((Sym)->Flags & (SC_REGISTER|SC_TYPE)) == SC_REGISTER)
+#  define SymIsRegVar(Sym)  (((Sym)->Flags & (SC_REGISTER|SC_TYPEMASK)) == SC_REGISTER)
```
oliverschmidt, line 252: It seems there should have been blanks here before, shouldn't they?

src/cc65/symtab.c
```diff
@@ -1109,7 +1109,7 @@ void EmitDebugInfo (void)
         }
         Sym = SymTab->SymHead;
         while (Sym) {
-            if ((Sym->Flags & (SC_CONST|SC_TYPE)) == 0) {
+            if ((Sym->Flags & (SC_CONST|SC_TYPEMASK)) == 0) {
```
oliverschmidt, line 1112: It seems there should have been blanks here before, shouldn't they?

acqn: Fixed.

src/cc65/symentry.h
acqn: Fixed.

src/cc65/symentry.h
acqn: Fixed.

src/cc65/declare.c
acqn: I chose to follow the "local style". Fixed anyways. (Same below.)

acqn:
So GitHub allowed me to go back in time and insert a push "before" the review.

src/cc65/declare.c
oliverschmidt: > I chose to follow the "local style". 

Which from my POV is in general the most professional approach!

> Fixed anyways.

I think that cc65 has in general such a style-consistent code base, that it makes sense here, so thanks :-)

oliverschmidt:
As far as I understand @greg-king5's comments were already taken care of, so merging...

Issue #1095, opened 2020-07-17
[cc65] Why are 'signed int' bit-fields unsigned in cc65?


acqn:
I suppose 'int' bit-fields may have implementation-defined signedness, but 'signed int' bit-fields should always be signed.

polluks:
Have a got a test case?

acqn:
A quick test case:
```c
#include <stdio.h>

struct S {
	signed int a : 3;
	signed int b : 3;
	signed int c : 3;
};

int main(void)
{
	struct S s = { -1, -1, -1 };

	printf("%d %d %d %d", s.a, s.b, s.c, (s.a > 0 || s.b > 0 || s.c > 0));
	return s.a > 0 || s.b > 0 || s.c > 0;
}

```

polluks:
By the way sim65 prints three spaces..?

jmr:
@acqn Is this one of the issues you're working on?

greg-king5:
Does cc65 even see a difference between `"int"` and `"signed int"`?

acqn:
@jmr No.

acqn:
@jmr Are you working on this? The changes in #1121 could impact this due to the changes of the "integer types".

jmr:
Sort of, but I didn't get very far yet since I was working on `static_assert` (#1124 ). I'll get back to this after #1121 is merged.

jmr:
I have the `signed int` bit-fields working. Now I just need to switch `int` back to `unsigned int`.

Issue #1093, opened 2020-07-17
[cc65] Underlying types of enums?


acqn:
Currently enums only use int as the underlying type. Should it be extended to other types if the value of its enumerators exceed the range of int?

jmr:
They have to be int.

https://port70.net/~nsz/c/c99/n1256.html#6.7.2.2

Not sure if other compilers provide this extension.

acqn:
Thanks for the standard reference.

The C99 standard says the constant-expression in
`enumeration-constant = constant-expression`
shall has a value representable as an int (https://port70.net/~nsz/c/c99/n1256.html#6.7.2.2p2), but the choice of the "_enumerated type_" (the _underlying type_ for variables of this enum type, if I get it correctly) is implementation-defined (https://port70.net/~nsz/c/c99/n1256.html#6.7.2.2p4), as long as the type is compatible with char or an signed/unsigned integer type and big enough to represent the values of all the members of the enumeration.

So an enum is allowed to use long as its _underlying type_, but the _constant-expression_ defining an _enumeration-constant_ can't be larger than what an int can represent. That's a weird limitation.

All compilers such as GCC, Clang, MSVC (and even CC65) that I've tested allow the _constant-expression_ to be larger than an int. Perhaps it's a popular extension.

jmr:
The idea is that you could use short or char if the values fit.

acqn:
```c
#include <stdio.h>

enum {
    N = -1,
    A = 2147483648,
    B,
    C = 65535
} a;

int main(void)
{
    printf("%u %u %u %u\n", sizeof(A), sizeof(C), sizeof(N), sizeof(a));
    printf("%ld, %d, %d", (long)A, B < 0, N < 0);
}
```
GCC output:
`
8 4 4 8
-2147483648, 0, 1
`
Clang output:
`
8 4 4 8
-2147483648, 0, 1
`
MSVC output:
`
4 4 4 4
-2147483648, 1, 1
`
GCC output:
`
2 2 2 2
-2147483648, 1, 1
`
While most of these can be considered natural, cc65's output doesn't seem quite right. How could sizeof(A) be 2? Perhaps it's another bug.

acqn:
> 
> 
> The idea is that you could use short or char if the values fit.

That would be the original intention for the standard to allow some size optimization. The various compiler implementations could be "exploiting" it to enhance enum.

jmr:
[...]

Issue #1050, opened 2020-06-19
[feature request] packed enums


jmr:
Because of the "use the shortest data type and preferably unsigned" recommendation, it would be nice if enums did this.

```
enum e { k };
int f(void) { return sizeof(enum e); }
```

[`sizeof(enum e)` is currently 2](https://godbolt.org/z/bmg8YY). 

[gcc and clang support `-fshort-enums`](https://godbolt.org/z/FwMERm) and `__attribute__((__packed__))` to make it 1.

acqn:
With PR #1121 merged, cc65 now always "pack" the enum variables by using the smallest integral underlying type preferring unsigned. Would we consider this resolved?

jmr:
I consider it fixed. I just added a test case in #1151 .

Issue #1048, opened 2020-06-17
enum names ignored


jmr:
The following code has [two errors](https://godbolt.org/z/XiY5iL): a redeclared enum type and an undeclared enum type:
```
enum e { x };
enum e { y };

int f() { return sizeof(enum undeclared); }
```

[No error or warning is currently issued.](https://godbolt.org/z/9dRMgA)

acqn:
Enum types didn't get registered the same way as structs/unions are.

Fixed in my working copy.

acqn:
Would we consider this resolved with PR #1121 merged?

PR #1121, opened 2020-07-26
[cc65] Fixes and extensions for enum types

acqn:
Resolved Issues:
- #1048, now rejected correctly at the two errors with:
```
Error: Multiple definition for enum 'e'
Error: Size of data type is unknown
```
- #1050, it now always uses the smallest type in size that can represent all members of the enum as the underlying type.
- #1093, supporting enumerators larger than `int` as an extension (still allowed but warned on under non-CC65 modes).
  - Possible enumerator constants types are `int`, `unsigned int`, `long` and `unsigned long`.
  - Possible underlying types of the enum are `unsigned char`, `signed char`, `unsigned int`, `int`, `unsigned long`, `long`. The smallest type in size that can represent all values of the enumerators of the enum, in preference from the left to the right will be chosen.
  - The `sizeof(enumerator)` bug is fixed too.
- Allows local enumerators/enums/enum types to be declared in functions' local scopes shadowing the global ones of the same types/names.
- Allows to use all integer types (including enums) of `int` size to declare bit-fields, although all of them will be treated as `unsigend int`. See also Issues #1047 and #1095.


oliverschmidt:
Travis CI says:

```
make[1]: Entering directory '/home/travis/build/cc65/cc65/test/val'
[...]
val/compare7.g.6502.prg
Error: Maximum number of cycles reached.
Makefile:61: recipe for target '../../testwrk/val/compare7.g.6502.prg' failed
make[1]: *** [../../testwrk/val/compare7.g.6502.prg] Error 126
```

acqn:
D*MN I hate git. The three patches will do.
[patches.zip](https://github.com/cc65/cc65/files/4977560/patches.zip)

acqn:
@

> 
> 
> Travis CI says:
> 
> ```
> make[1]: Entering directory '/home/travis/build/cc65/cc65/test/val'
> [...]
> val/compare7.g.6502.prg
> Error: Maximum number of cycles reached.
> Makefile:61: recipe for target '../../testwrk/val/compare7.g.6502.prg' failed
> make[1]: *** [../../testwrk/val/compare7.g.6502.prg] Error 126
> ```

I merge (rebased) the wrong commits into it. If you take a look at the original 3 pushes a few hours ago before this PR, they were just the same in content as the 3 of now.

acqn:
You may have been aware that there are multiple things using the term "type" in cc65, including but not limited to:
- Data/expression classes: `T_CLASS_` TypeCode flags
- Data/expression "basic types" (including ""raw, "unqulified" and "underlying"):  `T_TYPE_` TypeCode flags
- Data/expression "types" (including ""raw, "unqulified" and "underlying"):  `T_` TypeCode values
- Data/expression "type struct strings"): `type_` arrays of Type structs
- Storage class "ESU" types and "non-ESU" types: `SC_` flags
- Code generation type flags: `CF_` flags

That's unfortunate.

oliverschmidt:
> there are multiple things using the term "type" in cc65

If you are interested in refactoring the code in order to improve the situation I'm in general willing to merge PRs.

acqn:
> 
> 
> > there are multiple things using the term "type" in cc65
> 
> If you are interested in refactoring the code in order to improve the situation I'm in general willing to merge PRs.

No such a plan so far.

oliverschmidt:
I see. So if anybody else wants to help here, he's welcome...

src/cc65/datatype.c
```diff
@@ -245,6 +274,49 @@ const Type* GetStructReplacementType (const Type* SType)
 
 
 
+long GetIntegerTypeMin (const Type* Type)
+/* Get the smallest possible value of the integer type */
+{
+    if (IsSignSigned (Type)) {
+        return (long)(0xFFFFFFFF << (CHAR_BITS * SizeOf (Type) - 1U));
```
oliverschmidt, line 281: Maybe it's just me being too stupid, but it seems to me that this line deserves a comment how it works.

Apart from that I don't understand why it's not appropriate to use `LONG_MAX` here - given that `long` is used. Maybe that would be worth added to a comment as well.

src/cc65/declare.c
```diff
@@ -483,21 +555,116 @@ static void ParseEnumDecl (void)
...
+        } else if (SizeOf (NewType) < SizeOf (type_int)) {
+            /* Integer constants are not shorter than int */
+            NewType = type_int;
+        }
+
+        /* Warn if the incremented value exceeds the range of the previous
+        ** type.
+        */
+        if (IsIncremented   &&
+            EnumVal >= 0    &&
```
oliverschmidt, line 616: Is this type of aligning of operators to a tab stop(?) a style seen in cc65?

acqn: Yeah. I saw that frequently in the optimizer.... It's aligning to the logic AND/OR to the same tab stop "grids" according to their "levels".

src/cc65/datatype.c
acqn: The min possible signed value of the byte size N is `- pow (2, 8*N-1)`. I'll add the comment.

And I am not sure why you refer to `LONG_MAX` here, given that this funtion is used for calcaulating the `MIN` rather than the `MAX` for an arbitray integer type....

acqn: If you were questioning about why the check for `unsigned long` overflow doesn't simply compare the enumerator value to `ULONG_MAX` before the incrementation... well, I did that in earlier revisions but turned to the current method as I'd like just to warn when the "overflow has already happened", although that may just be unconvincing.

src/cc65/declare.c
```diff
@@ -517,21 +684,31 @@ static int ParseFieldWidth (Declaration* Decl)
     /* Read the width */
     NextToken ();
     ConstAbsIntExpr (hie1, &Expr);
+    if (!IsClassInt (Decl->Type)) {
+        /* Only integer types may be used for bit-fields */
+        Error ("Bit-field has invalid type ''");
```
jmr, line 689: It would be nice to print the type while you're changing this.

src/cc65/declare.c
```diff
@@ -517,21 +684,31 @@ static int ParseFieldWidth (Declaration* Decl)
...
+    /* Use is when we really support non-uint16_t bit-fields */
+    if (Expr.IVal > (long)(SizeOf (Decl->Type) * CHAR_BITS)) {
         Error ("Width of bit-field exceeds its type");
         return -1;
     }
-    if (Expr.IVal == 0 && Decl->Ident[0] != '\0') {
-        Error ("Zero width for named bit-field");
+#else
+    /* This is what we currenty do */
+    if (Expr.IVal > (long)(SizeOf (type_uint) * CHAR_BITS)) {
```
jmr, line 705: You changed this from `int` to `long`, but this will always fit in an `int`.

src/cc65/datatype.c
oliverschmidt: Sorry for being too unprecise. In fact I was only wondering if `0xFFFFFFFF` couldn't be replaced by something from `limits.h`.

acqn: > 
> 
> Sorry for being too unprecise. In fact I was only wondering if `0xFFFFFFFF` couldn't be replaced by something from `limits.h`.

The number used here is supposed to be -1, but due to quirks with signed integer shift I decided to use explicit hex literal (which is always unsigned) for it. It could be changed to `(long)((unsigned long)(-1L) << (CHAR_BITS * SizeOf (Type) - 1U))` if that's considered more appropriate.

src/cc65/declare.c
acqn: Strange mistake. Fixed.

src/cc65/declare.c
acqn: `Expr.IVal` is `long` while `SizeOf()` returns `unsigned int` (not `int`) which would make the whole right operand of the comaparison `unsigned int`. So at least one of the operands have to be converted to be the same as the other for the comparison, but neither of them is `int`.

jmr:
> Allows to use all integer types (including enums) to declare bit-fields, although all of them will be treated as` unsigend` int.

This seems a little dangerous. So `long` bit-fields will be accepted but treated as `unsigned int`? Why not change this part separately, later, after the support for non-`unsigned` bit-fields is really there?

acqn:
> 
> 
> > Allows to use all integer types (including enums) to declare bit-fields, although all of them will be treated as` unsigend` int.
> 
> This seems a little dangerous. So `long` bit-fields will be accepted but treated as `unsigned int`? Why not change this part separately, later, after the support for non-`unsigned` bit-fields is really there?

Non-int sized bit-field declarators are disabled now. (a6622cde5c604355b210f9f1d8c92be6f0fc0bf6)

src/cc65/declare.c
```diff
@@ -689,6 +689,13 @@ static int ParseFieldWidth (Declaration* Decl)
         Error ("Bit-field has invalid type '%s'", GetBasicTypeName (Decl->Type));
         return -1;
     }
+
+    if (SizeOf (Decl->Type) != SizeOf (type_uint)) {
+        /* Only int sized types may be used for bit-fields for now */
+        Error ("CC65 currently only supports unsigned int bit-fields");
```
oliverschmidt, line 695: The name of this project is 'cc65', not 'CC65'.

src/cc65/datatype.c
oliverschmidt: Thanks, at least I understand it better now :-)

src/cc65/declare.c
acqn: Will fix later.

jmr: 8eab280 already fixed it.

src/cc65/declare.c
```diff
@@ -517,21 +684,38 @@ static int ParseFieldWidth (Declaration* Decl)
...
+
     if (Expr.IVal < 0) {
         Error ("Negative width in bit-field");
         return -1;
     }
-    if (Expr.IVal > (int) INT_BITS) {
+    /* FIXME: We should compare with the width of the specified type */
+#if 0
+    /* Use is when we really support non-uint16_t bit-fields */
+    if (Expr.IVal > (long)(SizeOf (Decl->Type) * CHAR_BITS)) {
```
jmr, line 706: @acqn As part of the fix for #1095, I have the changes to pass the actual bit-field type around instead of hardcoding uint in various places. That should be ready in a day or two, but let me know if I should split out those changes so you can start working with them.

acqn: @jmr I am not working on #1047/#1134 as I realized that they are affected by #1095 for their signed part. Would you like to work on them? Anyways please feel free to modify the code here.

jmr: Yes, I'll work on char/long after signed is working.

Issue #1114, opened 2020-07-21
[cc65] A code example of multiple "multiple definition"-handling issues


acqn:
```c
void f(void) {}
void f(int);
void f(void);
void f(void) {}
int f(void) { return 0; } /* Too many errors here */

void g(void);
void g(int a) {}
int g(void) { return 0; } /* Too many errors here */

void h();
void h(int a);
void h(int a) {}
void h(void) {}
int h(int a) { return 0; } /* Too many errors here */

void i() {}
void i(int a);
void i(int a) {}

struct S {
    int a;
} y;

struct S {
    int a;
} x;                      /* It would be better if the error were reported two lines earlier */

typedef enum E {
    A
} E;

typedef int I;
typedef unsigned int I;
typedef signed int I;
typedef signed I;         /* BUG: This should NOT cause an error */

int I;
int f;

int main(void)
{
    extern int B;
    enum {
        A,                /* BUG: This should NOT cause an error */
        B                 /* BUG: This should cause an error */
    } a;
    int A;                /* BUG: This should cause an error */
    typedef int A;

E:                        /* BUG: This is wrongly interpreted as a declarator */
a:
A:
    f();
    return 0;
}
```
There are multiple issues exposed by this example. I'll update the Issue report later.

EDIT: added comments on the part where CC65 currenlty does wrong or not good enough. The inability to duplicate `typedef`'s could be a big one out of these.

acqn:
All of the `BUG`s here are fixed in my workig copy (depending on the enum type fix).

oliverschmidt:
Fixed with https://github.com/cc65/cc65/pull/1133.

Issue #1113, opened 2020-07-21
[cc65] Compiler crashes when calling functions "redefined" as other types


acqn:
```c
void f() {}

int f;

int main(void)
{
    f();
    return 0;
}
```
Output:
```
1113.c(3): Error: Conflicting types for 'f'
1113.c(3): Error: Global variable 'f' already was defined in the '' segment.
*crash*
```
It has the cause: caller of `AddGlobalSym()` tampers its return value when the symbol to add has already existed, which is similar to that of Issue #728.

I fixed/worked around this in my working copy by returning a "bogus" symbol entry to "deceive" the caller when it fails to add a symbol due to redefinition errors. It could also help to work around https://github.com/cc65/cc65/issues/728#issuecomment-426887640 better without eagerly aborting compiling.

acqn:
Fixed in my working copy (unfortunately with dependency on the enum type fix as part of the overhauled checking mechanism)

oliverschmidt:
Fixed with https://github.com/cc65/cc65/pull/1133.

PR #1133, opened 2020-07-29
[cc65] Fixes for multiple "redefinition" problems and labels with the same identiers as typedefs

acqn:
Resolved Issues:
- #1113, no more crash.
- #1114, no more missing/spurious errors on the "redefinitions (true or false)". Also the labels are correctly recognized now.
- Generally more informative error messages about "redefinitions", though they could be improved even better.

Example `1114a.c`:
```c
typedef int UOP;
typedef int UOP;

void f(void);

struct S {
    int a;
} y;

struct S {
    int a;
} x;                      /* TODO: It would be better if the error were reported two lines earlier */

typedef enum E {
    A
} E;

typedef int I;
typedef unsigned int I;
typedef signed int I;
typedef signed I;         /* FIXED: This should NOT cause an error */

int I;
int f;

int main(void)
{
    extern int B;
    enum {
        A,                /* FIXED: This should NOT cause an error */
        B                 /* FIXED: This should cause an error */
    } a;
    int A;                /* FIXED: This should cause an error */
    typedef int A;

E:                        /* FIXED: This was wrongly interpreted as an enum declarator */
a:
A:
    f();                  /* FIXED: This should NOT crash the compiler */
    return 0;
}

/* TODO: the warning messages about unused symbols should tell the kind of the symbols */

/* TODO: in general, the diagnostic messages should point to the exact locations of the symbols */
```

Old compiler output (just aborted before crashing as seen in #1113):
```
1114.c(2): Error: Multiple definition for 'UOP'
1114.c(12): Error: Multiple definition for 'S'
1114.c(19): Error: Multiple definition for 'I'
1114.c(20): Error: Multiple definition for 'I'
1114.c(21): Error: Multiple definition for 'I'
1114.c(23): Error: Multiple definition for 'I'
1114.c(24): Error: Conflicting types for 'f'
1114.c(24): Error: Global variable 'f' already was defined in the '' segment.
1114.c(30): Error: Multiple definition for 'A'
1114.c(34): Error: Multiple definition for 'A'
1114.c(36): Error: Identifier expected
1114.c(36): Fatal: Too many errors
```

New compiler output:
```
1114.c(12): Error: Multiple definition for 'S'
1114.c(19): Error: Conflicting types for typedef 'I'
1114.c(23): Error: Redefinition of typedef 'I' as different kind of symbol
1114.c(24): Error: Redefinition of function 'f' as different kind of symbol
1114.c(32): Error: Symbol 'B' is already different kind
1114.c(33): Error: Redefinition of 'A' as different kind of symbol
1114.c(34): Error: Redefinition of 'A' as different kind of symbol
1114.c(41): Warning: 'a' is defined but never used
1114.c(41): Warning: 'E' is defined but never used
1114.c(41): Warning: 'a' is defined but never used
1114.c(41): Warning: 'A' is defined but never used
```

### DIFF
--- a/src/cc65/declare.c
+++ b/src/cc65/declare.c
@@ -1256,11 +1256,23 @@ static void ParseTypeSpec (DeclSpec* D, long Default, TypeCode Qualifiers)
             break;
 
         case TOK_IDENT:
-            Entry = FindSym (CurTok.Ident);
-            if (Entry && SymIsTypeDef (Entry)) {
-                /* It's a typedef */
-                NextToken ();
-                TypeCopy (D->Type, Entry->Type);
+            /* This could be a label */
+            if (NextTok.Tok != TOK_COLON) {
+                Entry = FindSym (CurTok.Ident);
+                if (Entry && SymIsTypeDef (Entry)) {
+                    /* It's a typedef */
+                    NextToken ();
+                    TypeCopy (D->Type, Entry->Type);
+                    break;
+                }
+            } else {
+                /* This is a label. Use the default type flag to end the loop
+                ** in DeclareLocals. The type code used here doesn't matter as
+                ** long as it has no qualifiers.
+                */
+                D->Flags |= DS_DEF_TYPE;
+                D->Type[0].C = T_QUAL_NONE;
+                D->Type[1].C = T_END;
                 break;
             }
             /* FALL THROUGH */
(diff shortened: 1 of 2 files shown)


acqn:
`1114b.c`:
```c
void f(void) {}
void f(int);
void f(void) {}
int f(void) { return 0; }

void g(void);
void g(int a) {}
int g(void) { return 0; }

void h();
void h(int a) {}
void h(void) {}
int h(void) { return 0; }

void i() {}
void i(int a);

int f;

int main(void)
{
    f();
    return 0;
}
```
Note: this is currently affected by a bug (#1135).

PR #1135, opened 2020-07-29
[cc65] Fixed type mask usage

acqn:
As it's said (https://github.com/cc65/cc65/pull/1121#issuecomment-664000221), the type system is unfortunately complicated, and I'll add in that it's error-prone. This bug was found with a test intended for #1114/#1133 (only the relevent part here, full at https://github.com/cc65/cc65/pull/1133#issuecomment-665628371):

```c
void f(void) {}
void f(int);    /* Should fail */
```

### DIFF
--- a/src/cc65/typecmp.c
+++ b/src/cc65/typecmp.c
@@ -215,8 +215,8 @@ static void DoCompare (const Type* lhs, const Type* rhs, typecmp_t* Result)
         }
 
         /* Get the left and right types, signs and qualifiers */
-        LeftType  = GetUnderlyingTypeCode (lhs);
-        RightType = GetUnderlyingTypeCode (rhs);
+        LeftType  = (GetUnderlyingTypeCode (lhs) & T_MASK_TYPE);
+        RightType = (GetUnderlyingTypeCode (rhs) & T_MASK_TYPE);
         LeftSign  = GetSignedness (lhs);
         RightSign = GetSignedness (rhs);
         LeftQual  = GetQualifier (lhs);


oliverschmidt:
@jmr, @acqn: Sorry guys, I lost track of what you two are doing - and my time budget is way too tight to try change that. May I ask you to review each others PRs?

jmr:
> ```c
> void f(void) {}
> void f(int);    /* Should fail */
> ```

So this currently compiles, but will correctly fail to compile with this PR?

acqn:
> 
> 
> > ```c
> > void f(void) {}
> > void f(int);    /* Should fail */
> > ```
> 
> So this currently compiles, but will correctly fail to compile with this PR?

Exactly.

src/cc65/typecmp.c
```diff
@@ -215,8 +215,8 @@ static void DoCompare (const Type* lhs, const Type* rhs, typecmp_t* Result)
         }
 
         /* Get the left and right types, signs and qualifiers */
-        LeftType  = GetUnderlyingTypeCode (lhs);
-        RightType = GetUnderlyingTypeCode (rhs);
+        LeftType  = (GetUnderlyingTypeCode (lhs) & T_MASK_TYPE);
```
jmr, line 218: Can you get the bug to trigger on other cases? It seems like signedness should trigger this a lot, but I definitely don't understand what's going on.

jmr: It would be good to have more tests, because we're returning `TC_INCOMPATIBLE` less often.

Looks fine up to my limited knowledge, though.

acqn: It was not a design error or logical flaw like most other issues we've found. It was a "careless oops":
> I forgot that the "type"s to check here is not the same kind as the `T_<TYPE>` types that are frequently used elsewhere, but instead the more stripped-down `T_TYPE_<TYPE>` types, despite they are all of the `TypeCode` (AKA `unsigned long`) types.

You can consider this as a typical "messing up with the wrong kind of integer flags" mistake in C programming. That's what's going on here.

I haven't come up with a different way to expose the same bug, but I'm sure that would be redundant and the existing test code (the 2 lines in the OP) would be enough. Seriously, in my honest opinion, such kind of clearly known "oops" errors is not worth the trouble making "full-coverage regression case tests" _dedicated to reproducing the very same of them_. The right tool to prevent them from emerging again is ~~stronger type system in the language,~~ code reviews and unit test.

With regards to test cases, it would be more useful to test with more combinations of various code patterns. to reveal _unknown/unidentified_ problems.

Just my 2 cents.

oliverschmidt:
@jmr: Thanks!

Issue #1117, opened 2020-07-23
Bit Fields Don't Work


sci4me:
So I was trying to create a struct to represent a 65(c)22 VIA, and it worked fine until I tried to make bit fields for the individual DDR and OR/IR bits. When I assign any of the "single-bit" values, the byte they are contained within gets set to 1, no matter which bit I tried to set.

Here's the struct definition I came up with:
```c
typedef struct {
	union {
		struct {
			volatile u16 prb0 : 1;
			volatile u16 prb1 : 1;
			volatile u16 prb2 : 1;
			volatile u16 prb3 : 1;
			volatile u16 prb4 : 1;
			volatile u16 prb5 : 1;
			volatile u16 prb6 : 1;
			volatile u16 prb7 : 1;
		};
		volatile u8 prb;
	};
	union {
		struct {
			volatile u16 pra0 : 1;
			volatile u16 pra1 : 1;
			volatile u16 pra2 : 1;
			volatile u16 pra3 : 1;
			volatile u16 pra4 : 1;
			volatile u16 pra5 : 1;
			volatile u16 pra6 : 1;
			volatile u16 pra7 : 1;
		};
		volatile u8 pra;
	};
	union {
		struct {
			volatile u16 ddrb0 : 1;
			volatile u16 ddrb1 : 1;
			volatile u16 ddrb2 : 1;
			volatile u16 ddrb3 : 1;
			volatile u16 ddrb4 : 1;
			volatile u16 ddrb5 : 1;
			volatile u16 ddrb6 : 1;
			volatile u16 ddrb7 : 1;
		};
		volatile u8 ddrb;
	};
	union {
		struct {
			volatile u16 ddra0 : 1;
			volatile u16 ddra1 : 1;
			volatile u16 ddra2 : 1;
			volatile u16 ddra3 : 1;
			volatile u16 ddra4 : 1;
			volatile u16 ddra5 : 1;
			volatile u16 ddra6 : 1;
			volatile u16 ddra7 : 1;
		};
		volatile u8 ddra;
	};
	volatile u8 t1cl;
	volatile u8 t1ch;
	volatile u8 t1ll;
	volatile u8 t1lh;
	volatile u8 t2cl;
	volatile u8 t2ch;
	volatile u8 sr;
	volatile u8 acr;
	volatile u8 pcr;
	volatile u8 ifr;
	volatile u8 ier;
	volatile u8 _pra;
} __VIA;
```

[...]

PR #1138, opened 2020-07-30
Use char ops if possible for bit-field loads

jmr:
Set `CF_FORCECHAR` and change type to `char` once we have
shifted into a `char`.

This saves some unnecessary `ldx #0` instructions.

### DIFF
--- a/src/cc65/loadexpr.c
+++ b/src/cc65/loadexpr.c
@@ -193,10 +193,26 @@ void LoadExpr (unsigned Flags, struct ExprDesc* Expr)
         ** operations.
         */
         if (ED_IsBitField (Expr)) {
-            unsigned F = CF_INT | CF_UNSIGNED | CF_CONST | (Flags & CF_TEST);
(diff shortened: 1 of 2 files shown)


jmr:
@acqn

acqn:
Is this an optimization or a preparation for a solution of the remaining bit-field issues?

jmr:
Just an optimization that I noticed when implementing signed bit-fields.

acqn:
The removal of the unused CF_TEST flag actually makes me wonder if we have been handling direct tests with bit-fields correctly:
```c
struct {
    int a : 5;
} s;

if (s.a) ... /* I think this should be the case */
```
But that's a different question....

jmr:
Yes, I wondered the same thing. `CF_TEST` doesn't get removed, though. It still comes in through `Flags`.

I thought we had tests for #1117 , but we don't. I'll add some tests like that.

acqn:
I have no problem with this PR.

Issue #1139, opened 2020-07-30
Bad codegen with bitfield in if condition


jmr:
After our uncertainty with `CF_TEST` in #1138 , I wrote some tests and found a failure.

```
static struct overlap {
    unsigned int x : 10;
    unsigned int y : 10;
} o = {11, 22};

/* Test using bit-fields in if conditions. */
static void test_if(void)
{
    o.x = 0;
    o.y = 44;
    if (o.x) {
        printf("Bad, o.x is false\n");
        failures++;
    } else {
        printf("Good\n");
    }
}
```

I have to look in more detail, but the load looks strange. It seems like it should be `ldx` not `ora`.
```
;
; if (o.x) {
;
        lda     _o
        ora     _o+1
        tay
        txa
        and     #$03
        tax
        tya
        jeq     L000B
```

I will add a test/todo.

@acqn

acqn:
https://github.com/cc65/cc65/blob/2d5fd0fc6301ab0d0ee4cb3517f9e0c9dce540a5/src/cc65/loadexpr.c#L122-L124
This should only be done for non-bit-field loads.

What's going on here is that, when the `CF_TEST` flag is set, the `g_getstatic()` etc. will do some "optimization" only to set the Zero Processor Flag for the caller, destroying the data in `__AX__`. Since bit-fields could require extra steps (shift/mask) after loading the data, the premature "optimization" would be just wrong.

acqn:
The correct test would be to extract all bits of the field with `AND` and then `ORA` them together to get the Z flag, without needing to do shift. However, that would require quite some work.... A quicker fix would be to compare the loaded data with 0:
`if (o.x)` <=> `if (o.x != 0)`

jmr:
Thanks for the debugging. I don't think the `and` solution will be too hard. It's also nice to have a special case for `CF_TEST` because signed doesn't need to sign extend in that case (which is really slow).

PR #1140, opened 2020-07-30
Add test for #1139

jmr:


acqn:
We might also need to test the case when the bit-field doesn't need to be masked with `AND`.

jmr:
I have more test cases. This is just the one that failed. I can add the others, too.

Issue #1009, opened 2020-03-11
Warnings while building 64-bit cc65 using MSVC


madebr:
When trying to build a 64-bit cc65 using MSVC, there are a lot of warnings (that are errors due to `/WX`)

Most are due to `sizeof` and `strlen` returning `size_t` (64-bit) and storing it into a `int` (32-bit).
Some warnings are easy to fix, but changes trickle trough and cause bigger issues.

I've created a [msvc64 branch](https://github.com/madebr/cc65/tree/msvc64) in which I have added a x64 configuration.

jmr:
You can also get these warnings with gcc: `make USER_CFLAGS=-Wshorten-64-to-32`. `-Wconversion` will get you even more warnings.

mrdudz:
-Wshorten-64-to-32 ? What gcc version do i need for this?

jmr:
Oh, it could be a clang thing. gcc on OS X is really clang.

mrdudz:
-Wconversion is a bit too much :)

mrdudz:
is this still an issue?

madebr:
There are still warnings promoted to errors.

I copied the error log below:

<details>

<summary>Log of building cc65 with MSVC16 for Debug|x64</summary>

```
Build started...
1>------ Build started: Project: common, Configuration: Debug x64 ------
1>abend.c
1>addrsize.c
1>alignment.c
1>assertion.c
1>bitops.c
1>chartype.c
1>check.c
1>cmdline.c
1>C:\Users\maarten\source\repos\cc65\src\common\cmdline.c(129,1): error C2220: the following warning is treated as an error
1>C:\Users\maarten\source\repos\cc65\src\common\cmdline.c(129,1): warning C4267: 'initializing': conversion from 'size_t' to 'unsigned int', possible loss of data
1>coll.c
1>cpu.c
1>debugflag.c
1>exprdefs.c
1>fileid.c
1>filepos.c
1>filestat.c
1>filetime.c
1>filetype.c
1>fname.c
1>C:\Users\maarten\source\repos\cc65\src\common\fname.c(55,1): error C2220: the following warning is treated as an error
1>C:\Users\maarten\source\repos\cc65\src\common\fname.c(55,1): warning C4267: 'initializing': conversion from 'size_t' to 'unsigned int', possible loss of data
[...]

PR #1141, opened 2020-07-31
Handle bit-field test after shift/mask

### LINKED ISSUE #1139: Bad codegen with bitfield in if condition

jmr:
Previously, bit-field tests were incorrectly combined with load in
`if (x.bitfield)`.  Delay the test until after the shift/mask
is done.  Still combine tests with load if no shift/mask is required.

Fixes #1139


jmr:
@acqn

src/cc65/loadexpr.c
```diff
@@ -188,34 +200,57 @@ void LoadExpr (unsigned Flags, struct ExprDesc* Expr)
...
-                Expr->BitOffs + Expr->BitWidth != INT_BITS) {
-                g_and (F, (0x0001U << Expr->BitWidth) - 1U);
+            if (ED_NeedsTest (Expr)) {
+                /* If we need to do a test, then we avoid shifting (ASR only shifts one bit
+                ** at a time, so is slow) and just AND with the appropriate mask, then test
+                ** the result of that.
+                */
+
+                /* Avoid overly large shift. */
+                if (EndBit == sizeof (unsigned long) * CHAR_BIT) {
```
acqn, line 225: For the sake of consistency, perhaps `CHAR_BITS` (from "src/cc65/datatype.h") should be used here instead of `CHAR_BIT` (from <limits.h>), and maybe `SizeOf (type_ulong)` insteadof `sizeof (unsigned long)`.

Or maybe `LONG_BITS` (well, `ULONG_BITS` is missing, so this alternative is weak) to replace the whole expression on the right side of the comparison.

acqn: Oh, I see it. the point here is to avoid over-shifting on the compiler side.

> 
> 
> For the sake of consistency, perhaps `CHAR_BITS` (from "src/cc65/datatype.h") should be used here instead of `CHAR_BIT` (from <limits.h>), and maybe `SizeOf (type_ulong)` insteadof `sizeof (unsigned long)`.
> 
> Or maybe `LONG_BITS` (well, `ULONG_BITS` is missing, so this alternative is weak) to replace the whole expression on the right side of the comparison.

Oh, I see it. the point here is to avoid over-shifting on the compiler side.

oliverschmidt:
@acqn: Could you please me an email to the address given in my GH profile? Thanks!

src/cc65/loadexpr.c
jmr: > It is annoying that we didn't use int32_t/uint32_t instead of long/unsigned long

I updated the comment.

> Or maybe LONG_BIT

That's just POSIX, so somewhat less standard.

> It is annoying that we didn't use int32_t/uint32_t instead of long/unsigned long

They probably didn't exist when the code was written. @oliverschmidt would you merge a PR that did `s/unsigned long/uint32_t` (or `uint_fast32_t`), etc?

acqn: > They probably didn't exist when the code was written. @oliverschmidt would you merge a PR that did `s/unsigned long/uint32_t` (or `uint_fast32_t`), etc?

There was a proposal to deal with 64-bit compilation warnings (#1009) which I think should be a good chance to switch to stdint. Still... it could affect every single file in the code base. I have 200+ unmerged commits that could be impacted. XD

oliverschmidt:
> @oliverschmidt would you merge a PR that did s/unsigned long/uint32_t (or uint_fast32_t), etc?

I have in general no problem whatsoever with pure refactoring changes, as long as they keep (or even improve) consistency. There's no old code line or alike we want/need to keep close to.

src/cc65/loadexpr.c
greg-king5: >> It is annoying that we didn't use int32_t/uint32_t instead of long/unsigned long.
>
>They probably didn't exist when the code was written.

Uz was proud of the fact that he didn't use any C99 code in the tools.  He bragged that we could use any ANSI C compiler to build those tools.  Neither `<stdint.h>` nor `<stdbool.h>` are used anywhere.

mrdudz: Wow really? Ever since we started using those things in VICE we keep finding new subtle bugs related to surprise compiling due to "long" and "int"

greg-king5: I forgot to mention that Uz went overboard:  he even didn't use `NULL`!  It frustrates the heck out of me.  Every time that I see a zero being assigned to something, I wonder, "is that an integer or a pointer??"

jmr: > Uz was proud of the fact that he didn't use any C99 code in the tools.

That was probably reasonable until 2010 or so. It's 2020 now, so C99 if not C11 can be relied on.

groessler:
On 8/3/20 9:32 AM, Jesse Rosenstock wrote:
>
> That was probably reasonable until 2010 or so. It's 2020 now, so C99 
> if not C11 can be relied on.
>

There is no rule that just because a new language standard is available 
one has to use it.

I like the idea to "use legacy systems to develop for legacy systems", 
iow, use some rather old system to make 6502 programs. c89 is good 
enough imo.

Issue #1145, opened 2020-08-01
[cc65] Internal error with function type object


acqn:
```c
void f()
{
    f = 0; /* internal error */
}
```
Output:
```
Error: Invalid lvalue in assignment
Error: Incompatible types
Error: Size of data type is unknown
Internal compiler error:
Invalid type in CF flags: 0140, type = 0
```
Cause:
https://github.com/cc65/cc65/blob/e1043fac12151cdec415e76d78e23114d8f0b070/src/cc65/datatype.c#L735
`g_call()`, `g_callind()` and `g_enter()` are abusing this `CF_<FLAG>` "code generator type flag" for passing a special `CF_FIXARGC` parameter, which causes errors when other `CF_<FLAG>` users meet this unexpected flag. A fix is ready.

oliverschmidt:
Fixed with https://github.com/cc65/cc65/pull/1147.

PR #1147, opened 2020-08-01
[cc65] Avoid internal errors when using function-type objects in expressions

acqn:
Resolved Issue #1145.

### DIFF
--- a/src/cc65/datatype.c
+++ b/src/cc65/datatype.c
@@ -755,14 +755,11 @@ unsigned TypeOf (const Type* T)
 unsigned FuncTypeOf (const Type* T)
 /* Get the code generator flag for calling the function */
 {
-    switch (GetUnderlyingTypeCode (T)) {
-
-        case T_FUNC:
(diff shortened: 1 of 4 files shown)


src/cc65/datatype.c
```diff
@@ -751,6 +752,22 @@ unsigned TypeOf (const Type* T)
...
+{
+    switch (GetUnderlyingTypeCode (T)) {
+
+        case T_FUNC:
+            return (((FuncDesc*) T->A.P)->Flags & FD_VARIADIC) ? 0 : CF_FIXARGC;
+
+        default:
+            Error ("Illegal function type %04lX", T->C);
+            return 0;
+    }
```
oliverschmidt, line 766: I can't find a single-case-switch as recurring pattern in cc65. And e.g. GetFuncDesc() seems sort of similar regarding its structure and uses an if() statement. Is there a special reason for the switch() here?

acqn: No special reason for that. It was copy & paste of the original code. Will change.

PR #1152, opened 2020-08-02
LoadExpr: Set CF_FORCECHAR if test is required

jmr:
If we are testing, we do not need to load the high byte(s). This is an optimization.


jmr:
@acqn

acqn:
I think this is fine.

PR #1153, opened 2020-08-02
LoadExpr: Optimize <= 8-bit bit-field loads

jmr:
Set CF_FORCECHAR to do as many operations as char ops as possible.
Clear high byte at the end.

### DIFF
--- a/src/cc65/loadexpr.c
+++ b/src/cc65/loadexpr.c
@@ -117,12 +117,27 @@ void LoadExpr (unsigned Flags, struct ExprDesc* Expr)
         */
         unsigned EndBit = 0;  /* End bit for bit-fields, or zero if non-bit-field. */
         int AdjustBitField = 0;
+        unsigned BitFieldFullWidthFlags = 0;
         if (ED_IsBitField (Expr)) {
             EndBit = Expr->BitOffs + Expr->BitWidth;
             AdjustBitField = Expr->BitOffs != 0 || (EndBit != CHAR_BITS && EndBit != INT_BITS);
 
+            /* TODO: This probably needs to be guarded by AdjustBitField when long bit-fields are
+            ** supported.
+            */
             Flags |= (EndBit <= CHAR_BITS) ? CF_CHAR : CF_INT;
             Flags |= CF_UNSIGNED;
+
+            /* Flags we need operate on the whole bit-field, without CF_FORCECHAR.  */
+            BitFieldFullWidthFlags = Flags;
+
+            /* If we're adjusting, then only load a char (not an int) and do only char ops;
+            ** We will clear the high byte in the adjustment.  CF_FORCECHAR does nothing if the
+            ** type is not CF_CHAR.
+            */
+            if (AdjustBitField) {
+                Flags |= CF_FORCECHAR;
+            }
         } else if ((Flags & CF_TYPEMASK) == 0) {
             Flags |= TypeOf (Expr->Type);
         }
@@ -211,10 +226,7 @@ void LoadExpr (unsigned Flags, struct ExprDesc* Expr)
         ** so be sure to always use unsigned ints for the operations.
         */
         if (AdjustBitField) {
-            /* If the field was loaded as a char, force the shift/mask ops to be char ops.
-            ** If it is a char, the load has already put 0 in the upper byte, so that can remain.
-            */
-            unsigned F = Flags | CF_FORCECHAR | CF_CONST;
+            unsigned F = Flags | CF_CONST;
 
             /* We always need to do something with the low byte, so there is no opportunity
             ** for optimization by skipping it.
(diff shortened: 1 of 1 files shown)


jmr:
@acqn

acqn:
I am fine with the direction. Waiting for confliction resolution.

jmr:
I've messed up the history a bit. Hold on.

jmr:
Should be ok now.

src/cc65/loadexpr.c
```diff
@@ -115,6 +115,7 @@ void LoadExpr (unsigned Flags, struct ExprDesc* Expr)
         ** field is completely contained in the lower byte, we will throw away
         ** the high byte anyway and may therefore load just the low byte.
         */
+        unsigned OrigForceCharFlag = Flags & CF_FORCECHAR;
```
jmr, line 118: I added this, so tell me if there's a better way.

oliverschmidt: I'll keep my feet still until this is resolved...

acqn: What about setting up `F`  before adding the `CF_FORCECHAR` flag to `Flags`, for example right after Line 127?

acqn:
Is the `ED_TestDone (Expr);` on Line 247 duplicate with the same on Line 276?

jmr:
> Is the `ED_TestDone (Expr);` on Line 247 duplicate with the same statement on Line 276? It isn't changed this time but....

Yes, but that's not being changed here and is from #1141 ; #1161 will fix it.

src/cc65/loadexpr.c
jmr: Do you mean keeping both `F` and `Flags` at the outer scope? I don't like `F` having such an unnecessarily large scope when only the `AdjustBitField` branch needs it. I also want to do the loads as `CF_FORCECHAR` if we're adjusting since it will save the top-byte manipulations.

I could imagine something like
```
unsigned BitFieldLoadFlags = 0;
if (AdjustBitField) BitFieldLoadFlags = CF_FORCECHAR;
...
get_static (Flags | BitFieldLoadFlags | ..., ...);
...
if (AdjustBitField) {
    unsigned F = Flags | CF_CONST | CF_FORCECHAR;
    // adjustment code using F
    // final g_and using Flags | CF_CONST
}
```

acqn: You have to keep that kind of info in the large scope in some way, be it stored in `OrigForceCharFlag` , `F` or `BitFieldLoadFlags`.

jmr: Something has to be in the outer scope, it's a question of what. Try the new version.

src/cc65/loadexpr.c
```diff
@@ -250,11 +262,15 @@ void LoadExpr (unsigned Flags, struct ExprDesc* Expr)
...
-                ** g_and emits no code if the mask is all ones.
+                ** If it does end on a boundary, then zeros have already been shifted in,
+                ** but we need to clear the high byte for char.  g_and emits no code if the mask
+                ** is all ones.
                 */
-                if (EndBit != CHAR_BITS && EndBit != INT_BITS) {
-                    g_and (F, (0x0001U << Expr->BitWidth) - 1U);
+                if (EndBit == CHAR_BITS) {
+                    /* We need to clear the high byte, since CF_FORCECHAR was set. */
+                    g_and (BitFieldFullWidthFlags | CF_CONST, 0xFF);
```
acqn, line 271: If `BitFieldFullWidthFlags` is supposed to be used in this condition branch, is the operation with `F` on Line 261 wrong?
https://github.com/cc65/cc65/blob/0da27f297394ed7bbf01d7984c545e1c078f6509/src/cc65/loadexpr.c#L253-L262

jmr: `BitFieldFullWidthFlags` is only needed for the last operation. `g_asr` is the last operation only if `BitOffs == 0 && (EndBit == 8 || EndBit == 16)`, i.e. `!AdjustBitField`.

acqn: It appears to me this block can be only reached if `(AdjustBitField && !ED_NeedsTest (Expr))`:
`if (AdjustBitField ...)`
https://github.com/cc65/cc65/blob/0da27f297394ed7bbf01d7984c545e1c078f6509/src/cc65/loadexpr.c#L228-L236
`if (AdjustBitField ... && !ED_NeedsTest (Expr))`
https://github.com/cc65/cc65/blob/0da27f297394ed7bbf01d7984c545e1c078f6509/src/cc65/loadexpr.c#L253-L276
And then `F` is no longer used after the last `g_asr` on Line 255 but is still modifed on Line261.

jmr: Oh, I didn't look at the line numbers correctly in your first answer. yes, that `F |= CF_CHAR` is useless now, but `F` can still be used for the shift. This will all need some tweaking for `long` bit-fields later.

I removed the useless code.

acqn: I suppose `F == (BitFieldFullWidthFlags | CF_CONST  | CF_FORCECHAR)` when it is used with the `g_asr` here.
`F` could be moved to the `ED_NeedsTest (Expr)` case, and `BitFieldFullWidthFlags` could have `CF_CONST` set from the start so you could just use `BitFieldFullWidthFlags | CF_FORCECHAR` for that `g_asr`, but I am OK with the current code.

jmr: Yes, for now, but that probably won't be the case in the future with something like `long x : 4`. The type part will probably be different, too.

acqn: Uh, I am fine with the latest revision https://github.com/cc65/cc65/pull/1153/commits/5f8e50a454d167b98b709d0021a3d64a5b715608.

Sorry I haven't made it clear.

PR #1158, opened 2020-08-02
[cc65] Improved diagnostics

acqn:
- Increased upper limit of allowed errors before aborting from 10 to 20.
- Always show the count of errors/warnings if there are any errors.
- Tell what is the matter with certain errors.
- Show detailed types of the symbols that caused errors.
- Suppress "Statement has no effect" warnings on stuff that already caused errors.

Most of the changes are just in text (messages).

### DIFF
--- a/src/cc65/assignment.c
+++ b/src/cc65/assignment.c
@@ -127,7 +127,13 @@ void Assignment (ExprDesc* Expr)
 
     /* We must have an lvalue for an assignment */
     if (ED_IsRVal (Expr)) {
-        Error ("Invalid lvalue in assignment");
+        if (IsTypeArray (Expr->Type)) {
+            Error ("Array type '%s' is not assignable", GetFullTypeName (Expr->Type));
+        } else if (IsTypeFunc (Expr->Type)) {
+            Error ("Function type '%s' is not assignable", GetFullTypeName (Expr->Type));
+        } else {
+            Error ("Assignment to rvalue");
+        }
     }
 
     /* Check for assignment to const */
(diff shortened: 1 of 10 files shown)


acqn:
```make
Makefile:142: recipe for target '../../testwrk/misc/goto.g.6502.prg' failed

make[2]: *** [../../testwrk/misc/goto.g.6502.prg] Error 1
```
What is the expected outcome with this file?

mrdudz:
see the comment in the Makefile for these two:
```
# these need reference data that can't be generated by a host-compiled program,
# in a useful way
```
for goto.c the compiler output is checked, goto.ref contains the expected output. this one tests if the warnings the compiler outputs are the ones we expect - so you'll have to carefully (manually) examine the compiler output, and if it is correct then put it into goto.ref. CI fails if the two do not match.

for limits.c the expected output is in limits.ref ... it should not fail as far as i can tell (and CI should fail if so)

acqn:
The consistent failure with the `test/misc/goto.ref` is just unexciting: CRLF vs LF.

mrdudz:
Dunno, since thats a tool we only use for comparing reference output.... perhaps we can make it ignore CR ?

acqn:
Maybe then `tdiff` (stands for "**t**ext **diff**")? Or use different `.ref` files for Windows?

mrdudz:
Different .ref files sounds like overkill to me.... @oliverschmidt ?

oliverschmidt:
> The cause of the consistent failure [...] for me is just unexciting: CRLF vs LF.

Thanks for pointing out.

> since thats a tool we only use for comparing reference output.... perhaps we can make it ignore CR ?

That seems a totally appropriate approach to me.

> Maybe then tdiff (stands for "text diff")?

I don't see a reason to not just name it diff.

PR #1166, opened 2020-08-05
Fix vacuous comparison warning from 0df45fe

jmr:
```
cc65/symentry.c:306:60: warning: address of array 'Sym->Name' will always evaluate to 'true' [-Wpointer-bool-conversion]
    sprintf (TypeName, "%s %s", GetBasicTypeName (T), Sym->Name ? Sym->Name : "<unknown>");
                                                      ~~~~~^~~~ ~
```

### DIFF
--- a/src/cc65/symentry.c
+++ b/src/cc65/symentry.c
@@ -303,7 +303,8 @@ const char* GetSymTypeName (const Type* T)
     if (Sym == 0) {
         return GetBasicTypeName (T);
     }
-    sprintf (TypeName, "%s %s", GetBasicTypeName (T), Sym->Name ? Sym->Name : "<unknown>");
+    sprintf (TypeName, "%s %s", GetBasicTypeName (T),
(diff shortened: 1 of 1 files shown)


jmr:
@acqn

src/cc65/symentry.c
```diff
@@ -303,7 +303,7 @@ const char* GetSymTypeName (const Type* T)
     if (Sym == 0) {
         return GetBasicTypeName (T);
     }
-    sprintf (TypeName, "%s %s", GetBasicTypeName (T), Sym->Name ? Sym->Name : "<unknown>");
+    sprintf (TypeName, "%s %s", GetBasicTypeName (T), *Sym->Name != '\0' ? Sym->Name : "<unknown>");
```
acqn, line 306: I think the cc65 style is to use `Sym->Name[0]` instead of `*Sym->Name`.

jmr: Fixed.

acqn: No problem.

PR #1169, opened 2020-08-07
[cc65] Disallowed empty enum/struct/union types

### LINKED ISSUE #1098: [cc65] Empty enumerator-list?
```c
enum {
};
```
This is a bit different from empty struct/union type issue. I think the C standard requires the enumerator-list in an enum definition, and the example should be ill-formed.

acqn:
Technically speaking, "0-size" is not the same as "memberless" as a struct/union with only "0-size" members could be still "0-size", but there is no way to have such members to begin with as they would have been forbidden in the first place. So any complete struct/union type of 0 size is considered "empty".

This also resolves #1098. While it would be very easy to support empty enums as an extension, such an extension is apparently not useful or popular.


acqn:
So it's `test/misc` again.

acqn:
Empty struct/union tags can be disallowed in `src/cc65/declare.c` instead of `src/cc65/symtab.c`.

oliverschmidt:
@acqn: Do you consider this PR ready for merge?

acqn:
Certainly.

PR #1170, opened 2020-08-07
[cc65] Fixed missing "multiple definition" of local variables

acqn:
This fixed the missing check for multiple definitions of local variables in functions.
```c
void f()
{
    int i, i; /* BUG: Should be a "multiple definition" error */
}
```

### DIFF
--- a/src/cc65/symtab.c
+++ b/src/cc65/symtab.c
@@ -697,18 +697,17 @@ SymEntry* AddEnumSym (const char* Name, const Type* Type, SymTable* Tab)
             /* Existing symbol is not an enum */
             Error ("Symbol '%s' is already different kind", Name);
             Entry = 0;
-        } else {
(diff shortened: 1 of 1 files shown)


src/cc65/symtab.c
```diff
@@ -700,14 +700,16 @@ SymEntry* AddEnumSym (const char* Name, const Type* Type, SymTable* Tab)
         } else {
             /* Define the struct size if the underlying type is given. */
             if (Type != 0) {
-                if (Type !=0 && Entry->V.E.Type != 0) {
+                if (Type != 0 && Entry->V.E.Type != 0) {
```
greg-king5, line 703: The above two lines test `Type != 0` twice.

acqn: Fixed.

PR #1172, opened 2020-08-07
[cc65] Fixes for type comparisons

### LINKED ISSUE #1168: [cc65] Local enums/structs/unions types are wrongly judged in comparison
At the moment, cc65 does track ESU types in the correct scopes where they are declared/defined, but when comparing these types, it ignores the scopes and just assume they be identical if their tag names and field names match.

```c
struct X {
    int a;
} x, *px;

int main(void)
{
    struct X y, *py;

    /* Local 'struct X' shadows global 'struct X' and is an incompatible type */
    struct X {
        int a;
    } z, *pz;

    x  = y;     /* OK, same type */
    px = py;    /* OK, same type */
    x  = z;     /* BUG: Should issue an error */
    px = pz;    /* BUG: Should issue a warning */
    y  = z;     /* BUG: Should issue an error */
    py = pz;    /* BUG: Should issue a warning */

    return 0;
}
```
Cause: the current `DoCompare()` function gets wrong with:

https://port70.net/~nsz/c/c99/n1256.html#6.7.2.3p5:
> 5 Two declarations of structure, union, or enumerated types _which are **in different scopes** or use different tags declare **distinct types**_. Each declaration of a structure, union, or enumerated type which does not include a tag declares a distinct type.

It is currently ignoring the scope requirement and comparing the types field-by-field... which might the consequence of confusing the rules for that in separated translation units.

A fix for this is done.

acqn:
- Fixed potential problems with type comparison with different enum types.
- Fixed #1168.
- Fixed typedefs related to arrays/pointers:
```c
typedef int P[];
typedef int P[1]; /* Should be an error */
typedef int *P;   /* Should be an error */
```

### DIFF
--- a/src/cc65/symtab.c
+++ b/src/cc65/symtab.c
@@ -558,34 +558,46 @@ static int HandleSymRedefinition (SymEntry* Entry, const Type* T, unsigned Flags
     unsigned E_SCType = Entry->Flags & SC_TYPEMASK;
     unsigned SCType   = Flags & SC_TYPEMASK;
 
-    /* Some symbols may be redeclared if certain requirements are met */
-    if (IsTypeArray (T) && IsTypeArray (E_Type)) {
+    /* Existing typedefs cannot be redeclared as anything different */
+    if (E_SCType == SC_TYPEDEF) {
 
-        /* Get the array sizes */
-        long Size  = GetElementCount (T);
-        long ESize = GetElementCount (E_Type);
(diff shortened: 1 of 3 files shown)


oliverschmidt:
@acqn: Do you consider this PR ready for merge?

acqn:
Yes. This PR contains only the part that I was confident of.

PR #1175, opened 2020-08-12
[cc65] Utility for checking general datatype categories etc.

acqn:
These are used by several solutions to nown issues.

Note: some have names that may resemble the concepts in the ISO C standard, but they are not exactly the same.

### DIFF
--- a/src/cc65/datatype.c
+++ b/src/cc65/datatype.c
@@ -1037,10 +1037,69 @@ Type* ArrayToPtr (Type* T)
 
 
 
+int IsClassObject (const Type* T)
+/* Return true if this is a fully described object type */
+{
+    return !IsTypeFunc (T) && !IsClassIncomplete (T);
+}
+
+
+
+int IsClassIncomplete (const Type* T)
+/* Return true if this is an object type lacking size info */
+{
+    if (IsTypeArray (T)) {
+        return GetElementCount (T) == UNSPECIFIED;
+    }
+    return IsTypeVoid (T) || IsIncompleteESUType (T);
+}
+
+
+
 int IsClassArithmetic (const Type* T)
-/* Return true if this is an arithmetic type */
+/* Return true if this is an integer or real floating type */
 {
     return IsClassInt (T) || IsClassFloat (T);
+}
+
+
+
+int IsClassBasic (const Type* T)
+/* Return true if this is a char, integer or floating type */
+{
+    return IsRawTypeChar (T) || IsClassInt (T) || IsClassFloat (T);
+}
+
+
+
+int IsClassScalar (const Type* T)
+/* Return true if this is an arithmetic or pointer type */
+{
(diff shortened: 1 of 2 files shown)


src/cc65/datatype.c
```diff
@@ -1037,18 +1037,113 @@ Type* ArrayToPtr (Type* T)
...
 
+int IsClassObject (const Type* T)
+/* Return true if this is a fully described object type */
+{
+    return !IsTypeFunc (T) && !IsClassIncomplete (T);
+}
+
+
+
+int IsClassIncomplete (const Type* T)
```
greg-king5, line 1048: It makes more sense to me to swap the function names `IsClassIncomplete()` and `HasUnknownSize()`.  I can't think of `void` as being "incomplete".

Also, this function is a superset of the other function.  This one simply can return the value of the other one or `IsTypeVoid()`.

acqn: The standard requires `void` to be incomplete (https://port70.net/~nsz/c/c99/n1256.html#6.2.5p19), and `IsClassIncomplete()` is to reflect that concept. The reason why there is `HasUnknownSize()` is that  cc65 supports "void variables" as an extension. It looks more logical to me that `HasUnknownSize()` should be implemented with `IsClassIncomplete()` as `IsClassIncomplete() and !IsTypeVoid()`. What's your opinion?

src/cc65/datatype.c
```diff
@@ -1037,18 +1037,113 @@ Type* ArrayToPtr (Type* T)
...
+
+
+
+int IsRelationType (const Type* T)
+/* Return true if this is an arithmetic, array or pointer type */
+{
+    return IsClassArithmetic (T) || IsClassPtr (T);
+}
+
+
```
acqn, line 1105: There should be one more blank line.

acqn:
This PR was originally intended for getting the blocking part of multiple other PRs to be merged first. It is now going on the opposite way. I am considering splitting the controversial parts off from this.

src/cc65/datatype.c
greg-king5: It's inefficient.  That implementation would call `IsTypeVoid()` twice.  My idea is an implementation detail rather than a semantic detail.

acqn: What's your opinion on the naming?

As to compiler performance, my old rough profiling indicated that the heaviest CPU hog in non-optimized build was... `CHECK ()`. Surprised?

greg-king5: > What's your opinion on the naming? Do you still insist on swapping the two?

No, after reading the "Types" section of C99, I agree with your names.
> 
> As to compiler performance, my old rough profiling indicated that the heaviest CPU hog in non-optimized build was... `CHECK ()`. Surprised?

Yep.

Issue #1167, opened 2020-08-05
[cc65] Function prototypes?


acqn:
Besides the issue mentioned in #1142 and #1162, there are several other issues with function declarations in cc65:
```c
int f1(int (*)(char *), int (*)[]);
int f1(int (*)(), int (*)[3]);          /* BUG cause:  This simply replaces the previous function descriptor */
int f1(int (*)(int), int (*)[]);        /* BUG result: Compiles. Should be an error */

int f2(int (*)(), int (*)[3]);
int f2(int (*)(char *), int (*)[]);     /* BUG cause:  This simply replaces the previous function descriptor */
int f2(int (*)(char *), int (*)[2]);    /* BUG result: Compiles. Should be an error */

int f3(void);
volatile int f3(void);                  /* Question: Is this OK? */

void f4(int);
void f4(volatile int);                  /* Question: Is this OK? */
```
The second prototypes of `f1` and `f2` simply replace the previous ones.

acqn:
More:
```c
void scope1()
{
    extern void f(int (*)[1]);
}

void scope2()
{
    extern void f(int (*)[2]); /* UB */
}
```

oliverschmidt:
```
int f3(void);
volatile int f3(void);                  /* Question: Is this OK? */

void f4(int);
void f4(volatile int);                  /* Question: Is this OK? */
```
Yet again being naive: Is there a potentially useful meaning for those? I guess 'no'. I don't see why someone would have written that or why it should be allowed to write that.

oliverschmidt:
```
void scope1()
{
    extern void f(int (*)[1]);
}

void scope2()
{
    extern void f(int (*)[2]); /* UB */
}
```
I just don't know enough about C: Those two are declarations referring to definitions done somewhere else, right? Are such declarations in general allowed inside functions? Is this scenario special in that the declarations are functions? <...>?

polluks:
@oliverschmidt because of symmetry

acqn:
I got some answers from https://en.cppreference.com/w/c/language/function_declaration.

> 
> 
> ```
> int f3(void);
> volatile int f3(void);                  /* Question: Is this OK? */
> 
> void f4(int);
> void f4(volatile int);                  /* Question: Is this OK? */
> ```
> 
> Yet again being naive: Is there a potentially useful meaning for those? I guess 'no'. I don't see why someone would have written that or why it should be allowed to write that.

[...]

PR #1176, opened 2020-08-12
[cc65] Fixed several "multiple function declarations" issues

### LINKED ISSUE #1173: [cc65] Conflicting 'extern' declarations
'extern' declarations should not conflict with other _local_ declarations.

Currently, cc65 warns about conflicting 'static' vs 'extern' declarations in the file scope and proceed assuming them all 'extern', which is fine. But it doesn't give any diagnostics when there are conflicting 'static' vs 'extern' declarations in the function bloack scopes.

```c
void foo(int p);
static void foo(int a)  /* Warning */
{
    extern int k;
    int k;              /* BUG: should fail */
    int l;
    extern int l;       /* BUG: should fail */
    extern int m;
    static int m;       /* BUG: should fail */
    static int n;
    extern int n;       /* BUG: should fail */
    a = p;              /* Error on 'p' */
}

/* More examples */
static char a;
extern char a;

void bar()
{
    int a;
    int a();
    void vv();
    extern void vv(char []);
    void vv(char [1]);
}

void scope()
{
    extern void foo(int [3]);
    int foo;
    extern int m[3];
    extern int a = 0;
    a = 0;
}

int m[2];
```
Unlike in the file scope case, it wouldn't make any sense to just warn and proceed in function block scopes, because the 'extern' declarations would be referring to _different_ objects other than the conflicting local ones. So they should be forbidden with errors.

acqn:
- Fixed some parts of Issue #1167. The left-over part (https://github.com/cc65/cc65/issues/1167#issuecomment-670254608) will be in a separated PR.
- Fixed #1173.

### DIFF
--- a/src/cc65/codeinfo.c
+++ b/src/cc65/codeinfo.c
@@ -392,7 +392,7 @@ fncls_t GetFuncInfo (const char* Name, unsigned short* Use, unsigned short* Chg)
 
         /* Did we find it in the top-level table? */
         if (E && IsTypeFunc (E->Type)) {
-            FuncDesc* D = E->V.F.Func;
+            FuncDesc* D = GetFuncCompositeDesc (E);
 
             /* A variadic function will use the Y register (the parameter list
             ** size is passed there). A fastcall function will use the A or A/X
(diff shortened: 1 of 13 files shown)


acqn:
So it is #191.

oliverschmidt:
@greg-king5 has edited the initial comment one day ago but didn't add review comments as of now. Given that combined with no additional feedback from others makes me consider this PR as ready to be merged.

Issue #1123, opened 2020-07-26
"Result of comparison is constant" warning in constant expressions


jmr:
`int x[0 == 0];` [gives the warning](https://godbolt.org/z/v3f45d):
```
<source>(1): Warning: Result of comparison is constant
```

This isn't a useful warning, since the expression here has to be a constant.

acqn:
This was on my TODO list as a "codegen improvement" but at low priority.

acqn:
I got a hold on the old work and completed the part needed for this and #250. It looks to be working and resolved. Unfortunately it was on a very different branch and might need quite some modifications to be rebased onto the current stream.

acqn:
Fixed in #1220. It is also improved that the warnings now tell the result values (`true` or `false`).

PR #1182, opened 2020-08-14
[cc65] Fixed testing 'struct->field'

### LINKED ISSUE #1181: Testing struct member against NULL is broken
As hinted in Issue #1179 i was tracking down a compiler bug... i think i nailed it now :) This breaks my menu system in many interesting ways, would be nice to get a fix soonish (and i promise to work through recent issues and PRs and create more tests... :))

First of all, the last working commit is ba48dfe, 21084895 breaks it.

It seems that generally tests for null (eg) in this form are broken:
```
while (struct->member) { .... }
```

In the "working" compiles of my menusystem, the respective code has the test for null inline, eg like this:
```
; if(menu->helpline) {
	ldy     #$07
	jsr     ldaxysp
	ldy     #$02
	sta     ptr1
	stx     ptr1+1
	lda     (ptr1),y
	iny
	ora     (ptr1),y
	beq     L00AB
```

In the non working compiles it looks like this:
```
; if(menu->helpline) {
;
	ldy     #$07
	jsr     ldaxysp
	ldy     #$03
	jsr     ldaxidx
	beq     L00AB
```

While this is shorter, and the actual "load" is perhaps correct (i did not check this), the check for NULL randomly fails (because it doesnt actually happen, its pure coincidence it will sometimes appear to work right). This is probably most obvious when the accessed struct member is a 16bit (or even 32bit) type.

I created a testcase... another thing worth noting is that the testcase itself is kindof fragile - "cleaning up" may result in making it "work" (although the code is still wrong).

Also, it seems noteable that the "printf" is always broken too! that looks like a stack issue to me - which may or may not be related.

[menubug.zip](https://github.com/cc65/cc65/files/5071585/menubug.zip)

@acqn: I can perhaps create a better testcase, if you can hint me at what is actually going wrong here and how to test it better :)

acqn:
- Fixed #1181.

Note: Usually, the C parser recursively parses expression reusing the same `ExprDesc` object for the result as well as for the "left-hand subexpression" if it is binary or ternary. This could lead to problems like this if the info to and from the subexpression is not handled carefully. It could be avoided in the future by separating the info of the two expressions (See https://github.com/cc65/cc65/issues/1123#issuecomment-664114740).

### DIFF
--- a/src/cc65/assignment.c
+++ b/src/cc65/assignment.c
@@ -77,7 +77,7 @@ static int CopyStruct (ExprDesc* LExpr, ExprDesc* RExpr)
     /* Get the expression on the right of the '=' */
     hie1 (RExpr);
 
-    /* Check for equality of the structs */
+    /* Check for equality of the structs/unions */
     if (TypeCmp (ltype, RExpr->Type) < TC_STRICT_COMPATIBLE) {
         TypeCompatibilityDiagnostic (ltype, RExpr->Type, 1,
             "Incompatible types in assignment to '%s' from '%s'");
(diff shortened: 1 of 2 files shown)


mrdudz:
Works perfectly with my menu system and the derived testcase. Thanks a lot!

Test will follow!

PR #1183, opened 2020-08-14
Use C89 semantics for integer conversions

### LINKED ISSUE #170: Wrong implicit conversion of integers
While the standard says that an unsigned char should be promoted to int if an int can represent all values of an unsigned char, and similar for promotion from unsigned to long, cc65 promotes unsigned types to unsigned types of the larger range. That is, unsigned char is promoted to unsigned, and unsigned to unsigned long.

This can be easily tested. The following code should promote the operands to integer and produce signed results; but, if you look at the generated assembly, it uses the runtime routines for unsigneds, which generate unsigned results:

``` C
int main(void)
{
    unsigned char c = 2;
    unsigned u = 2;
    int a = -2;
    long l = -2;

    /* Generated code should use tosmulax but uses tosumulax */
    int r = c * a;
    /* Generated code should use tosmuleax but uses tosumuleax */
    long lr = u * l;

    return 0;
}
```

While, in the examples above, that has no ill effects, and is therefore only visible by looking at the generated code, there are other places where the compiler gets tests wrong, or produces invalid calculation results. For example:

``` C
long n = -95;
unsigned int d = 3;
int r = n/d; // produces 21813 instead of 31
```

A workaround is to manually force usage of the correct type:

``` C
int r = n / (long) d;
```

The disadvantage being that one must really understand the nature of the bug to apply the proper casts.

A fix is possible with not too much effort (needs rewrite of two functions in the compiler), but introduces two other problems:

The fixed code has a higher chance to trigger an error in the optimizer which causes invalid code transformations. This optimizer problem does actually exist without the fix, but chances for triggering it are rather low.
Fixing the conversions will have a negative impact on code quality. The reason is the integer promotions: The C standard says that operands to most binary operators must be converted to int. If that conversion is done to signed instead of unsigned, most operations on char will actually operate on signed ints. And on 6502 machines, code for a signed operation is usually larger and slower than the same code for an unsigned operation.

So, if you have plans to solve the issue, be at least prepared to have a thorough look at the generated code.

jmr:
Previously, the following rules were used for binary operators:
* If one of the values is a long, the result is long.
* If one of the values is unsigned, the result is also unsigned.
* Otherwise the result is an int.

C89 specifies the "usual arithmetic conversions" as:
* The integral promotions are performed on both operands.
* Then the following rules are applied:
  * If either operand has type unsigned long int, the other operand is
    converted to unsigned long int.
  * Otherwise, if one operand has type long int and the other has type
    unsigned int, if a long int can represent all values of an unsigned int,
    the operand of type unsigned int is converted to long int; if a long int
    cannot represent all the values of an unsigned int, both operands are
    converted to unsigned long int.
  * Otherwise, if either operand has type long int, the other operand is
    converted to long int.
  * Otherwise, if either operand has type unsigned int, the other operand is
    converted to unsigned int.
  * Otherwise, both operands have type int.
https://port70.net/~nsz/c/c89/c89-draft.html#3.2.1.5

As one example, these rules give a different result for an operator
with one long operand and one unsigned int operand.  Previously,
the result type was unsigned long.  With C89 semantics, it is just long,
since long can represent all unsigned ints.

Integral promotions convert types shorter than int to int (or unsigned int).
Both char and unsigned char are promoted to int since int can represent
all unsigned chars.
https://port70.net/~nsz/c/c89/c89-draft.html#3.2.1.1

Rename promoteint to ArithmeticConvert, since this is more accurate.

Fixes #170


acqn:
Uh, I had a different fix for this issue as I mentioned in #170....

acqn:
It seems that your version follows the correct C89 rules. Mine could be defective.

PR #1185, opened 2020-08-15
[cc65] Fixed function signatures in asm output

acqn:
- Fixed function signatures in asm output.
- Improved full type name production.

(1/1) commit pushed in PR but see below.

---
Caveat: this combined with #1187 will output the "composite" version of function parameter lists instead of the "definition" version of function parameter lists. If the "definition" version is desired, either the implementation of #1187 or some other part of the code base outside of this need be modified.

### DIFF
--- a/src/cc65/datatype.c
+++ b/src/cc65/datatype.c
@@ -83,7 +83,7 @@ static struct StrBuf* GetFullTypeNameWestEast (struct StrBuf* West, struct StrBu
 ** eastern part.
 */
 {
-    struct StrBuf Buf = STATIC_STRBUF_INITIALIZER;
+    struct StrBuf Buf = AUTO_STRBUF_INITIALIZER;
 
     if (IsTypeArray (T)) {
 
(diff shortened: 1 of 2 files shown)


acqn:
Some brain teaser:

```c
int (*(*(*__fastcall__ f(register int a, const int *(* volatile fp)(unsigned char, ...), const int (* const m)[]))[])(int, ...))[]
{
    return f; /* Just to test warning message */
}
```

Before:
```c
; ---------------------------------------------------------------
; __near__ __near__ __near__ function returning __near__ int [] * * [] * __near__ __fastcall__ f (int, volatile __near__ __near__ function returning __near__ const int * *, const __near__ const int [] *)
; ---------------------------------------------------------------
```
After:
```c
; ---------------------------------------------------------------
; int (*(*(*__near__ __fastcall__ f (int a, const int *(*volatile fp)(unsigned char, ...), const int (* const m)[]))[])(int, ...))[]
; ---------------------------------------------------------------
```

oliverschmidt:
> If the "definition" version is desired, either the implementation of #1187 or some other part of the code base outside of this need be modified.

I don't see the necessity to do so.

oliverschmidt:
As there isn't any negative feedback by now...

PR #1187, opened 2020-08-15
[cc65] Fixed calling functions with parameters first declared with an empty parameter list

acqn:
This fixes the regression in #1176:

```c
void f();
void f(int);

void g(void)
{
    f(0); /* BUG: this shouldn't fail */
}
```
(1/1) commit pushed in PR but see below.

---
Caveat: #1185 combined with this will output the "composite" version of function parameter lists instead of the "definition" version of function parameter lists. If the "definition" version is desired, either the implementation of this or some other part of the code base outside of #1185 need be modified.


oliverschmidt:
As there isn't any negative feedback by now...

PR #1189, opened 2020-08-15
[cc65] Fixed compiler CHECK failure when calling functions defined with duplicate param names

acqn:
There was the problem:
```c
void f(int a, int a);

void g()
{
    f(0, 1); /* Check failed: (Param->Flags & SC_PARAM) != 0 */
}
```
The cause was that `NewFunc()` adds two special symbols (non-NULL) onto the function's local symbol table, while `FunctionParamList()` was only expecting function parameters there despite it prevented from NULL. Since there are still possibilities that some other part of the compiler could insert symbols there before calling `FunctionParamList()`, it is a better and robuster way to use anonymous parameters to replace name-conflicted parameters to fix this than to just deal with the two special symbols. Now done (1/1).

Also clarified the terms "parameter" vs "argument" in FunctionParamList().

### DIFF
--- a/src/cc65/expr.c
+++ b/src/cc65/expr.c
@@ -296,32 +296,30 @@ static void WarnConstCompareResult (void)
 
 
 static unsigned FunctionParamList (FuncDesc* Func, int IsFastcall)
-/* Parse a function parameter list, and pass the parameters to the called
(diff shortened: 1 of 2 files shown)


greg-king5:
:+1: for the patch.
:tada: for the better use of "parameter" and "argument".

PR #1192, opened 2020-08-16
Treat signed int bit-fields as signed

### LINKED ISSUE #1095: [cc65] Why are 'signed int' bit-fields unsigned in cc65?

jmr:
Prior to this PR, `int`, `signed int`, and `unsigned int`
bitfields are all treated as `unsigned int`.

With this PR, `signed int` will be treated as `signed int`,
and the others remain unsigned.

Since `Type` does not distinguish between `int` and `signed int`,
add an extra `int* SignenessSpecified` param to `ParseTypeSpec`
so we can tell these apart for bit-fields and treat plain `int : N`
as `unsigned int : N` since it is more efficient to zero-extend
than sign-extend.

Fixes #1095

### DIFF
--- a/src/cc65/loadexpr.c
+++ b/src/cc65/loadexpr.c
@@ -123,7 +123,9 @@ void LoadExpr (unsigned Flags, struct ExprDesc* Expr)
             ** supported.
             */
             Flags |= (EndBit <= CHAR_BITS) ? CF_CHAR : CF_INT;
-            Flags |= CF_UNSIGNED;
+            if (IsSignUnsigned (Expr->Type)) {
+                Flags |= CF_UNSIGNED;
+            }
 
             /* Flags we need operate on the whole bit-field, without CF_FORCECHAR.  */
             BitFieldFullWidthFlags = Flags;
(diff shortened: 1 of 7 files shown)


src/cc65/loadexpr.c
```diff
@@ -255,20 +263,82 @@ void LoadExpr (unsigned Flags, struct ExprDesc* Expr)
...
+                    ZeroExtendMask = 0xFF;
                 } else if (EndBit != INT_BITS) {
-                    g_and (BitFieldFullWidthFlags | CF_CONST, (0x0001U << Expr->BitWidth) - 1U);
+                    ZeroExtendMask = (1U << Expr->BitWidth) - 1;
+                }
+
+                /* Handle signed bit-fields. */
+                if (IsSignSigned (Expr->Type)) {
+                    /* Push A, since the sign bit test will destroy it. */
+                    AddCodeLine ("pha");
```
jmr, line 288: I see that `AddCodeLine` is mostly used in `codegen.c`. Let me know if these should be moved to something like `g_extendbitfield` in that file.

acqn: Perhaps. The case for bitfields looks "inlined" a bit too much compared to others.

oliverschmidt: @jmr: I presume you want to add another commit with the proposed modification.

jmr: Sure. Let's do #1197 first, then I'll rebase on top of that.

jmr: Done.

oliverschmidt:
@acqn didn't provide additional feedback beside answering @jmr's question. From that perspective I consider this PR as ready to be merged now...

PR #1193, opened 2020-08-17
Add test of union of bit-field from mailing list

jmr:
https://sourceforge.net/p/cc65/mailman/message/36152700/

This currently works, but add a test to prevent future regressions.


mrdudz:
guess i can merge this, thanks! :)

PR #1194, opened 2020-08-17
[cc65] Fixed type checking in comparison operations

### LINKED ISSUE #1177: [cc65] Comparing with non-arithmetic types
```c
struct {
    int a;
} s;

union {
    char m[32];
} u;

int main()
{
	return s > u; /* BUG: no error?! */
}
```

acqn:
- Fixed #1177.


oliverschmidt:
Compared to other recent PRs this seems a rather "small" one - and without any negative feedback so far...

PR #1195, opened 2020-08-17
[cc65] Made 'char' a distinct type

### LINKED ISSUE #1163: [cc65] Distinct char type?
Well, in the ISO C the `char`, `signed char` and `unsigned char` types are 3 different types from each other. So the following should fail in cc65 (but currently it doesn't):
```c
char a;
unsigned char a;         /* should fail as 'char' is not the same type as 'unsigned char' */

void f(char *);
void f(unsigned char *); /* should fail for the same reason as above */
```
Would we fix this?

acqn:
This makes `char` a distinct type from `signed char` and `unsigned char`. It is till using `unsigned char` by default (which can be switched to `signed char` with compiler option `--signed-chars`) as the underlying type and compatible with them.

- Resolved #1163.

### DIFF
--- a/src/cc65/datatype.c
+++ b/src/cc65/datatype.c
@@ -59,6 +59,7 @@
 
 
 /* Predefined type strings */
+Type type_char[]        = { TYPE(T_CHAR),   TYPE(T_END) };
 Type type_schar[]       = { TYPE(T_SCHAR),  TYPE(T_END) };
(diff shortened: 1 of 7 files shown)


oliverschmidt:
It seems to me that all necessary discussion already happened in https://github.com/cc65/cc65/issues/1163 and that this PR has structure-wise a rather low complexity. So it seems appropriate to merge without further feedback...

PR #1197, opened 2020-08-17
Move bit-field adjustment to codegen.c

jmr:
Extract functions `g_testbitfield` and `g_extractbitfield` from `LoadExpr`.

This helps prepare for #1192, since `g_extractbitfield` will get much
longer and call `AddCodeLine`.

### DIFF
--- a/src/cc65/codegen.c
+++ b/src/cc65/codegen.c
@@ -33,6 +33,7 @@
 
 
 
+#include <limits.h>
 #include <stdio.h>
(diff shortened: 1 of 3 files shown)


oliverschmidt:
As far as I see this PR is the direct result of the discussion https://github.com/cc65/cc65/pull/1192#pullrequestreview-468094267. It seems to "only" move code (as discussed) and therefore doesn't "require" a review...

PR #1199, opened 2020-08-18
[cc65] Removed special-casing code for scaling up

acqn:
(1/2)

- Removed special-casing code for scaling up. Now it uses the normal multiplying code.

- [ ] Added support for changing multiplications with certain negative multipliers into bit-shifts + negations.

(Pending) This was an old changeset inspired by the #169 fix. It transforms some multiplications with negated power-of-2 operands to bit-shifts + negations when the code size factor settings allow.

### DIFF
--- a/src/cc65/codegen.c
+++ b/src/cc65/codegen.c
@@ -1496,54 +1496,14 @@ void g_scale (unsigned flags, long val)
 ** pointer points to.
 */
 {
-    int p2;
-
     /* Value may not be zero */
     if (val == 0) {
         Internal ("Data type has no size");
     } else if (val > 0) {
 
-        /* Scale up */
-        if ((p2 = PowerOf2 (val)) > 0 && p2 <= 4) {
-
-            /* Factor is 2, 4, 8 and 16, use special function */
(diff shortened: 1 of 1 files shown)


src/cc65/codegen.c
```diff
@@ -2590,13 +2550,26 @@ void g_mul (unsigned flags, unsigned long val)
...
     /* Do strength reduction if the value is constant and a power of two */
-    if (flags & CF_CONST && (p2 = PowerOf2 (val)) >= 0) {
-        /* Generate a shift instead */
-        g_asl (flags, p2);
-        return;
+    if (flags & CF_CONST) {
+
+        /* Deal with negative values if it's signed multiplication */
+        int Negation = (flags & CF_UNSIGNED) == 0 && (signed long)val < 0;
+        int p2 = PowerOf2 (Negation ? (unsigned long)-(signed long)val : val);
```
jmr, line 2558: Style nit: to me, `(0 - val)` is clearer and requires less thinking than `(unsigned long)-(signed long)val`.

PR #1200, opened 2020-08-18
[cc65] Fixed Issue #327

### LINKED ISSUE #327: Promoting u8 to s16 gives wrong result
This is a subset of #170 I believe, and fixing this shouldn't cause the general perf regression mentioned there.

```c
#include <nes.h>
#include <conio.h>
#include <stdio.h>
#include <stdint.h>

static const uint8_t arr[2] = {
	0,
	255
};

static int16_t get16() {
	return -arr[1];
}

int main() {

	cprintf("Value %d, should be -255", get16());

	while (1)
		waitvblank();

	return 0;
}
```

This program prints 1, not -255.

acqn:
- [x] Fixed #327.

Unfortunately, fixing this will hinder optimizations with s8 due to the fact that signed integer promotion introduces a conditional branch to set the high byte of the promoted s16, which will make most if not all optimizer steps bail out upon due to complecity dealign with conditional branches.


acqn:
There are two ways to resolve the .
- Use some "intrinsics/pseudo-insns" to replace the "signed promotion" sequences (and the likes such as the conditional `inx` in s8/u8 additions) so that the optimizer will see no extra branches introduced.
- Add some new optimization steps that can remove unnecessary "signed promotion" sequences and the likes before leaving the code to the rest steps.

jmr:
> Add some new optimization steps that can remove unnecessary "signed promotion" sequences

Which ones would be unnecessary?

PR #1208, opened 2020-08-21
[cc65] Made struct assignment less hackish

acqn:
(1/1) This basically does two things:
- Always mark the assigned struct as untested.
- Ensure the "expression type" of the assigned struct is a reference instead of a pointer (which doesn't matter for now but is not so future-proof in C3X or C4X or so, maybe).


oliverschmidt:
No feedback within three days - merging...

Issue #1209, opened 2020-08-21
backwards indirect goto does not compile


jmr:
```
int f (void)
{
    static void *x[1];
    L: if (x[0] != 0) return 0;
    x[0] = &&L;
    goto *x[0];
}
```
Results in:
```
Unresolved external 'L0004' referenced in:
  bugXXXX-ind-goto-rev.s(61)
  bugXXXX-ind-goto-rev.s(62)
```

This works with gcc.

It looks like something is wrong with the way `GotoStatement` is trying to prevent labels from getting deleted.

I'll send a PR with a test case soon.

clbr:
There is no such error because it didn't occur to me at the time someone would try something that crazy...

clbr:
The new sample still confuses me on why you would do that in the first place, but ok, it's a valid bug.

jmr:
I don't actually want to do any of the things in this bug, but there should be no reason not to allow them. I was trying to look at the computed goto label handling code to see how things should work because there are (unsurprisingly) similar issues for #1049 trying to use jump tables for switch. I looked at the code, said, "wait, how does this work?" and found that it doesn't. #1211 is equivalent to something that needs to be handled for #1049, so I'm more interested in a fix for that one.

acqn:
My fix for this depends on something tha depends on something that depends on something that depends on something that seemingly irrelevent to this one.... The "gap" between the current HEAD of the master branch and the fix is ~20 commits. I'll see if I can rebase it.

oliverschmidt:
If you plan to close the gap in the foreseeable future, I personally don't see why you should invest into changing the order of your commits just because of the existence of this issue.

acqn:
I've been closing the gap. There are still (3 + 1) + 21 = 25 commits to go before the fix branch could be merged and tested. That would be the 2 current PRs plus about 6 new ones.

greg-king5:
> I've been closing the gap. There are still (3 + 1) + 21 = 25 commits to go before the fix branch could be merged and tested. That would be the 2 current PRs plus about 6 new ones.

We "nosy bees" appreciate the progress report.

oliverschmidt:
+1

mrdudz:
yes! I am totally overwhelmed by how it is moving forward atm. Don't stop :)

PR #1214, opened 2020-08-24
Move #1209 test from err/ to misc/

jmr:
`misc/` is the correct place for tests that should compile but do not.

Revert `err/Makefile` changes from #1210.


test/err/Makefile
```diff
@@ -23,23 +23,23 @@ ifdef QUIET
...
 all: $(TESTS)
 
 $(WORKDIR):
 	$(call MKDIR,$(WORKDIR))
 
-$(WORKDIR)/%.prg: %.c | $(WORKDIR)
+$(WORKDIR)/%.s: %.c | $(WORKDIR)
 	$(if $(QUIET),echo err/$*.s)
-	$(NOT) $(CL65) -o $@ $< $(NULLERR)
+	$(NOT) $(CC65) -o $@ $< $(NULLERR)
```
oliverschmidt, line 42: I presume the revert of recent changes in this file aren't desired - are they?

Issue #1080, opened 2020-07-09
make -j


polluks:
The parallel option does not seem to work. Are our makefiles not smart enough?

oliverschmidt:
Sorry but "does not seem to work" doesn't qualify as a valid issue description.

polluks:
Well, it's not deterministic
````
...
atarixl - atr10.o
make[1]: *** [atarixl] Error 2
apple2enh - tgi/tgi_load.s
apple2enh - tgi/tgi_load_vectorfont.c
apple2enh - tgi/tgi_outcode.s
apple2enh - tgi/tgi_outtext.s
make[2]: *** Waiting for unfinished jobs....
apple2enh - tgi/tgi_outtextxy.s
apple2enh - tgi/tgi_pieslice.c
apple2enh - tgi/tgi_popxy.s
apple2enh - tgi/tgi_popxy2.s
apple2enh - tgi/tgi_setaspectratio.s
apple2enh - tgi/tgi_setcolor.s
apple2enh - tgi/tgi_setdrawpage.s
apple2enh - tgi/tgi_setpalette.s
apple2enh - tgi/tgi_setpixel.s
apple2enh - tgi/tgi_settextdir.s
apple2enh - tgi/tgi_settextstyle.s
apple2enh - tgi/tgi_setviewpage.s
apple2enh - apple2/tgi_stat_stddrv.s
make[1]: *** [atari] Error 2
apple2enh - apple2/tgi_stddrv.s
...
````
````
GNU Make 3.81
Copyright (C) 2006  Free Software Foundation, Inc.
This is free software; see the source for copying conditions.
There is NO warranty; not even for MERCHANTABILITY or FITNESS FOR A
PARTICULAR PURPOSE.

This program built for powerpc-apple-darwin9.0
````

mrdudz:
I can reproduce it, it randomly fails with -j

To me it looks like the ar65 command comes before all required objects have been compiled (which fails, of course)

polluks:
I was reading https://bugzilla.redhat.com/show_bug.cgi?id=212111 but it still fails
````
...
gmake[1]: *** [Makefile:118: atari] Error 2
c64 - c64-georam.o
c64 - c64-hitjoy.o
c64 - c64-numpad.o
c64 - c64-ptvjoy.o
c64 - c64-stdjoy.o
c64 - c64-joy.o
c64 - c64-swlink.o
c64 - c64-hi.o
[...]

Issue #301, opened 2016-05-29
Labels should be normalized


clbr:
When making a tiny change in C code, like adding a condition inside an if (), it often changes all asm labels, creating 1000+ line asm diffs. There is no need to change the labels, and there's seemingly no logic to it at all - no new labels were added, so why the offset?

There should be a normalization pass before outputting the final code, renumbering all labels consistently. This would help a lot in tracking down compiler bugs.

greg-king5:
New labels are added by the C-to-assembly translator; then, they are removed by the optimizer.

Also, cc65 uses single-pass translation.  Sometimes, it anticipates that it might need a label; so, it creates that label.  But later, it sees that the label wasn't needed.  Discarded labels aren't re-used.

acqn:
If the point is to make diff easier, then "normalizing" the labels simply by renumbering them wouldn't make it. Instead, it could make it harder. Imagine:
```
L1: op1 L4
L2: op2 L4
L3: op3     ; what if this line is removed by the optimizer?
L4: op4
L5: op5
```
After removing one line together with a label 
```
L1: op1 L4
L2: op2 L4
; only this removed line is the difference
L4: op4
L5: op5
```
But after renumbering all remaining labels:
```
L1: op1 L3
L2: op2 L3
; oops, now every line has changed!
L3: op4
L4: op5
```
Now every label after the removed one will be changed, which doesn't seem to be helpful with diff'ing.

The way I see could help with diff'ing is to "isolate" the changes. For example, to make "local labels" truly local to the functions where they live in, and use different naming schemes for different types of data/code.

mrdudz:
I dont see how this could work, because what gregg said.

Using differently named labels for different "things" (types, branch targets, locals/globals, whatever) seems like it could improve readability a lot though.

oliverschmidt:
I personally don't see any point in optimizing diffs of generated .s files. I'd be VERY reluctant to merge a PR adding code to maintained just for that feature. I personally would prefer to close this issue.

clbr:
The normalization pass would happen after any optimizations, therefore I don't see how Greg's reply would prevent it from working. It would be an improvement even in acqn's case I believe, because instead of changing almost every label in the file, the changes would start from the point you changed. Making it much faster to find what really changed.

Different pools for labels would help, but even in that case normalization would be necessary; it would just reduce the change clutter to inside one function, for example.

acqn:
[...]

Issue #1196, opened 2020-08-17
[cc65] Constant expressions in general


acqn:
The term "constant" is overloaded with meanings. This funfortunately creates confusions and leads to issues in cc65. We'd like to have some clean-ups.

Related Issue: #250.

Let's just forget about "C Language constant-expression" and see what the compiler can do.

Properties related to "constant-ness"of an expression:
- `Pure`               (Pr): No side-effects (borrowed concept from "pure function").
- `Static`             (St): No runtime code needed.
- `Immutable`          (Im): Never changes once determined.
- `Compile-Time-Known` (Kn): Value of which known to the compiler front end.

What is the relationship between them?
- `Static` -> `Pure`
- `Static` -> `Immutable`
- `Compile-Time-Known` -> `Immutable`
- `Immutable` == `Static` || `Compile-Time-Known`
- "C Language constant-expression" ->` Static` && `Compile-Time-Known`

Given the declarations:
```c
int a;
volatile int b;
const int c = 1;
#define d 1
enum { e = 1 };
int f() { return 1; }
extern int g();
```
The compiler can be sure of the value of the top-level expression for each statement (yeah I know these expression values are unused for the statements, but let's just focus on the expressions themselves) is:
```c
1;              /* Pure: Yes; Static: Yes;  Immutable: Yes; Compile-Time-Known: Yes  */
a;              /* Pure: Yes; Static: No;   Immutable: No;  Compile-Time-Known: No   */
b;              /* Pure: No?; Static: No;   Immutable: No;  Compile-Time-Known: No   */
c;              /* Pure: Yes; Static: No?;  Immutable: No?; Compile-Time-Known: No?  */
d;              /* Pure: Yes; Static: Yes;  Immutable: Yes; Compile-Time-Known: Yes  */
e;              /* Pure: Yes; Static: Yes;  Immutable: Yes; Compile-Time-Known: Yes  */
a == a;         /* Pure: Yes; Static: Yes;  Immutable: Yes; Compile-Time-Known: Yes  */
b == b;         /* Pure: No?; Static: No;   Immutable: No;  Compile-Time-Known: No   */
c == c;         /* Pure: Yes; Static: Yes;  Immutable: Yes; Compile-Time-Known: Yes  */
f();            /* Pure: Yes; Static: Yes?; Immutable: Yes; Compile-Time-Known: Yes? */
f() == f();     /* Pure: Yes; Static: Yes?; Immutable: Yes; Compile-Time-Known: Yes? */
[...]

PR #1220, opened 2020-08-26
[cc65] Fixed constant boolean expressions

### LINKED ISSUE #250: Array size compile-time optimization stops halfway
The following file errors out on line 3. It compiles fine on gcc with -ansi -pedantic, which forces it to hard c89 mode.

```
#define LZO_MAX(a,b)        ((a) >= (b) ? (a) : (b))
unsigned char c[2*4];
unsigned char b[2*LZO_MAX(8,sizeof(int))];
```

### LINKED ISSUE #1123: "Result of comparison is constant" warning in constant expressions

### LINKED ISSUE #1221: [cc65] Structs/unions as ternary operands
```c
int a;
struct S { int a; } s1, s2;
struct U { int a; } u1, u2;

int main()
{
    a ? s1 : s2; /* BUG: should be OK */
    a ? u1 : u2; /* BUG: should be OK */
    return 0;
}
```

acqn:
(8/8) commits.
- [x] Moved the almighty `bug250.c` from `test/misc` to `test/val`.
  - [x] Added integer boolean type string - no longer set the "expression tested" flag with constant results in comparison.
  - [x] Fixed #1123 - no more warnings on "const comparisons" that are desired in constant context.
  - [x] Warnings on expressions with no effects instead of statements with no effects.
  - [x] Fixed #250 (the OP part) - const ternary operator.
  - [x] Fixed #1221 - ternary couldn't use struct/union types as the true/false-branch expressions.
  - [x] Fixed #250 (https://github.com/cc65/cc65/issues/250#issuecomment-658003430) - logical AND and logical OR.
- [x] Added a constant AND/OR codegen test.

---
It might be worth mentioning that despite 1d7fcc67a498f65a4687099b87a40d402c039c78 saves nearly half of the labels, the new testcase d4025efaba798c652302bb49fb45506f8504b5c0 still easily reaches a big label number (>= L1131). The enhancements suggested in #301 might be desired.


acqn:
I think I really hate `test/misc`.

acqn:
Huh? So Travis CI doesn't respect the order in which the commits are pushed, but just reschedules the tests according to the "timestamps" of the commits? @oliverschmidt

acqn:
I give up. Enjoy the commit fusion.

acqn:
It seems to me that the "random" build failure tends to happen when there are multiple tests scheduled. I don't know if that indicates some hidden memory corruption bug since I can't reproduce it on my PC.

oliverschmidt:
> Huh? So Travis CI doesn't respect the order in which the commits are pushed, but just reschedules the tests according to the "timestamps" of the commits? @oliverschmidt

Either I get you wrong or I'd be pretty surprised if it were as you describe. The picture I have in mind is that Travis CI doesn't know anything about Git. When something is pushed to the upstream repo, a test of what was pushed is triggered. However, If your experience contradicts my understanding then my understanding is likely wrong...

jmr:
It says it builds the merge (so not exactly what was pushed if HEAD is different), but I'm not sure if that helps to explain anything.
https://docs.travis-ci.com/user/pull-requests/#how-pull-requests-are-built

oliverschmidt:
> It seems to me that the "random" CI build failure (https://travis-ci.org/github/cc65/cc65/jobs/721197860) tends to happen when there are multiple tests scheduled. I don't know if that indicates some hidden memory corruption bug or not since I can't reproduce it on my PC.

I re-ran that Travis CI job quite some times. It succeeded each time.

acqn:
> 
> 
> > It seems to me that the "random" CI build failure (https://travis-ci.org/github/cc65/cc65/jobs/721197860) tends to happen when there are multiple tests scheduled. I don't know if that indicates some hidden memory corruption bug or not since I can't reproduce it on my PC.
> 
> I re-ran that Travis CI job quite some times. It succeeded each time.

Could you re-run https://travis-ci.org/github/cc65/cc65/jobs/721079643?

acqn:
> 
> 
> It says it builds the merge (so not exactly what was pushed if HEAD is different), but I'm not sure if that helps to explain anything.
> https://docs.travis-ci.com/user/pull-requests/#how-pull-requests-are-built

It appears to me that the CI "caches" the results of some _commits_ (instead of "merged snapshots") and reuses them without redoing the tests.

acqn:
True order:
1. Added integer boolean type string. 17bbba7
2. Moved bug250.c from test/misc as it is to be fixed. b15ab34
3. Fixed warnings on const comparison. 3ea3887
4. Moved bug250.c from test/misc as it is to be fixed. b15ab34
5. Fixed constant expression checks with no-code requirement. 7255111
6. Fixed logical AND and logical OR. 85ea066
7. Avoided generating unnecessary true-case labels in logic AND/OR. f289ea6
8. Added test/ref/pr1220 for testing constant AND/OR code generation. b6a4715

oliverschmidt:
> Could you re-run https://travis-ci.org/github/cc65/cc65/jobs/721079643? That one was weird that it fore-triggered a bug pushed after the commits it was testing.

I re-ran that Travis CI job quite some times. It succeeded each time.

oliverschmidt:
> I give up. I don't want to fight test/misc, GitHub and Travis CI at the same time. 

If I understand things right, Travis CI entered the arena because the cc65 build wasn't fully parallelizable. Therefore I worked on that issue. I now closed https://github.com/cc65/cc65/issues/1080 and presume that you can therefore revert to a local build.

acqn:
Could it be possible that the failed tests were caused by flawed cc65 executable builds? The only way a revision test could demonstrate a "future" bug would be:
1. There was a different hidden bug that could demenstrate exactly the same symptoms.
2. The test was indeed done with the future commit built into the executable.

src/cc65/datatype.c
```diff
@@ -66,6 +66,7 @@ Type type_int[]         = { TYPE(T_INT),    TYPE(T_END) };
 Type type_uint[]        = { TYPE(T_UINT),   TYPE(T_END) };
 Type type_long[]        = { TYPE(T_LONG),   TYPE(T_END) };
 Type type_ulong[]       = { TYPE(T_ULONG),  TYPE(T_END) };
+Type type_bool[]        = { TYPE(T_INT),    TYPE(T_END) };
```
jmr, line 69: What should happen to `type_bool` if `_Bool` is introduced? Then it becomes `TYPE(T_INT)` or we just need to figure it out later?

src/cc65/exprdesc.h
```diff
@@ -104,29 +104,58 @@ enum {
...
 
+    /* Expression evaluation requirements.
+    ** Usage: (Flags & E_EVAL_<Flag>) == E_EVAL_<Flag>
+    */
+    E_MASK_EVAL         = 0xFC0000,
+    E_EVAL_NONE         = 0x000000,     /* No requirements */
+    E_EVAL_CONST        = 0x040000,     /* Result must be immutable */
+    E_EVAL_COMPILE_TIME = 0x0C0000,     /* Result must be known at compile time */
+    E_EVAL_PURE         = 0x100000,     /* Evaluation must have no side effects */
+    E_EVAL_STATIC       = 0x340000,     /* Evaluation must generate no code */
```
jmr, line 144: I think having `STATIC` and `CONST` names that mean different things from `static` and `const` in C is a bit confusing.

src/cc65/expr.h
```diff
@@ -54,23 +54,23 @@ int evalexpr (unsigned flags, void (*Func) (ExprDesc*), ExprDesc* Expr);
...
 ** printed, and the value is replaced by a constant one to make sure there
 ** are no internal errors that result from this input error.
 */
 
-void ConstAbsIntExpr (void (*Func) (ExprDesc*), ExprDesc* Expr);
-/* Will evaluate an expression via the given function. If the result is not
-** a constant numeric integer value, a diagnostic will be printed, and the
+ExprDesc StaticConstExpr (void (*Func) (ExprDesc*));
+/* Get an expression evaluated via the given function. If the result is not a
+** static constant expression, a diagnostic will be printed, and the value is
```
jmr, line 66: What is a "static constant expression"? Explain or include pointers to which definitions of these terms you're using.

src/cc65/datatype.c
acqn: It would use a new `T_BOOL` (that is  `T_TYPE_BOOL | T_CLASS_INT | T_SIGN_UNSIGNED | T_SIZE_BOOL`, and probably a `BOOL_BITS` that equals `1`. It would need special handling like `char` and `enum`.

src/cc65/expr.h
acqn: It could be read at the definition of `E_EVAL_C_CONST`. `Static` in such a context means "evaluated without any runtime code generated" as in #1196. What about changing the phrase
> static constant expression

to
> constant expression needing no runtime code

?

oliverschmidt:
@acqn: Do you consider the two remarks from @jmr resolved? If yes, please mark them as such.

acqn:
Not yet. I can rename them, but I haven't come up with good names.

src/cc65/exprdesc.h
acqn: There have been different meanings of `const`/`constant` in cc65 for a long time, which I think should be enough to keep people aware of that. And `static` in C mostly refers to object linkage rather than expression traits, except for `_Static_assert`, which exactly indicates the same "no runtime" virtual as used here.

Anyways, do you have suggestions for better names? `E_EVAL_IMMUTABLE` could be  good for `E_EVAL_CONST`, but `E_EVAL_NO_CODE`, `E_EVAL_CODELESS` or `E_EVAL_EMPTY_CODE` for `E_EVAL_STATIC`?

jmr: I didn't realize, but many/most of these are currently unused.

It would definitely help to expand the comments. "must be immutable" doesn't tell me much. Is `(i++, 0)` immutable? Having good documentation is more important than getting the right names right now. 

`E_EVAL_NO_CODE` seems good for `E_EVAL_STATIC`. I'll try to think about the other names.

src/cc65/expr.h
```diff
@@ -54,25 +54,25 @@ int evalexpr (unsigned flags, void (*Func) (ExprDesc*), ExprDesc* Expr);
...
 */
 
-void ConstAbsIntExpr (void (*Func) (ExprDesc*), ExprDesc* Expr);
-/* Will evaluate an expression via the given function. If the result is not
-** a constant numeric integer value, a diagnostic will be printed, and the
-** value is replaced by a constant one to make sure there are no internal
-** errors that result from this input error.
+ExprDesc StaticConstExpr (void (*Func) (ExprDesc*));
+/* Get an expression evaluated via the given function. If the result is not a
+** constant expression without runtime code generated, a diagnostic will be
```
jmr, line 66: Can you say something like "constant expression as defined by `E_EVAL_...`?

src/cc65/exprdesc.h
```diff
@@ -104,29 +104,58 @@ enum {
...
+    E_NEED_TEST         = 0x020000,     /* Expression needs a test to set cc */
 
+    /* Expression evaluation requirements.
+    ** Usage: (Flags & E_EVAL_<Flag>) == E_EVAL_<Flag>
+    */
+    E_MASK_EVAL         = 0xFC0000,
+    E_EVAL_NONE         = 0x000000,     /* No requirements */
+    E_EVAL_CONST        = 0x040000,     /* Result must be immutable */
+    E_EVAL_COMPILE_TIME = 0x0C0000,     /* Result must be known at compile time */
+    E_EVAL_PURE         = 0x100000,     /* Evaluation must have no side effects */
```
jmr, line 143: Pure usually means no side effects and same result for same inputs. `E_EVAL_NO_SIDE_EFFECTS` is a bit long, but `*p` isn't pure, is it?

src/cc65/exprdesc.h
acqn: The comment says "**Result** must be immutable". The result of the expression `(i++, 0)` is always `0`. Thus it is immutable.

src/cc65/expr.h
acqn: What about both?

src/cc65/exprdesc.h
acqn: The concept "Pure" was borrowed from "pure function" but it is about expressions, not just functions. It's true that the names could have better substitutes. I'll go with `E_EVAL_NO_SIDE_EFFECTS`.

jmr:
Thanks! The new comments are helpful.

acqn:
There was a potential subexpression issue in ternary operation similar to #1181. Now fixed in 74def4608aab0065d118e4a072e8aab9b6918cd6.

acqn:
There were two bugs with 74def4608aab0065d118e4a072e8aab9b6918cd6. Incoming hotfix.

PR #1228, opened 2020-08-29
[cc65] 'void' extension fixes

### LINKED ISSUE #1144: [cc65] How far do we extend support for `void` type family of objects?
Split from Issue #1142 (with modifications):
```c
void ex0    = { 0x1234, (char)12 };     /* our extension */
void ex1[1] = { {0x1234, (char)12} };   /* initialized 'void []', OK or not? */
extern void z[];                        /* forward declaration of 'void []'? */
```
Do we support `void`  object beyond the simplest `void` object? Do we support `void []`?

There are instances of usage of forward declarations of `void []` in the library code. So I suppose we at least support that too? But is that a real array of `void` elements, or just some pointer to `void` objects in disguise?

acqn:
(4/4) It's easier to leave `void` arrays supported than to forbid them while keep supporting `extern void` arrays....
- [x] Disallowed void arrays of elements of variant sizes.
- [x] Improved diagnostic info on assignment to void types.
- [x] Recursively checking for incomplete/unknown-sized types.
- [x] Disabled struct/union fields of 'void' type.

Resolved #1144.

```c
extern void v[];                            /* OK */
void ex      = { 12, (char)34 };            /* OK */
void ex1[1]  = { {12, (char)34} };          /* OK */
void ex2a[2] = { {12, (char)34}, {(char)56, (char)78, (char)90} };  /* OK - each "row" is 3 bytes in size */
void ex2b[2] = { {12, (char)34}, { 56 }, }; /* Error: 'void' array initialized with elements of variant sizes */
void ex3[3]  = { {12, (char)34}, };         /* OK - the last two "rows" will each be 3 bytes in size and zero-filled */

void x[][2] = { { { 1 } , { 2 }} };

struct S {
    int a;
    void v[1];                              /* Error: Field 'v' has incomplete type 'void[1]' */
};

union U {
    int a;
    void v;                                 /* Error: Field 'v' has incomplete type 'void' */
};
```

### DIFF
--- a/src/cc65/datatype.c
+++ b/src/cc65/datatype.c
@@ -979,7 +979,7 @@ int IsClassIncomplete (const Type* T)
 /* Return true if this is an object type lacking size info */
 {
     if (IsTypeArray (T)) {
-        return GetElementCount (T) == UNSPECIFIED;
+        return GetElementCount (T) == UNSPECIFIED || IsClassIncomplete (T + 1);
     }
     return IsTypeVoid (T) || IsIncompleteESUType (T);
 }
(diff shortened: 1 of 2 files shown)


oliverschmidt:
It seems to me that this PR doesn't ask for feedback - merging...

PR #1230, opened 2020-08-30
[cc65] Fixes for some optimization steps

acqn:
(2/2)
- [x] Fixed OptPush1 in case later code would rely on that pushax zeroes register Y.
There was the possibility that code following a `pushax` would rely on that register Y is set to `0` by it, whilst the opimization result `pushwysp` wouldn't set Y to 0. This fix checks for such cases and keeps the invariant on Y.
- [x] OptDupLoads shouldn't silently do optimizations.
`OptDupLoads` not only deletes code entries but also replaces the opcode at times, which was not correctly reported with `--debug-opt-output`. This PR fixes that.

### DIFF
--- a/src/cc65/coptind.c
+++ b/src/cc65/coptind.c
@@ -1147,8 +1147,9 @@ unsigned OptDupLoads (CodeSeg* S)
         /* Get next entry */
         CodeEntry* E = CS_GetEntry (S, I);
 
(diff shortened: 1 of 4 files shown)


oliverschmidt:
No feedback for 5 days (beside the rename by @greg-king5) -> merging...

PR #1232, opened 2020-08-31
[cc65] Fixed evaluation flags propagation

acqn:
(3/3)
- [x] Fixed AND/OR in certain cases where the 'E_NEED_TEST' flag set for usage only in subexpressions should be cleared.
Still the same old "reused subexpression" bug. Sigh.
- [x] Fixed evaluation flags propagation to subexpressions in assignments and function calls.
- [x] Updated `test/ref/pr1220.c` for testing this.

### DIFF
--- a/src/cc65/assignment.c
+++ b/src/cc65/assignment.c
@@ -136,10 +136,11 @@ static int CopyStruct (ExprDesc* LExpr, ExprDesc* RExpr)
 void Assignment (ExprDesc* Expr)
 /* Parse an assignment */
 {
-    ExprDesc Expr2;
     Type* ltype = Expr->Type;
 
+    ExprDesc Expr2;
     ED_Init (&Expr2);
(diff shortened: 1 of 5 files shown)


mrdudz:
It would be really nice to add a test for the stuff that was fixed too - so what was it exactly? :)

greg-king5:
:+1:

oliverschmidt:
Thumbs up from @greg-king5, no other feedback -> merging...

acqn:
> 
> 
> It would be really nice to add a test for the stuff that was fixed too - so what was it exactly? :)

When `expr1a` in `expr1a && expr1b ? expr2 : expr3` or `expr1a || expr1b ? expr2 : expr3` is non-constant and the result `expr2` or `expr3` is not boolean. d9ab002d1219b8e7c0d9e3ed4b6e5da9af6e3159 now tests such cases.

PR #1233, opened 2020-08-31
[cc65] Fixed logical-NOT in constant context.

acqn:
(3/3) See also Issue #1196.
- [x] Fixed usage of logical-NOT in constant context.
- [x] Object addresses as non-NULL pointers in constant context.
This PR makes `&object != NULL` as a true constant value.

### DIFF
--- a/src/cc65/expr.c
+++ b/src/cc65/expr.c
@@ -1911,6 +1911,9 @@ void hie10 (ExprDesc* Expr)
             if (ED_IsConstAbs (Expr)) {
                 /* Constant expression */
                 Expr->IVal = !Expr->IVal;
+            } else if (ED_IsAddrExpr (Expr)) {
+                /* Address != NULL, so !Address == 0 */
+                ED_MakeConstBool (Expr, 0);
             } else {
                 /* Not constant, load into the primary */
                 LoadExpr (CF_NONE, Expr);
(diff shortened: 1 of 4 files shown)


acqn:
@oliverschmidt Time-travel happened again. This time the CI checked ee158edfc9816fea36a0d12184d63175b71559c0 with the binary built from 869973de6044577314818da32dae0ccbc98c5828. Note that the failure was the result of a "_successful compilation_" of a non-standard-complian C source file, the opposite of those last two times, so it's more likely to confirm the theory that there is a bug/mis-feature in the CI rather than in the cc65.

acqn:
So I decided to keep relation operations between function pointes unsupported.

greg-king5:
> @oliverschmidt, would you please rerun the checks on c0b1476592c4349d3dafac299303092baa7f8ee4 and 6984d0cc0ec2802882fd803ae9334278d196cab1? It was likely that the CI got the wrong checks.

Re-running them is pointless.  Travis CI never checks a particular commit.  It always builds the _current_ state of the PR!  Therefore, we would see the result of commit 583331161cee3f0231028897b7c2689fb17c6c73.

oliverschmidt:
No content-related feedback in 4 days -> merging...

oliverschmidt:
Is issue #1196 to be considered fixed?

acqn:
> 
> 
> Is issue #1196 to be considered fixed?

#1196 includes part outside of the standard (i.e. exntsion). The required part (#250 and object address != 0) by the standard is fixed I think. So #1196 is still an issue for question/enhancement.

oliverschmidt:
Thanks for the feedback - I adjusted the #1196 classification.

PR #1235, opened 2020-09-03
Fix build on systems without <stdint.h>

Tevo45:
While porting cc65 over to Plan 9, I noticed that the compiler would not build when `<stdint.h>` was not included (due to certain macros not being defined by default, in my case), complaining about the missing `INT*_{MIN,MAX}` macros used on `src/cc65/declare.c`. Since there already is a compile-time branch for cases where the header is (supposedly) missing and the fact that given values don't vary between systems (I'd be surprised to see otherwise), I found it reasonable to just `#define` the missing constants. I just included the ones used in the code, but it might be desirable add the missing ones as well, for proofing against possible future use of the macros.

### DIFF
--- a/src/common/inttypes.h
+++ b/src/common/inttypes.h
@@ -38,7 +38,8 @@
 
 
 
-/* If we have <stdint.h>, include it; otherwise, adapt types from <stddef.h>.
+/* If we have <stdint.h>, include it; otherwise, adapt types from <stddef.h>
+** and define integer boundary constants.
 ** gcc and msvc don't define __STDC_VERSION__ without special flags, so check
 ** for them explicitly.  Undefined symbols are replaced by zero; so, checks for
 ** defined(__GNUC__) and defined(_MSC_VER) aren't necessary.
--- a/src/common/inttypes.h
+++ b/src/common/inttypes.h
@@ -57,6 +57,16 @@ typedef size_t uintptr_t;
 typedef ptrdiff_t intmax_t;
 typedef size_t uintmax_t;
 
+#define INT8_MAX (0x7F)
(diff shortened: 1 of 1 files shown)


jmr:
But plan9 does have `<stdint.h>`, right?
https://9p.io/sources/plan9/sys/include/ape/inttypes.h

It looks like you might need to use `pcc` to get the `ape/` headers:
http://doc.cat-v.org/plan_9/4th_edition/papers/ape

So update `#if (__STDC_VERSION__ >= 199901) || (__GNUC__ >= 3) || (_MSC_VER >= 1600)` to do also detect plan9?
https://github.com/cc65/cc65/blob/master/src/common/inttypes.h#L46

Your fix looks like it's still needed for systems that really don't have `<inttypes.h>`.

polluks:
Well, stdint simply means C99. I did my Amiga port of version 2.9.2 with a C89 compiler.
A port without APE feels more native IMHO.

Tevo45:
> But plan9 does have `<stdint.h>`, right?

It does, it just wasn't getting included by default. I ended up defining `__STDC_VERSION__` as `199901L` on CFLAGS. I don't think it should be a problem, since pcc/APE are mostly C99-compliant anyway. The compiler also doesn't define any macros for identifying the platform by default, so adding a check on the `#if` would still involve manually defining something else first.

oliverschmidt:
The comment above says `otherwise, adapt types from <stddef.h>`. With this change that comment isn't true anymore.

Tevo45:
> The comment above says `otherwise, adapt types from <stddef.h>`. With this change that comment isn't true anymore.

Updated

PR #1236, opened 2020-09-03
Change line endings from CRLF to LF

jmr:
test/ref/pr1220.c was somehow added with CRLFs.
Other files use just LF.


oliverschmidt:
I appreciate your effort, but this PR seems to conflict.

jmr:
Ok. What about now?

oliverschmidt:
Fine :-)

PR #1238, opened 2020-09-04
Allow char bit-fields

### LINKED ISSUE #1047: [feature request] char bitfields
Given the importance of keeping data as small as possible, it would be nice if char bitfields were supported, even though they are an extension. [Most](https://godbolt.org/z/ASMuuU) [compilers](https://godbolt.org/z/CyjbA6) [support](https://godbolt.org/z/y-9PjW) [them](https://godbolt.org/z/c7XMKd).

```
struct s {
    unsigned char x : 2;
    unsigned char y : 3;
};

int f(void) { return sizeof(struct s); }
```

[This currently produces](https://godbolt.org/z/o5Es6Y):
```
<source>(2): Error: Bit-field has invalid type
<source>(3): Error: Bit-field has invalid type
```

jmr:
These are not required to be supported (only int, signed int, and
unsigned int are required), but most compilers support it.

https://port70.net/~nsz/c/c89/c89-draft.html#3.5.2.1
https://port70.net/~nsz/c/c89/c89-draft.html#A.6.5.8

Fixes #1047


jmr:
I'll fix the test.

src/cc65/declare.c
```diff
@@ -767,9 +767,9 @@ static int ParseFieldWidth (Declaration* Decl)
     /* TODO: This can be relaxed to be any integral type, but
     ** ParseStructInit currently only supports up to int.
     */
-    if (SizeOf (Decl->Type) != SizeOf (type_uint)) {
+    if (SizeOf (Decl->Type) > SizeOf (type_uint)) {
         /* Only int sized types may be used for bit-fields for now */
-        Error ("cc65 currently only supports unsigned int bit-fields");
+        Error ("cc65 currently only supports char and int bit-fields");
```
greg-king5, line 772: I think that it should say, `"cc65 currently supports only char-sized and int-sized bit-fields"`.  (We don't want to discourage people from using `short`.)

greg-king5:
Please mention this extension in https://cc65.github.io/doc/cc65.html#s5.

polluks:
By the way our HTML looks fine on mobile phone but the PRE-tags...
Any CSS wizard here?

jmr:
> By the way our HTML looks fine on mobile phone but the PRE-tags...
> Any CSS wizard here?

Probably better just to rewrite as `.md`.

oliverschmidt:
Makefile:41: recipe for target '../../testwrk/err/bug1047.s' failed
4850make[2]: *** [../../testwrk/err/bug1047.s] Error 1
4851Makefile:20: recipe for target 'continue' failed

oliverschmidt:
> Probably better just to rewrite as .md.

Certainly not.

jmr:
> Makefile:41: recipe for target '../../testwrk/err/bug1047.s' failed
> 4850make[2]: *** [../../testwrk/err/bug1047.s] Error 1
> 4851Makefile:20: recipe for target 'continue' failed

Yes, that's what I meant by ["I'll fix the test."](https://github.com/cc65/cc65/pull/1238#issuecomment-687187118) I've pushed a fix and the CI is running again.

src/cc65/declare.c
jmr: Updated, but note that `short` bit-fields are treated as `unsigned short` like happens with int. Not sure if this is considered a bug since we're already in undefined extension territory.

acqn: Is every extended bit-field type (including `long`) declared without an `unsigned` treated as `unsigned`? It'll be OK if that's consistent.

jmr:
> > Probably better just to rewrite as .md.
> 
> Certainly not.

Then it wouldn't take a CSS wizard to make it look decent.

src/cc65/declare.c
jmr: `long` isn't currently supported, but the most obvious way to add it would make `long` bitfields unsigned. So, I agree, not a bug.

doc/cc65.sgml
```diff
@@ -807,6 +807,10 @@ This cc65 version has some extensions to the ISO C standard.
         <tt/_Static_assert/ is also available as the macro <tt/static_assert/ in
         <tt/assert.h/.
 
+<item>  cc65 supports bit-fields of any integral type that is int-sized or
+        smaller, and enum types with such types as their underlying type.
+        (Only <tt/int/, <tt/signed int/, and <tt/unsigned int/ are required.)
```
greg-king5, line 812: Perhaps a small clarification should be added here: `"... are required by the standard."`.

acqn: I think "enum types with such types as their underlying type" as documented here should be tested.

jmr: > I think "enum types with such types as their underlying type" as documented here should be tested.

Yes, these tests have been missing for years. 

Enum bit-fields apparently used to work: https://godbolt.org/z/KnEbcq

(Godbolt is using f95481f from 1.5 years ago as of when I made that link.)

Probably more tests for different underlying types should have been added when you did your enum work.

It will take a little work to fix the enum bit-fields, so I have just deleted mention of enum bit-fields for now. It's orthogonal to this PR anyway.

acqn: I wasn't aware of its missing from the tests either. It appeared to be working (aside from #1047) when I modified the check on bit-field type size. So I thought it was all right.

Anyways, are you going to work on fixing enum bit-fields?

jmr: I appear to have broken it, so I'll try to fix it.

oliverschmidt: > I appear to have broken it, so I'll try to fix it.

Do you see that fix part of this PR? Or do you consider this PR as ready to be merged?

jmr: Better to fix #1244 first, separately.

src/cc65/symtab.c
```diff
@@ -867,9 +867,10 @@ SymEntry* AddBitField (const char* Name, const Type* T, unsigned Offs,
         Entry->V.B.BitOffs  = BitOffs;
         Entry->V.B.BitWidth = BitWidth;
 
-        if (!SignednessSpecified) {
+        if (!SignednessSpecified && !IsTypeChar (Entry->Type)) {
```
acqn, line 870: This way of handling seems to be coupled with the underlying type of `char` (it could be `signed char` if compiled with `--signed-chars`).

I think whatever underlying signedness should be "normalized" to `unsigned` as long as it is not specified explicitly, for the reason that `char` bit-fields are integers rather than characters. This would keep consistency with other standard-integer-type bit-fields but leave `enum`-type bit-fields unique (for a sensible reason).

jmr: > `char` bit-fields are integers rather than characters.

What do you mean by this? `char` is both a character type and an integral type.

acqn: https://port70.net/~nsz/c/c99/n1256.html#6.7.2.1p9
> A bit-field is interpreted as a signed or unsigned integer type consisting of the specified number of bits.

I consider the idea "char bit-fields" to be an extension to allow (`signed`/`unsigned`) `char` to be used to declare bit-fields, which have no other difference from corresponding standard (`signed`/`unsigned`) `int` bit-fields other than how they are declared, which are just required to be interpreted as integer types but not character types.

jmr: Anyway, I made `char` unsigned in bit-fields, even with `--signed-chars`.

doc/cc65.sgml
oliverschmidt: https://github.com/cc65/cc65/issues/1244 is closed by now...

oliverschmidt:
As far as I can tell all improvement proposals have been taken into account -> merging...

Issue #1203, opened 2020-08-18
Possible special case optimization for basic arithmetic


mrdudz:
i have code like this quite often:
```
foo += 0x800;
```
ie, adding (or substracting) from a variable, with the lowbyte of the constant being 0 (or even the lower 16 or 24 bits). the code generated for it looks like this:
```
	lda     #$00
	clc
	adc     foo
	sta     foo
	lda     #$08
	adc     foo+1
	sta     foo+1
```
that is fine and correct of course - but we can safely skip adding the bytes that are 0, and generate something like this instead:
```
	lda     #$08
	clc
	adc     foo+1
	sta     foo+1
```
again, i have no idea how hard it is to do - but it would be nice to have :)

acqn:
A new optimizer step would make it.

jmr:
It looks like this should be doable in `g_addeqstatic` (and possibly other variants as well).

acqn:
I suggest to do this as an optimizer step instead of a code generation change, since there are some other optimizer steps that can only recognize certain few code patterns of additions.

acqn:
I was testing with the wrong revision of the "Optimization" branch... With the correct revision it gives:
```asm
	clc
	lda     _foo ; TODO: this could be removed
	sta     _foo ; TODO: this could be removed
	lda     #$08
	adc     _foo+1
	sta     _foo+1
```

oliverschmidt:
* I presume we all agree that the cc65 optimizer should in general assume that there are no side effects associated with reading or (re-)writing.

* I personally think that it would be a great use case for `volatile` to say that the cc65 optimizer guantees no spurious accesses.

kugelfuhr:
This can be closed. For the following code
```
unsigned foo;
int main(void)
{
    foo += 0x800;
    return 0;
}
```
the current compiler generates
```
        clc
        lda     #$08
        adc     _foo+1
        sta     _foo+1
        ldx     #$00
        txa
        rts
```

mrdudz:
Thanks for noting!

PR #1239, opened 2020-09-04
[cc65] Enhanced OptStackOps optimization

acqn:
(3/3) This was one of my oldest attempts on optimizer enhancements (refreshed once this year). It basically skips the comparison operations on the hi-bytes if they are equal. This is similar to _but not the same as_ that requested in Issue #1203 (which  would skip operations on the lo-bytes instead).

An enhanment to allow OptStack to optimize more cases on the stack is also stuffed in.

- [x] Utility functions about compare conditions.
- [x] Improved OptStackOps for optimizating further when operands have equal hi-bytes.
- [x] Extended support for more addressing modes in tos* optimizations.


oliverschmidt:
No feedback in 4 days -> merging...

PR #1240, opened 2020-09-04
[cc65] Sizeof fixes and optimizations

### LINKED ISSUE #1222: [cc65] 'sizeof' issues
`sizeof` should be applicable to most expressions, except when the expressions have incomplete types, bit-fields, or function designators.
```c
int a[1];
int b[sizeof ++a[42]]; /* should work as '++a[42]' is actually unevaluated */
```
A fix (???) is done (depending on PR #1220).

EDIT 1: Er, I am not sure what was the cause, but apparently the "BUG" in the OP doesn't happen in either of the current release version or the current HEAD revision. Nor in any revision that I've tested today. That's to say, I can't reproduce the "BUG" anymore! It could be a prototype/interim revision during PR #1220 where `sizeof` was once broken.

EDIT 2: The BUG was the result of a failure to propagate the `E_EVAL_UNVEAL` flag in expressions. It was fortunately prevented in the PRs and can't rise again as long as such a failure is no more around.

acqn:
(2/2) Resolved #1222.
- [x] Std-functions are no longer inlined if they are unevaluated.
- [x] Avoided referencing string literals with sizeof so that they are not output if unused elsewhere.

### DIFF
--- a/src/cc65/expr.c
+++ b/src/cc65/expr.c
@@ -593,7 +593,7 @@ static void FunctionCall (ExprDesc* Expr)
         }
 
         /* Check for known standard functions and inline them */
-        if (Expr->Name != 0) {
+        if (Expr->Name != 0 && !ED_IsUneval (Expr)) {
             int StdFunc = FindStdFunc ((const char*) Expr->Name);
             if (StdFunc >= 0) {
                 /* Inline this function */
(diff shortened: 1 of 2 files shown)


greg-king5:
:+1:

PR #1243, opened 2020-09-05
[cc65] Fixed check for processor flags usage in case of PHP

acqn:
(1/1) Opcode `php` uses the Z/N flags. Although this opcode isn't being generated by cc65 at the moment, it may appear in asm().

### DIFF
--- a/src/cc65/codeent.c
+++ b/src/cc65/codeent.c
@@ -476,6 +476,11 @@ int CE_UseLoadFlags (CodeEntry* E)
         }
     }
 
+    /* PHP will use all flags */
+    if (E->OPC == OP65_PHP) {
+        return 1;
+    }
+
     /* Anything else */
     return 0;
 }


oliverschmidt:
Sorry for the stupid question, but isn't this rather PLP?

acqn:
PHP pushes the current processor flags onto the stack: `Flags (read) -> Stack (write)`.
PLP pulls the processor flags from the stack: `Stack (read) -> Flags (write)`.

Therefore, PHP is a Z/N Flags reader (user), while PLP is a Z/N Flags writer (changer).

oliverschmidt:
> In the vocabulary of the cc65 code base (according to my reading), "use" means to read from, "change" means to write to, whilst "usage" refers to both "using" + "changing".

Ah, now I see. Thanks for the explanation.

PR #1247, opened 2020-09-06
[cc65] Reorganized optimization steps

acqn:
Moved all optimization steps from `cc65/codeopt.c` and some optimization steps from `cc65/coptind.c` into new separate files.


oliverschmidt:
This PR moves a significant amount of code - claiming to not change the behavior. I don't presume feedback coming in. So there's no point in waiting to merge...

PR #1248, opened 2020-09-07
Fix enum bit-field ICE #1244

### LINKED ISSUE #1244: ICE for enum bit-fields
> I triggered a check failure in https://github.com/cc65/cc65/blob/bb5b7f45921154b1655a6ba8648edda12025059e/src/cc65/symtab.c#L875 with
> ```c
> enum E {
>     L = 65535L  /* or U = 65535U */
> };
> 
> struct S {
>     enum E a : 16;
> } s;
> ```
> It seems a bit unclear to me whether this implicit `unsigned` "promotion" should apply to `enum`s. I would expect enum bit-fields to be intuitively using the same signedness of their underlying types (since we don't have `signed enum` or `unsigned enum`), but that could be just me.
> 
> Anyways, the underlying types of `enum`s should be checked with `GetUnderlyingType()`, not direct `->Type`.

_Originally posted by @acqn in https://github.com/cc65/cc65/pull/1238#discussion_r483909587_

jmr:
This previously resulted in an ICE:
```
cc65: Check failed: (Entry->Type->C & T_MASK_SIGN) == T_SIGN_SIGNED,
file 'cc65/symtab.c', line 874
```

This `CHECK` is in the code that deals with changing `int` bitfields to
`unsigned int`.

Work around this by treating enum bit-fields as having their signedness
specified, so this type change code does not get called.

Fixes #1244.

### DIFF
--- a/src/cc65/declare.c
+++ b/src/cc65/declare.c
@@ -1446,6 +1446,12 @@ static void ParseTypeSpec (DeclSpec* D, long Default, TypeCode Qualifiers,
             D->Type[0].C |= T_ENUM;
             SetESUSymEntry (D->Type, Entry);
             D->Type[1].C = T_END;
+            /* The signedness of enums is determined by the type, so say this is specified to avoid
+            ** the int -> unsigned int handling for plain int bit-fields in AddBitField.
+            */
+            if (SignednessSpecified) {
+                *SignednessSpecified = 1;
+            }
             break;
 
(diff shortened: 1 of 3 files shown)


jmr:
@acqn

greg-king5:
:+1:

acqn:
+1 +1 along with a long-missing test.

PR #1254, opened 2020-09-11
[cc65] Processor state tracking

acqn:
(3/3) commits.
- [x] Added processor flags usage tracking.
- [x] Added ZNRegs for tracking what register(s) Z/N flags currently reflect.
- [x] Added utility functions to check if the specified processor state is known to be a certain value.
- [x] Fixed processor states tracking for the BIT/TRB/TSB opcode.
- [x] Added new opcode descriptions about whether and how the opcode accesses memory.
- [x] Added debug output support for processor flags.

This is the first half of a changeset to add processor state tracking that doesn't require fixing the optimizer. It tracks the usage of the state of the processor flags but hasn't been utilizing them yet. The second half of the changeset is to start to use them.


acqn:
Since most of the fixes originally made for the second half have been pushed before, the remaining part is small (5 files/commits) and could be appended here without complicating this PR much. Should the second half be appended to this or be made a separate PR?

oliverschmidt:
> Should the second half be appended to this or be pushed in a separate PR?

I'd prefer a separate PR.

PR #1256, opened 2020-09-13
[cc65] Avoided unnecessarily boosting the code label numbers with boolean AND

acqn:
Greatly reduced the label numbers when `&&` is involved, to ease the label overflow number concern in PR #1220:
> It might be worth mentioning that despite that 1d7fcc67a498f65a4687099b87a40d402c039c78 saves nearly half of the labels, the new testcase d4025efaba798c652302bb49fb45506f8504b5c0 still easily reaches a big label number (>= L1141).

... as well as the problems described in Issue #301.

### DIFF
--- a/src/cc65/expr.c
+++ b/src/cc65/expr.c
@@ -3237,8 +3237,8 @@ static int hieAnd (ExprDesc* Expr, unsigned* TrueLab, int* TrueLabAllocated)
     int HasFalseJump = 0, HasTrueJump = 0;
     CodeMark Start;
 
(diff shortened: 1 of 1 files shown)


acqn:
The liggest label number in `test/ref/pr1220.c` is reduced from > `L1000` to < `L0400` by this. With PR #1257 in addition it will be further reduced to < `L0090`.

oliverschmidt:
Low risk change and no feedback in five days -> merging...

PR #1257, opened 2020-09-13
[cc65] Made local code/data labels really local to the functions where they live in

acqn:
Reset the code/data label numbers in every function to ease with the problem described in #301. 

Also greatly reduced the label numbers when `&&` is invloved to ease the label overflow number concern in PR #1220:
> It might be worth mentioning that despite [1d7fcc6](https://github.com/cc65/cc65/commit/1d7fcc67a498f65a4687099b87a40d402c039c78) saves nearly half of the labels, the new testcase [d4025ef](https://github.com/cc65/cc65/commit/d4025efaba798c652302bb49fb45506f8504b5c0) still easily reaches a big label number (>= L1131).

Note: String literals still have global data label numbers even though they were compiled with `--local-strings`.

### DIFF
--- a/src/cc65/asmlabel.c
+++ b/src/cc65/asmlabel.c
@@ -42,6 +42,17 @@
 /* cc65 */
 #include "asmlabel.h"
 #include "error.h"
+#include "segments.h"
+
+
+
+/*****************************************************************************/
+/*                                   Data                                    */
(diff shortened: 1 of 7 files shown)


oliverschmidt (commented):
No feedback in five days -> merging...

PR #1258, opened 2020-09-13
[cc65] Processor state tracking

acqn:
(3/3) commits.
- [x] Added processor flag usage/change-tracking for non-JSR/RTS code entries.
  - [x] Fixed and improved `OptPrecalc`. Now it respects the C/V/Z/N flags.
  - [x] Fixed optimizations impacted by added support of tracking processor flags.
- [x] Added processor state info to the `OPCDesc` table. However, since some opcodes are affected by the addressing mode, this info is unused in codegen/optimizer but just as quick in-code documentation/hints. The real processor state info used for optimizer instead of this is deduced elsewhere.
- [x] Fixed quick hack for known registers after calling certain runtime functions, and new quick hack for `tosshrax`.

This is the second half of the changeset (see also #1254) to add processor state tracking. It has a few fixes to the optimizer in order to work again.

### DIFF
--- a/src/cc65/codeent.c
+++ b/src/cc65/codeent.c
@@ -1418,10 +1418,10 @@ void CE_GenRegInfo (CodeEntry* E, RegContents* InputRegs)
                     Out->RegX = (In->RegX ^ 0xFF);
                 }
             } else if (strcmp (E->Arg, "tosandax") == 0) {
-                if (In->RegA == 0) {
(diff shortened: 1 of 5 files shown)


acqn:
As expected. So let's try reordering....

oliverschmidt:
No feedback in five days -> merging...

PR #1264, opened 2020-09-24
Simplify Gamate tile redefinition (as already done for PCE)

Fabrizio-Caruso:
This pull request greatly simplifies the (re)-definition of tiles for the gamate target by using an Assembly file for the tile data instead of a .inc file.

This is very similar to what has already been done by @greg-king5 for the pce target in
https://github.com/cc65/cc65/commit/ba0ef5938d267dc47ca6d30af18b596190b4f932

### DIFF
--- a/libsrc/gamate/vga.s
+++ b/libsrc/gamate/vga.s
@@ -1,8 +1,8 @@
 ; VGA charset for the Gamate conio implementation
 
-.export fontdata
+        .export fontdata
 
-.rodata
+        .rodata
 
 fontdata:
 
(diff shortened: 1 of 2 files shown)


libsrc/gamate/vga.s
```diff
@@ -1,6 +1,11 @@
-
 ; VGA charset for the Gamate conio implementation
 
+.export fontdata
+
+.rodata
```
oliverschmidt, line 5: Both .export and .rodata should be indented.

Fabrizio-Caruso: Done

Fabrizio-Caruso:
Thanks @oliverschmidt !

oliverschmidt:
Thanks for the contribution.

PR #1270, opened 2020-09-26
Fix ICE for bit-fields with typedef

### LINKED ISSUE #1267: ICE from typedefs in bit-field
Splitting this from #1094 because it doesn't have anything to do with nesting.

```
typedef unsigned u16;
struct X {
  u16 a : 3;
};
```

```
cc65: Check failed: (Entry->Type->C & T_MASK_SIGN) == T_SIGN_SIGNED || IsTypeChar (Entry->Type), file 'cc65/symtab.c', line 879
```

I'll look at this.

jmr:
Fixes #1267

Avoid ICE, but treat plain int bit-fields declared via typedef as
signed rather than unsigned.  It is more efficient to treat them
as unsigned, but this requires distinguishing int from signed int,
and this is curently not done.


mrdudz:
this PR should probably also move test/misc/bug1094.c to test/val and remove it from test/misc/makefile :)

jmr:
> this PR should probably also move test/misc/bug1094.c to test/val and remove it from test/misc/makefile :)

I didn't have that commit, so didn't get any failures locally. Should be fixed now.

test/val/bug1267.c
```diff
@@ -0,0 +1,79 @@
...
+    if (si.s != 3) {
+        printf ("Got si.s = %d, expected 3.\n", si.s);
+        failures++;
+    }
+
+    si.i = -1;
+    si.u = -2;
+    si.s = -3;
+
+    /* Note that this is another bug that i is signed. */
```
jmr, line 59: If/when this is merged, I will open another bug for this inconsistent handling of typedefed int bit-fields.

oliverschmidt:
There was no additional feedback for quite some days -> merging...

Issue #1211, opened 2020-08-22
ICE from CS_MergeLabels/CL_MoveRefs with computed goto


jmr:
```
int g(int x) {
    static const void *const labels[] = {&&L0, &&L1};
    goto *labels[x];
    L0: if (labels[0] != labels[1]) return 31; else return 13;
    L1: return 41;
}
```

Results in:
`cc65: Check failed: E->JumpTo == OldLabel, file 'cc65/codelab.c', line 115`

`E->JumpTo` is `NULL`.

@acqn If the #897 fix also fixes this, that's even better.

jmr:
I definitely don't understand what's going on with the label parsing and merging code, but it seems like either we have to parse data statements and know how to rewrite the labels in them, or allow multiple labels to remain after `CS_MergeLabels`. The latter seems like it should be easier. Will that break something?

acqn:
The current fix for #897 doesn't fix this, as it only fix tracking references to labels _by emitted code entries_, but not references to labels _by the data (that has no code at all)_. This is exactly
> but it seems like either we have to parse data statements and know how to rewrite the labels in them,

`CS_MergeLabels` does two things: merge labels _and_ remove "unreferenced" labels (which fails and results in the issues). It seems to me that it is the latter part that needs a fix, not the former (which does have a bug in the implementation as it relies on the latter, but not "in the design").

jmr:
It looks like there's a partial work-around that fixes the more realistic case of a user `&&Label` getting merged with a compiler label. `CS_MergeLabels` can be changed to pick the label with the null `JumpTo` reference as the destination for merging, then the data doesn't need to be rewritten. I'll test this more thoroughly in the next day or two.

acqn:
That approach would be fine.

mrdudz:
was this issue fixed?

I am confused by that the testcase was added to test/err - the programs in test/err are expected to not compile if the compiler works correctly. Tests that dont compile because of compiler errors should go to test/misc !

jmr:
I think the test is just in the wrong directory. The distinction between test/misc and test/err isn't obvious. Maybe they could be renamed.

polluks2:
Let's call it "test/ice/" :-)

Issue #1049, opened 2020-06-17
[feature request] jump tables for switch


jmr:
[The case dispatch of a `switch` looks like it's compiled to a series of `beq cmp`.](https://godbolt.org/z/d7rmz6)  I could easily be wrong, but this looks like four bytes per case, while a jump table would be two. (Assuming sufficiently dense case labels, etc.)

```
unsigned char f(unsigned char x) {
    unsigned char y;
    switch (x) {
        case 0x0: y = 0x31; break;
        case 0x1: y = 0x41; break;
        case 0x2: y = 0x59; break;
        case 0x3: y = 0x25; break;
        case 0x4: y = 0x65; break;
        case 0x5: y = 0x36; break;
    }
    return y;
}
```

```
        beq     L001F
        cmp     #$01
        beq     L0020
        cmp     #$02
...
L001F:  lda     #$31
        jmp     L0025
L0020:  lda     #$41
        jmp     L0025
...
```

PR #1271, opened 2020-09-27
CS_MergeLabels: Keep labels referenced by data

jmr:
Partial fix for ICE in #1211.  This may fix enough to allow #1049 to be
fixed.

When merging labels, keep the first label with a ref that has no JumpTo;
this is a data segment label, used by computed gotos.

The real fix is to track and rewrite labels in data, but this is more
involved.

### DIFF
--- a/src/cc65/codeent.h
+++ b/src/cc65/codeent.h
@@ -180,6 +180,16 @@ INLINE CodeLabel* CE_GetLabel (CodeEntry* E, unsigned Index)
 #  define CE_GetLabel(E, Index) CollAt (&(E)->Labels, (Index))
 #endif
 
+#if defined(HAVE_INLINE)
+INLINE void CE_ReplaceLabel (CodeEntry* E, CodeLabel* L, unsigned Index)
+/* Replace the code label at the specified index with L */
+{
+    return CollReplace (&E->Labels, L, Index);
+}
+#else
+#  define CE_ReplaceLabel(E, L, Index) CollReplace (&(E)->Labels, (L), (Index))
+#endif
+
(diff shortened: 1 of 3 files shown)


src/cc65/codeent.h
```diff
@@ -151,6 +151,15 @@ INLINE CodeLabel* CE_GetLabel (CodeEntry* E, unsigned Index)
 #  define CE_GetLabel(E, Index) CollAt (&(E)->Labels, (Index))
 #endif
 
+#if defined(HAVE_INLINE)
+INLINE void CE_ReplaceLabel (CodeEntry* E, CodeLabel* L, unsigned Index)
```
oliverschmidt, line 155: All other CE_ "entries" hav a brief comment. I'd appreciate sticking to that here too.

src/cc65/codeseg.c
```diff
@@ -935,8 +966,10 @@ void CS_MergeLabels (CodeSeg* S)
             continue;
         }
 
-        /* We have at least one label. Use the first one as reference label. */
-        RefLab = CE_GetLabel (E, 0);
+        /** Pick a label to keep, all references will be moved to this one, and the other labels
```
oliverschmidt, line 969: Please replace `/**` with `/*`.

oliverschmidt (changes requested):
Please follow up on my minor style nitpicking.

src/cc65/codeseg.c
jmr: Done.

src/cc65/codeent.h
jmr: Done.

oliverschmidt (approved):

Issue #960, opened 2019-10-13
[lynx][conio][feature/question] Some more conio support


Fabrizio-Caruso:
The Atari Lynx target lacks `conio` (except for `cgetc,` which does not work correctly as reported in issue #959).

It seems to me that all building blocks for `conio` are already there because we have functions like `tgi_outtextxy`, `tgi_setcolor` and `tgi_clear.`
What is missing to have at least `gotoxy` and `cputc`?

We may just need two ZP locations for `x` and `y` cursor position and then we could just leverage on `tgi_outtextxy` to perform `cputc` and maybe `cprintf.`

Fabrizio-Caruso:
@greg-king5 have you created a branch for this in your fork or somewhere else?
I suppose the conio does not interpret all character codes as TGI.
You mentioned some garbage characters. Which ones?

greg-king5:
No branch yet.  Here is what I get when I run the program that I built from ["samples/ascii.c"](https://github.com/cc65/cc65/blob/master/samples/ascii.c):
![ascii.png](https://user-images.githubusercontent.com/4145570/67127103-a5439880-f1c6-11e9-8a49-fdbea6ee9be3.png)

Fabrizio-Caruso:
Maybe we just need more characters and/or the correct mapping used by CONIO.
TGI for the Lynx has characters 33-127, which I thought was enough. I suspect that the other codes are mapped to data somewhere in memory where code is.

Fabrizio-Caruso:
@greg-king5 conio may be using characters outside the range 33-127 for some reason.

polluks:
Well, this character is 32, the blank.

Fabrizio-Caruso:
Sorry, TGI has character 32 and it displays it correctly. TGI handles characters starting from 32 until 127.
So, if conio displays a different character but still displays some other characters correctly, it is doing some remapping of the characters.

BrianPeek:
Hello!  The latest updates to cgetc for the Lynx seem to eat every other keypress...can anyone else confirm?  Thanks!

Fabrizio-Caruso:
@BrianPeek  Hi! I will check it and try to fix it. The previous version was doing worse: falsely detecting the last key pressed infinitely many times.

Fabrizio-Caruso:
[...]

PR #1272, opened 2020-09-27
Cosmetic change: separation of lynx font data from lynx code

Fabrizio-Caruso:
This is a cosmetic change to separate lynx TGI font data from lynx TGI code.

The font code should probably move in a different directory as it could be also used by a CONIO (to be implemented).

It seems to me that it cannot be used to pass a different set of font definitions at link-time.
I guess the font data ends up into the TGI driver anyway.
Nevertheless it makes the code cleaner and ready for CONIO.


oliverschmidt:
> Nevertheless it makes the code cleaner and ready for CONIO.

Unfortunately as this PR stands it doesn't do neither. If there's at some point a momentum for CONIO one should see how to approach this best. E.g. I see the font being not included but linked, having a label and being located by the TGI driver via a library reference.

Fabrizio-Caruso:
@oliverschmidt There is an open issue with some initial work in #960. @greg-king5 did something some time ago.
 
I do not see why font data should stay in the same file with the code but I understand this is a matter of personal taste.

oliverschmidt:
> I do not see why font data should stay in the same file with the code

Me too.

> but I understand this is a matter of personal taste.

No, it's a matter of correctness and style consistency. But I didn't (and in fact still don't) want to discuss the details with you as I don't see a major benefit in moving the font data at this point without CONIO support.

Issue #1277, opened 2020-10-03
dbg/dbg.c ,crash while building on OpenBSD


frakswe:
`apple2 - dbg/dbg.c`
`gmake[2]: *** [Makefile:283: ../libwrk/apple2/dbg.o] Segmentation fault (core dumped)`
`gmake[1]: *** [Makefile:118: apple2] Error 2`
`gmake: *** [Makefile:7: all] Error 2`
 
`~/source/cc65/libsrc $ gdb ../bin/cc65 cc65.core`
`GNU gdb 6.3`
`Copyright 2004 Free Software Foundation, Inc.`
`GDB is free software, covered by the GNU General Public License, and you are`
`welcome to change it and/or distribute copies of it under certain conditions.`
`Type "show copying" to see the conditions.`
`There is absolutely no warranty for GDB.  Type "show warranty" for details.`
`This GDB was configured as "amd64-unknown-openbsd6.7"...`
`Core was generated by  cc65'.`
`Program terminated with signal 11, Segmentation fault.`
`Loaded symbols for /home/frakswe/source/cc65/bin/cc65`
`Reading symbols from /usr/lib/libm.so.10.1...done.`
`Loaded symbols for /usr/lib/libm.so.10.1`
`Reading symbols from /usr/lib/libc.so.96.0...done.`
`Loaded symbols for /usr/lib/libc.so.96.0`
`Reading symbols from /usr/libexec/ld.so...Error while reading shared library symbols:`
`Dwarf Error: wrong version in compilation unit header (is 4, should be 2) [in module /usr/libexec/ld.so]`
`#0  strlen () at /usr/src/lib/libc/arch/amd64/string/strlen.S:125`
`125     /usr/src/lib/libc/arch/amd64/string/strlen.S: No such file or directory.`
`      in /usr/src/lib/libc/arch/amd64/string/strlen.S`
`(gdb) q`

What is strlen.S ?; 
[https://github.com/openbsd/src/blob/master/lib/libc/arch/amd64/string/strlen.S](url)
 
`       _ALIGN_TEXT`
`1:`
`        movq    (%rax),%rdx             /* get bytes to check */`

Testing cc65 on another random file does not crash:
 
`~/source/cc65/libsrc/apple2 $ ../../bin/cc65 closedir.c -o closedir.o`
`~/source/cc65/libsrc/apple2 $ ls closedir*`
`closedir.c  closedir.o`

Building cc65 with clang or gcc v8 , same result.

Why this?

[...]

PR #1282, opened 2020-10-04
[cc65] Better support for putting stuff on the zeropage

### LINKED ISSUE #243: Indirect array accesses not ZP-optimized
It seems accesses of the type "array[var]" do not get the one-byte (or more) reduction, even when both array and var are in ZP. The direct type, "array[13]", does get it.

All under -Oisr.

### LINKED ISSUE #261: import/export of zeropage symbols using pragmas in C does not work as expected
what i did was exporting a symbol in one C file, using pragmas to make it a zeropage symbol and also pushing it into a custom segment. and then i import it in another file, again using pragma to make it a zeropage symbol (as described in the docs):

``` C
extern char foo;
#pragma zpsym ("foo")
int n;
int main(void) {
    n = foo;
    return 0;
}
```

``` C
//#pragma bss-name (push,"ZEROPAGE")            /* no warning */
#pragma bss-name (push,"MYZEROPAGE")            /* warning */
char foo;
#pragma zpsym ("foo")
#pragma bss-name (pop)
```

now the odd thing is, the assembler gives this warning:

``` C
zpsym2.s(18): Warning: Symbol `foo' is absolute but exported zeropage
```

...which seems wrong (and is a bit irritating). also the generated code is correct (uses zeropage adressing).

whats interesting is that the warning goes away when putting the symbol into the regular "ZEROPAGE" segment (without changing anything else) - that seems to imply that its actually a bug, because for extra zeropage variables using an extra segment is pretty much required.

### LINKED ISSUE #917: [feature request] Option to have global variables in zeropage
Global variables in zeropage can produce more than >10% speed-ups. This can be implemented currently by a cumbersome procedure:

1. declaring then as extern 
2. using the zpsym pragma 
3. and then linking a .s file with their actual location in zero-page

It would be nice to trigger this behavior with a new option (-Oz ?) that would just take the first global variables in the order in which they are declared and map them in zero-page locations that are usually available for C-only projects for each specific target (or use a target-specific config file for this). 

An example of a nice >10% speed-up produced by this optimization is testable in my project:
https://github.com/Fabrizio-Caruso/Primitive-Pythagorean-Triples

Remark: ANSI C does not allow the register qualifier on global variables. So we cannot have this behavior as a result of the register qualifier on global variables.

acqn:
- [X] The instruction parser can now recognize ZP locations and set the addressing mode for them.
- [X] Support for optionally specifying address size for segments with:
`#pragma ***-name([push,] "name"[, "addrsize"])`
where "addrsize" is any addressing size supported in ca65.
- [X] Symbols in ZP segments will use '.exportzp' instead of '.export' if exported.

This resolves #261 and resolves #917 with the new features. It could be considered that this also "almost resolves" #243 (as in the OP) since this opens a path to the expected optimizations, although there are still some other factors that could prevent the optimizations from happening.

The new optional parameter of `#pragma ***-name`
---
`#pragma ***-name([push,] "name"[, "addrsize"])`

Example: https://github.com/cc65/cc65/issues/917#issuecomment-647326244.

The new optional "addrsize" string is to add an address size string for a segment in asm output and in theory can be anything, but only those recognized by ca65 make sense.

- Absolute address size strings
  - "abs"
  - "absolute"
  - "near"
- Default address size strings
  - "default"
- Far address size strings
  - "far"
- Long address size strings
  - "dword"
  - "long"
- Zeropage address size strings
  - "direct"
  - "zeropage"
  - "zp"

Moreover, symbols put on a zeropage specified with this pragma will be exported with `.exportzp` instead of `.export`.

### DIFF
--- a/src/cc65/codeinfo.c
+++ b/src/cc65/codeinfo.c
@@ -349,23 +349,23 @@ static const FuncInfo FuncInfoTable[] = {
 
 /* Table with names of zero page locations used by the compiler */
 static const ZPInfo ZPInfoTable[] = {
-    {   0, "ptr1",      REG_PTR1_LO,    REG_PTR1        },
-    {   0, "ptr1+1",    REG_PTR1_HI,    REG_PTR1        },
-    {   0, "ptr2",      REG_PTR2_LO,    REG_PTR2        },
-    {   0, "ptr2+1",    REG_PTR2_HI,    REG_PTR2        },
-    {   4, "ptr3",      REG_NONE,       REG_NONE        },
-    {   4, "ptr4",      REG_NONE,       REG_NONE        },
-    {   7, "regbank",   REG_NONE,       REG_NONE        },
-    {   0, "regsave",   REG_SAVE_LO,    REG_SAVE        },
-    {   0, "regsave+1", REG_SAVE_HI,    REG_SAVE        },
-    {   0, "sp",        REG_SP_LO,      REG_SP          },
-    {   0, "sp+1",      REG_SP_HI,      REG_SP          },
-    {   0, "sreg",      REG_SREG_LO,    REG_SREG        },
-    {   0, "sreg+1",    REG_SREG_HI,    REG_SREG        },
-    {   0, "tmp1",      REG_TMP1,       REG_TMP1        },
-    {   0, "tmp2",      REG_NONE,       REG_NONE        },
-    {   0, "tmp3",      REG_NONE,       REG_NONE        },
-    {   0, "tmp4",      REG_NONE,       REG_NONE        },
+    {   0, "ptr1",      2,  REG_PTR1_LO,    REG_PTR1    },
+    {   0, "ptr1+1",    1,  REG_PTR1_HI,    REG_PTR1    },
+    {   0, "ptr2",      2,  REG_PTR2_LO,    REG_PTR2    },
+    {   0, "ptr2+1",    1,  REG_PTR2_HI,    REG_PTR2    },
+    {   4, "ptr3",      2,  REG_NONE,       REG_NONE    },
+    {   4, "ptr4",      2,  REG_NONE,       REG_NONE    },
+    {   7, "regbank",   6,  REG_NONE,       REG_NONE    },
+    {   0, "regsave",   4,  REG_SAVE_LO,    REG_SAVE    },
+    {   0, "regsave+1", 3,  REG_SAVE_HI,    REG_SAVE    },
+    {   0, "sp",        2,  REG_SP_LO,      REG_SP      },
+    {   0, "sp+1",      1,  REG_SP_HI,      REG_SP      },
+    {   0, "sreg",      2,  REG_SREG_LO,    REG_SREG    },
+    {   0, "sreg+1",    1,  REG_SREG_HI,    REG_SREG    },
+    {   0, "tmp1",      1,  REG_TMP1,       REG_TMP1    },
+    {   0, "tmp2",      1,  REG_NONE,       REG_NONE    },
+    {   0, "tmp3",      1,  REG_NONE,       REG_NONE    },
+    {   0, "tmp4",      1,  REG_NONE,       REG_NONE    },
 };
 #define ZPInfoCount     (sizeof(ZPInfoTable) / sizeof(ZPInfoTable[0]))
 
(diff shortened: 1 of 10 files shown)


jmr:
> * [x]  Support for optionally specifying address size for segments with:
>   `#pragma ***-name([push,] "name"[, "addrsize"])`
>   where "addrsize" is any addressing size supported in ca65. 

Is there some `#pragma` documentation in `doc/cc65.sgml` that can be updated?

acqn:
> 
> 
> > * [x]   Support for optionally specifying address size for segments with:
> >   `#pragma ***-name([push,] "name"[, "addrsize"])`
> >   where "addrsize" is any addressing size supported in ca65.
> 
> Is there some `#pragma` documentation in `doc/cc65.sgml` that can be updated?

Yep. About the new optional address size parameter and the change of the behavior of `.export` as well.

oliverschmidt:
>> Is there some #pragma documentation in doc/cc65.sgml that can be updated?

> [...] I can take care of that if no one else voluteers to. (But look out for my English!)

Sorry for the misunderstanding. I sort of implied that your answer would mean that you provide a documentation enhancement. But now I see that it didn't. Therefore I explicitly ask you to provide some docs. I presume @greg-king5 is willing to make sure the new content is correct language-wise.

greg-king5:
> > > Is there some `#pragma` documentation in "doc/cc65.sgml" that can be updated?
> 
> > [...] I can take care of that if no one else volunteers to. (But, look out for my English!)
> 
> Sorry for the misunderstanding. I sort of implied that your answer would mean that you provide a documentation enhancement. But, now I see that it didn't. Therefore, I explicitly ask you to provide some docs. I presume @greg-king5 is willing to make sure the new content is correct language-wise.

Yep. :grin:

src/cc65/segments.c
```diff
@@ -85,6 +96,86 @@ static Collection SegmentStack = STATIC_COLLECTION_INITIALIZER;
...
+
+
+void DoneSegAddrSizes (void)
+/* Free the segment address sizes */
+{
+    SegAddrSize_t* A;
+    int I;
+    for (I = 0; I < (int)CollCount (&SegmentAddrSizes); ++I) {
+        A = CollAtUnchecked (&SegmentAddrSizes, I);
+        SB_Done (A->Name);
```
greg-king5, line 115: `SB_Done()` frees the string _inside_ the StrBuf.  That line must free the entire StrBuf; use `FreeStrBuf()`.

greg-king5 (changes requested):

src/cc65/codeinfo.c
```diff
@@ -377,6 +378,85 @@ static const ZPInfo ZPInfoTable[] = {
...
+/* Exam if the main part of the arg string indicates a ZP loc */
+{
+    unsigned short  ArgInfo = 0;
+    long            Offset = 0;
+    StrBuf          NameBuf;
+    SymEntry*       E = 0;
+    const ZPInfo*   Info = 0;
+
+    /* The actual name can never be longer than the arg string */
+    SB_InitFromString (&NameBuf, xstrdup (Name));
```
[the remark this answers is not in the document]
acqn: One more misuse of ` SB_InitFromString` like that in #1277. It doesn't need to pre-allocate the buffer.

src/cc65/codeinfo.c
```diff
@@ -377,6 +378,85 @@ static const ZPInfo ZPInfoTable[] = {
 
 
 
+int IsAbsAddrOnZP (long Address)
+/* Return true if the Address is within the ZP range of the platform.
+** FIXME: ZP range may vary on different platforms.
```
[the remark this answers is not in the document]
acqn: Weird.... The function name should have been ` IsAddrOnZP ` as I recall naming it. The current one seems to be an interim edit. Anyways it's an easy fix.

src/cc65/codeinfo.c
```diff
@@ -377,6 +378,85 @@ static const ZPInfo ZPInfoTable[] = {
...
+            return Offset >= 0 && Offset < (int)Info->Size;
+        }
+
+        /* Assume it be non-ZP */
+        return 0;
+    }
+
+    if ((ArgInfo & AIF_EXTERNAL) == 0) {
+        /* We don't support local variables on ZP */
+        return 0;
```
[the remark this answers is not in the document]
acqn: Theis piece of code was cut & pasted from somewhere outside into this function. Typically missing clean-ups for multiple function exits.

src/cc65/segments.c
acqn: Hmmm, `A->Name` was not intended to be allocated separately from `A`.

src/cc65/codeinfo.c
```diff
@@ -40,6 +40,7 @@
 #include "chartype.h"
 #include "coll.h"
 #include "debugflag.h"
+#include "xmalloc.h"
```
greg-king5, line 43: That `#include` isn't needed, now.

src/cc65/segments.c
```diff
@@ -85,6 +96,86 @@ static Collection SegmentStack = STATIC_COLLECTION_INITIALIZER;
...
 
+void InitSegAddrSizes (void)
+/* Initialize the segment address sizes */
+{
+    InitCollection (&SegmentAddrSizes);
+}
+
+
+
+
```
greg-king5, line 107: One blank line too many.

greg-king5 (approved):

src/cc65/codeinfo.c
acqn: Removed.

src/cc65/segments.c
acqn: Removed.

src/cc65/pragma.c
```diff
@@ -430,13 +433,35 @@ static void SegNamePragma (StrBuf* B, segment_t Seg)
...
+            SB_Skip (B);
+            SB_SkipWhite (B);
+
+            /* A string argument must follow */
+            if (!GetString (B, &A)) {
+                goto ExitPoint;
+            }
+
+            /* Get the address size string for the segment */
+            AddrSize = AddrSizeFromStr (SB_GetConstBuf (&A));
```
greg-king5, line 454: That expression actually gets the size, not the "size string".  The `GetString()` call, above, got the string.

greg-king5:
The code is good.  There will be no more nitpicking.  This PR will be merged soon after the comment is fixed.

greg-king5:
> The code is good. There will be no more nitpicking. This PR will be merged soon after the comment is fixed.

Oops!  I forgot about "doc/cc65.sgml".  I'll wait for that patch.

acqn:
> 
> 
> > The code is good. There will be no more nitpicking. This PR will be merged soon after the comment is fixed.
> 
> Oops! I forgot about "doc/cc65.sgml". I'll wait for that patch.

Sorry, I am a bit too busy currently to edit that file comforming to the syntax and style by myself. I've added some brief descriptions in the OP about the new features wishing for some volunteers. :p

oliverschmidt:
> @greg-king5: The code is good. There will be no more nitpicking. This PR will be merged soon after the comment is fixed.

> @acqn: Sorry, I am a bit too busy currently to edit that file comforming to the syntax and style by myself.

Based on those two statements it seems best to merge this PR now. I presume that the cc65 community agrees that given the value of contributions by @acqn it wouldn't be appropriate to consider the docs obligatory.

mrdudz:
may i suggest to mark the old "#pragma zpsym" deprecated then and issue a warning that points to the new pragma? that would be very useful, imho.

PR #1285, opened 2020-10-12
[cc65] Deferred postfix inc/dec operations till sequence points

### LINKED ISSUE #1229: Poor post-increment codegen when value not needed
```
static unsigned int si = 0;
void postinc_static_int(void) { si++; }
```

```
.proc   _postinc_static_int: near

.segment        "CODE"

        lda     _si
        ldx     _si+1
        jsr     incax1
        sta     _si
        stx     _si+1
        rts

.endproc
```

This could just be
```
        inc     _si
        bne     L000C
        inc     _si+1
L000C:  rts
```

https://godbolt.org/z/fPYoT4

It's probably not too hard to pass in a bool to the codegen indicating whether the value is actually needed. It's not needed for statements, the first operand of `,`, and probably other places.

acqn:
...since the C standard allows that. Resolves #1229.

Basically, cc65 has been doing post-inc/dec mostly like this (pseudo-code):
```asm
; use expr++;             // or expr--;
load expr into primary
backup expr in regsave
inc expr                  ; or dec expr
load regsave into primary
use expr
```
This PR instead (only if `expr` is not complicated that it needs much code to be loaded):
```asm
; use expr++;             // or expr--;
load expr into primary
use expr
backup expr in regsave    ; only if the original value of expr must still be kept in primary
inc expr                  ; or dec expr
load regsave into primary ; only if the original value of expr must still be kept in primary
```
This usually allows faster & smaller code. It is more complicated than a simple solution that like 63506b05cdad1960a2e702037b58674c4ecd4809 from https://github.com/cc65/cc65/issues/1229#issuecomment-683707744, but gives better results when the value of the post-inc/dec is used.

Note that deferred operations must still be called at sequence points even if the whole expressions containing them had constant values.

PS: This is a merge of several old patches and I hope nothing got wrong or missed out.

### DIFF
--- a/src/cc65/compile.c
+++ b/src/cc65/compile.c
@@ -320,6 +320,9 @@ static void Parse (void)
                 } else {
                     /* Parse the function body */
                     NewFunc (Entry, FuncDef);
+
+                    /* Make sure we aren't omitting any work */
+                    CheckDeferredOpAllDone ();
                 }
             }
 
@@ -394,6 +397,8 @@ void Compile (const char* FileName)
     /* Other standard macros */
     /* DefineNumericMacro ("__STDC__", 1);      <- not now */
     DefineNumericMacro ("__STDC_HOSTED__", 1);
+
+    InitDeferredOps ();
 
     /* Create the base lexical level */
     EnterGlobalLevel ();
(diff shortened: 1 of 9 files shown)


oliverschmidt:
Ne feedback for quite some days -> merging...

PR #1286, opened 2020-10-12
[cc65] Several new/changed optimization steps 

acqn:
- `OptSignExtended` and `OptShiftBack` are for optimizing code patterns that could be generated with #1070.
- `OpPushPop2` is for optimizing a code pattern that could be generated with #1285.
- `OptNegAX2` is changed to be more friendly to other optimization steps looking for addition patterns.

### DIFF
--- a/src/cc65/coptind.c
+++ b/src/cc65/coptind.c
@@ -1069,8 +1069,8 @@ unsigned OptTransfers4 (CodeSeg* S)
 
 
 
-unsigned OptPushPop (CodeSeg* S)
-/* Remove a PHA/PLA sequence were A is not used later */
+unsigned OptPushPop1 (CodeSeg* S)
+/* Remove a PHA/PLA sequence were A not used later */
 {
     unsigned Changes = 0;
     unsigned Push    = 0;       /* Index of push insn */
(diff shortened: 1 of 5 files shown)


src/cc65/coptind.c
```diff
@@ -1069,8 +1069,8 @@ unsigned OptTransfers4 (CodeSeg* S)
 
 
 
-unsigned OptPushPop (CodeSeg* S)
-/* Remove a PHA/PLA sequence were A is not used later */
+unsigned OptPushPop1 (CodeSeg* S)
```
jmr, line 1072: Maybe `OptPushPopAccumulator` and `OptPushPopStatus` would be more indicative names, but I see there's precedent for the numbered approach.

src/cc65/coptind.c
```diff
@@ -1385,3 +1474,149 @@ unsigned OptPrecalc (CodeSeg* S)
...
+
+    }
+
+    /* Return the number of changes made */
+    return Changes;
+}
+
+
+unsigned OptSignExtended (CodeSeg* S)
+/* Change
```
jmr, line 1535: It would be helpful to give example C code that can generate this assembly.

acqn:
```c
signed char a;
a / 2;
```

src/cc65/coptind.c
acqn: Since the two have clear comments and are only used once each in one place, I don't think the current naming convention is unacceptable.

oliverschmidt:
* As far as I can see @jmr's concerns have been taken care of.
* There's no additional feedback.

-> merging...

PR #1300, opened 2020-10-24
Default PREFIX to /usr/local/.

SvenMichaelKlose:
This is the traditional path on unixoids for additional user
system resources – having to set it manually every time for the
install target is rather unusual.  See also
https://www.gnu.org/software/make/manual/html_node/Directory-Variables.html

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -2,6 +2,8 @@ ifneq ($(shell echo),)
   CMD_EXE = 1
 endif
 
+PREFIX ?= /usr/local
+
 .PHONY: all mostlyclean clean install zip doc html info
 
 .SUFFIXES:
@@ -46,7 +48,6 @@ clean:
 	$(RM) -r ../html ../info
 
 install:
-	$(if $(PREFIX),,$(error variable "PREFIX" must be set))
 ifeq ($(wildcard ../html),../html)
 	$(INSTALL) -d $(DESTDIR)$(htmldir)
 	$(INSTALL) -m0644 ../html/*.* $(DESTDIR)$(htmldir)
--- a/libsrc/Makefile
+++ b/libsrc/Makefile
@@ -2,6 +2,8 @@ ifneq ($(shell echo),)
   CMD_EXE = 1
 endif
 
+PREFIX ?= /usr/local
+
 CBMS = c128   \
        c16    \
        c64    \
@@ -94,7 +96,6 @@ INSTALL = install
 
 define INSTALL_recipe
 
-$(if $(PREFIX),,$(error variable "PREFIX" must be set))
 $(INSTALL) -d $(DESTDIR)$(datadir)/$(dir)
 $(INSTALL) -m0644 ../$(dir)/*.* $(DESTDIR)$(datadir)/$(dir)
 
--- a/samples/Makefile
+++ b/samples/Makefile
@@ -4,6 +4,8 @@
 # This Makefile requires GNU make
 #
 
+PREFIX ?= /usr/local
+
 # Run 'make SYS=<target>'; or, set a SYS env.
 # var. to build for another target system.
 SYS ?= c64
@@ -332,7 +334,6 @@ INSTALL = install
 samplesdir = $(PREFIX)/share/cc65/samples
 
 install:
-	$(if $(PREFIX),,$(error variable "PREFIX" must be set))
 	$(INSTALL) -d $(DESTDIR)$(samplesdir)
 	$(INSTALL) -d $(DESTDIR)$(samplesdir)/geos
 	$(INSTALL) -d $(DESTDIR)$(samplesdir)/tutorial
--- a/src/Makefile
+++ b/src/Makefile
@@ -2,6 +2,8 @@ ifneq ($(shell echo),)
   CMD_EXE = 1
 endif
 
+PREFIX ?= /usr/local
+
 PROGS = ar65     \
         ca65     \
         cc65     \
@@ -105,7 +107,6 @@ $(RM) /usr/local/bin/$(prog)
 endef # UNAVAIL_recipe
 
 install:
-	$(if $(PREFIX),,$(error variable "PREFIX" must be set))
 	$(INSTALL) -d $(DESTDIR)$(bindir)
 	$(INSTALL) ../bin/* $(DESTDIR)$(bindir)
 


doc/Makefile
```diff
@@ -46,7 +48,6 @@ clean:
 	$(RM) -r ../html ../info
 
 install:
-	$(if $(PREFIX),,$(error variable "PREFIX" must be set))
```
spiro-trikaliotis, line 49: Why do you remove these error statements? Even with your default, if the user sets an empty PREFIX, this errors are helpful, aren't they?

libsrc/Makefile
```diff
@@ -94,7 +96,6 @@ INSTALL = install
 
 define INSTALL_recipe
 
-$(if $(PREFIX),,$(error variable "PREFIX" must be set))
```
spiro-trikaliotis, line 97: dto.

samples/Makefile
```diff
@@ -332,7 +334,6 @@ INSTALL = install
 samplesdir = $(PREFIX)/share/cc65/samples
 
 install:
-	$(if $(PREFIX),,$(error variable "PREFIX" must be set))
```
spiro-trikaliotis, line 335: dto.

src/Makefile
```diff
@@ -105,7 +107,6 @@ $(RM) /usr/local/bin/$(prog)
 endef # UNAVAIL_recipe
 
 install:
-	$(if $(PREFIX),,$(error variable "PREFIX" must be set))
```
spiro-trikaliotis, line 108: dto.

doc/Makefile
SvenMichaelKlose: That's essentially setting the wrong path. If someone clears PREFIX that's probably with intention and these error messages break the process – am guess that they'll never kick in anyhow.

polluks: ACK you want it, you got it :-)

oliverschmidt:
My understanding is that this approach would break "everything" !

The Makefiles (have to) make a distinction between "just" making things to run in place and generating an installable template to be run from another place. That distinction needs a switch to be set by the user. Instead of introducing yet another flag it was decided to have PREFIX serve two purposes:

1. Allow the user to deviate from the usual prefix.
2. Detect if generating an installable template is desired.

If PREFIX would be given a default then 2.) wouldn't be possible anymore.

See:

* https://github.com/cc65/cc65/commit/32daa3dd2b004759e379e296ff00478c934992e4
* https://github.com/cc65/cc65/blob/master/src/Makefile#L22-L30 and https://github.com/cc65/cc65/blob/master/src/Makefile#L63-L67

doc/Makefile
SvenMichaelKlose: Awesome! Saves quite some time typing. Thank you and stay healthy. :)

SvenMichaelKlose:
The PREFIX can still be set usual way, so nothing breaks at all.

oliverschmidt:
It seems you haven't read my comment :-(

SvenMichaelKlose:
Perhaps it just doesn't occur to me what an "installable template" is supposed to be.

oliverschmidt:
"installable template" is supposed to mean to build cc65 to be run from the location indicated by PREFIX. That implies from my POV that the build result is moved to that location in a separate step - the "installation".

In contrast without PREXFIX (even as default) cc65 is simply built to run from its current location, no moving, no installation.

SvenMichaelKlose:
How about a 'make PREFIX=""' then?

oliverschmidt:
Okay, so we agree that setting PREFIX to a non-empty default "removes" something, right?

Then we can move on the the question which thing is more common, more usual. My perspective so far is that it's common/usual to build cc65 to be used in it's current location, and that only tools for creating a package (https://software.opensuse.org/download.html?project=home%3Astrik&package=cc65) need/want to set PREFIX.

But maybe I'm wrong. What's your motivation to set PREFIX in the first place?

SvenMichaelKlose:
Actually it adds some appeal. ;)

I myself always install in regular places even during development, so all other stuff using cc65 around my place can be tested without additional overhead.

However, if an unset PREFIX during a build causes different results and only the install throws an error, then one's going to install wrongly compiled binaries easily and that can cause trouble which is hard to recover. So having to build locally with PREFIX="`pwd`" for development gives less surprises. That's a little more overhead for developers but the most common case really is people wanting to build and install cool new stuff. :)

oliverschmidt:
> [...] the most common case really is people wanting to build and install cool new stuff. :)

* cc65 doesn't support any of the usual, popular approaches to "download source, build from source, install result, delete source" out of the box. Therefore I don't see any non-cc65 insider to succeed in doing so anyway.

* There are https://software.opensuse.org/download.html?project=home%3Astrik&package=cc65 so I don't see the need to "build from source" for the usual non-cc65 insider anyway.

So we are from my POV left with people not so much working _with_ cc65 but working _on_ cc65. And for those it seems much more attractive to have a quick turnaround cycles with edit, build, test instead of edit, build, (remove old install), install, test.

Therefore I'm still believing that there are (much) more make cmdlines entered by humans which are now "correct" without PREFIX then then there would be with your proposed change.

SvenMichaelKlose:
You could reject updates of the documentation easily in the same spirit.

oliverschmidt:
> [...] reject updates of the documentation easily in the same spirit.

I don't see a relation to the documentation.

> You could reject [...]

Me closing this PR was a reaction on your comment "Awesome! Saves quite some time typing. Thank you and stay healthy. :)" which I misinterpreted as you consenting to my initial comment. Sorry for that! In general I would be willing to reopen this PR at any time, however...

I our comments so far you just didn't convince me. I don't think that it is appropriate to insinuate some general spirit. But as you seem to want to put this discussion into such a direction: There's always the option to fork this repo and do with it whatever you see fit.

Nevertheless: Thanks for your interest to contribute to this repo :-)

<eof>

mrdudz:
> I myself always install in regular places even during development, so all other stuff using cc65 around my place can be tested without additional overhead.

I do *never ever* install anything into my system that doesnt come as a regular package. And i expect a local build to just work from its build dir. So i am with Oliver on this question.

Setting prefix by default to some directory in the system is not a good idea, because you can NOT be sure that "/usr/local" is actually what the system uses. It may be "traditional" what is used, but there are certainly setups that disagree with that tradition. Requiring the user to provide the prefix is the way to prevent errors, and it always works.

SvenMichaelKlose:
Didn't want to continue this discussion really. Sorry. I'm glad you noticed that closing an issue before having it discussed is off-putting and leaving a wrong impression.

There's nothing wrong with keeping regular users from falling into traps which is what the "tradition" fixes and GNU is totally with it. This is NOT preventing errors – the error message should at least state that the build has to be done with the same prefix to be in sync – that can save upcoming insiders hours of frustration.
Keep it as you like. No hard feelings. But I'm quite curious which systems disagree with having a '/usr/local' and what the other errors that would be worth preventing are. I doubt they're something that might make one bang his head on the keyboard.
At least we agree that having to specify PREFIX all the time is unnerving. I'll go for the other tradition of setting up that make.conf...

oliverschmidt:
> I'm glad you noticed that closing an issue before having it discussed is off-putting and leaving a wrong impression.

For sure!

> No hard feelings.

Thanks :-) Me too!

greg-king5:
As Oliver said, `make install PREFIX="<whatever>"` is for package builders.  We "ordinary" people should use
```bash
sudo make avail
```
It is "set and forget".  You need to type it only once -- no matter how many times you update and rebuild cc65's tools.

polluks:
I didn't notice "avail" so far, maybe we could rename it as
- install → install-builder
- avail → install

oliverschmidt:
> I didn't notice "avail" so far

https://cc65.github.io/getting-started.html#GNU ;-)

> avail → install

https://github.com/cc65/cc65/commit/352ceb166cce2db217933dbef87b9a8b8c22d353 ;-))

oliverschmidt:
Maybe of interest: https://github.com/cc65/cc65/issues/1303

polluks:
@oliverschmidt oops, avail is avail since 2013

PR #1302, opened 2020-10-25
Added generic way of defining CLOCKS_PER_SEC

mzattera:
Hi

I'm porting cc65 to a new target and I'm creating code for library functions in time.h for the new target.
For the purpose, I need to use CLOCKS_PER_SEC. I could use a separate header to define it, but then the user will have to always include <time.h> AND my header, which means I must adapt all existing code that uses time.h, this makes porting cumbersome.
My proposal here is to define an external function _clk_tck()  that is called, should CLOCKS_PER_SEC still be undefined. It has some speed drawbacks but will make porting of existing code easier.

I made minor formatting changes to the file.

### DIFF
--- a/include/time.h
+++ b/include/time.h
@@ -84,8 +84,9 @@ extern struct _timezone {
 
 
 #if defined(__ATARI__)
-/* The clock depends on the video standard, so read it at runtime */
-unsigned _clocks_per_sec (void);
+   /* The clock depends on the video standard, so read it at runtime
+   */
+   unsigned _clocks_per_sec (void);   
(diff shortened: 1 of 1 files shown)


oliverschmidt:
https://github.com/cc65/cc65/commit/f723147f04f19ddc7e39bad9211a6f26082ffbc5 is supposed to fulfill your needs.

mzattera:
It does indeed, thanks!

oliverschmidt:
You're welcome. Thanks for the feedback :-)

PR #1305, opened 2020-10-25
Knock off two bytes from getcwd(), cbm_read() and cbm_write() – 6 in total.

SvenMichaelKlose:
Instead of XORing high and low size bytes with $FF and counting up the word, the two bytes are incremented and counted down.


oliverschmidt:
This PR contains the PREFIX thingy.

SvenMichaelKlose:
T'was worth trying. ;) Just a minute or two...

PR #1306, opened 2020-10-25
Have cbm_read() test on EOF.

SvenMichaelKlose:
Also remove redudtant status check.


oliverschmidt:
I cannot judge if this optimization is correct. Please CBM guys: Provide feedback!

mrdudz:
now the second test will branch to the "ok" exit also on a read error.

http://unusedino.de/ec64/technical/aay/c64/zp90.htm (this is the value returned by READST)

SvenMichaelKlose:
Guess I do not know how to fix this EOF issue...

SvenMichaelKlose:
Is it OK to use the "too few bytes" flag for that?

SvenMichaelKlose:
OK, folks. I hope this did the trick. :)

greg-king5:
Commodore file-systems are different from most other systems.  The End-Of-File signal is sent _with_ the last byte, not on an attempt to read _after_ the last byte.  Therefore, while errors must prevent a byte from being stored, EOF must not do it.  That's why the status is read twice (it isn't redundant).

The check below the "jsr BASIN" looks for errors; it won't let bad bytes be stored.  The check above the "jsr BASIN" looks for EOF; it is executed after the (possibly last) byte is copied from`tmp1` into the output buffer.

The two checks must not be combined.  Therefore, I close this PR.

greg-king5:
An idea occurred to me, the tests can be "half-combined".  Only one call to `READST` is needed:

After the `jsr`, the status byte would be copied into the X register.  The error flags would be tested, and the code would branch if they aren't zero.  Then, the data byte would be stored in the buffer, and "buffer_read" (`ptr3`) would be incremented.  Then, the X register would be tested, and the code would branch if it isn't zero.  Then, the buffer pointer would be incremented, the "size" would be decremented, and the loop would be repeated.

(I re-openned this PR.)

SvenMichaelKlose:
There's only one READST left now. Guess I fixed my mistake with the following commit. This IO stuff is doing my head in. Works fine for my project now. When it finds and end-of-file flag it exits now without errors. Any other flag kills the fun. May be ready to merge now. In the end it seems to be doing what's been promised in the code comments.

mrdudz:
At this point i strongly recommend coming up with a test program that checks all the border cases :)

SvenMichaelKlose:
That'd be a whole new project, wouldn't it? Like including the VICE emulator in the test runs…

oliverschmidt:
I presume that @mrdudz "only" proposes to have a test program posted here and executed interactively by you and the reviewers. I don't see that he wants to imply, that there's a an automated regression test - especially as he and me discussed the idea of emulators being used in regression tests and rejected it.

oliverschmidt:
BTW: I strongly agree with @mrdudz that things like an empty file, a one byte file etc. should have been experienced to actually work.

mrdudz:
> I presume that @mrdudz "only" proposes to have a test program posted here and executed interactively by you and the reviewers. I don't see that he wants to imply, that there's a an automated regression test - especially as he and me discussed the idea of emulators being used in regression tests and rejected it.

yes, of course. (it could still be included in targettest)

There are a few corner cases that apparently are not trivia nor obvious to understand and get right - a test program for those should be really trivial though, so that seems the obvious thing to do for me. The biggest "problem" will be producing a d64 with deliberate read errors on it (i could assist with that).

oliverschmidt:
> it could still be included in targettest

Of course. @SvenMichaelKlose: That means that the CI only builds the program, but doesn't run it.

> [...] i could assist with that

Thanks for your offer.

SvenMichaelKlose:
Some insights: the EOF flag is being set as soon as the last byte is **available**. This way there's chance to check if a file is empty or is just containing a single byte. This renders cbm_read() useless for reading larger files in chunks as it doesn't track the status flag; so when the EOF flag shows, one'd have to set a propriety EOF flag that tells cmb_read() that the last byte went out already.

A solution: keep a 10 byte list of logical file numbers. If the EOF flag is set, the LFN is added to this list – if cbm_read() finds it at init, it returns 0. cbm_close() needs to remove the LFN again.

polluks:
@SvenMichaelKlose This sounds laborious, why/how does it work in BASIC?

mrdudz:
In BASIC you have to check ST after every operation - so it works just the same.

targettest/cbm/bug-1306.c
```diff
@@ -0,0 +1,19 @@
+#include <stdio.h>
+#include <stdlib.h>
+
+#include <cbm.h>
+
+void
+main ()
+{
+    char * data = malloc (1024);
+    unsigned len;
```
greg-king5, line 10: `len` needs to be (signed) `int` instead of `unsigned`.

SvenMichaelKlose:
Got it fixed for cbm_read(). Empty files will always return a byte though. This problem seems to affect all of the CBM file I/0 code: one cannot read in chunks if the returned size of read bytes is expected to be 0 at the end of a file.

greg-king5:
@polluks:
> It sounds laborious, how does it work in BASIC?

@mrdudz:
> In BASIC, you have to check ST after every operation.

I agree with polluks.  I prefer a simpler, more low-level method:
1. Call `cbm_read()`.  Save the return value.
2. If it's negative, then there was a file error.
3. Otherwise, the program always should learn about EOF and read-errors by calling `cbm_k_readst()`.

mrdudz:
> I agree with polluks. I prefer a simpler, more low-level method:
> 
> 1. Call `cbm_read()`.  Save the return value.
> 2. If it's negative, then there was a file error.
> 3. Otherwise, the program always should learn about EOF and read-errors by calling `cbm_k_readst()`.

That removes much of the reason to use cbm_read at all for me. The whole point of having such function is to not have to bother with doing all this lowlevel stuff all the time.

polluks:
However we should finally document this function :-)
@SvenMichaelKlose What about the C code comment, wasn't it useful?

SvenMichaelKlose:
> However we should finally document this function :-)
It's being documented in the cbm.h header file. – well, I'll add a hint about empty files.

> @SvenMichaelKlose What about the C code comment, wasn't it useful?
The assembly code isn't mirroring it any more. It wasn't even before really. And that'll probably happen again with the next change. So, why waste time with that? :)

SvenMichaelKlose:
Anyone with a good idea where to put a MAX_LFN constant perhaps? It's now defined in cbm_close() as well. :(

greg-king5:
> > 3. Otherwise, the program always should learn about EOF and read-errors by calling `cbm_k_readst()`.
> 
> That removes much of the reason to use cbm_read at all for me. The whole point of having such function is to not have to bother with doing all this lowlevel stuff all the time.

I believe that `cbm_read()` should be more low-level than `read()`.  `read()` tracks EOF.  `cbm_read()` [and `cbm_close()`] shouldn't waste space with extra code that does what we use `read()` to do.  It should do multi-byte input with a single function call -- nothing more.

mrdudz:
> It should do multi-byte input with a single function call -- nothing more.

because we encourage writing programs that don't do error handling?

polluks:
@mrdudz Well, if you write platform specific code, you are at your own, error checking is simply optional.

mrdudz:
> Well, if you write platform specific code, you are at your own, error checking is simply optional.

what a silly argument for not even bothering to implement it in that function. a non trivial program ment for end user should handle errors. always. its not the exception, its the norm. and you can still choose to ignore the returned error, obviously - but requiring the user to fiddle with plain kernal calls *every single time* he uses this function is just beyond silly. the point of having a c library is to hide this kind of stuff.

polluks:
@SvenMichaelKlose funcref was already prepared...
I see, it was more or less pseudo code.

greg-king5:
> Some insights: the EOF flag is being set as soon as the last byte is **available**. That way, there's a chance to check if a file is empty, or is just containing a single byte. That renders `cbm_read()` useless for reading larger files in chunks, as it doesn't track the status flag.  So, when the EOF flag shows, one'd have to set a proprietary EOF flag that tells `cbm_read()` that the last byte went out already.

I should have remembered this sooner:  This function doesn't need to track EOF, because the DOS does it.  That's what this comment in the source file means:
> The Kernal routine BASIN sets ST to EOF if the end of file
> is reached the first time, then we must store tmp.
> **Every subsequent call returns EOF and READ ERROR in ST**, then
> we must exit the loop here _immediately_.

The DOS remembers that the last byte went out already.  It makes `cbm_read()` return zero when it's supposed to do that -- no matter how many chunks were read.

The only fix that `cbm_read()` needs is that the status flags must be checked properly:
- The error flags must be checked _before_ the just-read byte is stored and counted.
- The EOF flag must be checked _after_ the byte is stored and counted.

That fix will make sure that the function returns zero for most empty files and "not-found" files.

SvenMichaelKlose:
At the moment it returns -1 for files that haven't been opened for whatever reason. Testing the status bits before reading I can try again, of course. Good point right there. Speaking of low memory: that's when reading chunks comes in handy. Perhaps one day somebody will grab the dragon book and give this compiler some decent optimisations. BTW 1K stack for a VIC-20 with 3.5K main memory seems to be a little overkill to me. Any chance to have that reduced by default? In the end: ~32 bytes less free space for saving a lot of people's precious life time is a nice trade in my mind.

SvenMichaelKlose:
OK, I have checked: the KERNAL does not set the EOF flag until a read attempt has been performed, so there's no fix for empty files. And as CHKIN is being called before checking the flags in advance as well doesn't seem to gain anything.

oliverschmidt:
>1K stack for a VIC-20 with 3.5K main memory seems to be a little overkill to me.

The usual default is 2k - so it's already only half of the usual default. However I can easily follow you that 1k is still (way) to much.

>Any chance to have that reduced by default?

I'm open to a pull request further reducing it.

mrdudz:
Just to give a number - in my Chameleon menu project i use only 512 bytes for stack, and thats still more than enough :)

SvenMichaelKlose:
> Just to give a number - in my Chameleon menu project i use only 512 bytes for stack, and thats still more than enough :)

Suggested 256B in another pull request. INGLE, my VIC GUI, doesn't even come close to that either.

mrdudz:
please rebase this PR to current head. please also rename the testprog to "cbmread.c"

mrdudz:
> please rebase this PR to current head. please also rename the testprog to "cbmread.c"

please? :)

mrdudz:
added a slightly updated version of the test in targettest/cbm/cbmread.c ...

@greg-king5 @SvenMichaelKlose could you look at fixing the remaining problems please? This PR appears to go into the right direction, however it needs fixing

SvenMichaelKlose:
How on Earth do I get to check out that PRs code?

mrdudz:
I'd make a local branch and then apply https://github.com/cc65/cc65/pull/1306.diff to it

SvenMichaelKlose:
Doesn't apply. Got no brain cell left to get the science right.

greg-king5:
I close this PR again.  It isn't viable.  The original (repository) version works properly.  This PR broke it, then tried to fix what it broke.  It goes too far.  `cbm_read()` doesn't need to remember EOF.  The DOS remembers it.

mrdudz:
> I close this PR again. It isn't viable. The original (repository) version works properly. This PR broke it, then tried to fix what it broke. It goes too far. cbm_read() doesn't need to remember EOF. The DOS remembers it.

Not sure what you are trying to say, but the test program does NOT work correctly yet. It is clearly not fixed.

greg-king5:
For me, `cbm_read()` passes the test program.  What -- specifically -- fails for you?

Please note that the empty-file-test _cannot_ pass!  CBM DOS doesn't support empty files.  When it reads an empty file, it pretends that the file has one byte.

mrdudz:
We have already discussed it right here and found out your ideas of how this function should work is different from what we actually want this function to behave like. Just read eg Olivers comments. This function is broken until every tested thing actually works. This function should hide all the quirky crap of the cbm dos.

polluks:
I agree with Greg: cbm_read is CBM specific, hence the name, so CBM DOS restrictions are no problem.

mrdudz:
We have already discussed it. If you want to get dirty and deal with all the CBM DOS specific quirks, you can always use the kernal calls, you don't need this function.

PR #1307, opened 2020-10-25
Knock off two bytes from getcwd(), cbm_read() and cbm_write().

SvenMichaelKlose:


oliverschmidt:
I'm obviously too stupid to understand this PR so I ask others to provide feedback. Thanks in advance!

SvenMichaelKlose:
OK, I'll try to clear it up a blit: This one increments each byte of the word in advance to be able to make use of the zero flag with the 'dec' instruction, so no kind of load and 'cmp #$ff' is necessary. That enables us to have two 'inx' instead of two 'eor #$ff' – why that's supposed to work I don't understand in turn. :)

jmr:
Maybe I'm not adding anything that's not obvious, but `x ^ 0xFF == ~x == -x - 1`, and the comment said they were storing `-size-1`. I'm not sure why that was supposed be simpler. I also need to stare at it more to make sure both versions are the same, but didn't have time yet.

polluks:
@jmr simpler in the sense of shorter, faster.

libsrc/cbm/cbm_read.s
```diff
@@ -45,11 +45,11 @@
...
-        eor     #$FF
-        sta     ptr1
-        txa
-        eor     #$FF
-        sta     ptr1+1          ; Save -size-1
+        inx
+        stx     ptr1+1
+        tax
+        inx
+        stx     ptr1            ; Save size with each byte incremented.
```
jmr, line 52: Should this say something like "Save size, *decremented* on each byte"?

libsrc/cbm/cbm_read.s
```diff
@@ -92,9 +92,9 @@ _cbm_read:
         bne     @L3
         inc     ptr3+1          ; ++bytesread;
 
-@L3:    inc     ptr1
+@L3:    dec     ptr1
```
jmr, line 95: What does your code do for `cbm_read(_, _, 0);`? It looks like the first test decrements from 0 to 2¹⁶ − 1, so it will loop 2¹⁶ times.

libsrc/cbm/cbm_read.s
Marrin: No, both are incremented with `inx` before they are stored in `ptr1+1` and `ptr1`. The comment is about what was done to the values, not about what happens with them afterwards.

greg-king5: I think that the phrase is slightly confusing.  I recommend, "with both bytes incremented separately".

libsrc/cbm/cbm_read.s
SvenMichaelKlose: > What does your code do for `cbm_read(_, _, 0);`? It looks like the first test decrements from 0 to 2¹⁶ − 1, so it will loop 2¹⁶ times.
Nope. $0000 becomes $0101 at init and will exit immediately. I'll have to print out this thread and frame it because now I'm sure that I'm not alone with the way I was feeling about this hack.

libsrc/cbm/cbm_read.s
SvenMichaelKlose: @greg-king5 Sorted.

libsrc/cbm/cbm_read.s
greg-king5: At the beginning of the function, the two bytes are incremented, separately, to `$0101`.  The loop starts by decrementing the low byte to `$00`.  Then, the high byte is decremented to `$00`.  The loop immediately stops -- it reads nothing.

greg-king5: `ptr1+0` and `ptr1+1` aren't used as a 16-bit counter; they're two 8-bit counters.  `ptr1+0` counts a loop that's nested inside of a second loop that's counted by `ptr1+1`.

greg-king5:
Sven, there are many functions in the libraries that use the "-size-1" hack.  All of them can benefit from this better hack.

libsrc/cbm/cbm_read.s
jmr: I mis-parsed the comment and forgot about the increment when looking at the loop guard. The new comment is much better. Nice hack!

oliverschmidt:
Sorry, guys, you lost me here...

@greg-king5: Please proceed / merge as you see fit - thanks!

SvenMichaelKlose:
> Sven, there are many functions in the libraries that use the "-size-1" hack. All of them can benefit from this better hack.
Did a scan for similar comments only.  So far there wasn't more of it. Will gladly update stuff I might find next.

SvenMichaelKlose:
> Sorry guys, you lost me here...

The numerical series has been shifted by +1 to make use of the zero flag. Backdraw: the value being counted down is unusable for anything. That's what doing everybody's head in. Including mine.

SvenMichaelKlose:
Found more by grep'ing for "eor #$FF" but will get back to that after some nap.

libsrc/common/memcmp.s
```diff
@@ -33,7 +33,7 @@ _memcmp:
 
 ; Head of compare loop: Test for the end condition
 
-Loop:   inx                     ; Bump low byte of (-count-1)
+Loop:   dex                     ; Bump low byte of (-count-1)
```
greg-king5, line 36: Those two comments need to be changed; it no longer is "(-count-1)".

libsrc/common/strncpy.s
```diff
@@ -26,9 +26,9 @@
 
         ldx     tmp1            ; Load low byte of ones complement of size
```
greg-king5, line 27: That comment should be changed; it isn't "ones complement", now.

libsrc/common/strncpy.s
```diff
@@ -42,7 +42,7 @@ L2:     lda     (ptr1),y        ; Copy one character
 
 ; Fill the remaining bytes.
 
-L3:     inx                     ; Counter low byte
+L3:     dex                     ; Counter low byte
         beq     L6              ; Branch on overflow
```
greg-king5, line 46: It's not overflow, now.

libsrc/common/strnicmp.s
```diff
@@ -40,7 +34,7 @@ _strncasecmp:
 
 ; Start of compare loop. Check the counter.
 
-Loop:   inc     ptr3
+Loop:   dec     ptr3
         beq     IncHi           ; increment high byte
```
greg-king5, line 38: That comment is on the wrong line; and, it's "decrement", now.

libsrc/osic1p/bootstrap.s
```diff
@@ -50,16 +50,17 @@ LINEDIST        =       $20             ; Offset in video RAM between two lines
...
-        eor     #$FF
-        sta     count                   ; store (-size - 1)
-        lda     #>load_size
-        eor     #$FF
-        sta     count+1
-
-L1:     inc     count                   ; pre-count one's-complement upwards
+
+        ldx     #<load_size
+        inx
```
greg-king5, line 55: `load_size` is a constant; you can do the increment in the expression:
```asm
ldx #(<load_size) + 1
stx count
```

libsrc/osic1p/bootstrap.s
```diff
@@ -50,16 +50,17 @@ LINEDIST        =       $20             ; Offset in video RAM between two lines
...
+        stx     count
+        ldx     #>load_size
+        inx
+        stx     count+1                 ; save size with each byte incremented separately
+
+L1:     dec     count                   ; pre-count one's-complement upwards
```
greg-king5, line 61: Please fix the comment.

SvenMichaelKlose:
I bet the fixes could have been done faster than writing the comments. Whatever...

polluks:
@SvenMichaelKlose For sure but what about the educational aspects? ;-)

SvenMichaelKlose:
Just got a little sloppy… was working far too much beforehand. Will better fix that tomorrow morning.

greg-king5:
Only one "nit" remains.  I'll merge this, and pick that nit, myself.

Issue #1308, opened 2020-10-26
optimization: use more unsigned


jmr:
if the div/mod code generators could know that the LHS is guaranteed to be non-negative, then they could just take the "unsigned code paths"

_Originally posted by @acqn in https://github.com/cc65/cc65/issues/1298#issuecomment-716154438_

For uchar op uchar, we can use use the uint op uint code path, but then change the result type to int. This should give the same answer with faster code.

With a bit more effort, it can be extended to operations with constants that are representable by uchar.

It's a bit annoying and fiddly, since usually the integral promotions (char/uchar -> int) happen independently for each arg, before the usual arithmetic conversions, but now we need to look at both types together.

PR #1315, opened 2020-10-29
Use u16 codegen for u8 x u8 ops

jmr:
In g_typeadjust, before we apply the integral promotions, we check if
both types are unsigned char.  If so, we promote to unsigned int, rather
than int, which would be chosen by the standard rules.  This is only a
performance optimization and does not affect correctness, as the flags
returned by g_typeadjust are only used for code generation, and not to
determine types of other expressions containing this one.  All unsigned
char bit-patterns are valid as both int and unsigned int and represent
the same value, so either signed or unsigned int operations can be used.
This special case part is not duplicated by ArithmeticConvert.

Partial fix for #1308.

### DIFF
--- a/src/cc65/codegen.c
+++ b/src/cc65/codegen.c
@@ -1432,6 +1432,19 @@ unsigned g_typeadjust (unsigned lhs, unsigned rhs)
 
     /* Note that this logic is largely duplicated by ArithmeticConvert. */
 
+    /* Before we apply the integral promotions, we check if both types are unsigned char.
+    ** If so, we return unsigned int, rather than int, which would be returned by the standard
+    ** rules.  This is only a performance optimization and does not affect correctness, as
+    ** the flags are only used for code generation, and not to determine types of other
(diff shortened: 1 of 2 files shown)


jmr:
@acqn Does this look reasonable?

acqn:
> 
> 
> @acqn Does this look reasonable?

Hackish but reasonable to me. It's a good idea to pick the low-hanging fruits for now, and we might be able to optimize some more cases someday (if we could have an AST).

BTW, Is there a test case for mod?

jmr:
> we might be able to optimize some more cases someday (if we could have an AST).

Yeah, I was thinking about that, too. Is an `ExprDesc` just built up for one statement at a time currently?

> BTW, Is there a test case for mod?

Not currently. Is there a reason to expect it will be different, or is everything handled uniformly? In #1321, I add a test for `unsigned char x; x % 4`, so we can eventually make sure that's compiled to bitwise-and. I have a PR for that waiting, but want to get this merged first.

PR #1322, opened 2020-11-01
[cc65] Fixed internal representation of calculated constant results

### LINKED ISSUE #1310: surprising constant expression evaluation
`255 * 255 == -511` is false. I guess this is standard-compliant, since it involves signed integer overflow, but it would be nicer if it (1) gave a warning (2) also evaluated to true, since `int x = 255; x * x == -511` and `int y = 255 * 255; y == -511;` are both true.

```
#include <stdio.h>

static unsigned char failures = 0;

int main (void)
{
    const int two_fifty_five = 255;
    const int two_fifty_five_squared = 255 * 255;

    if (255 * 255 != -511) {
        fprintf (stderr, "Expected 255 * 255 == -511, got: %d\n", 255 * 255);
        failures++;
    }
    if (two_fifty_five * two_fifty_five != -511) {
        fprintf (stderr, "Expected two_fifty_five * two_fifty_five == -511, got: %d\n",
                 two_fifty_five * two_fifty_five);
        failures++;
    }
    if (two_fifty_five_squared != -511) {
        fprintf (stderr, "Expected two_fifty_five_squared == -511, got: %d\n",
                 two_fifty_five_squared);
        failures++;
    }
    printf ("failures: %u\n", failures);
    return failures;
}
```

Gives:
```
Expected 255 * 255 == -511, got: -511
failures: 1
```

acqn:
- [X] Fixed #1310 as well as some more cases.
Note: The numeric constant results of unary and binary calculations are stored inside `Expr->IVal`, which are as expected to be mathematically correct after being converted to its type. On the other hand, the offsets of expressions are not converted.
- [X] Fixed a bug that pointer subtraction results from two absolute addresses are calculated as `unsigned long`.
- [X] Test cases for the issues above.

### DIFF
--- a/src/cc65/expr.c
+++ b/src/cc65/expr.c
@@ -3347,7 +3347,7 @@ static void parsesub (ExprDesc* Expr)
                     Error ("Incompatible pointer types");
                 } else {
                     Expr->IVal = (Expr->IVal - Expr2.IVal) /
-                                      CheckedPSizeOf (lhst);
+                                      (long)CheckedPSizeOf (lhst);
                 }
                 /* Operate on pointers, result type is an integer */
                 Expr->Type = type_int;
--- a/src/cc65/expr.c
+++ b/src/cc65/expr.c
@@ -251,6 +251,42 @@ static unsigned typeadjust (ExprDesc* lhs, ExprDesc* rhs, int NoPush)
 
 
 
+static void LimitExprValue (ExprDesc* Expr)
+/* Limit the constant value of the expression to the range of its type */
+{
+    switch (GetUnderlyingTypeCode (Expr->Type)) {
+        case T_INT:
+        case T_SHORT:
+            Expr->IVal = (int16_t)Expr->IVal;
+            break;
+
+        case T_UINT:
+        case T_USHORT:
+        case T_PTR:
+        case T_ARRAY:
+            Expr->IVal = (uint16_t)Expr->IVal;
+            break;
+
+        case T_LONG:
+        case T_ULONG:
+            /* No need to do anything */
+            break;
+
+        case T_SCHAR:
+            Expr->IVal = (int8_t)Expr->IVal;
+            break;
+
+        case T_UCHAR:
+            Expr->IVal = (uint8_t)Expr->IVal;
+            break;
+
+        default:
+            Internal ("hie_internal: constant result type %s\n", GetFullTypeName (Expr->Type));
+    }
+}
+
+
+
 static const GenDesc* FindGen (token_t Tok, const GenDesc* Table)
 /* Find a token in a generator table */
 {
@@ -2159,15 +2195,19 @@ static void UnaryOp (ExprDesc* Expr)
         ED_MakeConstAbsInt (Expr, 1);
     }
 
-    /* Check for a constant expression */
+    /* Check for a constant numeric expression */
     if (ED_IsConstAbs (Expr)) {
-        /* Value is constant */
+        /* Value is numeric */
         switch (Tok) {
             case TOK_MINUS: Expr->IVal = -Expr->IVal;   break;
             case TOK_PLUS:                              break;
             case TOK_COMP:  Expr->IVal = ~Expr->IVal;   break;
             default:        Internal ("Unexpected token: %d", Tok);
         }
+
+        /* Limit the calculated value to the range of its type */
+        LimitExprValue (Expr);
+
     } else {
         /* Value is not constant */
         LoadExpr (CF_NONE, Expr);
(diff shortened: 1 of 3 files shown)


src/cc65/expr.c
```diff
@@ -2501,6 +2501,37 @@ static void hie_internal (const GenDesc* Ops,   /* List of generators */
                 }
             }
 
+            /* Limit the calculated value to the range of its type */
+            switch (GetUnderlyingTypeCode (Expr->Type)) {
+                case T_INT:
+                case T_SHORT:
+                    Expr->IVal = (signed short)Expr->IVal;
```
jmr, line 2508: Nit: Maybe `CHECK (sizeof (short) == 2)`, unless this is a global assumption.

jmr:
What about unary operators? Can't `~` and `-` make unwanted high bits? (Maybe this is already handled somewhere for all I know.)

There are tests in #1310 that will need to be moved to `val/` when that is merged.  The tests from #1312 that were `#if 0`-ed out should also be enabled.  https://github.com/cc65/cc65/pull/1312/files

src/cc65/expr.c
```diff
@@ -2501,6 +2501,37 @@ static void hie_internal (const GenDesc* Ops,   /* List of generators */
                 }
             }
 
+            /* Limit the calculated value to the range of its type */
```
jmr, line 2504: It looks like `+` and `-` are handled in `parseadd` and `parsesub`. Check everywhere `IVal` is assigned.

acqn: Nice catch. Fix for arithmetic `+` and `-` upcoming.

acqn:
> 
> 
> What about unary operators? Can't `~` and `-` make unwanted high bits? (Maybe this is already handled somewhere for all I know.)

True. Example:
```c
    if (-32768U != 32768U) {
        fprintf (stderr, "Expected -32768U == 32768U, got: %ld\n", (long)-32768U);
        failures++;
    }
```

There might be obscure ones: `p[n]`, `s.a` and `s->b`, but I suppose they would already become UB in that case.

src/cc65/expr.c
acqn: This should be a global static assertion, together with `sizeof (long) == 4`.

jmr: Apparently, this is supposed to compile with C89, so C11 is a bit out of the question.

I'm going to try to add `stdint.h`-like typedefs to `common/inttypes.h` and see if that gets merged. Then this can just be `int16_t`.

acqn:
There are quite a few naive checks like `Expr->IVal < 0` that would be broken by a simple "always store a C correct value". The issue requires much more work than this.

greg-king5:
> There are quite a few naive checks like `Expr->IVal < 0` that would be broken by a simple "always store a cc65-correct value" fix. The issue requires much more work than this.

`Expr->IVal = (signed short)Expr->IVal;` is evaluated as
`Expr->IVal = (signed long)(signed short)(signed long)Expr->IVal;`.  Won't normal sign extension handle negative-sign tests?

jmr:
> There are quite a few naive checks like `Expr->IVal < 0` that would be broken by a simple "always store a C correct value" fix. The issue requires much more work than this.

Then maybe split out the pieces that can be submitted already, and put comments where things are wrong but break other places. Then we can pick those apart later.

src/cc65/expr.c
jmr: > This should be a global static assertion, together with `sizeof (long) == 4`.

I've been assuming the assumption was `sizeof(long) >= 4`. `sizeof(long) == 8` on most 64-bit platforms, AFAIK. Using `int32_t` would make this clear. #1326 adds these for C89 compilers.

acqn: Resolved.

acqn:
You are right. I think this is done right and can be merged now.

test/val/bug1310.c
```diff
@@ -68,6 +68,21 @@ int main (void)
...
 
+    if (-32768U != 32768U) {
+        fprintf (stderr, "Expected -32768U == 32768U, got: %ld\n", (long)-32768U);
+        failures++;
+    }
+    if (~32767U != 32768U) {
+        fprintf (stderr, "Expected ~32767U == 32768U, got: %ld\n", (long)~32767U);
+        failures++;
+    }
     printf ("failures: %u\n", failures);
```
greg-king5, line 79: A `printf()` is below the next test.  Therefore, this line must be removed (it's redundant).

test/val/bug1310.c
```diff
@@ -68,6 +68,21 @@ int main (void)
...
+
+    if ((long*)0x1000 - (long*)0x2000 >= 0) {
+        fprintf (stderr, "Expected (long*)0x1000 - (long*)0x2000 < 0, got: %d\n", (long*)0x1000 - (long*)0x2000 >= 0);
```
greg-king5, line 82: The third argument should be only the subtraction.  It shouldn't have the comparison.

test/val/bug1310.c
acqn: Fixed.

test/val/bug1310.c
acqn: Fixed.

greg-king5 (approved):
This looks good; merging.

Issue #1298, opened 2020-10-24
Atari: Setting pixel data 10x slower than before


8bit-Dude:
This is a strange bug, probably one for Chris Groessler!  :-)

I have been using a 2018 version of CC65 for a while, and decided to upgrade to latest for next release of 8bit-Unity. Most functions work normally (such as loading bitmaps), but one function for setting pixels has become incredibly slow.
The code is as follows (compiled using default atarixl.cfg):
```c
#define BITMAPRAM1 (0x7010) // 7010-8f50 (bitmap frame 1)
#define BITMAPRAM2 (0xa010) // a010-bf50 (bitmap frame 2)

void SetPixel(unsigned int pixelX, unsigned int pixelY, unsigned char color)
{
	unsigned int offset;
	unsigned char shift, mask, col1, col2;	

	// Compute pixel location
	offset = 40*pixelY + pixelX/4;
	shift = 6 - 2*(pixelX%4);
	mask = 255 - (3 << shift);
	if ((pixelY+pixelX)%2) {
		col2 = (color%4) << shift;
		col1 = (color/4) << shift;
	} else {
		col1 = (color%4) << shift;
		col2 = (color/4) << shift;
	}

	// Set color/color2 in dual buffer
	POKE((char*)BITMAPRAM1+offset, (PEEK((char*)BITMAPRAM1+offset) & mask) | col1);
	POKE((char*)BITMAPRAM2+offset, (PEEK((char*)BITMAPRAM2+offset) & mask) | col2);
}
```
I wonder if the system is waiting for a VSYNC or something, whenever I touch the bitmap ram?

oliverschmidt:
I guess that the issue rather relates to different code generated by the compiler...

polluks:
Indeed, no job for Groessler. Firstly you may turn off the optimiser.

8bit-Dude:
I am trying to produce a sample code that replicates the issue without needing 8bit-unity.

jmr:
Some of it is probably from the changes in integral conversion semantics, but 10x seems extreme. It should be easy to put the above code in a loop and test with `sim65 -c`, if this function is indeed the problem.

8bit-Dude:
Alright, I have an interesting benchy right here:

`
#include <stdlib.h>
#include <conio.h>
#include <peekpoke.h>
#include <time.h>


#define BITMAPRAM1 (0x7010) // 7010-8f50 (bitmap frame 1)
[...]

PR #1328, opened 2020-11-10
Adjust type of int constants that fit in char

jmr:
When there is an integral constant like `3` in an expression, it has
type `int` according to the C spec, even though it can be represented
as an `unsigned char`.  Change codegen (`hie_internal` and `typeadjust`)
to treat it as `unsigned char` instead of `int` so that faster,
unsigned operations can be used.

For the test case in #1298, reduces the cycles per iteration from
4311 to 1884.  This is a great improvement, but still much worse than
the 1053 cycles/iter from `c4698df~`, so more work remains to be done.

Further partial fix for #1298 and #1308.


oliverschmidt:
@acqn, @greg-king5: Any feedback on this PR?

greg-king5:
:+1:

Issue #1320, opened 2020-10-31
Access violation in cc65


mzattera:
**cl65: Cannot execute 'D:\cc6529\bin\cc65': No error**

This error appeared in Windows snapshot I downloaded on Oct. 29th and later, but NOT in the snapshot I downloaded on Oct 19th.

It happens compiling javavm\interpreter.c in the attached sources. It sucessfully compiles other files.

Verions from Oct. 19th compiles the whole solution without issues.

[src.zip](https://github.com/cc65/cc65/files/5469292/src.zip)

mzattera:
Line that causes the error:

D:\cc6529\bin\cl65 -c --cpu 6502 -Oi -t none -I .\javavm -I .\platform\c64 -I .\util javavm\interpreter.c

oliverschmidt:
The issue at hand is caused by an access violation in cc65. It can be reproduced by compiling

```
char *var;

void foo(char, char);
char bar(void);

void main(void)
{
  foo(*var++, bar());
}
```
without any options.

The access violation happens in https://github.com/cc65/cc65/blob/master/src/cc65/datatype.c#L706 because `GetUnderlyingTypeCode()` is called with a NULL pointer.

In the VC++ debug build the callstack looks like this:
```
 	cc65.exe!GetUnderlyingTypeCode(const Type * Type) Line 706	C
 	cc65.exe!SizeOf(const Type * T) Line 747	C
 	cc65.exe!CheckedSizeOf(const Type * T) Line 834	C
 	cc65.exe!DoDeferred(unsigned int Flags, ExprDesc * Expr) Line 545	C
 	cc65.exe!FunctionParamList(FuncDesc * Func, int IsFastcall, ExprDesc * ED) Line 784	C
 	cc65.exe!FunctionCall(ExprDesc * Expr) Line 882	C
 	cc65.exe!hie11(ExprDesc * Expr) Line 1762	C
 	cc65.exe!hie10(ExprDesc * Expr) Line 2329	C
 	cc65.exe!ExprWithCheck(void(*)(ExprDesc *) Func, ExprDesc * Expr) Line 112	C
 	cc65.exe!hie_internal(const GenDesc * Ops, ExprDesc * Expr, void(*)(ExprDesc *) hienext, int * UsedGen) Line 2360	C
 	cc65.exe!hie9(ExprDesc * Expr) Line 2973	C
[...]

PR #1331, opened 2020-11-13
g_typeadjust: Use CF_CHAR for char args

jmr:
If lhs and rhs are either both signed char or both unsigned char,
return flags for that type instead of (unsigned) int.  The flags
are used only for codegen.  Currently, this does nothing, since
codegen treats chars as ints unless CF_FORCECHAR is set, but it
allows more efficient char x char -> int codegen to be added in
the future.

### DIFF
--- a/src/cc65/codegen.c
+++ b/src/cc65/codegen.c
@@ -1433,17 +1433,20 @@ unsigned g_typeadjust (unsigned lhs, unsigned rhs)
 
     /* Note that this logic is largely duplicated by ArithmeticConvert. */
 
-    /* Before we apply the integral promotions, we check if both types are unsigned char.
-    ** If so, we return unsigned int, rather than int, which would be returned by the standard
-    ** rules.  This is only a performance optimization and does not affect correctness, as
-    ** the flags are only used for code generation, and not to determine types of other
-    ** expressions containing this one.  All unsigned char bit-patterns are valid as both int
-    ** and unsigned int and represent the same value, so either signed or unsigned int operations
-    ** can be used.  This special case part is not duplicated by ArithmeticConvert.
+    /* Before we apply the integral promotions, we check if both types are the same character type.
+    ** If so, we return that type, rather than int, which would be returned by the standard
+    ** rules.  This is only a performance optimization allowing the use of unsigned and/or char
+    ** operations; it does not affect correctness, as the flags are only used for code generation,
+    ** and not to determine types of other expressions containing this one.  For codgen, CF_CHAR
+    ** means the operands are char and the result is int (unless CF_FORCECHAR is also set, in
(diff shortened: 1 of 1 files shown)


src/cc65/codegen.c
```diff
@@ -1432,17 +1432,19 @@ unsigned g_typeadjust (unsigned lhs, unsigned rhs)
...
+    ** means the operands are char and the result is int (unless CF_FORCECHAR is also set, in
+    ** which case the result is char).  This special case part is not duplicated by
+    ** ArithmeticConvert.
     */
-    if ((lhs & CF_TYPEMASK) == CF_CHAR && (lhs & CF_UNSIGNED) &&
-        (rhs & CF_TYPEMASK) == CF_CHAR && (rhs & CF_UNSIGNED)) {
-        return const_flag | CF_UNSIGNED | CF_INT;
+    if ((lhs & CF_TYPEMASK) == CF_CHAR && (rhs & CF_TYPEMASK) == CF_CHAR &&
+        (lhs & CF_UNSIGNED) == (rhs & CF_UNSIGNED)) {
+        const unsigned unsigned_flag = lhs & rhs & CF_UNSIGNED;
```
greg-king5, line 1446: `lhs` and `rhs` don't need to be ANDed.  The two flags are the same; therefore, either one can be used by itself.

jmr: I found the lack of symmetry disturbing, but ok.

oliverschmidt:
I think https://github.com/cc65/cc65/issues/1320 needs to be fixed before merging any other compiler-related PR.

oliverschmidt:
From my POV https://github.com/cc65/cc65/issues/1320 and https://github.com/cc65/cc65/issues/1348 have shown that we still have way too few tests to validate the compiler functionality. That fact combined with my missing understanding of the compiler internals leaves me pretty weak in my role as maintainer. @jmr: Can you accept that I rather don't want to merge any changes to the code generation and/or optimizer while there are still known unresolved regressions? I hope you know that my reluctance doesn't mean that I'm not interested in your contributions.

polluks:
"If you TDD your code you should get a 90%+ coverage as a bonus."
https://stackoverflow.com/questions/90002/what-is-a-reasonable-code-coverage-for-unit-tests-and-why

oliverschmidt:
Same thread:

"Code coverage is an objective measurement: Once you see your coverage report, there is no ambiguity about whether standards have been met are useful. Does it prove correctness? Not at all, but it has a clear relationship to how well-tested the code is, which in turn is our best way to increase confidence in its correctness. Code coverage is a measurable approximation of immeasurable qualities we care about."

I might be wrong easily, but I imagine that even if we have 100% code coverage in the compiler, we can still have regressions like the two i mentioned above. I don't see ((too) low) code coverage as our issue.

Issue #1332, opened 2020-11-14
Bitfield wrongly printed


mzattera:
The below code prints `b.msb` as `37`, still it evaluates as ==1 in the `if`.
Tested with version `V2.18 - Git 0f66f75`

```C
#include <stdio.h>

struct bitfield {
	unsigned int lsb : 1;
	unsigned int pad : 14;
	unsigned int msb : 1;
};

int main() {
	struct bitfield b;

	b.lsb = 1;
	b.pad = 0;
	b.msb = 1;

	printf("lsb =%u\n", b.lsb);
	printf("pad =%u\n", b.pad);
	printf("msb =%u\n", b.msb);

	if (b.msb == 1)
		printf("msb is 1\n");
	if (b.msb)
		printf("msb is true\n");

	return 0;
}
```

jmr:
I haven't had time to look in detail, but `unsigned x = b.msb; printf (..., x);` seems to work, so you have a work-around.

jmr:
Something related to varargs, but it doesn't always fail with varargs. See test cases in #1334.

greg-king5:
"msb-test.c":
```c
#include <stdio.h>

static const struct bitfield {
    unsigned int lsb: 1;
    unsigned int pad: 14;
    unsigned int msb: 1;
} b = {1, 0, 1};

static unsigned v;

int main(void)
{
    printf("b.lsb = %u\n", b.lsb);
    printf("b.pad = %u\n", b.pad);
    printf("b.msb = %u\n\n", b.msb);

    // This stops the printf()'s return value
    // from tricking the if () test into succeeding.
    v = 0x1234;

    if (b.msb == 1)
        printf("b.msb is 1.\n");

    if (b.msb)
        printf("b.msb is true.\n");

    v = 0x1234;
    return b.msb;
}
```
```
$ cl65 -t sim6502 msb-test.c
$ sim65 msb-test
b.lsb = 1
b.pad = 0
b.msb = 21

b.msb is true.
```
The program returned an "error-code" of 37.

Here is some Assembly that shows the bad code generation:
```asm
;
; if (b.msb == 1)
;
        lda     _b+1
        jsr     shrax4  ; <- .X register has garbage!
        jsr     shrax3
        ldx     #$00
        cpx     #$00
        bne     L0003
        cmp     #$01    ; <- .A now has some garbage
L0003:  jsr     booleq
        jeq     L0002
```
The bit-field is accessed as UCHAR, but the right-shifts are UINT.

[...]

PR #1334, opened 2020-11-15
Add test case for #1332

jmr:


greg-king5:
This test case needs to be changed because the bug isn't in the varargs handling.

jmr:
Is it in how the varargs functions are called? The only failures I've found so far involve varargs functions.

greg-king5:
It has nothing to do with varargs.  In the original issue example, the first `if ()` test actually is broken; it worked by accident.  I'm creating a new example for that issue that shows what really is happening.  (I posted here just to stop Oliver from merging this PR too soon.)

jmr:
I trimmed down the test cases, keeping one similar to yours and one similar to the original bug report.

greg-king5:
OK, my fix passes this test.

PR #1336, opened 2020-11-15
g_asr, g_asl: Use ROL/ROR for char shifts >= 6

jmr:
Instead of `val` right shifts, we can also do `8 - val` left rotates and
a mask.  This saves 4 instructions for `val == 7` and 1 instruction for
`val == 6`.

### DIFF
--- a/src/cc65/codegen.c
+++ b/src/cc65/codegen.c
@@ -3111,9 +3111,26 @@ void g_asr (unsigned flags, unsigned long val)
         switch (flags & CF_TYPEMASK) {
             case CF_CHAR:
                 if (flags & CF_FORCECHAR) {
-                    if ((flags & CF_UNSIGNED) != 0 && val < 8) {
-                        while (val--) {
-                            AddCodeLine ("lsr a");
+                    val &= 7;
+                    if ((flags & CF_UNSIGNED) != 0) {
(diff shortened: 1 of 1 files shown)


greg-king5:
You should count bytes and cycles (not instructions) when comparing new code versus old code.
The new code is worse for `val == 6`.  Only `val == 7` can benefit from it.

jmr:
I think what I wrote was correct, since all instructions involved are two cycles. You're right that it was more bytes for `val == 6`.

I've now rewritten it to be better for both 6 and 7, and included the byte and cycle counts.

oliverschmidt:
I think https://github.com/cc65/cc65/issues/1320 needs to be fixed before merging any other compiler-related PR.

IrgendwerA8:
Haven't examined the code in-depth, but wouldn't just clearing the carry (CLC) before rotation save another byte as you don't need the AND then?

oliverschmidt:
See https://github.com/cc65/cc65/pull/1331#issuecomment-735424994.

IrgendwerA8:
Now, where those mentioned issues are closed, what about merging this one?

oliverschmidt:
@greg-king5: Do you want to have another look at this PR or is just "everybody" waiting for me to merge?

PR #1346, opened 2020-11-19
Update to correct prerequisite for GEOS calcblksfree() (issue #1318)

### LINKED ISSUE #1318: GEOSLib - CalcBlksFree issue
Either the documentation isnt completely clear, or there is a bug in this function.  According to the docs, calling GetDirHead() before calling this is all that is needed.  But under my project, it continually returned an odd value.  

Checking some other documentation, you must first set r5 to the address of curDirHead (which for CBM, is 0x8200).  Upon doing this manually, the CalcBlksFree function returns the proper number of blocks free.

So instead of:

GetDirHead();
x = GetBlksFree();

it is:

GetDirHead();
r5 = 0x8200;  // or constant curDirHead
x = GetBlksFree();

I didnt know if this was a documentation issue, or if a code issue.  (or neither?  Perhaps it is assumed that a geos developer would know this?)  If a code issue, then adding:

LoadW r5, curDirHead  (or ca65 equivalent)

...before the jsr to _CalcBlksFree  should take care of it (in calcblksfree.s)

xlar54:
Fixes #1318.

### DIFF
--- a/libsrc/geos-common/disk/calcblksfree.s
+++ b/libsrc/geos-common/disk/calcblksfree.s
@@ -13,6 +13,10 @@
             .include "geossym.inc"
         
 _CalcBlksFree:
+        lda #>curDirHead
+        sta r5H
+        lda #<curDirHead
+        sta r5L
         jsr CalcBlksFree
         stx __oserror
         lda r4L


libsrc/geos-common/disk/calcblksfree.s
```diff
@@ -13,6 +13,10 @@
             .include "geossym.inc"
         
 _CalcBlksFree:
+        lda #>curDirHead
+        sta r5H
+        lda #<curDirHead
+        sta r5L
```
oliverschmidt, line 19: While the code is correct it doesn't fit the usual cc65 style. In cc65 Asm code we use the combination of A and X to represent 16-bit values (especially pointers). You'll often find the term AX to represent that (A for the low byte and X for the high byte). And when moving those 16-bite values around we start with the low byte. So the code here should read

```
lda #<curDirHead
ldx #>curDirHead
sta r5L
stx r5H
```

If you provide an update to this pull request I can merge it as-is.

oliverschmidt:
I replaced this pull request with the direct commit https://github.com/cc65/cc65/commit/79bdc2d51f1bf67830ef6fcde3e6cc6bf9c72bfa.

Issue #1345, opened 2020-11-19
More consistent configs


polluks:
Maybe
start → offset
VECTOR → VECTORS
````
grep VECTOR cfg/*
cfg/atari2600.cfg:    VECTORS:  load = ROM,            type = ro,                                start = $FFFA;
cfg/creativision.cfg:    VECTORS:  load = ROM, run = RAM, type = rw,  define = yes;
cfg/nes.cfg:    VECTORS:  load = ROMV,            type = rw;
cfg/pce.cfg:    VECTORS:  load = ROM,             type = ro,  start = $FFF6;
cfg/supervision-128k.cfg:    VECTOR:   load = ROM,      type = ro,                  offset = $3FFA;
cfg/supervision-16k.cfg:    VECTOR:   load = ROM,            type = ro,  offset   = $3FFA;
cfg/supervision-64k.cfg:    VECTOR:   load = ROM,      type = ro,                  offset = $3FFA;
cfg/supervision.cfg:    VECTOR:   load = ROM,            type = ro,  offset   = $7FFA;
````

greg-king5:
> ```
> $ grep -rF "VECTOR" cfg
> cfg/atari2600.cfg:    VECTORS:  load = ROM,            type = ro,                                start = $FFFA;
> cfg/creativision.cfg:    VECTORS:  load = ROM, run = RAM, type = rw,  define = yes;
> cfg/nes.cfg:    VECTORS:  load = ROMV,            type = rw;
> cfg/pce.cfg:    VECTORS:  load = ROM,             type = ro,  start = $FFF6;
> cfg/supervision-128k.cfg:    VECTOR:   load = ROM,      type = ro,                  offset = $3FFA;
> cfg/supervision-16k.cfg:    VECTOR:   load = ROM,            type = ro,  offset   = $3FFA;
> cfg/supervision-64k.cfg:    VECTOR:   load = ROM,      type = ro,                  offset = $3FFA;
> cfg/supervision.cfg:    VECTOR:   load = ROM,            type = ro,  offset   = $7FFA;
> ```

I agree with `VECTOR → VECTORS`.  But, I think that `offset → start` makes more sense.  Those vectors sit at fixed hardware-dictated addresses.

oliverschmidt:
> I agree with VECTOR → VECTORS.

+1

> But, I think that offset → start makes more sense. Those vectors sit at fixed hardware-dictated addresses.

+1

[...]

PR #1349, opened 2020-11-20
Unified configs for #1345

polluks2:


cfg/supervision-128k.cfg
```diff
@@ -33,5 +33,5 @@ SEGMENTS {
     ZEROPAGE: load = RAM,      type = bss, define   = yes;
     DATA:     load = RAM,      type = bss, define   = yes, offset = $0200;
     BSS:      load = RAM,      type = bss, define   = yes;
-    VECTOR:   load = ROM,      type = ro,                  offset = $3FFA;
+    VECTORS:  load = ROM,      type = ro,                  start  = $3FFA;
```
oliverschmidt, line 36: According to https://cc65.github.io/doc/ld65.html#ss5.7 the `offset` is relative to the start of the memory area. The `ROM` memory area starts at $C000. So the effective address of this section was $FFFA (which nicely matches the 6502 hardware ;-). In contrast the `start` is an absolute value - so at least from my POV this needs to read
```
start = $FFFA
```

cfg/supervision-128k.cfg
```diff
@@ -33,5 +33,5 @@ SEGMENTS {
     ZEROPAGE: load = RAM,      type = bss, define   = yes;
     DATA:     load = RAM,      type = bss, define   = yes, offset = $0200;
```
oliverschmidt, line 34: In general I don't understand the meaning of this line.
* Why is the `DATA` segment of type `bss` instead of type `rw`?
* The `RAM` memory area starts at $0000. So the `DATA` segment starts at $0200. Is that desired? Why?

cfg/supervision-128k.cfg
```diff
@@ -33,5 +33,5 @@ SEGMENTS {
...
     BSS:      load = RAM,      type = bss, define   = yes;
```
oliverschmidt, line 35: * As far as I understand this makes the `BSS` segment start right after the `ZEROPAGE` segment. In case the `ZEROPAGE` segment shouldn't be exactly $100 bytes this means the `BSS` partly lives in the zeropage. Is that desired? Why?
* If the `ZEROPAGE` segment should be e.g. $20 bytes the `BSS` segment can't be larger than $1E0 bytes because the `DATA'  segment is fixed to offset $0200. Is that desired? Why?

cfg/supervision-16k.cfg
```diff
@@ -21,7 +21,7 @@ SEGMENTS {
     RODATA:   load = ROM,            type = ro,  define   = yes;
     DATA:     load = ROM, run = RAM, type = rw,  define   = yes;
     FFF0:     load = ROM,            type = ro,  offset   = $3FF0;
-    VECTOR:   load = ROM,            type = ro,  offset   = $3FFA;
+    VECTORS:  load = ROM,            type = ro,  start    = $3FFA;
```
oliverschmidt, line 24: Same as for supervision-128k.cfg.

cfg/supervision-16k.cfg
```diff
@@ -21,7 +21,7 @@ SEGMENTS {
     RODATA:   load = ROM,            type = ro,  define   = yes;
     DATA:     load = ROM, run = RAM, type = rw,  define   = yes;
     FFF0:     load = ROM,            type = ro,  offset   = $3FF0;
```
oliverschmidt, line 23: If modifying `VECTORS` I suggest to modify `FFF0` too.

cfg/supervision-64k.cfg
```diff
@@ -26,5 +26,5 @@ SEGMENTS {
     ZEROPAGE: load = RAM,      type = bss, define   = yes;
     DATA:     load = RAM,      type = bss, define   = yes, offset = $0200;
     BSS:      load = RAM,      type = bss, define   = yes;
-    VECTOR:   load = ROM,      type = ro,                  offset = $3FFA;
+    VECTORS:  load = ROM,      type = ro,                  start  = $3FFA;
```
oliverschmidt, line 29: Same as for supervision-128k.cfg.

cfg/supervision.cfg
```diff
@@ -20,7 +20,7 @@ SEGMENTS {
     RODATA:   load = ROM,            type = ro,  define   = yes;
     DATA:     load = ROM, run = RAM, type = rw,  define   = yes;
     FFF0:     load = ROM,            type = ro,  offset   = $7FF0;
-    VECTOR:   load = ROM,            type = ro,  offset   = $7FFA;
+    VECTORS:  load = ROM,            type = ro,  start    = $7FFA;
```
oliverschmidt, line 23: Same as for supervision-128k.cfg.

cfg/supervision.cfg
```diff
@@ -20,7 +20,7 @@ SEGMENTS {
     RODATA:   load = ROM,            type = ro,  define   = yes;
     DATA:     load = ROM, run = RAM, type = rw,  define   = yes;
     FFF0:     load = ROM,            type = ro,  offset   = $7FF0;
```
oliverschmidt, line 22: If modifying `VECTORS` I suggest to modify `FFF0` too.

oliverschmidt (changes requested):
As far as I can see the pull request as-is would break the configurations.

polluks:
Hmm, why have 64 and 128 no features definition?

oliverschmidt:
> Hmm, why have 64 and 128 no features definition?

According to my understanding that's just a plain bug.

oliverschmidt:
What about my suggestion to have (nomen est omen)
`FFF0: load = ROM, type = ro, start = $FFF0;`

oliverschmidt (approved):

PR #1352, opened 2020-11-28
Add STATUS definition

ops:
Added definition of STATUS ($90) for Commodore targets.

### DIFF
--- a/asminc/c128.inc
+++ b/asminc/c128.inc
@@ -7,6 +7,7 @@
 ; Zero page, Commodore stuff
 
 TXTPTR          := $3D          ; Pointer into BASIC source code
+STATUS          := $90          ; Kernal I/O completion status
 TIME            := $A0          ; 60HZ clock
 FNAM_LEN        := $B7          ; Length of filename
 SECADR          := $B9          ; Secondary address
--- a/asminc/c64.inc
+++ b/asminc/c64.inc
@@ -9,6 +9,7 @@
 VARTAB          := $2D          ; Pointer to start of BASIC variables
 MEMSIZE         := $37          ; Pointer to highest BASIC RAM location (+1)
 TXTPTR          := $7A          ; Pointer into BASIC source code
+STATUS          := $90          ; Kernal I/O completion status
 TIME            := $A0          ; 60 HZ clock
 FNAM_LEN        := $B7          ; Length of filename
 SECADR          := $B9          ; Secondary address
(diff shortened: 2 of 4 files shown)


spiro-trikaliotis:
Why? The KERNAL call READST accesses exactly that variable.

Of course, one could argue that this is faster. But then, we should also add the definitions to circumcent RDTIM, CLRCHN, CKOUT, CHKIN, SETLFS, SETNAM, and so on...

ops:
Are you saying that there is never need to set STATUS value?

polluks:
See https://github.com/cc65/cc65/blob/2646c06f61b46ed142bbefc723acb8c2113a23f9/libsrc/c64/crt0.s#L66

oliverschmidt:
> Already done

@polluks: Once more ;-) I need to ask you to elaborate. Do you want to express that this pull request is not desirable (like @spiro-trikaliotis does) ?

oliverschmidt:
@polluks: Now I happen to read your original comment https://github.com/cc65/cc65/pull/1352#issuecomment-735401836 referring to https://github.com/cc65/cc65/blob/2646c06f61b46ed142bbefc723acb8c2113a23f9/libsrc/c64/crt0.s#L66 so I understand your point. However, there could easily be a difference between cc65 code and user code. Therefore that fact that cc65 code writes to `STATUS` doesn't imply that user code is supposed to do that too. I as well-known CBM noob still have no clue, so I'm asking for additional comments.

polluks:
"Already done" means we simply defined STATUS as ST.

polluks:
@oliverschmidt You wrote "Turned the 'ST' status byte into a symbol resolved by the linker in order to allow to use it from code living in the 'cbm' directory."
ops didn't notice this commit message. User code should use READST anyway.

mrdudz:
I am with spiro there - we have readst, and we can not be sure for certain that ST (not STATUS) is even $90 on all targets, with all kernals. Using the respective functions is exactly why they exist.

I also wonder what kind of code it is that wants to set ST - it seems this is a candidate for library code

polluks:
@mrdudz nearly
https://github.com/cc65/cc65/blob/2646c06f61b46ed142bbefc723acb8c2113a23f9/libsrc/pet/status.s#L5

mrdudz:
What is that supposed to tell me? This is library code.

Whatever, apparently those .inc files have all kind of things defined in them that fall into the same category, in that case i dont see a reason for why ST shouldnt be there.

oliverschmidt:
Okay, my take: There seems to be no mystery about that ZP location. So if someone thinks he wants to access it directly there's nothing hindering him from doing so anyway. Therefore we can as well make the symbol available as a simple matter of convenience. Strong opposite opinions?

Issue #479, opened 2017-09-02
ca65: can't reference .proc members before declaration


inexorabletash:
Example:

```
  lda foo::bar ; fails
  lda foo_bar ; workaround

.proc foo
bar .byte 0
.endproc
foo_bar := foo::bar ; for workaround
```

The workaround is to define e.g. `foo_bar := foo::bar`.

inexorabletash:
In the spirit of https://sourceforge.net/p/cc65/mailman/message/37044269/ inspired by @vrubleg I'll offer a USD50 bounty on this issue.

dmsc:
You can try the pull request, should work if you replace your example with:

 ```
   lda ::foo::bar ; fails
 
 .proc foo
 bar: .byte 0
 .endproc
 ```

Have Fun!

inexorabletash:
In the hope of attracting attention - the linked PR has at least one blocking issue around scope resolution; maybe someone can pick that up?

I'll double the bounty to USD100, as this continues to be a burden to work around.

dmsc:
Hi @inexorabletash !

I can try to look at this, but you will need to post a few test assembly files first, as in the last PR there was not clear what is the expected behavior.

Have Fun!

inexorabletash:
Thanks! I think the only blocker was the issue mentioned here:

https://github.com/cc65/cc65/pull/1358#issuecomment-747663702

Which has a repo in 

https://github.com/cc65/cc65/pull/1358#issuecomment-747194906

... but I can try to write up additional test cases.

dmsc:
Hi!

> Thanks! I think the only blocker was the issue mentioned here:
> 
> [#1358 (comment)](https://github.com/cc65/cc65/pull/1358#issuecomment-747663702)
> 

That should be fixed now in the PR. Please, test it.

Have Fun!

inexorabletash:
@dmsc - please email me (same username @ gmail) so I can send you the bounty!

PR #1358, opened 2020-12-13
Fixes Issue 479

dmsc:
This fixes issue #479 , it probably needs more testing in corner cases, but at lease the example in the documentation now works.
Note that to actually reference a scope not yet defined you *must* begin the name with "::".

### DIFF
--- a/doc/ca65.sgml
+++ b/doc/ca65.sgml
@@ -1143,6 +1143,11 @@ if you want to access the "other" symbol <tt/bar/, you would have to write:
         .endscope
 </verb></tscreen>
 
+The above example also shows that to search a scope that is not yet defined in
+the code above the usage, you must use the namespace token (<tt/::/) and
+specify the full scope name, to allow the assembler to create the scope at the
+correct place.
+
 
 <sect>Address sizes and memory models<label id="address-sizes"><p>
 
--- a/src/ca65/symbol.c
+++ b/src/ca65/symbol.c
@@ -136,10 +136,10 @@ SymTable* ParseScopedIdent (StrBuf* Name, StrBuf* FullName)
         SB_Append (FullName, Name);
 
         /* Search for the child scope */
-        Scope = SymFindScope (Scope, Name, SYM_FIND_EXISTING);
+        Scope = SymFindScope (Scope, Name, SYM_ALLOC_NEW);
         if (Scope == 0) {
             /* Scope not found */
-            Error ("No such scope: '%m%p'", FullName);
+            Error ("Can't create scope: '%m%p'", FullName);
             return 0;
         }
 


inexorabletash:
Note that with this PR the fully qualified name is needed, not just a `::` prefix. e.g. if the example is placed in a `.proc p1` scope, then `::p1::foo::bar` would be needed.

inexorabletash:
... which is particularly pesky if the source doesn't know it's full scope, e.g. if the source is being .included or otherwise instantiated in multiple places.

inexorabletash:
If there were a "current scope" operator/token that would help. Otherwise... this unfortunately turns into a bit of a refactoring hazard.

greg-king5:
> Note that, with this PR, the fully qualified name is needed, not just a `::` prefix. E.g., if the example is placed in a `.proc p1` scope, then `::p1::foo::bar` would be needed.
> 
> (This is indeed called out in the documentation change in the PR, to be clear.)

In that case, `lda p1::foo::bar` would work -- you don't need to reach out to the global level.
The requirement is that the first name in the scope path must be an existing ancestor.   (The global scope always exists.)

Your other issue would profit from a token that means "parent scope" (similar to relative directory paths).

oliverschmidt:
@greg-king5: If you consider this PR as desirable, please go ahead and merge it.

inexorabletash:
FYI, I'm seeing one bit of weirdness where a change from using the workaround to a correctly (?) scoped reference is producing a very unexpected error:

```diff
-        stx     test_rect_params2_top
+        stx     mgtk::test_rect_params2::top
```
```
../mgtk/mgtk.s(4590): Error: Range error (26683 not in [0..255])
```
I haven't had a chance to try and minimize the repro yet, but... that doesn't look like I'm doing anything with that change that could result in such an error, does it? (The assembled output is 8E 3C 68 so is a first pass assuming an 8-bit operand with the change, somehow?)

inexorabletash:
Ugh, okay, took a while but here's the minimized repro:
```asm
.setcpu "6502"
.org $2000

.proc s1

	symbol := $1

        stx     s1::s2::symbol      ; fails with Error: Range error (8197 not in [0..255])
	stx     s2_symbol           ; works

.scope s2
symbol:    .byte   0
.endscope

s2_symbol := s2::symbol

.endproc
```

Note how `symbol` is used in two scopes.

greg-king5:
That definitely is a bug!  It seems that the namespace path is ignored when ca65 is determining which address mode to use.  Instead of looking for `s1::s2::symbol`, it looked for only `symbol`.  It found `symbol := $1`, which is a zero-page address.

(Maybe, there is some "fallback" code in ca65's symbol finder: if it can't find the scope path, then look again, but in the local scope.)

mrdudz:
Another that isn't ready to merge apparently - so draft it is

dmsc:
Hi!

> That definitely is a bug! It seems that the namespace path is ignored when ca65 is determining which address mode to use. Instead of looking for `s1::s2::symbol`, it looked for only `symbol`. It found `symbol := $1`, which is a zero-page address.
> 
> (Maybe, there is some "fallback" code in ca65's symbol finder: if it can't find the scope path, then look again, but in the local scope.)

The problem is here: https://github.com/cc65/cc65/blob/master/src/ca65/studyexpr.c#L645 :

```C
        /* Since the symbol may be a forward, and we may need a statement
        ** about the address size, check higher lexical levels for a symbol
        ** with the same name and use its address size if we find such a
        ** symbol which is defined.
        */
        AddrSize = GetSymAddrSize (Sym);
        Parent = GetSymParentScope (Sym);
        if (AddrSize == ADDR_SIZE_DEFAULT && Parent != 0) {
            SymEntry* H = SymFindAny (Parent, GetSymName (Sym));
            if (H) {
                AddrSize = GetSymAddrSize (H);
                if (AddrSize != ADDR_SIZE_DEFAULT) {
                    D->AddrSize = AddrSize;
                }
            }
        } else {
            D->AddrSize = AddrSize;
        }
```

So, the code does this on purpose, searching the symbol in the parent scopes and returning the size of that symbol.

Removing that, the assembler stops working properly, as many symbols that should be ZP are converted to absolute :(

dmsc:
Ok, pushed a new commit that fixes the example in the comment above without changing the assembly of other cases.

mrdudz:
You try rebuilding https://github.com/mist64/msbasic - this one contains some ugly constructs that broke before (see #2208 )

inexorabletash:
Okay, starting to play with this now. As @greg-king5 mentioned in https://github.com/cc65/cc65/pull/1358#issuecomment-744596530 the approach requires the author to inform the assembler where the scope should be introduced before it is used, which is useful but still awkward. Examples:

```asm
.scope main
.proc foo

    lda params__bar ; works
    lda params::bar ; fails (even with this PR)
    lda foo::params::bar ; works (with this PR)
    lda params::bar ; after previous line, this now works since assembler knows where the scope is!

.proc params
bar: .byte 0
.endproc ; params
    params__bar := params::bar ; ugly workaround we'd like to eliminate

.endproc ; foo
.endscope ; main
```

And I found myself writing stuff like this:

```asm
.scope aux
   __dummy__ := ::main::__dummy__ ; pre-declare main in global scope 

   jsr main::foo ; now this just works
.endscope ; aux

.scope main
   __dummy__ := *
.proc foo
   rts
.endproc ; foo
.endscope ; main
```

So comments above still stand that this won't necessarily lead to cleaner code. Is there a reason the new scope can't be assumed to be a child of the current scope, if it's not given a qualified name? I haven't fully thought through the consequences of that. I guess that would fail in the second example (with `aux` and `main`). We're back to "need a way to refer to the current scope without naming it"

Sorry for appearing to move the goalposts, @dmsc!

inexorabletash:
After some pondering, the above can be simplified to:

```asm
.scope main
.proc foo
    foo::params::__define_scope__ = 0 ; pre-declare params in current scope
    lda params::bar

.proc params
bar: .byte 0
.endproc ; params

.endproc ; foo
.endscope ; main
```

And:

```asm
.scope aux
   ::main::__define_scope__ = 0; pre-declare main in global scope 
   jsr main::foo ; now this just works
.endscope ; aux

.scope main
.proc foo
   rts
.endproc ; foo
.endscope ; main
```

This pattern isn't so bad, especially in the latter case where there relatively few well-known scopes.

inexorabletash:
Here's an example of using this PR in my project: https://github.com/a2stuff/a2d/pull/890 

I wrapped up the above patterns into a macro, which would be improved by some way to reference "the current scope". But even without that I would call it an improvement over the current code. 

There are still cases where the current scope's name is unknown (e.g. it's in a file included in multiple places) so the workaround is still needed.

dmsc:
Hi!

> Okay, starting to play with this now. As @greg-king5 mentioned in [#1358 (comment)](https://github.com/cc65/cc65/pull/1358#issuecomment-744596530) the approach requires the author to inform the assembler where the scope should be introduced before it is used, which is useful but still awkward. Examples:
> 
> ```assembly
> .scope main
> .proc foo
> 
>     lda params__bar ; works
>     lda params::bar ; fails (even with this PR)

The main problem is *where* the "params" scope needs to be created.

See this example:

```
  .scope  foo
    .scope bar

      lda    foo::test  ; This is $200

      .scope foo
        test = $100
      .endscope

      lda    foo::test  ; This is $100

    .endscope
    test = $200
  .endscope
```

What value will the assembler use for "foo::test" depends on which scope is found first. You can't fix this without creating some symbol for "current scope", to avoid searching on parent scopes.

> Sorry for appearing to move the goalposts, @dmsc!

No problem :)

dmsc:
Pushed new commit that creates new scopes in the current scope if not found, allows this code to assemble:

```
.scope main
.proc foo

    lda params__bar ; works
    lda params::bar ; works (with this PR)
    lda foo::params::bar ; works (with this PR)

.proc params
bar: .byte 0
.endproc ; params
    params__bar := params::bar ; ugly workaround we'd like to eliminate

.endproc ; foo
.endscope ; main
```

Have Fun!

inexorabletash:
Nice! Trying it out now, so far so good. I did run into a couple places where explicit scoping was still needed, which might trip people up

As an aside: I'm trying this out by doing a comparison of the output binaries after each source change to ensure I don't introduce any differences. Without that approach I'm pretty sure I wouldn't have noticed these problems.

## Shadowing

Name shadowing still requires being explicit to identify the correct symbol. This is not new with this PR, but the prior workaround subtly avoided the problem.

```asm
;; in parent scope
.scope foo_params
bar:    .byte   123
.endscope ; foo_params

.proc DoTheThing

        ;; Oops, this would reference the parent scope!
        lda     foo_params::bar

        ;; Explicit scope necessary so local scope's params will be used
        lda     DoTheThing::foo_params::bar

;; in local scope - shadows the `foo_params` in the parent scope
.scope foo_params
bar:    .byte   456
.endscope ; foo_params

.endproc ; DoTheThing
```

Again, this is not new with this change and is a "me" problem for relying on shadowing.

I wonder if there would be a benefit to an optional warning if shadowing happens, or if there would be too many false-positives? This is a refactoring hazard as introducing shadowing by adding a symbol to a parent scope can inadvertently change what a symbol reference means. (Not this PR, though!)

## New Scope Inference

Here's a gotcha that _is_ new with this PR:

```asm
.scope foo
start:
        jmp bar::start ; where does this go?
.endscope ; foo

.scope bar
start:
        rts
.endscope ; bar
```

Prior to this PR, this would fail to assemble with `Error: No such scope: ‘bar’`. 

With this PR, this assembles, but the `bar::start` reference is actually a `foo::start` reference!

dmsc:
Hi!

> Nice! Trying it out now, so far so good. I did run into a couple places where explicit scoping was still needed, which might trip people up
> 
> ## Shadowing
> Name shadowing still requires being explicit to identify the correct symbol. This is not new with this PR, but the prior workaround subtly avoided the problem.

The only way to "fix" this is to create a new token to represent "current scope", then you could do:

```assembly
.proc DoTheThing
  ;; This would reference the parent scope!
  lda     foo_params::bar

  ;; Symbol for "current scope" could be ":::"
  lda    :::foo_params::bar
```

I think this would be controversial, and certainly beyond the scope of this PR (pun intended)

> 
> ## New Scope Inference
> Here's a gotcha that _is_ new with this PR:
> 
> ```assembly
> .scope foo
> start:
>         jmp bar::start ; where does this go?
> .endscope ; foo
> 
> .scope bar
> start:
>         rts
> .endscope ; bar
> ```

This is fixed in the last commit of the PR, the problem was that the scope was not marked as "closed", so any new symbol was being searched in any of the parent scopes.

Have Fun!

mrdudz:
It would be really nice to have a couple test programs that show (and check of course) the intended behaviour for those border cases... and the docs explaining them.

I am hardly ever using namespaces in assembler myself for that matter, so i couldn't even tell what makes sense or what this might break :)

dmsc:
Hi!

> It would be really nice to have a couple test programs that show (and check of course) the intended behaviour for those border cases... and the docs explaining them.

Yes, I asked @inexorabletash to provide tests and samples to add to the testsuite - we will see if he can provide those.

> 
> I am hardly ever using namespaces in assembler myself for that matter, so i couldn't even tell what makes sense or what this might break :)

Well, PROC are also scopes, so this will make possible to reference labels on inside PROCs that are after the current point, this could be valuable.

Have Fun!

inexorabletash:
Yep, I'll take a stab at some test cases. Looking at the `test/asm/listing` cases now, I'll see if I can come up with a handful that cover expected behavior and the edge cases noted here.

inexorabletash:
I added some tests into a [fork of this PR](https://github.com/inexorabletash/cc65/tree/issue-479) - you should be able to cherry pick https://github.com/inexorabletash/cc65/commit/939276669f85e392a82d9de4949a30579e1b3e53 into this PR.

Most of the cases are copy/paste from the documentation https://cc65.github.io/doc/ca65.html#s7 since capturing the behavior described there seemed important.

dmsc:
Hi!

> I added some tests into a [fork of this PR](https://github.com/inexorabletash/cc65/tree/issue-479) - you should be able to cherry pick [inexorabletash@9392766](https://github.com/inexorabletash/cc65/commit/939276669f85e392a82d9de4949a30579e1b3e53) into this PR.
> 

I left a few comments here: https://github.com/inexorabletash/cc65/commit/939276669f85e392a82d9de4949a30579e1b3e53

Have Fun!

dmsc:
Hi!

I added the tests by @inexorabletash to the PR, should be fairly safe to commit now.

Have Fun!

mrdudz:
Sorry dumb question (i dont have the time to dig through it again now) - is everything now in this single PR and i can ignore  the others that were linked?

Perhaps @kugelfuhr could skim over it and comment? Or someone else? :) As said before, i don't even use namepsaces myself, so it is very hard for me to tell what is good/intended/wanted behaviour and what isn't :)

inexorabletash:
It should all be in this single PR (changes + tests).

As the requester I'll comment: I think this is an improvement, but due to the single-pass nature of the compiler it has necessary compromises. New scopes need to be assumed to go somewhere, and if this assumption matches the use (a sub-scope of the current scope) then it works cleanly. For the bulk of how I use scopes, this is correct and significantly reduces the need for workarounds.

mrdudz:
Since there were no complaints for 2 weeks.... merging this. Let's see what it breaks :)

PR #1368, opened 2021-01-01
Add "return-type" warning suppression type.

thecodeme:
Now, it is possible to write:
```c
#pragma warn (return-type, off)
BYTE get_key() {
  asm("LDA $E425");
  asm("PHA");
  asm("LDA $E424");
  asm("PHA");
  asm("RTS");
}
#pragma warn (return-type, on)
```
that follows:
https://atariwiki.org/wiki/Wiki.jsp?page=How%20to%20read%20a%20Key%20from%20Keyboard%20with%20ATARI%20ROM%20Routines
```asm
...
JSR GETKEY
...

00010 ------------------------------
00020 *  READ CHAR FROM KEYBOARD   *
00030 *  CHAR WILL BE IN THE       *
00040 *  <A> REGISTER              *
00050 *  <X>,<Y> will be destroyed *
00060 ------------------------------
00070 *
00080 GETKEY   LDA $E425   GET ADDRESS-VECTOR
00090          PHA         FROM ROM HANDLER TABLE
00100          LDA $E424   PLACE ADDRESS AS
00110          PHA         RETURN ADDRESS TO STACK
00120          RTS         JUMP TO ROM-ROUTINE
00130 *
```
Options for cc65:
a) cc65 -Wreturn-type
b) cc65 -W-return-type
c) cc65 --list-warnings

### DIFF
--- a/doc/cc65.sgml
+++ b/doc/cc65.sgml
@@ -731,10 +731,7 @@ This cc65 version has some extensions to the ISO C standard.
 <item>  There is pseudo variable named <tt/__A__/ that indicates accumulator register. 
         It can be used to highlight the returned variable in non-void
         function where result is hidden in the <tt/asm/ section.
-        every other variable. They are most useful together with short
-        sequences of assembler code. 
-
-    For example, in the function below
+        For example, in the function below
 
         <tscreen><verb>
         BYTE get_key_via_atarixl_romcall_device_k(void)      \
--- a/doc/cc65.sgml
+++ b/doc/cc65.sgml
@@ -739,7 +739,7 @@ This cc65 version has some extensions to the ISO C standard.
             asm("PHA");         \
             asm("LDA $E424");   \
             asm("PHA");         \
-            asm("RTA");         \
+            asm("RTS");         \
             return __A__;
         </verb></tscreen>
 
--- a/doc/cc65.sgml
+++ b/doc/cc65.sgml
@@ -549,6 +549,8 @@ Here is a description of all the command line options:
   <tag><tt/remap-zero/</tag>
         Warn about a <tt/<ref id="pragma-charmap" name="#pragma charmap()">/
         that changes a character's code number from/to 0x00.
+  <tag><tt/return-type/</tag>
+        Warn about no return statement in function returning non-void..
   <tag><tt/struct-param/</tag>
         Warn when passing structs by value.
   <tag><tt/unknown-pragma/</tag>
@@ -724,6 +726,29 @@ This cc65 version has some extensions to the ISO C standard.
         </verb></tscreen>
 
         will give the high byte of any unsigned value.
+        <p>
+
+<item>  There is pseudo variable named <tt/__A__/ that indicates accumulator register. 
+        It can be used to highlight the returned variable in non-void
+        function where result is hidden in the <tt/asm/ section.
+        every other variable. They are most useful together with short
+        sequences of assembler code. 
+
+    For example, in the function below
+
+        <tscreen><verb>
+        BYTE get_key_via_atarixl_romcall_device_k(void)      \
+            asm("LDA $E425");   \
+            asm("PHA");         \
+            asm("LDA $E424");   \
+            asm("PHA");         \
+            asm("RTA");         \
+            return __A__;
+        </verb></tscreen>
+
+        <tt/return __A__;/ will indicate that result is stored in the accumulator
+        and suppress warning that function has no return statement. Instead, you can
+        also use <tt/#pragma warn (return-type, off) ... pragma warn (return-type, on)/
         <p>
 
 <item>  Inside a function, the identifier <tt/__func__/ gives the name of the
--- a/doc/cc65.sgml
+++ b/doc/cc65.sgml
@@ -709,7 +709,7 @@ This cc65 version has some extensions to the ISO C standard.
         places.
         <p>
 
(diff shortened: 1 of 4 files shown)


greg-king5:
We already have a way to do it:
```c
typedef unsigned char BYTE;

BYTE get_key(void) {
  asm("LDA $E425");
  asm("PHA");
  asm("LDA $E424");
  asm("PHA");
  return __A__;
}
```
(Hmm, that pseudo-variable isn't documented in https://cc65.github.io/doc/cc65.html#s5.)

thecodeme:
Good to know :)  Still "return-type" is a "standard" gcc solution so can be used without documenting. After all it should be also documented.

thecodeme:
I updated comment.  It was lack of **asm("RTS");**

The get_key() function will not directly return any argument.  It will be returned by system routine.
I've checked **return __A__;**.  Also works.

oliverschmidt:
I personally think that the inline assembler is heavily over-used. From my personal POV there's just no point in a function only consisting of `asm()` statements.

From that point of view nobody will be surprised that I personally don't consider a function that "returns into" another routine (which is already sort of obscure) written in inline assembler as a relevant use case.

Anyhow, the enhancement suggested here is simple and seems rather orthogonal. Therefore I'm not against merging it.

> [...] After all it should be also documented.

Indeed! If this pull request is extended to insert a description into https://github.com/cc65/cc65/blob/master/doc/cc65.sgml#L542-L562 (showing up at https://cc65.github.io/doc/cc65.html#option-W) I will merge it.

oliverschmidt:
> Hmm, that pseudo-variable isn't documented in https://cc65.github.io/doc/cc65.html#s5.

@greg-king5: It would be nice if you would take care of adding it.

thecodeme:
Thank you.  I will try to extend documentation during weekend.

**> I personally think that the inline assembler is heavily over-used.**
I am in favor of modular code, however, sometimes when writing snippets it makes it easier to read.

**> From that point of view nobody will be surprised that I personally don't consider a function that "returns into" another routine (which is already sort of obscure) written in inline assembler as a relevant use case.**
That was platform 6502 common practice to make a RTS call via stack and get a result from accumulator from OS routine. I remembered it yesterday after 40 years :)  
Of course, this is not clear at first and this optimization technique needs some additional explanation. As the C language is low level, and using pragmas helps achieve the expected compilation.

oliverschmidt:
> That was platform 6502 common practice to make a RTS call via stack [...]

Yes, e.g. the Applesoft interpreter does so for actual BASIC command dispatching. Therefore I wrote "*__sort of__* obscure" ;-)

thecodeme:
That is the Way :)

           |~
           |.---.
          .'_____`. /\
          |~xxxxx~| ||
          |_  #  _| ||
     .------`-#-'-----.
    (___|\_________/|_.`.
     /  | _________ | | |
    /   |/   _|_   \| | |
   /   /X|  __|__  |/ `.|
  (  --< \\/    _\//|_ |`.
  `.    ~----.-~=====,:=======
    ~-._____/___:__(``/| |
      |    |      XX|~ | |
       \__/======| /|  `.|
       |_\|\    /|/_|    )
       |_   \__/   _| .-'
       | \ .'||`. / |(_|
    LS |  ||.'`.||  |   )
       |  `'|  |`'  |  /
       |    |  |    |\/

oliverschmidt:
:-)

thecodeme:
Documentation has been updated.

thecodeme:
"RTA?"  ?

polluks:
@thecodeme looks like a typo

thecodeme:
> "RTA?" ?

Ok.  Thanx. Completed.

doc/cc65.sgml
```diff
@@ -726,6 +728,26 @@ This cc65 version has some extensions to the ISO C standard.
...
+            asm("PHA");         \
+            asm("RTS");         \
+            return __A__;
+        </verb></tscreen>
+
+        <tt/return __A__;/ will indicate that result is stored in the accumulator
+        and suppress warning that function has no return statement. Instead, you can
+        also use <tt/#pragma warn (return-type, off) ... pragma warn (return-type, on)/
+        <p>
+
```
oliverschmidt, line 750: Thanks for the effort to introduce this section. However, the content seems to specific to me. Please remove it from this pull request. I see two options:

1. Just don't care about documenting `__A__` as this lack of documentation isn't really related to the topic at hand.
2. Simply adjust the section above to not only talk about `__AX__` and `__EAX__` but rather about `__A__`, `__AX__` and `__EAX__`. There's no need to extend the example.

doc/cc65.sgml
```diff
@@ -709,7 +709,7 @@ This cc65 version has some extensions to the ISO C standard.
         places.
         <p>
 
-<item>  There are two pseudo variables named <tt/__AX__/ and <tt/__EAX__/.
+<item>  There are two pseudo variables named <tt/__A__, __AX__/ and <tt/__EAX__/.
```
oliverschmidt, line 712: This isn't correct. However, I'll fix it myself...

thecodeme: ok. thanx :)

oliverschmidt: Done: https://github.com/cc65/cc65/commit/a861d84011df82ac5f81a8c87ae3181fee393993

PR #1376, opened 2021-01-10
Define single character strings with symbolic names.

IrgendwerA8:
Helps for use cases like requested here:
https://atariage.com/forums/topic/314486-atari-strings-in-cc65/


oliverschmidt:
I can't comment on the question if the content of this PR is in general helpful or desirable. However, I see the following issues:

1. The cc65 code base is supposed to contain no tab characters, so please remove them.
2. I don't see the point creating a definition for "just every" byte value. Maybe you can elaborate on the use case of e.g. `CS_A`.
3. For all values with definitions @ https://github.com/cc65/cc65/blob/master/include/atari.h#L50-L89 I ask you to use the same name, so e.g. for $11 replace `CS_BRCORNER` with `CS_ULCORNER`. BTW: Presuming that `BR` stands for "Bottom Right" I'm wondering if it is actually correct - but maybe I'm missing the point ?!?

IrgendwerA8:
Thanks for the inspection. Regarding the issues:
1. I (again) forgot to switch the editor configuration. Sorry. Will fix that if the other points are cleared.
2. There are two arguments for that: Completeness and somewhat (surprise) better readability in mixed use case/formatting:
`CS_A CSI_T CS_A  CSI_R CS_I` vs.
`"A" CSI_T "A"  CSI_R "I"`
3. AFAIK there are no standard names for the graphic symbols (I would be glad if someone can prove me wrong!) I selected the names, resulting from a similar discussion from the tbxl-parser: https://github.com/dmsc/tbxl-parser#extended-string-constants
I see your point and the problem with the different interpretations: The `CS_BRCORNER` is a corner, connecting from bottom to the right, but this may be used typically to form the top left corner when printing a frame. I find the `BR` is more descriptive (describes the symbol, not the usage), but maybe others could chime in?

mrdudz:
IMHO:
- defining the normal characters (CS_A) is pointless. the compiler has builtin translation for string literals.
- definetly use the same names for the strings and the characters (note cbm.h also uses those)

Marrin:
Regarding standard names: hasn't Unicode added support for ”retro” platforms special characters? If there's no other standard, maybe get inspiration from the Unicode names of those characters.

greg-king5:
We want programs such as https://github.com/cc65/cc65/blob/master/samples/hello.c to be cross-target portable.

polluks:
@Marrin Unicode names are a little bit verbose...
https://de.wikipedia.org/wiki/Unicodeblock_Rahmenzeichnung

IrgendwerA8:
> Unicode names are a little bit verbose...
> https://de.wikipedia.org/wiki/Unicodeblock_Rahmenzeichnung

Yes, but at least they also describe the symbol and not the usage.

>We want consistency between the string macro names and the character macro names.

QFT.

>We want programs such as https://github.com/cc65/cc65/blob/master/samples/hello.c to be cross-target portable.

So adapt those programs too? I know, this is a breaking change, but I don't "feel well" using a misleading name for a symbol. Especially for symbols with a branch, it would be really awkward while a `CH_ULCORNER` and `CH_LLCORNER` would "match" to a `..._RBRANCH` then...

oliverschmidt:
I don't want to be the spoilsport but it's either using the existing character names or no single character strings at all. There's no room for discussion on this aspect (at least in this repo).

IrgendwerA8:
Ok!

PR #1386, opened 2021-01-19
[cc65] #1374 fix

### LINKED ISSUE #1374: Bad code generated when optimization is enabled.
The following code correctly prints "89ab" when optimization is disabled, but "ab" when optimization is enabled with the -O flag.

Seems something goes wrong with the evaluation of the right-hand side of the assignment to z.
```c
#include <stdint.h>
#include <stdio.h>

void main(void)
{
    uint8_t x = 0x89;
    uint8_t y = 0xab;
    uint16_t z = (x << 8) | y;

    printf("%x\n", z);
}
````

acqn:
Fixed #1374.


oliverschmidt:
Thanks - great to "see" you again !

I must admit that given https://github.com/cc65/cc65/pull/1322 I didn't expect you to return to cc65. One of those cases, where it's actually pretty nice to be wrong ;-)

oliverschmidt:
Given the complexity of this change (two chars) and that it fixes the regression at hand I'm merging it right away...

PR #1388, opened 2021-01-24
[cc65]Issue 1265 Fix

### LINKED ISSUE #1265: misadjusted stack from unprototyped function call
Calling an unprototyped function results in the stack pointer being decremented by 2. 

```
#include <stdio.h>

int main (void) {
  int x;

  printf ("%p\n", &x);
  f (x);
  printf ("%p\n", &x);

  return 0;
}

int f (int x) { return 1; }
```

```
% cl65 --standard c89 -t sim6502 foo.c -o foo && sim65 foo
0xffea
0xffe8
```

@acqn any insight here?

acqn:
- [X] Made `extern` declarations visible in the whole file scope.
- [X] Fixed a potential optimization bug regarding register usage with unprototyped function calls.
- [X] Fixed #1265: An undeclared function will be treated as if declared with `int` return type and an empty parameter list when first called. The bug was that it would also be errorneously treated as a variadic function, which could result in an unbalanced stack if the function is not really variadic.

### DIFF
--- a/src/cc65/codeinfo.c
+++ b/src/cc65/codeinfo.c
@@ -493,22 +493,28 @@ fncls_t GetFuncInfo (const char* Name, unsigned int* Use, unsigned int* Chg)
                 ** registers.
                 */
                 *Use = REG_EAXY;
-            } else if (D->ParamCount > 0 &&
+            } else if ((D->ParamCount > 0 ||
+                       (D->Flags & FD_EMPTY) != 0) &&
                        (AutoCDecl ?
                         IsQualFastcall (E->Type) :
                         !IsQualCDecl (E->Type))) {
                 /* Will use registers depending on the last param. If the last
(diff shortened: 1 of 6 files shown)


jmr:
> * Made `extern` declarations visible in the whole file scope.

What's an example of something that compiles now but didn't before? Test cases would help demonstrate this in addition to preventing future bugs.

acqn:
> 
> 
> > * Made `extern` declarations visible in the whole file scope.
> 
> What's an example of something that compiles now but didn't before? Test cases would help demonstrate this in addition to preventing future bugs.

The other way around:
```c
void f(void)
{
    extern char x;
}
int x; /* now error: conflicting types */
```

jmr:
Makes sense. I didn't realize C was that sensible in this case.

oliverschmidt:
Nobody complained about anything in the last six days -> merging.

PR #1391, opened 2021-02-01
Getchar management for telestrat target

jedeoric:
Hello,

Here is a fix to add getchar() function working on telestrat target

### DIFF
--- a/libsrc/telestrat/read.s
+++ b/libsrc/telestrat/read.s
@@ -8,21 +8,33 @@
 
     .include        "zeropage.inc"
     .include        "telestrat.inc"
+    .include        "fcntl.inc"
 
 ;int read (int fd, void* buf, unsigned count);
 
 .proc   _read
-    sta     ptr1 ; count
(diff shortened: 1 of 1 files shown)


oliverschmidt:
Maybe that approach is popular with other targets - I haven't checked. Anyhow, it is not correct. If you have a program doing:

```
close(0);
f = open("myfile");
```

then `f` is defined to be `0`. This way you can redirect stdin, stdout and stderr. In that scenario your code breaks. You can check e.g. the Apple II for an approach that is correct: https://github.com/cc65/cc65/blob/master/libsrc/apple2/filedes.s

The table that maps file descriptors to the internal data structures stores the information if the descriptor in question is mapped to a device or to a file. This allows e.g. read() to do the right thing: https://github.com/cc65/cc65/blob/master/libsrc/apple2/read.s#L47

jedeoric:
Ok, i did not see that point ;) i need to have a look

jedeoric:
i remove this PR, because i need more time to fix it

PR #1392, opened 2021-02-01
Add kbhit for Telestrat

jedeoric:
Hello,

Here is missing kbhit code for Telestrat

### DIFF
--- a/libsrc/telestrat/kbhit.s
+++ b/libsrc/telestrat/kbhit.s
@@ -11,9 +11,11 @@
 _kbhit:
         BRK_TELEMON XRD0
         bcs     @no_char_action
+        ldx     #$00 
         lda     #$01
         rts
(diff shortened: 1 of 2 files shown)


oliverschmidt:
Every function intended to be called by C needs to set the X register on return, see https://github.com/cc65/wiki/wiki/Parameter-passing-and-calling-conventions

jedeoric:
@oliverschmidt Fixed

polluks:
What about `jmp return1` and `jmp return0` to save some bytes?

groessler:
Better, put the "ldx #0" before the "bcs" and in "@no_char_action" do just a "txa; rts"

jedeoric:
Fixed here : https://github.com/cc65/cc65/pull/1393

PR #1393, opened 2021-02-02
kbhit optimized for telestrat

jedeoric:
As we discuss, here is kbhit optimized


oliverschmidt:
Now that your original code is merged this pull request can't be applied as-is. Please update your repo and add a commit to this pull request.

libsrc/telestrat/kbhit.s
```diff
@@ -10,12 +10,10 @@
 
 _kbhit:
         BRK_TELEMON XRD0
-        bcs     @no_char_action
         ldx     #$00 
+        bcs     @no_char_action
         lda     #$01
         rts
 @no_char_action:
```
oliverschmidt, line 17: Does A have a defined value at this point from BRK_TELEMON XRD0? If yes, it might be worth mentioning in a comment.

libsrc/telestrat/kbhit.s
```diff
@@ -10,12 +10,10 @@
...
-        lda     #$00
         tax
```
groessler, line 18: This needs to be "txa"

libsrc/telestrat/kbhit.s
jedeoric: It uses carry to set return value, but i have to look to source code to know what A returns. I try to have a look tonight

libsrc/telestrat/kbhit.s
jedeoric: Yes good point. I will fix it tonight. Thanks for the point :)

oliverschmidt: > This needs to be "txa"

I don't agree.

libsrc/telestrat/kbhit.s
oliverschmidt: Sorry, my comment above THAT useful. What I actually wanted to express and what I probably should have written is: "I consider your recent proposal broken. The only way it could potentially work is if BRK_TELEMON XRD0 would be well known to return a certain value in A that you only have to override if necessary. But I rather presume that this isn't the case." You see, I'm not actually interested in the fact what BRK_TELEMON XRD0 returns in A. There's no need that you try to find out. Rather just go ahead and write code here which works independently from BRK_TELEMON XRD0' setting of A.

oliverschmidt:
Trying to bring this back on track...

A simple, straight forward option to write this - as suggest in https://github.com/cc65/cc65/pull/1392#issuecomment-771753408 - is:

```
.import return0, return1

_kbhit:	_kbhit:
        BRK_TELEMON XRD0
        bcs     @no_char_action	
        jmp     return1
@no_char_action:
        jmp     return0
```

A "trying to be cool" option to write this is:

```
_kbhit:	_kbhit:
        BRK_TELEMON XRD0
        ldx     #$00
        txa
        asl
        eor     #$01
        rts
```

libsrc/telestrat/kbhit.s
groessler: why not?

libsrc/telestrat/kbhit.s
groessler: Sorry, Oliver, I cannot follow you.

libsrc/telestrat/kbhit.s
oliverschmidt: Just to make sure we're talking about the same code. This ias what you want to see? `tax` replaced with `txa`, right?

```
_kbhit:
        BRK_TELEMON XRD0
        ldx     #$00 
        bcs     @no_char_action
        lda     #$01
        rts
@no_char_action:
        lda     #$00
        txa        
        rts
```

My take is that the `tax` shouldn't be replaced with `txa`. Rather it should be removed altogether.

groessler: I was thinking about something like:

```
_kbhit:
        BRK_TELEMON XRD0
        ldx     #$00 
        bcs     @no_char_action
        lda     #$01  ; return one
        rts
@no_char_action:
        txa        ; return zero
        rts
```

oliverschmidt: But that's not what you wrote above ;-)

groessler:
That's the best version IMO. Smaller and faster than the others. But, `asl` should be `rol` :-)

oliverschmidt:
> But, asl should be rol :-)

How embarrassing :-(

I just directly pushed the fixed version: https://github.com/cc65/cc65/commit/855143123347fad6a09d08f995d7a2d5c7bd84de

jedeoric:
Thanks. For information, XRD0 returns in A ascii code of the key pressed

greg-king5:
> For information, XRD0 returns in A ascii code of the key pressed.

In that case, both `kbhit()` and `cgetc()` must be redesigned.  They must work together.

**`kbhit():`**
It sees that a key was tapped because it got a character!  `kbhit()` must not lose that character.  It must save that character somewhere.  If `kbhit()` is called again before `cgetc()` is called, then it must return "true" immediately because of that saved character -- it mustn't ask Telemon for a new character.

**`cgetc():`**
It must behave in a similar way.  If a saved character exists, then the function must return _that_ character -- it mustn't ask Telemon for a new character (and, it can skip the cursor code).  The function must "remove" the old character, in some way, so that the two functions will call Telemon the next time that they are called.

PR #1404, opened 2021-02-16
ar65: fix parallel builds by using a per-process temp file

groessler:
See also #1080.

Works for me, but not tested on Windows.

### DIFF
--- a/src/ar65/library.c
+++ b/src/ar65/library.c
@@ -36,8 +36,12 @@
 #include <stdio.h>
 #include <string.h>
 #include <errno.h>
+#if defined(_WIN32)
+#include <process.h>
+#else
 #include <sys/types.h>
 #include <unistd.h>
+#endif
 
 /* common */
 #include "cmdline.h"
--- a/src/ar65/library.c
+++ b/src/ar65/library.c
@@ -249,10 +249,8 @@ void LibOpen (const char* Name, int MustExist, int NeedTemp)
     if (NeedTemp) {
 
         /* Create the temporary library name */
-        NewLibName = xmalloc (strlen (Name) + strlen (".temp-") + 8 + 1);
-        strcpy (NewLibName, Name);
-        strcat (NewLibName, ".temp-");
-        sprintf (NewLibName + strlen (NewLibName), "%X", (unsigned int)getpid());
+        NewLibName = xmalloc (strlen (Name) + (sizeof (".temp-") - 1)  + 2 * sizeof (unsigned int) + 1);
+        sprintf (NewLibName, "%s.temp-%X", Name, (unsigned int)getpid ());
 
         /* Create the temporary library */
         NewLib = fopen (NewLibName, "w+b");
--- a/src/ar65/library.c
+++ b/src/ar65/library.c
@@ -36,6 +36,8 @@
 #include <stdio.h>
 #include <string.h>
 #include <errno.h>
+#include <sys/types.h>
+#include <unistd.h>
 
 /* common */
 #include "cmdline.h"
--- a/src/ar65/library.c
+++ b/src/ar65/library.c
@@ -249,9 +249,10 @@ void LibOpen (const char* Name, int MustExist, int NeedTemp)
     if (NeedTemp) {
 
         /* Create the temporary library name */
-        NewLibName = xmalloc (strlen (Name) + strlen (".temp") + 1);
+        NewLibName = xmalloc (strlen (Name) + strlen (".temp-") + 8 + 1);
         strcpy (NewLibName, Name);
-        strcat (NewLibName, ".temp");
+        strcat (NewLibName, ".temp-");
+        sprintf (NewLibName + strlen (NewLibName), "%X", (unsigned int)getpid());
 
         /* Create the temporary library */
         NewLib = fopen (NewLibName, "w+b");
--- a/src/ar65/library.c
+++ b/src/ar65/library.c
@@ -37,10 +37,10 @@
 #include <string.h>
 #include <errno.h>
 #if defined(_WIN32)
-#include <process.h>
+#  include <process.h>
 #else
-#include <sys/types.h>
-#include <unistd.h>
+#  include <sys/types.h>
+#  include <unistd.h>
 #endif
 
 /* common */


spiro-trikaliotis:
This might be an idea to do, and it makes the tool more robust, but it is not the root of the problem. It just hides the underlying root cause.

See my analysis on the mailing list.

In short:
The target geos-apple has a dependency on apple2enh.lib. It calls make "by hand" to handle it, which results in two makes of the apple2enh target in parallel, which fails.

The right solution to fix this problem is to fix the geos-apple target.

src/ar65/library.c
```diff
@@ -249,9 +249,10 @@ void LibOpen (const char* Name, int MustExist, int NeedTemp)
     if (NeedTemp) {
 
         /* Create the temporary library name */
-        NewLibName = xmalloc (strlen (Name) + strlen (".temp") + 1);
+        NewLibName = xmalloc (strlen (Name) + strlen (".temp-") + 8 + 1);
         strcpy (NewLibName, Name);
-        strcat (NewLibName, ".temp");
+        strcat (NewLibName, ".temp-");
+        sprintf (NewLibName + strlen (NewLibName), "%X", (unsigned int)getpid());
```
spiro-trikaliotis, line 255: Why don't you use something like mkstemp(3)?

groessler: Might be a good idea. But I was lazy and didn't want to bother about the issue that the temp file could be in another directory and file system than the output file.

groessler: > The right solution to fix this problem is to fix the geos-apple target.

Where exactly is this happening? I don't find it.

spiro-trikaliotis: See PR #1405.

spiro-trikaliotis: Unfortunately, I had to retract my PR #1405, as the fix generated an invalid apple2enh library.

groessler: So... is this PR still relevant, or should I retract is as well?

oliverschmidt: Apart from the question if we can/will fix the root cause of the issue at hand (apple2enh lib) I'm generally following the idea to add `getpid ()` to the temp file name.

However, I'm not that happy with the currently proposed code:

* When calling the heavyweight sprintf then we should make it do all the work. Something like `sprintf (NewLibName, "%s.%x.temp", Name, (unsigned int)getpid());`
* I pretty much dislike "magic constants" in the source code - like the `8` you introduced. Additionally I believe the code might produce a memory overwrite issue when `(unsigned int)` becomes 64-bit. As far as I understand replacing `8` with `2 * sizeof (unsigned int)` should take care of both aspects.

groessler: You are right. It was a "quick hack". I've updated the PR.

src/ar65/library.c
```diff
@@ -249,9 +249,8 @@ void LibOpen (const char* Name, int MustExist, int NeedTemp)
     if (NeedTemp) {
 
         /* Create the temporary library name */
-        NewLibName = xmalloc (strlen (Name) + strlen (".temp") + 1);
-        strcpy (NewLibName, Name);
-        strcat (NewLibName, ".temp");
+        NewLibName = xmalloc (strlen (Name) + (sizeof (".temp-") - 1)  + 2 * sizeof (unsigned int) + 1);
+        sprintf (NewLibName, "%s.temp-%X", Name, (unsigned int)getpid ());
```
spiro-trikaliotis, line 253: Wouldn't xsprintf() (src/common/xsprintf.*) be better, so it fails if there where a buffer overflow?

Personally, I would even prefer a helper function that first determines the needed length (snprintf(NULL, ...), then allocates a buffer for it, and then does the snprintf on that buffer, so the xmalloc would not need its own calculation. However, I think that currently, there is no such function in the code base, is it?

oliverschmidt: Chris stated, that the code base likely has other places which could benefit from better temp file names. Given that, I fully agree with you that it is desirable to have an utility function doing that snprintf(),xmalloc(),snprintf() thingy. Additionally one could check if it would work out to have a second layer of utility function which takes a path name and adds ".temp-<pid>" to an xmalloc'ed copy of that name (using the first utility layer).
I just didn't want to put the effort to do all that on Chris right here right now. But if he's willing to... ;-)

groessler:
On 2/17/21 8:23 AM, spiro-trikaliotis wrote:
>
> *@spiro-trikaliotis* commented on this pull request.
>
> ------------------------------------------------------------------------
>
> In src/ar65/library.c 
> <https://github.com/cc65/cc65/pull/1404#discussion_r577377102>:
>
> > @@ -249,9 +249,8 @@ void LibOpen (const char* Name, int MustExist, int NeedTemp)
>       if (NeedTemp) {
>   
>           /* Create the temporary library name */
> -        NewLibName = xmalloc (strlen (Name) + strlen (".temp") + 1);
> -        strcpy (NewLibName, Name);
> -        strcat (NewLibName, ".temp");
> +        NewLibName = xmalloc (strlen (Name) + (sizeof (".temp-") - 1)  + 2 * sizeof (unsigned int) + 1);
> +        sprintf (NewLibName, "%s.temp-%X", Name, (unsigned int)getpid ());
>
> Wouldn't xsprintf() (src/common/xsprintf.*) be better, so it fails if 
> there where a buffer overflow?
>

Isn't that xsprintf not more or less the same as snprintf? Maybe 
snprintf is not available on some older systems (old Windows, OS/2).


regards,
chris

groessler:
On 2/17/21 8:39 AM, Oliver Schmidt wrote:
> I just didn't want to put the effort to do all that on Chris right 
> here right now. But if he's willing to... ;-)


Hmm. Let's see. How should such a function be called?

regards,
chris

oliverschmidt:
>  Wouldn't xsprintf() (src/common/xsprintf.*) be better, so it fails if there where a buffer overflow?

Talking about src/common/xsprintf.c:

* https://github.com/cc65/cc65/blob/master/src/common/xsprintf.c#L651-L653 should occur much more towards the beginning of the file.
* https://github.com/cc65/cc65/blob/master/src/common/xsprintf.c#L673 looks like a C&P error (from https://github.com/cc65/cc65/blob/master/src/common/xsprintf.c#L658).

> Isn't that xsprintf not more or less the same as snprintf? Maybe snprintf is not available on some older systems (old Windows, OS/2).

I'd say that https://github.com/cc65/cc65/blob/master/src/common/xsprintf.c#L59-L62 answers that very question - doesn't it?

oliverschmidt:
> Hmm. Let's see.

:-)

> How should such a function be called?

Given that there's already https://github.com/cc65/cc65/blob/master/src/common/xmalloc.c#L101 I could imagine the "low level" function (doing the (x)snprintf(), xmalloc(), (x)snprintf()) in the same file called `xmallocprintf()`. A quick peek didn't clarify if https://github.com/cc65/cc65/blob/master/src/common/xsprintf.c#L634 supports xsnprintf(NULL, ...) - not sure how to procees best, if it doesn't...

Given that there's already https://github.com/cc65/cc65/blob/master/src/common/fname.c#L97 I could imagine the "high level" function for temp file names in the same file called `MakeTempFilename()`.

Just my two cents, Oliver

oliverschmidt:
@groessler: As far as I understand your PR requires quite some rebasing.

polluks:
Welcome in git hell ;-)

2021-03-10 22:27 GMT+01:00, Oliver Schmidt ***@***.***>:

groessler:
Welcome to git heaven :-)

groessler:
Naturally, `unistd.h` is not available on Windows. On U**x it provides the `getpid()` prototype. What's the preferred way to deal with this in cc65-land?

oliverschmidt:
See https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/getpid for the required header file.

Then proceed as e.g. at https://github.com/cc65/cc65/blob/master/src/common/filetime.c#L45-L51.

groessler:
Ok. Verified compilation with MSVC, too.

groessler:
I currently don't have time and energy to make this general temp file function we were talking before.

So please merge if you see fit for now.

src/ar65/library.c
```diff
@@ -36,6 +36,12 @@
 #include <stdio.h>
 #include <string.h>
 #include <errno.h>
+#if defined(_WIN32)
+#include <process.h>
+#else
+#include <sys/types.h>
+#include <unistd.h>
+#endif
```
oliverschmidt, line 44: Some nitpicking? ;-)
The source location I pointed you to uses `#  include` (note the two blanks) for "ifdef'ed" include statements.

oliverschmidt:
> Ok. Verified compilation with MSVC, too.

The TravisCI Windows build __does__ use MSVC. Nevertheless, thanks.

> I currently don't have time and energy to make this general temp file function we were talking before.

I see.

> So please merge if you see fit for now.

Will do...

groessler:
Not nitpicking, but correct!

Some very old compilers don't accept a preprocessor '#' which doesn't start at the first column!

PR #1405, opened 2021-02-16
Fix build problem for apple2enh on parallel builds

### LINKED ISSUE #1080: make -j

spiro-trikaliotis:
For parallel builds, the apple2enh library failed very often. This
fixes it. Fixes #1080.

### DIFF
--- a/libsrc/geos-apple/targetutil/Makefile.inc
+++ b/libsrc/geos-apple/targetutil/Makefile.inc
@@ -5,8 +5,7 @@ DEPS += ../libwrk/$(TARGET)/convert.d
 ../libwrk/$(TARGET)/convert.o: $(SRCDIR)/targetutil/convert.c | ../libwrk/$(TARGET)
 	$(COMPILE_recipe)
 
-../lib/apple2enh.lib:
-	@$(MAKE) --no-print-directory apple2enh
+../lib/apple2enh.lib: apple2enh
 
 ../target/$(TARGET)/util/convert.system: ../libwrk/$(TARGET)/convert.o ../lib/apple2enh.lib | ../target/$(TARGET)/util
 	$(LD65) -o $@ -C apple2enh-system.cfg $^


libsrc/geos-apple/targetutil/Makefile.inc
```diff
@@ -5,8 +5,7 @@ DEPS += ../libwrk/$(TARGET)/convert.d
 ../libwrk/$(TARGET)/convert.o: $(SRCDIR)/targetutil/convert.c | ../libwrk/$(TARGET)
 	$(COMPILE_recipe)
 
-../lib/apple2enh.lib:
-	@$(MAKE) --no-print-directory apple2enh
+../lib/apple2enh.lib: apple2enh
```
oliverschmidt, line 8: I'm afraid that this isn't a fix :-(

See https://github.com/cc65/cc65/blob/master/libsrc/Makefile#L66-L73

I guess your approach builds an apple2enh.lib with some GEOS files...

spiro-trikaliotis: *Argh*
I see.

I was too enthusiastic, then.

As I do not know anything about the Apple II, I could not even test it.

As I said, I do not understand your build system enough.

I am sorry. So I am retracting my PR.

spiro-trikaliotis:
This PR is invalid, as the generated output is wrong.

libsrc/geos-apple/targetutil/Makefile.inc
oliverschmidt: Apart from that it doesn't seem to work for me in the first place:

```
make geos-apple
../bin/ld65 -o ../target/geos-apple/util/convert.system -C apple2enh-system.cfg ../libwrk/geos-apple/convert.o ../lib/apple2enh.lib
ld65: Error: Input file '../lib/apple2enh.lib' not found
```

spiro-trikaliotis: That's interesting. I just re-tried, the compilation did not fail. It succeeded (but, most probably, the output is wrong, as you stated).

oliverschmidt: > I am sorry.

Don't! Thanks for
* The successful problem analysis
* The effort spent to provide a fix

greg-king5:
Spiro, this is what you tried to do:
```diff
diff --git i/libsrc/Makefile w/libsrc/Makefile
index d9ddb3ccd..431b34ba6 100644
--- i/libsrc/Makefile
+++ w/libsrc/Makefile
@@ -114,6 +114,8 @@ endef # ZIP_recipe
 zip:
 	$(foreach dir,$(OUTPUTDIRS),$(ZIP_recipe))
 
+geos-apple: apple2enh
+
 $(TARGETS):
 	@$(MAKE) --no-print-directory $@
 
```
If both `geos-apple` and `apple2enh` are "made", then `apple2enh` is built first.  Therefore, the rule in "Makefile.inc" isn't used.
If only `geos-apple` is made, then the rule in "Makefile.inc" is used.

Either way, "apple2enh.lib" is built only once!

spiro-trikaliotis:
@greg-king5, I thought I tried exactly that at first, but it failed. I do not remember exactly why it failed.
That's why I tried the approach in the PR, where I missed that the settings are wrong.

Or, possibly, I made another mistake?

oliverschmidt:
Just to make sure we're on the same page: By now the question isn't actually relevant anymore as the connection between geos-apple and apple2enh was removed by me.

PR #1406, opened 2021-02-21
[cc65] Fixed deferred post-inc and post-dec in unevaluated context

acqn:
This fixes the bug:
```c
#include <stdio.h>

int main(void)
{
    char i = 0;
    sizeof(i++);
    0 && i++;
    1 || i++;
    printf("%d\n", i);
    return i;
}
```
Expected output: `0`
Revision 131f96eb1eda192623b1b3e323a2a0739645b641 output: `3`

### DIFF
--- a/src/cc65/expr.c
+++ b/src/cc65/expr.c
@@ -376,6 +376,9 @@ void DoneDeferredOps (void)
 static void DeferInc (const ExprDesc* Expr)
 /* Defer the post-inc and put it in a queue */
 {
+    if (ED_IsUneval (Expr)) {
+        return;
+    }
     DeferredOp* Op = xmalloc (sizeof (DeferredOp));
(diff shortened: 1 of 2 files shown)


mrdudz:
Why not add this test program as an actual test program? :)

oliverschmidt:
Do I understand it right that the bug fixed here wasn't recorded as an issue so far?

mrdudz:
> Do I understand it right that the bug fixed here wasn't recorded as an issue so far?

sounds like it :)

I'd add the test - but in this case it will only create extra work IMHO, easier to add it together with the fix :)

acqn:
Testcase added.

oliverschmidt:
My gut feeling is that this PR doesn't require more time to wait for potential feedback. Please let me know if you consider my assessment is inappropriate. However, for now: Merging...

PR #1409, opened 2021-02-22
Fix crashes when .BANK() or ld65/dbgsyms.c look for a nonexistent export.

bbbradsmith:
I had the following code, which caused a null reference crash when the symbols in question were not available for import.
```
.import checkpoint_init
.import checkpoint_tick
.assert .bank(checkpoint_init)=.bank(checkpoint_tick),error,"init/tick in different banks for: checkpoint"
```

So in GetSegExprValInternal there is a case EXPR_SYMBOL which checks for a circular reference, and then if it's not, marks the symbol and recurses into GetSegExprValInternal. It seems at this point, the GetExprExport used before the recursion returned an Export with a NULL Expr member for some reason. (Crash was on "switch (Expr->Op)" with a NULL Expr.)

I haven't dug into why GetExprExport can return an Export with a NULL Expr member, or whether this is a real problem, but it seems sufficient to just reject this case as "TooComplex" here. The resulting error is very understandable and points to the specific line causing it, so it seems safe to do this?

```
ld65.exe: Warning: Cannot evaluate assertion in module '...', line 62
Unresolved external 'ceiling_init' referenced in:
  ...(62)
```

### DIFF
--- a/src/ld65/expr.c
+++ b/src/ld65/expr.c
@@ -466,7 +466,6 @@ static void GetSegExprValInternal (ExprNode* Expr, SegExprDesc* D, int Sign)
 
     if (Expr == 0)
     {
-        D->TooComplex = 1;
         return;
     }
 
--- a/src/ld65/expr.c
+++ b/src/ld65/expr.c
@@ -464,6 +464,12 @@ static void GetSegExprValInternal (ExprNode* Expr, SegExprDesc* D, int Sign)
 {
     Export* E;
 
+    if (Expr == 0)
+    {
+        D->TooComplex = 1;
+        return;
+    }
+
     switch (Expr->Op) {
 
         case EXPR_LITERAL:
--- a/src/ld65/expr.c
+++ b/src/ld65/expr.c
@@ -464,8 +464,7 @@ static void GetSegExprValInternal (ExprNode* Expr, SegExprDesc* D, int Sign)
 {
     Export* E;
 
-    if (Expr == 0)
-    {
+    if (Expr == 0) {
         return;
     }
 


greg-king5:
As far as I can see, it's sufficient just to return from the function (without evaluating `Expr->Op`).

`.bank()` wouldn't get a chance to look at `D.TooComplex`.  And, `PrintDbgSyms()` won't see a segment from a missing export.

`TooComplex` doesn't need to be set.

bbbradsmith:
Okay, removed the TooComplex and made it just return instead. Can confirm that still solves the problem case I had.

oliverschmidt:
@greg-king5: Thanks for the analysis - please merge as you see fit.

greg-king5:
I thought of a slightly simpler, slightly faster fix:  don't call the function recursively, in the first place, if it would crash.

In the `if` statement that looks for circular references, change the `else` to
```c
} else if (E->Expr != 0) {
```

bbbradsmith:
Do you think there might be other undiscovered cases where it's called with a null Expr, or is that particular recursion the only possible case where it can come up?

greg-king5:
> Do you think there might be other undiscovered cases where it's called with a null Expr, or is that particular recursion the only possible case where it can come up?

It's the only case that I know of where it's expected ("missing export" means "no definition for the name" -> "no expression" -> "NULL pointer").  If there actually is another case, then we need to discover it, and investigate it.  Therefore, I want the function to crash (instead of hiding it).

mrdudz:
> Therefore, I want the function to crash (instead of hiding it).

perhaps an assert would be more sensible?

oliverschmidt:
@greg-king5: How you do assess the state of this PR?

oliverschmidt:
@greg-king5: Thanks for adding the final touch-up.

mrdudz:
what about a testcase? :)

bbbradsmith:
Here is an extremely minimal test case:
[test_expr.zip](https://github.com/cc65/cc65/files/6601821/test_expr.zip)

The correct behaviour is that it should generate an error message and errorlevel 1, instead of silently crashing and generating an invalid errorlevel.

I don't know how to make a cc65 unit test that verifies something like this, though.

PR #1410, opened 2021-02-23
[cc65] Fixed signed char type comparisons with unsigned numeric constants

### LINKED ISSUE #1408: [cc65] Signed char type comparisons with unsigned numeric constants are wrong
```c
#include <stdio.h>

static unsigned int failures = 0;
static signed char x = -1;

int main(void)
{
    if (!(x > -2u)) {
        printf("x > -2u should be true\n");
        ++failures;
    }
    if (!(x > 0u)) {
        printf("x > 0u should be true\n");
        ++failures;
    }
    if (!(x > 255u)) {
        printf("x > 255u should be true\n");
        ++failures;
    }

    return failures;
}
```
Current output:
```
x > -2u should be true
x > 0u should be true
x > 255u should be true
Failures: 3
```

A fix depending on PR #1322 is done.

acqn:
Fixed #1408.

### DIFF
--- a/src/cc65/expr.c
+++ b/src/cc65/expr.c
@@ -2845,7 +2845,7 @@ static void hie_compare (const GenDesc* Ops,    /* List of generators */
             }
 
             /* Determine the type of the operation. */
-            if (IsTypeChar (Expr->Type) && rconst && RightSigned) {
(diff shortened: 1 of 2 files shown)


greg-king5:
:+1:

PR #1411, opened 2021-02-23
[cc65] Some comparison fixes

acqn:
- [X] Fixed expression type of the result of numeric constant comparison.
- [X] Fixed result type in certain contant expression addition cases.


oliverschmidt:
I don't expect any feedback by now, merging...

PR #1413, opened 2021-03-01
[Telestrat] chdir is now working

jedeoric:
add chdir management

### DIFF
--- a/libsrc/telestrat/syschdir.s
+++ b/libsrc/telestrat/syschdir.s
@@ -1,0 +1,34 @@
+;
+; Jede (jede@oric.org), 2021-02-22
+;
+; unsigned char _syschdir (const char* name, ...);
+;
+
+        .export         __syschdir
+        .import         addysp, popax
+		.importzp       tmp1
+
+        .include        "telestrat.inc"
+        .include        "zeropage.inc"
+        
+
+__syschdir:
+        ; Throw away all parameters except the name
+        dey
+        dey
+        jsr     addysp
+
+        ; Get name
+        jsr     popax
+		
+		stx     tmp1
+        ldy     tmp1
+        
+        ; Call telemon primitive
+        
+        BRK_TELEMON(XPUTCWD)
+        
+        rts
+
+
+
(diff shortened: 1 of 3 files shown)


jedeoric:
There is a problem with this PR because i did not update initcwd. I try to fix it (i wanted to do a separate PR, but it's not clear if i split into 2 PR)

polluks:
@oliverschmidt How about some more .gitattributes magic?
https://stackoverflow.com/questions/13819764/indentation-configuration-only-for-some-files/57244091#57244091

jedeoric:
I forgot to say that i finished all corrections.

libsrc/telestrat/syschdir.s
```diff
@@ -0,0 +1,34 @@
+;
+; Jede (jede@oric.org), 2021-02-22
+;
+; unsigned char _syschdir (const char* name, ...);
+;
+
+        .export         __syschdir
+        .import         addysp, popax
+	    .importzp       tmp1
```
oliverschmidt, line 9: Undesired tab char.

libsrc/telestrat/syschdir.s
```diff
@@ -0,0 +1,34 @@
...
+        stx     tmp1
+        ldy     tmp1
+        
+        ; Call telemon primitive
+        
+        BRK_TELEMON(XPUTCWD)
+
+        jsr     initcwd      ; Update cwd
+        
+        rts
```
oliverschmidt, line 34: Tail call optimization opportunity: `jsr <xxx> rts` -> `jmp <xxx>`.

libsrc/telestrat/syschdir.s
jedeoric: Fixed

libsrc/telestrat/syschdir.s
jedeoric: Fixed

oliverschmidt:
I had to revert this pull request as it doesn't even build :-(

greg-king5:
@jedeoric,
Definitions for `XGETCWD` and `XPUTCWD` need to be added to "asminc/telestrat.inc".

jedeoric:
@greg-king5 Sorry for the mistake, because i did not pushed the file. That is why it was working here, but not in this PR.

Anyway, i think that i have to open a new PR ? I doubt that i can open again this PR.

PR #1419, opened 2021-03-10
[Telestrat] chdir management

jedeoric:
This is the PR done last week. It was missing XGETCWD and XPUTCWD


oliverschmidt:
My Git Foo isn't sufficient to explain why, but GitHub says *Rebasing the commits of this branch on top of the base branch cannot be performed automatically as this would create a different result than a regular merge.* so I guess you need to do "something".

jedeoric:
For my part, i don't see any action under github, i only see "This branch has no conflicts with the base branch" 

I don't see what i must fix

oliverschmidt:
I decided to merge your squashed commits.

PR #1423, opened 2021-03-13
[cc65] Codegen fix for certain cases of object addresses as boolean

acqn:
Fixed code generation logic with certain cases of object addresses as boolean:

```c
#include <stdio.h>

int main(void)
{
    int a;
    while (&a) {
        return 0;
    }
    printf("Failure!\n");
    return -1;
}
```

Wrong with `while`, `do while` and `for` loops. This PR fixes them all.

### DIFF
--- a/src/cc65/expr.c
+++ b/src/cc65/expr.c
@@ -3646,7 +3646,7 @@ static int hieAnd (ExprDesc* Expr, unsigned* TrueLab, int* TrueLabAllocated)
                 */
                 DoDeferred (SQP_KEEP_NONE, Expr);
 
-                if (Expr->IVal == 0 && !ED_IsAddrExpr (Expr)) {
+                if (ED_IsConstFalse (Expr)) {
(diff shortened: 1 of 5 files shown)


mrdudz:
I'd be willing to knock up a testcase if you'd tell what exactly the problem was that is fixed by this =)

acqn:
There are two other bugs with ternary:
- `a ? (void)b++ : (void)c` has the `++` generated at the wrong place.
- `a ? b, c : d` doesn't compile.

oliverschmidt:
@mrdudz: Do you want me to wait with merging this PR until you have created the (for now failing) test case. Or should I rather go ahead and merge (and you go back locally to verify that the test case actually fails without this PR) ?

mrdudz:
easier if you just wait some days...

oliverschmidt:
Will do - thanks for the immediate feedback.

PR #1425, opened 2021-03-13
[cc65] Ternary fixes

acqn:
- [X] Fixed deferred post-inc/dec in the 'true-case branch' of a ternary operation, when the branch type is `void`.
```c
#include <stdio.h>

int main(void)
{
	int x = 0;
	x ? (void)x-- : (void)0;
	printf("%d\n", x);
	return x;
}
```
Expected output: `0`
Wrong output: `-1`
- [X] `a ? b, c  : d` now works as if written as `a ? (b, c) : d`.

### DIFF
--- a/src/cc65/expr.c
+++ b/src/cc65/expr.c
@@ -3973,7 +3973,7 @@ static void hieQuest (ExprDesc* Expr)
         /* Parse second expression. Remember for later if it is a NULL pointer
         ** expression, then load it into the primary.
         */
-        ExprWithCheck (hie1, &Expr2);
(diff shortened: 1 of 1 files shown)


oliverschmidt:
@mrdudz: As far as I understand it should be pretty straight forward to create a test case from the code given in the PR description. I'd appreciate, if you could take care of creating one.

mrdudz:
will do, as said in #1423, just give me a couple days :)

oliverschmidt:
I wasn't sure if you are aware if this PR... Thanks for the immediate feedback!

PR #1432, opened 2021-03-19
[cc65] Fixed and cleaned up codegen logic with arithmetic conversion in addition and subtraction

### LINKED ISSUE #1431: Internal Error / Stackmisalignment
```
/*
$ cl65 -Osir --codesize 180 -S -o main.s main.c
main.c(9): Internal compiler error:
Code generation messed up: StackPtr is -2, should be -4

Input: if (wcnt > btw) {

$ git bisect bad
aa6fdf58b8a17b747090fb521f3d9106e0c56d1c is the first bad commit
commit aa6fdf58b8a17b747090fb521f3d9106e0c56d1c
Author: acqn <acqn163@outlook.com>
Date:   Mon Feb 8 09:03:19 2021 +0800

    Addresses in constant subtraction expressions now work.
    Fixed codegen for cast type subtraction in constant expressions.
*/

unsigned long fptr;

int main(void)
{
    unsigned int btw;
    unsigned int wcnt;

    wcnt = 512U - (fptr % 512U);
    if (wcnt > btw) {
         wcnt = btw;
    }
}
```

aa6fdf58 breaks this ... @acqn please check :)

acqn:
Fixed #1431.

### DIFF
--- a/src/cc65/expr.c
+++ b/src/cc65/expr.c
@@ -2861,17 +2861,18 @@ static void parseadd (ExprDesc* Expr, int DoArrayRef)
                 AddDone = -1;
             }
 
+            /* Do constant calculation if we can */
             if (ED_IsAbs (&Expr2) &&
                 (ED_IsAbs (Expr) || lscale == 1)) {
                 if (IsClassInt (lhst) && IsClassInt (rhst)) {
-                    typeadjust (Expr, &Expr2, 1);
+                    Expr->Type = ArithmeticConvert (Expr->Type, Expr2.Type);
                 }
                 Expr->IVal = Expr->IVal * lscale + Expr2.IVal * rscale;
                 AddDone = 1;
             } else if (ED_IsAbs (Expr) &&
                 (ED_IsAbs (&Expr2) || rscale == 1)) {
                 if (IsClassInt (lhst) && IsClassInt (rhst)) {
-                    typeadjust (&Expr2, Expr, 1);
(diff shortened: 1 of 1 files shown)


mrdudz:
Now i cant built my project anymore because the code is blown up too much >_<

mrdudz:
@acqn are you sure this is the right thing to do? this kind of code explosion seems a bit excessive :)

acqn:
> 
> 
> Now i cant built my project anymore because the code is blown up too much >_<
> 
> Edit: actually quite a lot. depending on what --codesize param i use, its ~200 or even ~800 bytes more than before for a single segment. wth is happening there?

Any code samples?

mrdudz:
It's tedious to find out what exactly the problem is (afterall it could be a single byte in an often used codepath).... and since i solved it now by shuffling some code to another bank, i am not sure if i will bother finding out :) Its fine with me that code size increases when compiler bugs are being fixed - i just was a bit surprised, usually its just a handful bytes, not hundreds :)

So... no problem. There are more important things to look at.

oliverschmidt:
> [...] since i solved it now by shuffling some code to another bank [...]

https://github.com/cc65/ip65/commit/25dc32193ccc1ba14bcf955dcf761ed37dad8037 ;-)

> Its fine with me that code size increases when compiler bugs are being fixed

Me too.

> There are more important things to look at.

Full ACK.

PR #1434, opened 2021-03-19
Test for pr #1423

mrdudz:
needs #1423 merged first


mrdudz:
Lol, no idea where the unrelated stuff in the history comes from. git is just weird :)

greg-king5:
`test2()` can't fail!  `do { } while ();` always does a loop at least once.  Therefore, the function returns before it can test the object address.

I think that `test2()` isn't needed.  `while () {}` is compiled as `do { } while ();` with a jump to the `while ()`, at the beginning.  That makes `test2()` redundant.

mrdudz:
IF its compiled correctly, sure. If not, then not

PR #1435, opened 2021-03-24
[cc65] Pointer type comparison/conversion fixes and improvements

acqn:
- [X] Fixed the prototypes of `adler32()` and `crc32()` in the bundled zlib header so that the lib may still compile with pointer type conversion fix.
- [X] Fixed ternary result type detection with pointer types.
- [X] Fixed pointer type comparison and conversion, especially regarding qualifiers.
- [X] Improved diagnostics about type comparison and conversion.
- [X] Added new -W options to turn on/off warnings on certain pointer conversion cases:
  - pointer-sign: to a pointer type differing in pointee signedness. Default on.
  - pointer-types: to a pointer type incompatible. Default on.
- [X] Reorganized some type-comparison/conversion functions.


polluks:
Are we close to a new release?

mrdudz:
Once again, i'd be happy to create test cases - if you could elaborate what exactly this fixes (does it even? or is it just improvements for the code generator?)

oliverschmidt:
> Are we close to a new release?

I have no idea.

@acqn: Do you want to comment on the structure of your past and future changes? Is there a point (or even several) where you'd say: This is a some how especially consistent/haromized state which makes sense to be considered a release?

> Once again, i'd be happy to create test cases [...]

Thanks :-)

mrdudz:
Getting a silly amount of char* vs unsigned char* pointer conversion warnings now in my project :) GOOD! (and nothing broken as far as i can see - which is obviously also good!)

mrdudz:
Ok well, perhaps it's a bit too pedantic, eg
```
unsigned char *vgamodes[VGAMODE_NUMMODES] = {
{" 800x 600 @ 72Hz"},
}
```
this gives a
```
vga-test.c(13): Warning: Incompatible pointer conversion to 'unsigned char *' from 'char *'
```
BUT - char is unsigned by default! so IMHO no point to warn here - unless --signed-char was used.

groessler:
I recollect a discussion recently over here (in cc65 ML or so), that just "char" is different to "unsigned char" and "signed char".

greg-king5:
> I recollect a discussion recently over here (in cc65 ML or so), that just "char" is different to "unsigned char" and "signed char".

Sure, there's a conversion; but, I wouldn't call it "incompatible".

mrdudz:
> I recollect a discussion recently over here (in cc65 ML or so), that just "char" is different to "unsigned char" and "signed char".

Sure - but does that difference make the pointers incompatible, justifying the warning?

mrdudz:
Having said that - since it might still be interesting to fix those cases - ideally have this particular warning optional, ie disabled by default, but still available via some switch

groessler:
> Having said that - since it might still be interesting to fix those cases - ideally have this particular warning optional, ie disabled by default, but still available via some switch

Sounds good to me.

acqn:
I do understand that there is some interest on an option to restore the "old good char type behavior"... Maybe later.

mrdudz:
> I do understand that there is some interest in an option to restore the "old good char type behavior"... Maybe later.

just to clearify - generally more warnings are a good thing to me. however for this case ("incompatible pointers") the char type is kindof special. while "char" is different to "unsigned char" in subtle ways, their pointers are still compatible IF the default "char" type is "unsigned". So the warning should *only* be optional in this case, ie when the default char type matches the char type the pointer is being converted to/from. Converting a "char*" to "signed char*" should still produce a warning IF the default "char" type is "unsigned".

The point is - i have a lot of string initializations like the one above. And even worse, a lot of functions that take "unsigned char *" and then a string literal is passed to them. I know, if we are super pedantic this can be considered "wrong" - however now i am drowning in warnings that are not useful, and its tedious to fish out the warnings that may actually indicate a problem.

That said - i'd have to say thanks for looking at this stuff and fixing the compiler - its super helpful, even when i have to rework my code every now and then :)

oliverschmidt:
> So the warning should only be optional in this case [...]

That would be an opt-in. I could imagine that an opt-out could be (more?) approriate.

So "we" understand that there's a signficant code base assuming that `unsigned char` and `char` are the same when it comes to cc65 (at least with the default settings). In order to be able to maintain that code base without "drowning in warnings that are not useful" it would help if there would be a compiler option to suppress those very warnings.

But I'm not sure if that should be the default behavior - my gut feeling is it shouldn't...

mrdudz:
> That would be an opt-in. I could imagine that an opt-out could be (more?) approriate.

I dont agree :) The warning serves no purpose other than being (very) pedantic. Most of the time you probably dont want to see it.

with gcc for example, you have to explicitly enable it, its disabled by default even with -O3

oliverschmidt:
Can you please point me to the gcc option to enable it?

mrdudz:
i only checked that its not there with -O3 (which enables "commonly useful" warnings) and that -Wall does, there surely is a specific option too :)

acqn:
> 
> 
> Once again, i'd be happy to create test cases - if you could elaborate what exactly this fixes (does it even? or is it just improvements for the code generator?)

Some random cases:
```c
char *q;
signed char *sq;
unsigned char *uq;
long * lp;
int i;
int * ip;
int ** ipp;
int const ** icpp;
int * const * ipcp;
int const * const * icpcp;
void * vp;
void const * vcp;
void ** vpp;
int (*m0)[];
int (*m1)[1];
int (*m2)[2];
void (*fp)(void);
void __fastcall__ (*gp)(void);
void __cdecl__ (*hp)(void);

int (*mpm0[1])[];
int (*mpm1[1])[1];

uq = q;
sq = q;
i = vp - vcp;
i = (unsigned *)vp - (int *)vcp;
i = mpm1 - mpm0;
i = lp < ip;
i = lp < vp;
vp = (int *)vcp + 2;
vp = (i ? fp : gp);
fp = vp;
vp = (i ? fp : hp);
i = ip - vp;
i = vp - vpp;
i = m1 - m2;
ip = (i ? ip : vcp);
icpp = (i ? icpcp : icpp);
vp = (i ? ip : vpp);
ipp = (i ? ipp : vpp);
ip = (i ? ipcp : vp);
m0 = m1 = m2;
m2 = m1 = m0;
```

oliverschmidt:
> i only checked that its not there with -O3 (which enables "commonly useful" warnings) and that -Wall does enable it, there surely is a specific option too :)

Likely I'm once more missing the point... What I try:

```
unsigned char *uc;
char *c;

int main(void)
{
  return c == uc;
}
```

Using gcc both without any options as well as with `-funsigned-char` gives me:

```
comp.c: In function 'main':
comp.c:6:12: warning: comparison of distinct pointer types lacks a cast
   return c == uc;
            ^~
```

This is with:

```
gcc --version
gcc (i686-win32-sjlj-rev1, Built by MinGW-W64 project) 6.3.0
```

mrdudz:
Well, that is not what i tested - i used assignments, not comparisons :) sth like
```
unsigned char *uc;
signed char *sc;
char *c;

int main(void)
{
  s = uc;
  s = sc;
  sc = uc;
  sc = c;
  uc = c;
  uc = sc;
  return 0;
}
```
and a slightly less antique gcc :)
```
$ gcc --version
gcc (Gentoo 9.3.0-r2 p4) 9.3.0
```

oliverschmidt:
At least the more antique gcc misses a declaration of `s` in your code, maybe the new ones add it via AI ;-))

oliverschmidt:
So gcc has `-Wpointer-sign` to turn on the warning in question. It is implied by `-Wall` and by `-pedantic`. It has `-Wno-pointer-sign` to turn the warning in question explicitly off.

The pull request comment says:
> pointer-sign: to a pointer type differing in pointee signedness. Default on.

Based on the gcc behavior I second @mrdudz in asking to change this pull request to rather be:
> pointer-sign: to a pointer type differing in pointee signedness. Default __off__.

mrdudz:
>  I second @mrdudz

note that i am NOT proposing to disable that warning by default in the general case, but *only* for the "char" vs "unsigned char" case, which is the one that is hardly ever useful. (unless --signed-char is used, then it should be for "char" vs "signed char")

acqn:
> 
> 
> > Are we close to a new release?
> 
> I have no idea.
> 
> @acqn: Do you want to comment on the structure of your past and future changes? Is there a point (or even several) where you'd say: This is a some how especially consistent/haromized state which makes sense to be considered a release?
Perhaps. I'm still fixing some remaining bugs in the C compiler....

acqn:
> 
> 
> > I second @mrdudz
> 
> note that i am NOT proposing to disable that warning by default in the general case, but _only_ for the "char" vs "unsigned char" case, which is the one that is hardly ever useful. (unless --signed-char is used, then it should be for "char" vs "signed char")
Well, that warning is pretty much all you can get from that change of the `char` type for now, if not ever....

oliverschmidt:
> I'm still fixing some remaining bugs in the C compiler....

Thanks for the feedback. I see. It would be great to let me/us know when you are done with these fixes from your POV.

oliverschmidt:
> So, what is expected to be done for this PR to be merged/closed?

Thanks for your willingness to match expectations :-)

@mrdudz wrote:
> note that i am NOT proposing to disable that warning by default in the general case, but only for the "char" vs "unsigned char" case [...]

From my perspective You resolved that misunderstanding with your comment:
> Well, that warning is pretty much all you can get from that change of the char type for now, if not ever....

So from my perspective my comment https://github.com/cc65/cc65/pull/1435#issuecomment-809606316 still/again holds true saying:
> [...] pointer-sign: to a pointer type differing in pointee signedness. Default __off__.

Does that makes sense for you?

mrdudz:
>>[...] pointer-sign: to a pointer type differing in pointee signedness. Default off.

>Does that makes sense for you?

No. All other warnings of this type are (usually) useful. So if there is no way to just disable that one specific one, i'd prefer to see them all.

oliverschmidt:
I see. Then all particpants of this discussion prefer the PR as-is -> merging.

PR #1436, opened 2021-03-25
ca65 check listing (PR for #1427)

spiro-trikaliotis:
At first: THIS IS A PREVIEW ONLY, DO NOT APPLY THIS PR!

It is not fully tested yet, and I already know some issues!


spiro-trikaliotis:
Ok, building cc65 fails because of a segmentation fault. I still have work to do.

spiro-trikaliotis:
Revoking this pull request. I think this approach is a dead end. I am currently testing another one.

PR #1424, opened 2021-03-13
[cc65] More compile-time constant expressions regarding object addresses

acqn:
This enables most sensible cases discussed in Issue #1196, which are `Compile-Time-Known`.

Notable cases that _aren't_ supported:
```c
/* Taken from #1196 OP  */
int a;
int f() { return 1; }
extern int g();

a == a;         /* Pure: Yes; Static: Yes;  Immutable: Yes; Compile-Time-Known: Yes  */
f();            /* Pure: Yes; Static: Yes?; Immutable: Yes; Compile-Time-Known: Yes? */
g() * 0;        /* Pure: No;  Static: No;   Immutable: Yes; Compile-Time-Known: Yes  */
```
Notable cases that _are_ supported (https://github.com/cc65/cc65/issues/1196#issuecomment-675497430):
```c
/* Taken from #1196 reply */
struct S {
    int a;
    int b;
} s[2];

/* The compiler can be sure of: */
&s == &s[0];
&s[0] < &s[1];
&s[0].b > &s[0].a;
&s[0].b < &s[1].a;
```

PR #1442, opened 2021-03-29
[cc65] Fixed ICE on error cases such as '&func + int a'.

acqn:
Fixed a regression introduced by PR #1424.

### DIFF
--- a/src/cc65/expr.c
+++ b/src/cc65/expr.c
@@ -2854,27 +2854,31 @@ static void parseadd (ExprDesc* Expr, int DoArrayRef)
             } else {
                 /* OOPS */
                 AddDone = -1;
+                /* Avoid further errors */
+                ED_MakeConstAbsInt (Expr, 0);
             }
 
-            /* Do constant calculation if we can */
-            if (ED_IsAbs (&Expr2) &&
-                (ED_IsAbs (Expr) || lscale == 1)) {
-                if (IsClassInt (lhst) && IsClassInt (rhst)) {
-                    Expr->Type = ArithmeticConvert (Expr->Type, Expr2.Type);
+            if (!AddDone) {
+                /* Do constant calculation if we can */
+                if (ED_IsAbs (&Expr2) &&
+                    (ED_IsAbs (Expr) || lscale == 1)) {
+                    if (IsClassInt (lhst) && IsClassInt (rhst)) {
+                        Expr->Type = ArithmeticConvert (Expr->Type, Expr2.Type);
+                    }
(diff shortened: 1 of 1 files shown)


oliverschmidt:
@mrdudz: Do you want to provide feeback on this PR from the test perspective?

oliverschmidt:
No feedback for five days -> merging...

mrdudz:
@acqn could you give a more specific hint on what was fixed here? i tried
```
#include <stdlib.h>

void f1(int n)
{
    return;
}

int a;
void *p;
void (*func)(int);

int test1(void)
{
    func = f1;
    p = &func + a;
    return 0;
}

int main(void)
{
    test1();
    return EXIT_SUCCESS;
}
```
and this compiles just fine using 1bd04e84

mrdudz:
@acqn please?

greg-king5:
In your test, `&func` is the address of a pointer variable.  His fix is for the (direct) address of a function:
```c
p = f1 + a;
p = f1 + 17;
```

mrdudz:
Wtf? Once again i wonder.... why would you do this in the first place? :=)

oliverschmidt:
You can reframe this as: "Good that it's __SUCH__ an obscure thing, that causes an ICE." ;-)

mrdudz:
ok so i rewrote the test to this:
```
#include <stdlib.h>
#include <stdio.h>

void f1(void)
{
    return;
}

int a;
void *p;

unsigned long test1(void)
{
    p = f1;
    return (unsigned long)p;
}

unsigned long test2(void)
{
    a = 42;
    p = f1 + a;
    return (unsigned long)p;
}

unsigned long test3(void)
{
    p = f1 + 17;
    return (unsigned long)p;
}

unsigned long x, y, z;
int res = 0;

int main(void)
{
    x = test1();
    y = test2();
    z = test3();
    if ((y - x) != 42) {
        res++;
    }
    if ((z - x) != 17) {
        res++;
    }
    printf("res: %d\n", res);
    return res;
}
```
to my own surprise it does actually compile with gcc (tried --std=c89 and --std=c99). Only with --pedantic gcc produces a warning (but no error).

with cc65 it produces an error (but no ICE):
```
pr1442.c:22: Error: Invalid operands for binary operator '+'
pr1442.c:28: Error: Invalid operands for binary operator '+'
```

That raises the question what the standard says to this :) do we expect an error or not? @acqn ?

mrdudz:
@acqn Could you comment on this? What do we expect according to the standard - error? Warning? I'd like to add this test at the right place :)

PR #1444, opened 2021-03-31
[cc65]  Fixed composition of a function declaration with an empty parameter list and an old-style function definition

### LINKED ISSUE #1263: Erroneous error for K & R function definitions with enum parameters
This gives an error with `--standard c89`, but should be OK:
```c
enum E { I };
extern int f(enum E);
int f(e) 
  enum E e;
{
  return 1;
}
```
```
% cc65 --standard c89 /tmp/func3.c
/tmp/func3.c(5): Error: Conflicting function types for 'f'
/tmp/func3.c(7): Warning: Parameter 'e' is never used
```
This works with gcc, and used to work with cc65.
[gcc godbolt](https://godbolt.org/z/3rsPEz)
[old cc65 godbolt](https://godbolt.org/z/85de6E)

I didn't try to bisect, but maybe one of @acqn's function handling changes broke this.

It would also be nice if the error message said what the compiler thought the conflicting types were.

The line numbers are also a bit strange, since lines 5 and 7 are the opening and closing curly braces, respectively.

acqn:
Fixed #1263.
Note: only the one in https://github.com/cc65/cc65/issues/1263#issuecomment-698869640 was a bug. The one in the [OP](https://github.com/cc65/cc65/issues/1263#issue-707625858)  was not, for the reason that the underlying type of the `enum` is `unsigned char` instead of `int` in cc65.

### DIFF
--- a/src/cc65/expr.c
+++ b/src/cc65/expr.c
@@ -651,11 +651,11 @@ void DoDeferred (unsigned Flags, ExprDesc* Expr)
 }
 
 
-static unsigned FunctionParamList (FuncDesc* Func, int IsFastcall, ExprDesc* ED)
-/* Parse a function parameter list, and pass the arguments to the called
-** function. Depending on several criteria, this may be done by just pushing
-** into each parameter separately, or creating the parameter frame once, and
-** then storing into this frame.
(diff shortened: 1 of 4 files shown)


oliverschmidt:
No feedback in three days + turning an existing test case -> merging...

PR #1452, opened 2021-04-05
[cc65] Fixed local struct field access via the address of the struct

### LINKED ISSUE #1451: [cc65] struct member access via a local address ('->') doesn't work right
```c
#include <stdio.h>

typedef struct {
    int a;
    int b;
} S;

int main(void)
{
    S a = {2, 5};
    S b = {1, 4};
    S m[1] = {{6, 3}};
    S *p = &a;
	
    (&a)->a += b.a;
    p->b += b.b;
    m->a += b.a;

    printf("%d\n", (&a)->a);
    printf("%d\n", p->b);
    printf("%d\n", m->a);

    return 0;
}
```
Expected output:
```
3
9
7
```

Wrong output:
```
2
9
2
```

acqn:
Fixed #1451.

### DIFF
--- a/src/cc65/expr.c
+++ b/src/cc65/expr.c
@@ -1340,7 +1340,6 @@ static void Primary (ExprDesc* E)
 static void StructRef (ExprDesc* Expr)
 /* Process struct/union field after . or ->. */
 {
-    ident Ident;
     Type* FinalType;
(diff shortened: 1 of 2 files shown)


oliverschmidt:
Thanks for including a test :-)

PR #1457, opened 2021-04-10
Implemented .CHARSET control command

ReallyNiceGuy:
The target Commander X16 can work in PETSCII or ISO-8859-15 character set
Instead of having to switch back and forth using .CHARMAP commands,
one can use the .CHARSET control command to switch the translation from the
point of usage on.

Documentation:
.CHARSET command [tablename]
where:
  command can be:
    SAVE - to save the current state of the translation table into a temporary scratch pad
    RESTORE - to restore the contents of the scratch pad back into the translation table
    LOAD tablename - to load a defined translation table referenced by the string tablename
  Valid table names at this moment:
    "native" - Disables translation, uses the compilation unit characters directly.
    "target" - Loads the default translation table for the current target
    "atari"
    "osic1p"
    "petscii"


ReallyNiceGuy:
I just noticed that a define of and identifier named "save", "restore" or "load" will break it.
I would like to have some input on how to implement it better.
I was thinking about:
.charset "save"
.charset "restore"
.charset "atari"

or even:
.charsetsave
.charsetrestore
.charset "atari"

oliverschmidt:
Can't the same effect be reached by creating `.inc` files containing `.charmap` commands? I think of something like it was done with

* https://github.com/cc65/cc65/blob/master/include/cbm_petscii_charmap.h
* https://github.com/cc65/cc65/blob/master/include/cbm_screen_charmap.h
* https://github.com/cc65/cc65/blob/master/include/atari_atascii_charmap.h
* https://github.com/cc65/cc65/blob/master/include/atari_screen_charmap.h
* https://github.com/cc65/cc65/blob/master/include/ascii_charmap.h

for the C compiler?

ReallyNiceGuy:
Almost.
My rationale is that, as the compiler knows the target and the corresponding character tables, why do I need to include a file and duplicate the configuration that is already available? It should be much quicker than the include solution (not that it really matters in this day and age).
The issue I was trying to fix is when .charmap was used to change some characters, but I need to switch to another character set temporarily and then back, like in a library. This could be sidestepped by using .byte and hex codes directly, without even the include files that you suggest, but again, the programmer will have to do it manually.

What I found in the code is that the translation is done directly by the directive, so there is no way to implement a pseudo function like
`.asciiz .translate("native","This is the text")`
`.byte .translate("petscii","some text")`
in a backwards compatilbe way.
Actually I would prefer the following solution better than my commit, even if I need 2 more commands to keep backwards compatibility.
`.rawasciiz "This is the text"`
`.rawasciiz .translate("target","This is the text")`
`.rawbyte .translate("petscii","some text")`
These commands would not do automatic translation at all, they would use the bytes as found in the source code.

This pull request is more like an RFC.
I am willing to put the time and effort to implement the agreed solution.
Cheers

mrdudz:
reminds me of #1382

> My rationale is that, as the compiler knows the target and the corresponding character tables, why do I need to include a file and duplicate the configuration that is already available?

i wholeheartly agree with this. having an external include file for this is not only inconvenient - it's also a source for errors (who makes sure that include file actually matches the internal mapping, and how?)

oliverschmidt:
... and you'll never know if `stdio.h` is available and actually contains what you expect it to contain. Therefore all declarations needed should be built into the compiler for sure. And the same goes for this compiler & assembler & linker mess. Imagine their versions don't match! The only way to make sure that cc65 always works as desired is to have it consist of a single file. No idea why I didn't get that idea in all those years ;-))

mrdudz:
Now thats not strawman - it's strawmen.

The assembler already sets the mapping and allows to redefine it. There is no reason at all why it shouldnt be able to reset it.

ReallyNiceGuy:
> No idea why I didn't get that idea in all those years ;-))

Probably because you never had a target with more than one character set available.
I just started playing around with the Commander X16, so fresh eyes and all...

mrdudz:
and BTW - if putting the mapping definition into an include file is what we want, the target specific header files should simply include these files and the internal mapping removed entirely. Just like the internal linker configs were removed in favour of the external ones. Having two sources for the exact same thing is never a good idea - someone *will* forget to keep them in sync and cause a mess. (The absense of test cases for this tells me its already there :))

oliverschmidt:
> [...] the target specific header files should simply include these files and the internal mapping removed entirely [...]

I easily agree with this! To have both builtin charsets *and* charmap pragmas/commands - which make it at least for me unclear how they interact - isn't optimal. I always thought that the cc65 way is the "more abstract" way, not the way following closely the use case at hand - even if that means lower usability. I think the linker configs are _THE_ example.

groessler:
I don't agree. Portable C programs just using stdio or so don't include a target specific header file. Think of a hello-world which just includes stdio.h.

mrdudz:
> To have both builtin charsets and charmap pragmas/commands - which make it at least for me unclear how they interact - isn't optimal.

The problem here is that we arent only looking at the c-compiler but also the assembler. those two must have character mapping tables that match (exactly) for a target. That makes it kindof hard to get rid of the internal mapping (where one mapping can be shared between both easily by using the same sourcefile).

This is kindof a different issue than what the OP intended though :) I kindof agree that the assembler should be able to "reset" the mapping to whatever it was before it was redefined by the user. For anything further than this i'd point to using include files though. Which again makes this complicated - so we actually have to deal with two problems here:

a) how to make sure all possible sources for a mapping are exactly the same
b) what is the standard way to switch between them

My proposal would be to have an option for the assembler and compiler to output the respective include files, which should then be part of the build process. This makes sure the files represent the exact same mapping as the internal one. That then implies NOT having extra commands for the assembler/compiler as the OP intended.

The reason for doing it like this would be simply that neither the assembler nor the compiler will by default, without the user explicitly doing it, include anything - which would be required when the mapping was moved to external files.

ReallyNiceGuy:
> This is kindof a different issue than what the OP intended though :) I kindof agree that the assembler should be able to "reset" the mapping to whatever it was before it was redefined by the user.

I would still miss the save/restore pair.

> For anything further than this i'd point to using include files though. Which again makes this complicated - so we actually have to deal with two problems here:
> 
> a) how to make sure all possible sources for a mapping are exactly the same
> b) what is the standard way to switch between them
> 
> My proposal would be to have an option for the assembler and compiler to output the respective include files, which should then be part of the build process. This makes sure the files represent the exact same mapping as the internal one. That then implies NOT having extra commands for the assembler/compiler as the OP intended.

I find this more complicated than having a translate pseudo function. 
> 
> The reason for doing it like this would be simply that neither the assembler nor the compiler will by default, without the user explicitly doing it, include anything - which would be required when the mapping was moved to external files.

Can I suggest to have two X16 targets? Or, to make things more future proof, add a flag for the intended mapping for the compiler that gets saved on the object file and it is checked at link time?

mrdudz:
> I find this more complicated than having a translate pseudo function.

The issue i have with this is mostly that mappings can also be user defined - and there should be (only) one way to change the mapping. Options that only work for a special case and not the general case are always a bit questionable. I am not against it per se, but it should work generically.

(also, your "save" and "restore" should probably be "push" and "pop" similar to how the segment pragmas work)

mrdudz:
> The issue i have with this is mostly that mappings can also be user defined - and there should be (only) one way to change the mapping.

Forgot: example for this is "petscii" and "screencode" mapping for CBM targets... i think we can all agree we dont want the "screencode" one internally. However it is incredibly common to use it in assembly.

ReallyNiceGuy:
> > I find this more complicated than having a translate pseudo function.
> 
> The issue i have with this is mostly that mappings can also be user defined - and there should be (only) one way to change the mapping.

I agree, I just could not see a way to keep it backwards compatible without a new command. But starting the discussion is a very good outcome, anyways ;)

> Options that only work for a special case and not the general case are always a bit questionable. I am not against it per se, but it should work generically. It should also be possible to still solve the "one source for one mapping" problem when this option exists.

I am sorry, I don't understand what do you mean by "one source for one mapping".

> (also, your "save" and "restore" should probably be "push" and "pop" similar to how the segment pragmas work)

I agree, I didn't want to implement a stack at this stage.

mrdudz:
> I am sorry, I don't understand what do you mean by "one source for one mapping".

right now there are 3 places where the mapping is defined (not for all targets, thats another inconsistent thing)
- the internal table (included by the assembler, compiler)
- the .c include file (for c-code)
- the .inc include file (for assembly code)

from a maintainer viewpoint this is a terrible idea, because those are prone to "drift apart" and then all kinds of subtle problems may arise.

ReallyNiceGuy:
> > I am sorry, I don't understand what do you mean by "one source for one mapping".
> 
> right now there are 3 places where the mapping is defined (not for all targets, thats another inconsistent thing)
> 
> * the internal table (included by the assembler, compiler)
> * the .c include file (for c-code)
> * the .inc include file (for assembly code)
> 
> from a maintainer viewpoint this is a terrible idea, because those are prone to "drift apart" and then all kinds of subtle problems may arise.

I see it now, thank you.

If you only keep the internal tables, then you can have a pragma for C that loads the mapping. To keep compatibility, you can keep the .h files, but you only put the pragma inside.

My view is that just adding the `#pragma charset` and `.charset` would not be too intrusive, and it would keep full backwards compatibility.

I am not sure if I would like to have `.charmap "petscii"` as an option. It is still backwards compatible, but feels wrong to me to have behavior change because of the parameter type.

mrdudz:
oh, I forgot: there is even one more source - there are macros for putting eg screencode into assembler code. i wouldnt be surprised if those interact with the currently active mapping in funny ways and then not produce the expected result at all :)

> If you only keep the internal tables, then you can have a pragma for C that loads the mapping. To keep compatibility, you can keep the .h files, but you only put the pragma inside.

that'd be a smart way to solve it without needing a way to emit the file from the assembler/compiler at least - right

ReallyNiceGuy:
> that'd be a smart way to solve it without needing a way to emit the file from the assembler/compiler at least - right. we'd still need a good way to handle completely custom mappings though.

That is the user responsibility, I would say. He can make his own .h and .inc file with lots of pragmas.

groessler:
How would one then use an arbitrarty charmap (because of a redefined font or so)?

ReallyNiceGuy:
> How would one then use an arbitrarty charmap (because of a redefined font or so)?

You still have .charmap, nothing changes there.

oliverschmidt:
> The problem here is that we arent only looking at the c-compiler but also the assembler. those two must have character mapping tables that match (exactly) for a target. That makes it kindof hard to get rid of the internal mapping (where one mapping can be shared between both easily by using the same sourcefile).

I can't follow here. The C compiler generates hex constants in the assembler source from the char and string constants in the C source. So from my POV the C compiler and the assembler char mapping mechanisms can operate totally independently.

mrdudz:
> I can't follow here. The C compiler generates hex constants in the assembler source from the char and string constants in the C source. So from my POV the C compiler and the assembler char mapping mechanisms can operate totally independently.

you certainly want both to use the (exact) same mapping once you are combining assembler- and c-code. unless you expect me to use hex constants everywhere - which i hope you dont :)

oliverschmidt:
> The issue I was trying to fix is when .charmap was used to change some characters, but I need to switch to another character set temporarily and then back, like in a library.

I can't follow here, e.g. the source

```
lda #'A'
.charmap $41, $55
lda #'A'
.charmap $41, $41
lda #'A'
```

results in the listing

```
000000r 1  A9 41        lda #'A'
000002r 1               .charmap $41, $55
000002r 1  A9 55        lda #'A'
000004r 1               .charmap $41, $41
000004r 1  A9 41        lda #'A' 
```

So it seems very well possible to use .charmap to temporarily change the character mapping.

ReallyNiceGuy:
> So it seems very well possible to use .charmap to temporarily change the character mapping.

Maybe this is not important, but, let's say I wrote some library code that does things to the charmap.
It would be nice if I could just restore the charmap the way it was before my library was used. The user of the library would not need to be aware of changes in the charmap.

ReallyNiceGuy:
Charmap is dangerous and blunt, it seems. A pseudo function that translates on the spot looks like a better solution. I just can't see how to keep backwards compatibility.

oliverschmidt:
I seem to still miss the point completely...

* This thread seems to be related to X16 development. So any code intended for re-use on the X16 can use .charmap's to change the char mapping and then use .charmap's to change the char mapping back. So I don't understand why the user of the re-usable code would need to be aware of anything.

* At least to me the term 'library' refers to an archive of already compiled/assembled code. So even if the code in the library was using a custom char mapping based on .chamap's and wasn't setting the char mapping back, I don't understand how the user of the library would need to be aware of anything.

ReallyNiceGuy:
I was thinking about a macro package.

oliverschmidt:
I admit easily that I have close to no experience with complex macros. Can you please elaborate on what the macros you have in mind should archive? And I haven't understood especially if you want the users of the macros to be able to "actively" use both X16 charsets (and your macros be compatible with that) or if you want the users to stay with PETSCII and your macros need to use ISO<...> "internally".

ReallyNiceGuy:
> I admit easily that I have close to no experience with complex macros. Can you please elaborate on what the macros you have in mind should archive? And I haven't understood especially if you want the users of the macros to be able to "actively" use both X16 charsets (and your macros be compatible with that) or if you want the users to stay with PETSCII and your macros need to use ISO<...> "internally".

At this exact moment it is just theoretic, as I don't have anything concrete yet.
What I noticed was that, as a developer, I must take care of the state of the charmap at all times. If I am using my own macros and it is all my code, it is manageable. But if I use some 3rd party macro package (or if I develop one and share), it is impossible to know the state of the charset at any given time.
As a developer of the macro package, I can play around and change things, but I have no way to restore it to the previous state. This was my reason for the "save/restore" of the current state of character map.

Also, I don't like the idea of duplicating the character sets that the compiler already know into an external .inc file. That was the primary reason for creating the .charset in first place. I just exposed what the compiler already knew.

On the other hand, I also see the value in having this automatic translation done for strings.
I just wished that we could have more control on when to apply the translation, and also, how to apply it.

Maybe a cleaner and less controversial approach could be as follows:
The `.asciiz` and `.byte` commands are used when you want the compiler to use the current state of the charmap, as it is now.
Something like `.rbyte`,` .rbytez`, where you just use the bytes as they are in the source code.
Something like` .translate("codepage", "string to be translated")` to generate a string to be used with `.rbyte` (like `.string` does), but you interpret the` "string to be translated"` as they are in the source code.
Create `.charset "codepage","internalcodepage"`, to initialize an user defined charmap using an internal codepage.
Modify `.charmap` to accept the syntax `.charmap <int>,<int>,"codepage"`, to modify an user defined charmap. When "codepage" is missing, just behave as usual, changing the global charmap.
Where "codepage" is a string containing a valid internal or user defined code page name.
With all this in place, c-style literals would make more sense inside strings, if only '\xnn' for hex and '\nnn' for octal.

oliverschmidt:
> What I noticed was that, as a developer, I must take care of the state of the charmap at all times.

Full ACK.

> As a developer of the macro package, I can play around and change things, but I have no way to restore it to the previous state.

I presume you are referring with this statement to the following part of the previous discussion:

>> (also, your "save" and "restore" should probably be "push" and "pop" similar to how the segment pragmas work)

> I agree, I didn't want to implement a stack at this stage.

So we have `.PUSHCPU` (https://cc65.github.io/doc/ca65.html#ss11.91) and `.PUSHSEG` (https://cc65.github.io/doc/ca65.html#ss11.92). Not that I'm saying that this is necessarily the best way to approach the issue, but... Is it correct to say that

```
.pushcharmap $41
.charmap $41, $61
[...]
.popcharmap $41
```

does conceptually solve the issue?

And that by adding assembler include files to the cc65 distro with a) loads of `.pushcharmap <x> .charmap <x>, <y>` pairs and b) loads of `.popcharmap <x>` that approach would become sort of actually usable?

mrdudz:
> So we have `.PUSHCPU` (https://cc65.github.io/doc/ca65.html#ss11.91) and `.PUSHSEG` (https://cc65.github.io/doc/ca65.html#ss11.92). Not that I'm saying that this is necessarily the best way to approach the issue, but... Is it correct to say that
> 
> ```
> .pushcharmap $41
> .charmap $41, $61
> [...]
> .popcharmap $41
> ```
> 
> does conceptually solve the issue?
> 
> And that by adding assembler include files to the cc65 distro with a) loads of `.pushcharmap <x> .charmap <x>, <y>` pairs and b) loads of `.popcharmap <x>` that approach would become sort of actually usable?

I wouldnt do it per character, push/pop the state of the entire mapping is fine imho (and doesnt require those include files)

oliverschmidt:
> I wouldnt do it per character, push/pop the state of the entire mapping is fine imho (and doesnt require those include files)

I didn't want to actually propose the per char approach. I want to understand the issue and the solution on a conceptual level. Then I/we can move on to optimize the solution for feasibility/usability/<you name it>.

ReallyNiceGuy:
> > So we have `.PUSHCPU` (https://cc65.github.io/doc/ca65.html#ss11.91) and `.PUSHSEG` (https://cc65.github.io/doc/ca65.html#ss11.92). Note that I'm not saying that this is necessarily the best way to approach the issue, but... Is it correct to say that
> > ```
> > .pushcharmap $41
> > .charmap $41, $61
> > [...]
> > .popcharmap $41
> > ```
> > 
> > 
> > does conceptually solve the issue?
> > And that by adding assembler include files to the cc65 distro with a) loads of `.pushcharmap <x> .charmap <x>, <y>` pairs and b) loads of `.popcharmap <x>` that approach would become sort of actually usable?
> 
> I wouldnt do it per character, push/pop the state of the entire mapping is fine imho (and doesnt require those include files)

Yes, that is my understanding also.

oliverschmidt:
> The .asciiz and .byte commands are used when you want the compiler to use the current state of the charmap, as it is now.
Something like .rbyte, .rbytez, where you just use the bytes as they are in the source code.

Totally independent from all other aspects of this discussion I agree that it might be desirable to be able to define strings independently form the current char mapping. Unfortunately the already exsisting commands `.asciiz` and `.byte` don't allow for any solution both at least somewhat reasonable AND backward compatible :-(

oliverschmidt:
>>> And that by adding assembler include files to the cc65 distro with a) loads of .pushcharmap <x> .charmap <x>, <y> pairs and b) loads of .popcharmap <x> that approach would become sort of actually usable?

>> I wouldnt do it per character, push/pop the state of the entire mapping is fine imho (and doesnt require those include files)

> Yes, that is my understanding also.

Just to avoid misunderstandings: You both are talking about the following, right?

```
.pushcharmap
.charmap $01, $xx
.charmap $02, $xx
.charmap $03, $xx
[...]
.popcharmap
```

* This makes a lot of sense to me.
* I don't see why this doesn't need include files anymore. Do you want "everybody" to repeat all those individual `.charmap` commands again and again?

ReallyNiceGuy:
> > I wouldnt do it per character, push/pop the state of the entire mapping is fine imho (and doesnt require those include files)
> 
> I didn't want to actually propose the per char approach. I want to understand the issue and the solution on a conceptual level. Then I/we can move on to optimize the solution for feasibility/usability/[you name it].

I think it would work, but it would not be pretty.
Use case:
```
.macro sample
.pushcharmap
.include newcharmap.inc ; duplicated effort, if this new charmap is just direct translation from ascii to ascii,
                                         ; or reimplementation of ascii to petscii
DONICETHINGS             ; here we have issues:
                                         ; what if I want to translate some bytes,
                                         ; but others I want the charmap as the user left it?
.popcharmap
.asciiz "whatever"
.pushcharmap
.include newcharmap.inc ; contrived, but it might be inevitable,
                                         ; if I use another macro inside this macro somehow
.popcharmap
.endmacro
```

New approach from my comment:
```
.macro sample
.charset "localcharset","native" ; create a new charmap, initialized to "native" (or no translation)
.charmap $41,$61,"localcharset"
.charmap $61,$41,"localcharset"
DONICETHINGS
.rbytez .translate("localcharset","Aa") ; results in aA\000 in the code
.asciiz "I don't care for the mapping one bit here" ; If the user of my package changed the charmap,
      ; it is fine, this is what I expect
.endmacro
```

I hope this clarifies my thinking.
This avoids includes for things that are already in the compiler.

ReallyNiceGuy:
> * I don't see why this doesn't need include files anymore. Do you want "everybody" to repeat all those individual `.charmap` commands again and again?

I want to avoid the use of a bunch of charmaps.
I just want to have:
1. User defined charmaps
2. .translate pseudo function, which takes the name of the charmap (internal or user defined) and a string and returns another string, translated
3. .rbyte (as raw byte) which is exactly the same as .byte, just without the intrinsic call to translation
4. .rbytez as counterpart  to .asciiz

mrdudz:
> I don't see why this doesn't need include files anymore. Do you want "everybody" to repeat all those individual .charmap commands again and again?

i was referring to an include file full of "push" or "pop" commands just to save/restore the current mapping. ie the include file should provide the .charmap commands to change the mapping, no more no less. push/pop is up to the user when he needs it.

> 3. .rbyte (as raw byte) which is exactly the same as .byte, just without the intrinsic call to translation
> 4. .rbytez as counterpart  to .asciiz

i find this extremely kludgy to be honest, and against the principle of how ca65 is designed to work. ie what you want to do there can be done with push/pop - and isnt done often enough to justify extra specialized commands

groessler:
> Something like .rbyte, .rbytez, where you just use the bytes as they are in the source code.

This will then be dependent on the source code encoding (think EBCDIC :-) )

ReallyNiceGuy:
> i was referring to an include file full of "push" or "pop" commands just to save/restore the current mapping. ie the include file should provide the .charmap commands to change the mapping, no more no less. push/pop is up to the user when he needs it.

Fine
> 
> > 1. .rbyte (as raw byte) which is exactly the same as .byte, just without the intrinsic call to translation
> > 2. .rbytez as counterpart  to .asciiz
> 
> i find this extremely kludgy to be honest, and against the principle of how ca65 is designed to work. ie what you want to do there can be done with push/pop - and isnt done often enough to justify extra specialized commands

Ok, as I come as a new user, it seems to me that the intrinsic translation in .asciiz and .byte is the kludge.
I came to this issue like this:
I wanted to put a string in latin1 (the encoding of the source file) to print in X16, in latin1 mode.
So, the simplest way to do it now and have a readable string on the source code is to  redefine the charmap to correspond the bytes to themselves.
Then, I want to go back into petscii mode, and print something else. Now it even worse, because I have to get the include file from C, convert it to an .inc, include in my code at this point, then I can emit the string.

ReallyNiceGuy:
> > Something like .rbyte, .rbytez, where you just use the bytes as they are in the source code.
> 
> This will then be dependent on the source code encoding (think EBCDIC :-) )

Exactly! And I am happy with it. It is "native", meaning native to the developing source file.
But good luck in compiling a file encoded in EBCDIC with ca65 right now. ;)

I understand why it took so long to some brave soul invent Unicode...

ReallyNiceGuy:
> > Something like .rbyte, .rbytez, where you just use the bytes as they are in the source code.
> 
> This will then be dependent on the source code encoding (think EBCDIC :-) )

By the way, the translation tables as they are now assume ascii for the source code.

mrdudz:
> Ok, as I come as a new user, it seems to me that the intrinsic translation in .asciiz and .byte is the kludge.

I have to admit i find this a bit strange myself - from other assemblers i am used to ".byte" leaving the encoding alone, and ".text" (or whatever it uses) doing the translation. I wouldnt mind if ca65 would be changed to work like this either :)

ReallyNiceGuy:
> > I wouldnt mind if ca65 would be changed to work like this either :)
> 
> I'd support a change that .byte doesn't do any translations.

Backwards compatibility will be lost. That is why I suggested .rbyte . .rbytez is not necessary, it is not too hard to add a ,$0 at the end of the definition.

mrdudz:
> Backwards compatibility will be lost.

perhaps. We break this with the toolchain all the time though (most prominent example are the linker configs) so when updating the toolchain you will most likely have to adjust your sources in any case.

ReallyNiceGuy:
> > Backwards compatibility will be lost.
> 
> perhaps. We break this with the toolchain all the time though (most prominent example are the linker configs) so when updating the toolchain you will most likely have to adjust your sources in any case.

One thing is to have a broken config file, another is to have the source code result in something completely different from one version of the compiler to another.
I disagree to a behavior change to .byte

oliverschmidt:
> I just exposed what the compiler already knew.

> This avoids includes for things that are already in the compiler.

The charsets built into the assembler are right now an implementation detail. Allowing access to them by the programmer is something I don't consider desirable.

ReallyNiceGuy:
> > I just exposed what the compiler already knew.
> 
> > This avoids includes for things that are already in the compiler.
> 
> The charsets built into the assembler are right now an implementation detail. Allowing access to them by the programmer is something I don't consider desirable.

I believe that having .charmap exposes them already.

oliverschmidt:
>> I don't see why this doesn't need include files anymore. Do you want "everybody" to repeat all those individual .charmap commands again and again?

> i was referring to an include file full of "push" or "pop" commands just to save/restore the current mapping. ie the include file should provide the .charmap commands to change the mapping, no more no less. push/pop is up to the user when he needs it.

I see. So we would end up with a single include file containing

```
.charmap $00, $00
.charmap $01, $01
.charmap $02, $02
...
.charmap $FF, $FF
```

named `ascii_charmap.inc` that could be combined with the new commands in this way

```
.pushcharmap
.include "ascii_charmap.inc"
...
.popcharmap
```

ReallyNiceGuy:
> > > I don't see why this doesn't need include files anymore. Do you want "everybody" to repeat all those individual .charmap commands again and again?
> 
> > i was referring to an include file full of "push" or "pop" commands just to save/restore the current mapping. ie the include file should provide the .charmap commands to change the mapping, no more no less. push/pop is up to the user when he needs it.
> 
> I see. So we would end up with a single include file containing
> 
> ```
> .charmap $00, $00
> .charmap $01, $01
> .charmap $02, $02
> ...
> .charmap $FF, $FF
> ```
> 
> named `ascii_charmap.inc` that could be combined with the new commands in this way
> 
> ```
> .pushcharmap
> .include "ascii_charmap.inc"
> ...
> .popcharmap
> ```

Still more complicated than having .rbyte for this simple use case.

oliverschmidt:
I think the idea to use whatever assembler commands to define strings either mapped or unmapped has (beside backward compatibility reasons) the problem, that it doesn't work with characters like in `lda #'A'`.

In several other languages there are prefixes for different types of strings. Especially there's the notion of raw strings with `r"sample"`. I have no idea if this is feasible from the implementation effort perspective but with a raw character and string prefix we could have

```
lda #r'A'
.byte r"Hello"
.asciiz r"World"
```

At least this approach both solves the backward compatibility issue as well as the single-character issue.

oliverschmidt:
> Still more complicated than having .rbyte for this simple use case.

I'd be happy to merge two separate extensions:

1. `.pushcharmap` and `.popcharmap` to allow to revert to the char mapping in use before using `.charmap x,y`.
2. r-prefixed raw characters and strings.

ReallyNiceGuy:
> I think the idea to use whatever assembler commands to define strings either mapped or unmapped has (beside backward compatibility reasons) the problem, that it doesn't work with characters like in `lda #'A'`.
> 
> In several other languages there are prefixes for different types of strings. Especially there's the notion of raw strings with `r"sample"`. I have no idea if this is feasible from the implementation effort perspective but with a raw character and string prefix we could have
> 
> ```
> lda #r'A'
> .byte r"Hello"
> .asciiz r"World"
> ```
> 
> At least this approach both solves the backward compatibility issue as well as the single-character issue.

I didn't think about the `lda #'A'` case.
This seems better than creating a new command, for sure.

ReallyNiceGuy:
> > Still more complicated than having .rbyte for this simple use case.
> 
> I'd be happy to merge two separate extensions:
> 
> 1. `.pushcharmap` and `.popcharmap` to allow to revert to the char mapping in use before using `.charmap x,y`.
> 2. r-prefixed raw characters and strings.

I would be happy with this solution.

oliverschmidt:
> I would be happy with this solution.

:-))

Furher above you wrote

> This pull request is more like an RFC.
I am willing to put the time and effort to implement the agreed solution.

In case that's still true then it would be nice if you would contribute the extensions in question. However, there's one aspect left: "We" are pretty strict regarding code style. Please adhere more closely to the prerelevant code style in the code you contribute than you did in this PR.

ReallyNiceGuy:
> > I would be happy with this solution.
> 
> :-))
> 
> Furher above you wrote
> 
> > This pull request is more like an RFC.
> > I am willing to put the time and effort to implement the agreed solution.
> 

I can try, but I noticed that this will require a new type, unless I am mistaken.
The translation is hardcoded in the DoAsciiz and DoByte functions if the type is TOK_STRCON.

> In case that's still true then it would be nice if you would contribute the extensions in question. However, there's one aspect left: "We" are pretty strict regarding code style. Please adhere more closely to the prerelevant code style in the code you contribute than you did in this PR.

Sorry, I tried but old habits are hard to kill. Do you have an **indent** script to point to? I just did a very poor search on the repository and couldn't find one.

oliverschmidt:
> I can try, but I noticed that this will require a new type, unless I am mistaken.
The translation is hardcoded in the DoAsciiz and DoByte functions if the type is TOK_STRCON.

I imagine that this may result in a larger change. Please note that I'm just the maintainer of cc65, not the author. In fact, I can't assist you in how to best approach the necessary code changes.

> Sorry, I tried but old habits are hard to kill.

;-)

> Do you have an indent script to point to? I just did a very poor search on the repository and couldn't find one.

No, the only thing we have is https://github.com/cc65/wiki/wiki/C-and-Assembly-sources-style-guide

ReallyNiceGuy:
I am closing this pull request, revert my changes and start over.

Thanks for the input!

oliverschmidt:
Thanks for your interest to improve cc65 :-)

Issue #1454, opened 2021-04-06
Restarting getopt parsing.


everslick:
Should there be a way to restart the `getopt()` parsing? I know it is not standard, but in my case it would be really helpful. In some cases getopt() can be reset by setting `optind = 1;` but cc65 getopt has an static local variable `place` that cannot be reset from outside.

I would also be happy with a trick, like calling getopt with some defined values to reset the parser. That would not change the API, nor introduce new functions.

opinions?

polluks:
FYI https://stackoverflow.com/questions/60483737/whats-the-portable-way-to-reset-getopt

everslick:
I propose to replace the getopt implementation by this one:

https://github.com/z88dk/z88dk/blob/master/libsrc/stdlib/getopt.c

It does handle `optreset`, correctly uses `opterr` and has a clear license. drawback: slightly bigger.

polluks:
Indeed, 10%
* z88dk   6310
* cc65    5716

greg-king5:
I think that our implementation of `getopt()` shouldn't print any error messages.  The stdio library has a big footprint.  And, what if a programmer wants to use `getopt()` and conio on a platform where stdio and conio shouldn't be used together?

Therefore, I think that the calling program should show the messages (if the programmer wants them).

polluks:
> Therefore, I think that the calling program should show the messages (if the programmer wants them).

Isn't it the job of `opterr`? At least at runtime.

PR #1460, opened 2021-04-13
Added stdio/conio warning

polluks2:
"stdio and conio shouldn't be used together"
https://github.com/cc65/cc65/issues/1454#issuecomment-818268848

### DIFF
--- a/include/conio.h
+++ b/include/conio.h
@@ -54,6 +54,10 @@
 
 
 
+#ifdef _STDIO_H
+#warning Mixing stdio with conio can have unexpected results
+#endif
+
 #include <stdarg.h>
 #include <target.h>
 


mrdudz:
Adding such generic warning in the header doesn't make a lot of sense - stdio also has the file i/o, which is no problem to combine with conio (and its probably done all over the place)

oliverschmidt:
This statement may be true for some targets. On other targets it's definitiely wrong. See e.g. https://github.com/cc65/wiki/wiki/Apple-II-Screen saying:

_[...] shows how well the Apple II conio implementation interoperates with the Apple II ROM routines [...]_

Therefore let me close this pull request. Maybe you want to create one  which is specific to the target(s) with known issues.

PR #1463, opened 2021-04-16
Implemented raw strings

ReallyNiceGuy:
A raw string is denoted by the prefix 'r"'

Example:
  message: .byte r"This will not be converted"

Raw strings are not subject to automatic conversion to
the target charmap.


ReallyNiceGuy:
Not ready yet, I forgot to check for the case lda #'a' and similar.

ReallyNiceGuy:
I just noticed that we will need an .RSTRING pseudo function to create a raw string from an identifier.
From the documentation, this is fine, will work either way:
```
        ; Emulate other assemblers:
        .macro  section name
                .segment        .string(name)
        .endmacro
```
The segment string is not subject to translation anyways.

But the following will not necessarily work as intended:
```
        .macro  M       Arg
                ; Check if the argument string starts with r'a'
                ; If Arg is abc, the following is false
                .if (.strat (.string(Arg), 0) = r'a')
                ...
                .endif
        .endmacro
```
.STRAT is fine, as it uses an already defined string, and at that point, the string itself is already tagged as raw or not.

polluks2:
Smells like Python ;-)

oliverschmidt:
Pure coincidence ;-)

ReallyNiceGuy:
I am trying to write a test for this, but now I don't know how to check for pass or fail.
I noticed that `test/asm/paramcount.s` gets compiled, but even if I change it so that the compiled result is different, it doesn't fail.

This is my draft test for the raw string:
```
; Test ca65's handling of raw strings.

.charmap    $41,$61
.charmap    $61,$41

.macro  test_rawstring arg
    .byte .concat(.string(arg),.rstring(arg))
    .byte .strat(.string(arg),0),.strat(.string(arg),1)
    .byte .strat(.rstring(arg),0),.strat(.rstring(arg),1)
.endmacro

expected:
    .byte   $41, $61, $61, $41
    .byte   $41, $61, $61, $41
    .byte   $41, $61, $61, $41
    lda #$41
    lda #$61
    lda #$61
    lda #$41
got:
    .byte   "a","A",r"a",r"A" 
    test_rawstring  aA
    lda #'a'
    lda #'A'
    lda #r'a'
    lda #r'A'
```

ReallyNiceGuy:
> Smells like Python ;-)

Oh, my... This little comment gave me so much to think about.
I was reviewing my changes and I believe that the raw string is not a good solution.
Here is my explanation:
As the compiler works now, the translation is done only at .BYTE and .ASCIIZ. At any other point, a string is not interpreted, just used as is in the code page of the source file.
After the introduction of raw string, I had to change .CONCAT to do something sensible when receiving a normal string and a raw string. The route I took was to do the translation inside it. This changes everything! Now, If I use .CONCAT to create a identifier, it will give me different results depending on the type of the string I use.
I also had to take care of .STRING, so I created .RSTRING. But I forgot about .SPRINTF, and the same ideia would apply.
What we thought would be the simplest approach ended up being very complicated conceptually.

I would like to close this pull request and open an issue, so we can discuss it further.
I will remove the raw string changes and create a new pull request with only .PUSHCHARMAP and .POPCHARMAP, which are very straight forward and don't change anything in weird ways.

Thoughts?

oliverschmidt:
Separating the charmap stack from the rest is a good idea anyhow.

Regarding the raw string I'm sorry that my proposal obviously wasn't thought through well enough.

ReallyNiceGuy:
> Separating the charmap stack from the rest is a good idea anyhow.
> 
Agreed.

> Regarding the raw string I'm sorry that my proposal obviously wasn't thought through well enough.

No worries, I also didn't think it through.
The push/pop change brings me 99% up where I wanted to be.
I can now do:
```
.pushcharmap
.include "weirdcharmap.inc"
DONICETHINGS
.popcharmap
```

PR #1467, opened 2021-04-18
Implemented charmap stack

ReallyNiceGuy:
New commands:
 .PUSHCHARMAP: will push the current charmap state into an internal stack
 .POPCHARMAP: will restore the current charmap to the last pushed charmap

Details:
  The push and pop facilities are implemented directly inside the tgttrans.h,
  to facilitate its reuse on the C compiler.


src/common/tgttrans.c
```diff
@@ -76,6 +81,8 @@ static unsigned char Tab[256] = {
 
 
 
+
+
```
oliverschmidt, line 85: I don't think these two additional empty lines are desirable.

oliverschmidt:
I ask for two additions:

1. Documentation (in https://github.com/cc65/cc65/blob/master/doc/ca65.sgml)
2. That file asminc/ascii_charmap.inc discussed before containing
```
.charmap $00, $00
...
.charmap $FF, $FF
```
similiar to https://github.com/cc65/cc65/blob/master/include/ascii_charmap.h. When you look at e.g. https://github.com/cc65/cc65/blob/master/asminc/joy-kernel.inc you see that there are .inc files with the same type of header as .h files have. So you can really start out with ascii_charmap.h, do some Find&Replace and remove a few non-relevant things...

ReallyNiceGuy:
> I ask for two additions:
> 
> 1. Documentation (in https://github.com/cc65/cc65/blob/master/doc/ca65.sgml)

Sure, but English is not my native language. I will need review.

> 2. That file asminc/ascii_charmap.inc discussed before containing
> 
> ```
> .charmap $00, $00
> ...
> .charmap $FF, $FF
> ```
> 
> similiar to https://github.com/cc65/cc65/blob/master/include/ascii_charmap.h. When you look at e.g. https://github.com/cc65/cc65/blob/master/asminc/joy-kernel.inc you see that there are .inc files with the same type of header as .h files have. So you can really start out with ascii_charmap.h, do some Find&Replace and remove a few non-relevant things...

Sure.

oliverschmidt:
> Sure, but English is not my native language. I will need review.

I'd say adjusted copies of https://cc65.github.io/doc/ca65.html#ss11.91 and https://cc65.github.io/doc/ca65.html#ss11.87 are supposed to be sufficient. I don't see extensive explanations.

Marrin:
The term ASCII feels a bit weird here because ASCII is only byte values 0 to 127.  The `ascii_charmap.h` header file even has a comment `/* beyond ASCII */` for the second half of the translation table.

oliverschmidt:
> The term ASCII feels a bit weird here because ASCII is only byte values 0 to 127. The ascii_charmap.h header file even has a comment /* beyond ASCII */ for the second half of the translation table.

I noticed that very well. But this is - at least from my perspective - not about "being right". Rather it's about usability and practicality. And from that POV there's ASCII, PETSCII and ATASCII. The file in question is intended to switch from PETSCII or ATASCII to ASCII.

And as you noticed yourself the file with the comment you cite is called ascii_charmap.h. So do you really intend to question the .inc file being named ascii_charmap.inc? Or what is your intention?

asminc/ascii_charmap.inc
```diff
@@ -0,0 +1,291 @@
...
+;/* 2. Altered source versions must be plainly marked as such, and must not   */
+;/*    be misrepresented as being the original software.                      */
+;/* 3. This notice must not be removed or altered from any source             */
+;/*    distribution.                                                          */
+;/*                                                                           */
+;/*****************************************************************************/
+
+;/* No include guard here.  Each charmap header
+;** may be included many times in a source file.
+;*/
```
oliverschmidt, line 31: As far as I know there's nothing like include guards for asm .inc files. From that perspective the comment asks to be removed completely, doesn't it?

ReallyNiceGuy: I can have:
```
.ifndef ASCII_CHARMAP
.define ASCII_CHARMAP 1
.endif
```
But I will remove if you want, not a problem. I just converted the original files automatically.

Marrin:
It is only an ASCII mapping if the input is ASCII.  For 1 byte per character encodings it's an `any_charmap` or better an `identity_charmap` because each byte value is passed through as is, regardless of the encoding of the source code.

I don't know if this becomes that clear by the name `ascii_charmap` that this is the one you want if you want to transfer any byte value unaltered from source code to target binary.

asminc/ascii_charmap.inc
oliverschmidt: > But I will remove if you want, not a problem.

Please do so, thanks in advance!

ReallyNiceGuy: > As far as I know there's nothing like include guards for asm .inc files. From that perspective the comment asks to be removed completely, doesn't it?

Sorry, I clicked something and it resolved this issue, not my intention.

oliverschmidt:
The file name ascii_charmap.h was introduced by @greg-king5 - I suggest that you discuss that matter with him. My only goal is that the .inc file has the same base name as the .h file.

mrdudz:
> It is only an ASCII mapping if the input is ASCII. For 1 byte per character encodings it's an `any_charmap` or better an `identity_charmap` because each byte value is passed through as is, regardless of the encoding of the source code.
> 
> I don't know if this becomes that clear by the name `ascii_charmap` that this is the one you want if you want to transfer any byte value unaltered from source code to target binary.

*ANY* of the charmapping include files assume ASCII as input (actually, iso8859-15, i guess). So for all of them their name refers to what you get as a result when the input is ASCII.

oliverschmidt:
Thanks :-)

ReallyNiceGuy:
My pleasure!

PR #1469, opened 2021-04-18
Move all common project settings into cc65.props

### LINKED ISSUE #1466: Update Visual Studio projects to use shared settings from a props file
I propose to make a few updates to the vcxproj files.

First of all, I propose to move all the shared project settings to the [Directory.Build.props](https://docs.microsoft.com/en-us/visualstudio/msbuild/customize-your-build?view=vs-2019) or a custom props file. So, you change a setting in the props file, and it is used in all projects automatically. It is much better than when all the settings are duplicated for every project and project configurations.

According to project defaults, I propose three variants of better defaults.

Variant A (if we want to make it buildable on default VS2019):
1. Switch toolkit to v142 (VS2019).
2. Switch SDK version to 10.0. So, it will automatically choose the latest available SDK, no need to install specific version:
```
<WindowsTargetPlatformVersion>10.0</WindowsTargetPlatformVersion>
```
_(This method of selection of the latest SDK version is not supported by older VS2017, so it is VS2019 only.)_

Variant B (if we want to support XP):
1. Switch toolkit to v141_xp.
2. Use dynamic linking with `msvcrt.dll` to make files much smaller without requirement to install VS runtime libraries (but it will require to push a lib file from the Microsoft WDK7 to this repo).

Variant C (if we want to make it buildable on both VS2017 and VS2019):
1. Leave toolkit at v141 (VS2017).
2. Use this trick to select latest available Windows 10 SDK (so, at least you won't be enforced to install an older SDK version):
```
<WindowsTargetPlatformVersion>$([Microsoft.Build.Utilities.ToolLocationHelper]::GetLatestSDKTargetPlatformVersion('Windows', '10.0'))</WindowsTargetPlatformVersion>
```
_(It should be compatible with both VS2017 and VS2019.)_

What do you think about it?

vrubleg:
Resolves #1466. Should support both VS2017 and VS2019 without changing any settings.
As an addition, adds the `.vs` directory to the `.gitignore` to let the git ignore VS2019 temporary files.


vrubleg:
I moved almost all settings to the props file. Tested on the VS2019, but it should work on VS2017 also. Please check if I accidentally used a VS2019 feature without a fallback. If it fails, I'll install VS2017 to check what's wrong.

The cc65.props uses some conditions to set various settings. It sets just defaults. If some setting is changed in a project setting, it will be overwritten in a project file. Currently, cc65.props should be edited manually in a text editor. It is possible to make it editable from the VS GUI (similar to settings of a project), but the GUI is very limited, a few important settings are not available there, and it does not allow to use conditions which I used in a few of places.

When a new project is added to the solution, you just need to add this line manually:
```
  <Import Project="cc65.props" />
```
before this line:
```
  <Import Project="$(VCTargetsPath)\Microsoft.Cpp.Default.props" />
```
and it will apply all the defaults.

src/ld65.vcxproj
```diff
@@ -145,4 +116,4 @@
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
   <ImportGroup Label="ExtensionTargets">
   </ImportGroup>
-</Project>
+</Project>
```
oliverschmidt, line 119: Seems like you have unintentionally removed this final newline.

src/ar65.vcxproj
```diff
@@ -13,66 +13,37 @@
...
   </PropertyGroup>
   <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">
     <ClCompile>
-      <PrecompiledHeader>
-      </PrecompiledHeader>
-      <WarningLevel>Level3</WarningLevel>
-      <PreprocessorDefinitions>_CRT_NONSTDC_NO_WARNINGS;_CRT_SECURE_NO_WARNINGS;_CONSOLE;_DEBUG</PreprocessorDefinitions>
-      <AdditionalIncludeDirectories>common</AdditionalIncludeDirectories>
-      <TreatWarningAsError>true</TreatWarningAsError>
+      <PreprocessorDefinitions>_CONSOLE;_DEBUG;%(PreprocessorDefinitions)</PreprocessorDefinitions>
```
oliverschmidt, line 35: Is there a specific reason why you didn't move `_CONSOLE` to cc65.props (like you did with `_CRT_NONSTDC_NO_WARNINGS` and `_CRT_SECURE_NO_WARNINGS`) ?

src/cc65.props
```diff
@@ -0,0 +1,62 @@
...
+			<BufferSecurityCheck>false</BufferSecurityCheck>
+		</ClCompile>
+		<Link>
+			<GenerateDebugInformation>false</GenerateDebugInformation>
+			<EnableCOMDATFolding>true</EnableCOMDATFolding>
+			<OptimizeReferences>true</OptimizeReferences>
+			<LinkTimeCodeGeneration>UseLinkTimeCodeGeneration</LinkTimeCodeGeneration>
+		</Link>
+	</ItemDefinitionGroup>
+</Project>
```
oliverschmidt, line 62: Seems like you have unintentionally removed this final newline.

src/ar65.vcxproj
vrubleg: One day you may want to add a GUI project to this solution. I thought, that it would be smarter to leave it in the project files.

Also, the `common` project doesn't define `_CONSOLE`, but it defines `_LIB`, because it is a static library.

oliverschmidt: I see - thanks for the explanation.

oliverschmidt:
> Currently, cc65.props should be edited manually in a text editor. It is possible to make it editable from the VS GUI (similar to settings of a project), but the GUI is very limited, a few important settings are not available there, and it does not allow to use conditions which I used in a few of places.

Can you please elaborate what exactly we'd loose without the conditions? I must admit that at least for me being able to use the VS GUI to manipulate project settings is VERY desirable! And regarding the missing settings please remember that this isn't about creating delivery binaries.

vrubleg:
Later support of older versions of the Visual Studio also can be added in the future. As far as I can see, we potentially can support VS2008 and newer (if the source code is compatible with those old toolkits).

oliverschmidt:
> Support of older versions of the Visual Studio also can be added in the future.

Great. However, I don't see any reason to go back beyond VS2017.

vrubleg:
You can still change project settings as you did before, it will be just saved into the project file. If you want to make these new settings common for all projects, just copy this line and copy into the similar place of the cc65.props =)

vrubleg:
Allowed to edit the settings from the Property Manager (View -> Other Windows -> Property Manager).

But I would not recommend to use GUI for changing these common settings anyway. That GUI works, but I don't trust it =) At least, check if it placed the setting where you wanted.

The main drawback: when you add a new common setting which should be shared between release and debug configurations, Property Manager by default places it into the section for current configuration type only. So, you need to duplicate the value for both configurations or move it to the common section manually.

Tested on VS2019 only.

vrubleg:
There is a possibility to create three props files:
cc65_common.props
cc65_debug.props
cc65_release.props

So you can choose where to put your common settings this way from GUI. If you put them into common, it goes into all configurations. If you put them into debug or release, it goes into debug or release only.

src/cc65.props
vrubleg: @oliverschmidt seems like VS2019 removes them by default when you save a project file or a props file from GUI.

oliverschmidt:
> At least, check if it placed the setting where you wanted.

For sure!

> The main drawback: when you add a new common setting which should be shared between release and debug configurations, Property Manager by default places it into the section for current configuration type only. So, you need to duplicate the value for both configurations or move it to the common section manually.

For "normal" project settings, one can select multiple/all configurations to change release and debug simultaniously. Is that possible with the Property Manager too?

oliverschmidt:
> There is a possibility to create three props files:
cc65_common.props
cc65_debug.props
cc65_release.props

Thanks for the offer, but it seems that this isn't necessary/appropriate for this scenario. I rather see me/others to quickly change some setting to try something - and discard it later. Not so much maintaining settings that need to stay consistent over build configurations.

vrubleg:
> For "normal" project settings, one can select multiple/all configurations to change release and debug simultaniously. Is that possible with the Property Manager too?

This part lacks this feature. When you open Property Manager, you choose project and configuration type you want to edit, and then you change it. So you can change setting for one configuration at a time. But if a value is already placed in the section for all configurations, it is edited there. So, if you edit Output Directory, it will be edited for all configurations at once, Property Manager won't move this setting to configuration specific section.

src/cc65.props
oliverschmidt: I see :-(

oliverschmidt:
Thank you very much for the detailed and helpful explanation :-)

vrubleg:
> I rather see me/others to quickly change some setting to try something - and discard it later.

It can be done in project settings as it was done before. Overwriting default project settings works perfectly fine. The only drawback that you need to repeat it for all projects. Or modifying the default settings from the Property Manager for Debug and Release builds.

vrubleg:
Probably it is possible to enforce Property Manager to put new setting into the common section. I can try this.

So, now when you add a new default setting, it is automatically added to the current configuration only (Release/Debug).

Probably, it is possible to do that all the new default go to all configurations (Release and Debug). But in this case if you need separate values for Release and Debug, you'll need to move this setting in the props file manually.

vrubleg:
Oh, another weird thing about VS. Property Manager saves the props file when you build your project or try to close VS. It doesn't save the props file when you just press "Save" button. Seems like a bug.

oliverschmidt:
> Probably it is possible to enforce Property Manager to put new setting into the common section. I can try this.

> So, now when you add a new default setting, it is automatically added to the current configuration only (Release/Debug).

> Probably, it is possible to do that all the new default settings go to all configurations (Release and Debug). But in this case if you need separate values for Release and Debug, you'll need to move this setting in the props file manually into the section specific to your configurations Release and Debug.

> Also it is important to keep in mind that this file is imported really early (to let us set the toolkit and SDK version), so some VC++ specific variables may not be available, or some values may be overwritten later by the VC++ specific imports. So, if you are not sure if it worked fine, just check in the project settings which defaults are displayed.

A whole bunch of magic...

oliverschmidt:
* The Travis-CI VS 2017 based build seems to work still fine (https://travis-ci.org/github/cc65/cc65/jobs/767478705).
* It doesn't make sense to postpone merging until I personally have fully digested all consequences of this change from the user perspective.

Therefore -> merging...

vrubleg:
You could just pull my branch for testing =) I might use something that is not supported in VS2017 by accident.

Issue #1470, opened 2021-04-18
Use UTF-8 for source code


vrubleg:
Currently, source code in an ANSI encoding, and in some cases it has non-ASCII characters. On machines with non-Latin locale (it is most of non-western countries) [such comments](https://github.com/cc65/cc65/blob/master/src/ar65/error.c#L9) look like this in the Visual Studio:
```
/* (C) 1998-2003 Ullrich von Bassewitz                                       */
/*               RцmerstraЯe 52                                              */
```
but it should look like this:
```
/* (C) 1998-2003 Ullrich von Bassewitz                                       */
/*               Römerstraße 52                                              */
```
Not so good.

I propose to convert all the source files into UTF-8. There are not so many places where non-ASCII characters are used, so the conversion will affect a small amount of lines in the source code files. Visual Studio 2017+ supports UTF-8 natively. It is also possible to create an `.editorconfig` to tell Visual Studio that all the new text files should use UTF-8 whenever a non-ASCII character is used. UTF-8 is perfectly compatible with ASCII, so when non-ASCII characters are not used, the files will be identical to usual ASCII files.

What do you think?

oliverschmidt:
If we do any changes to the source files I suggest to remove the offending characters so that source files end up being pure ASCII. Please replace `Römerstraße` with `Roemerstrasse`.

mrdudz:
What we did in VICE was: remove all non ASCII stuff from Sourcecode, keep it in ASCII. UTF-8 caused too much subtle problems in various places.

vrubleg:
@mrdudz what kind of problems?

UTF-8 compatibility with ASCII is really good. If it is in the comments, any C compiler would be fine about it.

mrdudz:
> @mrdudz what kind of problems?

Mostly with other tools, the buildsystem, the documentation, scripts we are using etc. The C-Compiler is probably the only thing that didnt screw up one way or another :=)

oliverschmidt:
While this is probably an interesting discussion I personally for sure prefer ASCII source files.

@vrubleg: I'd appreciate if you could contribute a PR making the changes poposed in https://github.com/cc65/cc65/issues/1470#issuecomment-822030845

oliverschmidt:
Fixed with https://github.com/cc65/cc65/commit/f272bc8f42ebe26b7bab8e44665957eb690dfcf3.

oliverschmidt:
Thanks for pointing out!

[...]

PR #1472, opened 2021-04-19
Less non-ASCII characters.

vrubleg:
Ticket #1470


oliverschmidt:
Thanks for your effort. However I directly pushed the changes in question already.

Issue #1057, opened 2020-06-21
CA65: .IFREF can't see that a label was referenced after


vrubleg:
It is documented [here](https://cc65.github.io/doc/ca65.html#ss11.63), but it would make our lives a bit better if `.IFREF` were a bit smarter and allowed to check if a label was referenced anywhere in current translation unit.

greg-king5:
That enhancement probably will require us to change ca65 into an assembler that does a variable number of input passes.  The initial passes will look for references, and not set labels.  The last pass will set the labels.  Then, the output pass will resolve expressions.

vrubleg:
I use ca65 for my hobby project (for historical reasons: I started from cc65, and then switched to pure assembly). It is nice, and has quite powerful macros, but there are a few things which chould make using ca65 even more enjoyable, all of them are related to each other: #796, #1056, #1057. At this moment, I'm in the middle of other project, and can't switch to learning internals of ca65 and implementation of these enhancements. But if somebody knows internals of the assembler, likes these ideas, and wants to implement them, I would pay 50$ for each of the tickets (or 200$ if of them are implemented by the same person). Not much, but it's just a "thank you" for making ca65 better.

PR #1476, opened 2021-04-24
Add .REFERTO control command

vrubleg:
I propose to add new `.REF` control command which just marks a symbol as referenced and does nothing else. It is useful when a subroutine has a couple of entry points, and we want to include these pieces of code when they are used only. Example:
```
.DEFINE REF .REF

.ENUM
	ppu_mask
	ppu_ctrl        ; + ppu_scroll_base
	ppu_scroll_x    ; from 0 to 255
	ppu_scroll_y    ; from 0 to 239
.ENDENUM

test:

	LDA #10
	JSR ppu_scroll_set_b0_x
	RTS

.IFREF ppu_scroll_set_b0_x

ppu_scroll_set_b0_x:

	STA ppu_scroll_x
	CLC
	REF ppu_scroll_select_base0_x

.ENDIF
.IFREF ppu_scroll_select_base0_x

ppu_scroll_select_base0_x:

	LDA ppu_ctrl
	AND #%11111110
	STA ppu_ctrl
	RTS

.ENDIF
```

The `ppu_scroll_set_b0_x` should fall through into the `ppu_scroll_select_base0_x`, which itself is a separate usable function. If you remove the `REF ppu_scroll_select_base0_x`, the `ppu_scroll_select_base0_x` won't be included into the binary and it will cause an error. So, we need somehow explicitly reference the `ppu_scroll_select_base0_x` without emitting any machine code.

The name of the command can be changed. I can imagine such names:
- `REF`.
- `.REF` (this PoC).
- `.REFERENCE`.
- `.REFERENCED` (it is accepted as an alias to `.REF` in this PoC also).
- `.MAKEREF`.
- `.MAKEREFERENCE`.
- `.MAKEREFERENCED`.

Looking at the `.DEFINE` command, probably `.REFERENCE` is the most logical name for this command. But I would add the short `.REF` also. (Also, I would probably add `.DEF` as an alias for the `.DEFINE` control command, for consistency.)

I'm actually going to use it as `REF` (without `.`) pseudo-instruction in my code, because it will look nicer in place of a replaced `JMP` (which could be used there, but it will emit a couple of useless bytes). I thought that this name probably won't be accepted into the mainline version of the ca65, so this PoC code uses `.REF`. It can be used as `REF` in your code easily using a simple `.DEFINE REF .REF`, so I don't insist on using `REF`, any name is fine for me.

If you are positive about adding such a feature, I'll update this PR to reflect this new command in the documentation, and to change name of the instruction if required.

### DIFF
--- a/src/ca65/pseudo.c
+++ b/src/ca65/pseudo.c
@@ -1730,7 +1730,7 @@ static void DoPushSeg (void)
 
 
 
-static void DoReferenced (void)
+static void DoReferTo (void)
 /* Mark given symbol as referenced */
 {
     SymEntry* Sym = ParseAnySymName (SYM_ALLOC_NEW);
@@ -2165,7 +2165,8 @@ static CtrlDesc CtrlCmdTab [] = {
     { ccNone,           DoPushCharmap   },
     { ccNone,           DoPushCPU       },
     { ccNone,           DoPushSeg       },
-    { ccNone,           DoReferenced    },      /* .REFERENCED */
+    { ccNone,           DoUnexpected    },      /* .REFERENCED */
+    { ccNone,           DoReferTo       },      /* .REFERTO */
     { ccNone,           DoReloc         },
     { ccNone,           DoRepeat        },
     { ccNone,           DoRes           },
--- a/src/ca65/pseudo.c
+++ b/src/ca65/pseudo.c
@@ -1730,6 +1730,18 @@ static void DoPushSeg (void)
 
 
 
+static void DoReferenced (void)
+/* Mark given symbol as referenced */
+{
+    SymEntry* Sym = ParseAnySymName (SYM_ALLOC_NEW);
+    if (Sym)
+    {
+        SymRef (Sym);
+    }
+}
+
+
+
 static void DoReloc (void)
 /* Enter relocatable mode */
 {
@@ -2153,7 +2165,7 @@ static CtrlDesc CtrlCmdTab [] = {
     { ccNone,           DoPushCharmap   },
     { ccNone,           DoPushCPU       },
     { ccNone,           DoPushSeg       },
-    { ccNone,           DoUnexpected    },      /* .REFERENCED */
+    { ccNone,           DoReferenced    },      /* .REFERENCED */
     { ccNone,           DoReloc         },
     { ccNone,           DoRepeat        },
     { ccNone,           DoRes           },
--- a/src/ca65/pseudo.c
+++ b/src/ca65/pseudo.c
@@ -1734,8 +1734,7 @@ static void DoReferTo (void)
 /* Mark given symbol as referenced */
 {
     SymEntry* Sym = ParseAnySymName (SYM_ALLOC_NEW);
-    if (Sym)
-    {
+    if (Sym) {
         SymRef (Sym);
     }
 }
--- a/doc/ca65.sgml
+++ b/doc/ca65.sgml
@@ -3765,7 +3765,7 @@ See: <tt><ref id=".ASCIIZ" name=".ASCIIZ"></tt>,<tt><ref id=".BYTE" name=".BYTE"
   See: <tt><ref id=".POPSEG" name=".POPSEG"></tt>
 
 
-<sect1><tt>.REFERTO</tt><label id=".REFERTO"><p>
+<sect1><tt>.REFERTO, .REFTO</tt><label id=".REFERTO"><p>
 
   Mark a symbol as referenced.
 
@@ -3779,11 +3779,24 @@ See: <tt><ref id=".ASCIIZ" name=".ASCIIZ"></tt>,<tt><ref id=".BYTE" name=".BYTE"
   Example:
 
   <tscreen><verb>
+        .ifref NegateValue              ; If this subroutine is used
+        NegateValue:                    ; Define it
+                lda     #0
+                sec
+                sbc     Value
+            .ifref ResetValue           ; If the ResetValue is also used
+                jmp     SetValue        ; Jump over it
+            .else
+                .refto  SetValue        ; Ensure that SetValue will be included
+            .endif
+        .endif
+
         .ifref ResetValue               ; If this subroutine is used
         ResetValue:                     ; Define it
                 lda     #0              ; Set a default value
-                .referto SetValue       ; Ensure that SetValue will be included
+                .refto  SetValue        ; Ensure that SetValue will be included
         .endif
+
         .ifref SetValue                 ; If this or previous subroutine is used
         SetValue:
                 sta     Value
--- a/doc/ca65.sgml
+++ b/doc/ca65.sgml
@@ -3213,7 +3213,8 @@ See: <tt><ref id=".ASCIIZ" name=".ASCIIZ"></tt>,<tt><ref id=".CHARMAP" name=".CH
         .endif
   </verb></tscreen>
 
-  See also: <tt><ref id=".REFERENCED" name=".REFERENCED"></tt>
+  See also: <tt><ref id=".REFERENCED" name=".REFERENCED"></tt>, and
+  <tt><ref id=".REFERTO" name=".REFERTO"></tt>
 
 
 <sect1><tt>.IMPORT</tt><label id=".IMPORT"><p>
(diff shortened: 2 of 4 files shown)


greg-king5:
`.DEFINE` is a hack.  It was adapted from the C-language preprocessor.  I think that it shouldn't be abbreviated.  `.DEF()` should continue to be the abbreviation for `.DEFINED()`.

`.REF()` should continue to mean `.REFERENCED()`.  I believe that the short forms are more useful in expressions than as control commands.

I vote for the Assembly command `.REFERTO`.

vrubleg:
@greg-king5 @oliverschmidt what about `.REFTO` as an alias to `.REFERTO`?

vrubleg:
Made a version with `.REFERTO`.

src/ca65/pseudo.c
```diff
@@ -1730,6 +1730,18 @@ static void DoPushSeg (void)
 
 
 
+static void DoReferTo (void)
+/* Mark given symbol as referenced */
+{
+    SymEntry* Sym = ParseAnySymName (SYM_ALLOC_NEW);
+    if (Sym)
+    {
```
greg-king5, line 1738: Remember our slightly different style about where to put left braces.

vrubleg: Fixed.

vrubleg:
Ouch, I used the shorter `.REFTO` in the docs by mistake. Forgot that we chose longer variant. `.REFTO` felt so natural when I was writing that part though... Will fix.

vrubleg:
Updated the docs. Provided an example how it can be used also.

oliverschmidt:
@greg-king5: To me the value of this PR isn't clear - but that may very well just mean that I haven't thought about it enough. So if you consider this PR as valuable please merge as you see fit - thanks :-)

vrubleg:
There is no way to refer an implicitly used block of data or code now. If somebody uses `.IFREF` a lot, this new command makes coding a bit easier. It is just an explicit way to say that this code uses some other code or data implicitly. 

I use this approach when a function has two entry points (one of them just sets some default arguments and falls through into the other) quite often. And also I create libraries which use `.IFREF` for every piece of code or data, so when nothing is used, nothing is included into the final binary.

oliverschmidt:
> And also I create libraries which use .IFREF for every piece of code or data, so when nothing is used, nothing is included into the final binary.

That's pretty much the point I don't get. My understanding is that this is the very task of an archive (aka library) created with ar65.

vrubleg:
I use ca65 as a "flat" assembler. All the libraries are just includes in this case. It is a supported use case, according to the docs:

> 11.62 .IFREF
> Conditional assembly: Check if a symbol is referenced. Must be followed by a symbol name. The condition is true if if the the given symbol was referenced before, and false otherwise.
> **This command may be used to build subroutine libraries in include files** (you may use separate object modules for this purpose too).
> Example:
> ```
>            .ifref  ToHex                   ; If someone used this subroutine
>            ToHex:  tay                     ; Define subroutine
>                    lda     HexTab,y
>                    rts
>            .endif
>```

This approach adds much more control on the layout of functions and data. For example, I can create functions with a few entry points, and ensure that only used parts will be in the final binary. Many of these tricks are not possible when you link a few object files.

Why I chose ca65 if I don't use cc65? Initially, I wanted to use cc65, because I wanted to make a project for NES, and I thought that learning 6502 assembly would be too much. But it turned out that I had to learn 6502 assembly anyway to do what I wanted. It became clear that 6502 assembly is really easy to code, so I switched to pure ca65. It has many powerful features for pure assembly coding, so I didn't look for another assembler.

oliverschmidt:
It's by no means unusual to use ca65 as assembler without even looking at cc65. See  e.g. https://github.com/cc65/wiki/wiki/Applications-written-in-assembler-with-ca65

But using ca65 without cc65 doesn't imply to not make use of features like libraries or linker configurations.

However, you pointed why you don't want an archive-type library so that's now clarified.

vrubleg:
Well, I use a linker configuration to create segments for a proper NES file =)

polluks:
> > And also I create libraries which use .IFREF for every piece of code or data, so when nothing is used, nothing is included into the final binary.
> 
> That's pretty much the point I don't get. My understanding is that this is the very task of an archive (aka library) created with ar65.

Indeed, it isn't very useful to move features of the toolchain into the assembler.

vrubleg:
`.IFREF` is already here, this approach is already supported by ca65. I just suggest how it can be improved.

vrubleg:
@greg-king5 what do you think about this change?

greg-king5:
> @greg-king5, what do you think about this change?

I think of `.REFERTO` as a forward reference with a use that's similar to the use of forward function declarations in C.
It might help with issue #1057.

greg-king5:
I think that some more code should be added to the document's example.  This could be put above `ResetValue`:
```asm
.ifref NegateValue
NegateValue:
        lda     #0
        sec
        sbc     Value
        jmp     SetValue
.endif
```
You could say that it shows the difference between the traditional way of referring to a label and the new way.  The first way creates some code where it refers to `SetValue`, while the second way doesn't add anything where it refers to `SetValue`.

mrdudz:
> I think that some more code should be added to the document's example. This could be put above `ResetValue`:
> 
> ```assembly
> .ifref NegateValue
> NegateValue:
>         lda     #0
>         sec
>         sbc     Value
>         jmp     SetValue
> .endif
> ```
> 
> You could say that it shows the difference between the traditional way of referring to a label and the new way. The first way creates some code where it refers to `SetValue`, while the second way doesn't add anything where it refers to `SetValue`.

Pardon my ignorance - but shouldnt this code simply go into a seperate source (and object) file and then into a library?

oliverschmidt:
> [...] shouldnt this code simply go into a seperate source (and object) file and then into a library?

I already asked this very question in https://github.com/cc65/cc65/pull/1476#issuecomment-826362198. From the answer:

> For example, I can create functions with a few entry points, and ensure that only used parts will be in the final binary, and they are in the right order. Many of these tricks are not possible when you link a few object files. There is always just one object file in my projects, and all the libraries are just includes, so I have 100% control of everything in the final binary.

My understanding is that one of the aspects is that an optional entry point in a separate source file requires a jump while the approach outlined here allows to fall-through (or at least get away with a branch).

groessler:
> My understanding is that one of the aspects is that an optional entry point in a separate source file requires a jump while the approach outlined here allows to fall-through (or at least get away with a branch).

Yes. That's also the only advantage I see from having separate object files in a library. Sometimes it might be handy, but (guessed) 95% of all projects can do without it.

groessler:
> I think of .REFERTO as a forward reference with a use that's similar to the use of forward function declarations in C.
It might help with issue #1057.

But in general, the change appears to be useful. If somebody wants to use .ifref" etc.

oliverschmidt:
In cases like this I tend to balance the complexity added to the code base with the desirability of the feature. Here the added complexity is minimal. So even a low desirability is enough - and everybody seems to agree that at least this level is given.

So my perspective is: If the change looks technically correct and the feature in question is well enough documented then it "wants" to be merged.

@greg-king5: You have some ideas on documentation improvements. Please merge when those are taken care of - thanks!

vrubleg:
> Pardon my ignorance - but shouldnt this code simply go into a seperate source (and object) file and then into a library? If anything, the documentation should provide examples that show why this is not done in the first place.

There are different approaches. One of them is C-like, relies on a linker. The other one is FASM-like, which doesn't have linker at all. Both approaches have advantages and disadvantages. I like the FASM-like approach for pure assembly projects, because it allows me to have more control and save a few bytes here and there. When I write in assembly, it is usually my hobby, and I enjoy every moment when I saved a byte or two and reduced amount of cycles required for execution. Sometimes it makes code a bit more complicated, but if I were afraid of complicated code, I wouldn't use assembly language =)

As in my example, `.IFREF` + `.REFERTO` help to create a library function with two entry points, one of them sets default argument values, and the other one does some complex work without using a `JMP` at the end of the first part.

Original idea of this pattern comes from desire to have default arguments for functions, exactly like when you code in C++. I use this pattern a lot for this purpose. When I was reversing Need For Speed III, I saw that its compiler (Watcom C++) sometimes used this pattern also.

> I think that some more code should be added to the document's example. This could be put above `ResetValue`:
> 
> ```assembly
> .ifref NegateValue
> NegateValue:
>         lda     #0
>         sec
>         sbc     Value
>         jmp     SetValue
> .endif
> ```
> 
> You could say that it shows the difference between the traditional way of referring to a label and the new way. The first way creates some code where it refers to `SetValue`, while the second way doesn't add anything where it refers to `SetValue`.

I don't see huge examples in the docs, so I tried to keep the example as small as possible to show the idea. If we add the `NegateValue` entry point, it should be a bit more complicated to handle the case when the `ResetValue` is not used in the code, so it would look like this:
```assembly
        .ifref NegateValue
        NegateValue:
                lda     #0
                sec
                sbc     Value
            .ifref ResetValue
                jmp     SetValue
            .else
                .referto SetValue
            .endif
        .endif

        .ifref ResetValue               ; If this subroutine is used
        ResetValue:                     ; Define it
                lda     #0              ; Set a default value
                .referto SetValue       ; Ensure that SetValue will be included
        .endif

        .ifref SetValue                 ; If this or previous subroutine is used
        SetValue:
                sta     Value           ; Complex processing can be done here
                rts
        .endif
```

This example is quite long, I'm not sure if it is OK to include it into the docs. Also it shows why I propose to have a `.REFTO` shorthand as an alias. `.REFERTO` is longer than 7 characters, and breaks 8-character alignment of the argument for one char, it is visible in the example above. I don't insist on having this alias, absence of this alias is not a blocker, but even the example above would look nicer with a `.REFTO`.

mrdudz:
> There are different approaches. One of them is C-like, relies on a linker. The other one is FASM-like, which doesn't have linker at all. Both approaches have advantages and disadvantages. I like the FASM-like approach for pure assembly projects, because it allows me to have more control and save a few bytes here and there. When I write in assembly, it is usually my hobby, and I enjoy every moment when I saved a byte or two and reduced amount of cycles required for execution. Sometimes it makes code a bit more complicated, but if I were afraid of complicated code, I wouldn't use assembly language =)

I totally get the drift :) IMHO ca65 is the wrong tool for the job in this case (because everything in the toolchain is centered around the linker and its not really tailored to many of the ugly things you'd do in hand written assembly, eg 64tass does a much better job at that). In any case, as said, it should be clearly documented what this is good for.

greg-king5:
I don't think that anyone has said that one wants a limit on the length of examples.  They're small simply because it's hard to add useful flesh to their skeletons (we're good programmers, not great writers).

I accept the `.REFTO` alias.  It does seem like an appropriate complement to the `.REF()` function alias.

vrubleg:
Added the `.REFTO` alias and updated the docs to include that huge example of code.

vrubleg:
Glad to see this feature merged. Will update my code to use it immediately =)

Issue #1473, opened 2021-04-19
Unexpected result when including file from inside .macro


ReallyNiceGuy:
When using .include inside a .macro, the contents of the included file are added after the macro expansion, no matter where the .include was done.

Ex:
```
.org $080D
.segment "STARTUP"
.segment "INIT"
.segment "ONCE"
.segment "CODE"

.macro LITERAL2 arg
    .pushcharmap
    .include "ascii_charmap.inc"
    .byte arg
    .popcharmap
.endmacro

LITERAL2 "text in latin1"
```
This results in:
```
ca65 V2.19 - Git f272bc8f4
Main file   : min.asm
Current file: min.asm

000000r 1               .org $080D
00080D  1               .segment "STARTUP"
00080D  1               .segment "INIT"
00080D  1               .segment "ONCE"
00080D  1               .segment "CODE"
00080D  1    
00080D  1    
00080D  1               .macro LITERAL2 arg 
00080D  1                   .pushcharmap
00080D  1                   .include "ascii_charmap.inc"
00080D  1                   .byte arg 
00080D  1                   .popcharmap
00080D  1               .endmacro
00080D  1    
```
**NOTE: this is still in PETSCII**
```
00080D  1  CC 4F 52 45  LITERAL2  {"Lorem Ipsum",0}
000811  1  4D 20 C9 50  
000815  1  53 55 4D 00  
000819  1    
000819  2               ;/*****************************************************************************/
000819  2               ;/*                                                                           */
000819  2               ;/*                              ascii_charmap.inc                            */
000819  2               ;/*                                                                           */
...
000819  2               ;/*****************************************************************************/
000819  2
000819  2               ;/* ASCII */
000819  2               .charmap $00, $00
000819  2               .charmap $01, $01
...
000819  2               .charmap $FD, $FD
000819  2               .charmap $FE, $FE
000819  2               .charmap $FF, $FF
000819  2
000819  1
```
**NOTE: Now we are in latin1, but we expected to be in PETSCII because of the .pushcharmap and .popcharmap pair.**

greg-king5:
This issue is related to issue #231.

ReallyNiceGuy:
My patch will not fix issue #231. It will make it fail to compile, as the include file is processed at the time of the definition, not at the time of the expansion.

oliverschmidt:
Fixed with https://github.com/cc65/cc65/pull/1477.

Issue #231, opened 2015-11-14
ca65: scope created inside a macro doesn't affect files included inside it


tokumaru-x:
The contents of the included file aren't placed inside the new scope when this macro is used:

```
.macro IncludeHidden FilePath
    .scope
        .include FilePath
    .endscope
.endmacro
```

mckaygerhard:
that since are happened, to track what changes in the code?

ReallyNiceGuy:
After application of patch #1477 , this will result in a compilation error.
The contents of a file referenced by an .INCLUDE is processed immediately, as if it was inline on the source code, so it has no access to the argument of the macro at expansion time.

Please close the issue.

PR #1477, opened 2021-04-26
Parse file included inside a macro at definition time

ReallyNiceGuy:
Fixes issue number #1473.
The contents of the included file will be parsed at macro definition time, as if it is part of the current file.

### DIFF
--- a/src/ca65/macro.c
+++ b/src/ca65/macro.c
@@ -491,6 +491,23 @@ void MacDef (unsigned Style)
     */
     while (1) {
 
+        /* Check for include */
+        if (CurTok.Tok == TOK_INCLUDE) {
+            /* Include another file */
+            NextTok ();
+            /* Name must follow */
+            if (CurTok.Tok != TOK_STRCON) {
+                ErrorSkip ("String constant expected");
+            } else {
+                SB_Terminate (&CurTok.SVal);
+                if (NewInputFile (SB_GetConstBuf (&CurTok.SVal)) == 0) {
+                    /* Error opening the file, skip remainder of line */
+                    SkipUntilSep ();
+                }
+            }
+            NextTok ();
+        }
+
         /* Check for end of macro */
         if (Style == MAC_STYLE_CLASSIC) {
             /* In classic macros, only .endmacro is allowed */
@@ -508,22 +525,6 @@ void MacDef (unsigned Style)
             if (TokIsSep (CurTok.Tok)) {
                 break;
             }
-        }
-
-        if (CurTok.Tok == TOK_INCLUDE) {
-            /* Include another file */
-            NextTok ();
-            /* Name must follow */
-            if (CurTok.Tok != TOK_STRCON) {
-                ErrorSkip ("String constant expected");
-            } else {
-                SB_Terminate (&CurTok.SVal);
-                if (NewInputFile (SB_GetConstBuf (&CurTok.SVal)) == 0) {
-                    /* Error opening the file, skip remainder of line */
-                    SkipUntilSep ();
-                }
-            }
-            NextTok ();
         }
 
         /* Check for a .LOCAL declaration */
--- a/src/ca65/macro.c
+++ b/src/ca65/macro.c
@@ -510,6 +510,22 @@ void MacDef (unsigned Style)
             }
         }
 
+        if (CurTok.Tok == TOK_INCLUDE) {
+            /* Include another file */
+            NextTok ();
+            /* Name must follow */
+            if (CurTok.Tok != TOK_STRCON) {
+                ErrorSkip ("String constant expected");
+            } else {
+                SB_Terminate (&CurTok.SVal);
+                if (NewInputFile (SB_GetConstBuf (&CurTok.SVal)) == 0) {
+                    /* Error opening the file, skip remainder of line */
+                    SkipUntilSep ();
+                }
+            }
+            NextTok ();
+        }
+
         /* Check for a .LOCAL declaration */
         if (CurTok.Tok == TOK_LOCAL && Style == MAC_STYLE_CLASSIC) {
 


spiro-trikaliotis:
Are you really sure that your patch works without breaking anything else (other than making #231 fail to compile)?

I am currently working on the parser, too, in order to fix #1427. What I found out is that the behaviour of the parser is rather complicated, and it is rather hard to follow.

To me, your solution looks "too easy". Furthermore, you add include file processing at a second point in the sources, that is, you add some type of redundancy. To me, it does not look right, although I cannot tell exactly why at the moment.

ReallyNiceGuy:
> Are you really sure that your patch works without breaking anything else (other than making #231 fail to compile)?
> 
At least the automated tests pass. As it wasn't working at all before, I have a strong belief that it won't break any sensible code.

> I am currently working on the parser, too, in order to fix #1427. What I found out is that the behaviour of the parser is rather complicated, and it is rather hard to follow.
I completely agree with you.

> 
> To me, your solution looks "too easy".
At this point, it is very easy, because I just start injecting the contents of the included file on the macro itself.

> Furthermore, you add include file processing at a second point in the sources, that is, you add some type of redundancy. To me, it does not look right, although I cannot tell exactly why at the moment.
I thought about it and could not find a sane way to follow the normal path. The DoInclude function is working in normal token mode, but in the macro processing, it is in RawTokenMode.

oliverschmidt:
> At least the automated tests pass. [...]

The tests focus almost exclusively on the C compiler so that doesn't say much :-(

oliverschmidt:
Any additional thoughts on how to move forward with this PR?

ReallyNiceGuy:
> Any additional thoughts on how to move forward with this PR?

I was thinking about my own comment and it doesn't make much sense to have the include work at expansion time.
Otherwise, the parsing would have to be done incrementally, and I doubt that this is what was intended.

oliverschmidt:
> [...] doesn't make much sense to have the include work at expansion time.

I can't really follow you :-( Please elaborate.

ReallyNiceGuy:
I said above:

>  It could be improved by including the file at the point of use, but, to be honest, I didn't figure out on how to implement it.

But then I thought about it and concluded that this idea is too complicated.
The question we have to ask is: what does it mean to have an .include inside a macro.
The implementation in this patch is:
  pull the contents of the include and put inside the macro, as if it was in the same source file. What follows is the that rules apply as if you had typed the file in at that point (a new macro definition will not work, for example)

The option is:
  pull the contents of the include file only at the expansion of the macro. This would allow the issue #231 to work as the reporter intended, but now you can define a macro inside the expansion of a macro, inside the expansion of a macro, etc. I am not sure if this is what .include was intended for.

I hope this makes sense.

spiro-trikaliotis:
The behaviour of ca65 with macros is as follows:
When parsing a macro, it convert them to tokens and stores them. Other than that, it does nothing. Most importantly, it does not interpret any token other than the corresponding .endmacro, or an .local (for traditional/classic macros only).

That's all done inside of the while (1) loo where @ReallyNiceGuy added his .include processing.

That's why I have my doubts with this solution:

1. Does it break non-classic macros?
2. As everything is processed at use time, why not this? What are the side effects?

I did some test and immediately found a regression. Have a look at the attached zip and the files a1.a65 and a2.65.

In a1.a65, there is a .repeat inside of the .macro. It works flawlessly.
In a2.a65, the .repeat inside of the .macro has been put into a.inc instead. Note that a1.a65 and a2.a65 result in the identical commands once tokenized (ommitting the .include directive itself).

While a2.a65 assembles witthout this PR, it does not with this PR. The .repeat instructions are not recognized. They are not recognized because they are handled now at definition time, and the loop does not know about the .repeat. They must be handled at use time, as they are in the original code.

[broken.tar.gz](https://github.com/cc65/cc65/files/6406699/broken.tar.gz)

ReallyNiceGuy:
This is the compilation with the changes in this PR with your slightly modified examples.
Notice that the compilation result is exactly the same.
[pr_test.tar.gz](https://github.com/cc65/cc65/files/6406857/pr_test.tar.gz)

ReallyNiceGuy:
> @ReallyNiceGuy: You must slow down a little bit and give me time to react... This is no chat here. ;)

Sorry, I just posted as soon as I got a test finished. I don't expect an immediate reaction, no worries there.

> You are right, I managed to confuse the "original" (cc65's git master) ca65 and your version. So, to correct my statement: Your version assembles the file, the original does not. So, your code is an improvement.
> 
> I am not sure how and why this happened. I must have had a bad brain at that moment.

It happens to everybody, also, no worries.
> 
> So, it seems your version is better than the original. I will do some additional tests to find out what happens if you nest macros and includes (and macros in the includes and includes in that macro and so on).

Nested macros defined inside the include will fail to compile, in the same manner as nested macros in the same source file.

> 
> I still believe handling it at usage time would be the better approach, as we are splitting the logic between two places now, which will make the code less manageable, IMHO. It is already very complicated.

I only handle the TOK_INCLUDE and the subsequent string with the file name, nothing else.
The issue I was having with the expansion code was that it was way more intrusive. I would have to start getting tokens from a file while in the middle of the expansion, and it was not pretty. Handling at the definition time was easy, and the only things I can think of that we miss with this approach are:
1. I cannot have macro definitions in the files included inside a macro (fine by me, might cause havoc if allowed)
2. I cannot use a macro to include a file defined at expansion time by an argument to that macro (maybe there are use cases for this one, issue #231 comes to mind), but I don't see it as that bad.

spiro-trikaliotis:
You are right, I overlooked that nested macros are not possible now, too. So, this is not a regression.

I tested some more things. I did not find any problems with this patch. I still don't like the approach of handling it at definition time, but I also see your point that it would be harder to do it right at expansion time.

From my point of view (if this matters at all!), this patch can be merged.

oliverschmidt:
> From my point of view (if this matters at all!), this patch can be merged.

@spiro-trikaliotis: Thank you very much for the thoughts you put into assessing this PR! In fact it matters a lot as it triggers me merging it ;-)

@ReallyNiceGuy: Thanks for being open to answer all questions coming up.

oliverschmidt:
I'm wondering if https://sourceforge.net/p/cc65/mailman/message/37306831/ is a result of merging this PR. The message says:

```
wanted to do some work with the ancheron VM...started with the basics...
makefile is given which uses ca65
but I am on windows.  
gnu make installed.
cc65 installed.  

I run make which would execute:
rm -r obj
rm -r bin
mkdir  obj
mkdir  bin
ca65 -l obj/acheron.o.lst src/acheron.asm -g -o obj/acheron.o -I src
ld65 -Ln bin/labels -C src/acheron.cfg -m obj/map -o bin/acheron.prg obj/acheron.o
# Display size information for each component. Only the above line is necessary for the build.
grep '^[^_]* Size=' obj/map
grep '\._size__' bin/labels
dir bin/acheron.prg

but at the ca65 step I get this:
rm -r obj
RemoveDirectory(C:\retro\6502\acheronvm-master\acheronvm-master\obj)
rm -r bin
RemoveDirectory(C:\retro\6502\acheronvm-master\acheronvm-master\bin)
mkdir  obj
mkdir  bin
ca65 -l obj/acheron.o.lst src/acheron.asm -g -o obj/acheron.o -I src
src/acheron.asm:38: Error: String constant expected
115 / 128 opcodes defined
make: *** [obj/acheron.o] Error 1

it's having a issue with this macro:
.macro reportSize label, filename
 label:
 .include filename
 .ident (.concat("_size__", .string(label))) = * - name
 .export .ident (.concat("_size__", .string(label)))
 .endmacro


specifically the .include filename

Any ideas?  
```

@ReallyNiceGuy: What do you think?

ReallyNiceGuy:
Yes, it is. The error is caused by the .include being resolved at definition time, not at usage time.

I will check what was the result before the patch and how to modify the macro to make it work as expected again.

ReallyNiceGuy:
He can remove this macro completely, it is not used anywhere.

ReallyNiceGuy:
I created a PR to the project directly.

oliverschmidt:
> I created a PR to the project directly.

I really appreciate your super-quick and very helpful reaction! Thanks :-)

ReallyNiceGuy:
👍

PR #1487, opened 2021-05-05
Extend #pragma wrapped-call to support "bank" argument

mrdudz:
This adds the option to use "bank" in place of the numeric value in the wrapped-call pragma. So instead of 
```
#pragma wrapped-call (push, mytrampoline, 23)
void somefunc1(void);
#pragma wrapped-call (pop)
```
you can also write
```
#pragma wrapped-call (push, mytrampoline, bank)
void somefunc1(void);
#pragma wrapped-call (pop)
```
in that case the compiler will use the assemblers .bank directive to load the value defined in the bank attribute in the linker config - which removes the requirement to maintain those values manually.

I used "bank" in place of the numeric value (instead of making the value an optional parameter) because this makes it easy to add an additional parameter without breaking anything. For example at some point in the future an additional parameter might be added to pass 16- or more bits of banking info (right now it is always 8 bit).

Please review and let me know how to improve what if this is somehow not meeting the general ideas how things should be done in the compiler - i have very little experience with the compiler internals :)

That said it would be great to get this merged soonish because it will be incredibly useful for what i am doing right now :)

### DIFF
--- a/doc/cc65.sgml
+++ b/doc/cc65.sgml
@@ -1388,7 +1388,10 @@ parameter with the <tt/#pragma/.
   the <tt/Y/ register if it wraps any variadic functions (they have "<tt/.../"
   in their prototypes).
 
-  The identifier is an 8-bit number that's set into <tt/tmp4/.
+  The identifier is an 8-bit number that's set into <tt/tmp4/. If the identifier
+  is "bank", then a <tt><htmlurl url="ca65.html#.BANK" name=".bank"></tt> operator will be used
+  to determine the number from the bank attribute defined in the linker config,
+  see <htmlurl url="ld65.html#MEMORY" name="Other MEMORY area attributes">.
 
   The address of a wrapped function is passed in <tt/ptr4/.  The wrapper can
   call that function by using "<tt/jsr callptr4/".
--- a/doc/cc65.sgml
+++ b/doc/cc65.sgml
@@ -1389,9 +1389,11 @@ parameter with the <tt/#pragma/.
   in their prototypes).
 
   The identifier is an 8-bit number that's set into <tt/tmp4/. If the identifier
-  is "bank", then a <tt><htmlurl url="ca65.html#.BANK" name=".bank"></tt> operator will be used
+  is "bank", then a <tt><url url="ca65.html#.BANK" name=".bank"></tt> operator will be used
   to determine the number from the bank attribute defined in the linker config,
-  see <htmlurl url="ld65.html#MEMORY" name="Other MEMORY area attributes">.
+  see <htmlurl url="ld65.html#MEMORY" name="Other MEMORY area attributes">. Note that
+  this currently implies that only the least significant 8 bits of the bank attribute
+  can be used.
 
   The address of a wrapped function is passed in <tt/ptr4/.  The wrapper can
   call that function by using "<tt/jsr callptr4/".
--- a/src/cc65/funcdesc.h
+++ b/src/cc65/funcdesc.h
@@ -58,7 +58,7 @@
 /* Bits that must be ignored when comparing funcs */
 #define FD_IGNORE   (FD_INCOMPLETE_PARAM | FD_OLDSTYLE | FD_OLDSTYLE_INTRET | FD_UNNAMED_PARAMS | FD_CALL_WRAPPER)
 
-
+#define WRAPPED_CALL_USE_BANK   0x0100U /* WrappedCall uses .bank() */
 
 /* Function descriptor */
 typedef struct FuncDesc FuncDesc;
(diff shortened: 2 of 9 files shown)


doc/cc65.sgml
```diff
@@ -1388,7 +1388,10 @@ parameter with the <tt/#pragma/.
   the <tt/Y/ register if it wraps any variadic functions (they have "<tt/.../"
   in their prototypes).
 
-  The identifier is an 8-bit number that's set into <tt/tmp4/.
+  The identifier is an 8-bit number that's set into <tt/tmp4/. If the identifier
+  is "bank", then a <tt><htmlurl url="ca65.html#.BANK" name=".bank"></tt> operator will be used
+  to determine the number from the bank attribute defined in the linker config,
+  see <htmlurl url="ld65.html#MEMORY" name="Other MEMORY area attributes">.
```
mrdudz, line 1394: please let me know if there is a better way to reference across documents - htmlurl is a bit kludgy

src/cc65/funcdesc.h
```diff
@@ -72,6 +72,7 @@ struct FuncDesc {
     struct FuncDesc*    FuncDef;        /* Descriptor used in definition     */
     struct SymEntry*    WrappedCall;    /* Pointer to the WrappedCall        */
     unsigned char       WrappedCallData;/* The WrappedCall's user data       */
+    int                 WrappedCallUseBank;/* Flag: does WrappedCall use .bank() or literal value */
```
mrdudz, line 75: Perhaps there is a better way to add flags. Please elaborate :)

src/cc65/pragma.c
```diff
@@ -535,12 +536,15 @@ static void WrappedCallPragma (StrBuf* B)
         goto ExitPoint;
     }
 
-    if (!GetNumber (B, &Val)) {
+    /* Next must be either a numeric value, or "bank" */
+    if (HasStr (B, "bank")) {
+        UseBank = 1;
+    } else if (!GetNumber (B, &Val)) {
```
mrdudz, line 542: Perhaps it should be changed to "Value or 'bank' required for wrapped-call identifier"?

src/cc65/funcdesc.h
greg-king5: A "boolean" variable makes sense here.  The code to handle it is simpler than the code would be for any other way to add that flag.

src/cc65/pragma.c
greg-king5: I agree.  But, I would say, "the word 'bank'":
```
"Value or the word 'bank' required for wrapped-call identifier"
```

doc/cc65.sgml
oliverschmidt: I think it needs to be briefly mentioned that this implies ignoring the top 24 bits of the 32-bit bank attribute.

src/cc65/wrappedcall.c
```diff
@@ -64,13 +64,13 @@ static IntPtrStack WrappedCalls;
...
 
-void PushWrappedCall (void *Ptr, unsigned char Val)
+void PushWrappedCall (void *Ptr, unsigned char Val, int UseBank)
 /* Push the current WrappedCall */
 {
     if (IPS_IsFull (&WrappedCalls)) {
         Error ("WrappedCall stack overflow");
     } else {
-        IPS_Push (&WrappedCalls, Val, Ptr);
+        IPS_Push (&WrappedCalls, Val | (UseBank << 8), Ptr);
```
oliverschmidt, line 73: This seems pretty ugly to me. If you need to "encode" the new flag and the normal value into one variable here anyway I suggest to do that in the first place.

The way the code is written now, it implies that there is a value and additionally an optional flag meaning that the value needs to be interpreted in a different way.

However, this isn't true. Rather there are two types of values: The normal value and a special value. Therefore I suggest to extent the `unsigned char Val` to `unsigned Val` and set it to the value of a new macro (which is define as 0x0100) - and get rid of `UseBank` altogether.

src/cc65/funcdesc.h
oliverschmidt: I'm surprised: There's already `unsigned Flags` in line 66. All that would be necessary would be to define an additional `FD_...` macro. However, I suggest to get rid of a flag/boolean value alltogether - see my other comment.

oliverschmidt:
https://git-scm.com/book/en/v2/Git-Branching-Rebasing

You have a local copy of the upstream repo. The Upstream repo was changed (by me) since you made changes to your copy. Thefore you need to update your local copy. Only then it is possible to move the change you have done in your local copy to the upstream repo.

Updating your local copy means that your local changes are "removed", then your local copy being updated and finally your local changes being "re-applied". This process is done by git and called 'rebasing' your local changes.

In this case I have changed a line you have changed too. Therefore there's a conflict. You have to resolve that conflict in order to be able to re-apply your changes to your local copy.

groessler:
Hmm. It works well for me normally.
Occasionally something slips in which I hadn't noticed and which wasn't intended. Therefore I'm always looking at the diff before I make a PR. (Take this "always" with a grain of salt. The normal "always", not the "always always" :-) )

Issue #1494, opened 2021-05-09
Output errors so emacs can parse the line number


audetto:
Currently errors are written to console like this
```
hello.c(5): Error: Unexpected newline
hello.c(6): Error: ')' expected
```
which emacs does not parse.

Most compilers output it as 

```hello.c:5: Error: Unexpected newline
hello.c:6: Error: ')' expected
hello.c:6: Error: ';' expected
```

which can be clicked on emacs easily.
Change is trivial and I am happy to propose a PR, but I wonder if this has already been discussed in the past.

oliverschmidt:
Apart from the line number notation... Are you specifically talking about `Unexpected newline` or is this an example for a certain class of error messages?

audetto:
yes, just the `file:line` notation that emacs accepts by default.
I am sure one can configure emacs to parse `file(line)` but patching `error.c` seems much simpler and portable solution.

https://github.com/cc65/cc65/blob/5d84a4ba137e32163390ccd058c32ad169b4ca94/src/cc65/error.c#L130

I see other tools with the same syntax (`ca65`), but I have only used `cc65` so far.

oliverschmidt:
As far as I understand you'd like to see cc65 conforum (more) to https://www.gnu.org/prep/standards/html_node/Errors.html

I'm personally open for a change - but it needs to include all parts of the cc65 tool chain.

audetto:
yes, I agree on both points.

oliverschmidt:
I suggest that we wait a bit if there are strong opposite opinions. If there aren't, I'm willing to merge a PR consistently changing the syntax.

mrdudz:
This has been bothering me recently as well, my editor of choice also likes file:line better - so +1 for changing it :)

spiro-trikaliotis:
vim recognizes both formats (just tested), so I have no objections. :)

audetto:
Thank you guys.
This is done now,

PR #1497, opened 2021-05-10
Standard formatting of error messages.

audetto:
https://www.gnu.org/prep/standards/html_node/Errors.html

Issue: https://github.com/cc65/cc65/issues/1494


audetto:
I am still looking if I have missed anything, but the idea is this.

BTW, if you think the string "%s:%u:" should be a macro for future configuration and customisation, it can be done.

src/ld65/exports.c
```diff
@@ -774,17 +774,19 @@ static void PrintUnresolved (ExpCheckFunc F, void* Data)
...
                      "Unresolved external '%s' referenced in:\n",
-                     GetString (E->Name));
+                     name);
             while (Imp) {
                 unsigned J;
                 for (J = 0; J < CollCount (&Imp->RefLines); ++J) {
                     const LineInfo* LI = CollConstAt (&Imp->RefLines, J);
                     fprintf (stderr,
-                         "  %s(%u)\n",
+                         "  %s:%u: Error: Unresolved external '%s'\n",
```
greg-king5, line 786: It's redundant.  Maybe you should remove the previous error line, and show only the list of reference lines.

src/ld65/exports.c
```diff
@@ -774,17 +774,19 @@ static void PrintUnresolved (ExpCheckFunc F, void* Data)
         if (E->Expr == 0 && E->ImpCount > 0 && F (E->Name, Data) == 0) {
             /* Unresolved external */
             Import* Imp = E->ImpList;
+            const char * name = GetString (E->Name);
```
oliverschmidt, line 777: The predominant style is `char*`, not `char *`.

audetto:
I made the suggested changes, added one more case in `exports.c`.

If one searches `%s(%u)` in the codebase, it still appears in a few places:

1. `_afailed.c`, `mousedemo.c`, `mouse-test.c`: I think these run in the guest os, so not sure they are needed, but could be for uniformity
2. `exports.c`: https://github.com/cc65/cc65/blob/5d84a4ba137e32163390ccd058c32ad169b4ca94/src/ld65/exports.c#L693 I am not too sure what they are
3. `preproc.c`: https://github.com/cc65/cc65/blob/5d84a4ba137e32163390ccd058c32ad169b4ca94/src/cc65/preproc.c#L693 is this a line number?

polluks2:
Yes, number 2 uses GetSourceLine().
No, number 3 is a counter.

oliverschmidt:
1. Please adjust `_afailed.c`. Please leave `mousedemo.c` and `mouse-test.c` alone. They are not about line numbers at all.
2. As @polluks2 wrote.
3. As @polluks2 wrote.

PR #1501, opened 2021-05-14
[cc65] Less superfluous error messages

acqn:


oliverschmidt:
I don't see this PR receiving feedback. Additionally we don't have barly any tests checking for error messages. Therefore I opt to merge right away.

PR #1505, opened 2021-05-16
[ccc65] #1504 fix

### LINKED ISSUE #1504: [cc65] Some compilation failures
```c
#include <stdio.h>

int main(void)
{
    int i = 0, *p = &i;
    switch (i) case 0: case 1: i = 21;  /* Should be OK but fails */
    p++[0] += 21;                       /* Should be OK but fails */
    printf("%d\n", i);
    return i != 42;
}
```
Expected output:
```
42
```

This currently fails to be compiled with cc65.

acqn:
Fixed #1504.


oliverschmidt:
This seems to be the type of PR not initiating a fruitful discussion - and it fixes issues demonstrated by exsisting tests. I opt to merge now.

PR #1509, opened 2021-05-17
C64 Retro Replay EMD driver

mrdudz:
Just stumbled about this ... it is not ready to merge yet, but perhaps someone wants to pick it up - seems a nice weekend project :)

todo:
- test if it still works with current libs
- rework a bit so the size of the RAM is properly detected. it should work also with just 8k (regular Action Replay and Nordic Power) and 64k (modern reimplementations)


mrdudz:
seems to work ok now, will merge soonish if noone complains :)

Issue #650, opened 2018-05-17
Have RUN/STOP + RESTORE call exit()


oliverschmidt:
On the Apple II the cc65 startup code makes CTRL + RESET call exit(). See https://github.com/cc65/cc65/blob/master/libsrc/apple2/crt0.s for how it avoids re-entrance.

It seems desirable to have on CBM machines RUN/STOP + RESTORE work in the same way. Especially as the CBM C Library contains a destructor to close all open files (just like the Apple II C Library does).

mrdudz:
Just stumbled about this... i wonder what the intend is. Usually runstop+restore "force exits" a basic program, in most ml programs it will do nothing (and in those that use NMI for something it will often cause a crash). Incase the respective pointer stays intact in low mem, it will even "close all files" (and do a general i/o reset) so there is nothing left to do as far as i can see.

polluks:
A SIGKILL should be generated =)

oliverschmidt:
As everybody knows, I'm no CBM insider. I understand that cc65 programs in general want to behave like BASIC programs. So from my perspective it seems totally normal to have that key combo "force exit" a cc65 just in the same way it "force exits" a BASIC program. And for that purpose calling exit() seems the totally normal approach. It doesn't matter too much if there are any built-in exit handlers in the C library which need to be called because the user can at any point use atexit() to add his own handlers, which should to be called.

mrdudz:
> It doesn't matter too much if there are any built-in exit handlers in the C library which need to be called because the user can at any point use atexit() to add his own handlers, which should to be called.

OK - so we want runstop/restore call the atexit handlers, and then exit - and not just "force exit" like it would happen with a basic program (which really is a hard abort, more like what a segfault would do in a modern OS).

I am not sure this can be implemented in a 100% safe way though... the RESTORE key has quite some hardware bounce so the nmi triggering two or more times in a short time is to be expected (sometimes), which might screw up the exit code.

Other than that it should be trivial to implement, in the same way as in that apple2 code, just using the 0138/9 vector and doing a similar check for the runstop key as the ROM does: http://unusedino.de/ec64/technical/aay/c64/romfe43.htm

oliverschmidt:
> OK - so we want runstop/restore call the atexit handlers, and then exit - and not just "force exit" like it would happen with a basic program (which really is a hard abort, more like what a segfault would do in a modern OS).

I see, I misinterpreted what you mean with "force exit". But now we're on the same page :-)

[...]

PR #1515, opened 2021-05-21
kindof hacky implementation of what was requested in issue #650

mrdudz:
This is a somewhat naive implementation of what was requested in issue #650

After doing this, i have to say i dont really like it, and the behaviour feels strange to me. It is already a significant amount of code for (imho) very little gain. I'd propose a classic "NMI lock" instead (which simple disables run/stop+restore)

Problems with this implementation:
- it fails with anything that banks out the kernal (for whatever reason). To handle this correctly a second nmi handler is needed that handles those NMIs slightly different (ie not chain to the system NMI). not sure how to implement this in a generic way.
- it will probably behave unexpected with anything that uses a custom NMI (rs232 drivers are the classic example).
- unsure how it will behave when a kernal NMI is needed (kernal rs232)
- exit code will always be -1 when exiting with runstop/restore
- on exit it prints "break in 816", probably some kernal flag needs to be cleared to supress this

As said, i find the whole behaviour kindof weird, i'd expect runstop/restore to either do a basic warmstart - or nothing at all. (and since re-running the program will not work for most non trivial programs, i dont really see the point either)

oh well, here it is :)

### DIFF
--- a/libsrc/c64/crt0.s
+++ b/libsrc/c64/crt0.s
@@ -46,13 +46,26 @@ Start:
 
         jsr     callmain
 
-; Back from main() [this is also the exit() entry]. Run the module destructors.
+        ; Avoid a re-entrance of donelib. This is also the exit() entry.
+_exit:
+        sta     stsave            ; Save the return code
 
-_exit:  pha                     ; Save the return code on stack
+        ldx     #<exit
+        lda     #>exit
+        jsr     setnmi
+
+; Run the module destructors.
+
         jsr     donelib
 
-; Copy back the zero-page stuff.
+; Restore the original NMI vector.
+exit:   ldx     #$01
+:       lda     nmisave,x
+        sta     $0318,x
+        dex
+        bpl     :-
 
+; Copy back the zero-page stuff.
         ldx     #zpspace-1
 L2:     lda     zpsave,x
         sta     sp,x
@@ -61,7 +74,7 @@ L2:     lda     zpsave,x
 
 ; Place the program return code into BASIC's status variable.
 
-        pla
+        lda     stsave
         sta     ST
 
 ; Restore the system stuff.
@@ -75,6 +88,37 @@ L2:     lda     zpsave,x
 
         rts
 
+.segment        "CODE"
+setnmi:
+        stx     $0318
+        sta     $0319
+        rts
+
+exitnmi:
+        pha
+        ; save $01
+        lda     $01
+        pha
+        ; enable i/o (switch to standard memory config)
+        lda     #$36
+        sta     $01
+        ; check run/stop
+        lda     #$7f
+        sta     $dc00
+        lda     $dc01
+        bpl     :+              ; branch if pressed
+        ; restore $01
+        pla
+        sta     $01
+        pla
+        ; continue with saved system vector
+        jmp     (nmisave)
+:
+        ; throw away saved $01 value and akku (keep memory config)
+        pla
+        pla
+        lda     #$ff    ; exit code (-1)
+        jmp     _exit
 
 ; ------------------------------------------------------------------------
 
@@ -90,6 +134,13 @@ L1:     lda     sp,x
         dex
         bpl     L1
 
+; save original NMI vector
+        ldx     #$01
+:       lda     $0318,x
+        sta     nmisave,x
+        dex
+        bpl     :-
+
 ; Set up the stack.
 
         lda     #<(__MAIN_START__ + __MAIN_SIZE__)
@@ -102,6 +153,11 @@ L1:     lda     sp,x
         lda     #14
         jsr     BSOUT
 
+; set NMI vector to our exit nmi
+        ldx     #<exitnmi
+        lda     #>exitnmi
+        jsr     setnmi
+
 ; Call the module constructors.
 
         jmp     initlib
@@ -112,6 +168,8 @@ L1:     lda     sp,x
 
 .segment        "INIT"
 
-mmusave:.res    1
-spsave: .res    1
-zpsave: .res    zpspace
+mmusave:    .res    1
+spsave:     .res    1
+zpsave:     .res    zpspace
+stsave:     .res    1
+nmisave:    .res    2


oliverschmidt:
First of all thanks for the effort :-)

Just to recapture: I said that this functionality would be desirable because of e.g. closing files. Then you said, that this is done anyway. Then I mentioned `atexit()`. So at this point we sort of agrred that the functionality would be at least somewhat desirable. So far we didn't touch the question, how well it can be implemented in the first place.

Now you bring up - from my very limited CBM knowledge POV very substantial - reasons to rather not implement this functionality. I can follow your reasoning well and am open to closing my issue. It's basically up to you (and mybe other CBM gurus around) to decide...

mrdudz:
Please dont call me guru =D

From my point of view its not really worth the effort. It CAN be implemented correctly (working in all cases), but this will require significantly more effort and code space.

The strongest argument hower to me is, that no matter what we do, its not possible to re-run most programs anyway, so whats the point of doing all this?

And yes, the normal system NMI handler does a BASIC warm start, which closes all files, initializes I/O and some other things.

oliverschmidt:
The ability to re-run a program is from my POV pretty much unrelated to proposed feature. To me it is desirable as it allows to do all cleanup steps that may be necessary. I'm primarily thinking about avoiding file system inconsistencies or dangling IRQ pointers. If the default NMI handler already takes care of those (and similiar) things, then the proposed feature is probably already (partly) implemented.

greg-king5:
> ..., which closes all files, initializes I/O and some other things.

It closes all computer files.  It does nothing to disk units' files.  Basically, computer files are only metadata (the primary/secondary addresses that everybody read about recently) that tell the computer where to find "the real files".  STOP-RESTORE cancels that data.  The computer forgets; the drive files become orphans.  Write files that aren't closed properly are unterminated (like C strings without '\0').

If a user is lucky, and remembers about the difference between computer and drive files, then that one can tell _each_ drive's (separate) DOS to close its files.  ... (if the user remembers) ...

Therefore, an argument can be made for this kind of code.  It would remember for us.

mrdudz:
> I'm primarily thinking about avoiding file system inconsistencies

The CBM DOS will produce so called "splat files" (not closed) if the files are not closed by explicitly sending CLOSE - neither the default NMI handler nor this implementation will prevent this. if you want this, then even more code needs to be added, ie an atexit handler that explicitly closes all open files. Again, i dont think its worth the effort - users wont expect this to work (nor does it work like this in any other program).

mrdudz:
> The ability to re-run a program is from my POV pretty much unrelated to proposed feature. To me it is desirable as it allows to do all cleanup steps that may be necessary.

To me, it raises the question what the usecase for pressing runstop/restore is.

oliverschmidt:
I think/thought it is: "Quit the currently running program and return to the BASIC prompt."

mrdudz:
> I think/thought it is: "Quit the currently running program and return to the BASIC prompt."

But to do what? As mentioned before elsewhere - most C64 Programs can not be "shut down" like this, very few programs offer an explicit "exit" feature, and as a result most users will not expect this feature to exist. They will either powercycle or press the reset button - and programs that "force exit" on runstop/restore are mostly considered "unprofessional" (which is why many either disable this completely, or use it to restart the program). To me handling this looks like a very academical thing to do, with little relevance in practise (and as said before, handling it properly will be quite some effort).

Oh and... regarding those open files - when you'd exit the program in the middle of a file operation, you'd likely end up with a broken (incomplete) file anyway, regardless if its properly closed or not - so closing it doesnt really solve anything (unless you expect the program to be able to handle such incomplete files, which in turn most programs will not). Any "corruption" (as in splat files) can easily be fixed by using a "validate" DOS command - which again most users will be familiar with.

mrdudz:
Another thought: perhaps instead of putting this code into crt0 this way, it would be a better idea to implement it seperately as signal handling. Then crt0 can simply disable (lock) NMI by default, and those who want to have such exit handling on run/stop restore can use a signal handler to do whatever needs to be done. Sounds more reasonable to me :)

oliverschmidt:
Again, I'm not insisting on anything. Maybe it's really all about user expectation. On the Apple II, a user runs a program, exits the program, and runs another program. Instead of figuring out how to exit this and that program "cleanly", users tend to use the Reset key to "quickly" get back to the prompt. Maybe it's more of a Reset thing on the C64 too(?).
And regarding the files: Please note that I was talking about file __system__ inconsistencies, not file __content__ inconsistencies. But maybe that's just no issue on the C64...

mrdudz:
> Instead of figuring out how to exit this and that program "cleanly", users tend to use the Reset key to "quickly" get back to the prompt. Maybe it's more of a Reset thing on the C64 too(?).

As said, most ppl will probably either powercycle or press reset - since most programs have no option to exit. This is even more true these days, where cartridges like 1541U or TC64 are commonly used to load software - there is simply no point in exiting to the BASIC prompt, you just enter the cartridge menu directly and load another program.

> Please note that I was talking about file system inconsistencies, not file content inconsistencies. But maybe that's just no issue on the C64...

Its no issue. The DOS has a builtin "validate" command which will clean up the filesystem from incomplete/not closed files, and nothing breaks if those "splat files" exist.

oliverschmidt:
So if there are no objections I suggest that I close my issue.

@mrdudz: Again thanks for the effort you put into this PR - for the waste basket after all :-(

polluks:
CBM simply has no ProDOS bye equivalent.

oliverschmidt:
See https://github.com/cc65/cc65/pull/1515#issuecomment-846590551.

mrdudz:
(incase someone is wondering, i am abusing open/closing the PR for triggering the github actions - since there is no way to do this manually

greg-king5:
> (incase someone is wondering, i am abusing open/closing the PR for triggering the github actions - since there is no way to do this manually

If you click on a check's `Details` link, then you go to its log.  Above the log, on the right side, is a button that reruns the Github actions for that PR.

mrdudz:
OOOH! Damn, didnt see this. Good to know :)

PR #1517, opened 2021-05-23
Creativision, fixes #1488

### LINKED ISSUE #1488: [creativision][doc] Doc or link to doc for creativision-specific functions?
I see that some PSG functions are provided in the lib:
https://github.com/cc65/cc65/blob/master/include/creativision.h
https://github.com/cc65/cc65/blob/master/libsrc/creativision/psg.s

but most of them are not documented at all and no example is provided.

For example, 
What does `psg_outb` do exactly? Can it be used to right into PSG registers?

It would be nice to have an example that shows how to use  `bios_playsound`?
What is a note tripple? 3 notes? Do I have to play 3 notes? How do I play a single note?

polluks2:


oliverschmidt:
There seem to be no comments -> merging...

PR #1528, opened 2021-05-30
Fix Atari keyboard code for cursor up key

IrgendwerA8:


groessler:
Test program?

IrgendwerA8:
How?

groessler:
A simple program in which one can "walk" around the screen with the cursor keys. Something like that....

groessler:
Not required. But would be nice...

oliverschmidt:
I'm no ATARI guy but it seems reasonable that KEY_SHIFT should not be part of KEY_UP. Therefore I'm going to merge...

PR #1529, opened 2021-06-01
[cc65] OptStackOps fix

acqn:
This fixed an issue in `OptStackOps` when the pushed TOS is loaded before the matching operation.


oliverschmidt:
No feedback for two days -> merging... (and of course thanks!)

PR #1530, opened 2021-06-01
[cc65] Bit-field fix

### LINKED ISSUE #1462: [cc65] Bit-fields are still broken
It seems to me that bit-field is still broken with '++/--' and '@='...

Demo:
```c
#include <stdio.h>

typedef struct {
    signed int a : 3;
    signed int b : 3;
    signed int c : 3;
} T;

int failures = 0;

void test()
{
    T a = {2, 5, -1};
    T b = {1, 4, -1};
    T m[1] = {{6, 3, -1}};
    T *p = &a;

    a.c += b.a;
    p->c += b.b;
    m->c += b.c;

    if (a.c != -4) {
        ++failures;
    }
    printf("%d\n", a.c);

    if (p->c != -4) {
        ++failures;
    }
    printf("%d\n", p->c);

    if (m->c != -2) {
        ++failures;
    }
    printf("%d\n", m->c);

    ++a.a;
    p->b++;
    m->c--;

    if (a.a != 3) {
        ++failures;
    }
    printf("%d\n", a.a);

    if (p->b != -2) {
        ++failures;
    }
    printf("%d\n", p->b);

    if (m->c != -3) {
        ++failures;
    }
    printf("%d\n", m->c);

    printf("Failures: %d\n", failures);
}

int main(void)
{
    test();
    return failures;
}
```

acqn:
Fixed #1462.
It's possible to do some optimization with the code generation in the future.
Note: _This depends on PR #1529. Without it this will fail in `OptStackOps`._


oliverschmidt:
I'd like to merge this PR but CI says:

```
err/cc65150311-10.s
err/cc65150311-1.s
err/static-2.s
err/cc65150311-7.s
err/void-empty.s
err/cc65150311-8.s
err/pr1110.s
err/bug1098.s
err/duplicate-global.s
err/cc65150311-3.s
err/pr1189.s
1 errors and 2 warnings generated.
err/bss-name-conflict.s
err/bug1145.s
err/cc65150311-11.s
err/front.s
7 errors and 41 warnings generated.
err/cc65091001.s
```

I presume that these errors are just about bugs being fixed. @mrdudz: Maybe you can verify this?

mrdudz:
uhm, the programs in err/ are supposed to error out to begin with... not sure what the actual problem is now *shrug*

mrdudz:
oh yeah, your paste was missing the one important line :)
```
Makefile:41: recipe for target '../../testwrk/err/cc65091001.s' failed
```
will investigate...

@oliverschmidt (as a sidenote, on the travis-ci.org page i see "We're pushing back the shutdown date to June 15th. Please consider migrating to travis-ci.com." - whatever that means, you are aware of this?)

oliverschmidt:
> oh yeah, your paste was missing the one important line :)

Sorry :-(

> will investigate...

Thanks :-)

> [...] you are aware of this?

Sort of ;-) Thanks for reminding me (again).

mrdudz:
It seems to be an actual bug introduced by this PR, the test in question checks and invalid operation like
```
char var = 0xf0;
char fn(char bar)
{
    char* ptr = (char*)0xf;
    var |= ptr; /* should throw an error here */
```
with this PR merged it compiles... which is at least very questionable :) without this PR the compiler generates the error:
```
$ ../../bin/cc65 -o ../../testwrk/err/cc65091001.s cc65091001.c
cc65091001.c:18: Error: Invalid right operand for binary operator '|='
```

oliverschmidt:
Thanks for the analysis!

acqn:
There is another problem here: bit-fields haven't been int-promoted/arithmetic-converted properly. More work is needed.

oliverschmidt:
> There is another problem here

:-(

> More work is needed

Once more: Thanks in advance!

mrdudz:
> bit-fields haven't been int-promoted/arithmetic-converted properly.

how would a testcase for this problem look like? I have to admit that i almost never use bitfields myself, so i need some help here :)

polluks:
@mrdudz I used it once to rearrange pixels.

acqn:
> 
> 
> > bit-fields haven't been int-promoted/arithmetic-converted properly.
> 
> how would a testcase for this problem look like? I have to admit that i almost never use bitfields myself, so i need some help here :)

I added some minimalist test cases for that, and even _fixed_ an existing test case:
https://github.com/cc65/cc65/blob/ed9f9ccbabea58a2e2a688434290d0ad026acd13/test/val/enum-bitfield.c#L71-L78
The test case assumes that the `unsigned` bit-field `e10ubf.x` should be promoted to `unsigned int`, which is not standard-compliant. Since all of its possible values can be represented with a `int`, it is actually supposed to be promoted to be an `int` instead. This latest PR fixed both the compiler and the test case.

acqn:
Hmm, the offset + width detection has a flaw. To be fixed.

oliverschmidt:
So this PR is NOT to be merged at this time, right?

acqn:
Well, the current way cc65 declare bit-fields ensures that their bit-offsets could never excess 8, so that this is perhaps OK for merge as it is now. But maybe I can make this a little more futrue-proof....

oliverschmidt:
I was wondering because you set the 'draft' flag. I can do it either way - just let me know what you think is more appropriate.

acqn:
Well, I think this is OK as it is now.

jmr:
> Well, the current way cc65 handles bit-field declarations ensures that their bit-offsets could never be 8 or bigger, so maybe, this PR is OK for merge as it is now. But perhaps I can make this a little more future-proof.... That won't take too long.

It's like this because the offset is reduced modulo the bit width of the native word size. If 16-bit support is going to be added soon (or already exists and I just don't know about it), then this could be generalized. All the extraction uses 8-bit instructions, so there would be quite a bit more changes necessary.

PR #1540, opened 2021-06-09
[cc65] Fixed result types of certain unary operations broken with #1530

acqn:
PR #1530 broke the result types of certain unary operations.
Now they are fixed and checked with new test cases.

In fact, most of the fix consists of removal of the superfluous type hacks.

### DIFF
--- a/src/cc65/assignment.c
+++ b/src/cc65/assignment.c
@@ -155,7 +155,6 @@ void DoIncDecBitField (ExprDesc* Expr, long Val, unsigned KeepResult)
     unsigned    Mask;
     unsigned    ChunkFlags;
     const Type* ChunkType;
-    const Type* ResType;
 
     /* If the bit-field fits within one byte, do the following operations
     ** with bytes.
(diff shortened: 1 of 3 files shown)


oliverschmidt:
Cool - fixing issues by removing code!

Reminds me of https://youtu.be/MHwmAbKy5ug @ 1:11 :-)

oliverschmidt:
@acqn:

Sorry but I sort of lost track :-(

How do you assess the state of the bit-field code with this this PR applied? Can we consider it fixed? Or do you see more adjustments necessary?

In case you consider the bit-field code okay, are there other things you'd like to see done on your side before we declare the next cc65 revision?

acqn:
I am not quite confident with the bit-field fix and there should always be more tests. I am suspecting potential errors when it involves boolean test but haven't come up with any ideas. Anyway, with regards to the original Issue #1462, I consider it fixed.

The code generated is suboptimal regarding codesize/time, and there are several possible optimization on the current codebase, but most optimization methods lack some code information that would require changing parse/generator. I'm experimenting with them on my private branch too.

PR #1541, opened 2021-06-10
provide PATH_MAX

groessler:


mrdudz:
MMMh i wonder what context this refers to... and wouldnt it have to be system specific too?

groessler:
For example buffer size for getcwd().

256 is the minimum allowed value by POSIX (https://pubs.opengroup.org/onlinepubs/009695399/basedefs/limits.h.html)

mrdudz:
> 256 is the minimum allowed value by POSIX

thanks, then it makes sense (perhaps add a comment that says that?)

groessler:
added a comment

oliverschmidt:
> For example buffer size for getcwd().

I understand that, but... doesn't the user also have the right to expect that he can set paths up to PATH_MAX? I'm wondering if POSIX compliance is more relevant than functionality. To me it would seem much more appropriate to define PATH_MAX to be FILENAME_MAX.

groessler:
Does make sense, too. Yes, I guess functionality is more important than compliance.

I can make the change, and on the way move FILENAME_MAX definition from stdio.h to limits.h. I think it's more appropriate there. Agree?

mrdudz:
Are you implying eg CMD Harddrives can not use a path? making it FILENAME_MAX makes no sense to me at all

groessler:
I think FILENAME_MAX is a misnomer. It's used to set the size of the "cwd" buffer. So it's in fact pathname max.

OTOH, on the Atari at least, the maximum length of a name is determined by the DOS. So, anyway there, one cannot naively use PATH_MAX and expect it to work in all cases. Which would again be an argument to leave PATH_MAX at 256.

oliverschmidt:
> Does make sense, too. Yes, I guess functionality is more important than compliance.

:-)

> I can make the change [...]

:-)

> [...] move FILENAME_MAX definition from stdio.h to limits.h. I think it's more appropriate there.

- That would likely cause regressions for users.
- I blieve that FILENAME_MAX belogns into stdio.h in just the same way that PATH_MAX belongs into limits.h. I think it's most appropriate that one header includes the other.

Just my gut feeling - I haven't read specs:

1. Move https://github.com/cc65/cc65/blob/master/include/stdio.h#L67-L78 into limits.h (or course with FILENAME_MAX replaced with PATH_MAX).
2. Include limits.h from stdio.h.
3. Add `#define FILENAME_MAX PATH_MAX` to stdio.h.

Other thoughts?

mrdudz:
that sounds like it should be something like
```
#if FILENAME_MAX > 256
#define PATH_MAX FILENAME_MAX /* it cant be less than the path */
#else
#define PATH_MAX 256 /* POSIX minimum */
#endif
```

groessler:
Btw., I've noticed that FILENAME_MAX is too small for the Atari when used with a "subdirectory capable" DOS.
Similarly, CBM version appears too small when used with CMD.
I'm talking here about the `cwd` buffer size.

groessler:
we should split these two definitions. FILENAME_MAX seems to be the max lenght of one path element. PATH_MAX is the overall maximum length of a fully qualified path name.

mrdudz:
I have to admit i never actually use FILENAME_MAX - always PATH_MAX :)

still seems sane that PATH_MAX can never be smaller than FILENAME_MAX

oliverschmidt:
1. This for sure isn't a retro-specific problem - see e.g. https://insanecoding.blogspot.com/2007/11/pathmax-simply-isnt.html?m=1

2. "FILENAME_MAX seems to be the max lenght of one path element. PATH_MAX is the overall maximum length of a fully qualified path name." makes of course sense from a semantic POV, but it doesn't fit to given standards.

3. The idea of 2.) doesn't solve the problem of 1.) - in general both retro and modern systems can have different file systems. This is a run-time property, which can't be handled correctly by a compile-time constant. So whatever we come up with here is per definition "wrong" anyhow. Please keep that in mind for the discussion!

oliverschmidt:
> still seems sane that PATH_MAX can never be smaller than FILENAME_MAX

I can't find hints on the net saying that it is appropriate to make any difference between the two whatsoever.

mrdudz:
> > still seems sane that PATH_MAX can never be smaller than FILENAME_MAX
> 
> I can't find hints on the net saying that it is appropriate to make any difference between the two whatsoever.

Whether they can (or have) to be the same is another question... But if PATH_MAX refers to a fully qualified name and FILENAME_MAX to just a part of it, it seems that the logical consequence that the later cant be smaller than the former at least :)

mrdudz:
> Whether they can (or have) to be the same is another question... But if PATH_MAX refers to a fully qualified name and FILENAME_MAX to just a part of it, it seems that the logical consequence that the later cant be smaller than the former at least :)

Forgot - this is why i brought up the "shouldnt it by system specific" part: for CBM DOS the maximum filename lenght is 16, maximum path length is (i think) 256 (it might be smaller, ie limited by the longest string you can send in a DOS command). It seems reasonable to define them as such - for CBM targets that is.

oliverschmidt:
> But if PATH_MAX refers to a fully qualified name and FILENAME_MAX to just a part of it, it seems that the logical consequence that the later cant be smaller than the former at least :)

As I already wrote there no evidence whatsoever that this interpretation of the semantics holds true - or do you have authoriatative links?

mrdudz:
> As I already wrote there no evidence whatsoever that this interpretation of the semantics holds true - or do you have authoriatative links?

? i am just referring to what was written here

groessler:
It's a mess.

I'm now convinced that PATH_MAX should be set to FILENAME_MAX. (In the systems I looked (Linux/FreeBSD), they aren't defined one by the other, but the constants written out directly. But for our target dependent different values it's nicer to define PATH_MAX from FILENAME_MAX).
So I would follow Oliver's suggestion, to include limits.h in stdio.h (btw., on the systems I looked FILENAME_MAX is defined in stdio.h, too).
I'm going to change FILENAME_MAX on the Atari to 63 (max. SpartaDOS path length). I could also make a change for CBM to 256 if you want. Should this be for 'CBM' or do we need to distinguish the different CBM machines?

Agree (except CBM)?
Should I change CBM?

oliverschmidt:
That's the very point. We can make up or own little happy world - or we can look what the rest of the world does. So far I thought we were into the latter, because of user expecations, program portability , <you name it>.

groessler:
How shall I interpret your answer, Oliver?

mrdudz:
Of course if there is a standard for this, then we should stick to it - we dont need to discuss this (i hope)

So, i guess first we need to find out what FILENAME_MAX is supposed to be exactly. if (and _only_ if) it is allowed to be different from PATH_MAX, then we can discuss further what it should be :o)

oliverschmidt:
> It's a mess.

... and always have been ;-)

> I'm now convinced that PATH_MAX should be set to FILENAME_MAX. (In the systems I looked (Linux/FreeBSD), they aren't defined one by the other,

That's my expereince too!

> But for our target dependent different values it's nicer to define PATH_MAX from FILENAME_MAX).

That's waht I thought too.

> So I would follow Oliver's suggestion, to include limits.h in stdio.h (btw., on the systems I looked FILENAME_MAX is defined in stdio.h, too).

:-)

> I'm going to change FILENAME_MAX on the Atari to 63 (max. SpartaDOS path length).

ACK

> I could also make a change for CBM to 256 if you want.

Hard to tell. I see people trying to put a PATH_MAX buffer on the stack... And there are almost never sub directories available...

oliverschmidt:
> How shall I interpret your answer, Oliver?

It was an answer to @mrdudz' comment.

mrdudz:
> And there are almost never sub directories available

actually the sd2iec is quite a common device now, and several filebrowsers for it are written using cc65 :)

groessler:
FILENAME_MAX vs. PATH_MAX:

```
OpenBSD:
/usr/include/stdio.h:#define    FILENAME_MAX    1024    /* must be <= PATH_MAX <sys/syslimits.h> */
/usr/include/sys/syslimits.h:#define    PATH_MAX                 1024   /* max bytes in pathname */

FreeBSD:
/usr/include/stdio.h:#define    FILENAME_MAX    1024    /* must be <= PATH_MAX <sys/syslimits.h> */
/usr/include/sys/syslimits.h:#define    PATH_MAX                 1024   /* max bytes in pathname */

Linux:
/usr/include/x86_64-linux-gnu/bits/stdio_lim.h:#define FILENAME_MAX 4096
/usr/include/linux/limits.h:#define PATH_MAX        4096        /* # chars in a path name including nul */
```

Somebody should check Windows.

In all examples the defines are the same.

groessler:
> And there are almost never sub directories available...

The "almost" is the crucial part. Since if they _are_ available, and cwd happens to be longer than 16, the program will crash since it overwrites the buffer.

mrdudz:
```
/* must be <= PATH_MAX <sys/syslimits.h> */
```
hear hear :)

mrdudz:
i found this
```
But be careful, FILENAME_MAX might be not as large as you might expect,
e.g. on my system (HP-UX B.11.00 A 9000/712) FILENAME_MAX (from <stdio.h>)
equals 14 (that's fourteen).
```
amazing what a rabbit hole this question is. funky :)

oliverschmidt:
Indeed :-)

mrdudz:
also
```
FILENAME_MAX

which expands to an integer constant expression that is the size
needed for an array of char large enough to hold the longest
file name string that the implementation guarantees can be opened;
222)
____________________

222) If the implementation imposes no practical limit on the
length of file name strings, the value of FILENAME_MAX
should instead be the recommended size of an array intended
to hold a file name string. Of course, file name string
contents are subject to other system-specific constraints;
therefore all possible strings of length FILENAME_MAX cannot
be expected to be opened successfully.

**Note that the C standard is blind to the concept of path, therefore
the path is implicitly part of the file name.**
```

groessler:
I've made the changes and left CBM alone.

mrdudz:
The more i read the more i am also convinced that FILENAME_MAX should indeed refer to the fully qualified filename, and making it equal to PATH_MAX is the way to go.

oliverschmidt:
> The more i read the more i am also convinced that FILENAME_MAX should indeed refer to the fully qualified filename, and making it equal to PATH_MAX is the way to go.

That's what I came up with too when I did that excerise in my work life.

So I think it's safe to say that this PR as it stands for sure doesn't make anything worse -> merging...

mrdudz:
> I've made the changes and left CBM alone.

make it 256

groessler:
> make it 256

For all CBMs (`#ifdef __CBM__`) or just selected machines?

mrdudz:
> For all CBMs (#ifdef __CBM__) or just selected machines?
I admit i dont know how the harddrives for IEEE machines work. However, since you actually can connect IEC devices to those too (with interfaces), doing it for all CBM machines is probably fine

groessler:
done

groessler:
@mrdudz, do there 256 bytes include the terminating 0?

I've set the CBM value now to 255. See `libsrc/common/_cwd.s`. It claims some runtime lib code cannot handle values > 255. Although I think 256 should be ok. And the comment is inconsistent:

```
; NOTE: Some of the code working with directories is not able to handle
; strings longer than 255 chars, so don't make __cwd larger than 256 without
; checking the other sources.
```

Shouldn't it be `so don't make __cwd larger than 2565 without checking the other sources.`?

mrdudz:
255 is fine, iirc also in CBM DOS the terminating character is included in those 256

groessler:
With this CBM change to 255, the CI test fails with 

```
$ ./bin/cc65 -Ors --codesize 500 -T -g -t c64 samples/enumdevdir.c
samples/enumdevdir.c:30: Error: Too many local variables
samples/enumdevdir.c:35: Error: Too many local variables
samples/enumdevdir.c:86: Error: Too many local variables
samples/enumdevdir.c:91: Error: Too many local variables
4 errors and 0 warnings generated.
```

Any ideas? There are now not "more" local variables, but "bigger" local variables. The sample puts two (three altogether) local vars like this on the stack:

```
    char olddir[FILENAME_MAX];
    char curdir[FILENAME_MAX];
```

Is there some stack size check in the compiler?

mrdudz:
the total size of local variables is limited to 255 - the fix is to make them global (or static)

groessler:
you mean "space for local variables", aren't you?

groessler:
> the fix is to make them global (or static)

This doesn't work in this example. The `printdir()` function calls itself recursively, and on quick glance at least a static/global `olddir` would be destroyed.

The vars need to be allocated from the heap I guess.

mrdudz:
> The vars need to be allocated from the heap I guess.

in that case, yes, sounds like it.

groessler:
I made the change (untested, please test).

oliverschmidt:
>> I see people trying to put a PATH_MAX buffer on the stack...

> Error: Too many local variables

;-))

samples/enumdevdir.c
```diff
@@ -16,31 +16,47 @@
...
+    char *curdir;
     DIR *dir;
     struct dirent *ent;
     char *subdirs = NULL;
     unsigned dirnum = 0;
     unsigned num;
 
-    getcwd (olddir, sizeof (olddir));
+    olddir = malloc (FILENAME_MAX);
+    if (olddir != NULL) {
```
oliverschmidt, line 30: This seems indeed untested. As far as I know it needs to be `==` instead of `!=`.

samples/enumdevdir.c
```diff
@@ -16,31 +16,47 @@
...
+
```
oliverschmidt, line 31: When another commit is necessary anyway please remove this blank line.

samples/enumdevdir.c
```diff
@@ -16,31 +16,47 @@
...
         ** directory name - as done for files.
         */
         printf ("  Dir  %s\n", newdir);
-        return;
+        free (olddir);
+        return 0;
+    }
+
+    curdir = malloc (FILENAME_MAX);
+    if (curdir != NULL) {
```
oliverschmidt, line 48: Same as above.

samples/enumdevdir.c
```diff
@@ -16,31 +16,47 @@
...
+
```
oliverschmidt, line 49: Same as above.

samples/enumdevdir.c
```diff
@@ -65,18 +81,28 @@ void printdir (char *newdir)
...
 
 
 void main (void)
 {
     unsigned char device;
-    char devicedir[FILENAME_MAX];
+    char *devicedir;
+
+    devicedir = malloc (FILENAME_MAX);
+    if (devicedir != NULL) {
```
oliverschmidt, line 101: Same as above.

samples/enumdevdir.c
```diff
@@ -65,18 +81,28 @@ void printdir (char *newdir)
...
+
```
oliverschmidt, line 102: Same as above.

samples/enumdevdir.c
```diff
@@ -16,31 +16,47 @@
 #include <cc65.h>
 
 
-void printdir (char *newdir)
+int printdir (char *newdir)
```
oliverschmidt, line 19: Can you please make use of https://github.com/cc65/cc65/blob/master/include/stdbool.h ?
Success -> true
Failure -> false

oliverschmidt (changes requested):
Thanks for taking care if thingy starting with a single line in limits.h ;-)

samples/enumdevdir.c
groessler: Hmm. I added this blank line since after the `if (chdir (newdir))` there is one.

samples/enumdevdir.c
groessler: Stupid...

groessler:
did the changes

oliverschmidt (approved):

oliverschmidt:
Thanks :-)

PR #1546, opened 2021-07-05
Implement digit separators

Nufflee:
Digit separators are a C++14 feature added to cc65 as an extension. They're a generally useful feature for separating thousands in decimal and bit groups in binary literals.

Example:
```c
unsigned char foo = 0b11'00'11;
unsigned int bar = 42'000;
```

I implemented the feature together with tests and documentation.
The feature is implemented on the preprocessor level meaning that the digit separator characters (`'`) are removed before the code is scanned and parsed as the compiler doesn't need to be aware of their existance.

### DIFF
--- a/doc/cc65.sgml
+++ b/doc/cc65.sgml
@@ -853,6 +853,16 @@ This cc65 version has some extensions to the ISO C standard.
         unsigned char foo = 0b101; // sets it to 5
         </verb></tscreen>
 
+<item>  Digit separators in literals, a C++14 feature, are accepted. They are
+        often used to separate thousands in decimal or bit groups in binary literals.
+        They can be disabled with the <tt><ref id="option--standard"
+        name="--standard"></tt> option.
+
+        <tscreen><verb>
+        unsigned char foo = 0b11'00'11'00; // sets foo to 0b11001100
+        unsigned int bar  = 42'069;        // sets bar to 42069
+        </verb></tscreen>
+
 </itemize>
 <p>
 
--- a/src/cc65/preproc.c
+++ b/src/cc65/preproc.c
@@ -914,6 +914,12 @@ static unsigned Pass1 (StrBuf* Source, StrBuf* Target)
                 SB_AppendStr (Target, Ident);
             }
         } else if (IsQuote (CurC)) {
+            /* Remove digit separators between two digits. This is a C++14 feature supported as an extension in the CC65 standard mode. */
+            if (IS_Get (&Standard) >= STD_CC65 && (IsDigit (SB_LookAtLast (Target)) && IsDigit (NextC))) {
+                NextChar ();
+                continue;
+            }
+
             CopyQuotedString (Target);
         } else if (CurC == '/' && NextC == '*') {
             if (!IsSpace (SB_LookAtLast (Target))) {
--- a/test/val/digit-separator.c
+++ b/test/val/digit-separator.c
@@ -1,0 +1,13 @@
+#define DO_TEST(actual, expected) \
+    if (actual != expected) { \
+        return 1; \
+    }
+
+int main() {
+    DO_TEST(1'2'3'4'5, 12345);
+    DO_TEST(0x1'2'3'4, 0x1234);
+    DO_TEST(0'1'2'3'4, 01234);
+    DO_TEST(0b1'0'1'1'0'1'0'0, 0b10110100);
+
+    return 0;
+}


mrdudz:
> C++14

the what now? *shrug*

Nufflee:
The C++14 standard, just like the binary literals feature I derived this from. This is not a GNU C extension unlike the binary literals but it should be trivial and useful enough otherwise.

greg-king5:
What about hexadecimal letters?
```c
var = 0xd'eed; // should work
var = d'eed; // should this fail?
```

mrdudz:
Can you point to a formal description of how this is supposed to work? Can i basically put single quotes anywhere in any numeric constant? I find this very odd to be honest :)

spiro-trikaliotis:
@mrdudz, an "almost" formal description can be found here: 
https://en.cppreference.com/w/cpp/language/integer_literal

And, yes, it seems you can add the separators wherever you want. I find it odd, too.

Nufflee:
Thank you everyone for taking a look!

@greg-king5 hexadecimal letters should be supported, that is my fault for not making sure `IsDigit` looks for them. I should use `IsXDigit` instead. As for the second remark, you're right and the `'` character should not be removed, even though it will be removed by my code. I'm not sure what the best solution for this would be. Either the preprocessor would have to validate that the current token is a valid literal, or this should be delegated to the scanner. The issue in the latter case is that `CopyQuotedString` would run which would result in inline comments being sent to the scanner as well and I'm not sure how to avoid that.

@mrdudz I worked off of the formal description from cppreference linked by @spiro-trikaliotis, but [here is the most formal description from the C++14 standard](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2013/n3781.pdf).

Nufflee:
I checked what the gcc C++ preprocessor does in this case and it seems like it validates the literal, and in case it is invalid and contains a digit separator, it emits a warning and copies the rest of the input as if it were a string.

`invalid-literal.cpp`:
```c
int foo = d'eed; // comment
```

```
$ g++ -E invalid-literal.cpp

...

invalid-literal.cpp:1:12: warning: missing terminating ' character
    1 | int foo = d'eed; // comment
      |            ^
int foo = d'eed; // comment
```

...and in the valid identifier case it emits the expected result.
`valid-identifier.cpp`:
```c
int foo = 0xd'eed; // comment
```

```
g++ -E valid-identifier.cpp

...

int foo = 0xd'eed;
```

In conclusion, having the preprocessor validate literals in this case is not unseen and I could implement it if everyone else agrees.

groessler:
Sorry, does anybody really need this?

Nufflee:
Sorry, does anybody really need C(++)?

oliverschmidt:
I think the benefit of a PR needs to be balanced with the amount/complexity of new code added and the number of regressions envisioned.

The benefit of this PR is likely by most considered very small, but on the other hand, it consists of only three lines of actual code...

polluks:
We are a serious project, no fight for fame like C++14 =)

Nufflee:
@oliverschmidt I completely agree with you and I appreciate you engaging in a constructive discussion unlike many others. In fact @mrdudz *literally* edited my comment responding to concers raised above, completely removing it and replaced it with something of his own. They don't seem to be a collaborator to the project so I have no idea why they have comment edit privileges. This is completey unacceptable in an open source project and I do appreciate all of the work everyone has put into cc65 but I will personally never even consider contributing to this project again. 
I was just trying to make an honest contribution addressing a little thing I saw missing but I ended up being made fun of and got my comments constructively addressing issues raised effectively deleted. This is a toxic environment and unlike anything I'm looking for when spending a significant amount of my free time understanding the codebase and implementing a feature for everyone to use.

image

mrdudz:
> In fact @mrdudz _literally_ edited my comment responding to concers raised above, completely removing it and replaced it with something of their own. They don't seem to be a collaborator to the project so I have no idea why they have comment edit privileges. 

I am really sorry, i don't know what happened there. I just wanted to "quote and reply", not edit your comment at all. (And i honestly dont need nor want the permission to edit other peoples comments, i wonder why its even possible)

mrdudz:
> @mrdudz I worked off of the formal description from cppreference linked by @spiro-trikaliotis, but [here is the most formal description from the C++14 standard](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2013/n3781.pdf).

Thanks, the example given at the bottom is enough of a reason for me to rather not have this extension.

Nufflee:
@mrdudz I'm not here to argue over this at all. I do not want to continue working on this PR but if anyone else sees any value in it, I'd be glad to have it taken over and finished up by them. No need to reopen it once I've closed it though.

mrdudz:
> @mrdudz I'm not here to argue over this at all.

Nor am i, i can only say sorry again - it wasnt my intend to edit your comment. Entirely my fault.

groessler:
> but you can still see this malicious edit in the "edited" dropdown (and in the screenshot above)

Probably it doesn't really show, maybe blame github for this, but I don't see any "malicious edit" in the screenshot. I'm seeing two line deleted, and replaced with two lines with the same content.

oliverschmidt:
Let me add two - final - remarks to this:

* In this repo nobody has ever edited somebody else's comments intentionally (aside from adding clarifications to the opening comment of an issue). @mrdudz made it pretty clear that this wasn't the case here too.

* So far there seems to be no interest in this PR. @Nufflee explicitly asked to not reopen it for potential further discussion. Apart from that, as always ANY interest in contributing to cc65 is appreciated! Maybe this attitude hasn't become clear in all comments ;-)

oliverschmidt:
And a personal remark: This PR came perfectly formatted and with test right from the start so I could have merged it without any additional review cycles - thanks :-)

Nufflee:
@oliverschmidt As I said, I always appreciate constructive discussion and criticism so I will address this one last thing. I don't want to be involved with this PR any more so I closed it, but I can certainly open a feature request issue for this and link this PR in for further discussion.

> And a personal remark: This PR came perfectly formatted and with test right from the start so I could have merged it without any additional review cycles - thanks :-)

The issue with this approach is that my tests were flawed - they didn't contain any hexadecimal letters and my implementation introduced a bug/regression when there is a `'` character between two (hexadecimal) digits in an identifier as I explained above. A (additional) review cycle is certainly very important to bring up and address issues like these.

Regarding the edit situation, I would again suggest you to disable edit permissions for non-collaborators (ie. not members of the cc65 organization) so things like this cannot happen again.

oliverschmidt:
> Regarding the edit situation, I would again suggest you to disable edit permissions for non-collaborators (ie. not members of the cc65 organization) so things like this cannot happen again.

@mrdudz certainly doesn't fall into this category.

Nufflee:
> @mrdudz certainly doesn't fall into this category.

I'm not sure what you mean, but they show up as a "Contributor" as opposed to you who shows up as a "Collaborator". But anyway, that is the least important part of my comment.

PR #1558, opened 2021-09-08
saves 2 bytes in the standard c64 joystick driver

gorlik:
remove redundant code and add jmp to the common sequence
two more bytes could be saved at the expense of longer sequence with interrupts disabled by moving sei/cli


mrdudz:
you could save one more byte with zero impact by moving the "end" label to the cli and removing the cli before the jmp :)

gorlik:
Yes. However, what you suggest will make the interrupt disabled sequence 3 cycles longer in the joy1 path
I opted to favor interrupt disabled time over extra size savings.

mrdudz:
Uh, yes. i'd really favour a saved byte here though, those 3 cycles won't matter :=)

Issue #1556, opened 2021-08-30
[Feature] Output struct field information in dbgfile or something


empathicqubit:
I've been trying to write a debugger to connect to VICE, but I was having some trouble getting metadata on struct fields. Right now I'm using clang-query to parse through the preprocessed code files, which works kind of, but it's a really ugly hack and clang is a huge dependency. Obviously after the compile stage the struct information gets optimized out. It would be nice if I could get the names and types of the struct fields from the dbgfile or from `-d`. I was sorting through how to add this to cc65 and found the appropriate entries in the symbol table, but I wanted to get an opinion on how it should be done before I do it, or if I've completely missed something obvious.

empathicqubit:
Something naive such as this under cc65/symtab.c->LeaveGlobalLevel:

```c
        SymEntry* Entry = TagTab0->SymHead;
        do {
            if(!(Entry->Flags & SC_STRUCT) || !Entry->V.S.SymTab) {
                continue;
            }
            PrintSymTable(Entry->V.S.SymTab, stdout, Entry->Name);
        } while((Entry = Entry->NextSym));
```
Spits out useful data, but there's probably a better place for it. Stuffing it in the debug output means that it would only appear once when using `make`.

empathicqubit:
So I have one idea that wouldn't be too disruptive to the existing code: Add something similar to the above, and allow redirection of the table dumps to a specific file if requested using a switch.

PR #1560, opened 2021-09-10
Add --debug-tables <filename> option and output struct fields

empathicqubit:
Add an option to output symbol table information to a file instead of stdout. Add struct field and function scoped symbol information to output. This is merely an example, and should probably be thought about more thoroughly.

#1556


empathicqubit:
I've been using this for a little while in another project without any issues since the last force-push, so I think it's ready to be looked at, at least.

mrdudz:
It'd really be great if you'd also add documentation for this switch, and the format of the info it produces

empathicqubit:
Okay, I added the switch along with an example to cc65.html

empathicqubit:
Is there anything further that needs to be done with this? I'm not sure what the test failure is, since I didn't really change the compile behavior, only added logging.

mrdudz:
i think that is the test-fails-randomly issue that was fixed, i will close/reopen for a test

PR #1564, opened 2021-09-22
VIC-20 and C128 cfg added for ASM programming

skeetor:
Added config file for ASM for VIC-20 and C128.

### DIFF
--- a/cfg/c128-asm.cfg
+++ b/cfg/c128-asm.cfg
@@ -1,0 +1,20 @@
+FEATURES {
+    STARTADDRESS: default = $1c01;
+}
+SYMBOLS {
+    __LOADADDR__: type = import;
+}
+MEMORY {
+    ZP:       file = "", start = $0002,  size = $00FE,      define = yes;
+    LOADADDR: file = %O, start = %S - 2, size = $0002;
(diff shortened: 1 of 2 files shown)


mrdudz:
perhaps the vic20 config could have a bunch of comments on how to modifiy it for some common memory configs? or maybe even add a vic20-asm-8k.cfg at least? mmh

skeetor:
> perhaps the vic20 config could have a bunch of comments on how to modifiy it for some common memory configs? or maybe even add a vic20-asm-8k.cfg at least? mmh
Problem is, that I don't really understand the details of the cfg file. So I just added a simple version, which I looked in the c64 file for reference.

oliverschmidt:
I'd (obviously) prefer the style I used for the Apple II target: https://github.com/cc65/cc65/blob/master/cfg/apple2.cfg

A lot of variables with default values. This:
- Allows for easy usage for standard use-cases
- Enables the user to move away from standard use-cases
- Serves as documentation, what are things useful to adjust
- Doesn't spam the cfg file directory with loads of derivates

The variables are a nice anchor point for documentation too: https://cc65.github.io/doc/apple2.html#s4

PR #1599, opened 2021-11-28
VIC-20 TGI needs 16K to be usable

ops:
With 8K expansion tgidemo can't be compiled. I think it is reasonable to require 16k expansion.

### DIFF
--- a/cfg/vic20-tgi.cfg
+++ b/cfg/vic20-tgi.cfg
@@ -1,5 +1,5 @@
 # Memory configuration which supports the "vic20-hi.tgi" driver.
-# Memory configuration for a VIC-20 with, at least, 8K expansion RAM.
+# Memory configuration for a VIC-20 with, at least, 16K expansion RAM.
 FEATURES {
     STARTADDRESS: default = $1201;
 }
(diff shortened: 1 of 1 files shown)


greg-king5:
> With 8K expansion, tgidemo can't be compiled. I think it is reasonable to require 16k expansion.

I think that it's reasonable _to require_ the _minimum_ size that's needed for graphics.  When a programmer needs more, that one can set it on a command line.  Look at what the samples makefile does to build its TGI programs on the VIC-20:
https://github.com/cc65/cc65/blob/1d36f255ecf401f1c69ce5f337b936fd0cf08154/samples/Makefile#L362-L374

elderling:
I agree that  the sample TGI linker file should remain at 8K. That being said, I think that there should be a less-obscure path to successfully compiling tgidemo for the VIC-20. Linking, especially with the VIC-20, is tricky for new (and sometimes seasoned) users. When I first tried to make tgidemo work for the VIC-20 I thought it was just broken until I started experimenting with linker settings.

mrdudz:
The sensible thing here is:
- put a comment into the linker config explains that you might need more memory
- document this accordingly

ops:
> The sensible thing here is:
> 
>     * put a comment into the linker config explains that you might need more memory
> 
>     * document this accordingly

I'd like the other way around:

- increase memory in the default config
- document that if the program is trivial, 8k may be enough (add D __HIMEM__=0x4000)

WayneParham:
I personally like having a minimum configuration but also an easy way to show how to increase it.  So I lean towards the documentation approach.  I also like having samples in the Makefile in targettest that show examples of each configuration.

mrdudz:
> I personally like having a minimum configuration but also an easy way to show how to increase it. So I lean towards the documentation approach. I also like having samples in the Makefile in targettest that show examples of each configuration.

Exactly this, the documentation should be changed/made more clear (perhaps also the tgidemo source needs a comment like this) - not the default config changed

PR #1620, opened 2021-12-24
Add standard C library function strnlen().

SvenMichaelKlose:


polluks:
Please take care "standard" means POSIX...

libsrc/common/strnlen.s
```diff
@@ -0,0 +1,54 @@
+; size_t __fastcall__ strnlen (const char* s, size_t maxlen);
+
+.export     _strnlen
+.import     popax
+.importzp   ptr1, tmp1, tmp2, tmp3, tmp4
+
+.proc _strnlen
+        ; Fetch string pointer.
+        sta     ptr1
+        stx     ptr1+1
```
greg-king5, line 10: You have the parameters backwards.  This is the _last_ one, `maxlen`.

libsrc/common/strnlen.s
```diff
@@ -0,0 +1,54 @@
...
+
+        ; Get maximum length.
+        jsr     popax
+        sta     tmp3
+        inc     tmp3
+        inx
+        stx     tmp4
+
+        ;;; Loop over string.
+        ldy     #0
```
greg-king5, line 25: You can save a couple of bytes by using this .Y value to clear the function's return value.

libsrc/common/strnlen.s
```diff
@@ -0,0 +1,54 @@
...
+        bne     l2
+        dec     tmp4
+        beq     done
+l2:
+
+        lda     (ptr1),y
+        beq     done
+
+        ; Step to next character.
+        inc     ptr1
```
greg-king5, line 38: Save a byte here by incrementing the .Y index instead of the pointer's low byte.

libsrc/common/strnlen.s
```diff
@@ -0,0 +1,54 @@
...
+        bne     l1
+        inc     ptr1+1
+l1:
+
+        ; Increment return value.
+        inc     tmp1
+        bne     next
+        inc     tmp2
+
+        jmp     next
```
greg-king5, line 48: At this point, `tmp2` won't be zero.  Therefore, you can use `bne` instead of the longer `jmp`.

include/string.h
```diff
@@ -50,6 +50,7 @@ char* __fastcall__ strcpy (char* dest, const char* src);
 size_t __fastcall__ strcspn (const char* s1, const char* s2);
 char* __fastcall__ strerror (int errcode);
 size_t __fastcall__ strlen (const char* s);
+size_t __fastcall__ strnlen (const char* s, size_t maxlen);
```
greg-king5, line 53: @polluks is right.  Please move this line down to the "Non-standard" section of the header.

greg-king5 (changes requested):

libsrc/common/strnlen.s
SvenMichaelKlose: Oh, damn! Was under the impression that this code was already in use at my place. Sorry.

include/string.h
SvenMichaelKlose: It's conforming to POSIX.1-2008. Still sure?

greg-king5 (approved):

mrdudz:
@greg-king5 in clear cases like this, feel free to just merge :)

PR #1621, opened 2021-12-25
Fix VIC build, remove compiler warnings, add LST files for VICE emulator targets.

SvenMichaelKlose:


targettest/heaptest.c
```diff
@@ -6,7 +6,7 @@
 
 
 
-static unsigned char* V[256];
+static char* V[256];
```
greg-king5, line 9: This program uses binary bytes, not text characters.  Those type declarations should be `unsigned char`, not `char`.  "heaptest.c" shouldn't be changed.

greg-king5: OK, I was wrong.  The program does need to be fixed.  But, it's the `Alloc()` function's declaration that's wrong.

SvenMichaelKlose: No clue what's wrong with Alloc(), I'm afraid.

targettest/gamate/ctest.c
```diff
@@ -9,6 +9,9 @@ unsigned short n;
 
 int main(int argc, char *argv[])
 {
+    (void) argc;    // Suppress warnings.
```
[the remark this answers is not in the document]
SvenMichaelKlose: Compiler rejects a void return value of main().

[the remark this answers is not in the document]
greg-king5: Remove the `return 0;` at the end of the function.

targettest/heaptest.c
greg-king5: `Alloc()` wants to return the value of `P`.  Therefore, its return type must match the type of `P`.  The first warning says that it doesn't match.  `P` is correct; therefore, it's the return type that needs fixing.

targettest/Makefile
```diff
@@ -386,7 +424,7 @@ testcode.atr: testcode
 # Clean-up rules
 
 mostlyclean:
-	@$(DEL) *.lbl *.map *.o 2>$(NULLDEV)
+	@$(DEL) *.lbl *.map *.lbl *.o 2>$(NULLDEV)
```
greg-king5, line 427: `*.lbl` already is there!

targettest/gamate/ctest.c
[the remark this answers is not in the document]
SvenMichaelKlose: Yessir.

targettest/heaptest.c
SvenMichaelKlose: Am confused. Am not getting any warnings.

greg-king5: That's because you changed some things.  You need to go back to the original repository version:
```
git restore heaptest.c
```
Then, you'll see five warnings when you compile "heaptest.c".  The first warning is the important one; the others are only consequences of the first one.  It points to https://github.com/cc65/cc65/blob/6ac4aa4e2076189ad5b5f3a9d33e1c5910d4e37c/targettest/heaptest.c#L31
It says that the `return` and the `P` have different pointer types.  The type of the `return` is declared by https://github.com/cc65/cc65/blob/6ac4aa4e2076189ad5b5f3a9d33e1c5910d4e37c/targettest/heaptest.c#L13
The bug is that an `unsigned` is missing from that line.

SvenMichaelKlose: So I guess I'd have been easier to fix oneself than to write a ticket. I'm out.

mrdudz:
looks good enough for me :)

polluks2:
> looks good enough for me :)

See https://stackoverflow.com/questions/7656425/makefile-ifeq-logical-or

mrdudz:
feel free to provide further improvement in a PR

PR #1622, opened 2021-12-29
modify listing to include file information for better correlating the…

ebadger:
… relative addresses with the --dbgfile output

I've built a 6502 virtual machine for my home brew hardware project.
I'm adding debug functionality to my emulator, and am working to get source line debugging.
i.e. while stepping through instructions, the assembly source is displayed with a line indicator.
Current --dbgfile output is not sufficient.   By adding file info to the ca65 -l  output, I can use the listing file in conjunction with --dbgfile output to resolve the source code.

### DIFF
--- a/src/ca65/listing.c
+++ b/src/ca65/listing.c
@@ -53,7 +53,6 @@
 #include "segment.h"
 
 
-
 /*****************************************************************************/
 /*                                   Data                                    */
 /*****************************************************************************/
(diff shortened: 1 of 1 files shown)


mrdudz:
as also said in #1560 - it'd be really great if you'd add some documentation about this (yes i realize there is none yet - someone has to start :))

mrdudz:
Mmmh this PR also seems to contain target support for some custom target? *shrug*

mrdudz:
ok so, this is really too much of a hack - the debug info should get updated instead (if it isn't already sufficient)

and although the homebrew target is interesting (@ebadger please still post some info about it!) it is a bit too obscure to include in the main branch.

so.. closing

PR #1623, opened 2021-12-31
Save two bytes

polluks2:


mrdudz (changes requested):
please add a comment that this is a BIT abs

PR #1626, opened 2022-01-07
Fix Pop() implementation in src/sim65/paravirt.c (fixes #1625)

### LINKED ISSUE #1625: Paravirt functions not returning properly
I'm using `sim65` V2.18 on macOS (installed via [`homebrew`](https://formulae.brew.sh/formula/cc65)).  Whenever I `jsr` to a paravirt function (other than `PVExit`), the paravirt function works, but then doesn't seem to return to the correct address.  `sim65` eventually terminates with something like `Error: Illegal opcode $03 at address $10000`, presumably after the PC wanders off into oblivion.

Looking at [paravirt.c](https://github.com/cc65/cc65/blob/5d84a4ba137e32163390ccd058c32ad169b4ca94/src/sim65/paravirt.c), I suspect the problem is with the last line of `ParaVirtHooks()`:

```c
/* Simulate RTS */
Regs->PC = Pop(Regs) + (Pop(Regs) << 8) + 1;
```

If I'm not mistaken, evaluation order of arguments within an expression like that is unspecified in C, and I think what's happening is that on my machine/OS, the two `Pop()` calls are happening in the wrong order, leading to setting the PC to e.g. `$3412` instead of `$1234`.  If my guess is correct, then I believe the fix would be to force the correct pop order by changing that line to something like:

```c
/* Simulate RTS */
unsigned lo = Pop(Regs);
unsigned hi = Pop(Regs);
Regs->PC = lo + (hi << 8) + 1;
```

mdsteele:
The Pop() function was not handling stack pointer wrap around correctly.

Also, change the simulated RTS implementation in ParaVirtHooks() to
explicitly sequence the two Pop() calls in the correct order.

### DIFF
--- a/src/sim65/paravirt.c
+++ b/src/sim65/paravirt.c
@@ -105,7 +105,7 @@ static void SetAX (CPURegs* Regs, unsigned Val)
 
 static unsigned char Pop (CPURegs* Regs)
 {
-    return MemReadByte (0x0100 + ++Regs->SP);
+    return MemReadByte (0x0100 + (++Regs->SP & 0xFF));
 }
 
 
@@ -327,5 +327,7 @@ void ParaVirtHooks (CPURegs* Regs)
     Hooks[Regs->PC - PARAVIRT_BASE] (Regs);
 
     /* Simulate RTS */
-    Regs->PC = Pop(Regs) + (Pop(Regs) << 8) + 1;
+    unsigned lo = Pop(Regs);
+    unsigned hi = Pop(Regs);
+    Regs->PC = lo + (hi << 8) + 1;
 }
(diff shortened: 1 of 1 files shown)


src/sim65/paravirt.c
```diff
@@ -327,5 +327,7 @@ void ParaVirtHooks (CPURegs* Regs)
     Hooks[Regs->PC - PARAVIRT_BASE] (Regs);
 
     /* Simulate RTS */
-    Regs->PC = Pop(Regs) + (Pop(Regs) << 8) + 1;
+    unsigned lo = Pop(Regs);
+    unsigned hi = Pop(Regs);
+    Regs->PC = lo + (hi << 8) + 1;
```
greg-king5, line 332: Don't overdo the fix.  There needs to be a sequence point _between_ the two `Pop()s`.  Therefore, only the first one needs to be outside the PC expression.

polluks: BTW this is C99 style.

mdsteele: Done.

mrdudz:
Merging, thanks!

PR #1633, opened 2022-01-23
document special case on sub 40-col modes

sprive:


sprive:
Documents unexpected behavior that's not a bug

Issue #1570, opened 2021-09-27
[nes] background colors seem wrong (white is grey, red is violet, etc.)


Fabrizio-Caruso:
Background colors seem wrong mapped (white shows grey, red is violet, etc.)
If I compile
```
#include<conio.h>

int main()
{
    bgcolor(COLOR_WHITE);
    clrscr();

    gotoxy(2,2);
    cprintf("hello world\n");
    while(1){};
    return 0;
}
```
I get:

![image](https://user-images.githubusercontent.com/29057615/134865129-4bf4b4dc-65e4-435b-812f-48c8788635b2.png)

whereas with `bgcolor(COLOR_RED)` I get a violet background.

polluks:
Indeed, the colours of libsrc/nes/color.s do not match very well
https://nesdoug.com/2018/09/05/05-palettes/

PR #1644, opened 2022-01-31
Added more purple (rain)

polluks2:
NES was missing TGI colors also.


mrdudz:
IMHO it'd make more sense to either remove VIOLET or PURPLE, whats the point of having both?

Fabrizio-Caruso:
Could it be possible to include a fix for white shown as grey and red shown as purple?
Issue: #1570

polluks:
> IMHO it'd make more sense to either remove VIOLET or PURPLE, whats the point of having both?

Maybe it's too late, we already have both. I don't like redundancy either.

mrdudz:
> Maybe it's too late, we already have both.
That is not a good reason for not cleaning it up (and even less for polluting the code even further)

Fabrizio-Caruso:
Purple and violet are different colors even though they may look similar to the human eye.
Purple is reddish because it is red and blue whereas violet is a pure spectral color with no red in it.
On many 8 bit systems you only get one of the two. So their associated macros may coincide if you want to approximate them with a limited range of displayable colors.
The NES seems to have both (the reddish/pinkish one is purple and not violet).

mrdudz:
> Purple and violet are different colors even though they may look similar to the human eye.

The same applies to most other colors defined by those macros

However, we have a portable API there which tries to provide 16 distinct colors for the most part

> Purple is reddish because it is red and blue whereas violet is a pure spectral color with no red in it.

That entirely depends how you look at it. On a color circle (and in terms of RGB) both are mixtures of red and blue

Marrin:
As all the names (but one) and the order in which they are defined are the names of the C64 colours as listed in the handbook, I'd suggest VIOLET is the one to go if there should be only one, because in the C64 documentation that colour is called PURPLE.  And also in the abbreviation on the keyboard. :-)

polluks:
> That is not a good reason for not cleaning it up (and even less for polluting the code even further)

This is a consistency clean-up =)

mrdudz:
> I'd suggest VIOLET is the one to go if there should be only one, because in the C64 documentation that colour is called PURPLE. 

i agree

mrdudz:
Looks good to me now - are you sure the NES changes are actually correct? (the TGI colors may not necessarily match the CONIO colors)

mrdudz:
> are you sure the NES changes are actually correct?

@polluks i can only repeat that question. Also please fix the conflicts so it can be merged :)

PR #1648, opened 2022-02-02
Added tail call optimization

polluks2:


WayneParham:
Verified on a Sym-1 system.  I built and tested binaries created with a version of the cc65 compiler that has this change.  Ran compiled binaries that executed the code in these changes.  All works.

Good catch, Stefan!  Nice little change!

polluks2:
Maybe the assembler should warn about TCO.

mrdudz:
feel free to implement it and make a PR :)

PR #1654, opened 2022-02-07
Move the platform-test enumerations from Github Actions to makefiles.

greg-king5:


samples/Makefile
```diff
@@ -340,7 +340,7 @@ endif
 
 define SUBDIR_recipe
 
-@$(MAKE) -C $(dir) --no-print-directory $@
+@+$(MAKE) -C $(dir) --no-print-directory $@
```
mrdudz, line 343: Why the + ?

targettest/Makefile
```diff
@@ -699,7 +699,7 @@ endif
 
 define SUBDIR_recipe
 
-@$(MAKE) -C $(dir) --no-print-directory $@
+@+$(MAKE) -C $(dir) --no-print-directory $@
```
mrdudz, line 702: Again, why the + ?

samples/Makefile
greg-king5: It blocks the `-j2` in subdirectories.  Without that `+`, we get a warning message:  `Jobserver unavailable. Use '+' in parent makefile rule.`

PR #1661, opened 2022-02-11
Fixed makefiles with cmd.exe

acqn:
Only tested with `mingw32-make` under `cmd.exe`.


mrdudz:
That test failure is surprising to me *shrug* how did it pick up CC wrong there?

mrdudz:
ehrm now i understand. these tools are not ment to be compiled with cc65 - these are utilities that run on the host, not on the target

PR #1662, opened 2022-02-11
[cc65] Struct fix

acqn:
Fixed the regression:
```c
typedef int field_type;

struct
{
    field_type : 1; /* ICE */
};
```


mrdudz:
A simple test for this would be nice

PR #1675, opened 2022-02-17
[cc65] Fixed bitwise shift with numeric constant operand(s)

acqn:
- Fixed bitwise shift when the rhs is a numeric constant with side-effects.
- "Fixed" signed numeric constant bitwise shift when the result should overflow so that we always have 2's complement results.

### DIFF
--- a/src/cc65/expr.c
+++ b/src/cc65/expr.c
@@ -216,8 +216,11 @@ void LimitExprValue (ExprDesc* Expr)
             break;
 
         case T_LONG:
+            Expr->IVal = (int32_t)Expr->IVal;
+            break;
+
         case T_ULONG:
-            /* No need to do anything */
+            Expr->IVal = (uint32_t)Expr->IVal;
             break;
 
         case T_SCHAR:
--- a/src/cc65/shiftexpr.c
+++ b/src/cc65/shiftexpr.c
@@ -64,11 +64,11 @@ void ShiftExpr (struct ExprDesc* Expr)
     CodeMark Mark1;
     CodeMark Mark2;
     token_t Tok;                        /* The operator token */
-    const Type* EffType;                /* Effective lhs type */
     const Type* ResultType;             /* Type of the result */
     unsigned ExprBits;                  /* Bits of the lhs operand */
     unsigned GenFlags;                  /* Generator flags */
     unsigned ltype;
+    int lconst;                         /* Operand is a constant */
     int rconst;                         /* Operand is a constant */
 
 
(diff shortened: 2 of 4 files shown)


mrdudz:
Would be nice to have a test for this (which fails before and works after this fix)

acqn:
Hmm, so what's with this test? It passes for me.

spiro-trikaliotis:
Please note that some of the tested left-shifts are UB.
cf. n1256, 6.5.7 "Bitwise shift operators", clause 4:

"The result of E1 << E2is E1 left-shifted E2 bit positions; vacated bits are filled with
zeros. If E1 has an unsigned type, [...]. If E1 has a signed
type and nonnegative value, and E1 × 2^{E2} is representable in the result type, then that is
the resulting value; otherwise, the behavior is undefined."

Note that 0x8000 (signed 16 bit int) * 2 is not representable in signed 16 bit; thus, this is UB.

acqn:
> Please note that some of the tested left-shifts are UB. cf. n1256, 6.5.7 "Bitwise shift operators", clause 4:
> 
> "The result of E1 << E2is E1 left-shifted E2 bit positions; vacated bits are filled with zeros. If E1 has an unsigned type, [...]. If E1 has a signed type and nonnegative value, and E1 × 2^{E2} is representable in the result type, then that is the resulting value; otherwise, the behavior is undefined."
> 
> Note that 0x8000 (signed 16 bit int) * 2 is not representable in signed 16 bit; thus, this is UB.

True. But we can make the results always "2's complement" which is a legitimate outcome of UB.
And failing the test doesn't make sense according to cc65's implementation...

mrdudz:
> But we can make the results always "2's complement" which is a legitimate outcome of UB.

That seems reasonable to me

> And failing the test doesn't make sense according to cc65's implementation...

Not sure if i understand this one - but to me it makes sense to also test if UB actually produces what we expect cc65 to produce. Perhaps add a comment explaining this in the test where needed

spiro-trikaliotis:
> @spiro-trikaliotis does this look ok to you now? :)

Well, I did not have a detailed look into the implementation. I only wanted to say that this patch wants to achieve more than the standard asks a compiler to do.

Of course, we are free to use 2's complement implementation (who would not do so in the last 40 years?), so if it is an agreement to rely on it, I'm fine.

Again, I did not have a detailed look into the patch nor into the test cases, if they are enough. I doubt I will have much time to do so in the near future, so feel free not to wait on my "ok".

I do not have any objections, though.

mrdudz:
good enough for me :)

PR #1676, opened 2022-02-17
smarter manufacturer

polluks2:
some kind of promotion

### DIFF
--- a/libsrc/lynx/exehdr.s
+++ b/libsrc/lynx/exehdr.s
@@ -16,8 +16,12 @@
         .word   __BANK1BLOCKSIZE__                      ; bank 1 page size
         .word   1                                       ; version number
         .asciiz "Cart name                      "       ; 32 bytes cart name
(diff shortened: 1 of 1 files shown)


mrdudz:
Wouldnt it make more sense to have some external object that the user can replace as needed?

mrdudz:
I am not convinced this is better, sorry. Especially since it is using the known broken CC65 version symbol :)

IMHO the string should simply be blank by default, and there should be a way to override it with an external symbol.

greg-king5:
> I am not convinced this is better, sorry. Especially since it is using the known broken CC65 version symbol. :)

Then, don't use the version -- just `"cc65"`.
> 
> IMHO, the string should simply be blank by default, and there should be a way to override it with an external symbol.

I don't think that it can be done for a string literal.

mrdudz:
> I don't think that it can be done for a string literal.

why not? just have it in a seperate object and then link it - just like we do with other things that can be replaced by the user

mrdudz:
merging this, since the version symbol is fixed now

PR #1679, opened 2022-02-19
Reorganized test/asm

spiro-trikaliotis:
Reorganized test/asm to allow for easier adding of test cases by simply dropping a file into test/asm/listing/

It also adds more reference files:

for a <test>.s file, you can add:
<test>.bin-ref as a reference for the resulting binary
<test>.list-ref as a reference file for the resulting listing output (remove the first line, because it is a version info of ca65!)

Additionally, it tests if the resulting .bin is the same if called with and without a listing file. That is, it can be used to find out if the generation of a listing file changes compilation (which it shouldn't).


mrdudz:
please update test/asm/readme.txt to contain info about what exactly goes into which subdirectory, and how to add tests

spiro-trikaliotis:
> please update test/asm/readme.txt to contain info about what exactly goes into which subdirectory, and how to add tests

done.

mrdudz:
looks good to me!

Issue #1584, opened 2021-10-22
[bug][tgi][atari_lynx] TGI_COLOR_BLACK is transparent and not a solid color


Fabrizio-Caruso:
Issue #956 was never fully solved.
On a white background, it is not possible to display black text.
The TGI_COLOR_BLACK is still implemented as transparent. 
TGI_COLOR_BLACK should be a real solid color.

The problem is reproduced with this example:
```c
#include <tgi.h>
#define CLI() asm("\tcli")
int main(void)
{
    tgi_install (tgi_static_stddrv);
    tgi_init ();        
    CLI();
    while (tgi_busy())  {  };
    tgi_setpalette(tgi_getdefpalette());
    tgi_setbgcolor(TGI_COLOR_WHITE);
    tgi_outtextxy(8,8,"    "); 
    tgi_setcolor(TGI_COLOR_BLACK);
    // tgi_setcolor(TGI_COLOR_RED);
    tgi_outtextxy(8,8,"hello");
        
    while(1){};
    return 0;
}
```
@greg-king5 We still have a problem with the black color... Do you know what is still wrong?

Fabrizio-Caruso:
@oliverschmidt Who are the TGI and ATARI_LYNX experts that could fix this bug?

polluks:
Please take a look
![image_thumb15](https://user-images.githubusercontent.com/4707573/142412220-4ba0a219-8e32-4863-939d-4e40304333f0.png)
https://atarilynxdeveloper.wordpress.com/2012/04/25/programming-tutorial-part-6colors/

Fabrizio-Caruso:
@polluks I know how to display black on white by avoiding some commands but the code above should work. It works with red but not with black,

Fabrizio-Caruso:
@greg-king5, @oliverschmidt could this issue be tagged as "bug" ?
If not why?

karrika:
There is actually only one solution that would make the text work correctly with index 0 (TGI_BLACK).
The solution is to create the characters as shaped sprites. This would add complexity to the text font as some characters need to be split up into two sprites. This is because we cannot draw holes if we don't use index 0 as transparency.
The fix would be possible to implement.

Fabrizio-Caruso:
@mrdudz conio? There is still no CONIO support for this target  (#960). @greg-king5 has made some experiments but there was a problem with some characters.

Fabrizio-Caruso:
[...]

PR #1686, opened 2022-02-24
Mark index 0 as TRANSPARENT. Let palette start from index 1

karrika:
#1584 proposed fix

### DIFF
--- a/include/lynx.h
+++ b/include/lynx.h
@@ -52,24 +52,25 @@
 
 
 /* Color defines */
-#define COLOR_BLACK             0x00
-#define COLOR_RED               0x01
-#define COLOR_PINK              0x02
-#define COLOR_LIGHTGREY         0x03
-#define COLOR_GREY              0x04
-#define COLOR_DARKGREY          0x05
-#define COLOR_BROWN             0x06
-#define COLOR_PEACH             0x07
-#define COLOR_YELLOW            0x08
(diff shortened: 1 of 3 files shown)


greg-king5:
Please update `"lynx/tgi_colors.s"`.

Fabrizio-Caruso:
@mrdudz Could this PR be merged? If not, what is missing?

mrdudz:
A lost a bit track of this - do we have a test program for this thing now? As with other things, it would be good to have a portable test program that can be used to check (and compare) the output on different platforms to make sure its consistent.

Fabrizio-Caruso:
@mrdudz I suppose we already have such test:
https://github.com/cc65/cc65/blob/master/samples/tgidemo.c

mrdudz:
I was more thinking of something that uses more different text- and background colors... perhaps tgidemo can be expanded a bit to do that?

karrika:
> I was more thinking of something that uses more different text- and background colors... perhaps tgidemo can be expanded a bit to do that?

The idea of bgcolour only exists in the Lynx. The concept is implemented with tgi_ioctl.

I may have added it in the early days as a way to print black. You were able to print black text on white paper this way.

mrdudz:
> The idea of bgcolour only exists in the Lynx. The concept is implemented with tgi_ioctl.

TGI has no background color? OK (shows how often i use it =P)

merging then :)

Issue #805, opened 2018-11-27
Q) Is 7800 supported with cc65?


zezba9000:
As the title says. Does the cc65 compiler support the Atari 7800 features or only 2600 features?

Or am I asking the wrong question with my limited knowledge of the hardware so far?

Fabrizio-Caruso:
The Atari 5200, Atari 2600 consoles and Atari 8-bit computers are supported.
I wish Atari 7800 were supported.
You can use the none target to create binaries for the Atari 7800 but you have to do more work.

Fabrizio-Caruso:
This is the only example of an attempt to use CC65 to code a game for the Atari 7800:
https://sourceforge.net/projects/rfk7800/files/rfk7800/

zezba9000:
Ok tnx. Backed it up for ref.

zezba9000:
Its odd there isn't more 7800 and Jaguar support for C. Both those console can just do more and have a lot more ram to work with.

polluks:
@Fabrizio-Caruso There was a BBC target long time ago, see `TGT_BBC`. It needs a cleanup for V2.17.

groessler:
The 7800 seems to have just 4k of RAM. Very strange, since the older 5200 had 16k. What did the developers smoke? But probably it was forced upon them by upper management.

AFAICT it supports some kind of character mode. The font for this mode would typically have to be stored in the program ROM. Then a light version of conio might be possible.

Who volunteers?

karrika:
I am currently trying to port over Wizzy from Lynx to 7800. Creating a new target was not difficult. There is also a stdjoy driver now that appears to work. The TGI driver is next.

If I get my game ported and working I may have a look at making a pull request.

Fabrizio-Caruso:
@karrika Thanks for the news! Even if you do not port TGI and stdio, but you have a working crt0.s and cfg, could you please share it and/or integrate it in CC65? Some CC65 targets have minimal libs such as for example Supervision (no conio, no stdio, no TGI). Supervision is a usable target (I have written the hello world example in CC65).
Keep up the good work!

karrika:
> @karrika Thanks for the news! Even if you do not port TGI and stdio, but you have a working crt0.s and cfg, could you please share it and/or integrate it in CC65? Some CC65 targets have minimal libs such as for example Supervision (no conio, no stdio, no TGI). Supervision is a usable target (I have written the hello world example in CC65). Keep up the good work!

[...]

PR #1692, opened 2022-03-01
#805 Atari7800 target

karrika:
In answer to the request #805. A first version for the atari7800. Published just to attract any fellow developers to contribute to this.

The target compiles and runs on emulators.

In order to produce a working executable I link the stuff like:
```
$(target) : $(objects)
        $(CL) -t atari7800 -o $@ $(objects) -l atari7800
        sign7800 -w $(target)
```
The config will automatically insert a default header file. So you just need to use a program for signing the data (sign7800).


Fabrizio-Caruso:
Thanks a lot! Could you please also add a hello world example or something similar? It would help a lot people to get started.

mrdudz:
> Thanks a lot! Could you please also add a hello world example or something similar? It would help a lot people to get started.

In any case it should be added to the samples makefile :) (But indeed, some kind of hello world would be good)

mrdudz:
> So you just need to use a program for signing the data (sign7800).

please document this accordingly

karrika:
> > So you just need to use a program for signing the data (sign7800).
> 
> please document this accordingly

The only thing I know about this code is what I wrote in the first note.

sign7800 -w wizzy.a78

will sign the binary so it can be played on the real 7800. I just downloaded the sign7800 zip file and typed "make". Perhaps documenting the Atari encryption protection is outside the scope of cc65?

karrika:
> > Thanks a lot! Could you please also add a hello world example or something similar? It would help a lot people to get started.
> 
> In any case it should be added to the samples makefile :) (But indeed, some kind of hello world would be good)

Yep. I hope to create a small TGI driver for the 7800. It would allow me to run the same C-code on Lynx and the 7800 to some extent. (Like Hello World!)

mrdudz:
> The only thing I know about this code is what I wrote in the first note.
>
> sign7800 -w wizzy.a78
>
> will sign the binary so it can be played on the real 7800. I just downloaded the sign7800 zip file and typed "make".

Then this should be mentioned in the documentation

karrika:
> > > Thanks a lot! Could you please also add a hello world example or something similar? It would help a lot people to get started.
> > 
> > 
> > In any case it should be added to the samples makefile :) (But indeed, some kind of hello world would be good)
> 
> Yep. I hope to create a small TGI driver for the 7800. It would allow me to run the same C-code on Lynx and the 7800 to some extent. (Like Hello World!)

I prefer to not create too large pull requests at once. So I will work on tgi + joy drivers locally for my wizzy game only and not add more content to this pull request.

libsrc/atari7800/exehdr.s
```diff
@@ -0,0 +1,46 @@
+;
+; Karri Kaksonen, 2022
+;
+; This header contains data for emulators
+;
```
mrdudz, line 5: Please elaborate what this actually produces - IMHO the toolchain should produce binaries that run as is on actual hardware, not some kind of file with extra headers that only work in some emulated environment - producing those should be an extra step

karrika: The 7800 has a lot of compile time configuration details. Without the header info the build is pretty useless. Headers are always generated on the Lynx target also by default. Having a header is the way a ROM burner knows how to burn the build to a working cart.

The a78 header appears to be the "standard" way to pass the information of the compilations. Therefore it is added to the default config file. By simply removing it from the config file you can generate just the binary.

What might be a good next step would be to have some defines in the config file that would automatically trigger the right bits in the header file.

mrdudz: The header should perhaps be better a weak import, like in [atari-cart.cfg](https://github.com/cc65/cc65/blob/master/cfg/atari-cart.cfg) and [carthdr.s](https://github.com/cc65/cc65/blob/master/libsrc/atari/carthdr.s), then the user has full control about the process.

Any any case, what you say should be stated in the .cfg file and the documentation

karrika: I don't really get it. I tried to use weak definition for EXEHDR and even tried to see what happens if I add CARTSIZE to trigger the definition of EXEHDR. Both experiments just add complexity for the user. The definition EXEHDR: type = import; is so clear. You want to import an EXEHDR.

If you have some insight in why you want to use more complicated techniques for triggering the inclusion of EXEHDR please educate me.

mrdudz: using weak makes it optional, you can produce a binary with or without the extra header just by linking it or not - same as with the atari cartridge

karrika: I believe the reason is because of the use cases. An ordinary user wants a workflow that is simple and helps to streamline the development. Picking the choices in the cfg file for many details is a good way to customize the build. Reflecting the choices in the header is a good way for documenting them.

But it is of course up to you what you want to have in your repository.

I pushed the changes I want to have in there.

mrdudz: > Picking the choices in the cfg file for many details is a good way to customize the build.

However, then you will have to change the default cfg file. For cc65 the general rule always has been: once you do this, you are entirely on your own, and building the supported "normal" files will never require this. Ideally you shouldnt have to do this at all for common usecases.

In this case it seems like the way to go is something like:

- exehdr in a seperate file which is imported "weak" by the startup code
- exehdr itself imports a bunch of symbols "weak" which allow to configure the header

now the user can forceimport the header in his code (or just put it on the linker cmdline), and define the symbols he wants to change to configure it - no changes in the provided library or cfg file needed.

karrika: I pushed the changes a minute ago.

karrika:
Not in progress. I just tried to address the concerns with the EXEHDR. In my opinion this could be merged now.

karrika:
> Not in progress. I just tried to address the concerns with the EXEHDR. In my opinion this could be merged now.

I prefer to close this pull request soon. Does this still need some work?

mrdudz:
Am Donnerstag, 17. März 2022, 10:53:14 CET schrieb Karri Kaksonen:
> > Not in progress. I just tried to address the concerns with the EXEHDR. In
> > my opinion this could be merged now.
> I prefer to close this pull request soon. Does this still need some work?

please keep in mind maintaining this is a hobby project and time is very 
limited - i will look at it this weekend hopefully

(right now github doesnt even let me use its website for some reason, i 
couldnt merge this right now at all.... weird)

-- 

Wenn eine vom Kopienräuber geraubte Kopie eine Raubkopie ist, dann ist ein vom 
Kopiererräuber geraubter Kopierer ein Raubkopierer.

karrika:
> please keep in mind maintaining this is a hobby project and time is very limited - i will look at it this weekend hopefully

I fully understand. Take your time.

mrdudz:
Looks ok to me now, thanks!

Issue #1680, opened 2022-02-19
cc65: segfault when building on MSYS2 MinGW64


Gumball2415:
On trying to build cc65 on my Windows laptop, a segmentation fault occurred when the makefile tried to compile `common/_scanf.c`.
```
...
apple2 - common/_scanf.c
make[2]: *** [Makefile:283: ../libwrk/apple2/_scanf.o] Segmentation fault
make[1]: *** [Makefile:118: apple2] Error 2
make: *** [Makefile:7: all] Error 2
```

I tried running gdb to see what's wrong:
```
$ gdb -args ../bin/cc65.exe -t apple2 -Or -W error --create-dep ../libwrk/apple2/_scanf.d --dep-target ../libwrk/apple2/_scanf.o -o ../libwrk/apple2/_scanf.s common/_scanf.c
GNU gdb (GDB) 11.2
Copyright (C) 2022 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-w64-mingw32".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<https://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from ../bin/cc65.exe...
(gdb) r
Starting program: C:\cc65\bin\cc65.exe -t apple2 -Or -W error --create-dep ../libwrk/apple2/_scanf.d --dep-target ../libwrk/apple2/_scanf.o -o ../libwrk/apple2/_scanf.s common/_scanf.c
[New Thread 10924.0x21ac]

Thread 1 received signal SIGSEGV, Segmentation fault.
0x00007ffcd317d2f1 in strlen () from C:\WINDOWS\System32\msvcrt.dll
(gdb) bt
#0  0x00007ffcd317d2f1 in strlen () from C:\WINDOWS\System32\msvcrt.dll
#1  0x00007ff758da31f8 in FormatStr ()
#2  0x00007ff758da3605 in xvsnprintf ()
#3  0x00007ff758da3b7c in xsprintf ()
#4  0x00007ff758d64b56 in GetLabelName ()
[...]

PR #1694, opened 2022-03-02
Fix cc65 segfault on 64-bit LLP64 Windows builds

nyanpasu64:
Currently cc65 segfaults on LLP64 Windows builds (eg. MinGW but not MSYS2), because pointers are casted to `uintptr_t`, narrowed to `unsigned long` in function arguments (which truncates to 32-bit on Win64), and casted back to pointers (to invalid memory) and dereferenced.

This PR fixes two occurrences of `unsigned long` in `g_getimmed` and `g_defdata`, and makes `make` succeed without segfaulting. I'm told the resulting cc65 binary works properly in some real-world use-cases.

There are many remaining occurrences of `unsigned long` in codegen.h's function arguments. I don't know if this PR makes cc65 behave correctly in all cases, or if there are other cases where pointers are cast to `uintptr_t` then `unsigned long` which break on Win64 (but are either not triggered or don't cause segfaults in unit tests).

See #1680.


nyanpasu64:
Looking at https://github.com/cc65/cc65/runs/5389072616?check_suite_focus=true, you only perform 32-bit MSVC CI, and don't seem to run tests (personally I'd change that). And your Visual Studio project setup (https://github.com/cc65/cc65/blob/master/src/cc65.vcxproj) seemingly lacks a 64-bit mode entirely (not 100% sure). Do you want to add Win64 support to your .vcxproj and CI or not?

nyanpasu64:
The Linux CI doesn't like my format strings. I don't know what format string for uintptr_t is accepted by both MinGW's error checking and xvsprintf's format string parsing. Is it safe to replace the `&` calls with casts to uint8/16/32_t or unsigned char/short/int, and adjust the format strings accordingly? (I like this idea in general, but uint8/16/32_t don't have dedicated format strings, making it suspect in this case.)

mrdudz:
> Looking at https://github.com/cc65/cc65/runs/5389072616?check_suite_focus=true, you only perform 32-bit MSVC CI, and don't seem to run tests (personally I'd change that). And your Visual Studio project setup (https://github.com/cc65/cc65/blob/master/src/cc65.vcxproj) seemingly lacks a 64-bit mode entirely (not 100% sure). Do you want to add Win64 support to your .vcxproj and CI or not?

I can not comment on anything VS - i am not using it. If you think the project files can be improved, by all means do it and make a PR for it, so someone who uses it can look at it.

As for the CI - sure, it can (and should be) more thorough. What we have right now is only the starting point to get things going again at all - if you can improve it, please do.

> The Linux CI doesn't like my format strings. I don't know what format string for uintptr_t is accepted by both MinGW's error checking and xvsprintf's format string parsing. Is it safe to replace the & 0xFF... calls with casts to uint8/16/32_t or unsigned char/short/int, and adjust the format strings accordingly?

a) casts should not be needed and they only hide problems - please do not use them :)
b) look at eg https://embeddedgurus.com/stack-overflow/2011/02/formatted-output-when-using-c99-data-types/ and https://www.cplusplus.com/reference/cinttypes/ - i think PRIu64PTR is what you need to use. The syntax is a bit weird, excerpt from VICE:
```
    fprintf(stdout, "DBGCART: exit(%d) cycles elapsed: %"PRIu64"\n", n, maincpu_clk);
```

nyanpasu64:
> a) casts should not be needed and they only hide problems - please do not use them :)

Normally I'd agree, but casting is lossless when you're already performing bit masking:

```
cc65/codegen.c: In function ‘g_defdata’:
cc65/codegen.c:4406:45: error: format ‘%lX’ expects argument of type ‘long unsigned int’, but argument 2 has type ‘uintptr_t’ {aka ‘long long unsigned int’} [-Werror=format=]
 4406 |                 AddDataLine ("\t.byte\t$%02lX", val & 0xFF);
      |                                         ~~~~^   ~~~~~~~~~~
      |                                             |       |
      |                                             |       uintptr_t {aka long long unsigned int}
      |                                             long unsigned int
      |                                         %02I64X
cc65/codegen.c:4410:45: error: format ‘%lX’ expects argument of type ‘long unsigned int’, but argument 2 has type ‘uintptr_t’ {aka ‘long long unsigned int’} [-Werror=format=]
 4410 |                 AddDataLine ("\t.word\t$%04lX", val & 0xFFFF);
      |                                         ~~~~^   ~~~~~~~~~~~~
      |                                             |       |
      |                                             |       uintptr_t {aka long long unsigned int}
      |                                             long unsigned int
      |                                         %04I64X
cc65/codegen.c:4414:46: error: format ‘%lX’ expects argument of type ‘long unsigned int’, but argument 2 has type ‘uintptr_t’ {aka ‘long long unsigned int’} [-Werror=format=]
 4414 |                 AddDataLine ("\t.dword\t$%08lX", val & 0xFFFFFFFF);
      |                                          ~~~~^   ~~~~~~~~~~~~~~~~
      |                                              |       |
      |                                              |       uintptr_t {aka long long unsigned int}
      |                                              long unsigned int
      |                                          %08I64X
```

Where the value after `& 0xFF/FFFF/FFFFFFFF` is *logically* 8/16/32-bit. What host CPU architectures does cc65 support, and do they all have 32-bit or more long? Or should I cast all of them to `unsigned long`?

> i think PRIu64PTR is what you need to use.

- I don't think PRIu64PTR exists ([searching across all of GitHub](https://github.com/search?q=PRIu64PTR&type=code) there's only a single occurrence in a toy kernel). https://stackoverflow.com/a/9053835 instead suggests PRIXPTR. I found a list at https://pubs.opengroup.org/onlinepubs/009696899/basedefs/inttypes.h.html, and https://stackoverflow.com/q/12970293 says it's found on MSVC since VS2013. I didn't find it in the MSVC tree but instead "C:/Program Files (x86)/Windows Kits/10/Include/VERSION/ucrt/inttypes.h".
- Unfortunately on MinGW64, https://github.com/cc65/cc65/blob/master/src/common/inttypes.h prevents the system inttypes.h from being seen (even if I switch `#include "inttypes.h"` to `<inttypes.h>`), and itself fails to define PRI... macros. If I rename it to prevent it from being found, then cc65 compiles fine and finds PRIXPTR on MinGW64. Is this an acceptable solution or will it break building the compiler for platforms without a real inttypes.h?
- PRIXPTR is a macro designed for the system printf. The widths of int/long/pointer are consistent across printf/xprintf, and the format specifiers are xprintf-compatible on all systems I've checked so far (though not guaranteed to be the case):
	- On Win64, MSVC and MinGW's PRIXPTR = PRIX64 = `llX`, and xprintf supports `%lld/u/x/...`.
	- Linux has `/usr/include/inttypes.h`. 32-bit Linux uses `ll` for 64-bit and no prefix for 32-bit pointers, and 64-bit Linux uses `l` for 64-bit integers and pointers.
	- macOS 12 has `/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/inttypes.h` (it's ridiculous, you can't make this up) and only supports `lX` (since Mac is 64-bit only... I'm still mad at Apple for breaking 32-bit Wine).

So do you want to cast to unsigned long (which is safe because you're already masking to the lowest 8/16/32 bits), or remove `src/common/inttypes.h` and use PRIXPTR?

mrdudz:
> So do you want to cast to unsigned long (which is safe because you're already masking to the lowest 8/16/32 bits), or remove src/common/inttypes.h and use PRIXPTR?

I don't see a reason to duplicate a standard header file in our source - so indeed inttypes.h should be removed. I am not sure why this one even exists. If there is really a certain compiler that doesnt have this, IMHO its a much better idea to create such header file for the system the compiler runs on/compiles for and provide it to whoever makes the compiler :)

That said i don't know what consequences that may have - i don't see a problem with it though. @spiro-trikaliotis maybe you can comment on this?

nyanpasu64:
Just to verify, `include/inttypes.h` is used by cc65 to build 6502 apps, and `src/common/inttypes.h` is (for the time being) used by gcc/clang/msvc to *build* cc65?

mrdudz:
> Just to verify, include/inttypes.h is used by cc65 to build 6502 apps, and src/common/inttypes.h is (for the time being) used by gcc/clang/msvc to build cc65?

yes. at least i hope building cc65 doesnt use anything from include/ :)

spiro-trikaliotis:
> If there is really a certain compiler that doesnt have this, IMHO its a much better idea to create such header file for the system the compiler runs on/compiles for and provide it to whoever makes the compiler :)

Older version of MSVC did not have inttypes.h. In fact, it is a C99 header, so anything before that does not contain it.
I know of other projects that include an inttypes.h and even a stdint.h so they work on older compilers.

Have a look here, for example: https://stackoverflow.com/questions/126279/c99-stdint-h-header-and-ms-visual-studio

Thus, the question is: Do we still support old compilers that do not have stdint.h and inttypes.h? If so, we should include them (and update if necessary). If not, just drop them and use the compiler supplied files instead.

mrdudz:
> Do we still support old compilers that do not have stdint.h and inttypes.h?

IMHO that is a pointless idea - at some point every compiler was old and didnt contain certain standard header files. We are talking about something at least 20 years old here, i see little to no reason to care about it at all.

spiro-trikaliotis:
> We are talking about something at least 20 years old here

Looking at the link I provided, it seems MSVC added this in MSVC 2010 only. That is, MS compilers that are older than 12 (and not 20) years will exhibit this problem.

Of course, this is still a long time, so we do not need to take care of it (if there are not other reasons why we need these old compilers).

mrdudz:
As we happen to stumble about this in VICE too - here is another thing that could be done:

In some strategic place (i dont know where would be ideal - a header that every source includes would be best) put:
```
#if defined(__MINGW32__) || defined (__MINGW64__)
#define __USE_MINGW_ANSI_STDIO 1
#endif 
```
that makes eg %z work in msys2 (but it will also only work in non ancient msvc)

nyanpasu64:
I'm deleting `src/common/inttypes.h`. Should I change `#include "inttypes.h"` to `<inttypes.h>`? It occurs in 4 files: `src/cc65/codegen.c`, `src/cc65/codegen.h`, `src/cc65/exprdesc.h`, `src/common/xsprintf.c`.

mrdudz:
> Should I change #include "inttypes.h" to <inttypes.h>

yes

nyanpasu64:
I have some concerns about removing [src/common/inttypes.h](https://github.com/cc65/cc65/blob/2d0bc67c419059255af02cb43219421f10c35ab7/src/common/inttypes.h):

- On modern compilers, the header does literally nothing but include  <stdint.h>. (Oddly in GCC's headers, intmax_t and friends, which cc65 actually uses, is 64-bit not 128-bit.)

```
#if (__STDC_VERSION__ >= 199901) || (__GNUC__ >= 3) || (_MSC_VER >= 1600)
#include <stdint.h>
#else
...
#endif
```

Should I replace `#include "inttypes.h"` with `<stdint.h>` (and keep `<inttypes.h>` in parallel for `src/cc65/codegen.c`)?

- On older compilers, inttypes.h includes `<limits.h>` and `<stddef.h>`. I assume nothing depends on these headers, since it's not included on newer compilers and yet the code builds. The rest of the file is designed to emulate stdint.h, not inttypes.h.
- Should I reorder/move the include line out of the "local header" group, and into the <system header> group?

mrdudz:
> Should I replace #include "inttypes.h" with <stdint.h> (and keep <inttypes.h> in parallel for src/cc65/codegen.c)?

I dont understand the question. You want to use <stdint.h> instead of "inttypes.h" at some places and use <inttypes.h> at other places? That doesnt sound like the right thing to do to me :)

> Should I reorder/move the include line out of the "local header" group, and into the group?

Not sure i understand this either - but sure, the system headers included with <> should come first

nyanpasu64:
> You want to use <stdint.h> instead of "inttypes.h" at some places and use <inttypes.h> at other places? That doesnt sound like the right thing to do to me :)

Currently "inttypes.h" is a shim around <stdint.h> and nobody uses system <inttypes.h>. So I will replace all occurrences of "inttypes.h" with <stdint.h>, and then add <inttypes.h> to codegen.c since I'm going to be using `PRIXPTR` there.

mrdudz:
> > You want to use <stdint.h> instead of "inttypes.h" at some places and use <inttypes.h> at other places? That doesnt sound like the right thing to do to me :)
> 
> Currently "inttypes.h" is a shim around <stdint.h> and nobody uses system <inttypes.h>. So I will replace all occurrences of "inttypes.h" with <stdint.h>, and then add <inttypes.h> to codegen.c since I'm going to be using `PRIXPTR` there.

To be honsest - as long as the CI agrees i dont have a string reason to disagree. Someone who knows the standard (headers) better than i do and uses a more obscure compiler might (or maybe not) :)

nyanpasu64:
Correction: according to https://pubs.opengroup.org/onlinepubs/009696899/basedefs/inttypes.h.html, inttypes.h includes stdint.h. So I *could* alternatively replace all "inttypes.h" with <inttypes.h> and it won't *break* code. But should I do that or not? (Well at least I'm not going to include stdint.h *and* inttypes.h in one file.)

It appears `sitest.c` is run under cc65 rather than the host compiler. It appears to test cc65's target `<inttypes.h>`, but the Makefile says it fails to compile. So I won't be touching that.

mrdudz:
> So I could alternatively replace all "inttypes.h" with <inttypes.h> and it won't break code. But should I do that or not? (Well at least I'm not going to include stdint.h and inttypes.h in one file.)

you could do that, i guess

> It appears sitest.c is run under cc65 rather than the host compiler. It appears to test cc65's target <inttypes.h>, but the Makefile says it fails to compile. So I won't be touching that.

there are no tests that test the host compiler. i dont recall why it fails :)

Compyx:
`sitetest.c` fails to compile because it contains `#warn` instead of `#warning`. With `#warning` it compiles with gcc on Linux and msys2.

mrdudz:
> Another thing: Why do I need to use PRIXPTR instead of casts for printing hex literals in `g_defdata`, when `g_getimmed` already uses casts in `(unsigned char) Val` and `(unsigned char) (Val >> 8)`?

Generally we have to start somewhere - i dont know why the other code is like it is. Ideally the other code is changed as well :) Perhaps previous maintainers had different preferences (reminds me to brush up a codestyle doc), i never looked at the code in that detail since recently.

mrdudz:
> Otherwise can the code be reviewed and possibly merged? (I didn't test Visual Studio build, I only edited the .vcxproj by hand.)

Wouldnt hurt if someone could check those VS files... @spiro-trikaliotis perhaps?

spiro-trikaliotis:
> Wouldnt hurt if someone could check those VS files... @spiro-trikaliotis perhaps?

Why me? I do not have a native Windows anymore.

Ok, I started VS 2009 in a VM (which is a PITA!). The Debug and the Release build compile flawlessly, with and without this PR merged.

I did not test the resulting binaries, though.

mrdudz:
> Why me? I do not have a native Windows anymore.

You are the only one i remember who uses (or used) VS :)

merging then...

spiro-trikaliotis:
> Ok, I started VS 2009 in a VM (which is a PITA!).

Correction: It is VS *2019*.

> You are the only one i remember who uses (or used) VS :)

Yes, I am still using VS6. ;)

PR #1695, opened 2022-03-03
[Build] Fixed test/asm/listing/Makefile with mingw32-make.exe

acqn:
Fixed test/asm/listing/Makefile with (at least some versions of) mingw32-make.exe that may have trouble with comments lead with tabulators.


mrdudz:
The problem is not make here - anything that isnt at the start of the line will be passed to the shell - so those lines would only work in sh or bash or whatever interprets # as comment, but not eg cmd.exe

PR #1698, opened 2022-03-03
Corrected size of BASROM definition in Sym1 config files

WayneParham:
BASIC in ROM on a Sym-1 computer resides at location C000 and is 8Kb wide.  However, the configuration files incorrectly label "BASROM" as being 4Kb.  This change corrects that.

There is no change in cc65 behavior because the only segment used is "RAM," which is set for lower memory.   But it's nice for the memory definitions to be right anyway.


mrdudz:
I'll have to take your word for it :)

PR #1699, opened 2022-03-03
Added Sym-1 extended memory sample program and documentation

WayneParham:
This pull request adds a Sym-1 sample program to show how to access memory above 32K and add it to the heap.

I have tested it on a Sym-1 with 64KB and it works as designed.


polluks2:
Would you wrap the lines for easy reading?

WayneParham:
Will do.

mrdudz:
Thanks!

PR #1707, opened 2022-03-19
Add Atari Lynx versions of three samples programs.

greg-king5:
The general samples programs `"hello.c"`, `"mandelbrot.c"`, and `"tgidemo.c"` can't be built by the `lynx` target because they use conio output or stdio.  I created special Lynx versions of those programs that don't use that I/O.


mrdudz:
looks good, thanks!

Issue #1705, opened 2022-03-14
[atari7800] clock() function


karrika:
The Atari 7800 only has NMI interrupts. There is no timer connected to IRQ's.

You can create NMI interrupts at the end of the previous display zone by setting a bit in the Display List List entry.

A pretty general way to have a steady time source could be to have two NMI's on the display list. One when the first visible scanline starts. Another when the screen drawing is complete.

You want to mess with pointers and offsets to implement scrolling and moving sprites during the vblank period.

Once the display drawing starts we have time for other things like updating time.

Would it be a good idea to call the list for interruptors at the 2nd NMI and implement the clock() function as in other targets?

Opinions?

karrika:
> A side effect would be waitvsync() for free?

Yes.

karrika:
Perhaps we also need to run a counter to measure how long the complete display cycle is to decide if the console is PAL or NTSC. Just to know how long the clock tick is.

karrika:
Thank you!

PR #1709, opened 2022-03-22
[Atari7800] clock()

karrika:
Issue #1705 The clock() function requires NMI interrupts to work in this implementation. The NMI calls irq.s in order to be able to use interruptors like for other targets. We also need to find out if the machine is a PAL or a NTSC unit. As there is no DLL structures at startup we poll the vsync bits for one frame and count the lines.

I did test the functions by creating a game. But there is not yet enough data in the repo for running the test samples.

Having PAL/NTSC detection, irq's and clock() are very good building blocks for any applications for the Atari7800.

### DIFF
--- a/cfg/atari7800.cfg
+++ b/cfg/atari7800.cfg
@@ -10,7 +10,7 @@ SYMBOLS {
     __ENCRYPT_BOTTOM__:   value = $ff7a, type = export;
     __ENCRYPT_SIZE__:     value = $80, type = export;
     __MEMORY_TOP__:       value = __ENCRYPT_BOTTOM__, type = export;
-    __INIT_SIZE__:        value = 107, type = export;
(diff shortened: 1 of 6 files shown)


mrdudz:
I'd like to see the clock functions in seperate sourcefiles, not all might be needed (eg clockspersec), and we should only ever link what is really used.

Also since you did pal/ntsc detection, please implement unsigned char get_tv (void) too (returns the pal/ntsc flag)

mrdudz:
are you done? :)

karrika:
> are you done? :)

I hope so :)

At least stuff appears to work with this setup. I also did measure the time of the clock() and compared to my stopwatch the times appear to be consistent.

mrdudz:
ok :)

Issue #1710, opened 2022-03-23
[atari7800] conio


karrika:
The Atari 7800 has only 4k of RAM so most of times you don't need anything like conio. But it would be useful for testing stuff to have a minimal conio implementation. What I had in mind is a 40x28 character display using the 320A mode of the Atari 7800. The font would be in a separate vga.s file that would allow you to swap it for your own font as in other systems.

The choice of 28 lines comes from the fact that this is the maximum number of lines we can show on a NTSC console is 224. On a PAL we show the same amount and draw a blank border at the bottom of the screen. 

Due to very limited RAM the text colour could only be changed on a line-by-line basis. You can also change the colour for the left and the right half of the line. The background can also have a defined colour but it affects the entire screen.

I would like to implement a basic system first to output text and assign colours. No scrolling or blinking cursor in the first version. Perhaps there is a way to have these functions in separate files to minimize memory footprint if they are not needed in the application.

The goal would be to run Hello World from the samples directory. Perhaps you could even run some character mapped games on it.

Comments?

Fabrizio-Caruso:
That is cool! So, there is no way to assign a single foreground color to each character? 
How are all the existing colorful games possible? Do they rely on sort of sprites as the Lynx? 
Having a replaceable VGA.s is nice. My framework implements an abstraction layer for single per-character tile graphics. So it is perfect for me except maybe the lack of per-character color.

karrika:
There is a way to assign a single foreground color to every character. But that needs a separate zone structure for every character. I would like to have one zone for 20 characters. The downside is less colours.
You won't like the VGA.s as all characters needs to be upside-down and interleaved by 256 bytes. So the block is 256 characters 8x8 bits. Single bit per pixel. There is no way to have an abstraction layer for single per-character tile graphics. You need to compile all your characters into a block and put it in memory aligned to 256 bytes. So one 8x8 sprite for index 0 has its bytes like:
0 - row 7
256 - row 6
512 - row 5
768 - row 4
1024 - row 3
1280 - row 2
1536 - row 1
1792 - row 0

The next character in the font, index 1 would be
1 - row 7
257 - row 6
513 - row 5
769 - row 4
1025 - row 3
1281 - row 2
1537 - row 1
1793 - row 0

and so on up to index 255 in the font. The same is true to all graphics in Atari 7800.

Fabrizio-Caruso:
@karrika I will implement the abstraction layer as long as I understand how it is structured.

[...]

PR #1715, opened 2022-04-02
[atari7800] conio

karrika:
A very crude conio target with coloured fonts. #1710 
This pull request has only the basic screen buffer and a clrscr() function. The font is very limited. Just upper chase characters and the numbers. Plus 6 characters for punctuation and building boxes on screen.
The screen is a 20 x 28 character screen that works on NTSC and PAL.

Currently the only access is to address the screen buffer directly.


karrika:
Ok. So no scrolling in conio. There appears to be tons of routines in conio. I don't believe I have time to implement them all right now. The current methods can output characters and set colours. You can also position your cursor on the screen and make it visible/invisible.

I wonder if someone would have time to test the functions a bit. I can print stuff on screen and the code appears to work.

karrika:
Using this conio implementation with a pre-written adventure game compiles, but the result shows that this conio is still seriously broken. I changed this to draft for a while...

karrika:
The problem was that I forgot that runlib stuff is unavailable for drivers. Now my tests work as they should.

karrika:
The weird bugs are gone and I even added a blinking cursor for a little style in the conio implementation. The code works nicely with a text adventure from the 90's. But of course it would be nice to run some more tests.
The implementation is now complete from my point of view. So this can be merged.

Fabrizio-Caruso:
@mrdudz Could this PR be merged?

mrdudz:
So here is my review :) A few things need minor fixing:

- the functions in `setcursor.s` should be seperated (generally always seperate functions from each other), ie `gotox`, `gotoy`, `wherex`, `wherey` 

- `textcolor` should be seperated from `cputc.s`

- all functions that return `char` must still return 0 in X - see eg [wherex.s](https://github.com/cc65/cc65/blob/master/libsrc/cbm/wherex.s),ie `textcolor`, `gotox`, `gotoy`, `wherex`, `wherey` 

- please add comments that explain the binary format of the font to `libsrc/atari7800/font160.s` (see eg [neschar.s](https://github.com/cc65/cc65/blob/master/libsrc/nes/neschar.s) or [soft80_charset.s](https://github.com/cc65/cc65/blob/master/libsrc/c64/soft80_charset.s))

- we do not include binary blobs or random scripts in the sourcetree, please remove the following files:
`libsrc/atari7800/buildfont160.py`
`libsrc/atari7800/font160.data`
`libsrc/atari7800/font160.png`
`libsrc/atari7800/font160flipped.png`

Fabrizio-Caruso:
@mrdudz I understand but maybe buildfont160.py would be really useful for a user. CC65 already includes some small target-specific scripts or tools, e.g., gamate. Maybe if you really think it is bad to have it, it would be nice to at least have a link to a repository with the script and some example data.

mrdudz:
> @mrdudz I understand but maybe buildfont160.py would be really useful for a user. CC65 already includes some small target-specific scripts or tools, e.g., gamate. Maybe if you really think it is bad to have it, it would be nice to at least have a link to a repository with the script and some example data.

That is again something that belongs into the wiki - if we really want to include a tool to produce charset data, it should be a target agnostic native tool written in C (similar to the existing [chrcvt65](https://github.com/cc65/cc65/tree/master/src/chrcvt65)) that produces charsets for all the different conio implementations (ideally) from a single source.

karrika:
Here is a small demo of Pirate Cove implemented with conio.
https://raz0red.github.io/js7800/?cart=https://atariage.com/forums/applications/core/interface/file/attachment.php?id=934364

I actually implemented it both with th 160A mode conio but the flattened text was hard to read so I also implemented conio as a monochrome font.

I will read through the comments here and fix the stuff.

Fabrizio-Caruso:
@karrika will you make both CONIO library implementations available somewhere?

karrika:
The Pirate Cove game sources are publically available through the page where I published the game https://atariage.com/forums/topic/334129-pirate-cove-for-atari-7800/

mrdudz:
Another comment: please make sure to not use TABs for indention. I noticed the recently merged 7800 files have them (i will fix that)

karrika:
> So here is my review :) A few things need minor fixing:
> 
> * the functions in `setcursor.s` should be seperated (generally always seperate functions from each other), ie `gotox`, `gotoy`, `wherex`, `wherey`

Producing a separate gotoy will create confusion as the data structure holding the hpos of the cursor is duplicated to the end of each line. This also means that you must first set gotoy and then gotox. Because of this it is safest to only provide gotoxy()

karrika:
Is there some cc65 way for providing two different conio implementations?

Perhaps a define at compile time? Like -DCONIO320 for picking a 320 pixel per scanline conio.
Or a separate atari7800conio320.lib that could be referenced by the linker before the atari7800.lib.

mrdudz:
have a look at how the soft80 conio for c64 is implemented - basically you override the functions you need to override by linking them. you could have a look at my tetris thing to see how that works (it produces a regular and a soft80 binary): https://github.com/mrdudz/portris

karrika:
The names of the modules are something like soft80mono_cgetc. Where/how does this override the stock cgetc function? 30 minutes of reading did not reveal it for me.

greg-king5:
> The names of the modules are something like soft80mono_cgetc. Where/how does this override the stock cgetc function? 30 minutes of reading did not reveal it for me.

It's the exported labels that override.  "soft80mono_cgetc" exports `_cgetc` before "c64.lib" does.  Therefore, "soft80mono_cgetc" is used instead of the function in "c64.lib".

That behavior is another reason why it's a good idea to separate functions into their own files.  Separate modules makes it easier to override a particular one.

polluks:
@greg-king5 Maybe also a hint for cc65-intern.sgml?

karrika:
> > The names of the modules are something like soft80mono_cgetc. Where/how does this override the stock cgetc function? 30 minutes of reading did not reveal it for me.
> 
> It's the exported labels that override. "soft80mono_cgetc" exports `_cgetc` before "c64.lib" does. Therefore, "soft80mono_cgetc" is used instead of the function in "c64.lib".
> 
> That behavior is another reason why it's a good idea to separate functions into their own files. Separate modules makes it easier to override a particular one.

It took me ages to find the re-mapping in the extras/soft80.s

I would very much prefer to have a separate soft80 directory where you build a soft80.lib using the final cputc etc labels. You can then present that lib for the linker before c64.lib.

For a casual programmer this is a show stopper. You cannot find the modules of interest by simple "grep cputc * | grep export".

mrdudz:
> I would very much prefer to have a separate soft80 directory where you build a soft80.lib using the final cputc etc labels. You can then present that lib for the linker before c64.lib.

The philosophy of cc65 has always been to have one big library (eg c64.lib) for a target, and then override what you want to replace. This has pros and cons. If we split the library into several, that creates other problems.

The real problem here is lack of documentation - as so often.

karrika:
> > I would very much prefer to have a separate soft80 directory where you build a soft80.lib using the final cputc etc labels. You can then present that lib for the linker before c64.lib.
> 
> The philosophy of cc65 has always been to have one big library (eg c64.lib) for a target, and then override what you want to replace. This has pros and cons. If we split the library into several, that creates other problems.
> 
> The real problem here is lack of documentation - as so often.

Not too happy about it but that is your choice. In order to be consistent with other inputs for the cc65 I added a mono conio in this style.

I tested it on Pirate Cove and it works on PAL and NTSC.

karrika:
The build checks seems to have a problem. It fails with:
"dangling spaces" in samples/cbm/Makefile

I believe that my code should be ok. Got to go now...

mrdudz:
> I believe that my code should be ok. Got to go now...

should work again... apparently @greg-king5 broke it by updating the checker scripts

karrika:
Thanks! The mono conio has the full VGA font (256 characters). It also has the proper aspect ratio that makes it easy to read.
PirateCove

The font has only 1 colour but you can set the hardware palette in Atari 7800 to almost anything creating a nice atmosphere in a text game.

For me this is ready to be merged.

mrdudz:
Would you add the comment that explains the font format also into the mono_font file please?

Also, perhaps the font160.s file could be renamed to something more descriptive... conio_font perhaps? just incase another font is added at some point in the future.

the rest looks good to me!

karrika:
A small test code that works for me.

```
#include <atari7800.h>
#include <conio.h>

void main ()
{
        cputs("Hello World!");
        while (1) ;
}

```
Compile by:
cl65 -t atari7800 --force-import __EXEHDR__ atari7800-mono.o main.c -o test.a78

During my tests I have sometimes encountered builds where the screen is blank. The cure was to copy mono_conio.s to the application and link it in as one of the game files. I don't know why this happens.  My guess is that there could be some kind of race in the constructiion part of the code. But I have no example build that fails like this.

If you encounter a black screen then you get it working by linking in conio.s of mono_conio.s

Any pointers or ideas of how to fix this is appreciated.

Most of the time the build just works without problems.

karrika:
I believe I know the reason for the black screens occasionally. The blinking cursor operates before conio is running. I will add a guard tonight.

karrika:
The CPU is kind of slow in the Atari 7800. The previous fix made conio unstable so I had to revert it. Now it appears to be stable enough to work with audio and conio. (As long as you keep your effects reasonable for the console)

So ready to be merged. (Before I find the next problem...)

karrika:
I was able to create a black screen version also. The current PR works correctly and does not have this occasional black screen problem.

karrika:
The latest change moves IRQ processing to the start interrupt of when the screen is being drawn. This should make it more robust when you keep adding sounds and music.

karrika:
Is there still concerns?
I believe that all requests have been addressed and this can be merged.

mrdudz:
looks fine! thanks!

PR #1716, opened 2022-04-02
[Tentative: please review] Fix Gamate CONIO: non-solid characters bug

### LINKED ISSUE #1714: [bug][gamate][conio] On this target conio non-black characters have transparency (not solid background except for black)
CONIO on the Gamate is not consistent with other targets.
For instance COLOR_WHITE behaves as transparent. Printing on top of a some black text has no effect.
By the way only two colors are defined (COLOR_BLACK, COLOR_WHITE) with values 0 and 3 but two more "colors" (grey levels) also work as expected (in terms of luminocity) but also suffer from the same problem.

This is reproduced by the following minimal example:
```


#include <conio.h>
#include <unistd.h>

int main(void)
{
    clrscr();
    gotoxy(2,2);
    textcolor(COLOR_BLACK);
    cprintf("hello");
    sleep(1);
    textcolor(COLOR_WHITE);
    gotoxy(2,2);
    cprintf("ciao!");
    
    while(1){};
    return 0;
}
``` 

An ugly and probably inefficient fix that seems to work for me is to delete with a "black space character" before displaying a character in a different color.

Fabrizio-Caruso:
This tentative pull request fixes #1714.
As a side effect, it also makes all grey colors work correctly (to be further tested).

The fix (and the bug) can be reproduced with
```
#include <conio.h>
#include <unistd.h>

int main(void)
{
    clrscr();
    gotoxy(2,2);
    textcolor(COLOR_BLACK);
    cprintf("hello");
    sleep(1);
    textcolor(COLOR_WHITE);
    gotoxy(2,2);
    cprintf("ciao!");
    sleep(1);
    gotoxy(2,2);
    textcolor(1);
    cprintf("hello");
    sleep(1);
    textcolor(2);
    gotoxy(2,2);
    cprintf("ciao!");
    sleep(1);
    gotoxy(2,2);
    textcolor(COLOR_BLACK);
    cprintf("hello");
    sleep(1);
    textcolor(COLOR_WHITE);
    gotoxy(2,2);
    cprintf("ciao!");    
    while(1){};
    return 0;
}
```

### DIFF
--- a/libsrc/gamate/cputc.s
+++ b/libsrc/gamate/cputc.s
@@ -89,28 +89,32 @@ putchar:
         adc     #>(fontdata-$f8)
         sta     ptr3+1
 
-        lda     CHARCOLOR
-        and     #1
-        beq     @skip_plane1
-
         lda     #LCD_XPOS_PLANE1
         clc
         adc     CURS_X
         sta     LCD_X
 
-        ldy     #$f8
+        lda     CHARCOLOR
+        and     #1
+        beq     @delete1
+
+        ldy     #$F8
 @copylp1:
         lda     (ptr3),y
         eor     RVS
         sta     LCD_DATA
         iny
         bne     @copylp1
+        
+        beq @skip_delete1
 
-@skip_plane1:
+@delete1:
+        lda   #$00
+        sta   LCD_DATA
+        iny
+        bne @delete1
 
-        lda     CHARCOLOR
-        and     #2
-        beq     @skip_plane2
+@skip_delete1:
 
         lda     #LCD_XPOS_PLANE2
         clc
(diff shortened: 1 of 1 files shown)


mrdudz:
looks ok, thanks!

Fabrizio-Caruso:
@mrdudz thanks for the quick review and merge!
I am happy I could help with this contribution.

Fabrizio-Caruso:
My code is fishy as well.
This bug is fixed for BLACK and WHITE. The test I provided works fine and my games still work as long as they only use BLACK (and WHITE). Using the other two colors (with color values 1 and 2) is broken in my games. It looks like I am deleting more then I am supposed to delete.

Fabrizio-Caruso:
I have found one problem that should fix colors 1 and 2. Color 3 (COLOR_BLACK) was already good.
I am now looking at color 0 (COLOR_WHITE). If we assume that the background is always the same "white" (lighest green), then COLOR_WHITE should just delete the tile.

Fabrizio-Caruso:
The new PR fixes the display of grey/green levels:
https://github.com/cc65/cc65/pull/1717

PR #1727, opened 2022-04-23
[Telestrat] Cleaning and send fd to write/read

jedeoric:


libsrc/telestrat/close.s
```diff
@@ -1,16 +1,17 @@
...
+        .include        "telestrat.inc"
+        .include        "errno.inc"
+        .include        "fcntl.inc"
 
-; int open (const char* name, int flags, ...);    /* May take a mode argument */
+; int __fastcall__ close (int fd);
 .proc _close
-    BRK_TELEMON XCLOSE  ; launch primitive ROM
-    rts
+        jsr     popax       ; Get fd
```
polluks, line 14: Pop? This is a fast call...

jedeoric: Fixed

PR #1728, opened 2022-04-23
[Telestrat] Kbhit fix

jedeoric:
There is a wrong behavior reported from someone. 

When kbhit() is called, cgetc() returns the char.

In telestrat target, there is a keyboard buffer. When we get the char, this char is removed from this buffer. Cgetc() does not work.

This PR is done in order to fix this. If kbhit had been called, it set the char into ram (store_char). If cget is called and store_char is equal to 0, it calls Telemon keyboard routine, if it's a different value, it means that kbhit was called previously and cgetc returns store_char


mrdudz:
I can't comment on the code changes, so if noone steps up soon i will just merge :)

Just to reiterate, the intended behaviour is like this:

- if there is no character available (in the buffer), then kbhit should return 0 and cgetc should wait until a char is available
- if there is a character available, then kbhit returns 1 (and does not remove the char from the buffer), and cgetc returns the character from the buffer and removes it from the buffer

libsrc/telestrat/kbhit.s
```diff
@@ -6,12 +6,21 @@
 
         .export         _kbhit
 
+        .export         store_char
+
         .include        "telestrat.inc"
 
 _kbhit:
         BRK_TELEMON XRD0
```
greg-king5, line 14: It's possible that `kbhit()` could be called several times before `cgetc()` is called.  Therefore, `kbhit()` should not call Telemon if a character already is waiting.

libsrc/telestrat/kbhit.s
```diff
@@ -6,12 +6,21 @@
...
-        eor     #$01
+        bcs     @no_char_action
+        sta     store_char
+        lda     #$01
+        rts
+@no_char_action:
+        tax
         rts
+.data
+store_char:
```
greg-king5, line 25: If you move that variable to "cgetc.s", then `kbhit()` won't be linked into programs if it actually isn't called.

("samples/ascii.c" calls `cgetc()` and doesn't call `kbhit()`.)

Issue #1726, opened 2022-04-21
ca65: .macpack doesn't look for files in cc65 share dir


rofl0r:
```
ca65 -I common -o 01-implied.o 01-implied.a
common/macros.a:28: Error: Cannot open include file 'longbranch.mac': No such file or directory
```
the file is available in /share/cc65/asminc/longbranch.mac
i.e. in one of cc65's own directories.

strace from the above:
```
access("common/longbranch.mac", F_OK)   = -1 ENOENT (No such file or directory)
access("longbranch.mac", F_OK)          = -1 ENOENT (No such file or directory)
access("common/longbranch.mac", F_OK)   = -1 ENOENT (No such file or directory)
access("longbranch.mac", F_OK)          = -1 ENOENT (No such file or directory)
writev(2, [{"", 0}, {"common/macros.a:28: Error: Cannot open include file 'longbranch.mac': No such file or directory\n", 96}], 2common/macros.a:28: Error: Cannot open include file 'longbranch.mac': No such file or directory
) = 96
```

the expected behaviour is that the assembler always looks for files included via .macpack in it's own include dir (if the above search in explicitly specified include dirs fails), much like e.g. gcc search for headers it ships inside its own directory tree, e.g. when you `#include <xmmintrin.h>`

spiro-trikaliotis:
Well, you are doing something differently than I do:
```
$ ca65 --version
ca65 V2.19 - Git 451acb342
$ cat test.a 
        .macpack longbranch
$ ca65 -o test.o test.a
$ 
```
I am using the pre-built package from OBS (https://build.opensuse.org/package/show/home:strik/opencbm) on Debian 11.

spiro-trikaliotis:
I forgot one important line:
```
$ ca65 -I common -o test.o test.a`
$
```

spiro-trikaliotis:
What platform are you using? I see that on Windows, the behaviour is slightly different than on non-Windows.

The default search paths are, in this order:

1. the contents of the environment variable CA65_INC
2. the contents of the subdir asminc/ of the environment variable CC65_HOME
3. on non-Windows: What is defined in the (compilation variable) CA65_INC
4. on Windows: the asminc/ directory of the directory where the ca65.exe resides

So, the platform is important!

spiro-trikaliotis:
Forget it, you told about strace, so this is not Windows...

[...]

PR #1731, opened 2022-04-25
build: properly quote strings passed as cpp macros

rofl0r:
until now, the strings intended to be hardcoded into the binary,
such as directory names and build id, were passed unquoted, which
means they're interpreted by the preprocessor as C tokens, rather
than strings, which can result in all sorts of "interesting"
behaviour such as interpreting paths starting with // as C++-style
comment.
this was then worked around using a stringize macro which turned
the tokens into a string (if they happened to be in a compatible
format).

adresses #1726

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -68,9 +68,9 @@ $(info BUILD_ID: $(BUILD_ID))
 
 CFLAGS += -MMD -MP -O3 -I common \
           -Wall -Wextra -Wno-char-subscripts $(USER_CFLAGS) \
-          -DCA65_INC="$(CA65_INC)" -DCC65_INC="$(CC65_INC)" -DCL65_TGT="$(CL65_TGT)" \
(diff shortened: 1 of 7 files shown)


mrdudz:
Is the STRINGIFY macro still used for anything after that patch? If not - it should be removed in this patch, too :)

Other than that it looks good to me

rofl0r:
> Is the STRINGIFY macro still used for anything after that patch? If not - it should be removed in this patch, too :)

done. i think this is better kept as a spearate commit to not "do two things in one commit"

mrdudz:
perfect!

If noone objects i will merge this soon

rofl0r:
> If noone objects i will merge this soon

cool. i actually think there's nothing objectionable here, the holy win32 pathsearch is not affected by this.

mrdudz:
fine, merging...

PR #1734, opened 2022-04-25
[Telestrat] Kbhit fix

jedeoric:
There is a wrong behavior reported from someone.

When kbhit() is called, cgetc() returns the char.

In telestrat target, there is a keyboard buffer. When we get the char, this char is removed from this buffer. Cgetc() does not work.

This PR is done in order to fix this. If kbhit had been called, it set the char into ram (store_char). If cget is called and store_char is equal to 0, it calls Telemon keyboard routine, if it's a different value, it means that kbhit was called previously and cgetc returns store_char

@[greg-king5](https://github.com/greg-king5) greg-king5 [22 hours ago](https://github.com/cc65/cc65/pull/1728#discussion_r857194457)
It's possible that kbhit() could be called several times before cgetc() is called. Therefore, kbhit() should not call Telemon if a character already is waiting.

=> Fixed

[greg-king5](https://github.com/greg-king5) [22 hours ago](https://github.com/cc65/cc65/pull/1728#discussion_r857195483)
If you move that variable to "cgetc.s", then kbhit() won't be linked into programs if it actually isn't called.

("samples/ascii.c" calls cgetc() and doesn't call kbhit().)

=> Fixed

### DIFF
--- a/libsrc/telestrat/cgetc.s
+++ b/libsrc/telestrat/cgetc.s
@@ -5,24 +5,38 @@
 
     .import    cursor
 
+    .export    store_char
+
     .include   "telestrat.inc"
 
 .proc _cgetc
-        ; this routine could be quicker if we wrote in page 2 variables,
-        ; but it's better to use telemon routine in that case, because telemon can manage 4 I/O
-        ldx     cursor       ; if cursor equal to 0, then switch off cursor
+        ; This routine could be quicker if we wrote in page 2 variables,
+        ; But it's better to use telemon routine in that case, because telemon can manage 4 I/O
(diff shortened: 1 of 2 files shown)


libsrc/telestrat/cgetc.s
```diff
@@ -5,24 +5,38 @@
...
+        ldx     #$00
+        rts
+@out:
+        ldx     #$00
+        stx     store_char
         rts
 .endproc
+.data
+store_char:
+        .byte 0
```
mrdudz, line 42: You can save a byte by putting this into .bss (which is always initialized to 0 by crt0)

mrdudz:
seems fine, merging. Thanks!

PR #1735, opened 2022-04-27
build: allow empty prefix

rofl0r:
there was some concern that this will break windows' way of doing
file lookups relatively from the binary, rather than via hardcoded
locations, but so far each occurence adding e.g. "CA65_INC" to
the pathsearch is already shielded with an `#ifndef _WIN32.`

addressing #1726

### DIFF
--- a/src/common/searchpath.c
+++ b/src/common/searchpath.c
@@ -263,11 +263,14 @@ static char* GetProgPath(char* pathbuf, char* a0)
 #endif
 
 void AddSubSearchPathFromBin (SearchPaths* P, const char* SubDir)
-{
-/* Windows only:
-** Add a search path from the running binary, adding a subdirectory to
+/* Add a search path from the running binary, adding a subdirectory to
 ** the parent directory of the directory containing the binary.
+**
+** currently this will work on POSIX systems and on Windows. Should
+** we run into build errors on systems that are neither, we must add
+** another exception below.
 */
+{
     char* Ptr;
     char Dir[PATH_MAX];
 
@@ -277,7 +280,7 @@ void AddSubSearchPathFromBin (SearchPaths* P, const char* SubDir)
         return;
     }
 
-#else
+#else /* POSIX */
 
     GetProgPath(Dir, ArgVec[0]);
 


rofl0r:
@oliverschmidt would you mind testing the win32 CI build artefacts whether my change broke win32 pathsearch ?


ah, it seems the CI was relying on some "build-in-tree prefix behaviour" for running the testsuite without make install, i figure ?

mrdudz:
> it seems the CI was relying on some "build-in-tree prefix behaviour" for running the testsuite without make install, i figure 

yes. i am not sure how critical it is tho preserve this behaviour - or if we should fix the makefiles for the tests and samples instead (they should all work in-tree without having to install, on both windows and linux)

src/Makefile
```diff
@@ -26,7 +26,7 @@ PROGS = ar65     \
 .SUFFIXES:
 
 bindir  := $(PREFIX)/bin
-datadir := $(if $(PREFIX),$(PREFIX)/share/cc65,$(abspath ..))
+datadir := $(PREFIX)/share/cc65
```
spiro-trikaliotis, line 29: This patch breaks the ability to run cc65 without installing it. That is, just compile it in-place and let it run there.

I am not sure if this feature is really needed, though.
However, one side-effect of this is that "make test" does not work as long as cc65 is not installed yet, because it does not find its files. That's why the $(abspath ..) was there.

groessler: I regularily run cc65 from its build directory. In fact, I think I haven't "installed" it for 15 years (at least). Please preserve that possibility!

rofl0r: no worries, this PR was made to see "whether something breaks, and what specifically"

mrdudz: > I regularily run cc65 from its build directory. In fact, I think I haven't "installed" it for 15 years (at least). Please preserve that possibility!

Now that you say it - i don't think i have *ever* installed it :=)

rofl0r:
> then extend [AddSubSearchPathFromWinBin](https://github.com/cc65/cc65/blob/ba5859f7243076a90fd0eaa8376b85911b7a75b3/src/common/searchpath.c#L162)

unlike windows, there's no portable way to get the exe name except by accessing argv[0] passed to main(). therefore i added some code to the cmdline parser unit, and it seems to work now. the application root path is exported as CC65_HOME, which is already used for pathsearch.

mrdudz:
> > then extend [AddSubSearchPathFromWinBin](https://github.com/cc65/cc65/blob/ba5859f7243076a90fd0eaa8376b85911b7a75b3/src/common/searchpath.c#L162)
> 
> unlike windows, there's no portable way to get the exe name except by accessing argv[0] passed to main(). therefore i added some code to the cmdline parser unit, and it seems to work now. the application root path is exported as CC65_HOME, which is already used for pathsearch.

Please look at the code i linked to - i feel like doing something equivalent in AddSubSearchPathFromWinBin is a lot cleaner than abusing CC65_HOME for this. CC65_HOME is deprecated in the first place, and might get removed sooner or later. And i dont want to introduce another thing that is entirely different in windows vs linux.

src/Makefile
polluks: By the way GNU says `make check` should work without installation.
However we called this target "test". Maybe it's time to be standard conform.

rofl0r:
implemented as requested. how does it look to you now ?

mrdudz:
> it finds the absolute path of a binary that was started from the shell, using $PATH. for example if PATH is /bin:/sbin:/usr/local/bin, and ca65 was started using just `ca65` from the shell and resides in /usr/local/bin, it will return /usr/local/bin/ca65.

But why? How is that even useful? That function should return the path to executable when it was started in the sourcetree - anywhere. There is most likely no PATH pointing at it in this case. If there is no path in argv[0] it should rather prepend the current working directory - again, see the code i linked.

rofl0r:
> But why? How is that even useful? That function should return the path to executable when it was started in the sourcetree - anywhere. There is most likely no PATH pointing at it in this case (the binary is likely not installed). If there is no path in argv[0] it should rather prepend the current working directory - again, see the code i linked.

this code was written to always find the binary no matter how it was started. i use it for other projects too. there's 3 possible ways to start a binary in unix:
1) via PATH
2) via absolute path. in this case argv[0] starts with '/'.
3) via relative path. this for example happens when you start the binary from source tree like this `./cc65` or `bin/cc65`

checking these 3 cases always works, and is portable across all POSIXy systems. unlike the /proc method you propose, which is platform specific, and doesn't work if /proc isn't mounted (for example inside a container or bubble-wrapped chroot).

this is why i went with this route to begin with when i had to implement  the functionality for other programs i wrote. the path-search code is from 
https://github.com/rofl0r/debuglib/blob/b519f6a59fa25aa564a952ace5c6a690c91f96d5/examples/idfake.c .

i could leave away the path search, but then the functionality would work only for the "run relative from builddir" case, but not universally.

mrdudz:
> * via PATH
> * via absolute path. in this case argv[0] starts with '/'.
> * via relative path. this for example happens when you start the binary from source tree like this `./cc65` or `bin/cc65`

OK - as said that code seriously lacks comments

mrdudz:
I took the opportunity to update contributing.md a bit (sorry, i know it is a bit annoying when new rules appear in the middle of the process), some other things i noticed:

- in pointer declarations/definitions the asterisk should stick to the type, not the variable, ie ```char* foo```
- there should be an empty line between variable declarations and the code
- there should be a comment for each function explaining what it does and what parameters and return type are - and that comment goes between function declaration and body (probably somewhat unexpected/unusual)
- if statements should be always written using braces, ie instead of
```
if (foo) bar();
```
write
```
if (foo) {
    bar();
}

**Comment by mrdudz** [comment]
> I like it. Using /proc if available might give a small speedup, but it's good that it works generally. One point: On Windows one can use `\` and `/` as directory separator (not in cmd.exe though). Maybe Windows should check both `\` and `/`.

Please - one thing at a time.

Actually, i think on windows we can really just replace any ```\``` by ```/``` in strings we "see" in environment variables or so, and use ```/``` in the code everywhere else (that may or may not simplify some code)

**Comment by mrdudz** [comment]
> > in pointer declarations/definitions the asterisk should stick to the type, not the variable, ie char* foo
> 
> that's ugly :-(

Please lets not discuss the codestyle here - this rule existed before and its what the code uses.

**Comment by groessler** [comment]
> Actually, i think on windows we can really just replace any \ by / in strings we "see" in environment variables or so, and use / in the code everywhere else (that may or may not simplify some code)

I think so too

**Comment by groessler** [comment]
> > > in pointer declarations/definitions the asterisk should stick to the type, not the variable, ie char* foo
> > 
> > 
> > that's ugly :-(
> 
> Please lets not discuss the codestyle here - this rule existed before and its what the code uses.

Ok. Didn't know that it existed before.

**Comment by rofl0r** [comment]
i think the style should be good now. if you have the same opinion, let me know so i can squash the style fixups into the "[implement AddSubSearchPathFromWinBin() counterpart for unix](https://github.com/cc65/cc65/pull/1735/commits/621deced6d9e94333132ad4126d55f41c8ebf962)" commit.

**Comment by polluks2** [comment]
> * if statements should be always written using braces

but one-liner IMHO
@mrdudz where is the right place for discussing code style?

**Comment by mrdudz** [comment]
> > * if statements should be always written using braces
> 
> but one-liner IMHO @mrdudz where is the right place for discussing code style?

Right now i will not discuss it at all - as that traditionally serves little purpose other than spawning flame wars. I will first try to make the written rules more complete by looking at what style is already used in the code. And then whoever proposes changes to the existing style will have to be the one who fixes the entire codebase according to the proposed change. I don't agree with some of the existing style decisions either for that matter - but i even less agree with mixing codestyles in the same codebase, which implies that the written rules must always match the codebase.

File: src/common/searchpath.c [hunk 2]
```diff
@@ -157,18 +163,125 @@ void AddSubSearchPathFromEnv (SearchPaths* P, const char* EnvVar, const char* Su
...
+    }
+    for (;;) {
+        o = buf;
+        l = buflen;
+        while (l && *p && *p != ':') {
+            *(o++) = *(p++);
+            l--;
+        }
+        snprintf (o, l, "/%s", bin);
+        if (access (buf, X_OK) == 0) return 1;
```
mrdudz, line 221: please use braces

mrdudz:
> i think the style should be good now. if you have the same opinion, let me know so i can squash the style fixups into the "[implement AddSubSearchPathFromWinBin() counterpart for unix](https://github.com/cc65/cc65/pull/1735/commits/621deced6d9e94333132ad4126d55f41c8ebf962)" commit.

Besides the one case of forgotten braces it looks ok to me

rofl0r:
> Besides the one case of forgotten braces it looks ok to me

cool, fixed and rebased.

mrdudz:
Added reading from /proc/self first to make the edgy "make avail" work with the make targets left alone for now. @greg-king5 please test if this works for you.

One thing that seems to be a good idea now would be rewriting GetProgPath() so it would return the full path (including the binary name), so the binary name can be removed in all codepathes with a single call to MyDirname() after the ifdef in AddSubSearchPathFromBin(). @rofl0r could you do this? i don't quite understand why GetProgPath() would call it on the (unused) return value but not for the copy in pathbuf. Seems in the current code MyDirname() is not used at all, and pathbuf already removes the binary name?

src/common/searchpath.c
```diff
@@ -296,10 +296,23 @@ void AddSubSearchPathFromBin (SearchPaths* P, const char* SubDir)
     }
     *Ptr = '\0';
 
-#else
+#elif defined(__linux__)
```
rofl0r, line 299: this isn't necessary; in fact it's just cargo culting. 2 years ago when i needed the functionality here i came across the stackoverflow answer mentioning this method, i looked at it in digust and instead went to find a portable method - the one presented in this PR.
this addition here will make not only add a method that works only on a single platform, whereas the other one works everywhere, it also relies on /proc being mounted - which isn't always the case. so it works less reliable, i.e. linux is now at a disadvantage vs e.g. HAIKU or the BSDs.

rofl0r: oh ok, i didn't notice that GetProgPath is still called as a fallback, so i retract the latter.

mrdudz: The fallback is always needed as a last resort - however to make that edgy usecase work, using argv[0] will not do the trick

rofl0r: > using argv[0] will not do the trick

why ?

rofl0r: oh, i see this was added to make `make avail` work, which i didnt study yet. i'll study it in a couple hours, though i'm quite sure it will work even without the /proc hack.

mrdudz: Look at what the "make avail" target does.... it creates symlinks to (only) the binaries in "/usr/local/bin", nothing else is installed. Now when PATH is pointing to /usr/local/bin and you run the binary as "cc65", the code would add "/usr/local/" to the searchpath - but there isnt anything. The binary could still be located in /home/user/cc65/bin and the includes and libs in /home/user/cc65/

mrdudz: > oh, i see this was added to make `make avail` work, which i didnt study yet. i'll study it in a couple hours, though i'm quite sure it will work even without the /proc hack.

As i said before, the only way to make it work is adding some more symlinks (to the lib/include/asminc/cfg dirs) and then compile with PREFIX=/usr/local - but it'd rather not do this, its even more ugly than what it does now. The proc lookup is fine and works for all practical purposes. If you can make it work without doing either - i'd really like to know how :)

rofl0r: without having studied the issue in much detail, i'd think realpath() would solve even the case where "absolute path involving symlinks" is involved.

from man 3p realpath()

```
     The  realpath()  function  shall  derive, from the pathname pointed to by
     file_name, an absolute pathname  that  resolves  to  the  same  directory
     entry,  whose  resolution  does not involve '.', '..', or symbolic links.
```

anyway, i'll be afk for a couple hours, then will get back to this and see whether we can find an elegant solution.

mrdudz: Some quick testing seems to suggest the solution could be:
1) make GetProgPath() always return the path including the binary name
2) call realpath on the returned path
3) strip the binary name from it

Interesting!

That said, we'll still need some ifdef magic around 2) - because not everything that isnt windows is POSIX (AmigaOS for example)
Also: the

rofl0r: i resolved it slightly different - GetProgPath now always passes through realpath(), so we don't need to call it twice in the case of starting the binary using a relative path, like bin/cc65.
but it does return the full binary name now, so we can use the previous existing code to strip it, just like the windows code. please take a look at the last commit.

mrdudz:
This looks good - with one thing left open:

> That said, we'll still need some ifdef magic around 2) - because not everything that isnt windows is POSIX (AmigaOS for example)

I am willing to accept that this feature doesn't work on those systems - however breaking compilation is a bit extreme (especially since until now the toolchain pretty much compiled on anything that has a C compiler).

Unless there is a single symbol we can check to find out if we are on a POSIX system (i don't think there is, is there? iirc they are defined in unistd.h - so we have a chicken and egg problem for those), there are two somewhat ugly ways to do this:

a) put the POSIX stuff in ifdefs that check all POSIX systems we support. that might get ugly soon and might need a long list
b) check first the NON POSIX systems we know about, then fall back to the POSIX code. perhaps put a #warn in there to inform the user about the disabled feature.

I tend to favour b) - simply because the number of non POSIX systems ppl would actually use these days to compile cc65 on is very limited (Amiga is the only one i can think of right now)

(The symbols to check for would be  __amigaos__ and __amigaos4__)

@rofl0r what do you think?
@polluks once its implemented, can you check on amiga?

rofl0r:
> b) check first the NON POSIX systems we know about, then fall back to the POSIX code. perhaps put a #warn in there to inform the user about the disabled feature.

indeed; it makes much more sense to add an exception for the exception, instead of having to maintain a list of regular systems.

which compiler is used to create the amiga binaries? in case it's @bebbo's https://github.com/bebbo/amiga-gcc someone who has built it can just run `amiga-gcc -dM -E - < /dev/null` and look at the list of preprocessor macros it spits out so we can identify it.

mrdudz:
> which compiler is used to create the amiga binaries? 

STORMC, VBCC or GCC

rofl0r:
> STORMC, VBCC or GCC
> Not sure it's a good idea to identify the compiler rather than the platform though :)

right, but the compiler needs to provide some sort of platform detection macro; maybe/hopefully there is one all 3 compilers agreed upon

rofl0r:
> right, but the compiler needs to provide some sort of platform detection macro; maybe/hopefully there is one all 3 compilers agreed upon

in case there isn't, we might document that an amiga user has to pass -DHAVE_AMIGAOS to USER_CFLAGS when running make.

bebbo:
maybe
```
#if defined __M68K__ || defined __m68k__
```
the first is for vasm the latter for gcc (and maybe storm too)

rofl0r:
> maybe
> #if defined __M68K__ || defined __m68k__

thanks for your suggestion, but this condition is also true for m68k linux, we'd need something more specific

mrdudz:
i asked [here](https://www.a1k.org/forum/index.php?threads/83811/) - hopefully will give us some hints

bebbo:
> > maybe
> > #if defined **M68K** || defined **m68k**
> 
> thanks for your suggestion, but this condition is also true for m68k linux, we'd need something more specific

for the amiga-gcc version there is `__amiga__`

mrdudz:
> and the most used lib also provides getprogname

we wont add specific support for amiga (or even something that only exists in some libs for amiga), that goes a bit too far

mrdudz:
Oh and i just stumbled about [https://cc65.github.io/getting-started.html#GNU](the docs) - this should be fixed/changed as well. The way its outlined in the docs imply that "make avail" is what you _must_ use, which is just wrong (and it doesnt even mention "make install", wth). After this patch here its one of 3 ways you can use cc65 - and certainly the least clean and/or recomended one.

rofl0r:
> uhm, that's POSIX, isn't it?

```
$ man 3p getprogname
No entry for getprogname in section 3p of the manual.
```

doesn't seem to be the case. there's `__progname_full, __progname , program_invocation_name, program_invocation_short_name` which all seem to be various non-portable extensions provided by glibc and the BSDs (they're usually aliased to argv[0]), but i've never seen `getprogname`.

bebbo:
> > uhm, that's POSIX, isn't it?
> 
> ```
> $ man 3p getprogname
> No entry for getprogname in section 3p of the manual.
> ```
> 
> doesn't seem to be the case. there's `__progname_full, __progname , program_invocation_name, program_invocation_short_name` which all seem to be various non-portable extensions provided by glibc and the BSDs (they're usually aliased to argv[0]), but i've never seen `getprogname`.

nm then. it's BSD not POSIX^^

polluks2:
(Some) POSIX support for Amiga:
There's no perfect solution, it depends on the compiler
- gcc uses ixemul
- vbcc uses vbcc_PosixLib
- SAS/C, my first cc65 port, is itself not supported anymore

rofl0r:
> (Some) POSIX support for Amiga:
> There's no perfect solution, it depends on the compiler

interesting. so it's well possible this already works for amiga users. if so, i'd propose to merge this and wait whether anyone complains.

mrdudz:
> (Some) POSIX support for Amiga: There's no perfect solution, it depends on the compiler
> 
> * gcc uses ixemul
> * vbcc uses vbcc_PosixLib
> * SAS/C, my first cc65 port, is itself not supported anymore

So can you test if it works or not? From my experience with amiga things, i somehow doubt it would :P

mrdudz:
So, after evaluating this a bit the result is simply that the situation on Amiga is a lot more messy than predicted, and i am not interested in sorting out this mess right now either.

Should noone object i will merge this PR as is. In the unlikely case that someone really runs into build problems we can fix it then - at least it will show us if there are actually ppl using cc65 on those obscure systems :)

PR #1742, opened 2022-05-05
Mkdir : fix registers calls

jedeoric:
This PR enhances mkdir calls for Orix and Telestrat target.

AY is the string, and X is the mode : 0 for X means that only one level of the folder will be created


polluks2:
Maybe _sysmkdir should use fastcall?

jedeoric:
I will open a new PR to fix it

polluks2:
You may also fix apple2/sysmkdir.s and atari/sysmkdir.s

PR #1745, opened 2022-05-17
Fix set but unused variable

Compyx:
In ReadOptStats() the variable Lines is used to count lines, but nothing is done with the result.


mrdudz:
good find :)

PR #1748, opened 2022-05-18
Fix switched meaning of '\n' and '\r' for osic1p target

### LINKED ISSUE #1747: Wrong encoding of CR and LF in conio for the osic1p target
According to ASCII code 0x0A is LF and code 0x0D is CR.

As an example the c64 conio implementation handles CR and LF like this:

https://github.com/cc65/cc65/blob/93ee91fb4b339337fae6ea99e025858a48acc988/libsrc/c64/cputc.s#L23-L30

Compare that for example to the implementation for nes:

https://github.com/cc65/cc65/blob/93ee91fb4b339337fae6ea99e025858a48acc988/libsrc/nes/cputc.s#L31-L38

It looks like the c64 conio library implements an incorrect encoding of CR and LF.

The following targets have this problem: c128, c16, c64, cbm510, cbm610, osic1p, pet, plus4, vic20

smuehlst:
When conio was implemented for the osic1p target, the cbm
target was used as a blueprint. But it was overlooked that the
cbm target encodes CR as 0x0A and LF as 0x0D, while the
osic1p target uses the ASCII encoding of CR and LF. Therefore
conio output of '\n' moved the active position to the start of the
line and '\r' moved the active position to the next line.

This change implements the correct semantics of '\n' and '\r'
in conio for the osic1p target.

Fixes #1747.

### DIFF
--- a/libsrc/osic1p/osiscreen.inc
+++ b/libsrc/osic1p/osiscreen.inc
@@ -78,13 +78,13 @@ _cputcxy:
 
 ; Plot a character - also used as internal function
 
-_cputc: cmp     #$0A            ; CR?
+_cputc: cmp     #$0D            ; CR?
         bne     L1
         lda     #0
         sta     CURS_X
         beq     plot            ; Recalculate pointers
 
-L1:     cmp     #$0D            ; LF?
+L1:     cmp     #$0A            ; LF?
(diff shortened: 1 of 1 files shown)


mrdudz:
OK :) Thanks!

PR #1755, opened 2022-05-30
cl65: fix regression in --print-target-path

### LINKED ISSUE #1754: What happened to "cl65 --print-target-path"?
I used to follow the pattern

`java -jar $(AC) -p  $@ program.system sys < $(shell cl65 --print-target-path)/apple2enh/util/loader.system`

in Apple][ makefiles.

Now I get

`/share/cc65/target` rather than an absolute path.

Bisection lands here: https://github.com/cc65/cc65/commit/8669710c0f9909f8b3b825e3d0693b98563979fd

This seems to be broken too:

https://github.com/oliverschmidt/A2Stream/blob/main/Makefile

How is it supposed to work now?

rofl0r:
since the "empty prefix means run from current dir" hack was removed,
cl65 --print-target-path no longer prints the path relative to the
binary, but the hardcoded library path which points to prefix, because
the code adds the hardcoded path first to the pathsearch, but then actually
only returns the first entry rather than looking whether the path exists.

closes #1754

### DIFF
--- a/src/cl65/main.c
+++ b/src/cl65/main.c
@@ -1212,6 +1212,7 @@ static void OptPrintTargetPath (const char* Opt attribute ((unused)),
 /* Print the target file path */
 {
     char* TargetPath;
+    char* tmp;
 
     SearchPaths* TargetPaths = NewSearchPath ();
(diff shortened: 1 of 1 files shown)


mrdudz:
thanks!

PR #1760, opened 2022-06-03
Fixed term #1553

polluks2:


README.md
```diff
@@ -1,7 +1,7 @@
 # About cc65
 
 cc65 is a complete cross development package for 65(C)02 systems, including
-a powerful macro assembler, a C compiler, linker, librarian and several
+a powerful macro assembler, a C compiler, linker, archiver and several
```
spiro-trikaliotis, line 4: Was there any consensus on this?
Yes, ar(1) (of gcc) is called "archiver", because it can do more than just a library manager (like LIB.exe from Microsoft).

However, despite its name, ar65 is more like lib.exe than ar(1), because it is not that generic as ar(1), so "librarian" or "library manager" fits much better to it, IMHO.

spiro-trikaliotis, line 4: Was there any consensus on this?
Yes, ar(1) (of gcc) is called "archiver", because it can do more than just a library manager (like LIB.exe from Microsoft).

However, despite its name, ar65 is more like lib.exe than ar(1), because it is not that generic as ar(1), so "librarian" or "library manager" fits much better to it, IMHO.

polluks2: Because of this definition
https://github.com/cc65/cc65/blob/dcdf7ade0829a1a62bcb1a38f11dcf60722aad9c/doc/index.sgml#L12

spiro-trikaliotis: Ok, I see. Yes, consistency is important here.
I still don't like the name archiver, because ar65 is not as versatile as ar, but if it is the consensus to do so, then this is the way to go.

mrdudz:
I prefer "archiver" myself, simply because the tool is called ar65 :) In any case, being consistant is the much more important thing here, so merging

PR #1770, opened 2022-06-16
lynx: fix TGI_COLOR_DARKBROWN

devkev:
Commit 4e406b744 in #1686 (see also discussion in #1584) removed the definition of TGI_COLOR_DARKBROWN, however the merge commit e0cc3650e seems to have added it back in (perhaps accidentally?).  Unfortunately, TGI_COLOR_DARKBROWN refers to COLOR_DARKBROWN, which is what was removed (to make space for COLOR_TRANSPARENT) by 4e406b744.  This means anything referring to TGI_COLOR_DARKBROWN will fail to compile.

This PR updates TGI_COLOR_DARKBROWN to map instead to COLOR_BROWN.  This seems better (to me) than removing it, since that would break backward compatibility with existing code using TGI_COLOR_DARKBROWN.

OTOH, since this approach will cause the actual on-screen color of TGI_COLOR_DARKBROWN to change, perhaps it would be better to break existing code, and thereby force any devs currently using TGI_COLOR_DARKBROWN to figure out the most appropriate solution for their case - whether that is changing their code to use TGI_COLOR_BROWN (or some other color) instead of TGI_COLOR_DARKBROWN, or using a custom palette.  Let me know if this approach is preferred instead, or feel free to just make that change (ie. removing TGI_COLOR_DARKBROWN completely) yourself before squashing/merging.

cc @karrika


karrika:
I suggest deleting references to darkbrown completely as it is not available anymore. As I am on the road and will lose internet for 10 days I cannot participate in the PR.

mrdudz:
> I suggest deleting references to darkbrown completely

done, and merged

PR #1771, opened 2022-06-16
Some test cases for ca65

spiro-trikaliotis:
Started with some test cases for ca65.


mrdudz:
Looks good!

However, i missed this before, i see a like using "tail" in that Makefile - can we get rid of this? Makefiles are supposed to work in cmd.exe (ie windows) too :)

spiro-trikaliotis:
Yes, of course, we can get rid of it.
I have the following ideas:

1. Add an option to ca65 to *not* print the header, or
2. Add an option to ca65 to output a "non-versioned" header line (no version no git revision, that is)
3. add an option to isequal so it skips one line of a file

It's the line with the version information that tail is removing in the listing file, that's why I need it.

Do you agree? Any other option?

mrdudz:
i like 3 best, it seems simple enough :) (make it "skip n lines", that may be handy later)

spiro-trikaliotis:
I'm getting into another problem: ld65 prints out lines that include a path, for example, to none.cfg. Thus, my reference files are not correct.
I will have to come up with a solution to ignore these. One option could be to include some marker into the reference file that is given to isequal that says "here comes something that I am not interested in".

mrdudz:
is this ready?

spiro-trikaliotis:
It can be added, yes.
Of course, it is not complete. However, I don't have much time to work on it at the moment.

As the tests are valid and work, I think it is better to add them (and add another PR once more tests are on their way).

mrdudz:
ok

PR #1777, opened 2022-06-20
Floating point support

mrdudz:
this is NOT ready for merge yet.

please read the readme at https://github.com/mrdudz/cc65/tree/fptest/libsrc/float/

any help appreciated, ideally via PR to the above branch


mrdudz:
Are you talking about a floating point library written in C? that is of limited interest... technically we'll need something like that in the future to implement the builtin operators without relying on the hosts floating point implementation - but is like 3 steps further :)

bereiden:
Yes, the algorithms are written in C, primarily for testing and accuracy performance measurements.  My "vision" (for lack of a better term) would be to use the macro assembler to hand compile the C sources, mainly for size and speed reasons.  Currently, all of the operations within the algorithms are either adds/subtracts or bit shifts so it should be relatively easy to write macros to perform those tasks.  I did run cc65 on a portion of the code and it looks like there would be plenty of opportunities to tune up the resultant assembly code.  That being said, my 6502 chops are no where near others who contribute to this project, so that is where a team approach would be beneficial.

I understand you're at the first stages of implementation, but maybe when the time comes, you might want to look over the code I'm developing.  I would think we would want to natively support the functions in <math.h>, which is my ultimate goal.  That, and I was curious how to implement these functions in an efficient manner -- one of my many faults.

As an example of what I am working on, here is the routine that divides two floats:

```
/***************************************************************************************************
**  float32_div - Division
***************************************************************************************************/

float32 float32_div (float32 a, float32 b) {

//  Save input parameter values to the temporary storage locations

    fa = a;
    fb = b;

    if (float32_check () == 0) return fa;

//  Check for division by 0

    if (fb & (~SIGN_MASK) == 0) return qNAN;

    sa = sa ^ sb;                              //  Compute sign bit of result
    xa = MAX_EXPONENT + xa - xb;               //  Compute exponent of result

//  Improve results for subnormals by shifting the a input as far left as possible

    while ((fa & IMPLIED_ONE) == 0) {
       fa <<= 1;
       xa  -= 1;
    }

//  For the division algorithm below, the b input must be greater than or equal to the a input

    while (fb < fa) {
       fb <<= 1;
       xa  += 1;
    }

//  Check for data over/underflow

//  Underflow:  if (((xa - 127) - (xb - 127)) < -126)   => (xa - xb) < 128
//  Overflow:   if (((xa - 127) - (xb - 127)) >  127)   => (xa - xb) > 381

    if (xa < (EXPONENT_BIAS - FRACTION_BITS - 1)) {  //  Underflow?  Return ±0
       fa = (sa == 0) ? 0 : SIGN_MASK;
       return fa;
    }
    if (xa > 3 * EXPONENT_BIAS) {                    //  Overflow?  Return ±INF
       fa = (sa == 0) ? EXPONENT_MASK : EXPONENT_MASK | SIGN_MASK;
       return fa;
    }

//  Perform the multiplication using the following algorithm:
//     x = fa;  y = fb;  z = 0;
//     for (idx = 0; idx <= 32; i++) {
//        if (x > 0) {
//           x = x - y*2^(-i);
//           z = z +   2^(-i);
//        } else {
//           x = x + y*2^(-i);
//           z = z -   2^(-i);
//        }
//     }
//     return z;

    iter = IMPLIED_ONE;
    tmp  = fa;
    fa   = 0;

    for (idx = FRACTION_BITS + GUARD_BITS + 1; idx != 0; idx--) {
       if ((tmp & SIGN_MASK) == 0) {
          tmp -= fb;
          fa  += iter;
       } else {
          tmp += fb;
          fa  -= iter;
       }
       fb   >>= 1;
       iter >>= 1;
    }

//  Adjust the result for subnormals

    while (xa < (EXPONENT_BIAS + 1)) {
       xa++;
       fa >>= 1;
    }

    xa -= EXPONENT_BIAS;

    float32_normalize ();

    return fa;
}
```

mrdudz:
While i understand writing the math library is the most interesting part for you - it's the least needed part for what we want to do here, unfortunately :)

That said, i still see some potential use for the stuff you are doing:

- at some point the compiler will have to use internal fp functions, so it can work on platforms that use a different native implementation than ieee754. This requires the basic math primitives, type conversion, comparisons (see the readme in the branch i linked). This is not terribly urgent right now, as we can (for the time being) get away with assuming the host computer uses 754 floats too.
- much more needed is a runtime library for the target. this contains the same basic primitives. having this code available as C code might be interesting as a reference - or even to build a quick'n' dirty (and very slow =D) library for the target for testing purposes. Obviously once we have this and proved working, functions can be rewritten in optimized assembly one by one.
- last not least there is the math library. this contains all the trig functions etc. for this its important to understand that for the time being it is 100% optional and before we have the other things mostly working, having a math library is a bit pointless (eg because we cant even test it before the rest works). For this part having the functions in clean C would be really useful and would save a lot of time, i guess :). However, as for the primitives, having this stuff in clean C would probably help with creating the assembly versions.

If you want to help pushing this forward, it would be nice to have the basic primitives (type conversions, add/sub/mul/div and a few more - again see the linked branch) in C, so we can build a temporary runtime with them - which may speed up the work on the compiler, since then you can test things on the console using sim65 (right now i'll have to use the c64 binary and -emulator). It would also be a requirement for adding proper regression tests, which is very much needed for this kind of thing :) Ideally make a new directory for your stuff in the test/float directory, put your stuff in there - preferably one file per function - and then send a PR to my branch over there :)

BTW, as for "most efficient implementation" - for our usecase that would almost certainly mean "the smallest implementation". I am pretty impressed by the CBM kernal code now for that matter, which squeezes not only the basic primitives, but also a complete math library, into a few kilobytes.

mrdudz:
For whoever might be following this: another "milestone" was done: the "softfloat" lib is now compiled into the regular library, and a bunch of tests work in the regression test suite. From now on it is mostly about bugfixing (in the compiler).

acqn:
Will the debugging/logging stuff be kept or removed when this is finished? My AST branch would conflict badly with them. XD

mrdudz:
I can rip that out again :) Just helps me a bit to get familiar with what is going on :)

mrdudz:
Ok, so check samples/mandelfloat.c ... things start working :)

acqn:
How will integer <-> floating conversion do with this patch once finished? Will `int` be implicitly converted to `float` or `double` ?

mrdudz:
> How will integer <-> floating conversion do with this patch once finished? Will `int` be implicitly converted to `float` or `double` ?

Not sure i understand the question :) Right now what i am doing is basically looking at what the test snippets produce in gcc, and then try to make cc65 do the same thing :)

But yes, so far i am implicitly converting integers from/to float (there is no double yet). That said this "float" is actually a double internally (which afaik is just fine).

As for "floats vs doubles" i haven't put much thought into it yet, right now i only want to add the datatype :)

I am also planning to merge this "early" for that matter - ie not wait until all and everything is done and working, but rather clean up and merge what i have at a certain point (which is coming closer now...)

mrdudz:
To add on this... i have not even looked at the standard yet - but if it allows to do that, i would prefer "float" being the "main" floatingpoint datatype, ie all implicit conversions go from/to float and if we ever add "double" too, then this would only be used when explicitly cast to.

acqn:
The C standard ~converts~ _promotes_ `float` to `double` in _default argument promotion_ for prototype-less and ellipsis arguments.

mrdudz:
I see, so if a double type exists, then we must do it this way.

BUT its correct that "internally" the compiler is allowed to always use the highest precision possible, right?

mrdudz:
Sooo. In a recent burst of activity i pushed this a little bit further. It is now in a state that things start working, library support is more or less "complete" (as in: implemented in C), it is possible to compile simple programs (with some effort at  working around the remaining bugs), and for someone who wants to help working in a certain area the "chicken and egg" problems have been solved, so (more or less) independent work on eg the library is now possible. I am actually thinking about cleaning everything up now and merge it in this state perhaps - as that makes it more visible and more likely that others will jump on the boat.

More details can be found at https://github.com/mrdudz/cc65/tree/fptest/libsrc/float/

mrdudz:
taking over from #2014

> Are they the one same type (ie. double is just an alias of float) or two distinct types (like short vs int) with the same implementation in that case?

BTW, i'd lean to the later, ie internally have different types, and then the compiler has some (for now always enabled) flag that makes the compiler use the same implementation for both types.

pm100:
@mrdudz - wow glad I checked here. I was about to start on this myself, just for kicks.

Whats the state of play? 

I started with the well known wozniak float code, which is of course not ieee, but at least it works and gets me into all the compiler and library changes

jgbaldwinbrown:
Hi mrdudz,

I just stumbled on this thread and realized I'd been working in private on a similar project. Looking at the work in this thread, it looks like you may have already made my work obsolete, but I thought I'd describe it here in case it had some use. In particular, I think my project might be a way to get 64-bit floating point support into the compiler for doubles without having to implement a 64-bit integer type.

I had the idea to port Berkeley softfloat to cc65 to get floating point support, but I started with softfloat version 2 instead of version 3.  The advantage of version 2 is that it was designed for environments that only have 32-bit integers but still need 64-bit doubles. I modified it to compile and run in cc65 by making all struct passing done with pointers. I've only barely tested it, but it seems to produce correct answers for 32-bit and 64-bit floats.

I don't know if this is of use in your project, but I thought I'd list some of the advantages and disadvantages of using softfloat2 over softfloat3:

### Advantages of softfloat2

- Easy to port (designed for 32-bit environments, float64 is already a struct that contains 2 32-bit fields. I ported it in just a few hours).
- Already functional.
- 64-bit support out of the box.

### Disadvantages of softfloat2

- Older codebase means more errors and fewer optimizations, so code is slower and could be incorrect in edge cases.
- No 16-bit or 128-bit float, which softfloat3 has.

It may be possible to use softfloat2 code just for 64-bit doubles, assuming we can convert between softfloat2 64-bit doubles and softfloat3 32-bit floats.  For what it's worth, I think you could entirely use softfloat2 to implement floats and doubles, but that would mean throwing out the excellent work you've done to port softfloat3, so I'm not really recommending that.

Anyway, what do you think? Would there be value in integrating softfloat2 here? Here's the link to my softfloat2 repository:

[https://github.com/jgbaldwinbrown/cc65_softfloat2](https://github.com/jgbaldwinbrown/cc65_softfloat2)

Cheers,
Jim

mrdudz:
> EDIT: My mistake, I realized you are also using version 2 of Berkeley softfloat. For what it's worth, my 64-bit implementation does seem to be working, so I think it makes even more sense to integrate the changes into your code so we can implement doubles along with floats. What do you think?

Well. So far i have just completely ignored 64-bit - and i am not planning to work on it (either in the compiler or library) anytime soon (i don't even have time right now to work on whats there for that matter).

That said, if you'd make a PR to the fptest branch which adds your 64bit float support to the softfloat library, i will happily merge it (provided it follows the API naming scheme i invented for it)

jgbaldwinbrown:
Awesome! I'll work on putting a pull request together. It will probably take a little while to integrate everything nicely.

mrdudz:
You didn't do any work on the compiler yet by any chance? :)

jgbaldwinbrown:
No, sorry, I'm only now trying to look at the compiler code. I'll do my best to help, though!

mrdudz:
@kugelfuhr if you really have some time to spend - it would be awesome if you could have a look at this :) The main goal right now is to make it so that it will not affect anything that does not use floats, and perhaps fix some idiot things that i did because i have no idea what i am doing there (i really don't). The idea is to merge it ASAP even when it is not complete, so we have the initial momentum and enable others to test and look at it. You may look at floating-point.md to get an idea of what has been done already. (I am sure someone who knows the code better can fix the remaining bits rather easily... however, i won't have the time to even look at it anytime soon - if ever. Even this cleanup now was stretching my time budget :/)

kugelfuhr:
I had a rough look at it which leads to a few questions:

What is the overall goal for this change? Support for many variants of floating point? Portability to non 6502 platforms? Or just one format and 6502 as a target?

What was the reason for choosing the different float formats?

How complete is the support in the compiler? Does it generate (at least mostly) correct code for float arithmetic (+-*/) and float <-> integer conversions? The readme lists some things that do not work but doesn't say what works. Is the non working stuff caused by missing runtime support or by missing support in the compiler?

Why aren't doubles working? Or do they? I got this from the readme, please correct me if I'm wrong. Is there a reason for not supporting double? It is the target type for any floating point default promotions, so no support for double sounds wrong to me (at least when it comes to the C compiler).

It seems that on one side  a lot of effort has been put into supporting a plethora of different float formats while on the other side not even floats and doubles are distinguished in the compiler. Could you please explain a bit about the reasoning behind all this?

Why I'm asking: The changes seem pretty large. They look quite ambitious on one side (float formats/runtime code) and pretty sparse on the other (compiler). Maybe stripping them down is an option. This would improve chances to have them merged without breaking too much stuff.

kugelfuhr:
Please do not edit my posts instead of replying!!

For one, its bad style. Second, part of what I wrote is lost. Third and probably most important: I don't get any notifications from github if you do this. I found your edit when I was wondering why there was no reply and reviewed the issue.

Is there a chance to restore my original post and add your text as a reply as it should be?

mrdudz:
> Please do not edit my posts instead of replying!!
> 
> For one, its bad style. Second, part of what I wrote is lost. Third and probably most important: I don't get any notifications from github if you do this. I found your edit when I was wondering why there was no reply and reviewed the issue.
> 
> Is there a chance to restore my original post and add your text as a reply as it should be?

Oh damn, sorry. That wasn't my intention. Sometimes i seriously hate this web interface :)

mrdudz:
> What is the overall goal for this change? Support for many variants of floating point? Portability to non 6502 platforms? Or just one format and 6502 as a target?

This change: support for IEEE754 32bit floats (only, no doubles yet) for 6502

> What was the reason for choosing the different float formats?

In the future, the compiler might want to support the native floating point format used by the various targets. Right now there is no such thing - also the CBM kernal float library pretends to use IEEE754 format (and converts forth and back internally) right now.

However, this is only something to "keep in mind" when implementing things.

> How complete is the support in the compiler? Does it generate (at least mostly) correct code for float arithmetic (+-*/) and float <-> integer conversions? The readme lists some things that do not work but doesn't say what works. 

Mhh it is... good enough to make the sample programs i added compile and work. And the (admittedly brief) tests i added, of course. So yes, "mostly correct", depending on your definition of "mostly", with some bugs and some open ends :)

> Is the non working stuff caused by missing runtime support or by missing support in the compiler?

Mostly the compiler. The runtime should be pretty complete (even including math.h) - but only for 32bit floats right now. (the softfloat library also supports doubles, should that be needed in a next step)

> Why aren't doubles working? Or do they? I got this from the readme, please correct me if I'm wrong. Is there a reason for not supporting double? It is the target type for any floating point default promotions, so no support for double sounds wrong to me (at least when it comes to the C compiler).

Making doubles work isn't as straightforward, that would mean adding all kind of support for 64bit values/parameters (for 32bit floats i could re-use a lot of the stuff that was used for 32bit integers). For a start it seemed fine to me to assume that double=float (i know i know....). That said, i have not even tested this yet. I don't think its a big problem right now to be honest.

> It seems that on one side  a lot of effort has been put into supporting a plethora of different float formats while on the other side not even floats and doubles are distinguished in the compiler. Could you please explain a bit about the reasoning behind all this?

See above. You can mostly ignore the different float formats right now. I mostly added the CBM float lib because i already had this laying around (however, as said, it pretends to work with IEEE754 - there is no specific support in the compiler, you'd just link another lib). And the lib in float/IEEE754 is just empty templates for the most part - prepared for whoever might want to look at implementing this stuff in assembly.

So basically, "softfloat" is the main runtime code to consider right now, if only because we can be relatively confident that it works correctly - the rest is just extra bonus. However, it doesnt affect anything to have the other stuff there or not :)

kugelfuhr:
Thank you for the cleanup!

Before I talk about the issue itself, let me clear up a misconception. You seem to mix up `float` and `double` used in the C compiler with the IEEE "float" and "double" floating point formats. While many compilers nowadays map the C data types to the IEEE data types, they have nothing to do with each other. At least not as far as the C language is concerned.

A C compiler would be free to implement its `float` and `double` data types both using IEEE "double". It could also choose CBM 40 bit floating point for `float` and IEEE 80 bit floating point for `double`.

This is important since when I was talking about support for the `double` data type of the C compiler, you seemed to imply it means support for the 64 bit IEEE floating point data type - which is a different story.

As far as the compiler is concerned, `float`, `double` and `long double` are just some data types it knows and treats as specified in the language description. `double` is especially important, because it's the type used for default argument promotions. So for example when calling `printf` and passing it a value of type `float` it will get promoted and passed as a `double`. This is similar as with `char` which is always passed as an `int` in an ellipsis or in absence of a prototype.

This means that support for the `double` type is mandatory for a C compiler. You could go without `float` and `long double` and this would just mean something is missing. But without `double` some important parts will simply be wrong.

So while supporting all three floating point data types in the compiler is more or less required, this doesn't mean you have to support three different floating point formats in the library. They are distinct types as far as the compiler is concerned but may be mapped to one, two or three actual floating point formats in the library.

kugelfuhr:
Before I start working on this, let me see if we agree on some design decisions (besides having floating point support in general). I haven't found any design discussions regarding floating point. Feel free to point me to any discussion threads if they exist.

First: While floating point support is desirable, it has to be as small as possible to be useful. If the required routines are too large, they aren't usable at all on some platforms and will severely decrease the available memory on all others.

This has some consequences for the library. A fp library needs quite some space. `printf` support for floating point even more. So it would be advisable not to support more than one floating point format, since this would effectively double the size required for the library routines. With two formats, even when using just `float` in a program, the fp library for `double` gets linked in whenever `printf` is used. This is because the default argument promotion for `float` is `double`, so `printf` will always get a `double` on the stack when a `float` was passed and will have to handle it.

Second: The C standard requires at least 10 digits accuracy for the data types `double` and `long double`. This in turn requires at least an 48 bit floating point data type. For the `float` data type, 32 bits are sufficient. Since supporting more than one floating point data type is a bad idea (see above), this leaves two options:

1. Either use the >= 48 bit data type for `float`, `double` and `long double`, or
2. ditch standard compliance in this case and use the 32 bit floating point implementation for all C floating point types.

I would strongly suggest to go with option 2 for several reasons:

* There is no ready to use 48 bit floating point library.
* One could use 64 bit floats instead but these are even larger and slower.
* A 48 or 64 bit fp library will always be larger than a 32 bit fp library.
* A 6502 is no number crunching platform. So while having floats is nice, the reduced accuracy shouldn't be a problem in many cases.
* The compiler can use existing routines to shuffle around floating point values since they have the same size as a long.

If we agree in these two points, it means that:

1. The implementation in the compiler must distinguish between the floating point data types, but the actual floating point format for all three types is identical. This simplifies the code in the compiler and in the library.
2. Target specific formats are more or less ruled out if you don't want to sacrifice the advantages from 1.

If we still agree on that, here's a third design decision: Which floating point format to use. Your implementation uses 32 bit IEEE floats, but I think this format has some drawbacks:

* No ready to use implementation in 6502 assembler.
* The type supports NAN and INF which is great for some applications but does also need additional checks in most or all routines dealing with floating point. Which in turn increases size and complexity of the library with no real benefit for most users.

So the final question would be if using another floating point format would be acceptable. Or did I overlook some advantages of the IEEE floats?

pm100:
There is , I think, 48 bit floating point in some MS basic versions for 6502

________________________________

mrdudz:
OK let me answer this first, the other will take a bit :)

> Before I talk about the issue itself, let me clear up a misconception. You seem to mix up `float` and `double` used in the C compiler with the IEEE "float" and "double" floating point formats. While many compilers nowadays map the C data types to the IEEE data types, they have nothing to do with each other. At least not as far as the C language is concerned.
[...]
> So while supporting all three floating point data types in the compiler is more or less required, this doesn't mean you have to support three different floating point formats in the library. They are distinct types as far as the compiler is concerned but may be mapped to one, two or three actual floating point formats in the library.

Thanks for those explanations. I have to admit i am not familiar with this stuff at all, even as a user of various C compilers - since i am hardly ever using floats (or doubles :)) in the type of code i am usually writing (and if so, its usually trivial enough that none of those things matter to me).

I am aware that the datatypes in the compiler have nothing to do with the implementation of the floating point format itself. What i didn't know is that float and double being the same format is actually standard compliant (i have seen this option in compilers - but assumed this is a non standard thing). That makes some things easier, perhaps :)

mrdudz:
> Before I start working on this, let me see if we agree on some design decisions (besides having floating point support in general). I haven't found any design discussions regarding floating point. Feel free to point me to any discussion threads if they exist.

Not sure this was ever discussed in detail - if so, that was many years ago on the old mailing list. I don't have a link, sorry-


> First: While floating point support is desirable, it has to be as small as possible to be useful. If the required routines are too large, they aren't usable at all on some platforms and will severely decrease the available memory on all others.

Yes. This is also why support for the ROM routines (eg of CBM machines) is essential, IMHO.


> This has some consequences for the library. A fp library needs quite some space. `printf` support for floating point even more. So it would be advisable not to support more than one floating point format, since this would effectively double the size required for the library routines. With two formats, even when using just `float` in a program, the fp library for `double` gets linked in whenever `printf` is used. This is because the default argument promotion for `float` is `double`, so `printf` will always get a `double` on the stack when a `float` was passed and will have to handle it.

Understood. I'd also propose to implement a separate "printf" family of functions (with float support), which is then only linked when floats are being used (ie it should be part of the extra floating point library). That way the overhead introduced by float support does not end up in programs that do not use floats at all.


> Second: The C standard requires at least 10 digits accuracy for the data types `double` and `long double`. This in turn requires at least an 48 bit floating point data type. For the `float` data type, 32 bits are sufficient. Since supporting more than one floating point data type is a bad idea (see above), this leaves two options:
> 
> 1. Either use the >= 48 bit data type for `float`, `double` and `long double`, or
> 2. ditch standard compliance in this case and use the 32 bit floating point implementation for all C floating point types.
> 
> I would strongly suggest to go with option 2 for several reasons:
> 
> * There is no ready to use 48 bit floating point library.
> * One could use 64 bit floats instead but these are even larger and slower.
> * A 48 or 64 bit fp library will always be larger than a 32 bit fp library.
> * A 6502 is no number crunching platform. So while having floats is nice, the reduced accuracy shouldn't be a problem in many cases.
> * The compiler can use existing routines to shuffle around floating point values since they have the same size as a long.

yes yes yes, all of that :)


> If we agree in these two points, it means that:
> 
> 1. The implementation in the compiler must distinguish between the floating point data types, but the actual floating point format for all three types is identical. This simplifies the code in the compiler and in the library.

yes


> 2. Target specific formats are more or less ruled out if you don't want to sacrifice the advantages from 1.

Well, yes. This was always something i'd consider as "sometime in the future perhaps". For something like the CBM floats it would certainly require (much) more work at the compiler and runtime, even more so if optional.

(This does not rule out supporting ROM functions using a wrapper, like i did)

> If we still agree on that, here's a third design decision: Which floating point format to use. Your implementation uses 32 bit IEEE floats, but I think this format has some drawbacks:
> 
> * No ready to use implementation in 6502 assembler.

True - but is there one for another format? I am not aware of any implementation that we could use, everything i could find could not be used (for one reason or another).


> * The type supports NAN and INF which is great for some applications but does also need additional checks in most or all routines dealing with floating point. Which in turn increases size and complexity of the library with no real benefit for most users.

But we could just ignore this detail, when we are not standard compliant anyway (most people don't use those things either)


> So the final question would be if using another floating point format would be acceptable. Or did I overlook some advantages of the IEEE floats?

I picked IEEE754 floats, because IMHO the big advantages are:
* there is a ready to use library (even if only written in C right now) that "just works" - and which comes with a license that allows us to just use it.
* tests for this format exist, and the format is common and its properties well defined
* the host most likely uses the same format, which reduces the chance of introducing strange (and very hard to track down) bugs and side effects caused by converting forth and back between different formats

Being able to test is a strong point there imho. Not having to write a library before anything works at all as well :)

But generally, sure, a different format is fine - IF that has significant advantages. However "we'll have to implement the entire library from scratch and make sure it works correctly" is a rather big disadvantage, imho.

kugelfuhr:
Ok, thanks for the reply!

> Understood. I'd also propose to implement a separate "printf" family of functions (with float support), which is then only linked when floats are being used (ie it should be part of the extra floating point library). That way the overhead introduced by float support does not end up in programs that do not use floats at all.

I would suggest a different solution using some compiler and linker magic. 

* `printf` is changed to call an additional subroutine via a modifiable vector when it is done parsing format specifiers. This vector will usually point to an `rts`.
* When the compiler detects that `printf` is used with floating points, it emits an `.import __PRINTF_FLOAT_FORMATS__` or whatever into the generated assembly.
*  A library module that exports this symbol is then linked in. This module contains a `.constructor` that sets the `printf` vector to a new routine that understands the additional floating point formats.

This process is automatic and transparent and works for all simple cases. For all others, an additional compiler option and/or `#pragma` can be added that forces the compiler to add the `.import`.

Regarding the library: One example is [Woz floating point library](https://archive.org/details/dr_dobbs_journal_vol_01/page/n207/mode/2up). The copyright needs to be checked but I think that can be worked out. The library has been proven to work reliably and was written in a time when memory was precious, so it is really small. The basic routines (add, sub, multiply, divide and conversion from 16 bit int to float) are less than 300 bytes(!). It does also come with `log` and `exp` functions which can be used to implement many fp functions using simple formulas. For example

```c
log2(x) = log(x) / log(2)
log10(x) = log(x) / log(10)
pow(x, y) = exp(log(x) * y)
sqrt(x) = pow(x, 0.5)
```

There are also other libraries based on the old MSBASIC code.

mrdudz:
> This process is automatic and transparent and works for all simple cases. For all others, an additional compiler option and/or #pragma can be added that forces the compiler to add the .import.

I don't like solutions that work sometimes and sometimes not to be honest - in particular when that adds a bunch of extra complexity. Having to link the float library is simple and clean, and works always. (However, such mechanism to hook into printf could still be used)

> Regarding the library: One example is [Woz floating point library](https://archive.org/details/dr_dobbs_journal_vol_01/page/n207/mode/2up). The copyright needs to be checked but I think that can be worked out

If you find a complete version of it, go ahead. All i could find was kind of incomplete and didn't really work either.

> There are also other libraries based on the old MSBASIC code.

Those can't be used though (because of the copyright) - and i think they are all 40 bits too

kugelfuhr:
> Having to link the float library is simple and clean, and works always. (However, such mechanism to hook into printf could still be used)

This means you have to create and maintain two versions of `printf` - not something I would prefer. But this can be decided later.

> If you find a complete version of it, go ahead. All i could find was kind of incomplete and didn't really work either.

Ok.

mrdudz:
> This means you have to create and maintain two versions of printf - not something I would prefer. But this can be decided later.

No, you misunderstood - printf could still implement the float stuff in subroutines, which are then stubbed out via rts in the regular lib - and the floating point lib would only contain those stub replacements. very similar to how eg the soft80 library replaces a bunch of conio functions.

regarding the Woz float routines - VBCC apparently uses them for 6502 as well - i could not find the source for its libraries though (and the license of VBCC is somewhat limiting, so i didnt look further either). Maybe you have better luck.

kugelfuhr:
> No, you misunderstood - printf could still implement the float stuff in subroutines, which are then stubbed out via rts in the regular lib - and the floating point lib would only contain those stub replacements. very similar to how eg the soft80 library replaces a bunch of conio functions.

Ok.

kugelfuhr:
> @kugelfuhr if you really have some time to spend - it would be awesome if you could have a look at this :)

Well, I would, but there doesn't seem to be a way to fork a fork of a repository in github. At least not if one has already forked the original project. I could create another account but this seems ugly. Any ideas on how to get involved?

rofl0r:
```
git remote add mrdudz https://github.com/mrdudz/cc65
git fetch mrdudz fptest
git checkout fptest
```

kugelfuhr:
> ```
> git remote add mrdudz https://github.com/mrdudz/cc65
> git fetch mrdudz fptest
> git checkout fptest
> ```

This will of course work but doesn't solve the problem. Once I have a copy and made changes, how do I get them back? I'm pretty sure I cannot just push changes (or another branch) into mrdudz' fork. Maybe I'm overlooking something obvious?

mrdudz:
> > ```
> > git remote add mrdudz https://github.com/mrdudz/cc65
> > git fetch mrdudz fptest
> > git checkout fptest
> > ```
> 
> This will of course work but doesn't solve the problem. Once I have a copy and made changes, how do I get them back? I'm pretty sure I cannot just push changes (or another branch) into mrdudz' fork. Maybe I'm overlooking something obvious?

This way you'll make a PR against the cc65 repo, not this fork...? IIRC thats how it is supposed to work on github (its a bit weird and limiting indeed)

I could swear there was a (somewhat tedious) way to fork a forked repo and then send PR to that fork though (which is what you want, i guess) - but i cant find the solution right now either :(

kugelfuhr:
It seems to work if you do not already have a repo with the same name since the fork has always the same name as the original. Forking this fork would create another fork named "cc65" and this doesn't seem to work.

kugelfuhr:
So how do you want me to proceed?

mrdudz:
For me a PR to the original cc65 repo is fine, i won't have time to work on this anyway

rofl0r:
> I'm pretty sure I cannot just push changes (or another branch) into mrdudz' fork.

well, you could if you had r/w privileges on this repo and this pr has "allow edits from maintainers" ticked on mrdudz' end.

though the simple way would be to just open a new pr with your branch

btw you dont need a new account to make another fork on github, you could just as well make a new "organization" using your existing account and then fork the repo to there. though that wouldnt solve the "can't push to someone else's repo" either.

kugelfuhr:
JFTR and to keep you updated: I've had a look at the code. Getting it into shape is unfortunately a lot more work than expected.

One example: [This simple looking line](https://github.com/cc65/cc65/pull/1777/files#diff-72c2270a00b4d57a91ca0e847feef08f438ffbe37e91492d0301d50d587585bcR196) may cause undefined behavior. See  ISO/IEC 9899:1999 (E), 6.3.1.4p1. A working solution needs much more code. Similar things in other places.

GorillaSapiens:
is this still being worked on?  i wanted to look at the changes, so i did a
<pre>
git clone https://github.com/mrdudz/cc65.git
cd cc65
git checkout fptest
git diff master
</pre>
and saw a LOT of stuff unrelated to floats.

it's hard to do a good review when i can't quite tell what is and isn't float related.

i did notice include/_float.h has "radiants".  the correct spelling is "radians".

mrdudz:
> and saw a LOT of stuff unrelated to floats

There shouldnt be, everything is related (mind you, there are added libraries and samples and other things - but its all related). 

I haven't looked at it since September, and as said before, i don't have the time to do it in the foreseeable future.

I have no idea how far @kugelfuhr worked on it in the meantime for that matter

mrdudz:
```
git clone https://github.com/mrdudz/cc65.git
cd cc65
git checkout fptest
git diff master
```
Now i see this again - i never synced this repo with upstream since september, so of course this way you will see all changes (in reverse) made since then. What you have to do is something like
```
git clone https://github.com/mrdudz/cc65.git
cd cc65
git fetch upstream
git checkout master
git merge upstream/master
git checkout fptest
git merge master
git diff master
```
(perhaps you need to set up "upstream" after cloning and point it to the regular cc65 repo, no idea if that is preserved)

GorillaSapiens:
> (perhaps you need to set up "upstream" after cloning and point it to the regular cc65 repo, no idea if that is preserved)

note for anyone following along, looks like upstream is not preserved, you need to
<pre>
git remote -v                                                                                   # to see if it is there
git remote add upstream https://github.com/cc65/cc65.git          # to add if it is not
</pre>

GorillaSapiens:
looking at ./libsrc/float/ieee754/fbnegeax.s

this is just a boolean negation of a float, right?

from https://www.h-schmidt.net/FloatConverter/IEEE754.html i see that
0.0f -> 0x00000000
1.0f -> 0x3f800000

it seems like fbnegeax should test for zero, return 0x3f800000 if true, or return 0 otherwise.

(forgive me, my 6502 asm is rusty and this might have problems.  this is untested)
<pre>
fbnegeax:
    stx tmp1
    ora tmp1
    ora sreg
    ora sreg+1
    bne nonzero

; argument was zero, return 1.0f
    lda #$3f
    sta sreg+1
    lda #$80
    sta sreg
    lda #0
    tax
    rts

nonzero:
; argument was nonzero, return 0.0f
    lda #0
    tax
    sta sreg
    sta sreg+1
    rts
    
</pre>

i don't see an fnegeax.s in libsrc/float/ieee754, but that's trivial, you just flip the sign bit, right?

<pre>
fnegeax:
    sta tmp1
    lda #$80
    eor sreg+1
    sta sreg+1
    lda tmp1
    rts
</pre>

GorillaSapiens:
it would probably help a lot if the stubs in libsrc/ieee754/*.s had a short comment indicating what they're supposed to do.

GorillaSapiens:
is eaxufloat.s supposed to take an unsigned long and convert it to float?
once that works, other things can just call into it
this may work, but i haven't tested it
<pre>
    .include "ieee754.inc"
    .export eaxufloat

    .importzp       sreg, tmp1, tmp2

eaxufloat:
    sta tmp1 ; byte 0
    stx tmp2 ; byte 1
             ; byte 2 is in sreg
             ; byte 3 is in sreg+1

    ; check for special case of 0
    ora tmp2
    ora sreg
    ora sreg+1
    bz  special_zero

    ; general case
    ldy #$97 ; 24 + 127 = 141 = x97  ; Y will now track magnitude

magloop:
    lda sreg+1
    bz slide_left

    cmp #$1
    beq magdone

slide_right:
    iny
    clc
    ror sreg+1
    ror sreg
    ror tmp2
    ror tmp1
    jmp magloop
    
slide_left:
    dey
    clc
    rol tmp1
    rol tmp2
    rol sreg
    rol sreg+1
    jmp magloop

magdone:
    sty sreg+1  ; this gets shifted down later
    lda tmp1
    and #$1
    tay         ; Y now tracks rounding

    clc
    ror sreg+1
    ror sreg
    ror tmp2
    ror tmp1

                ; do we need to round?
    tya
    ora tmp1
    bz  finish

round_up:       ; IEEE 754 rounds to the nearest even
    lda tmp1
    clc
    adc #$1
    sta tmp1
    lda tmp2
    adc #$0
    sta tmp2
    lda sreg
    adc #$0
    sta sreg
    lda sreg+1
    adc #$0
    sta sreg+1

finish:
              ; sreg and sreg+1 are already set
    lda tmp1  ; restore A / byte 0
    ldx tmp2  ; restore X / byte 1
    rts

special_zero: ; 0x00000000
    lda #$0
    tax
    sta sreg
    sta sreg+1
    rts
</pre>

mrdudz:
> it would probably help a lot if the stubs in libsrc/ieee754/*.s had a short comment indicating what they're supposed to do.

please refer to floating-point.md

https://github.com/cc65/cc65/blob/3cfcdf94d1de143c0480c360476c928223bf9bf4/Floating-point.md

also note that  ./libsrc/float/ieee754/ is not really working right now :)

PR #1781, opened 2022-06-24
Telestrat : change default start address

jedeoric:
This change is to set to $1000 the default address, and now --start-addr works from command line


mrdudz:
looks good, thanks!

PR #1785, opened 2022-07-04
[cc65] Avoid generating asm from C code that has no observable effects

acqn:
This just avoids unneccessary loading the value of the expression of the statement. It doesn't avoid all unneccessary code generation.


mrdudz:
Would be nice to have some kind of test for this... not that i have an idea how :)

mrdudz:
no complaints for 2 weeks (sorry for the delay) - merging

Issue #1489, opened 2021-05-06
Macro using __AX__ gets wrong data from pointer in cc65 2.19; works in cc65 2.18


demesos:
I experienced an issue with cc65 V2.19 where a macro using the `__AX__=` syntax got the wrong data when fed a `char *`.
This example program works when compiled cc65 V2.18  - Git 1074d35 by outputting "hello world" to the screen on the C64 target:
```C
#include <stdint.h>

// convert the printable characters from ASCII to screen code (simplified, incomplete version)
#define toscreencode(C) (__AX__ = (C),   \
                         asm("and #$3f"),\
                         __AX__)   

void main (void)
{
  char *p = "hello world";
  uint16_t scaddr=1024+11*40;	//output to line 11 on the screen

  while (*p != 0) {
    (*(unsigned char *)(scaddr++)=(toscreencode(*p)));
    p++;
  }    
}
```
But when compiled with  cc65 V2.19  - Git 216bb22, the macro receives the values `B9 BA BB BC` and so on, which looks like index values instead of the actual data.

greg-king5:
In order to allow better optimization, post-increment and post-decrement operations are delayed until cc65 reaches a C "sequence point".  Normally, that's at a semicolon.  But, the comma operator also is a sequence point.  Therefore, the increment of `scaddr` is done _after_ `(C)` is loaded, and _before_ the Assembly instruction -- the value in `__AX__` becomes invalid.

You can avoid that problem by making an extra sequence point before loading into `__AX__`:
```c
#define toscreencode(C) (__AX__, __AX__ = (C),   \
                         asm("and #$3f"),\
                         __AX__)

    (*(unsigned char *)(scaddr++)=(toscreencode(*p)));
```
(cc65 2.18 didn't delay post-counts.)

oliverschmidt:
Thanks for the analysis and explanation :-)

demesos:
Thanks for the explanation, but why is *p considered a post-decrement/post-increment operation? The post-decrement operation on scaddr++ is not the part that fails.

demesos:
Thanks for digging into it and for the explanation. The code fix works, but it would require the coder to know about this issue any time a macro is written that potentially can be used with a post-x operation before. Would it make sense and be possible to have the compiler introducing an automatic sequence point before every __A__= and __AX__=? Not sure if this can cause other issues though.

oliverschmidt:
[...]

PR #1799, opened 2022-07-21
[cc65] Post-inc/post-dec won't affect __A__/__AX__/__EAX__ in the same statement

acqn:
Restores the old behavior (#1489) of postfix-increment/decrement operations.

The aim: if a pseudo primary register variable i.e. `__A__`, `__AX__` or `__EAX__` is used, any postfix-increment/decrement operations in the same _full expression_ will stay from tampering the content in them.

Due to some technical limitation, the range where the pseudo primary register variables affect postfix-increment/decrement operations is extended to the whole _statement_ rather than the supposed _full expression_, which could result in worse code generation and/or optimization. This drawback could be resolved in future.

### DIFF
--- a/src/cc65.vcxproj
+++ b/src/cc65.vcxproj
@@ -110,6 +110,7 @@
     <ClInclude Include="cc65\scanner.h" />
     <ClInclude Include="cc65\scanstrbuf.h" />
     <ClInclude Include="cc65\segments.h" />
+    <ClInclude Include="cc65\seqpoint.h" />
     <ClInclude Include="cc65\shiftexpr.h" />
     <ClInclude Include="cc65\stackptr.h" />
     <ClInclude Include="cc65\standard.h" />
@@ -190,6 +191,7 @@
     <ClCompile Include="cc65\scanner.c" />
     <ClCompile Include="cc65\scanstrbuf.c" />
     <ClCompile Include="cc65\segments.c" />
+    <ClCompile Include="cc65\seqpoint.c" />
     <ClCompile Include="cc65\shiftexpr.c" />
     <ClCompile Include="cc65\stackptr.c" />
     <ClCompile Include="cc65\standard.c" />
--- a/src/cc65/expr.c
+++ b/src/cc65/expr.c
@@ -31,6 +31,7 @@
 #include "macrotab.h"
 #include "preproc.h"
 #include "scanner.h"
+#include "seqpoint.h"
 #include "shiftexpr.h"
 #include "stackptr.h"
 #include "standard.h"
@@ -677,6 +678,7 @@ void DoDeferred (unsigned Flags, ExprDesc* Expr)
     int         I;
     unsigned    Size = 0;
     int         Count = GetDeferredOpCount ();
+    unsigned    StmtFlags = GetSQPFlags ();
 
     /* Nothing to be done */
     if (Count <= 0) {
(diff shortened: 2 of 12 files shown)


mrdudz:
would be nice to have a simple testcase for this... like the one in #1489 but for use in the testbench

mrdudz:
here is the testprog i came up with...

Whats really funky... i noticed that i can trigger strange problems by using the now commented out memsets. @acqn before i merge this PR, would be nice if you could check.

```c
#include <stdlib.h>
#include <stdio.h>
#include <stdint.h>
#include <string.h>

#define testasm1(C) (__AX__ = (C),   \
                     asm("and #$3f"),\
                    __AX__)

#define testasm2(C) (__A__ = (C),   \
                     asm("and #$3f"),\
                    __A__)

uint8_t src[32] = { 0x10, 0x41, 0x62, 0x83, 0xb4, 0xf5,  0xe6,  0xc7, 0, 0 };
uint8_t src2[32] = { 0xc1, 0xc2, 0xc3, 0xc4, 0xc5, 0xc6,  0xc7,  0xc8, 0, 0 };


uint8_t ref1[32] = { 0x10, 0x01, 0x22, 0x03, 0x34, 0x35,  0x26,  0x07, 0, 0 };
uint8_t ref2[32] = { 0x01, 0x02, 0x03, 0x04, 0x05, 0x06,  0x07,  0x08, 0, 0 };

uint8_t dest[32];
int res = 0;

void dotest1a(uint8_t *s, uint8_t *d)
{
    printf("dotest1a\n");
    while (*s != 0) {
        *d = (testasm1(*s));
//        printf("%04x:%02x\n",d,*d);
        d++;
        s++;
    }
}

void dotest1b(uint8_t *s, uint8_t *d)
{
    printf("dotest1b\n");
    while (*s != 0) {
        *d = (testasm2(*s));
        d++;
        s++;
    }
}

void dotest2a (void)
{
  char *p = &src2[0];
  uint16_t scaddr=&dest[0];     //output to line 11 on the screen

    printf("dotest2a\n");
    while (*p != 0) {
        (*(unsigned char *)(scaddr++)=(testasm1(*p)));
        p++;
    }
}

void dotest2b (void)
{
  char *p = &src2[0];
  uint16_t scaddr=&dest[0];     //output to line 11 on the screen

    printf("dotest2b\n");
    while (*p != 0) {
        (*(unsigned char *)(scaddr++)=(testasm2(*p)));
        p++;
    }
}

int docmp(uint8_t *s, uint8_t *r)
{
    int res = 0;
    while (*s != 0) {
//        printf("is %02x\n", *s);
        if (*r != *s) {
            printf("is %02x expected %02x\n", *s, *r);
            res++;
        }
        r++;
        s++;
    }
    return res;
}

int main(void)
{
//    memset(dest, 0, 24);
//    memset(dest, 0x11, 24);
    dotest1a(src, dest);
    res = docmp(dest, ref1);
//    memset(dest, 0, 24);
//    memset(dest, 0x22, 24);
    dotest2a();
    res += docmp(&dest[0], &ref2[0]);
// 
//     memset(dest, 0x33, 24);
    dotest1b(&src[0], &dest[0]);
    res = docmp(&dest[0], &ref1[0]);
//     memset(dest, 0x44, 24);
    dotest2b();
    res += docmp(&dest[0], &ref2[0]);
    return res;
}
```

acqn:
I''ll investigate this.
BTW, I wonder if we should change the term `asm statement` as it is actually an expression...

acqn:
So I see no problem with `memset`...

But there is indeed an issue regarding the topic here - the "expected behavior" that `__EAX__` is unaffected by post-inc/post-dec in the same statement is not always "guaranteed" with this PR:

```c
#include <stdio.h>

#ifndef __CC65__
unsigned int __AX__;
#endif

int main(void)
{
    int a = 0;
    int* p = &a;

    *p++ = (__AX__ = a++ + 42, /* <- sequence point here! */ __AX__);
    printf("%d\n", a);

    return a != 42;
}
```
Even with this PR, the result is not the expected `42`, but `1`.

Whilst this case can be fixed with more tweaks, it exposes an issue on how we should really deal with sequence points. Since the solution of that issue may impact (or be impacted by) this PR, I'm going to delay this PR.

mrdudz:
Oh sh*t. It's a bug in the test?

> But I'll hold on this PR to see what that issue (https://github.com/cc65/cc65/issues/1819) ragarding sequence point will result in.

I will fix it and commit the test into TODO then

mrdudz:
added the fixed test in todo

mrdudz:
looks good, thanks!

PR #1810, opened 2022-07-24
conio peek for Atari, Atmos, NES, OSIC1P

mrdudz:
Another one from the attic, someone who is familiar with the targets in question may want to have a look and fix it. It seems to match the implementation we agreed one requires adding some screencode to charcode conversion. And perhaps adding a cpeekrevers function.
NOT ready to merge yet


mrdudz:
Atari 800 complete now (tested and conio test updated), replaces what is in #532

mrdudz:
completely replaces #532 now

mrdudz:
ok well, after briefly testing those things, i will merge them. A hand ful functions are still missing after this (see targettest/conio.c)

PR #1825, opened 2022-08-07
Revert: Parse file included inside a macro at definition time

ReallyNiceGuy:
The reverted change broke too many use cases where users
were expecting the compiler quirky behavior to be the correct
behavior.

A better solution is needed for the quirk which does not break
current usage.


mrdudz:
It'd really be nice to have a test program for this (the old/intended behaviour), to make sure it will still work as correctly in the future, incase someone tries to fix it for real.

ReallyNiceGuy:
I agree, but it is out of the scope of this pull request, I believe.

Issue #1823, opened 2022-08-06
POSIX defines NULL as ((void *)0)


alejandro-colomar:
https://github.com/cc65/cc65/blob/4de40a20cb641a80de9b03222a0c686443178d43/include/stddef.h#L57

Hi,

POSIX defines `NULL` as `((void *)0)` (it doesn't require the outer parentheses, but they're safer, and most implementations, including glibc, use them), which is safer than `0` because it is a pointer type and makes it easier to detect issues in code:

```
int i = NULL;
```

Would you mind redefining NULL to be of type `void *`?

<https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/stddef.h.html>

rofl0r:
see also: http://ewontfix.com/11/

tl;dr - posix and C are at odds here.

alejandro-colomar:
That article recommends not using `NULL` by programmers for a reason: ISO C defines it in an unsafe way.

All of what that article says seems to be fixed by POSIX's definition, which requires that NULL is of a pointer type (`void *`).

Then, the article continues by saying that in C++, due to the rules around `void *`, that's not possible.  That a non-issue in C; we didn't make `void *` an unusable type.  That's why most C libraries have a different definition of NULL for C and C++.

BTW, musl libc changed the definition of NULL after that article was written, and they now use `void *` again for C.

alejandro-colomar:
Recommending your users to not use NULL at all is not incompatible with defining it in a better way.  I don't think that defining NULL in one of the worst ways that it can be defined, just because it's allowed by ISO C, and force users to not use it at all because it's broken is a good thing to do.

In fact, it's what is motivating ISO C to develop yet another null pointer constant `nullptr`, which doesn't add anything good.

mrdudz:
Not sure how the current way is a problem though - could you show a simple test case where it is?

alejandro-colomar:
The article here shows one example: <http://ewontfix.com/11/>  It requires you to cast `NULL` when passing it to a vararg (...) function.  POSIX `NULL` doesn't need to be casted, since the alignment and bit representation of `void *` and `char *` are the same.

Another simpler one is that the compiler can't detect broken code such as:

```
int i = NULL;
```

Of course the above is so broken that no humans in their right mind would write it, but another more reallistic case would be passing NULL to a function accepting an integer.  It's not that the code is not going to work --it will work--, but the compiler should probably warn about it.

mrdudz:
I see. Please make a PR :)

[...]

PR #1826, opened 2022-08-08
Make NULL conform to POSIX

alejandro-colomar:
POSIX.1-2008 tightened the definition of NULL to be 0 cast to the
type 'void *'.

Defining NULL as 0 is problematic, because it requires users to
cast NULL to a pointer type before passing it to variadic
functions.  Using POSIX's definition is safer, because NULL can be
used in all contexts without a cast, due to the alignment of
'void *' and 'char *' being the same.  It also helps the compiler
be able to detect when NULL is being used in an integer context.

Link: <http://ewontfix.com/11/>
Signed-off-by: Alejandro Colomar <alx.manpages@gmail.com>

### DIFF
--- a/include/stddef.h
+++ b/include/stddef.h
@@ -54,7 +54,7 @@ typedef unsigned size_t;
 
 /* NULL pointer */
 #ifndef _HAVE_NULL
-#define NULL    0
+#define NULL    ((void *) 0)
 #define _HAVE_NULL
 #endif
 


alejandro-colomar:
Closes: <https://github.com/cc65/cc65/issues/1823>

alejandro-colomar:
I wrote a branch for C++, which we can discard if you don't support C++.  I should also explain a bit more in the commit message why it's different in C++.

mrdudz:
Looks like either the library or the compiler needs further fixing (and no, forget about c++)

alejandro-colomar:
The library seems okay.  It's the compiler that refuses to compare a `void *` to a `struct DIR *`.  I don't see why this should not be allowed, especially when `void *` can be converted to any pointer type automatically.

acqn:
> It's the compiler that refuses to compare a `void *` to a `struct DIR *`. I don't see why this should not be allowed, when `void *` can be converted to any pointer type automatically. Especially, when the `void *` value is a null pointer. (Even ISO C requires that `(void *) 0` represents a null pointer, so one could write that expression instead of `0` or `NULL` and expect it to be a null pointer.)

That's bug. Will be fixed soon.

alejandro-colomar:
That's sooner than I expected :)  Easy fix!

alejandro-colomar:
Thanks! =)

PR #1832, opened 2022-08-20
[cc65] Preprocessor directive #define fixes

### LINKED ISSUE #1822: [cc65] Redefined macros cannot be all undefined with a single #undef?
```c
#define F 1
#define F 1
#define F 1

#undef F
#if defined F
#error #undef F fails!
#endif

#define F 0

#include <stdio.h>

int main(void)
{
    printf("%d\n", F);
    return F;
}
```
currently fails.

The cause is that every macro definition entry even with the same name is added to the macro table whilst only one of them can be removed with a single `#undef`.

A fix is done. Should it come in #1817 or a separate PR?

acqn:
- [X] Fixed #1822. Redefined macros and `#undef` now work as expected.
- [X] Disallowed `defined` as a macro name.
- [X] Added check for `##` at either end of of macro expansion.

### DIFF
--- a/src/cc65/preproc.c
+++ b/src/cc65/preproc.c
@@ -1053,6 +1053,12 @@ static void DoDefine (void)
     /* Remember if we're in C89 mode */
     C89 = (IS_Get (&Standard) == STD_C89);
 
+    /* Check for forbidden macro names */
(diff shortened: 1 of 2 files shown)


mrdudz:
you are on a roll :) thanks!

PR #1834, opened 2022-08-21
[cc65] Minor improvements, fixes and cleanups with the preprocessor

acqn:
See commit comments.
Note: A better preprocessor is coming, but this is still worth pushing as a step stone.


mrdudz:
cheers!

PR #1840, opened 2022-08-24
[cc65] Fixed function parameters declared as function types rather than function pointers

### LINKED ISSUE #1838: Wrongly generated code when passing a function to a function
I have to admit i have not seen this before, i have only ever seen function pointers declared as such. I stumbled about this in the test program used to verify the Berkeley softfloat library:
```
#include <stdlib.h>
#include <stdio.h>
 
int bar( int a )
{
//    printf("a = %d\n", a);
    return 12;
}
 
static void foo(int function(int))
{
    /*int n = */function(42);
//    printf("n = %d\n", n);
}
 
{
    foo(bar);
    return EXIT_SUCCESS;
}
```

cc65 gets it "almost" right. but not quite:
```
.proc   _bar: near

        jsr     pushax
        ldx     #$00
        lda     #$0C
        jmp     incsp2

.proc   _foo: near

        jsr     pushax
        lda     _function             ; <- compiler is confused here :)
        ldx     _function+1
        jsr     pushax
        ldx     #$00
        lda     #$2A
        pha
        lda     (sp,x)
        sta     jmpvec+1
        ldy     #$01
        lda     (sp),y
        sta     jmpvec+2
        pla
        jsr     jmpvec
        jsr     incsp2
        jmp     incsp2

.proc   _main: near

        lda     #<(_bar)
        ldx     #>(_bar)
        jsr     _foo
        ldx     #$00
        txa
        jmp     incsp4
```

acqn:
Fixed #1838.

### DIFF
--- a/src/cc65/expr.c
+++ b/src/cc65/expr.c
@@ -1133,10 +1133,6 @@ static void Primary (ExprDesc* E)
                     /* Enum or some other numeric constant */
                     E->Flags = E_LOC_NONE | E_RTYPE_RVAL;
                     E->IVal  = Sym->V.ConstVal;
-                } else if ((Sym->Flags & SC_FUNC) == SC_FUNC) {
-                    /* Function */
-                    E->Flags = E_LOC_GLOBAL | E_RTYPE_LVAL;
-                    E->Name  = (uintptr_t) Sym->Name;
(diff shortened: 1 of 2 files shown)


mrdudz:
cool, thumbs up for the prompt fix!

PR #1842, opened 2022-08-27
[cc65] Added support for the popular __COUNTER__ macro

### LINKED ISSUE #1812: [cc65] Extension: predefined macro __COUNTER__
`__COUNTER__` is a commonly supported predefined macro in many major C compilers (eg. [GCC](https://gcc.gnu.org/onlinedocs/cpp/Common-Predefined-Macros.html)), which has well-established and understood behavior (except when involved with precompiled headers, maybe) and been used in practice.

For those who might be unfamiliar with it: `__COUNTER__` is a "magical" macro that evaluates to an integer literal whose value is starting from 0 and increased by one every time it is found in a source code text. It is a convenient way to provide a unqinue identifier in conjunctions with the `##` operator.

Is this feature desired for cc65?

acqn:
Resolved #1812 (see it for details).

### DIFF
--- a/src/cc65/compile.c
+++ b/src/cc65/compile.c
@@ -396,9 +396,10 @@ void Compile (const char* FileName)
         DefineNumericMacro ("__EAGERLY_INLINE_FUNCS__", 1);
     }
 
-    /* Placeholders for __FILE__ and __LINE__ macros */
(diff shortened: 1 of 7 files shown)


mrdudz:
great, thanks!

Issue #1252, opened 2020-09-08
inline asm: cc65 chokes on label ref if space/tab follows


marcobaye:
this works:
```
asm(
"label: inx\n"
"       bne label\n"
);
```
but this does not:
```
asm(
"label: inx\n"
"       bne label        \n"
);
```
cc65 compiles it, but ld65 then complains about an unresolved external.
cc65 seems to count trailing spaces or TABs in label refs as part of the label name, so the matching label is not found and therefore not marked as referenced. the label definition is then dropped, at least it does not show up in the .s file, and so ld65 rightfully complains.

Here's a patch to fix it, though I guess it's not the "right" way to do it (it removes all trailing whitespace from inline asm lines).
```
diff --git a/src/cc65/codegen.c b/src/cc65/codegen.c
index 9e510272..7a9d5930 100644
--- a/src/cc65/codegen.c
+++ b/src/cc65/codegen.c
@@ -4370,5 +4370,19 @@ void g_switch (Collection* Nodes, unsigned DefaultLabel, unsigned Depth)
 void g_asmcode (struct StrBuf* B)
 /* Output one line of assembler code. */
 {
-    AddCodeLine ("%.*s", (int) SB_GetLen (B), SB_GetConstBuf (B));
+    int len = (int) SB_GetLen(B);
+    const char *buf = SB_GetConstBuf(B);
+
+    /* remove whitespace at end of line so it does not interfere with label refs */
+    while (len) {
+       switch (buf[len - 1]) {
+       case '\n':
+       case ' ':
+       case '\t':
+           --len;
+           continue;
+       }
+       break;
+    }
+    AddCodeLine ("%.*s", len, buf);
 }
```

acqn:
At the moment, labels are not supposed to be written directly in the asm strings. C labels should be used instead.

jmr:
> At the moment, labels are not supposed to be written directly in the asm strings. C labels should be used instead.

Then is the right solution to give an error/warning for labels in inline assembly?

Why aren't they supposed to be used, if they apparently work, at least in some cases?

acqn:
You are right. They should be fixed so that they could work in all cases.

oliverschmidt:
What about name clashes with labels generated by the compiler?

[...]

PR #1843, opened 2022-08-27
fix for issue #1252

marcobaye:
this patch removes end-of-line whitespace from asm lines, which happens to fix issue #1252.


mrdudz:
Fixed the root cause of the problem - and applied your patch anyway, since it seems to be desirable to remove trailing whitespace from inline asm anyway.

cheers :)

PR #1849, opened 2022-09-01
[cc65] Reverted E_ADDRESS_OF logic fix

### LINKED ISSUE #1847: Strange results when using the arrow operator on a structure pointer
I get strange results when using the arrow operator on a structure pointer.

```C
#include <stdio.h>

struct TestStruct {
    char a;
    char b;
};

struct TestStruct s1[2];
struct TestStruct* s1Ptr;


int main() {
    s1[1].b=2;
    s1Ptr=s1;
    printf("%x\n",s1[1].b);
    printf("%x\n",s1Ptr[1].b);
    printf("%x\n",(&s1[1])->b);
    printf("%x\n",(&s1Ptr[1])->b);
    printf("%x\n",(&s1[1])->b);
    printf("%x\n",(&(s1Ptr[1]))->b);
    return 0;
}
```

The result of this should be as below.
```
2
2
2
2
2
2
```

When I run it on cc65 I get:
```
2
2
2
99cd
2
99cd
```
99cd is the same as &(s1[1].b) .
It worked fine before, but it seems to be a recent bug.

acqn:
Fixed #1847 by reverting 6a3b9a9880599acc38fd721a24c5864ec3194a3e. Testcase added.

### DIFF
--- a/src/cc65/exprdesc.c
+++ b/src/cc65/exprdesc.c
@@ -93,7 +93,8 @@ int ED_IsLocPrimaryOrExpr (const ExprDesc* Expr)
 int ED_IsIndExpr (const ExprDesc* Expr)
 /* Check if the expression is a reference to its value */
 {
-    return (Expr->Flags & E_ADDRESS_OF) == 0 && !ED_IsLocNone (Expr);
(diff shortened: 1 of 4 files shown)


mrdudz:
Thanks for the prompt fix!

PR #1856, opened 2022-09-19
Big clean-up of driver return codes (#1794)

polluks2:
Now nonobvious code should contain an assert.

### DIFF
--- a/libsrc/apple2/mou/a2.stdmou.s
+++ b/libsrc/apple2/mou/a2.stdmou.s
@@ -133,8 +133,8 @@ next:   inc     ptr1+1
         bcc     :+
 
         ; Mouse firmware not found
-        lda     #<MOUSE_ERR_NO_DEVICE
-        ldx     #>MOUSE_ERR_NO_DEVICE
+        lda     #MOUSE_ERR_NO_DEVICE
+        ldx     #>$0000
         rts
 
         ; Check Pascal 1.1 Firmware Protocol ID bytes
--- a/libsrc/apple2/emd/a2.auxmem.s
+++ b/libsrc/apple2/emd/a2.auxmem.s
@@ -73,7 +73,8 @@ INSTALL:
         and     #$f0
         cmp     #$80
         bne     @L1
-        lda     #EM_ERR_OK
+        .assert EM_ERR_OK = 0, error
+        txa
         rts
 @L1:    lda     #EM_ERR_NO_DEVICE
 ;       rts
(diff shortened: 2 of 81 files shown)


libsrc/apple2/mou/a2.stdmou.s
```diff
@@ -133,8 +133,8 @@ next:   inc     ptr1+1
         bcc     :+
 
         ; Mouse firmware not found
-        lda     #<MOUSE_ERR_NO_DEVICE
-        ldx     #>MOUSE_ERR_NO_DEVICE
+        lda     #MOUSE_ERR_NO_DEVICE
+        ldx     #>$0000
```
groessler, line 137: Why did you change this (and the other similar ones)? Looks OK to me. Your change doesn't look nicer (worse in my eyes), and it breaks if MOUSE_ERR_NO_DEVICE is >= 0x100.

polluks2: Please look at our long discussion, MOUSE_ERR_NO_DEVICE >= 0x100 is impossible because the return value is char.

groessler: Ok. (I don't remember that specific discussion right now.)
But "#>$0000" for just "0" looks really ugly to me.

polluks2: This is an intended stumbling block because it isn't a high byte of an int.

mrdudz:
> > But "#>$0000" for just "#0" looks really ugly to me.

> This is an intended stumbling block because it isn't a high byte of an int but integer promotion

I have to admit i find this particular thing ugly too. The reason is demonstrated right here: it is not obvious what this is about. Actually not only that, it even looks like something is wrong there. I'd prefer something like this:
```
lda #FOO
ldx #0 ; return value is char
```
... because it achives the same goal and is a lot easier to read/understand.

polluks2:
Well, this patch is just a unification. The "ugly" syntax is already common, please try
`grep -r "#>\$0000" libsrc/`

mrdudz:
> The "ugly" syntax is already common

That doesn't mean it can't be changed though (and it should be easy too with some inline sed)

polluks2:
Feel free to use sed after a merge (don't forget adjusting the comments).

mrdudz:
> Feel free to use sed after a merge (don't forget adjusting the comments).

I am not willing to merge a PR that requires me to fix it after that, sorry. I am already working in other corners of the project and my time is limited. This is exactly why you should discuss and ask before doing larger changes (and its exactly why i started contributing.md) - it makes everyones life easier.

groessler:
I'm with @mrdudz.  @polluks2, please fix that yourself. But I very much like the non-"#>0000" changes.

mrdudz:
closed via [ffa83c3](https://github.com/cc65/cc65/commit/ffa83c32a4d4cb2f7bd7749aa3e2766c370e60a7)

PR #1862, opened 2022-09-30
[cc65] Renamed some more testcases

acqn:
Renamed a few testcases so that it's clearer what they were intended for.


mrdudz:
thanks!

PR #1863, opened 2022-09-30
[cc65] Fixed symbol visibility and usage problems with extern/static declarations

acqn:
- [X] Fixed block-scope `extern` declarations visibility:
```c
int c = 1;
int main(void)
{
    void* c = 0;
    {
        extern int c;   /* BUG: Conflicting types for 'c' */
        return c - 1;   /* BUG: Converting pointer to integer without a cast */
    }
}
```
- [X] Fixed missing file-scope `static` function codegen output in a rare care:
```c
static void f(void) {}  /* Generated functions with internal linkage are not always kept in cc65 */
int main(void)
{
    void f(void);       /* Function with internal linkage declared in block scope */
    f();                /* BUG: Unresolved external to _f in ld65 */
}
```

### DIFF
--- a/src/cc65/locals.c
+++ b/src/cc65/locals.c
@@ -465,8 +465,8 @@ static void ParseOneDecl (const DeclSpec* Spec)
         /* The default storage class could be wrong. Just clear them */
         Decl.StorageClass &= ~SC_STORAGEMASK;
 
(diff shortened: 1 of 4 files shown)


mrdudz:
Thanks!

Issue #1875, opened 2022-10-13
Loss of integer promotion when multiplying two smaller literal values into a larger type?


bbbradsmith:
I ran into a change of behaviour re: integer promotion, and I'm not sure if it's desired/intended. If I multiply two literals into something larger than an int, the don't seem to be promoted anymore, but they used to. Example code:
```
int test(unsigned long int v)
{
	if (v >= (256*256)) return 1;
	return 0;
}
```
Compiled with -O, giving the warning "Condition is always true":
```
;
; {
;
	jsr     pusheax
;
; if (v >= (256*256)) return 1;
;
	jsr     ldeax0sp
	jsr     return1
	beq     L0002
	ldx     #$00
	lda     #$01
	jmp     incsp4
;
; return 0;
;
L0002:	ldx     #$00
	txa
;
; }
;
	jmp     incsp4
```
Old output (2.17 - Git a6b04f6), no warnings:
```
;
; {
;
	jsr     pusheax
;
; if (v >= (256*256)) return 1;
;
	jsr     ldeax0sp
	cmp     #$00
	txa
	sbc     #$00
	lda     sreg
	sbc     #$01
	lda     sreg+1
	sbc     #$00
	lda     #$00
	tax
	bcc     L0001
	lda     #$01
	jmp     incsp4
;
; }
;
L0001:	jmp     incsp4
```
Two workarounds which will produce the same code as previously:
```
v >= (65536)
v >= ((unsigned long int)256*256)
```

This does give a warning, at least, so there's some diagnosis, but isn't the literal expression supposed to promote like this automatically under C's promotion rules? Was this capability removed intentionally, or was this accidental collateral of something else?

mrdudz:
To me it looks like it is working as expected.... 256*256 gives 65536 - ie 0x10000 - which does not fit into a 16bit int, so it results in 0 and the warning is saying just that. (perhaps there could be another warning that tells about the overflow?)

bbbradsmith:
To me it's not working as expected... code that was correctly making a 32 bit comparison before now isn't. In this case it gave a warning, but here's a related example that gives no warning at all:
```
unsigned long int test2()
{
	return 65536;
}

unsigned long int test3()
{
	return 256*256;
}
```
Result:
```
.proc	_test2: near
; return 65536;
	ldx     #$00
	stx     sreg+1
	lda     #$01
	sta     sreg
	txa
; }
	rts
.endproc

.proc	_test3: near
; return 256*256;
[...]

PR #1877, opened 2022-10-14
Rephrasing warnings for suspicious decimal integer constants

bbbradsmith:
Rephrasing two warnings about integer constants.

I felt the existing warning implied that long or unsigned long integer constants were _themselves_ a bad thing, but they are in fact valid code and are properly handled by the compiler.

After understanding the intent of the warnings is to point out suspicious cases where the programmer _probably_ should have used a suffix (e.g. do you expect 65535 to be an unsigned int or a signed long?). I have reworded the two warnings in a way that I think makes this purpose clearer.

(Investigated during #1875.)


acqn:
The "xx is `long`/`unsigned long`" warning text was "stolen" from a version of gcc (or maybe clang). It was indeed not as clear as this new one. Nice improvement!

PR #1878, opened 2022-10-17
Speed up sim65 by 10%

clbr:
Approx 40% of make test is spent in sim65. 10% of sim65 was spent in MemReadByte. Inline it for nice speed gains.

"make test" takes 10% less time after this, even though the expected increase was only 4%.


mrdudz:
Good catch!

PR #1879, opened 2022-10-20
[cc65] SymEntry cleanups

acqn:
Renamed functions (https://github.com/cc65/cc65/pull/1879/commits/a2dfa7c721ae0a0c3393c58740e0e286ec7e9b4c):

now in _src/cc65/datatype.c_
- `GetSymType` and `GetESUSymEntry` -> `GetESUTagSym`
- `GetSymTypeName` -> `GetTagSymName`
- `SetESUSymEntry` -> `SetESUTagSym`

now in _src/cc65/symentry.c_
- `CvtRegVarToAuto` -> `SymCvtRegVarToAuto`
- `ChangeSymType` -> `SymChangeType`
- `ChangeAsmName` -> `SymChangeAsmName`
- `HasAnonName` -> `SymHasAnonName`

Note: It was actually `GetSymType` that got renamed to `GetESUTagSym` rather than `GetESUSymEntry` would look like to be (from the name). The two renamed (or removed) functions were doing almost the same thing except that `GetESUSymEntry` had a narrower constraint asserting the type of the symbol should be `ESU` (`enum`, `struct` or `union`) than `GetSymType` had that would simply return NUL if the type was non-`ESU`.

### DIFF
--- a/src/cc65/compile.c
+++ b/src/cc65/compile.c
@@ -478,7 +478,7 @@ void Compile (const char* FileName)
         for (Entry = GetGlobalSymTab ()->SymHead; Entry; Entry = Entry->NextSym) {
             if ((Entry->Flags & (SC_STORAGE | SC_DEF | SC_STATIC)) == (SC_STORAGE | SC_STATIC)) {
                 /* Assembly definition of uninitialized global variable */
-                SymEntry* TagSym = GetSymType (Entry->Type);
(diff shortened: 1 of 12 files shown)


acqn:
A fix for #1888 on the current head would be conflicted with this.

mrdudz:
Since there were no complaints - i am just merging it :)

PR #1885, opened 2022-10-27
Allow specifying range end as a size

clbr:
Usability enhancement. Sometimes giving a size is faster than calculating the end absolute address.


mrdudz:
thanks!

PR #1896, opened 2022-11-02
[cc65] Fixed diagnostic messages on enumerator overflow

### LINKED ISSUE #1890: cc65: Overflow in enumerator value is not detected
The following code compiles without problems. The numeric overflow of the enumerator value 'b' is not detected and the variable 'c' has the value zero:

```C
#include <limits.h>
enum { a = ULONG_MAX, b } c = b;
```

acqn:
- [x] Resolved #1890.
- [x] An enumerator that would be incremented greater than ULONG_MAX now causes an error.

### DIFF
--- a/src/cc65/declare.c
+++ b/src/cc65/declare.c
@@ -611,20 +611,20 @@ static SymEntry* ParseEnumDecl (const char* Name, unsigned* DSFlags)
 
         } else {
 
-             /* Defaulted with the same signedness as the previous member's */
+            /* Defaulted with the same signedness as the previous member's */
             IsSigned = IsSignSigned (MemberType) &&
                        (unsigned long)EnumVal != GetIntegerTypeMax (MemberType);
 
-            /* Enumerate. Signed integer overflow is UB but unsigned integers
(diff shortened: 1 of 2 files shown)


mrdudz:
thanks!

clbr:
test/err/bug1890.c is CRLF, causing the same issues as before. Please fix. @acqn, most of these erroneous CRLF files were added by you, please double-check your files are LF from now on.

mrdudz:
@clbr perhaps you can produce some test cases that explicitly check this (ie contain valid code with "interesting" line endings) and open a seperate Issue for this? IMHO the compiler shouldnt barf on this - i don't know what the standard says, but ignoring every kind of line ending is something that is almost trivial to do :))

clbr:
Not compiler, git. See issue 1858.

mrdudz:
I see. fixed and check added in https://github.com/cc65/cc65/commit/79fd078ae340da74dbc5111ad04e919d0951c5e9 :)

acqn:
OK

PR #1897, opened 2022-11-02
[cc65] Fixed errors in the conditional operator "? :" parser

### LINKED ISSUE #1893: cc65: Compiler accepts a ternary expression where it shouldn't
The compiler accepts a ternary expression on the left side of an assignment which yields an rvalue and is therefore not assignable. This is true even in c89 and c99 mode:

```C
int main(void)
{
    int a, b, c;
    a == 1? b : c = 3;
    return 0;
}
```

The C standard doesn't allow this. And it's difficult to imagine what might be the correct translation for this code. gcc for example refuses to translate it:

```
$ gcc -Wall -o test test.c
test.c: In function ‘main’:
test.c:4:19: error: lvalue required as left operand of assignment
    4 |     a == 1? b : c = 3;
      |                   ^
```

acqn:
Fixed #1893.

### DIFF
--- a/src/cc65/expr.c
+++ b/src/cc65/expr.c
@@ -3869,9 +3869,9 @@ static void hieQuest (ExprDesc* Expr)
 
             ED_FinalizeRValLoad (&Expr2);
         } else {
-            /* Constant boolean subexpression could still have deferred inc/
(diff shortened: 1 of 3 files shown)


mrdudz:
thanks!

PR #1899, opened 2022-11-02
[cc65] Fixed compatibility checks on "pointer to pointer" vs "pointer to array" etc.

acqn:
Fixed #1895.

PR #1900, opened 2022-11-02
[cc65] Fixed endlessly repeated error messages when a declaration lacks a required identifier

### LINKED ISSUE #1889: cc65: Error recovery has degraded with recent changes
The current git version of the compiler has serious problems with error recovery. Here is an example:

```C
inline enum { a };
```

When translated with version 2.18, an error is detected and the compiler recovers from this error:

```
test.c(1): Error: Identifier expected
test.c(1): Warning: Implicit 'int' is an obsolete feature
test.c(1): Error: ';' expected
```

If this one liner is translated with the current git version, the compiler outputs the same error over and over again until it runs into the "Too many errors" limit:

```
test.c:1: Error: Identifier expected
test.c:1: Warning: Implicit 'int' is an obsolete feature
test.c:1: Error: ';' expected
test.c:1: Error: Identifier expected
test.c:1: Warning: Implicit 'int' is an obsolete feature
test.c:1: Error: ';' expected
test.c:1: Error: Identifier expected
test.c:1: Warning: Implicit 'int' is an obsolete feature
test.c:1: Error: ';' expected
test.c:1: Error: Identifier expected
test.c:1: Warning: Implicit 'int' is an obsolete feature
test.c:1: Error: ';' expected
test.c:1: Error: Identifier expected
test.c:1: Warning: Implicit 'int' is an obsolete feature
test.c:1: Error: ';' expected
test.c:1: Error: Identifier expected
test.c:1: Warning: Implicit 'int' is an obsolete feature
test.c:1: Error: ';' expected
test.c:1: Error: Identifier expected
test.c:1: Warning: Implicit 'int' is an obsolete feature
test.c:1: Error: ';' expected
test.c:1: Error: Identifier expected
test.c:1: Warning: Implicit 'int' is an obsolete feature
test.c:1: Error: ';' expected
test.c:1: Error: Identifier expected
test.c:1: Warning: Implicit 'int' is an obsolete feature
test.c:1: Error: ';' expected
test.c:1: Error: Identifier expected
test.c:1: Warning: Implicit 'int' is an obsolete feature
test.c:1: Error: ';' expected
test.c:1: Error: Identifier expected
test.c:1: Fatal: Too many errors
```

This is not only true for the given example but there are many others.

acqn:
Fixed #1889.

### DIFF
--- a/src/cc65/declare.c
+++ b/src/cc65/declare.c
@@ -1831,7 +1831,13 @@ static void Declarator (const DeclSpec* Spec, Declaration* D, declmode_t Mode)
             NextToken ();
         } else {
             if (Mode == DM_NEED_IDENT) {
+                /* Some fix point tokens that are used for error recovery */
+                static const token_t TokenList[] = { TOK_COMMA, TOK_SEMI };
+
                 Error ("Identifier expected");
+
+                /* Try some smart error recovery */
+                SkipTokens (TokenList, sizeof(TokenList) / sizeof(TokenList[0]));
             }
             D->Ident[0] = '\0';
         }
--- a/src/cc65/declare.c
+++ b/src/cc65/declare.c
@@ -1419,9 +1419,8 @@ static void ParseTypeSpec (DeclSpec* D, long Default, TypeCode Qualifiers,
             } else {
                 if (CurTok.Tok != TOK_LCURLY) {
                     Error ("Identifier expected");
-                } else {
-                    AnonName (Ident, "enum");
                 }
+                AnonName (Ident, "enum");
             }
             /* Remember we have an extra type decl */
             D->Flags |= DS_EXTRA_TYPE;
--- a/test/misc/Makefile
+++ b/test/misc/Makefile
@@ -155,6 +155,11 @@ $(WORKDIR)/goto.$1.$2.prg: goto.c $(ISEQUAL) | $(WORKDIR)
 	$(CC65) -t sim$2 -$1 -o $$@ $$< 2>$(WORKDIR)/goto.$1.$2.out
 	$(ISEQUAL) $(WORKDIR)/goto.$1.$2.out goto.ref
 
+$(WORKDIR)/bug1889-missing-identifier.$1.$2.prg: bug1889-missing-identifier.c $(ISEQUAL) | $(WORKDIR)
+	$(if $(QUIET),echo misc/bug1889-missing-identifier.$1.$2.error.prg)
+	-$(CC65) -t sim$2 -$1 -o $$(@:.error.prg=.s) $$< 2> $(WORKDIR)/bug1889-missing-identifier.$1.$2.out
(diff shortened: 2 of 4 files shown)


mrdudz:
what about making a simple test from the case given in #1889 ?

acqn:
> what about making a simple test from the case given in #1889 ?

That would be good, but..I have no idea how to do that given how poorly I am doing with make.

mrdudz:
MMMh shouldnt the "err" directoy work for this? I have to admit i have not tried with a case like this yet, ie checked if it can "see" the difference between "error" and "timeout because of endless loop". I will give it a try later tonite

mrdudz:
Same advice as in #1899 - Have a look at the examples added in https://github.com/cc65/cc65/commit/44b2e4f3316c34c6b18190af109f120f9eb45cdb - you can add a test in "ref" with a .cref file that contains the expected output (see readme)

acqn:
Or maybe we could change the exit value from Fatal() to tell whether there are "Too many errors"?

mrdudz:
We could, but that'd not really help - the makefile only sees "failure", it can not check the exit code without further shell specific magic which we have to avoid so it can still work on windows/cmd.exe

acqn:
The makefile has already been misfunctioning with on Windows for a while...
- `check`: fails on cmd.exe . Seemingly working right in MSYS2 though.
- `test`: fails in test/asm tests on both of cmd.exe and MSYS2 .

mrdudz:
> Just FYI, the makefile has already been malfunctioning on Windows for a while...

damn. (shows how many ppl use that) - i will have a look

acqn:
> Same advice as in #1899 - Have a look at the examples added in [44b2e4f](https://github.com/cc65/cc65/commit/44b2e4f3316c34c6b18190af109f120f9eb45cdb) - you can add a test in "ref" with a .cref file that contains the expected output (see readme)

I tried to add the test there, but (still) came across errors with the recent makefile changes.
- (Affects this) `test/ref`: the recently added `custom-reference.cref` and `custom-reference-error.cref` fail in cmd.exe due to the good old `/` vs `\` issue for `copy`.
- (Bonus!) `test/asm/listing`: output of `010-paramcount.bin` just differs from the reference output.

So I put the test in test/misc instead.

mrdudz:
> (Bonus!) test/asm/listing: output of 010-paramcount.bin just differs from the reference output.

I noticed this before - interestingly the output really _is_ different when cmd.exe is used - likely some kind of buffering issue/difference. Perhaps the tests should have references for stdout and stderr in separate files.

will play around with this and try to fix it

mrdudz:
> * (Affects this) `test/ref`: the recently added `custom-reference.cref` and `custom-reference-error.cref` fail in cmd.exe due to the good old `/` vs `\` issue for `copy`.
>
> So I put the test in test/misc instead.

https://github.com/cc65/cc65/commit/ebb33ee8b445d3531f1b9a787b60c353fecf3f31 should really fix it now :)

will merge this PR now, and move the test to test/ref - please open a separate issue if you still have problems with test/err on cmd.exe

PR #1902, opened 2022-11-02
Silencing confusing util/zlib 'warning' target

silverdr:
This PR silences the `warning` message emitted from `util/zlib` when building cc65 and adds explanations in `readme.txt` file


mrdudz:
makes sense, thx!

PR #1903, opened 2022-11-02
Silencing "parameter 'Flags' set but not used" warning

silverdr:


silverdr:
Yeah, I'd very much prefer to simply declare `... Flags __attribute((__unused__))` but Windows compiler doesn't like it

mrdudz:
in VICE we use macros for this kind of thing...

```
#ifdef _WIN32
#define ATTRIBUTE_UNUSED
#else
#define ATTRIBUTE_UNUSED __attribute((__unused__))
#endif
```

 I someohow prefer this, less cluttler in the actual code (and no need to have the same comment about it dozens of times)

The only question for cc65 would be where to put this kind of macro (maybe similar things do already exist?)

polluks:
Right now we have no common header
```
src/cl65/main.c:#if defined(_WIN32)
src/cl65/main.c:#if defined(_WIN32)
src/cl65/main.c:#if defined(CL65_TGT) && !defined(_WIN32)
src/sim65/paravirt.c:#if defined(_WIN32)
src/ca65/incpath.c:#if defined(CA65_INC) && !defined(_WIN32)
src/ar65/library.c:#if defined(_WIN32)
src/common/filestat.c:#if defined(_WIN32)
src/common/filestat.c:#if defined(_WIN32)
src/common/searchpath.c:#if defined(_WIN32)
src/common/searchpath.c:#ifdef _WIN32
src/common/searchpath.c:#if defined(_WIN32)
src/common/filetime.c:#if defined(_WIN32)
src/common/filetime.c:#if defined(_WIN32)
src/cc65/incpath.c:#if defined(CC65_INC) && !defined(_WIN32)
src/ld65/filepath.c:#if defined(LD65_LIB) && !defined(_WIN32)
src/ld65/filepath.c:#if defined(LD65_OBJ) && !defined(_WIN32)
src/ld65/filepath.c:#if defined(LD65_CFG) && !defined(_WIN32)

**Comment by silverdr** [comment]
> I someohow prefer this, less cluttler in the actual code (and no need to have the same comment about it dozens of times)

If there's something ready that could be used instead that would of course be better. A cursory grep didn't reveal anything to me though.

**Comment by rofl0r** [comment]
why not just use the portable (void) ? the VICE macro above breaks any compiler that isn't GNUC compatible.

**Comment by mrdudz** [comment]
> why not just use the portable (void) ? 

Because it is ugly and requires the same comment every time it occurs. the macro speaks for itself. Also with the macro you can easily remove it at one place, should that ever be needed for whatever.

> the VICE macro above breaks any compiler that isn't GNUC compatible.

That was just an example - of course the real macro should be guarded properly by the respective compiler checks and would not break anything that way.

As said, the question boils down to "where to put it". Is there some common header file that is already included by "all" sources in the tree?

**Comment by mrdudz** [comment]
should be fixed in https://github.com/cc65/cc65/commit/e5777f7ccfdef0869845c4a92d3a8dc602371de9

**Comment by silverdr** [comment]
> should be fixed in [e5777f7](https://github.com/cc65/cc65/commit/e5777f7ccfdef0869845c4a92d3a8dc602371de9)

```
cc65/codegen.c:4565:43: warning: parameter 'Flags' set but not used [-Wunused-but-set-parameter]
void g_testbitfield (ATTR_UNUSED(unsigned Flags), unsigned BitOffs, unsigned BitWidth)
```

silverdr:
https://github.com/cc65/cc65/pull/1909
updates the above

mrdudz:
Oops missed the clang case. thanks!

PR #1904, opened 2022-11-05
Remove CFLAGS and ASFLAGS from tests

ssterling:
By using the CFLAGS environment variable as set by the user, incompatible arguments such as `-march=native` may be included, stopping the compilation.
This pull request removes the CFLAGS and ASFLAGS variables from the targettest Makefile.

### DIFF
--- a/targettest/Makefile
+++ b/targettest/Makefile
@@ -114,11 +114,11 @@ DISK_atarixl   = testcode.atr
 %: %.s
 
 .c.o:
-	$(CC) $(CFLAGS) -Ors --codesize 500 -T -g -t $(SYS) $<
+	$(CC) -Ors --codesize 500 -T -g -t $(SYS) $<
 	$(AS) $(<:.c=.s)
 
 .s.o:
-	$(AS) $(ASFLAGS) -t $(SYS) $<
+	$(AS) -t $(SYS) $<
 
 .PRECIOUS: %.o
 


mrdudz:
i wonder in what context this patch was actually the solution to the supposed problem - because CFLAGS, ASFLAGS and LDFLAGS etc are used all over the place for target code, eg in "test", "targettest", "samples"

I agree that, depending on what you are doing, there can be conflicts with the environment - but i am not convinced this patch is the solution. We might want to rename the flags, for example, instead of removing them.

ssterling:
I tried reproducing this outside of Gentoo’s ebuild and somehow couldn’t even get targettest to fire as of 5d1ad05.  I ran `export CFLAGS="-march=native"; make`, logged the entire process and found no mention of `minimal.c` or `targettest`.  Here are the logs: [make.log](https://github.com/cc65/cc65/files/9944865/make.log)

ssterling:
Meanwhile:

```console
$ make CFLAGS="-march=native"
BUILD_ID: Git 5d1ad05cb
ar65/add.c
In file included from ar65/add.c:38:
ar65/error.h:42:10: fatal error: attrib.h: No such file or directory
   42 | #include "attrib.h"
      |          ^~~~~~~~~~
compilation terminated.
make[1]: *** [Makefile:155: ../wrk/ar65/add.o] Error 1
make: *** [Makefile:6: all] Error 2
```

But `export CFLAGS="-march=native"; make`—which is theoretically analogous to what’s going on in ebuild—works just fine.

Given all these seemingly unrelated problems, I’m going to assume my toolchain is somehow screwed up and that this isn’t a cc65 issue.

groessler:
There is a difference between `make CFLAGS=x` and `export CFLAGS=x; make`
If there is a `CFLAGS+=Y` in the Makefile, Y will be appended in the 2nd case, but not in the 1st.

Issue #1768, opened 2022-06-14
Unexpected "Error: Division by zero"


pmor13:
This code:
```
#if 1 || (8 / 0)
#endif
```
leads to:
```
Error: Division by zero
```
while per C11 the `(8 / 0)` is not evaluated (emphasis added):
> If the first operand compares unequal to 0, the **second operand is not evaluated**.

Hence, the `Error: Division by zero` may be unexpected.

Note: the same for `#if 0 && (8 / 0)`, `1 ? 0 : 1 / 0`, and so on.

acqn:
Not only in PP, but also:
```c
int a = 1 || (8 / 0);
int b = 0 && (8 % 0);
int c = 1 ? 42 : (0 % 0);
int d = 1 || a / 0;
int e = 0 && b % 0;
int f = 1 ? 42 : (a %= 0, b /= 0);
```
since the PP is reusing "almost the same" expression parser.
Somewhat clean fix would be best with separation of the parser and the codegen.

mrdudz:
fixed by #1913

PR #1913, opened 2022-11-12
[cc65] Improved diagnostics on div-by-zero/bitwise-shift in unevaluated context and overall

### LINKED ISSUE #1887: cc65: Div by zero in dead code - no output
The compiler refuses to generate output if there's a division by zero in code that is never called. Even if the code would have been removed if the compiler would process the input further.

```C
#include <stdio.h>

static int never_called(int a)
{
    return a / 0;
}

int main(void)
{
    printf("Hello world\n");
    return 0;
}
```

Strictly speaking the compiler is correct since division by zero triggers ub. So it's not a bug. But I don't think it's a good idea either.

acqn:
- [X] Fixed the issue that the always-unevaluated subexpression of a conditional operation with constant condition would still be evaluated.
- [X] Fixed https://github.com/cc65/cc65/issues/1768#issuecomment-1175221466 - no more division-by-zero/modulo-with-zero/shift-by-negative/shift-by-too-wide diagnostics in unevaluated context.
- [X] Resolved #1887 (maybe?) by [making the division-by-zero/modulo-with-zero errors warnings](https://github.com/cc65/cc65/issues/1887#issuecomment-1297198947).
- [X] Improved diagnostics for bitwise-shift:
- - [X] Added missing checks for `<<=` and `>>=`.
- - [X] Added hints on the behavior when the shift count is out of the range of the integer-promoted type.
- - [X] Added additional checks against bit-field widths.

Note: Division-by-zero/modulo-with-zero/shift-by-negative/shift-by-too-wide are handled differently for the preprocessor.
- Division-by-zero/modulo-with-zero cause errors in preprocessor expressions when evaluated.
- Shift-by-negative results in shifting in the opposite direction.
- Shift-by-too-wide will behave as if to shift bits one-by-one till all original bits were shifted out. (See #1918)

### DIFF
--- a/src/cc65/assignment.c
+++ b/src/cc65/assignment.c
@@ -345,7 +345,12 @@ static void OpAssignBitField (const GenDesc* Gen, ExprDesc* Expr, const char* Op
 
                         /* Here we simply "wrap" the shift count around the width */
                         Expr2.IVal &= ExprBits - 1;
-                   }
+
+                        /* Additional check for bit-fields */
+                        if (Expr2.IVal >= (long)Expr->Type->A.B.Width) {
+                            Warning ("Shift count %ld >= width of bit-field", Expr2.IVal);
+                        }
+                    }
                 }
 
                 /* Adjust the types of the operands if needed */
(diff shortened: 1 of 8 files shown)


mrdudz:
Thanks!

Issue #1912, opened 2022-11-11
tests that compare stdout and stderr with a reference may fail


mrdudz:
Some tests pipe the stdout and stderr of the compiler or assembler into a file and compare the result with a reference. This has the subtle problem that both stdout and stderr and buffered i/o streams that may or may not work the same on different OSs or even shells. That means the order of the lines in the resulting file is not guaranteed.

In practise it is "mostly" not an issue, but "we" still stumbled about one test where different behaviour can be triggered depending on running it in cmd.exe or bash (msys): ```test/asm/listing``` output of ```010-paramcount.bin``` differs from the reference output when running the test from cmd.exe.

The solution is most likely to have two reference files and not redirect into one file.

Fixing all the tests seems like a huge effort for little gain right now - but fixing enough that at least this one failing one works again on both would be nice - @spiro-trikaliotis perhaps you can have a look?

spiro-trikaliotis:
One question: As I do not have a cmd.exe environment, can I have the output files (testwrk/asm/listing/) of a run with cmd.exe, so I can determine what fails exactly?

mrdudz:
sure, with cmd.exe 
```
asm\listing>type ..\..\..\testwrk\asm\listing\010-paramcount.err
 .paramcount = 3
 .paramcount = 5
 .paramcount = 3
 .paramcount = 5
 .paramcount = 1
 .paramcount = 5
010-paramcount.s:18: Warning: User warning: r1 is blank!
010-paramcount.s:14: Note: Macro was defined here
010-paramcount.s:8: Note: Macro was defined here
010-paramcount.s:19: Warning: User warning: r1 is blank!
010-paramcount.s:14: Note: Macro was defined here
010-paramcount.s:8: Note: Macro was defined here
010-paramcount.s:20: Warning: User warning: r1 is blank!
010-paramcount.s:14: Note: Macro was defined here
010-paramcount.s:8: Note: Macro was defined here
```
hower the reference file (which apparently matches the output in msys or linux):
```
test\asm\listing>type ref\010-paramcount.err-ref
 .paramcount = 3
 .paramcount = 5
010-paramcount.s:18: Warning: User warning: r1 is blank!
010-paramcount.s:14: Note: Macro was defined here
010-paramcount.s:8: Note: Macro was defined here
 .paramcount = 3
 .paramcount = 5
010-paramcount.s:19: Warning: User warning: r1 is blank!
010-paramcount.s:14: Note: Macro was defined here
[...]

PR #1916, opened 2022-11-12
Split stdout and stderr in test/asm/

spiro-trikaliotis:
Some tests pipe the stdout and stderr of the compiler or assembler into a file and compare the result with a reference. This has the subtle problem that both stdout and stderr and buffered i/o streams that may or may not work the same on different OSs or even shells. That means the order of the lines in the resulting file is not guaranteed.

In practise it is "mostly" not an issue, but "we" still stumbled about one test where different behaviour can be triggered depending on running it in cmd.exe or bash (msys): test/asm/listing output of 010-paramcount.bin differs from the reference output when running the test from cmd.exe.

The solution is most likely to have two reference files and not redirect into one file.

This patch fixes the issue (cf. #1912) for the directory test/asm/.


rofl0r:
commit title speaks of stdin where it should be stderr

spiro-trikaliotis:
@rofl0r good catch... ;)

spiro-trikaliotis:
@mrdudz Do you plan to comment on this PR, merge it, or doing anything else? I want to work on the assembler, but this PR is blocking me somehow - I have to rebase my work, and I do not want to do it twice, so I am waiting on the outcome of this PR...

mrdudz:
i'll have a look and merge it in the next days (certainly weekend)... busy atm, sorry :)

Issue #1914, opened 2022-11-12
hidden bug in g_typecast() ?


mrdudz:
While debugging some type conversion code in my fptest branch, i stumbled over this:
look here: https://github.com/cc65/cc65/blob/master/src/cc65/codegen.c#L1511
It uses ```rhs``` for the conversion to long and int, and then ```lhs``` for the conversion to char (and for the error message).  I don't claim i understand how all of this works in detail, but that seems to be wrong, and it should be rhs in all cases instead? (At least changing it that way fixed something related for me :))

@acqn perhaps you can take a look and comment?

Also, if this is really a bug - how do we trigger it from a test program (in head)? I found it when examining why this would not work as expected:

```
float f = 42;
unsigned char c;
c = f;
```
So what it does with the original code is: it uses lhs for the conversion to char, ie char to char, so it will just copy one byte. Fixed as above results in the correct conversion code being emitted via g_regchar.

Now that i wrote it, the bug may not show when converting integers to chars - at least not in the cases when copying just the lsb will be a valid conversion anyway. Maybe something that requires/checks sign extension can trigger a problem(?)

acqn:
> ...and it should be rhs in all cases instead?

No. See below.

> Also, if this is really a bug - how do we trigger it from a test program (in head)?

Not a bug as far as I see (though I got puzzled too when I saw it first time).

For most integer type conversions in codegen, the "old type" info about signedness is needed in order to do sign-extension correctly, while the "new type" info about signedness is just unneeded. The reason why `g_regchar` is different from `g_regint` and `g_reglong` is that it doesn't need to promote the value from a "smaller" type (thus no need for the "old type" info), but on the other hand, it has to _always_ keep the "converted value" in AX as a full `int` (thus it needs the "new type" info about signedness to decide whether X should hold `0x00` or `0xFF`) since we are _almost always_ using the `char` value in the primary as an `int` (unless `CF_FORCECHAR` is specified) from the whole AX without promotion/conversion first in codegen.

For the type error message inside `g_typecast`, `lhs` is used because the switch statement is done on it.

> (At least changing it that way fixed something related for me :))

If you do need the "new type" info for conversion, you might have to change the interface of these functions to take more parameters.

> I found it when examining why this would not work as expected:
```c
float f = 42;
unsigned char c;
c = f;
```
Hmm, perhaps you may need to convert it from `float` to `int` first, then to `char` (but still `int` under the hood)?

> @acqn perhaps you can take a look and comment?

[...]

PR #1917, opened 2022-11-13
[cc65] Clearer comments on and usage of code generator flags with type conversions in the primary register

acqn:
Improve parameter naming and comments to reduce confusion.
Moved some code from `g_regchar` to `g_regint` to reflect the mechanism discussed in https://github.com/cc65/cc65/issues/1914#issuecomment-1312646436 better.


mrdudz:
Thank you!

PR #1918, opened 2022-11-13
[cc65] Fixed bitwise-shift in preprocessor

acqn:
- [X] Out-of-range shift count checking was insufficient.
- [X] Right-shift for signed integers was just wrong for whatever reasons.


mrdudz:
thanks!

PR #1927, opened 2022-11-20
Character defines for osic1p target

smuehlst:
Added defines for the line graphics characters and the enter key for the osic1p target.

Added `#ifdefs` to the `targettest/conio.c` program so that it compiles for platforms without cursor keys.


mrdudz:
thanks!

Issue #1933, opened 2022-11-27
As of 11/26/2022 - build fails with MINGW?


mrdudz:
### Discussed in https://github.com/cc65/cc65/discussions/1929

<div type='discussions-op-text'>

<sup>Originally posted by **xlar54** November 27, 2022</sup>
Hello, Im getting the following when trying to build the current master branch:

```
apple2 - apple2/targetutil/convert.c
cc65.exe: Internal error: Invalid format specifier in xvsnprintf, file 'common/xsprintf.c', line 616
make[2]: *** [apple2/targetutil/Makefile.inc:5: ../libwrk/apple2/convert.o] Error 1
make[1]: *** [Makefile:119: apple2] Error 2
make: *** [Makefile:7: all] Error 2
```

Is this a known issue? </div>

xlar54:
Ityped the following:

make clean
git pull
make

mrdudz:
What kind of setup is it? 32 or 64 bit? What windows version? bash or cmd.exe?

The snapshot is built with mingw and that seems to work - so i wonder what you are doing differently there.

mrdudz:
Are you using cmd.exe by any chance? I am playing with testing that in the github actions right now - and i am getting the same error! see https://github.com/cc65/cc65/actions/runs/3560655016/jobs/5980841299

mrdudz:
I tracked it down to the recent printf-specifier hack in codegen.c. This is really super annoying :(

mrdudz:
should be fixed, please try again

xlar54:
works like a champ, thank you!

bbbradsmith:
Should this be closed?

mrdudz:
indeed, thx for the reminder

PR #1937, opened 2022-11-27
run regression tests also on cmd.exe

mrdudz:
The makefiles should also be tested when using cmd.exe

### DIFF
--- a/.github/workflows/build-on-pull-request.yml
+++ b/.github/workflows/build-on-pull-request.yml
@@ -67,3 +67,38 @@ jobs:
 
       - name: Build app (release)
         run: msbuild src\cc65.sln -t:rebuild -property:Configuration=Release
+
+      - name: Install MINGW64
+        uses: msys2/setup-msys2@v2
+        with:
+          msystem: MINGW64
+          update: true
+          install: >-
+            git
+            base-devel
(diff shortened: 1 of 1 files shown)


mrdudz:
Interestingly, this is blocked by #1933

mrdudz:
Oh well, i am out of ideas. It seems there is no way to actually do this. If someone else knows how, please tell

mrdudz:
I want the github action to run the regression tests using cmd.exe as well, to make sure that is working (it broke in the past without getting noticed).

groessler:
I meant the "problem" (iow, whats not working), not the "motivation".

So to reproduce, checkout this branch on Windows, then run "gmake" and see what's happening?

mrdudz:
> I meant the "problem" (iow, whats not working), not the "motivation".

The problem is with the github action itself. Somehow it will not (always) use cmd.exe and always ends up using bash (which doesnt help in this case)

> So to reproduce, checkout this branch on Windows, then run "gmake" and see what's happening?

If only :) Locally it works fine. You can use a tool called "act" ( https://github.com/nektos/act ) to run the GHA locally and test it - but unfortunately that also doesnt reproduce GH 100%, which is why i gave up.

mrdudz:
closed via #2401

PR #1943, opened 2022-12-08
Fix for #1941

### LINKED ISSUE #1941: [bug] Shifts by zero bits are broken
I am assuming that right-shifing an unsigned char should have no effect on an unsigned char. This is the case in my code when compiled with any compiler except CC65.

I have this macro:
```
#define initial_letter_drop() \
do \
{ \
    uint8_t i; \
    uint8_t k; \
    \
    _XL_PRINTD(0,1,3,(uint8_t) MIN_INITIAL_DROP + (level>>0)); \
    k = MIN_INITIAL_DROP + (level>>0); \
    _XL_WAIT_FOR_INPUT(); \
    for(i=0;i<(uint8_t) MIN_INITIAL_DROP + (level>>0);++i) \
    { \
        drop_letter(); \
        short_pause(); \
    } \
} while(0)
```
It produces a infinite loop when compiled with CC65. It can be fixed if I use `k` instead of `(uint8_t) MIN_INITIAL_DROP + (level>>0)`.

I will try to reproduce it with a minimal example.

Has any recent commit changed how the right shift operator works?

laubzega:
Fixes #1941 by making sure the expression type remains unchanged if shift of 0 bits is requested.

### DIFF
--- a/src/cc65/shiftexpr.c
+++ b/src/cc65/shiftexpr.c
@@ -174,10 +174,9 @@ void ShiftExpr (struct ExprDesc* Expr)
             */
             if (Expr2.IVal == 0 || lconst) {
 
-                /* Set the type */
-                Expr->Type = ResultType;
-
                 if (lconst) {
+                    /* Set the type */
+                    Expr->Type = ResultType;
 
                     /* Evaluate the result */
                     switch (Tok) {
--- a/test/val/bug1941.c
+++ b/test/val/bug1941.c
@@ -6,8 +6,9 @@
 uint8_t foo=1U;
 uint8_t bar=3U; // You need it to reproduce the issue
 
+_Static_assert (sizeof (foo >> 0) == sizeof (int), "Result type of (foo >> 0) should be int");
+
 int main(void)
 {
     return(foo == foo << 0 ? EXIT_SUCCESS : EXIT_FAILURE);
 }
-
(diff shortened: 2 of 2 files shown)


mrdudz:
It would be super nice if you would add a test for this problem to the PR

acqn:
The result of shift-by-zero shall always be integer-promoted, or things like this would fail:
https://github.com/cc65/cc65/blob/22d435b68966b86929d3c393f3bbc1593b261cee/test/val/bug1941-shift-by-zero.c#L10

mrdudz:
So the proposed fix needs a fix, i guess :)

laubzega:
@acqn - would you be so nice and merge our PRs? Unless it can wait a few days (I am traveling rn).

mrdudz:
what is missing here? do we basically want to merge https://github.com/acqn/cc65/commit/22d435b68966b86929d3c393f3bbc1593b261cee ? :)

mrdudz:
ping? :)

mrdudz:
@acqn @laubzega please comment, i can do whatever needs to be done, but i need to know what exactly :)

mrdudz:
@acqn @laubzega PING! are you still on vacation or something? :)

mrdudz:
@acqn @laubzega i took the patch from acqn and made a new PR #1983 - please check/test/confirm ;)

PR #1952, opened 2022-12-22
Introduced the notion of a standard serial driver.

oliverschmidt:
There's no target with more than one serial driver (and I don't see that change anytime soon) so it's a no-brainer to apply the standard driver concept to serial drivers.

### DIFF
--- a/include/apple2.h
+++ b/include/apple2.h
@@ -171,7 +171,7 @@ extern struct {
 extern void a2_auxmem_emd[];
 extern void a2_stdjoy_joy[];     /* Referred to by joy_static_stddrv[]   */
 extern void a2_stdmou_mou[];     /* Referred to by mouse_static_stddrv[] */
-extern void a2_ssc_ser[];
+extern void a2_ssc_ser[];        /* Referred to by ser_static_stddrv[]   */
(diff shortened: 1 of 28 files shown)


mrdudz:
I have my doubts with swithlink being the standard driver here (it is by far not the standard interface - that would be the userport). But its easy to change :)

At this point i wonder how the setup looks like that was used to test this stuff (i am still looking for one for that matter)

polluks:
> I have my doubts with swithlink being the standard driver here (it is by far not the standard interface - that would be the userport). But its easy to change :)

Indeed, the Swiftlink is not built-in but we have no userport driver right now.

mrdudz:
If you want to have a look, there is one here: https://github.com/mrdudz/cc65-c64-up2400

polluks:
Terminal's 9600 baud seems to be critical for this driver. How about a 2400 fallback?
At least ^[ works as documented :-)

oliverschmidt:
That userpoet code is waiting ~ a decade to be transformed into a cc65 driver by now. So sorry for not taking it into account ;-)

mrdudz:
I know i know. Same for the serialsurfer code :)

oliverschmidt:
In case the userport driver would show up at some day - and then of course would become the default - then my terminal program could simply set to 2400 for all targets.

Issue #1957, opened 2022-12-30
[sp65] 1BPP pixel sprite generation buggy In Atari Lynx sprites


karrika:
I just tried to help a friend to convert his old sources to the latest cc65 and found out that the sp65 is really buggy in the current sp65.
![Buggy](https://user-images.githubusercontent.com/6996727/210056909-4d9cc3ec-e51e-46bd-9c1d-cb133f8c497d.png)
I have been developing the sp65 for my own work and the generation of optimized sprites is a really complex thing. The solution I have is still far from optimal - but a lot better than the current version we have. It does not produce a better version than optimizing by hand but it does produce a correct version.
![Fixed](https://user-images.githubusercontent.com/6996727/210057123-a49c0405-9cc8-4c1a-b60a-3310c7ff5424.png)
Anyway, I plan to bring up the code to a working level and make a pull request.

mrdudz:
again, this can be closed by #1958 - correct? (if you make an issue and a related PR, please reference one from the other)

karrika:
Yes.

mrdudz:
closed by #1958

PR #1958, opened 2022-12-30
[sp65] Sprite generation fixed for 1BPP Lynx sprites. Fix for #1957

karrika:
The lynxsprite.c is fairly complex today as it is pretty hard to choose between literal and packed formats. Sometimes the packed format produces bad sprites due to errors in the Lynx hw. In this version I have also tackled the problem of how to choose the right kind of a sprite. The Lynx supports 1BPP, 2BPP, 3BPP and 4BPP sprites. It also supports BACKGROUND and SHAPED sprites where all pixels are opaque. It also supports sprites with transparency where index 0 of the sprite is transparent. In order to control the process I added a parameter pen=01 which determines which colours of the bitmap are mapped to witch colours in the sprite. There is also two new convenient words for setting the anchor point of a sprite. ax=mid,ay=mid puts the anchor in the middle of the sprite, ax=0,ay=0 puts the anchor point in the top left corner, ax=max,ay=max puts the anchor point in the bottom right corner. I know there is a need to add all this info to the documentation also. I am working on it.

### DIFF
--- a/src/sp65/bitmap.h
+++ b/src/sp65/bitmap.h
@@ -75,6 +75,9 @@ struct Bitmap {
     unsigned    Width;
     unsigned    Height;
 
+    /* Bits per pixels */
+    unsigned    BPP;
+
     /* Palette for indexed bitmap types, otherwise NULL */
     Palette*    Pal;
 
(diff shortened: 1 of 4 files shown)


mrdudz:
trigger checks...

mrdudz:
looks good, thanks!

PR #1961, opened 2022-12-30
Teach da65 about table units

clbr:
Another da65 usability enhancement.

```
  Split the table into sections of this size. For example, if you have a
  ByteTable of size 48, but it has logical groups of size 16, specifying
  16 for UNIT adds newlines after every 16 bytes. UNIT is always in bytes.
```


mrdudz:
looks good, thanks!

PR #1968, opened 2023-01-12
Allow line_bynumber to return more than one result

mseabold:
If a macro is defined in an include file, each source file that includes and references it will get a unique line info entry in the debug information. This leads to multiple line ids that all reference the same source id/line number combination. However, cc65_line_bynumber() only searches for the first entry it can find that matches the source and line. Here, I have changed it to work more like cc65_symbol_byname. It now finds the first index of a matching entry, then copies out all matching entries.


mrdudz:
Please also update the documentation accordingly

mseabold:
I added "otherwise a list of line infos" to the function comment in the header and source. Is there other documentation that should be updated that I missed?

mrdudz:
I always forget that this stuff is not documented at all (in the actual documentation, i wasn't referring to the comments in the source). That really needs to change though - would be awesome if you could fill in this gap, now that you have apparently understood how it works and what it does.....
But since thats a bit much to ask for just to have this merged, i will do that now. cheers

PR #1977, opened 2023-01-20
Sim65 : new command line option to disable terminal echo + buffering AND kbhit implementation

cgimenez:
Hi, I started using Sim65 to prototype a simple game.

The more I use Sim65 the more I love it :
- instant startup
- host native speed
- run on terminal under VS Code
- advanced display (hello ANSI sequences)

Something that I was missing is :
- getchar() : disabling echo and buffering (using ENTER to get the char)
- no kbhit implementation

Notes : 
- these two commits are closely related
- untested under Linux (but I could) but is likely to works
- no guards for other OSes (eg. Windows), this patched version will certainly fail to compile except on MacOS and Linux - And I have no Windows host to make tests

First PR ever so if there's something more that I could do, or do better, please tell.

### DIFF
--- a/src/sim65/main.c
+++ b/src/sim65/main.c
@@ -35,10 +35,8 @@
 
 #include <string.h>
 #include <stdio.h>
-#include <unistd.h>
 #include <stdlib.h>
 #include <errno.h>
-#include <termios.h>
 
 /* common */
 #include "abend.h"
@@ -52,7 +50,10 @@
 #include "memory.h"
 #include "paravirt.h"
 
-
+#ifndef(_WIN32)
+#include <unistd.h>
+#include <termios.h>
+#endif
 
 /*****************************************************************************/
 /*                                   Data                                    */
@@ -82,9 +83,11 @@ static const unsigned char HeaderVersion = 2;
 /*                                   Code                                    */
 /*****************************************************************************/
 
+#ifndef(_WIN32)
 void Exit() {
     tcsetattr(STDIN_FILENO, TCSANOW, &OldTermAttrs);
 }
+#endif
 
 static void Usage (void)
 {
@@ -92,7 +95,7 @@ static void Usage (void)
             "Short options:\n"
             "  -h\t\t\tHelp (this text)\n"
             "  -c\t\t\tPrint amount of executed CPU cycles\n"
-            "  -t\t\t\tDisable term echo and buffering\n"
+            "  -t\t\t\tDisable term echo and buffering (*Nix only)\n"
             "  -v\t\t\tIncrease verbosity\n"
             "  -V\t\t\tPrint the simulator version number\n"
             "  -x <num>\t\tExit simulator after <num> cycles\n"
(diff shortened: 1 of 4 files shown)


src/sim65/main.c
```diff
@@ -77,21 +85,27 @@ static const unsigned char HeaderVersion = 2;
...
+}
+#endif
 
 static void Usage (void)
 {
     printf ("Usage: %s [options] file [arguments]\n"
             "Short options:\n"
             "  -h\t\t\tHelp (this text)\n"
             "  -c\t\t\tPrint amount of executed CPU cycles\n"
+            "  -t\t\t\tDisable term echo and buffering (*Nix only)\n"
```
polluks, line 100: Maybe the output of the two lines should be optional?

cgimenez: Yeap, makes sense ! Done

cgimenez:
Since that was my first PR I can see now that I did it the wrong way : I should not have commits in my master local branch, but instead have created a specific branch for the PR, because now my master is polluted by me.

I would like to have this PR deleted and restart from scratch and for what I read I need to (but not sure) :
- close de PR
- delete the origin branch (master in that case)
- destroy my forked repository

Can somebody confirm please ?

mrdudz:
Not sure what exactly the question is :)

That said, i am very reluctant to merge something like this when it is not confirmed to work on "the big three" - i don't want features that work on one host but not on another. It should be well possible to do this - perhaps even without requiring ifdef magic and windows specific terminal hacks.

mrdudz:
> My question is outside of the scope of cc65 : this PR is my first PR ever and I think that I did not followed the usual way to do it which is : fork cc65, git clone forked then create a branch for the PR (eg. kbhit), commit on that branch, then create a PR from that.
Instead I did the commits in my local cc65 master branch which is bad. So I wonder how I could have this PR deleted (and then I will delete my cc65 forked repository, and restart from zero : fork again, create a local kbhit branch, and do another PR), no matter if this PR is interesting or no...

First of all: you can make a PR this way. Of course it has drawbacks on your end :)

Then - cant you close the PR yourself (close button at the button of the page)?

> disabling terminal input buffering and echo can be done outside of sim65 (see bash script below). I did put this option in the PR for conveniency but you're right, that must be avoided
kbhit() on the other hand is really missing on sim65. The blocking getchar() is a no go. But if there's another solution to implement this function correctly and in a multi-platform way I will be glad to hear !

I agree. Have a look at the answer here: https://stackoverflow.com/questions/69393895/how-to-use-non-blocking-way-to-check-stdin-is-empty-or-not - i think that should work in both linux and windows (and macOS)

polluks:
@mrdudz I like the poll approach, but NetBSD and macOS say
`poll failed on stdin: Undefined error: 0`

cgimenez:
> @mrdudz I like the poll approach, but NetBSD and macOS say `poll failed on stdin: Undefined error: 0`

Which version of macOS ? I'm on 10.15 and did not had this error....

polluks:
13.2
I was using the example of Stack Overflow

mrdudz:
I cant believe select/poll does not work on macOS. Or even NetBSD. (Select/Poll come from BSD) Pretty much all half portable network code uses them.

groessler:
The example is wrong. It has "ret > 0" when it should use "ret >= 0".

groessler:
Timeout value of 0 is wrong.

polluks:
INFTIM is not POSIX

PR #1991, opened 2023-02-17
Remove .exe from file output

### LINKED ISSUE #1990: "make test" no longer works with MSYS2 due to asm/listing tests
I used to be able to run "make test" with MSYS2 but it's no longer viable due to the asm/listing tests that have been added.

This is because ld65 will be ld65.exe on the windows environment, and diff against the log will fail.

This is the first test that fails:
```
diff -u ref/030-assert-success.ld65err2-ref ../../../testwrk/asm/listing/030-assert-success.ld65-err2
```

It expects lines of this form:
```
ld65: Warning: 030-assert-success.s:3: Code not at $0000
```

Instead it gets a .exe on ld65 and fails the comparison:
```
ld65.exe: Warning: 030-assert-success.s:3: Code not at $0000
```

spiro-trikaliotis:
Whenever a tool like ld65 wants to output its name into a log file, it uses the name of command-line parameter 0. However, this parameter also includes the .exe suffix if it is on Windows.

This patch removes the .exe suffix, so that the output is clean and consistent across platforms.

This fixes #1990.

### DIFF
--- a/src/common/cmdline.c
+++ b/src/common/cmdline.c
@@ -181,6 +181,17 @@ void InitCmdLine (int* aArgCount, char*** aArgVec, const char* aProgName)
             /* Use the default */
             ProgName = aProgName;
         }
+        else {
+            /* remove .exe extension, if there is any
+            **
+            ** Note: This creates a new string that is
+            ** never free()d.
+            ** As this is exactly only string, and it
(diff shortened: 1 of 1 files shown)


mrdudz:
looks good :)

Issue #1989, opened 2023-02-16
OptStackOps optimization case bypasses comparison generating incorrect code


bbbradsmith:
In the following example, the "fail" variation produces incorrect code with -O. This appears to be a regression since the 2.19 release.

([godbolt](https://godbolt.org/z/r86be4MWE))
```
unsigned char i,r,j;
void fail()
{
    if ((i & 0x1f) == (r ? 0 : 16)) j -=8;
}
void pass()
{
    if ((i & 0x1f) == (unsigned char)(r ? 0 : 16)) j -= 8;
}
```

Produces:
```
.proc   _fail: near
        lda     _i
        and     #$1F
        sta     ptr1
        lda     _r
        beq     L0005
        lda     #$00
        jmp     L0004
L0005:  lda     #$10
        cmp     ptr1
L0004:  bne     L0002
        lda     _j
        sec
        sbc     #$08
        sta     _j
L0002:  rts

.proc   _pass: near
        lda     _i
        and     #$1F
        sta     ptr1
        lda     _r
        beq     L0005
        lda     #$00
        jmp     L0006
L0005:  lda     #$10
L0006:  cmp     ptr1
        bne     L0002
        lda     _j
        sec
        sbc     #$08
        sta     _j
L0002:  rts
```

Note in the failure case how "jmp L0004" jumps over "cmp ptr1" instead of to it. A is never compared to ptr1 and instead the Z flag will be the result of "lda #$00" which causes it to branch never (i.e. the == acts as if always true instead of performing its intended comparison).

bbbradsmith:
With --debug-opt-output it seems to fail at the first optimization rule applied: OptStackOps

```
=========================================================================
Initial code for function 'fail':
.segment	"CODE"

	ldx     #$00
	lda     _i
	ldx     #$00
	and     #$1F
	jsr     pushax
	lda     _r
	jeq     L0003
	ldx     #$00
	lda     #$00
	jmp     L0004
L0003:	ldx     #$00
	lda     #$10
L0004:	jsr     toseqax
	jeq     L0002
	ldx     #$00
	lda     _j
	sec
	sbc     #$08
	sta     _j
L0002:	rts

=========================================================================
Code after applying 'OptStackOps':
.segment	"CODE"

	ldx     #$00
	lda     _i
	ldx     #$00
	and     #$1F
	sta     ptr1
	lda     _r
	jeq     L0003
	ldx     #$00
	lda     #$00
	jmp     L0004
[...]

PR #1992, opened 2023-02-18
OptStackOps label migration fix

bbbradsmith:
Addresses bug found in #1989 affecting a rare case of OptStackOps optimization step where a label was generated exactly on the replaced stack operator.

Opt_a_toscmpbool had one branch which neglected to update its Op replacement insertion pointer after inserting code above it to replaced the Push.

Opt_a_tosicmp appeared to have a version of the same bug, though I could not come up with a test case to trigger it. Have added a prophylactic fix for it.


mrdudz:
thanks!

PR #1994, opened 2023-02-19
make avail: check for presence of binaries before creating symbolic links

Compyx:
To avoid creating broken symlinks, first check if the binaries exists in `bin/` and exit when they don't, with a message to first run `make`.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -110,22 +110,16 @@ $(RM) /usr/local/bin/$(prog)
 
 endef # UNAVAIL_recipe
 
-define AVAIL_check_prog
-
-@if [ ! -e ../bin/$(prog) ]; then \
-    echo "failed to locate $(prog), please run make first"; \
-    false; \
-fi
-
-endef # AVAIL_check_prog
-
 install:
 	$(INSTALL) -d $(DESTDIR)$(bindir)
 	$(INSTALL) ../bin/* $(DESTDIR)$(bindir)
 
 avail:
-	$(foreach prog,$(PROGS),$(AVAIL_check_prog))
+ifneq ($(patsubst %,../bin/%,$(PROGS)),$(wildcard $(patsubst %,../bin/%,$(PROGS))))
+	$(error executables are missing, please run make first)
+else
 	$(foreach prog,$(PROGS),$(AVAIL_recipe))
+endif
 
 unavail:
 	$(foreach prog,$(PROGS),$(UNAVAIL_recipe))
--- a/src/Makefile
+++ b/src/Makefile
@@ -110,11 +110,21 @@ $(RM) /usr/local/bin/$(prog)
 
 endef # UNAVAIL_recipe
 
+define AVAIL_check_prog
+
+@if [ ! -e ../bin/$(prog) ]; then \
+    echo "failed to locate $(prog), please run make first"; \
+    false; \
+fi
+
+endef # AVAIL_check_prog
+
 install:
 	$(INSTALL) -d $(DESTDIR)$(bindir)
 	$(INSTALL) ../bin/* $(DESTDIR)$(bindir)
 
 avail:
+	$(foreach prog,$(PROGS),$(AVAIL_check_prog))
 	$(foreach prog,$(PROGS),$(AVAIL_recipe))
 
 unavail:


mrdudz:
The idea is good - but i am not sure i like it is using the shell (our Makefiles don't use the shell).

@oliverschmidt any idea how this can be done just using make? I'd have to look it up and play around (but its probably possible... i guess)

oliverschmidt:
I.e. https://stackoverflow.com/questions/5553352/how-do-i-check-if-file-exists-in-makefile-so-i-can-delete-it

BTW: A consistent name for `AVAIL_check_prog` would be `PROG_CHECK_recipe`.

Compyx:
I tried the `ifeq` approach first, but the problem with that is that's is evaluated at parse time and we're trying to do the checks inside a rule. But i'll give it another shot.

rofl0r:
you could just foreach( wildcard ../bin ... instead foreach($PROGS)

Compyx:
I'm not sure what you mean.

Are you talking about the `foreach` loop that installs the symlinks? I assume the PROGS list is there to make sure it installs only symlinks to those programs, not everything that happens to be in ../bin/. And the current PR doesn't contain a `foreach` after the last commit.

rofl0r:
> Are you talking about the foreach loop that installs the symlinks?

right.

if you think that there may be stuff in ../bin that's not in $PROGS (though, why should it?) you can maybe use a filter.

Compyx:
If we use a glob to install the symlinks, then if the user didn't run `make` before running `make avail` there will be no warning and no symlinks installed. The point of this PR is to warn the user about running `make` first, not to silently do nothing.

rofl0r:
> The point of this PR is to warn the user about running make first, not to silently do nothing.

in that case you can just add $(PROGS) to prerequisites of the avail target, so make avail will build the progs first if they dont exist.

```make
avail: $(PROGS)
```

Compyx:
I know.

The disadvantage of that is that `make avail` requires root privileges, so when we add $(PROGS) as prerequisite for `make avail`, the binaries and all intermediate objects will be built as root, which means the user cannot run `make` or `make clean` again without first changing the ownership all of build objects back to theirs with `sudo` or `su`.

Which is why I didn't simply add $(PROGS) as dependency of `avail`.

Compyx:
Any movement on this? It's not exactly brain surgery I'm performing in this PR ;)

Compyx:
Thanks! =)

inexorabletash:
FYI, broke my workflow:

https://github.com/a2stuff/build-install-ca65-action/blob/016c3208a7ebfaa7b6b98f7b090dc481bca12239/action.yml#L8

which only does:

`sudo make -C /tmp/cc65 ca65 ld65 avail`

mrdudz:
> FYI, broke my workflow:
> 
> https://github.com/a2stuff/build-install-ca65-action/blob/016c3208a7ebfaa7b6b98f7b090dc481bca12239/action.yml#L8
> 
> which only does:
> 
> `sudo make -C /tmp/cc65 ca65 ld65 avail`
> 
> ETA: I updated my workflow to "manually" link ca65 and ld65 targets. Just a heads-up that others may have taken similar shortcuts.

To be honest - i am thinking about removing this "avail" target altogether, because it is very non-unix and "install" is the right thing to do anyway.

Compyx:
I agree removing `make avail` and having a proper `make install` (including not having to manually provide `PREFIX`, but defaulting to `/usr/local/`) would definitely remove some of the unpleasant surprises of trying to install cc65. It would also simplify distributing cc65 since `make avail` depends on the source tree remaining present, which is a very peculiar requirement, to put it mildly.

spiro-trikaliotis:
> It would also simplify distributing cc65 since make avail depends on the source tree remaining present, which is a very peculiar requirement, to put it mildly.

But distributing cc65 is no problem (provide a `PREFIX` and use `make install`).
Please, if the install target is changed, please make sure to do it in a way that the packetizer can react on it in a timely manner. Having a default `/usr/local` is not a problem, but removing make targets or changing their behaviour might be.

mrdudz:
> Having a default `/usr/local` is not a problem, but removing make targets or changing their behaviour might be.

I don't think any packager uses "make avail" :)

spiro-trikaliotis:
> I don't think any packager uses "make avail" :)

Me neither. But: It was mentioned that removing `make avail`  (and adding a default PREFIX) would simplify distribution. I don't think so: Providing a PREFIX should be very familiar for any packetizer, and `avail` is nothing one would do. So, I do not see how this would simplify distributions.

However, changing behaviour of `install` *beyond* providing a default `PREFIX` might be problematic, depending upon what it is exactly.

As I am packetizing cc65 myself, I only wanted to raise my voice to be careful when doing changes here.

Compyx:
Last time I used `make install` the assembler couldn't find the `.mac` files it required, they were installed in `$PREFIX/share/cc65`, but the assembler failed with a message about longjump. So I'd like to fix `make install` to the point you actually get a usable installation of cc65 that doesn't rely on the source tree being present.

Perhaps @mrdudz can move this into a discussion topic?

mrdudz:
> Perhaps @mrdudz can move this into a discussion topic?

Interestingly.... i cant do that here, for some reason (perhaps because it is merged?) - so better open a new discussion topic :)

Issue #1366, opened 2020-12-22
overwrite segment appends data incorrectly?


bbbradsmith:
Trying to use the overwrite segment type, it seems to append extra data after the segment it is trying to overwrite?

```
# CFG
MEMORY
{
	A: start = $0, size = $8, file = %O, fill = yes, fillval = $55;
	B: start = $8, size = $8, file = %O, fill = yes, fillval = $66;
}
SEGMENTS
{
	AA: load = A, type = ro;
	BA: load = B, type = ro;
	AO: load = A, type = overwrite, start = $4;
}

; assembly
.segment "AA"
.byte 0,1,2,3,4,5,6,7

.segment "AO"
.byte $33,$33

.segment "BA"
.byte $11,$12

; produces:
00 01 02 03 33 33 06 07 55 55 11 12 66 66 66 66 66 66
```
Notice in the product two extra "fill" bytes seem to appear at the end of the MEMORY block containing an overwrite. The overwrite itself seems to have happened correctly, but we now have extra bytes inserted in the middle of the file to compensate?

[test.zip](https://github.com/cc65/cc65/files/5728505/test.zip)

Notification to @laubzega whom I think created this feature?

bbbradsmith:
Another note: the extra fill seems to only occur if the containing MEMORY block was already full. E.g. if I remove one byte from "AA" in the example, we've only got 1 extra byte instead of 2. If I remove two bytes, I get no extra bytes, though there are still 2 padding bytes, they are what we'd expect with the specified size and fill.

If I remove 3 bytes from "AA" I get instead:
```
ld65: Internal Error: Invalid file offset for memory area B: 8/7
```

Not sure quite what's happening in that case. So, it seems like without a full MEMORY block it produces an internal error. With one that's nearly full or full we silently get some erroneous extra fill.

laubzega:
Hi Brad, thanks for the detailed report. I'll look into this during Xmas break. @oliverschmidt, can you please assign the issue to me? Thanks!

oliverschmidt:
@laubzega: Thanks for your quick reaction - and willingness to look into the issue.

HertzDevil:
This line:

https://github.com/cc65/cc65/blob/4acdc9ced9ded4c37e46d2bc531bea31894c1ccc/src/ld65/config.c#L2081-L2084

[...]

PR #1995, opened 2023-02-19
ld65 fix overwrite segments adding wrong padding or causing internal errors

bbbradsmith:
Fix for #1366 based on HertzDevil's suggestion. Overwrite segments no longer will cause accidental extra padding, and no longer produce internal errors. If the overwrite extends into area not filled by the previous segments, it will gracefully extend as should be expected.

### DIFF
--- a/src/ld65/bin.c
+++ b/src/ld65/bin.c
@@ -193,8 +193,16 @@ static void BinWriteMem (BinDesc* D, MemoryArea* M)
                     NewAddr += M->Start;
                 }
                 if (DoWrite || (M->Flags & MF_FILL) != 0) {
-                    /* Seek in "overwrite" segments */
                     if (S->Flags & SF_OVERWRITE) {
+                        /* Seek in "overwrite" segments. Fill if the seek position has not been reached yet. */
+                        unsigned long FileLength;
+                        unsigned long SeekTarget = NewAddr - M->Start + M->FileOffs;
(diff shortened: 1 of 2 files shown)


mrdudz:
> If the overwrite extends into area not filled by the previous segments, it will gracefully extend as should be expected.

just to make sure: in the example given in #1366 it would still limit the size of the segment AO to 8 bytes, correct?

bbbradsmith:
Yes, the size of the memory area is enforced, and you will get an overflow error if your overwrite segment would go past the end of it.

What I meant by gracefully extend is that previously this feature required that for every byte in the overwrite segment, a previous segment must have already placed data there. Fill bytes from unfilled parts of the memory area did not count and could not be overwritten. With my changes this is no longer a requirement. It should simply overwrite bytes in the memory area, without having to worry about whether a previous segment was filling those bytes or not.

bbbradsmith:
One thing I didn't mention. Though I did test it locally with many permutations of overlap, I couldn't find anything suitable in the existing test/ set that I could easily add a relevant test to.

So, in the future I do think it could use some tests to prevent regression, so I've left an issue as a note about it with #2008 but I didn't think it was worth delaying this PR with that want, since the feature previously felt quite broken, and the existing tests should hopefully at least be verifying that the changes didn't break other features of the linker.

mrdudz:
Ok, sounds good then. And yes, i agree there should be more tests :)

Issue #1390, opened 2021-01-25
wrong MEMORY block definition caused 26GB file to be created on disk


shazz:
Hi,

As I'm generating the memory configuration file by code, I made a mistake which created a funny behavior with cl65.
If the MAIN memory section was declared like this:

`MAIN:     file = %O, start = %S,     size =  - %S,   fill = yes;`

it creates a file of at least 26GB on disk :)

groessler:
The bug is in your cfg file.

mrdudz:
Should probably still error out after a reasonable maximum size :)

shazz:
I agree the configuration is faulty but that's not a good reason to fill my disk :)

greg-king5:
ld65 saw that attribute as
```c
size = (unsigned long)(0L - (long)S)
```
I agree with @mrdudz.  ld65 should reject negative sizes before it casts them to `unsigned`.

mrdudz:
closed via #1996

PR #1996, opened 2023-02-19
ld65 error if computed memory size is negative

bbbradsmith:
Simple error check to address issue #1390


bbbradsmith:
This seems to have exposed an error with one of the dasm tests.

mrdudz:
looks good, thanks!

Issue #1759, opened 2022-06-03
confusing sentences about .A16 and .A8


jeffythedragonslayer:
I think this sentence in the documentation for `.A16` is kind of confusing:

> Switch the accumulator to 16 bit.

ditto for this one for `.A8`:

> Switch the accumulator to 8 bit.

The notes below them may not make sense to newbies who haven't yet learned what an operand or addressing mode is.  I think these sentences would be better phrased as:

> Tells the assembler that the accumulator is 16 bits wide

and

> Tells the assembler that the accumulator is 8 bits wide

because `.A16` and `.A8`, being assemble-time directives, definitely do not switch the CPU accumulator width.  These directives only change what the assembler thinks the accumulator width is.  It is possible to use these directives to lie to the assembler.  The word "switch" is also used as a verb in the documentation for `.I16` and `.I8`.

jeffythedragonslayer:
opened in wrong repo - closing and opening in doc

jeffythedragonslayer:
I couldn't find an "Issues" tab in the doc repo for some reason so I'm reopening this.

greg-king5:
This is the correct repo.

The source files for the documents are here (https://github.com/cc65/cc65/tree/master/doc).
Fixes to the docs are done here.

mrdudz:
fixed via #1997

PR #1997, opened 2023-02-21
ca65 doc for A8/A18/I8/I16: use "assume" instead of "switch" to avoid implying that a hardware state is changed

bbbradsmith:
See: #1759


mrdudz:
thanks!

Issue #1563, opened 2021-09-21
FR: ca65: add warnings-as-errors option


inexorabletash:
I ran into this while moving some code around:
```
ca65 -o target.o target.s
target.s:1234: Warning: "jmp (abs)" across page border
ld65 --config asm.cfg -o target target.o
ld65: Warning: target.s:1234: "jmp (abs)" across page border
```
... but the build didn't fail so I thought _"okay, that's just warning, that's nice"_

And then I read about [what the JMP instruction does when the target straddles page boundaries](http://www.6502.org/tutorials/6502opcodes.html#JMP) and freaked out. This wouldn't cause CI to fail which seems... bad.

I can work around this by failing the build if STDERR is non-empty, but can we get a "treat warnings as errors" option?

I guess this should apply to both ca65 and ld65, and should apply to the current warning level.

skeetor:
IMO this should be an error, not just a warning.

mrdudz:
Adding this option sounds like a good first-time contribution project :)

bbbradsmith:
In lieu of a first-time contributor, I wanted to have this feature too. PR #1999

mrdudz:
MMh is this really how you'd expect this to work? Like this it will first spit out all warnings, and then error out - instead of error out after a warning (like it would do when it would be an error instead), right?

bbbradsmith:
I put the error as late as I could before it generates and output file, that way you get a list of all warnings and not just the first one. I don't think there would be any advantage in stopping only at the first warning?

As in many compilers, we also have in ca65 types of errors that do not immediately terminate the build, so that it can continue and generate a more complete/useful list of errors before eventually terminating with an error. I feel the same applies to warnings-as-errors?

Though, I'm realizing now that OP's request is actually for jmp (abs) across a page to be an error instead of a warning? I'd probably agree with that one. Not sure when it would ever be useful to only warn on that? If it really has to be suppressed, there's always the manual option:
```
.byte $6C ; jmp (abs)
.word pointer
```

bbbradsmith:
I added promotion of jmp (abs) warning to an error to PR #1999 as it seemed convenient to do there instead of a separate PR. This also suppresses the error on 65C02 and later CPUs which don't have that quirk.

mrdudz:
closed via #1999

PR #1999, opened 2023-02-21
--warnings-as-errors for ca65 and ld65

bbbradsmith:
Addresses #1563


bbbradsmith:
In the second commit I promoted the "jmp (abs)" wrap warning to an error, which was the actual original request in #1563. I've also limited it to only 6502 or unknown CPU types, since this behaviour was fixed on 65C02 and later.

groessler:
I don't know if I like "jmp (abs)" as an error. warnings-as-errors looks good enough to me.

mrdudz:
> I don't know if I like "jmp (abs)" as an error. warnings-as-errors looks good enough to me.

I agree, by default only actually invalid code should produce an error

bbbradsmith:
I'm confused about when "jmp (abs)" with a page wrap would ever be valid code?

mrdudz:
> I'm confused about when "jmp (abs)" with a page wrap would ever be valid code? It's universally described as a bug in the processor, and fixed in 65C02 and all subsequent revisions.

It's valid code in the sense that is syntactically correct - and thus should be accepted by the assembler.

bbbradsmith:
This doesn't have anything to do with syntax? This is generally a link-time error, indicating the CPU will do the wrong thing if "abs" is one of several specific addresses only. It's only an assemble-time error if the "abs" address is known at assemble time.

I can't conceive of a case where the page wrap is intended behaviour, except when writing code intended to specifically test it. The resulting code will produce invalid results due to the linking problem that is not visible at all in the syntax.

mrdudz:
> Having to use --warnings-as-errors to catch this case seems like an irritating workaround to me. While --warnings-as-errors I think has value by itself, it's not a good solution to catching the jmp (abs) wrap case.

I'd say its the expected behaviour (and every other assembler i know works like this)

bbbradsmith:
I feel like we can't possibly be describing the same thing? Can you point me at an example of code somewhere that uses jmp (abs) with a wrap expected, that is not simply a test for the 6502's bug behaviour?

To illustrate:
```
jmp ($8005) ; valid, jumps to address at $8005,$8006
jmp ($80FF) ; error on 6502, jumps to address at $80FF,$8000
jmp (table) ; how do we know if this is an error? will it fetch table,table+0 or table,table-255?
```
There is no way for the programmer to see that table might straddle a page boundary. The linker decides where to put it, only the linker knows where it's going. Why would we want to expect that 1 in 256 locations for table are going to jump somewhere radically different?

mrdudz:
> how do we know if this is an error? will it fetch table,table+0 or table,table-255?

Thats why you get the warning

groessler:
@bbbradsmith, why is the warning not sufficient?

bbbradsmith:
I'm very surprised by the question. I'd ask why the warning is sufficient?

If you JMP to the wrong location, the program _will crash_. What other possibility is there except error?

mrdudz:
To me this is no different than eg ```lda($ff),y``` - when i write this, i don't what the assembler to produce an error.

groessler:
I don't think it's the job of the toolchain to outwit the user.

If such a jmp is by mistake, the warning should suffice.
If it's deliberate, the user should be allowed to do it.

bbbradsmith:
I would also want an error for lda (zp),Y with zp crossing a page, but it is at least a much less severe case. There is exactly one affected location, and it will only happen if you forget to reserve 2 bytes instead of 1 for a pointer, and that point happens to fall exactly on $FF.

With "jmp (abs)" there are a lot more potential affected locations, and the result is jumping to completely unexpected location. I've had this warning come up and not been able to catch it right away for the same reason as #1563. Warnings are helpful but easy to miss, errors are a much more direct indication of the problem. If you want to catch all warnings right away, --warnings-as-errors is very helpful, but there's definitely a reason we have both warnings and errors, and this assembler does not have much ability to filter or control spurious warnings.

I asked above why you would ever do it deliberately in code. I really can't see any possible use case for it, except as a CPU test for the bug. Have you seen a practical example that makes it worth doing?

There are multiple ways to do it deliberately:
```
.byte $6C, $FF, $80 ; jmp ($80FF) with intentional wrap

.pc02
jmp ($80FF) ; no error because 65C02 doesn't have the bug
```

I would agree that this specifically looks deliberate:
```
lda ($FF), Y
```

But the normal usage of that instruction is:
```
lda (label), Y
```

I would never interpret that as deliberate wrapping on sight. The linker knows where label is, not the coder. I feel like the linker here might "outwit" the user by putting label on a page boundary.

bbbradsmith:
I'll remove the lda (abs) wrap warning promotion from the PR and start a separate issue for this.

mrdudz:
Thanks, this way i can merge it

PR #2000, opened 2023-02-21
Document expected linker configuration and special segments used by cc65

bbbradsmith:
I couldn't find any explanation of the special segments used by the compiler, so I added a description of the expected linker configuration to the cc65-internals document.


mrdudz:
cool, thanks!

Issue #892, opened 2019-05-11
Automatic promotion of 65816's JSR to JSL is a bad idea.


greg-king5:
(Oops, Github posted this issue before I could write the description.)

Ca65 shouldn't promote `JSR` opcodes automatically to `JSL` when the operand is a far address.  That action "hides" the fact that the addressed subroutine must return through `RTL` instead of `RTS`.  Ca65 should give an error message instead.  That will force a programmer to correct the instruction -- which will remind that person to fix the corresponding `RTS` instructions.

bbbradsmith:
I don't like it either, though the use of `JSR` as a mnemonic for both near and far jumps seems to have some historical use in older assemblers, unfortunately. This applies to `JMP` as well.

We also have the `.smart` mode which automatically replaces `RTS` with `RTL` in a `far` context. With that mode enabled, I'd think it would have to be acceptable to have `JSR` promote like that.

However, as I mentioned in #879 I'd personally like an option to disable `JSR` and `JMP` as mnemonics for far jumps. I think even in `.smart` mode, some might want `REP`/`SEP` tracking without the `RTS` replacement too, so it might help if these features were more individually selectable with a separate directive to control the processor flags assumption vs. the jump/return assumptions.

The pull #885 that I submitted at least fixes a big source of unintended errors through that ambiguity, as an accidentally under-promoted `JSR` was not being detected as an error at all.

bbbradsmith:
I created PR #2001 to prevent JMP / JSR and RTS from assuming long instruction variations by default, and then created a new feature "long_jsr_jmp_rts" to re-enable it for compatibility with older code.

However, while looking into it, I also discovered that the existing promotion of JMP/JSR only works for .segment and .import declared as far, and ignores the .proc declared as far, which is a bit vexing, as I think it makes the old behaviour even more error prone than it seemed.

What I discovered is that the proc : far attribute is stored as the "AddrSize" property of a scope, but this is only checked in exactly one place in the code: https://github.com/cc65/cc65/blob/d0f17ba60251debf8d91fc6bdac103ffb6ede7c8/src/ca65/instr.c#L1635

So, it seems that the only function of "proc : far" is to turn RTS into RTL under smart mode, and otherwise it does nothing. It really should also affect JMP/JSR in the same way (promoting them to JML/JSL if the label is in such a scope). I consider this a bug that it does not.

[...]

PR #2001, opened 2023-02-21
ca65 jsr/jmp/rts will not promote to jsl/jml/rtl by default, new feature long_jsr_jmp_rts to re-enable this ability

bbbradsmith:
This makes the behaviour suggested by #892 the default, i.e. JSR and JMP do not emit long instructions automatically, and RTS will not become RTL automatically under smart mode.

It also adds a feature long_jsr_jmp_rts that re-enables the former behaviour for compatibility with former code that was using it.


mrdudz:
looks good, thanks!

PR #2003, opened 2023-02-21
supress spurious "large alignment" warnings

bbbradsmith:
There is a warning that attempts to diagnose accidentally "large" alignments when combining mismatched alignment sizes. It put an arbitrary size limit on all alignments that anything > 256 should be a warning.

However, if you request an alignment of 1024 or 32768 or 65536, and that size isn't being increased, this should not be a warning. You're getting exactly what you explicitly asked for. There should be no need to use the cumbersome "--large-alignment" command line option on both the assemble and link to accommodate this need.

I have modified the warning to only trigger if the combined alignment is greater than any of the alignments you actually asked for.

### DIFF
--- a/doc/ca65.sgml
+++ b/doc/ca65.sgml
@@ -2112,14 +2112,15 @@ Here's a list of all control commands and a description, what they do:
   </verb></tscreen>
 
   the assembler will force a segment alignment to the least common multiple of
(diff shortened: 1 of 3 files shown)


mrdudz:
makes sense to me, thanks!

Issue #2005, opened 2023-02-24
jmp (abs) on a page boundary should be an error on 6502 rather than a warning because of the well known CPU bug


bbbradsmith:
I had originally implemented with as part of PR #1999 thinking it was an uncontroversial add-on, but after some resulting debate about it, felt it was inappropriate baggage for that PR. I felt I should raise it as its own issue rather than let it block that PR.

New PR for just this change: #2006 

I will try to summarize the debate had at #1999 but please read for full context.

On the 6502:
```
jmp (label)
```

This can jump to an unexpected location, if the linker happens to place _label_ on a page boundary. E.g. if label is $80FF it will jump to the addressed fetched from $80FF,$8000 rather than $80FF,$8100. Guides for 6502 programming seem to universally describe this as a bug of the CPU, and the 65C02 and later revisions indeed treated it that way and corrected it. Example: [NMOS 6502 Opcodes: JMP](http://www.6502.org/tutorials/6502opcodes.html#JMP) article in bold states "AN INDIRECT JUMP MUST NEVER USE A VECTOR BEGINNING ON THE LAST BYTE OF A PAGE".

We have a warning for this. My PR promoted this warning to an error, but also made the exception that 65C02 and other processor variations that don't have this bug won't produce the error.

I felt it should be an error because it produces invalid code that will crash the program, in a way that the programmer can't predict because it was the linker's job to place the label. The user who opened issue #1563 implied they felt it should be an error as well, but was more explicitly requesting ```--warnings-as-errors```, which as a general solution can promote this warning to an error. Of course that promotes all other warnings as well, and we don't have very good warning filtering/suppression control on this assembler to begin with.

Some points from mrdudz:

> It's valid code in the sense that is syntactically correct - and thus should be accepted by the assembler.
> I'd say its the expected behaviour (and every other assembler i know works like this)
> To me this is no different than eg lda($ff),y - when i write this, i don't want the assembler to produce an error.

Some comments from groessler:

> why is the warning not sufficient?
> I don't think it's the job of the toolchain to outwit the user.
> If such a jmp is by mistake, the warning should suffice.
> If it's deliberate, the user should be allowed to do it.

My feeling is that the we should validate more than just syntax in this case. I would also argue that ```lda (zp), Y``` should have an error against the similar wrap, but I consider it a much less severe case. I would accept that ```lda ($FF), Y``` does look deliberate but I feel that ```lda (label), Y``` is the normal case.

[...]

PR #2006, opened 2023-02-24
jmp (abs) wrap warning promoted to an error, suppressed on 65C02/etc

bbbradsmith:
This change may be controversial, I opened issue #2005 to discuss it.

Originally this was an add-on to #1999 but because of the unexpected debate, I have separated it. This is also related to issue #1563 .


bbbradsmith:
It was [requested here](https://github.com/cc65/cc65/issues/2005#issuecomment-1685080985) that `--relax-checks` should suppress this error, so I have added this, as well as a fix to the documentation for that feature which said it suppressed another error which it does not.

mrdudz:
Looks OK to me this way, thanks

Issue #1973, opened 2023-01-15
printf("%c", '\0') doesn't _write '\0' to fd 0


patrickleboutillier:
I stumbled upon this while making a custom target:

`printf("%c", '\0') ; // does not result in a call to _write`

It seems like the character is placed in a 1-byte long string and then treated as such, which causes the '\0' not to be printed since it is the string terminator.

I checked in gcc and that call actually prints '\0' to fd 0. I know I can call _write directly as a workaround, but I spent some time investigating this so I though I should report it. I didn't find any mention of this in the docs.

mrdudz:
Clearly a bug :)

groessler:
I think the problem is here (libsrc/common/_printf.s, line 665++):

```
; We have the argument, do argument string formatting

HaveArg:

; ArgLen = strlen (Str);

        lda     Str
        ldx     Str+1
        jsr     _strlen                 ; Get length of argument
        sta     ArgLen
        stx     ArgLen+1
```

`strlen()` doesn't work in this case....

groessler:
a hackish solution: (maybe even good?)

```
--- a/libsrc/common/_printf.s
+++ b/libsrc/common/_printf.s
@@ -505,7 +505,9 @@ DoFormat:
         sta     Buf                     ; Place it as zero terminated string...
         lda     #0
         sta     Buf+1                   ; ...into the buffer
-        jmp     HaveArg                 ; Done
+        tax
+        lda #1
+        jmp x1
 
 ; Is it an integer?
 
@@ -671,6 +673,7 @@ HaveArg:
         lda     Str
         ldx     Str+1
         jsr     _strlen                 ; Get length of argument
+x1:
         sta     ArgLen
         stx     ArgLen+1
 
```

mrdudz:
closed via #2007

PR #2007, opened 2023-02-24
fix for issue #1973 (printf("%c", '\0') doesn't _write '\0' to fd 0)

groessler:


polluks:
The code is even smaller :-)

mrdudz:
merging, thanks!

PR #2009, opened 2023-02-24
ca65: allow .feature to both enable and disable

bbbradsmith:
This PR now allows ca65's .feature to disable a feature mid-file. Previously it could only be enabled mid-file, or at the command line.

This uses the existing SetBoolOption syntax, which is fully backward compatible with how .feature previously worked, allowing now an optional + or - to override the default enable with an explicit enable or disable.

Command line --feature is the same as before, as all features are disabled by default, there is no need to be able to disable them from the command line.

If there is a need to disallow mid-file disable of any of these, code could be added to the switch case in feature.c, but I haven't identified any that would need this currently.

All of the existing features for the most part appear to have very localized effects, usually just applying to the parsing of a single line at a time, so there shouldn't be any unexpected/unintended effects if disabled later.

The ```org_per_seg``` feature seems slightly different. I think this should normally be enabled or disabled only between segments, but it was already possible to enable mid-segment. I believe this will just switch to the last known per-seg or global org state when enabled/disabled, which I think makes sense, but might not be totally obvious. Anyway, this doesn't seem to be a new caveat, just the reverse case of an existing one.


bbbradsmith:
Noticed .feature doesn't have any tests currently, I'll add some.

mrdudz:
Looks ok, merging this - it shouldnt have impact on existing code

PR #2010, opened 2023-02-25
optional BRK signature on all 6502 CPUs, not just 65816

bbbradsmith:
I noticed we have the option of a 2nd byte for BRK but only on 65C816? (Working since d13d068)

Added the same capability to all 6502 derived CPUs, and also for 65C816 COP which is functionally the same as a BRK.

Assemblers seem to disagree on whether 6502 BRK should be 1 or 2 bytes, but having both options accomodates both usages. I think this is useful for all of them, not just 65C816. Also added the 1-byte option for COP since it is functionally equivalent to BRK.

This should be fully compatible with pre-existing ca65 code, since BRK with a parameter was an error previously.

Also added some information about MVN and MVP syntax to the ca65 document while I was documenting this optional BRK behaviour. (As they've been since: ad6c2db)


mrdudz:
while i am not opposing the idea itself, i'd rather have it match the usual syntax of an immediate parameter - this way it looks like zeropage addressing

bbbradsmith:
> while i am not opposing the idea itself, i'd rather have it match the usual syntax of an immediate parameter - this way it looks like zeropage addressing

The 65816 version has already supported this as a direct parameter for 5 years, so it would be a breaking change to switch to only # prefix. However, it should be easy to add it as a 3rd variation.

So far in other assemblers I've only seen the BRK parameter be direct, not with the # syntax.

mrdudz:
> The 65816 version has already supported this as a direct parameter for 5 years, so it would be a breaking change to switch to only # prefix. However, it should be easy to add it as a 3rd variation.

"3rd variation" sound like a terrible idea to me. And we are not afraid of breaking changes when they restore sanity :=)
Is there an "official" syntax for this in 65816 world? If not, immediate still makes a lot more sense to me (because that's what it works like)

bbbradsmith:
I don't agree that it's insane at all, and it doesn't function like an immediate. It's not a byte used by the hardware, so the semantics of it entirely depend on what the BRK handler wants to do with it.

WDC's official datasheet specifically refers to the operand as type S for signature, and not # for immediate, but it does not provide assembler syntax.

Assembler syntaxes I've come across so far:
* ca65 - direct only
* wla-dx - direct only
* 64tass - prefers immediate (not sure if it accepts both)
* Merlin - accepts both
* ACME  - direct only
* asar - immediate only
* Eyes & Lichty - "Programming the 65816" apparently notates it in the direct form

I personally don't see anything wrong with accepting both? There isn't really any ambiguity, it just allows two different style preferences, and it makes it easiest for people coming from other assemblers. It feels as normal as accepting both <tt>LSR</tt> and <tt>LSR A</tt> to me.

bbbradsmith:
Added the immediate style as a variation in 3d41a5b to show what the change would look like if both styles were accommodated. Can amend this whatever permutation we have to go with, but I do think accepting both would probably be the most useful, since it's clear that the preferences vary for this one.

bbbradsmith:
Since @greg-king5 implemented the 65816 version of this in the first place they might have useful commentary about the syntax.

mrdudz:
Greg is MIA...

Anyway, i am merging this as is :)

PR #2011, opened 2023-02-25
Using .define with .endmacro in a macro definition (bug fix)

Movax12:
When using `.define` to create a new symbol that references the `.endmacro` command outside of a macro, it works as might be expected:
```
.define startmac .macro
.define endmac .endmacro

startmac test
    .out "Test"
endmac 

test
```
This will result in console output:
`Test`

Attempting to do the same, but defining the endmac symbol inside a macro however:
```
.macro makeStartMac
    .define startmac .macro
.endmacro

.macro makeEndMac 
    .define endmac .endmacro
.endmacro
```
This is as far as the assembler gets while resulting in:
```
... : Error: Unexpected '.ENDMACRO'
```
While it is unlikely that most users will run into this issue, ca65's behavior is not consistent in this case. This patch fixes this behavior.


Movax12:
Going to redo this pull request, as my macro.c file looks to be too different from recent changes.

PR #2012, opened 2023-02-25
Fix endmacro with define in macro body

Movax12:
When using .define to create a new symbol that references the .endmacro command outside of a macro, it works as expected:
```
.define startmac .macro
.define endmac .endmacro

startmac test
    .out "Test"
endmac 

test
```
This will result in console output:
`Test`

Attempting to do the same, but defining the endmac symbol inside a macro however:
```
.macro makeStartMac
    .define startmac .macro
.endmacro

.macro makeEndMac 
    .define endmac .endmacro
.endmacro
```
This is as far as the assembler gets while resulting in:

`... : Error: Unexpected '.ENDMACRO'`
While it is unlikely that most users will run into this issue, ca65's behavior is not consistent in this case. This patch fixes this behavior.

Note: The current code also allows `.endmacro` to be anywhere at the end of a line. For example, this works:
```
.macro sayHi 
    .out "Hi!" .endmacro
```

### DIFF
--- a/src/ca65/macro.c
+++ b/src/ca65/macro.c
@@ -590,7 +590,7 @@ void MacDef (unsigned Style)
         NextTok ();
     }
 
-/* Reset the Incomplete flag now that parsing is done */
+    /* Reset the Incomplete flag now that parsing is done */
     M->Incomplete = 0;
 
 Done:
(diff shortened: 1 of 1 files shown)


Movax12:
I now realize that the last point is the real problem and was not addressed by this request.
"Note: The current code also allows .endmacro to be anywhere at the end of a line. "
New request incoming.

PR #2013, opened 2023-02-25
Fix endmacro not at start of line

Movax12:
When defining a new classic macro, ca65 looks for `.endmacro`  without checking for an EOL separator.
This allows macros to be ended with `.endmacro` anywhere at the end of a line, and also terminates a `.define` that defines a symbol that references `.endmacro`.
This request addresses this issue.


bbbradsmith:
I'm a little bit confused about what this fixes or permits. Is this valid code, and would it produce a macro that has an .endmacro token inside it?
````
.macro inc2
    inc
    inc .endmacro
.endmacro
```

Movax12:
You example would be equivalent of:
```
.macro inc2
    inc
    inc 
.endmacro
.endmacro
```

bbbradsmith:
I see, so I did understand the code correctly.

I wasn't sure why you wanted .endmacro to be permitted inside a .macro, which it sounded like you were explicitly carving out permission to do.

Is .endmacro token within a .macro something you think has value to be permitted? What was confusing to me is why the not-start-of-line .endmacro is being added to the macro instead of generating an error. The language chosen says "ignored" but it's not ignored, it's actually added to the macro.

So, is that desired behaviour, or are you relying on it generating an error elsewhere when used? If it's not desired, should it just error at the not-start .endmacro? I think that would be easier to diagnose for the user.

```
.macro inc1
    inc .endmacro
```

This, for example, is going to produce an unclosed macro error now, right? I don't think the error would explain to me how this needs to be fixed, because it looks closed.

So, is a classic .macro allowed to contain .macro and .endmacro? Do we want to use this to use .macro to create more macros? Or are we just trying to fix a current loophole in the syntax that should just be an error?

Movax12:
The problem that led to this was the inconsistency with the following examples:
This works: (not part of a macro body)
```
.define endmac .endmacro
``` 
This does not:
```
.macro newEndmacro
    .define endmac .endmacro
.endmacro
```
This will work with this patch.

The assembler will not let you define a macro within a macro, because the `.endmacro` cannot be part of a macro definition twice. I don't believe having an `.endmacro` included as part of the macro definition is a problem, as any valid tokens can be a part of a macro definition. I also don't see a problem with an unclosed macro error, as I don't think the syntax of ca65 implies that control commands can be used at the end of a line.

Movax12:
> The language chosen says "ignored" but it's not ignored, it's actually added to the macro.
How about:
/* In classic macros, only .endmacro is allowed, but do no exit the macro definition if not at the start of a line */
?

bbbradsmith:
I agree that control commands aren't valid at the end of a line, what I think is a problem is that it's only an implication.

So what I meant with this example:
```
.macro inc1
    inc .endmacro
```
It will produce an error. Good.

However, the error is: ```Error: '.ENDMACRO' expected``` The line number will be the end of the file. No useful information about where to find your incorrect usage or what it might be. It's just "guess which implication you misinterpreted". This situation is probably worth improving.

But I wasn't sure if you were merely trying to disallow .endmacro off the end of the line or explicitly permit it by creating a special case for it, which you clarified with this example:

```
.macro newEndmacro
    .define endmac .endmacro
.endmacro
```

So, you do want to have .endmacro tokens allowed inside a .macro, and you do want to use "not at the start of the line" as a special case for that. It's creating a situation where .endmacro has a special function, under a condition that seems impossible to guess.

So, if this special case was indeed the goal, I'm making three suggestions:

1. Document this special case in ca65.sgml (i.e. explain .endmacro has special meaning at the end of a line within a macro, provide example)
2. Add a test/ case for this, which will make sure this desired special case doesn't get broken by accident, and provides indirectly an example of what its intent was.
3. If possible, improve the unclosed macro error for people who weren't intending this special case.

I think for 3, using LIError instead of Error to provide the starting .macro line location might be relatively simple to do, and would give the user a much better clue as to where the offending code might be. If the special case is also documented, that's probably enough.

bbbradsmith:
> How about: `/* In classic macros, only .endmacro is allowed, but do no exit the macro definition if not at the start of a line */`

Ah, yes it was the comment in the code that made it unclear for me what intent of the change was, though ultimately I think this intent should be documented in ca65.sgml so that it's user-facing.

I'm unclear on what "only .endmacro is allowed" means in that sentence. I think this is how I'd describe what I understand of the intent:
`/* In classic macros, if .endmacro is not at the start of the line it will be added to the macro definition instead of closing it. */`

mrdudz:
It would be great if you would add a test for this to this PR

mrdudz:
looks good... one more thing, would you rename the tests into something more descriptive? that makes it easier to tell what is what (like "endmacro.s" or sth like that)

bbbradsmith:
Also looks good to me.

mrdudz:
cheers

PR #2027, opened 2023-03-07
ca65 jmp (abs) wrap warning only applies to 6502, later CPUs do not have this bug

bbbradsmith:
`jmp (abs)` has a warning case currently which is only relevant to the original 6502. All later 6502-family CPUs have corrected the hardware bug.

This pull request eliminates the warning for later CPUs.

A separate PR #2006 had included this along with promotion of that warning to an error, but since it was contested (see issue #2005), I have separated just the CPU filter change for this PR.


mrdudz:
cheers

Issue #2008, opened 2023-02-24
test suite should include some tests of custom linker CFG


bbbradsmith:
While working on #1995 I couldn't find anything in our existing test/ suite that would allow me to assemble a file, link with a custom CFG, and then compare the resulting binary against a reference.

test/asm/listing/ was the closest, but it always links with -t none (none.cfg / none.lib).

While I think we do have a lot of tests of the linker being done as a part of other tests, there don't seem to be any that directly address customizing CFG files, so features like "overwrite" that aren't part of the standard C support CFG don't currently get any test at all.

Ideally, we should also have an err/ variation of custom CFG tests that can verify that error conditions are correctly produced.

bbbradsmith:
Was addressed by #2028

PR #2028, opened 2023-03-07
Linker test capability and overwrite segment tests

bbbradsmith:
This adds to test/asm/listing an additional capability to provide a custom CFG file for the linker. (#2008)

This new feature is used to create tests for the "overwrite" segment feature following up on #1995.


mrdudz:
Just curious.... What are the changes in config.c for?

mrdudz:
I see just wondering.... looks good, merging

PR #2032, opened 2023-03-17
[Telestrat Target]  Add fseek to telestrat target (Orix)

jedeoric:
This PR add fseek to telestrat target (Orix Os)


mrdudz:
Looks good to me, will merge soon if noone objects

PR #2037, opened 2023-03-27
Fixes and improvements to KIM-1 i/o routines.

jefftranter:
Tested on real KIM-1 hardware.

read.s:
- Remove commented out line.
- Remove unused check for bell character.
- Remove echo of newline (hardware always echoes entered characters).
- This fixes gets() and fgets() so they return when CR is entered.

write.s:
- Fix check for adding return after linefeed (failed to work because OUTCHR changes A)
- Remove unused check for bell character.

kim1.inc:
- Add symbol for monitor entry

crt0.s:
- Jump to KIM-1 monitor by address rather than using BRK (which relies on vector being set in RAM)


mrdudz:
i can't judge this :) looks ok, thanks!

PR #2042, opened 2023-03-29
Add Additional KIM-1 Functions

jefftranter:
Add KIM-1 functions to write to the 7-segment LED display and get keypresses from the keypad.
Includes sample program illustrating how to use them. Tested on real KIM-1 hardware.


mrdudz:
looks good, thanks!

PR #2049, opened 2023-04-06
Add profiling support to sim65

clbr:
Sample output:
```
CPU cycles total: 31868040

14.3%   4570624 _fwrite
13.6%   4326104 pushw0sp
9.9%    3164228 addysp1
9.7%    3092198 _tochr
9.4%    2992608 _decomp_1bit
8.5%    2702204 _decomp_ancestor
6.2%    1972920 stax0sp
5.8%    1856208 _decomp_3bit
5.6%    1775678 tosudiva0
3.8%    1224926 _stdin
```


util/profile/sim65-prof.c
```diff
@@ -0,0 +1,189 @@
...
+static int byaddr(const void *ap, const void *bp)
+{
+    const struct label_t *a = (struct label_t *) ap;
+    const struct label_t *b = (struct label_t *) bp;
+
+    if (a->addr < b->addr)
+        return -1;
+    if (a->addr > b->addr)
+        return 1;
+    return 0;
```
mrdudz, line 50: please always use braces for if statements

util/profile/sim65-prof.c
```diff
@@ -0,0 +1,189 @@
...
+static int bycount(const void *ap, const void *bp)
+{
+    const struct label_t *a = (struct label_t *) ap;
+    const struct label_t *b = (struct label_t *) bp;
+
+    if (a->count > b->count)
+        return -1;
+    if (a->count < b->count)
+        return 1;
+    return 0;
```
mrdudz, line 62: please always use braces for if statements

util/profile/sim65-prof.c
```diff
@@ -0,0 +1,189 @@
...
+
+static struct label_t *findlabel(const unsigned short addr)
+{
+    unsigned i;
+    for (i = 1; i < numlabels; i++) {
+        if (labels[i - 1].addr <= addr && labels[i].addr > addr)
+            return &labels[i - 1];
+        else if (i == numlabels - 1 && labels[i].addr <= addr)
+            return &labels[i];
+    }
```
mrdudz, line 73: please always use braces for if statements

util/profile/sim65-prof.c
```diff
@@ -0,0 +1,189 @@
...
+
+    if (optind + 2 > argc)
+        usage (argv[0]);
+
+    profile = argv[optind];
+    label = argv[optind + 1];
+
+    // samples
+    FILE *f = fopen (profile, "rb");
```
mrdudz, line 112: Please use C comments only

util/profile/sim65-prof.c
```diff
@@ -0,0 +1,189 @@
...
+
+    if (fread (samples, 1, sizeof(samples), f) != sizeof(samples)) {
+        printf("Invalid profile\n");
+        return 1;
+    }
+
+    fclose (f);
+
+    // labels
+    f = fopen (label, "r");
```
mrdudz, line 126: Please use C comments only

util/profile/sim65-prof.c
```diff
@@ -0,0 +1,189 @@
...
+        for (i = 0; buf[i]; i++) {
+            if (buf[i] == '\n') {
+                buf[i] = '\0';
+                break;
+            }
+        }
+
+        if (strstr (buf, ".__")) // skip metalabels
+            continue;
+
```
mrdudz, line 144: Please use C comments only.
Please use braces with all IF statements.

util/profile/sim65-prof.c
```diff
@@ -0,0 +1,189 @@
...
+        if (numlabels >= MAX_LABELS) {
+            printf ("Too many labels\n");
+            break;
+        }
+    }
+
+    fclose (f);
+
+    // prep
+    qsort (labels, numlabels, sizeof(struct label_t), byaddr);
```
mrdudz, line 161: Please use C comments only

util/profile/sim65-prof.c
```diff
@@ -0,0 +1,189 @@
...
+
+        struct label_t *cur = findlabel(i);
+        cur->count += samples[i];
+        total += samples[i];
+    }
+
+    qsort (labels, numlabels, sizeof(struct label_t), bycount);
+
+    // output
+    printf ("CPU cycles total: %u\n\n", total);
```
mrdudz, line 177: Please use C comments only

util/profile/sim65-prof.c
```diff
@@ -0,0 +1,189 @@
...
+    for (i = 0; i < numlabels; i++) {
+        if ((i >= 10 || labels[i].count < total / 100) && !verbose)
+            break;
+        if (!labels[i].count)
+            break;
+        printf ("%.1f%%\t%8u\t%s\n", labels[i].count * 100.0f / total,
```
mrdudz, line 183: Please always use braces with IF statements

mrdudz:
First, i really like this. A profiler seems to be a natural addition to this tool chain.

However, besides the trivial code style issues, there are some more things i'd like to see

- a complete "how to" example in the samples and documentation
- the profile file format should be documented (yes i realize its trivial right now :))
- the tool should probably be called "prof65" and live in the regular bin directory instead (which means it would also get installed). the "utils" directory is really only for tools that don't really belong to the tool chain.

util/profile/sim65-prof.c
polluks: or `return a->addr > b->addr;`

clbr:
> or `return a->addr > b->addr;`

That wouldn't work, the other direction (-1) may also be used by qsort.

polluks:
> or `return a->addr > b->addr;`
> That wouldn't work, the other direction (-1) may also be used by qsort.

Indeed, you need at least the first if.

clbr:
The current PR autobot failure is due to a bogus warning. Please disable that warning in your runner (stringopt-truncation).

error: output may be truncated... is completely useless here because we want it to be truncated. The following line ensures null-termination.

util/profile/sim65-prof.c
Compyx: That would be incorrect, it's not a boolean return value. Perhaps you meant `return a->addr - b->addr`?

src/prof65/main.c
```diff
@@ -0,0 +1,198 @@
...
+
+    if (optind + 2 > argc) {
+        usage (argv[0]);
+    }
+
+    profile = argv[optind];
+    label = argv[optind + 1];
+
+    /* samples */
+    FILE *f = fopen (profile, "rb");
```
spiro-trikaliotis, line 118: Defining a variable here is C99 style and won't compile with a C89/C90 compiler.
As we insist on C89'/90 style comments, should we insist on C89/C90 style var definitions?

mrdudz: > All declarations in a block must be at the beginning of that block.

yes, we do :)

mrdudz:
> The current PR autobot failure is due to a bogus warning. Please disable that warning in your runner (stringopt-truncation).
> 
> error: output may be truncated... is completely useless here because we want it to be truncated. The following line ensures null-termination.

I'd rather see the code fixed/changed so the warning disappears. Disabling warnings is almost never a good idea.

clbr:
With memcpy, you have to calculate the max len first, because reading
more than there is could segfault. Which is more code and ugly. But ok.

kugelfuhr:
While the general idea behind the tool is great, the code has several shortcomings.

1. The source differs in style from the other tools.

2. Common command line options supported by all other tools aren't accepted by prof65. Namely `--help`, `--verbose`, `--version`, `-V`.

3. The maximum number of labels is hard coded. This could be easily be changed by using a dynamic data structure like a Collection.

4. Despite labels being sorted by address, `findlabel()` uses a linear search. It would be easy to use a binary search here to drastically improve performance at almost no cost in complexity and code size.

5. There is absolutely no error checking on the format of the label file. If the format is wrong, prof65 will simply generate garbage instead of outputting an error message. I would suggest to use `sscanf()` to parse the input line to
have at least limited error checking. This would also simplify the code and remove the `strncpy` issue:

```C
        if (sscanf(buf, "al %6hx .%31s", &cur->addr, cur->name) != 2) {
            /* print error message */
```

6. Using a small header with a magic value and maybe a size for the profile file would improve resilience with just little overhead. Currently any file larger than 256KB can be used as a profile file and will simply generate garbage.

7. The constant "64 * 1024" is used in more than one place and could be derived from one source, so it's easier to change if - for example - the 65816 should be supported.

8. The functionality of the prof65 tool is so minimal that it might be worth to merge it into sim65.

9. Use of the vice label file is highly sub optimal. The debug file would be a much better choice since it identifies C functions, contains sizes of functions and allows profiling down to single source lines. Usage of this file would also help to improve support for debug information.

GorillaSapiens:
this looks awesome.  what's needed to push it over the finish line?

kugelfuhr:
I am not the one to decide this, but my suggestion would be not to merge it in its current form. I have listed a few things that should be changed in my eyes, but after this long time it became obvious to me that the last point mentioned above is the most important one.

Debug info generation has always been something that was there, but more or less invisible. Once in a while people using it complained when it got broke, but apart from that it didn't get much attention. Adding profiling support to sim65 is THE chance to make use of debug information inside the cc65 tool chain. Using it would enable profiling as a whole, based on functions (with much more detail than implemented in this PR), based on source code lines and much more. It would make bugs in debug info immediately visible, since it is actually used. It would also to make it more apparent that debug info is available and what can be done with it. Sim65 could serve as an example on how to use it.

On the other side, using the VICE label file would cement usage of a hack that was never thought for things like this. Once it is in, there is almost no chance to remove it later.

So yes, I like what was done in this PR. And I think it is important and useful enough to do it right. Even if a first implementation uses only a subset of the available debug information, it would open the door for more improvements in the future.

clbr:
As the original author, the debug info mentioned by kugelfuhr would not
have been useful to me. What I implemented covers my use cases fully.

It's of course fine if anyone wants to change it to do so.

mrdudz:
I very much agree with @kugelfuhr - the idea is good and this can perhaps be used as a foundation for further work... but indeed, it leaves some things to be desired. And i don't like the idea of letting such stopgap solution (using the label file) sneak in, as that usually means it will stay forever

GorillaSapiens:
so what i'm hearing is you want it to use the --dbgfile and not the -Ln VICE label file.  is that correct?

kugelfuhr:
> so what i'm hearing is you want it to use the --dbgfile and not the -Ln VICE label file. is that correct?

Personally, I think that this is a much better solution, yes. But there are other things. See https://github.com/cc65/cc65/pull/2049#issuecomment-1516115314 for more shortcomings.

If the profiling feature should go into a separate tool (as in this PR) instead of merging it into sim65 (as I would suggest), also a better format for the data exported by sim65 is necessary. There are several other interesting questions, a profiler could answer. For example how often a function is called. The current output doesn't allow this, it contains just cycle counts per address. And it has to be extended for the 65816 anyway.

mrdudz:
I wonder if it wouldnt make more sense to have sim65 spit out profiling data that is compatible with eg gprof - did something like that for a different target (the Playstation...) years ago, and it was almost trivial to do. BUT also depends on what kind of info you want to get out of it.

kugelfuhr:
Using gprof sounds like a good idea in first sight. But it would require users (especially Windows users) to get a gprof executable somehow. And gprof requires a call graph - something that is not easily available in the debug info or sim65.

PR #2058, opened 2023-04-19
Update lynxsprite.c

icepic:
Remove compiler warning about (signed) char wrapping around to -1 when set to 255.


mrdudz:
Thanks!

PR #2063, opened 2023-04-24
Add features: nested macros, newline sequence, etc

Movax12:
I've made some changes, mostly to the .macro definition routines to add some features and also found and fixed some bugs.

Quick summary:

Add a "newline" token represented by `.\` that behaves as a end of line separator unless defining a .define style macro, where it is captured as part of the macro definition. Allow for nested macro definitions. Some bug fixes and code changes related to macro definitions.

More detail:

Add features:

1) Add "newline" token sequence `.\` to allow for more than one statement per line. It will behave exactly as a true end of line, with one exception in (2).
2) Allow .define style macros to store the newline token when it is found in a definition. Definition is terminated by a true line ending.
3) Allow "nested" macros: Other macros can be defined while expanding a macro.
4) Add .insert token. Only valid inside a classic macro body. This allows defining the .macro token in a macro without triggering a nested macro block. It will also allow the name for a .define definition to be expanded if it the name is a .define style macro, otherwise .define style macros are disabled while reading the name given after a .define statement.

Bug fixes:

1) Control commands that affect a macro definition must be at the beginning of the line to be active: .local, .macro, .endmacro (already fixed) and also .insert
2) A .define or .undefine found at the beginning of a line when in a macro body will result in .define style macros being disabled while reading the next token (the macro name). Currently they are not, and a previously defined .define style macro matching the macro name will result in a macro expansion during the definition. (This does already work this way outside of macro definitions.)

Other info:

1) Any control commands that affect macro definition will be skipped when in a nested macro. (Enclosing macro would have to be expanded to define the nested macro and have its corresponding control commands active.)
2) The .define command will only result in define style macros being disabled and newline tokens being read as described when it is at the beginning of the line.
3) The .paramcount function will not be processed if found in a nested macro while expanding the enclosing macro. The .paramount function will always return the number of parameters for the .macro it is used in. If used in a .define style macro (and not inside a .macro definition) it will return the number of parameters for the .define macro (Matches current behavior.)
4) When expanding .define style macros, any newline tokens found will be converted into end of line separator tokens, unless defining a new .define.
5) Symbols defined in the macro using .local are accessible within nested macros, and can be used to change the definition of the nested macro. New symbols defined with .local in the nested macro can use duplicate names from the enclosing macro and will internally be assigned a new name.
6) Function MacSkipDef is removed, due to the inconsistent error message when .endmacro is not found. If there is an error, the MacDef function will still continue to read the macro definition and check for a missing .endmacro. Any other errors will result in the macro being set as as incomplete. 

Some simple examples:

```
.macro myMacro1 foo
    .local bar
    bar .set 0
    .if foo = 1
        bar .set 2
    .endif
    .macro myMacro2
        .out .sprintf("Bar is %d", bar)
    .endmacro
.endmacro

myMacro1 1
myMacro2        ; console output: Bar is 2

; build again:
myMacro1 0
myMacro2        ; console output: Bar is 0
```

This becomes two statements: `clc` (newline separator) `adc #$10`
```
.define add(op) clc .\ adc op
add #$10
```

Three statements:
```
lda #<foo .\ ldx #>foo .\ jsr load
```

I've been testing these changes for some time and am confident I have discovered and fixed any issues. There shouldn't be anything in these changes that breaks anyone's current code. If this request looks like it will be accepted, I will add documentation and tests.

### DIFF
--- a/src/ca65/macro.c
+++ b/src/ca65/macro.c
@@ -364,35 +364,22 @@ static void FreeMacExp (MacExp* E)
 
 
 
-static void MacSkipDef (unsigned Style)
-/* Skip a macro definition */
-{
-    if (Style == MAC_STYLE_CLASSIC) {
-        /* Skip tokens until we reach the final .endmacro */
-        while (CurTok.Tok != TOK_ENDMACRO && CurTok.Tok != TOK_EOF) {
-            NextTok ();
-        }
-        if (CurTok.Tok != TOK_EOF) {
-            SkipUntilSep ();
-        } else {
-            Error ("'.ENDMACRO' expected");
-        }
-    } else {
-        /* Skip until end of line */
-        SkipUntilSep ();
-    }
-}
-
-
-
 void MacDef (unsigned Style)
 /* Parse a macro definition */
 {
     Macro* M;
     TokNode* N;
     FilePos Pos;
+    int L;
     int HaveParams;
     int LastTokWasSep;
+    int NestedMacroCount;
+    unsigned int MacroErrorCount;
+
+    /* For classic macros, this indicates when in a nested macro. 
+    ** Only end the macro defintion with .endmacro when this is zero
+    */
+    NestedMacroCount = 0;
 
     /* For classic macros, remember if we are at the beginning of the line.
     ** If the macro name and parameters pass our checks then we will be on a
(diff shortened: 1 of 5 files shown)


mrdudz:
Oh and despite that - the patch should also contain updates to the documentation :)

Movax12:
OK, I will attempt to split this into smaller pieces...soon.

PR #2068, opened 2023-04-28
Remove MacSkipDef, capture all MacDef errors

Movax12:
Recent changes added a clearer error for a missing .endmacro when defining a macro: The error message will now supply the line and name of the macro.  Example: `main.s:70: Error: '.ENDMACRO' expected for macro 'mymacro'`
When _some_ errors are found in the MacDef routine, it will call MacSkipDef which searches for an .endmacro and, if not found, displays the error `'.ENDMACRO' expected` without any other info. The call to MacSkipDef is always followed by a `return`, leaving the macro's `Incomplete` flag set.
But, some errors in MacDef don't call the MacSkipDef routine, and as well, errors can be generated by ConsumeSep, ConsumeRParen, or an invalid control command. These errors result in the `Incomplete` flag being cleared, despite errors in the macro definition. 
These changes allow all errors generated when MacDef is running to properly set the `Incomplete` flag, as well as a consistent '.ENDMACRO' expected error.

### DIFF
--- a/src/ca65/macro.c
+++ b/src/ca65/macro.c
@@ -364,27 +364,6 @@ static void FreeMacExp (MacExp* E)
 
 
 
-static void MacSkipDef (unsigned Style)
-/* Skip a macro definition */
-{
-    if (Style == MAC_STYLE_CLASSIC) {
-        /* Skip tokens until we reach the final .endmacro */
-        while (CurTok.Tok != TOK_ENDMACRO && CurTok.Tok != TOK_EOF) {
-            NextTok ();
-        }
-        if (CurTok.Tok != TOK_EOF) {
-            SkipUntilSep ();
-        } else {
-            Error ("'.ENDMACRO' expected");
-        }
-    } else {
-        /* Skip until end of line */
-        SkipUntilSep ();
-    }
-}
-
-
-
 void MacDef (unsigned Style)
 /* Parse a macro definition */
 {
@@ -393,6 +372,7 @@ void MacDef (unsigned Style)
     FilePos Pos;
     int HaveParams;
     int LastTokWasSep;
+    unsigned MacroErrorCount;
 
     /* For classic macros, remember if we are at the beginning of the line.
     ** If the macro name and parameters pass our checks then we will be on a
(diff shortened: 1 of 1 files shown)


mrdudz:
Could you give an example of such a macro? I am not sure if/how we could create a test for this - but it would be nice to have, i guess

Movax12:
Currently, the MacDef routine leaves the Incomplete flag set if it finds an issue when checking for a valid identifier, a duplicate macro name, if names of instructions are allowed, or a missing .endmacro
The Incomplete flag is only checked by MacExpandStart when the macro is attempted to be expanded, and errors with "Cannot expand an incomplete macro".
I think this can only happen currently with a duplicate macro name. A missing .endmacro means EOF was reached. There is no way to call the macro with the others. But this example:
```
.macro myMacro
    .local foo!
.endmacro
myMacro
```
Results in:
 `Error: Unexpected trailing garbage characters`
With proposed changes it will:
`Error: Unexpected trailing garbage characters`
`Error: Cannot expand an incomplete macro`
Similar:
```
.macro myMacro one, one
    .local foo
.endmacro
myMacro
```
Results in:
 `Error: Duplicate symbol 'one'`
With proposed changes it will:
`Error: Duplicate symbol 'one'`
`Error: Cannot expand an incomplete macro`

This isn't a big problem, as the build will still fail, but better to be consistent?
So currently, any errors looking for an end of line separator, problems with the parameter names or .local names, will result in the Incomplete flag being cleared.

The only thing I am unsure of (after having some more time to think about it), is if the macro name is not a valid identifier, could that cause any serious issues with the hash function when `M = NewMacro (&CurTok.SVal, Style);` is executed? (line 432)
I also don't know how to make a test for this/don't think we can?

mrdudz:
> The only thing I am unsure of (after having some more time to think about it), is if the macro name is not a valid identifier, could that cause any serious issues with the hash function when M = NewMacro (&CurTok.SVal, Style); is executed? (line 432)

I wouldn't know either, unfortunately :(

mrdudz:
I'd really like to merge this - but i am reluctant because of no test. *shrug* What should i do? :)

mrdudz:
Yay, so #2777 also fixed this one:

```
.macro myMacro one, one
    .local foo
.endmacro
myMacro
```
now gives
```
bla.asm:1: Error: Duplicate macro parameter ‘one’
bla.asm:4: Error: Macro ‘myMacro’ contains errors and cannot be expanded
bla.asm:1: Note: Definition of macro ‘myMacro’ was here
```

PR #2069, opened 2023-04-30
Source listing in bootstrap.s is incorrect.

jefftranter:
The OSI C1P alternative boot file format works, but the code in the source listing does not match the ASCII-coded hex translation (which is actually used). This is confusing to anyone trying to maintain the code.

Also, the source code does not assemble when ASM is defined.

With these changes the source file should correctly match what is used at run time.

### DIFF
--- a/libsrc/osic1p/bootstrap.s
+++ b/libsrc/osic1p/bootstrap.s
@@ -34,7 +34,7 @@ ram_top         :=      __MAIN_START__ + __MAIN_SIZE__
 
 .ifdef ASM
 
-        .include        "osic1p.inc"
+        .include        "screen-c1p-24x24.s"
         .macpack        generic
 
 load            :=      $08             ; private variables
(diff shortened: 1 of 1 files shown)


libsrc/osic1p/bootstrap.s
```diff
@@ -45,20 +45,22 @@ GETCHAR         :=      $FFBF           ; gets one character from ACIA
...
-        ldx     #(<load_size) + 1
-        stx     count
-        ldx     #(>load_size) + 1
-        stx     count+1                 ; save size with each byte incremented separately
+        lda     #<load_size
+        eor     #$FF
+        sta     count
+        lda     #>load_size
+        eor     #$FF
+        sta     count+1
```
mrdudz, line 59: Why not eor with $ffff at compile time?

jefftranter:
I revised the pull request.

mrdudz:
Just now i actually realise what this PR is trying to do... would'nt it make more sense to fix the ASCII listing at the bottom to contain the original code (which is more optimal!) rather than restoring the old (less optimal) code by translating back the ascii stuff?

jefftranter:
I have to confirm, but my recollection was that the source version (not the ASCII listing) didn't work properly.

mrdudz:
> I have to confirm, but my recollection was that the source version (not the ASCII listing) didn't work properly.

Well, but you could surely test - and if it doesnt work, fix - it? It's not very complex code afterall :)

mrdudz:
looks good now, thanks!

Issue #2060, opened 2023-04-21
Pointer math bug/limits


davepl:
This code works perfectly until n becomes more than about  32768, which makes me suspect a pointer math bug.  As long as n is only 16 bits in value, it works fine.  Any thoughts, or is it a compiler bug?

void SETBIT(byte *p, unsigned long n)
{
   *(p + (n >> 3)) |=  (0b10000000 >> (n & 7));
}

mrdudz:
First, please give a full minimal example to demonstrate the problem, and tell exactly what happens, and what you expect to happen.

Also, are you sure this isnt the result of integer promotion?

This seems to be correct at least:
```
#include <stdio.h>

unsigned char *test(unsigned char *p, unsigned long n)
{
    return p + (n >> 3);
}

int main(void)
{
    printf("%p\n", test(0x1234, 0x12340000UL));
    printf("%p\n", test(0x1234, 0xba980000UL));
}

$ cl65 -t sim6502 -o bug2060.prg bug2060.c
$ sim65 bug2060.prg
0x9234
0x1234
```

davepl:
My bad for not including the link to the code example.  Here it is:

https://github.com/davepl/cc65/blob/KIMLife/samples/kim1/kimGFX.c￼
cc65/kimGFX.c at KIMLife · davepl/cc65
github.com

The line draw which calls SETPIXEL can’t draw more than 4K into the screen buffer.  Yet ClearScreen() works fine.

To work around it I rewrote it as follows, which works.  But it seems to me both should work, unless I’ve simply got a but in the GitHub version above.  But given the arbitrary nature of the cutoff, it seemed odd.

- Dave

￼

mrdudz:
Sorry, but this is far too complex to debug anything - hence "minimal example". Ideally strip it down to a single expression that does not work as intended.

kugelfuhr:
The problem is not in the snippet shown. It is in the surrounding code and caused by an error in the code generation of the compiler

The problematic code is

```C
void SETPIXEL(byte * p, int x, int y, byte b)
{
   if (b)
      SETBIT(p, (long) y * SCREEN_WIDTH + x);
   else
      CLRBIT(p, (long) y * SCREEN_WIDTH + x);
}
```

This is correctly translated by an older cc65 version using "-T -Oirs" as:

```asm
;
; SETBIT(p, (long) y * SCREEN_WIDTH + x);
;
        ldy     #$08
        jsr     pushwysp                ; push p
[...]

PR #2070, opened 2023-05-02
RHS primary integer promotion must happen after loading the primary, not before

bbbradsmith:
Probable fix for #2060

This also includes a test at test/val/bug2060.c

If the unit tests pass, I expect this is the correct resolution. However, I would feel more confident if someone else could take a look and evaluate whether the logic of this change is correct.

We should also figure out at what change this was first broken, which the test case can assist with.

### DIFF
--- a/src/cc65/expr.c
+++ b/src/cc65/expr.c
@@ -3090,9 +3090,10 @@ static void parseadd (ExprDesc* Expr, int DoArrayRef)
                 Expr->Type = Expr2.Type;
             } else if (!DoArrayRef && IsClassInt (lhst) && IsClassInt (rhst)) {
                 /* Integer addition */
-                flags = typeadjust (Expr, &Expr2, 0);
                 /* Load rhs into the primary */
                 LoadExpr (CF_NONE, &Expr2);
+                /* Adjust rhs primary if needed  */
+                flags = typeadjust (Expr, &Expr2, 0);
             } else {
                 /* OOPS */
                 AddDone = -1;
--- a/test/val/bug2060.c
+++ b/test/val/bug2060.c
@@ -1,0 +1,56 @@
+/* Test of bug: https://github.com/cc65/cc65/issues/2060 */
+
+#include <stdio.h>
+
+#define W 320
+
+unsigned long test1(unsigned char* p, unsigned long n)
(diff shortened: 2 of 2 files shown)


bbbradsmith:
I suspect this bug was introduced in f5972df as part of #1424

mrdudz:
Unfortunately @acqn seems to be missing since December, no idea what happened

I have not worked on those things, so i really have no idea - if in doubt, i'd try to construct more tests that rely on correct behaviour of this piece of the code

bbbradsmith:
As far as I can tell this is really the case that it affects. A non-constant right hand side that needs to be promoted from a 16-bit int to a 32-bit one.

I can't think of any other way to trigger that code, so I don't have any good ideas for other useful tests besides the one I already added.

mrdudz:
perhaps @kugelfuhr can have a look? :)

kugelfuhr:
> perhaps @kugelfuhr can have a look? :)

Well, I fear that I cannot be of much help in this case. If you look at f5972dfd084d4669dc4d936b19c52819efcdcbed, you will find the following comment just before the added line:

```C
                /* Integer addition. Note: Result is never constant.
                ** Problem here is that typeadjust does not know if the
                ** variable is an rvalue or lvalue, so if both operands
                ** are dereferenced constant numeric addresses, typeadjust
                ** thinks the operation works on constants. Removing
                ** CF_CONST here means handling the symptoms, however, the
                ** whole parser is such a mess that I fear to break anything
                ** when trying to apply another solution.
                */
```

(this comment was silently removed in a later change). In concordance with this comment I would say that the code is too complex for me to make any safe statements. The change by @bbbradsmith looks good, but I cannot say if it has any bad side effects. What makes me scratching my head is that the `LoadExpr()` line was added in f5972dfd084d4669dc4d936b19c52819efcdcbed without any hint why it was necessary at all.

My suggestion would be to simply merge the code so more people use it and possible problems can be found. The `parseadd()` function should also be marked as "needs refactoring". It has about 400 lines which is way too large to understand it easily.

mrdudz:
> My suggestion would be to simply merge the code

I should probably wait until that other thing(tm) is resolved =)

Issue #2022, opened 2023-03-06
cc65: Returning structs by value generates invalid code


kugelfuhr:
The following code:
        
```c
struct x { int a; char b; };
    
struct x f(void)
{
    struct x x = { 1, 2 };
    return x;   /* Returned in eax with high byte = junk */
}

int main(void)
{
    struct x x;
    x = f();    /* Overwrites x with 4 bytes */
    return 0;
} 
```

Looks like this after translation with "-Oirs -T":

```
; ---------------------------------------------------------------
; struct x __near__ f (void)
; ---------------------------------------------------------------

.segment        "CODE"

.proc   _f: near

.segment        "CODE"

;
; struct x x = { 1, 2 };
;
        jsr     decsp3
        ldy     #$02
L0002:  lda     M0001,y
        sta     (sp),y
        dey
        bpl     L0002
;
; return x;   /* Returned in eax with high byte = junk */
;
        jsr     ldeax0sp
;
; }
;
        jmp     incsp3
 
.segment        "RODATA"

M0001:
        .word   $0001
        .byte   $02
        
.endproc

; ---------------------------------------------------------------
; int __near__ main (void)
; ---------------------------------------------------------------

.segment        "CODE"

.proc   _main: near

.segment        "CODE"

;
; x = f();    /* Overwrites x with 4 bytes */
;
        jsr     decsp3
        jsr     _f
        jsr     steax0sp
;
; return 0;
;
        ldx     #$00
        txa
;
; }
;
        jmp     incsp3
 
.endproc
```

As you can see, the struct is 3 bytes in size but is handled by eax which causes one additional byte to be read/written. The write might cause erratic behavior or program crashes.

bbbradsmith:
According to the documentation [cc65 cannot return structs by value](https://github.com/cc65/cc65/blob/805e98a7aa6fbd2b419de5192163f7d47c1f3beb/doc/cc65.sgml#L809), but apparently this is not true?

Was this feature (including the noted bug) added at some point without documenting it? Is there a past version which rejects any struct return with an error?

mrdudz:
Unfortunately there was a time when feature additions were accepted without requiring to update documentation accordingly - i think you are correct in all points (however there are still certain restrictions regarding the size of the structs that work when passing into and returning from a function. not documented of course ;/)

bbbradsmith:
I think this capability may have been added in #1102

bbbradsmith:
[...]

PR #2072, opened 2023-05-02
Forbid 3-byte struct pass/return by value, document struct value pass/return

bbbradsmith:
#2022 indicated a serious error in the handling of a 3-byte struct as a return value.

I have simply disabled the 3-byte case for both pass and return. It will not create an error, the same way other unsupported struct sizes do.

No problem was yet identified with 3-byte struct _passing_, but given the nature of the discovered error I don't feel we have demonstrated that the passing case is safe either. If it is to be eventually supported, it will require a full review of the ramifications of passing/returning such a struct. For now I feel it is better to exclude both.

While investigating this I also documented the supported struct sizes for pass/return, since it was a completely undocumented feature presumably added in #1102 by @acqn.

Also added a simple test case for pass and return of structs by value in the 3 supported sizes.

### DIFF
--- a/doc/cc65.sgml
+++ b/doc/cc65.sgml
@@ -806,9 +806,8 @@ and the one defined by the ISO standard:
 
 <item>  The datatypes "float" and "double" are not available.
         <p>
-<item>  C Functions may not return structs (or unions), and structs may not
-        be passed as parameters by value. However, struct assignment *is*
-        possible.
+<item>  C Functions may return structs (or unions) by value, but only of
+        1, 2 or 4 byte sizes.
         <p>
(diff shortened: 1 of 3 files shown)


mrdudz:
Thanks! This is also a good example for an ideal PR - Fix, Documentation, Test :) Cheers!

mrdudz:
I have reverted this one - see #2079

bbbradsmith:
I protest that it's not a bug, and reverting has reintroduced the bug it was preventing. Commentary at #2079

mrdudz:
merged again, as discussed in #2079

Issue #2026, opened 2023-03-07
cc65: Compiler does no range check on integer literals


kugelfuhr:
An error message for the following code might be helpful but there is none.

```c
int a = 0xFFFFFFFFFFFFFFFFFFFFFFFFFFF;
int b = 10000000000000000000000000000;
```

bbbradsmith:
Added an error for this in PR #2073

kugelfuhr:
This issue can be closed. It was fixed in commit bf5b37a3b2b0a286e3cc14d1435f0ddbaa6cee4d.

mrdudz:
ok

Issue #2021, opened 2023-03-06
cc65: Using floats silently generates wrong code


kugelfuhr:
The following works without a warning or error and stores zero into p (instead of 3):

```c
void* p = (int)3.0;
```

bbbradsmith:
This is another case I was surprised, because I did not think cc65 supported float constants at all.

When looking into this, I made PR #2073 with several things that should improve the situation. This particular problem was a combination of the float parsing not working properly, and also the constant conversion assuming everything was an int and reading a 0 integer value from the float expression instead of attempting to convert the floating point value. I tried to add a brief not in the document that float constants are usable.

bbbradsmith:
Sorry if this is redundant, but it feels like the PR fixing this may have gotten lost: #2085 

Changes:
* Fixed the broken floating point parsing.
* Allow float constants to convert to int, with a warning if an implicit conversion would lose precision.
* Gives a descriptive error if floating point constant arithmetic is attempted (since it's not supported).
* Adds test cases to demonstrate that things work as expected.
* Documents the compiler's abilities in this matter.

The PR was waiting on a review vs. some in-progress comprehensive floating point support, but I feel the changes are minor and do not really conflict with that work. It's not really adding much, mostly just fixing up what was already there. The main addition was the conversion to int, which is currently allowed by the compiler anyway but silently produces garbage.

mrdudz:
#2085 is merged so this can be closed :)

PR #2073, opened 2023-05-03
Numerical constant errors and improvements

bbbradsmith:
This is a few improvements to do with numerical constants.

* Added an error for numerical constants that overflow the 32-bit internal limitation. See: #2026
* Added optional warning ```const-overflow``` for a numerical constant that overflows its storage, off by default. (I believe this is a default warning for clang, but "pedantic" for GCC.)
* Fix incomplete/broken parsing of floating point constants, also allows unary +/- for negative floating point.
* Allow floating point constants to be implicitly converted to integers, with a warning if precision is lost.
* Better error message when unsupported floating point arithmetic is attempted.
* Documenation and simple test of floating point constants.

The floating point stuff was addressing #2021 which revealed that at some point partial support for floating point constants was attempted. However, it was both incomplete and broken, and instead of producing any error it just silently turned them all into 0 valued integers. I've tried to complete the implementation of their parsing, and allow simple conversion to integers. The warning should avoid any surprise about the precision loss, and the lack of expression arithmetic support now has a more intuitive error. I also added a small test of the feature and added a note about it to the documentation.

### DIFF
--- a/doc/cc65.sgml
+++ b/doc/cc65.sgml
@@ -754,6 +754,8 @@ Here is a description of all the command line options:
         Warn about unused function parameters.
   <tag><tt/unused-var/</tag>
         Warn about unused variables.
+  <tag><tt/const-overflow/</tag>
+        Warn if numerical constant conversion implies overflow. (Disabled by default.)
   </descrip>
 
   The full list of available warning names can be retrieved by using the
--- a/doc/cc65.sgml
+++ b/doc/cc65.sgml
@@ -807,6 +807,9 @@ and the one defined by the ISO standard:
 <itemize>
 
 <item>  The datatypes "float" and "double" are not available.
+        Floating point constants may be used, though they will have to be
+        converted and stored into integer values.
+        Floating point arithmetic expressions are not supported.
         <p>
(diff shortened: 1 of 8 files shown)


bbbradsmith:
As a side note, if someone wanted to take floating point constant support a bit further, and allow arithmetic in their expressions, promoting int to float, etc. [datatype.c:ArithmeticConvert](https://github.com/cc65/cc65/blob/8f2927cf07d92966c9f61b3e0d65e2509b382d98/src/cc65/datatype.c#L660) appears to be the entry point for this.

mrdudz:
Would you please split this PR and put the floating point stuff into a seperate one? One of my side quests is actually fixing the float support, and that would perhaps make it a bit easier for me (see https://github.com/mrdudz/cc65/tree/fptest )

bbbradsmith:
As requested I have split and rebased this into a pair of pull requests: #2084 and #2085 

The individual commits were each a separate functional chunk, so even though both of these commits have multiple changes, looking at individual commits should show each isolated component change, if you need to investigate or be selective.

mrdudz:
thanks, closing this one then

Issue #194, opened 2015-08-06
Suppress "Suspicious address expression" for 65816 direct page


pinobatch:
On 65816, it is common to access several addresses in the same 256-byte page of bank $00 using direct page addressing mode. However, ca65 raises a "Range error" when using this with addresses not in the true zero page, and attempts to work around this error result in a "Suspicious address expression" warning. The docs mention neither this warning nor a way to suppress it.

```
.p816
VMDATA = $002118
.smart
.segment "CODE"
  sep #$20
  ; First set the direct page base
  pea linebuf & ~$FF
  pld
  ; Now try to add to direct page
  lda <linebuf      ; This is suspicious
  lda z:<linebuf    ; So is this
  lda z:linebuf     ; And this just makes a range error
  pea VMDATA & ~$FF
  pld
  sta <VMDATA       ; But this isn't
forever: jmp forever
.segment "BSS"      ; linker script puts this in $000200-$001FFF
linebuf: .res 32
```

greg-king5:
Uz put this in a comment in the source code:
> If the instruction has a one-byte operand, and immediate addressing is
> allowed but not used, check for an operand expression in the form
> `<label` or `>label`, where label is a far or absolute label. If found,
> emit a warning. That warning protects against a typo, where the '`#`'
> for the immediate operand is omitted.

You can avoid the warning by putting a "dummy" operation into the expression:

```gas
        .p816
VMDATA  := $002118
        .smart  on
.code
        sep     #%00100000
        ; First, set the direct-page base.
        pea     linebuf & ~$FF
        pld
        ; Now, use direct-page addressing.
        lda     z:<linebuf-0

        pea     VMDATA & ~$FF
        pld
        sta     <VMDATA+256     ; Note: operator precedence makes this a 16-bit address!
forever:
        jmp     forever
.bss                            ; linker script puts this in $000200-$001FFF
linebuf:
        .res    32
```

greg-king5:
@oliverschmidt:  This issue should get the "**question**" label.

pinobatch:
Is there a label for bugs in the manual?

oliverschmidt:
Nope - I'd say the label 'bug' with an explanation that it's about the manual should be fine - at least for now...

oliverschmidt:
If this issue is supposed to be a bug in the manual then it should be updated to explicitly state where the bug in the manual is located.

Otherwise I'd consider the question answered - and therefore close this issue.

greg-king5:
[...]

PR #2074, opened 2023-05-03
Explicit z: should suppress "Suspicious address expression" warning

bbbradsmith:
When using ```z:``` prefix to explicitly request zeropage, it is not suspicious to use ```<``` to truncate an address. (Especially useful for 65816, Hu6280, etc.)

Other cases of the warning are left intact, as it is legitimately trying to warn against a missing ```#```.

Addresses: #194

### DIFF
--- a/src/ca65/ea.h
+++ b/src/ca65/ea.h
@@ -43,6 +43,10 @@
 /*****************************************************************************/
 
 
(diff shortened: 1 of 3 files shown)


mrdudz:
thanks!

Issue #2015, opened 2023-03-06
cc65: Structs with duplicate fields do not cause errors


kugelfuhr:
cc65 happily compiles structs with duplicate fields:

```c
struct x {
    int a;
    int a;
} x;
    
void f(void)
{
    x.a = 3;
}
```

If one of the fields has its type changed to char (for example), it complains about "conflicting types, 'char' vs 'int'".

bbbradsmith:
Seemed very simple to fix. PR #2075

PR #2075, opened 2023-05-03
Error for struct/union with a duplicate member

bbbradsmith:
A struct or union can't have 2 members with an identical name. Fixes: #2015


mrdudz:
seems sane, thanks!

Issue #392, opened 2017-03-02
ca65 C-style macros respect reserved names as params


clbr:
```
.define debug(message) .out message
.define debug(x) .out x
```

The latter line fails with "Error: `)' expected", same if the param is named "y". There is no such limit with normal C macros, and the error message is confusing - ca65 should allow params be named anything, or at the very least print a better error.

raybellis:
This applies to `.macro` as well.  Despite there being examples such as:
```
.macro  ldaxy   a, x, y
```

in the documentation this syntax appears to be illegal and generates this error:
```
test.s(1): Error: Unexpected trailing garbage characters
```

If I replace the parameters with `aa, ax, ay` it works as documented.

bbbradsmith:
I've revised the broken example in the doc with #2076 and made a note that these shouldn't be used for parameter names.

Otherwise should this be considered a bug? I assumed it's intended that A, X, Y retain special meaning and can't really be used in macros like this.

Perhaps a better error could help.

mrdudz:
IMHO it's a bug - and the bug is, that the assembler produces the wrong error message (the error message should say "can not use reserved names as parameters" or something like that)

mrdudz:
closing this, errors are much better now

PR #2076, opened 2023-05-03
Don't use a,x,y in macro parameter example, document why not.

bbbradsmith:
The documentation has a non-working example macro due to the use of a, x, and y as parameters. Replacing the names so the code can work, and added a brief note to avoid those parameter names.

Partially addresses: #392


mrdudz:
thanks!

Issue #1894, opened 2022-11-01
ca65: Corrupted object files when compiling source files with Macintosh line endings (CR)


Gumball2415:
Macintosh style line endings (CR) in a source assembly file seem to produce incomplete/corrupt object files.
Here is a short demo program:
[hello_NES.zip](https://github.com/cc65/cc65/files/9912587/hello_NES.zip)

and here are the commands used to compile the program:
```
ca65 -g -o hello_NES.o hello_NES.asm
ld65 -v -t nes --dbgfile hello_NES.dbg hello_NES.o -o hello_NES.nes
```
I'm using cc65 version at commit 911ff40f6, built with MinGW 64-bit MSYS2.

When the source assembly file contains only CR line endings, this is the resulting object file:
[hello_NES_cr.zip](https://github.com/cc65/cc65/files/9912564/hello_NES_cr.zip)

And correspondingly, here is the resulting object file with only LF line endings:
[hello_NES_lf.zip](https://github.com/cc65/cc65/files/9912567/hello_NES_lf.zip)

bbbradsmith:
So, what happens in this case is the first line starts with ```;``` beginning a comment. The [scanner's comment loop](https://github.com/cc65/cc65/blob/805e98a7aa6fbd2b419de5192163f7d47c1f3beb/src/ca65/scanner.c#L1357) discards characters until '\n' or EOF. Because your file has no '\n' the comment will last until the end of the file, producing no errors and an empty object.

So, that's the problem. I don't have any quick suggestion about how to address it. I'm a little curious how these line endings came about... I haven't personally had to deal with this kind of line endings in a very long time, and I haven't spotted any of my current Mac tools doing it. Is there a place this still comes up in normal practice?

bbbradsmith:
As far as I can tell, ca65's scanner doesn't really look for anything but '\n' to mark a new line. It [uses "r" with fopen](https://github.com/cc65/cc65/blob/805e98a7aa6fbd2b419de5192163f7d47c1f3beb/src/ca65/scanner.c#L504) and reads the file [with fgetc](https://github.com/cc65/cc65/blob/805e98a7aa6fbd2b419de5192163f7d47c1f3beb/src/ca65/scanner.c#L386).

[...]

PR #2078, opened 2023-05-03
ca65 support for three line ending types: \r, \r\n, \n

bbbradsmith:
Created as a test to address #1894

This should accommodate all 3 line ending types transparently.

Not sure whether it is needed as a feature or not, but if it is desired, the change here seems somewhat minimal. I only implemented this as a test to see if it was viable.

Have not investigated whether cc65 would require a corresponding change. It's scanner is a bit different.

### DIFF
--- a/src/ca65/scanner.c
+++ b/src/ca65/scanner.c
@@ -112,6 +112,7 @@ struct CharSource {
     CharSource*                 Next;   /* Linked list of char sources */
     token_t                     Tok;    /* Last token */
     int                         C;      /* Last character */
+    int                         SkipN;  /* For '\r\n' line endings, skip '\n\ if next */
     const CharSourceFunctions*  Func;   /* Pointer to function table */
     union {
         InputFile               File;   /* File data */
@@ -325,6 +326,7 @@ static void UseCharSource (CharSource* S)
     Source      = S;
 
     /* Read the first character from the new file */
+    S->SkipN    = 0;
     S->Func->NextChar (S);
 
     /* Setup the next token so it will be skipped on the next call to
@@ -386,6 +388,10 @@ static void IFNextChar (CharSource* S)
         while (1) {
 
             int N = fgetc (S->V.File.F);
+            if (N == '\n' && S->SkipN)
+                N = fgetc (S->V.File.F);
+            S->SkipN = 0;
+
             if (N == EOF) {
                 /* End of file. Accept files without a newline at the end */
                 if (SB_NotEmpty (&S->V.File.Line)) {
@@ -401,8 +407,11 @@ static void IFNextChar (CharSource* S)
 
             /* Check for end of line */
             } else if (N == '\n') {
-
                 /* End of line */
+                break;
+            } else if (N == '\r') {
+                /* End of line, skip '\n' if it's the next character */
+                S->SkipN = 1;
                 break;
 
             /* Collect other stuff */


src/ca65/scanner.c
```diff
@@ -386,6 +388,10 @@ static void IFNextChar (CharSource* S)
         while (1) {
 
             int N = fgetc (S->V.File.F);
+            if (N == '\n' && S->SkipN)
+                N = fgetc (S->V.File.F);
```
mrdudz, line 392: please use braces

mrdudz:
I am a bit confused by the SkipN thing - wouldnt it make sense to just skip any amount of \n or \r after any \n or \r? Not seldomly you get a whack mixture of those (when someone edits a file on a different OS it was created with, with some stupid editor). Is there a case where any of those would NOT be a line ending?

bbbradsmith:
I don't think it's possible to account for all combinations of mixed lines but I believe \r\n\n should be treated as 2 lines, not 1. I think \r will be followed by at most one \n before the next line.

There are plenty of goofy theoretical mixed cases but skipping at most 1 \n after \r  is the most useful way I can think of dealing with it.
```
\r (old mac blank line)
\n (linux blank line) < this is now an \r\n pair indistinguishable from a windows line
\n (linux blank line)
\r\n (windows blank line)
\n (linux blank line) < don't skip this
\n (linux blank line) < don't skip this
```

The main case this PR is for was the issue case where I guess they have a file made by an old Mac editor, and maybe they're trying to build it on a new Mac or other platform? (OP has not yet commented on how the situation arose.)

A mixture of blink lines with old mac + linux endings would be ambiguous no matter what, but I also don't expect to ever see that case in practice.

A mixture of linux + windows line endings I've definitely seen many times, but this should be handled fine by this PR. However, that case was basically already handled by \r being treated as whitespace. The PR is really only needed to address the confusion of what happened with the old mac file.

You mentioned wanting to treat \n\r as a line ending? Is that a real world case I'm not aware of? I don't think we should skip \r after \n. The potential mixture of linux + windows line endings are a good reason not to do that.

mrdudz:
> I don't think it's possible to account for all combinations of mixed lines but I believe \r\n\n should be treated as 2 lines, not 1. I think \r will be followed by at most one \n before the next line.

Well, yes - in an attempt to be able to still properly determine the line number that makes sense. On the other hand, such file is simply broken (perhaps a warning could be generated?)

> You mentioned wanting to treat \n\r as a line ending? Is that a real world case I'm not aware of? I don't think we should skip \r after \n. 

I am really trying hard to remember where i needed this.... but i cant. I might remember wrong, obviously :) It might have been the result of editing a broken file in a broken editor, resulting in an even more broken file :=)

In any case, is it easily possible to produce a warning (once preferably) when the parser sees a mixture of line endings? If so, it would be great if you could add that.

bbbradsmith:
A warning like that wouldn't be too difficult to implement, but it doesn't seem worthwhile to me.

I was trying to make the changes to a core loop like this as minimal as I could, and I don't even really want to suggest merging this without clarification on how the user in the original issue came across the old mac line endings case.

I can't actually think of a way in which a file will be "broken" as a result of mixed endings. Mixed linux + windows files were fine before and after this PR, and the line numbering wasn't inconsistent. Old mac \r-only endings are broken before this PR, which is why I investigated this change, but that was broken without even considering mixed lines issues.

If you want to warn for it it, I think it would require tracking whether it's seen each of the 3 line ending types, and then emitting a warning when the file is closed if more than one was used. Maybe you'd also want to track the first line each one appeared on so you can point to it in the warning? I feel like the extra code isn't worth the limited utility.

Tools are pretty good at catching mixed lines these days. Programming text editors are often aggressive about mixed ending warnings. Github repositories automatically fix line endings by default. Etc. I don't think we really need ca65 to catch it for us.

mrdudz:
Ok, i agree its not worth it to spend a lot of effort into producing a proper warning for such an edge case. I'll merge this as is :)

Issue #1778, opened 2022-06-21
Macro was defined in different places?


jeffythedragonslayer:
The "Note: Macro was defined here" message reports different line numbers sometimes.  I think it should report the line that .macro is on.

With broken_macro.s:
```
.macro blah
blah2
blah2
blah2
.endmacro

blah
```

ca65 .\broken_macro.s output:

```
.\broken_macro.s:2: Note: Macro was defined here
.\broken_macro.s:7: Error: Symbol 'blah2' is already defined
.\broken_macro.s:3: Note: Macro was defined here
.\broken_macro.s:7: Error: ':' expected
.\broken_macro.s:3: Note: Macro was defined here
.\broken_macro.s:7: Error: Symbol '.size' is already defined
.\broken_macro.s:3: Note: Macro was defined here
.\broken_macro.s:7: Error: Symbol 'blah2' is already defined
.\broken_macro.s:4: Note: Macro was defined here
.\broken_macro.s:7: Error: ':' expected
.\broken_macro.s:4: Note: Macro was defined here
.\broken_macro.s:7: Error: Symbol '.size' is already defined
.\broken_macro.s:4: Note: Macro was defined here        
```

bbbradsmith:
When there's a macro error it reports the line where the macro was put to use (in this case line 7) and it also reports the definition line of the macro that generated the error (in this case lines 2 3 and 4 are all generating their own errors).

So, the given error locations are correct, even though it's confusing. (The macro system is pretty confusing overall.)

I agree the error could be improved, though I don't have any implementation suggestion offhand. I don't see a bug here, though, it's reporting all of the relevant error lines correctly.

mrdudz:
I actually agree with the OP - the "Macro was defined here" should always point to the line the macro was started (ie the line containing .macro), not some random line inside of the macro.

Perhaps we can change the wording of the error message to make more clear it is pointing to a line inside the macro? I don't have a good idea on what that would be right now though :)

bbbradsmith:
I don't agree. The line isn't random, it's the line that produced the error. That's really important. If you have a 100 line macro and the error is on line 50, you need to know about line 50, not line 1 where it says ".macro".

The wording isn't great, I agree with that much, but the actual line numbers chosen aren't random or wrong. They're the 2 relevant lines to the error:
* The line of the code that invoked the macro.
* The line of the macro source that produced the problem.

The line with ".macro" on it isn't an error. It's not the line that needs to be implicated.

[...]

PR #2081, opened 2023-05-03
ca65: Suppress '.size' error for multiply-defined symbols

bbbradsmith:
In #1778 I [made a comment](https://github.com/cc65/cc65/issues/1778#issuecomment-1533191532) that the multiply-defined symbol errors were also generating a secondary '.size' error which adds to the user's confusion.

This change skips generating the ```.size`` symbol for a symbol that has already received the multiply-defined error.

Previously a doubly-defined label would produce a confusing second error about '.size':
```
blah:
blah:
```
Error:
```
.\file.s:2: Error: Symbol 'blah' is already defined
.\file.s:2: Error: Symbol '.size' is already defined
```

This identifies the case as a symbol that was marked multiply-defined via the SF_MULTDEF, which is only applied in the case of an error, so I don't believe this change would hide any error that isn't already redundant.

### DIFF
--- a/src/ca65/main.c
+++ b/src/ca65/main.c
@@ -858,7 +858,11 @@ static void OneLine (void)
             /* The line has switched the segment */
             Size = 0;
         }
-        DefSizeOfSymbol (Sym, Size);
+        /* Suppress .size Symbol if this Symbol already has a multiply-defined error,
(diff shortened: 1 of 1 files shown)


mrdudz:
Thanks!

PR #2085, opened 2023-05-03
Numerical constant errors and improvements (float)

bbbradsmith:
A requested split rebase of https://github.com/cc65/cc65/pull/2073, including only the portions relating to float:

- Fix incomplete/broken parsing of floating point constants, also allows unary +/- for negative floating point.
- Allow floating point constants to be implicitly converted to integers, with a warning if precision is lost.
- Better error message when unsupported floating point arithmetic is attempted.
- Documenation and simple test of floating point constants.

The floating point stuff was addressing #2021 which revealed that at some point partial support for floating point constants was attempted. However, it was both incomplete and broken, and instead of producing any error it just silently turned them all into 0 valued integers. I've tried to complete the implementation of their parsing, and allow simple conversion to integers. The warning should avoid any surprise about the precision loss, and the lack of expression arithmetic support now has a more intuitive error. I also added a small test of the feature and added a note about it to the documentation.

### DIFF
--- a/doc/cc65.sgml
+++ b/doc/cc65.sgml
@@ -805,6 +805,9 @@ and the one defined by the ISO standard:
 <itemize>
 
 <item>  The datatypes "float" and "double" are not available.
+        Floating point constants may be used, though they will have to be
+        converted and stored into integer values.
+        Floating point arithmetic expressions are not supported.
         <p>
 <item>  C Functions may not return structs (or unions), and structs may not
         be passed as parameters by value. However, struct assignment *is*
--- a/src/cc65/datatype.c
+++ b/src/cc65/datatype.c
@@ -670,6 +670,10 @@ const Type* ArithmeticConvert (const Type* lhst, const Type* rhst)
     ** floating point types are not (yet) supported.
     ** The integral promotions are performed on both operands.
     */
+    if (IsClassFloat(lhst) || IsClassFloat(rhst)) {
+	    Error ("Floating point arithmetic not supported.");
+		return type_long;
+    }
     lhst = IntPromotion (lhst);
     rhst = IntPromotion (rhst);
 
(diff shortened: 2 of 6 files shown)


bbbradsmith:
Linking to [my comment](https://github.com/cc65/cc65/pull/2073#issuecomment-1532451440) about potential floating point arithmetic support for future reference.

mrdudz:
I'll have to work on this one later, and see how to manage merging this into the main branch while still keeping my WIP floating point support somewhat working.

bbbradsmith:
I looked over [mrdudz:cc65:fptest](https://github.com/cc65/cc65/compare/master...mrdudz:cc65:fptest) and I don't think this really conflicts with it in any significant way. I think it would be beneficial to merge this in the short term because it fixes some serious silent errors. Currently, using floating points in code compiles without error but produces completely bogus values (not even rounded ones).

A summary of my changes:
1. doc/cc65.sgml - Note about how much floating point stuff is and isn't supported.
2. datatype.c - trying to do FP vs FP arithmetic in a constant expression emits an error telling you that it's unsupported.
3. expr.c - allows float to convert to int in constant expressions
4. scanner.c - fixed the existing floating point parser, which produced very incorrect values.
5. typeconv.c - handles float to int conversion with a warning if the rounding changes the value (unless it's an explicit cast)
6. test/val/float-const-convert.c - a test program demonstrating these changes work as intended

So, this change isn't really adding a whole lot. There is a new ability to cast float to int in constant expressions, which is two short cases in expr.c. Otherwise, it's adding errors/warnings to guard against things it can't do, and fixes the broken floating point parsing.

mrdudz:
sorry, its just my limited time budget :) i am merging it now, and will try to catch up in my branch

Issue #2086, opened 2023-05-04
Evaluate and restore ability to pass and return 3-byte structs by value


bbbradsmith:
#2072 disabled 3-byte structs passed or returned by value because of a discovered bug #2022. This could potentially be re-enabled, but the code path for the 3-byte case is in need of a review.

The disabled implementation was allowing 3-byte structs to be passed like a 4-byte value, but this had a flaw that could write 4 bytes instead of 3 to the calling stack as it returns, corrupting memory. It's possible the corresponding pass-by-value could have a similar problem and it must be carefully reviewed.

A test has been created [test/val/struct-by-value.c](https://github.com/cc65/cc65/blob/master/test/val/struct-by-value.c) which has a commented out 3-byte case, which should be un-commented if this feature is restored. (Also don't forget to update [the documentation](https://github.com/cc65/cc65/blob/master/doc/cc65.sgml).)

.

#2079 covers the broader topic of allowing larger structs to be passed this way. Apparently larger structs were supported by value as late as 2.19 but we have since lost the ability. (Of course it was never documented that we had this.)

If this is re-introduced, perhaps the same arbitrary-sized pass/return could be used for the 3-byte case as well.

mrdudz:
BTW, you can place tests that do NOT work YET in test/todo - that way not only do we have a collection of things that need to be fixed in the sourcetree, but also a PR fixing it can simply move the test from test/todo to test/val

bbbradsmith:
The 3-byte struct passing is a compiler error, so it wouldn't be able to run the test for todo.

mrdudz:
Yes ok. NOW it is :)

mrdudz:
the (very few) cases that do not compile, but should, are in misc right now

bbbradsmith:
Okay #2088 should give the requested test I think.

acqn:
> The disabled implementation was allowing 3-byte structs to be passed like a 4-byte value, but this had a flaw that could write 4 bytes instead of 3 to the calling stack as it returns, corrupting memory. It's possible the corresponding pass-by-value could have a similar problem and it must be carefully reviewed.

The quickest way to reenable it is to pass them like a 4-byte value on the caller side (as before) with the parameter size/offset error on the callee side fixed. This way we might save some code size but at a cost of one byte of stack space for each of such parameters.

Or we may pass them 3-byte-precisely (copying the bytes one by one, using `memcpy` or some brand new 3-byte R/W runtime routines). 

Some design decision must be made.

[...]

PR #2088, opened 2023-05-04
guard test to ensure 3-byte struct isn't re-enabled without evaluation by accident

bbbradsmith:
An anti-test for this was requested in #2086
This ensures that 3-byte struct by value does not compile. When it breaks, there are notes in the test about what should be verified correct before allowing 3-byte struct by value.


mrdudz:
thanks!

Issue #2065, opened 2023-04-25
Documentation update availability?


bbbradsmith:
How often does [the website documentation](https://cc65.github.io/doc/) get updated?

It seems like [the doc repository](https://github.com/cc65/doc) was getting regularly updated, every month or two, but it looks like that stopped completely in 2021 and never resumed.

I've seen that the SGML does get built by the continuous integration tests, so we're validating that new document changes still work, but it seems that the resulting HTML documentation is not available anywhere publicly. One would have to build it for themselves, and unfortunately the relevant tools seem to be unavailable on Windows. I've only been able to do this on linux. (Haven't tried WSL, maybe it would be easier with that.)

I feel like a few years of changes are kind of unknown because the only publicly available document is so out of date, e.g. I was pinged today in #2064 about a newer feature addition that I did make sure to document, but there seems to be no normal way for a user to discover that document.

Ideally I feel it would be best to include the HTML documentation with releases and snapshot builds, so that one can always find documentation that matches the version they are using.

Sort of related, I was wondering if the reason it hasn't been updated in 2 years is that there hasn't been a release tagged in even longer... but it does seem that it was continually being updated unrelated to (and beyond) the releases until it stopped, so it doesn't look like it was the reason at all.

Maybe it is possible to automate this, update them once a month with a script?

mrdudz:
It should indeed get updated automatically from the github actions - i currently dont have the time to find out how exactly that would work however. So if someone knows how, go ahead, make a PR :)

mrdudz:
see discussion at https://github.com/cc65/doc/pull/1

The commands in the travis script oliver linked can probably be modified to work in the github action

bbbradsmith:
Well the github action already seems to have that: [snapshot-on-push-master.yml](https://github.com/cc65/cc65/blob/b743b1a8c270e68c8635711bcbab66a9ddb4f25c/.github/workflows/snapshot-on-push-master.yml#L105_

In June 2022 greg-king5 [commented out the push](https://github.com/cc65/cc65/commit/16d259d2b48a96f0521219d5ad35f5d25325e828), though it was previously an "-n" push so it would have been a dry run with no actual push. It seems like he was trying to get it to work but gave up.

The build logs look like it's successfully creating commits for new doc files, it's just never pushing anything.

So, my guess is that the issue was maybe permissions for the push or something? Unfortunately the logs for the runs back then have long since expired.

[...]

PR #2089, opened 2023-05-05
Workflow docs artifact, windows test, sim65 fix

bbbradsmith:
I was trying to investigate some workflow possibilities, but it ended up combining with a few related things.

Changes:
* Added a docs HTML artifact to pull requests, so that someone submitting document changes can use that to preview them.
* Upload a snapshot docs HTML zip to sourceforge alongside the snapshot build.
* Remove ".zip" from artifact filenames, because that's automatically added. (We had .zip.zip files.)
* Added "libtest" target to makefiles to build only the platform libraries needed to run "test".
* Fixed test/todo using testwrk/val for its temporary files by mistake.
* Added test/standard to test makefile, seemed like it was left out by mistake.
* Restrict sim65 to 32-bit values for cycles so that it's the same on both platforms.
* Fixed sim65 crash caused by test/val/constexpr.c using close(-1) in a weird way. MSVC's close(-1) will terminate with a fatal error, rather than return -1.
* Added workflow for windows test, but manually triggered.

So, the initial goal was to generate the HTML docs as an artifact for PR reviews, and also see if the windows version could have its test set run by the build. The docs part worked well. The windows test... was a little weird.

So, it's not hard to add the windows test to the PR and snapshot workflows, but it was taking about 25 minutes on the build machines, seems about 5x slower than the Linux tests. I'm not sure what the cause of this speed disparity is, but it's a little bit disturbing. I kinda suspect the performance difference is to do with "make", and not really a problem with the cc65 windows binaries, but I don't have much information about why at this point.

So, I left comments about that in the workflows, but I don't think it should be enabled, because we probably don't want to be waiting an extra half hour for the CI report every time. Instead I created a separate workflow that can be manually triggered whenever anyone wants to investigate.

The Windows version test was almost fine. The only problem that came up was test/val/constexpr.c, which ended up causing sim65 to crash. After accommodating this with a fix to sim65, the Windows build does manage to run the test set without error.

I also added a "libtest" target to the makefiles which only builds the 3 platform libraries needed to run the test. I also noticed that test/standard wasn't ever added to the test makefile, so I hooked that up too. (Luckily it didn't cause any new errors to appear.)

.

I did my best to test the workflow changes as thoroughly as I could on my own fork, so I don't expect them to fail here, though obviously I could not test the stuff involving secret credentials.

### DIFF
--- a/.github/workflows/build-on-pull-request.yml
+++ b/.github/workflows/build-on-pull-request.yml
@@ -54,7 +54,7 @@ jobs:
           make -j2 bin USER_CFLAGS=-Werror CROSS_COMPILE=x86_64-w64-mingw32-
 
   build_windows:
-    name: Build (Windows)
+    name: Build and Test (Windows)
     runs-on: windows-latest
 
     steps:
@@ -72,3 +72,8 @@ jobs:
 
       - name: Build app (release)
         run: msbuild src\cc65.sln -t:rebuild -property:Configuration=Release
+
+      - name: Build the platform libraries.
+        run: make -j2 lib QUIET=1
+      - name: Run the regression tests.
+        run: make test QUIET=1
--- a/.github/workflows/build-on-pull-request.yml
+++ b/.github/workflows/build-on-pull-request.yml
@@ -43,6 +43,11 @@ jobs:
       - name: Build the document files.
         shell: bash
         run: make -j2 doc
+      - name: Upload a documents snapshot.
+        uses: actions/upload-artifact@v3
+        with:
+          name: docs
+          path: ./html
       - name: Build 64-bit Windows versions of the tools.
         run: |
           make -C src clean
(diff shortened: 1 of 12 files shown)


mrdudz:
Hooray for the effort - but... would you split this into one PR per issue? I see at least 3 different things addressed here, which makes it kinda hard to manage.

bbbradsmith:
Hrm, ironically I was trying to make it _more_ manageable by grouping the two workflow goals into only one PR. From [your last comment](https://github.com/cc65/cc65/issues/2065#issuecomment-1535376785) about the docs workflow I was trying to avoid making multiple separate workflow PR that have to be merged...

Anyway, there were two primary goals for workflow, the docs artifact, and a viable windows test.

The windows test is dependent on the sim65 fix, so I couldn't really separate it from that. There were a couple of really small makefile tweaks etc. that are technically independent but didn't seem like they should need their own PR. I tried to separate each individual minor element of this into its own commit to make them easier to review, but I expected making many PRs out of it would have been unhelpful.

But... that's okay. I can take this part and put it back together into several PRs if that is what is needed. Will do.

mrdudz:
Well, at least the sim65 fix can be seperate, and the test makefile fixes. and then the workflow changes should not depend on each other.

PR #2090, opened 2023-05-05
add test/standard to test makefile

bbbradsmith:
Seems omitted by accident? The test passes, but "make test" had not been verifying it because of this.

Split from #2089


mrdudz:
Indeed, i must have simply forgotten to add it. weird :)

PR #2091, opened 2023-05-05
test/todo makefile uses testwrk/val by mistake

bbbradsmith:
These tests were using the wrong working files folder.

Split from #2089


mrdudz:
Thanks!

PR #2092, opened 2023-05-05
Windows test workflow, sim65 fixes for Windows differences

bbbradsmith:
This creates a manual-dispatch Windows test workflow that we can trigger as needed to help verify the integrity of that build. It was not added to the PR or snapshot workflows because it currently takes about 25 minutes on the github build machines.

The Windows build actually did pass the regression tests, but it exposed 2 issues with sim65. The first is the issue of "long" being 64-bit on the Linux builds but 32-bit on Windows. Some of the tests were using 5000000000 as their requested cycle count. I put checks into sim65 to disallow this so that the two platforms are not allowed to behave differently on this input. The second issue is that [test/val/constexpr.c](https://github.com/cc65/cc65/blob/a7676bdff5f07ee940e6054373d989f8737dd5e3/test/val/constexpr.c) is using ```close(-1)``` for an expected ```-1``` return value. On MSVC ```close(-1)``` will terminate the program instead of returning any value.

Also added a "libtest" makefile target that builds only the 3 platform libraries needed for "make test". This allows a faster setup for when we only wants to run the tests.

Summary:
* Restrict sim65 to 32-bit values for cycles so that it's the same on both platforms.
* Fixed sim65 crash caused by test/val/constexpr.c using close(-1) in a weird way. MSVC's close(-1) will terminate with a fatal error, rather than return -1.
* Suppress uninitialized variable warning in sim65.
* Added "libtest" target to makefiles to build only the platform libraries needed to run "test".
* Added workflow for windows test, but manually triggered because it's slow.

Split from #2089

### DIFF
--- a/.github/workflows/build-on-pull-request.yml
+++ b/.github/workflows/build-on-pull-request.yml
@@ -67,3 +67,10 @@ jobs:
 
       - name: Build app (release)
         run: msbuild src\cc65.sln -t:rebuild -property:Configuration=Release
+
+      # Disabled because this is 5x slower than the corresponding Linux test.
+      # windows-test-manual.yml provides a manually dispatched alternative.
+      #- name: Build the test libraries.
+      #  run: make -j2 libtest QUIET=1
+      #- name: Run the regression tests.
+      #  run: make -j test QUIET=1
--- a/.github/workflows/snapshot-on-push-master.yml
+++ b/.github/workflows/snapshot-on-push-master.yml
@@ -29,6 +29,13 @@ jobs:
       - name: Build app (release)
         run: msbuild src\cc65.sln -t:rebuild -property:Configuration=Release
 
+      # Disabled because this is 5x slower than the corresponding Linux test.
+      # windows-test-manual.yml provides a manually dispatched alternative.
+      #- name: Build the test libraries.
+      #  run: make -j2 libtest QUIET=1
+      #- name: Run the regression tests.
+      #  run: make -j test QUIET=1
+
   build_linux:
     name: Build, Test, and Snapshot (Linux)
     if: github.repository == 'cc65/cc65'
--- a/.github/workflows/windows-test-manual.yml
+++ b/.github/workflows/windows-test-manual.yml
@@ -26,18 +26,18 @@ jobs:
       - name: Build app (MSVC release)
         run: msbuild src\cc65.sln -t:rebuild -property:Configuration=Release
 
-      - name: Build utils (make)
+      - name: Build utils (MinGW)
         shell: cmd
         run: make -j2 util
 
-      - name: Build the platform libraries
+      - name: Build the platform libraries (make lib)
         shell: cmd
         run: make -j2 lib QUIET=1
 
-      - name: Run the regression tests
+      - name: Run the regression tests (make test)
         shell: cmd
         run: make test QUIET=1
 
-      - name: Test that the samples can be built
+      - name: Test that the samples can be built (make samples)
         shell: cmd
         run: make -j2 samples
--- a/.github/workflows/windows-test-manual.yml
+++ b/.github/workflows/windows-test-manual.yml
@@ -6,7 +6,7 @@ on:
 
 jobs:
   build_windows:
-    name: Build, Test (Windows)
+    name: Build, Test (Windows MSVC)
     runs-on: windows-latest
 
     steps:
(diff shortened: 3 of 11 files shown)


bbbradsmith:
The changes to sim65, and the addition of make libtest are both useful changes by themself, should the Windows test workflow be rejected.

I kept them together in this PR because they are a prerequisite for the Windows test to pass, and vice versa the Windows test also demonstrates that they work.

mrdudz:
This somehow still does way too much at once (sorry). Basically all the points you list should be seperate...

In any case:

- The sim65 "fix" is too strict. It should check the requested number of cycles against ULONG_MAX, and then produce an error if its more than that. Its fine to support more when its more than 32bit. If it errors out in case the requested number is too big, that is good enough to find and fix the problem.

- The makefiles should probably contain a comment that say why the number of max cycles is what it is.

- I'm not sure the extra "libtest" target makes sense, building the libs is fast enough even on windows :)

- I'm not sure if the windows tests should'nt just run in the normal workflow anyway. 20-30 minutes isnt really that much (with VICE it takes that long just to complete the build... and its not even a terribly large project).

mrdudz:
Oh and i forgot: The windows tests would make a lot more sense of they'd totally run on cmd.exe (i failed on doing this - it works locally but not with the github runner).

bbbradsmith:
```make lib``` takes _many_ minutes for me on windows locally. I've been kinda shocked to see how fast the github runners seem to do it. So, ```make libtest``` has already been very beneficial for me in the past day of running a lot of tests.

Though perhaps it's unnecessary for the manual build workflow to use it. It could maybe build all libs, utils, and samples like the Linux test. I haven't tested it that far yet... 30 minutes is a long wait for the build machine, and it's hours locally.

**Comment by bbbradsmith** [comment]
Okay, I removed the 32-bit limitation from sim65, and added a comment above the 3 values that had to be reduced.

I do agree that 64-bit cycle counts could potentially be useful, and there is no disadvantage for the single-platform user to allow it to go as high as it can.

My rationale for the restriction is that this had unknowingly been giving the wrong cycle counts on the Windows build, and I wanted to prevent that from happening again. However, as long as we can run the Windows test now and then, we'll be able to catch it with just the strtoul overflow check.

**Comment by bbbradsmith** [comment]
Expanded the windows manual dispatch to use the full set of tests equivalent to the master snapshot (util, libs, samples, test). Also set it to use cmd shell.

Not mergeable yet, as I need to test it on my fork first (can't run it in this PR).

**Comment by bbbradsmith** [comment]
```make -j2 samples``` contains a surprise! Somehow my local cmd.exe make attempt built samples, but [that test](https://github.com/bbbradsmith/cc65/actions/runs/4897982409) had a surprise in store at _geos_.

```
(earlier successes omitted)
...
../bin/ld65   -o sieve -t c64 -m sieve.map sieve.o c64.lib
../bin/ld65   -o terminal -t c64 -m terminal.map terminal.o c64.lib
../bin/ld65  -t c64 -o tinyshell tinyshell.o getsp.o c64.lib
../bin/ld65   -o tgidemo -t c64 -m tgidemo.map tgidemo.o c64.lib
../../bin/cl65 -t c64 -o hello hello.c text.s
../../bin/sp65 -r logo.pcx -c geos-bitmap -w bitmap.c,ident=bitmap
../../bin/cl65 -t geos-cbm -O -o filesel.cvt -m filesel.map fileselres.grc filesel.c
../../bin/cl65 -t geos-cbm -O -o geosver.cvt -m geosver.map geosverres.grc geosver.c
grc65: Unknown section 'xample'
grc65: Unknown section 'xample'
mingw32-make[2]: *** [makefile:114: geosver.cvt] Error 1
mingw32-make[2]: *** Waiting for unfinished jobs....
mingw32-make[2]: *** [makefile:108: filesel.cvt] Error 1
mingw32-make[1]: *** [makefile:378: samples] Error 2
mingw32-make: *** [makefile:31: samples] Error 2
```

bbbradsmith:
Failing samples turned out to be a serious bug in grc65. If was using ```fseek(F, -1, SEEK_CUR)``` to rewind one byte, but the file was a text file. This gave unreliable results, but not on all platforms. It was also missing a buffer overflow check.

bbbradsmith:
Okay, with that fix to grc65, samples is now building fine. I am expecting the windows manual build to come through green. Active test here: [windows-test-manual.yml action](https://github.com/bbbradsmith/cc65/actions/runs/4898893536)

bbbradsmith:
Now that the dependent issues were successfully solved and verified, I have split this PR as was requested:
* #2098 
* #2099 
* #2100 
* #2101

PR #2093, opened 2023-05-05
Workflow docs artifact and snapshot

bbbradsmith:
Summary:
* Added a docs HTML artifact to pull requests, so that someone submitting document changes can use that to preview them.
* Upload a snapshot docs HTML zip to sourceforge alongside the snapshot build.
* Remove ".zip" from artifact filenames, because that's automatically added. (We had .zip.zip files.)

Split from #2089


mrdudz:
now i wonder what https://github.com/cc65/doc/archive/refs/heads/master.zip is and what produces it (this is linked from the webpage) - isnt that the same? *shrug*

bbbradsmith:
That's just a direct link to the "source code" zip. At the top of any github repository page, there's a green button that says "code". If you click it, the "download zip" link points to that. That's how I found that link.

[When I suggested using it](https://github.com/cc65/cc65.github.io/pull/2), I did so because it was a thing that conveniently already existed. I hadn't considered the higher effort problem of fixing the build machines to generate it in lock-step with the snapshot.

mrdudz:
Actually what i'd prefer to pushing the file to sourceforge, would be pushing it to the ghpages repo instead (where the rest of the website is hosted). The same should probably happen with the snapshot - The idea is getting rid of sourceforge completely.

bbbradsmith:
I don't mind that as a goal, but is that something you are asking me to add to this PR? I feel like that's an orthogonal goal that can be managed separately.

This PR is really just about making the generated docs available to PR checks as artifacts + preserving the generated doc at time of build alongside the snapshot. I think where we happen to store the snapshot can be easily changed when we're ready to do so.

mrdudz:
Mh yeah ok. Would you fix the conflict though? :)

bbbradsmith:
Resolved.

mrdudz:
and merged - thanks!

PR #2094, opened 2023-05-05
More specific git commit error suppression in doc upload

bbbradsmith:
The bash if solution to avoid the git commit error when there are no changes has the potential to suppress other git commit errors.

The diff-index approach targets only the one error we want to eliminate (error because of no changes to commit).

As mentioned in #2065 [here](https://github.com/cc65/cc65/issues/2065#issuecomment-1535364766)

### DIFF
--- a/.github/workflows/snapshot-on-push-master.yml
+++ b/.github/workflows/snapshot-on-push-master.yml
@@ -115,9 +115,8 @@ jobs:
           git config push.default simple
           git add -A
           # prevent failure when there is nothing to commit
-          if git commit -m "Updated from https://github.com/cc65/cc65/commit/${GITHUB_SHA}" ; then
-            git push
-          fi
+          git diff-index --quiet HEAD || git commit -m "Updated from https://github.com/cc65/cc65/commit/${GITHUB_SHA}"
+          git push
 
     # enter secrets under "repository secrets"
       - name: Upload snapshot to sourceforge


mrdudz:
Did you test both cases? ie when docs changed and when they did not? when i did that, it would never update anything :)

bbbradsmith:
I believe that I tested the equivalent on my test repositories yesterday before I made [this comment](https://github.com/cc65/cc65/issues/2065#issuecomment-1535364766) about it. I also believe that I tested the cases of changes, no changes, and missing credentials for the push.

However, I suppose I was hasty and deleted my two test repositories already, think I wouldn't need them again. Unfortunately that means I can no longer check the test runs to verify whether or not I actually did test this.

I can probably recreate the test, it will just take a bit of time...

bbbradsmith:
I found some local files I hadn't deleted that proved I had done he test, but anyway I recreated it.

This is the workflow: [pushtest.yml](https://github.com/bbbradsmith/pushtest1/blob/main/.github/workflows/pushtest.yml)

It has two jobs, one that will always succeed, whether or not there are changes to push to the second repository. The other job will always fail at the push step, because it doesn't have the personal access token.

Demonstration:
* [a push with changes to commit will push and pass](https://github.com/bbbradsmith/pushtest1/actions/runs/4899459623)
* [a push with no changes to commit will pass](https://github.com/bbbradsmith/pushtest1/actions/runs/4899454341)

Target repository showing the successful pushes:
* [bbbradsmith/pushtest2](https://github.com/bbbradsmith/pushtest2/commits/main)

mrdudz:
ok, let's try :) Thanks!

PR #2096, opened 2023-05-05
Bugfix for the .ISMNEMONIC, .ISMNEM builtin function

Movax12:
This fixes a bug where attempting to match the first instruction for the current CPU table with `.ismnemonic` does not return a match. This is usually 'ADC'.
Example of bug:
```
.if .ismnemonic(adc) ; doesn't match
 adc #$01 ; never output
.endif
```


mrdudz:
It would be nice to have a test for this in the PR as well

Movax12:
Not sure if that is the best way to conduct the test - but it should do the job!

mrdudz:
> Not sure if that is the best way to conduct the test - but it should do the job!

In tests basically anything goes, as long as they test what they are supposed to test - so thanks for adding one!

merging...

PR #2097, opened 2023-05-05
Add support for 48x12 video mode on Challenger 1P

jefftranter:
Add support for 48x12 video mode on Challenger 1P.
Tested on real C1P hardware.


mrdudz:
I'll have to trust you on it :) thanks!

PR #2098, opened 2023-05-06
grc65 fix flawed text parsing

bbbradsmith:
Was using fseek(F,-1,SEEK_CUR) which is invalid for text files, behaviour unreliable across platforms. Added check for internal buffer overflow.

Discovered while working to support #2101 I found that [make samples failed](https://github.com/cc65/cc65/pull/2092#issuecomment-1536868347) with grc65 reporting a strange fragment of a word that should have been an ignored comment. The seek was skipping back several bytes and not just one character.

Split from #2092

### DIFF
--- a/src/grc65/main.c
+++ b/src/grc65/main.c
@@ -851,10 +851,9 @@ static char *filterInput (FILE *F, char *tbl)
     int a, prevchar = -1, i = 0, bracket = 0, quote = 1;
 
     a = getc(F);
-    while (1)
-    {
+    while (1) {
         if (i >= BLOODY_BIG_BUFFER) {
-            AbEnd ("File too large for internal parsing buffer (%d bytes).",BLOODY_BIG_BUFFER);
+            AbEnd ("File too large for internal parsing buffer (%d bytes)",BLOODY_BIG_BUFFER);
         }
         if ((a == '\n') || (a == '\015')) a = ' ';
         if (a == ',' && quote) a = ' ';
--- a/src/grc65/main.c
+++ b/src/grc65/main.c
@@ -855,12 +855,20 @@ static char *filterInput (FILE *F, char *tbl)
         if (i >= BLOODY_BIG_BUFFER) {
             AbEnd ("File too large for internal parsing buffer (%d bytes)",BLOODY_BIG_BUFFER);
         }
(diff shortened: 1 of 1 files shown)


src/grc65/main.c
```diff
@@ -850,8 +850,12 @@ static char *filterInput (FILE *F, char *tbl)
     /* loads file into buffer filtering it out */
     int a, prevchar = -1, i = 0, bracket = 0, quote = 1;
 
-    for (;;) {
-        a = getc(F);
+    a = getc(F);
+    while (1)
+    {
```
mrdudz, line 855: please put the opening brace after the while()

src/grc65/main.c
```diff
@@ -850,8 +850,12 @@ static char *filterInput (FILE *F, char *tbl)
...
+        if (i >= BLOODY_BIG_BUFFER) {
+            AbEnd ("File too large for internal parsing buffer (%d bytes).",BLOODY_BIG_BUFFER);
+        }
         if ((a == '\n') || (a == '\015')) a = ' ';
         if (a == ',' && quote) a = ' ';
         if (a == '\042') quote =! quote;
```
mrdudz, line 861: While at it, would you add braces here too, please?

polluks: Indeed, you could also combine both space assignments.

src/grc65/main.c
bbbradsmith: [1df7ab0](https://github.com/cc65/cc65/pull/2098/commits/1df7ab0352a43546c6ece76830594a687681c938)

src/grc65/main.c
bbbradsmith: I'm unclear about what this means. Where's the missing braces in my change?

bbbradsmith: Oh, you're reviewing the old pre-existing code. Sorry, that was confusing to me.

bbbradsmith: [532681c](https://github.com/cc65/cc65/pull/2098/commits/532681c9613af15b1993e08db12761855c9707c6)

mrdudz:
Looks good now! Yeah i commented on code not touched by you, sorry for the confusion :)

Issue #2087, opened 2023-05-04
long type has inconsistent type across various target platforms


bbbradsmith:
Most Windows builds (32 or 64 bit) will have a 32-bit long type. As least some Linux builds have 64-bit long type.

While working on #2084 I discovered different behaviour between [our automated builds[(https://github.com/cc65/cc65/actions/workflows/snapshot-on-push-master.yml) where the Windows build showed different results than the Linux build being used for our unit tests.

I resolved the difference that was relevant for that PR, which was about parsing constant literals, by restricting their size on input. However, this is just the tip of the iceberg.

The first suggestion I would have is running the unit tests also on the Windows build action to catch disparities like this. If that proves impractical, maybe a 32-bit Linux build could substitute.

A second suggestion, which would be a large amount of work, is to review [ExprDesc.IVal](https://github.com/cc65/cc65/blob/master/src/cc65/exprdesc.h), which is long type, and this affects how almost everything is evaluated. I would suggest changing this to an explicitly sized type, and then reviewing everywhere this data travels.

Beyond that I'm sure there's many places this could be relevant, but I don't think anyone will ever have the time to do such a full review. For similar reasons I think a full overhaul of the code to use explicit type sizes aggressively is probably out of the picture. At least unit testing the 2 main cases I think could help catch most of the issue.

bbbradsmith:
I was able to run the full make test with the MSVC windows builds, though there were two issues with sim65 that had to be addressed. (One of them was indeed about the size of long.)

PR #2092 adds a workflow that allows a Windows test to be manually triggered. I did not make it part of the PR/snapshot workflows because it was exceptionally slow.

So, the first suggestion is at least mostly viable. Thankfully the tests seem to indicate no unknown issues, and the manual workflow can be dispatched as often as we feel like checking up on it.

PR #2099, opened 2023-05-06
sim65 fix platform-dependent issues

bbbradsmith:
sim65 is suffering from #2087 issue of long being 64-bit on some platforms, 32-bit on others. This allowed a higher range of cycle count depending on the platform, and some of our tests were asking for values that were too large for 32-bit targets.

Also [test/val/constexpr.c](https://github.com/cc65/cc65/blob/master/test/val/constexpr.c) was making strange use of ```close(-1)``` expecting it to return ```-1```. MSVC instead treats this invalid parameter as a security issue and terminates the program instead of returning. Have added compensation for this.

Also fixed a spurious uninitalized variable warning.

This was one of the fixes needed to get #2101 working.

Split from #2092

### DIFF
--- a/src/sim65/main.c
+++ b/src/sim65/main.c
@@ -189,6 +189,7 @@ static unsigned char ReadProgramFile (void)
     }
 
     /* Get load address */
+    Val2 = 0; /* suppress uninitialized variable warning */
     if (((Val = fgetc(F)) == EOF) ||
         ((Val2 = fgetc(F)) == EOF)) {
         Error ("'%s': Header missing load address", ProgramFile);
--- a/src/sim65/main.c
+++ b/src/sim65/main.c
@@ -36,6 +36,7 @@
 #include <string.h>
 #include <stdlib.h>
 #include <errno.h>
+#include <limits.h>
 
 /* common */
 #include "abend.h"
@@ -140,6 +141,10 @@ static void OptQuitXIns (const char* Opt attribute ((unused)),
 /* quit after MaxCycles cycles */
 {
     MaxCycles = strtoul(Arg, NULL, 0);
+    /* Guard against overflow. */
+    if (MaxCycles == ULONG_MAX && errno == ERANGE) {
+        Error("'-x parameter out of range. Max: %lu",ULONG_MAX);
+    }
 }
 
 static unsigned char ReadProgramFile (void)
--- a/src/sim65/main.c
+++ b/src/sim65/main.c
@@ -64,6 +64,12 @@ const char* ProgramFile;
 /* exit simulator after MaxCycles Cycles */
 unsigned long MaxCycles;
 
+/* maximum number of cycles that can be tested,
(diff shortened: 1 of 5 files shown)


src/sim65/main.c
```diff
@@ -140,6 +141,10 @@ static void OptQuitXIns (const char* Opt attribute ((unused)),
 /* quit after MaxCycles cycles */
 {
     MaxCycles = strtoul(Arg, NULL, 0);
+    /* Guard against overflow. */
+    if (MaxCycles == ULONG_MAX && errno == ERANGE) {
+        Error("'-x parameter out of range. Max: %lu",ULONG_MAX);
+    }
 }
```
mrdudz, line 149: mmmh, while this works - wouldnt it be better (as in simpler) to just make MaxCycles a uint64_t ? That'd also preemptively solve future overflow problems (we actually had a lot of these in VICE)

rofl0r: or unsigned long long, which is guaranteed by C spec to be at least 64 bit, while having its own dedicated format string mnemonic "%llu" - using uint64_t requires macros like PRIu64 which requires hacks for MSVC.

bbbradsmith:
```
    /* Guard against overflow. */
    if (MaxCycles == ULONG_MAX && errno == ERANGE) {
        Error("'-x parameter out of range. Max: %lu",ULONG_MAX);
    }
```
> @mrdudz  mmmh, while this works - wouldnt it be better (as in simpler) to just make MaxCycles a uint64_t ? That'd also preemptively solve future overflow problems (we actually had a lot of these in VICE)

> @rofl0r or unsigned long long, which is guaranteed by C spec to be at least 64 bit, while having its own dedicated format string mnemonic "%llu" - using uint64_t requires macros like PRIu64 which requires hacks for MSVC.

I considered both of these possibilities, and even tried implementing both of them them actually, but both of them require a series of changes and have complications. Changing MaxCycles alone is not enough to support it, there's corresponding variables and function returns across multiple files that interact with it that end up having to be the larger type as well.

For uint64_t there is a strtoui64 but this seemed to be Microsoft specific. It did also mean adding inttypes.h to multiple files, and as rofl0r mentions it's problematic for printf. Using unsigned long long is more viable (though also C99 territory), but still requires quite a bit more than just changing MaxCycles.

The change I chose for this diagnoses the uncaught error problem as simply as I could without the complexity of trying to alter the function of the program. Augmenting this to universally support 64-bit is a bigger task, IMO. If we want it I would rather it be a separate unrelated task than this one which has a less ambitious goal.

src/sim65/main.c
bbbradsmith: I feel this is not simply a review of the code linked above so I have transferred it to a discussion below.

mrdudz:
OK, its fine like this, i didnt think it was more than a trivial change to the pointer type. Not very important to have it always 64bit either. merging...

PR #2100, opened 2023-05-06
libtest target alternative to lib

bbbradsmith:
Just a small change to add a make target that is only the lib needed to run tests. This saves me about 20 minutes when I only want to run the regression tests.

Split from #2092


mrdudz:
OK, seems reasonable... 20 minutes locally is quite an improvement :)

Issue #2095, opened 2023-05-05
make test very slow on Windows


bbbradsmith:
During investigation of #2092 it was found that "make test" runs about 5x slower on the Github action builds for Windows than Linux. I'm not sure of the cause.

It could be that the MSVC builds are really that much slower, in which case I think there must be some incorrect compiler settings, as there really should not be a 5x speedup between this and MinGW GCC.

The other stronger suspicion I have is that **"make" is somehow much slower on the Windows platform**. If that's the case I'm not sure if there's much we can do about it, unless there's something problematic about the way our particular makefiles are written that is an issue for Windows only.

Either way, it would be nice to understand it, and even nicer to fix it.

Edit: for quick reference I've made the correct answer **bold** above. (The binaries are fine, make is the problem.)

mrdudz:
> The other stronger suspicion I have is that "make" is somehow much slower on the Windows platform. 

That is the case. Windows is awfully slow at spawning processes (which happens a lot during make). I don't think there is a way to fix this. VICE builds really really slow on Windows for that matter (for the same reasons).

That said, the build environment doesn't seem to ever truly use a native windows shell - at least not for the subshells used by "make". see #1920 and #1937 for that matter

bbbradsmith:
So I made a test on my fork, building and testing both MSVC and GCC in the same job (just to be sure it's the same machine at least), and then another job in the reverse order. The time differences are not significant between the two, so I think that at least puts to rest the idea that the MSVC is horribly misconfigured. Details follow, if curious:

# Details

Starting from #2092 which fixes sim65 so that the MSVC test can pass.

I created [a workflow](https://github.com/bbbradsmith/cc65/blob/0749a59968af620b29f439b8d43fd30b46399702/.github/workflows/windows-performance-test.yml) that looked like this:
```
name: Windows Performance Test
on: [pull_request]

jobs:

  build_msvc_gcc:
    name: Build MSVC then GCC
    runs-on: windows-latest

    steps:
      - name: Git Configure
        shell: bash
        run: git config --global core.autocrlf input
      - name: Checkout Source
        uses: actions/checkout@v3

      - name: MSVC Add msbuild to PATH
        uses: microsoft/setup-msbuild@v1.1
      - name: MSVC Build app (release)
        run: msbuild src\cc65.sln -t:rebuild -property:Configuration=Release
      - name: MSVC make lib
        run: make -j2 lib QUIET=1
      - name: MSVC make test
        run: make test QUIET=1

[...]

Issue #1920, opened 2022-11-18
Tests should also run on windows with cmd.exe


mrdudz:
The github actions should perform a full build/test also when running on windows - that is to make sure the makefiles all keep working with cmd.exe (which we just restored... hopefully).

mrdudz:
I tried a bunch of things in #1937 but couldnt get it to work - seems GHA is always using msys bash, not cmd.exe

bbbradsmith:
Has this been resolved now with #2108 and is #1937 still relevant or should it be closed?

mrdudz:
Those 3 interconnected PRs are still confusing me :)

I think the original issue is solved. Kind of - i am not really comfortable with the CI partially running via cron instead of the PR trigger though, since that means failing checks won't be noticed eventually and something merged that shouldn't be... (not sure if all ppl who can merge would even see the result) it's a bit besides the point of having CI for PRs ;)

bbbradsmith:
Well, the only reason I suggested the manual/cron version was that it adds another 20 minutes to the windows job. If that's accceptable, though, you could definitely move the test into the PR/snapshot jobs.

mrdudz:
closed via #2401

PR #1029, opened 2020-05-02
Commander x16

CJLove:
This is a candidate set of changes to add support for the Commander X16 rom version r37.  See https://github.com/cc65/cc65/issues/1020.

Comments welcome, or if you want to proceed with @greg-king5's set of changes when they're available that is fine too.

Testing has included updated versions of the cc65-sprite and cc65-audio programs from the x16-demo repo (https://github.com/commanderx16/x16-demo), my Lode Runner port to the cx16 (https://github.com/CJLove/x16-LodeRunner) and a handful of small test programs I created which exercise functions touched by this set of change (https://github.com/CJLove/x16-cc65-testprogs)

PR #2101, opened 2023-05-06
workflow for manually dispatched Windows build and test

bbbradsmith:
This will help generally with additional build testing, and diagnosing cross-platform issues. The immediate goal was to help diagnose one particular issue (#2087) with type sizes that change for different platform targets.

In the process of testing this, I had to discover and resolve some existing cross-platform issues to get this workflow to be able to build green. I put all of these together in one PR initially (#2092) but as requested I have split them. These two additional PRs will have to be merged before this test can successfully run:
* #2098 
* #2099 

However, because make is apparently very slow on the Windows platform (#2095), I made this into an action that can be manually dispatched, rather than add it to the automatic snapshot or PR workflows. An extra 25 minutes of test time doesn't seem reasonable for every single PR.

So, because it's basically a way to test the build on-request, I's still okay to merge without those two pre-requisites.

I also considered running it as a scheduled action, maybe once every 5 days if there are any changes. However,  notifications for a scheduled action go to the person who committed it, and GitHub doesn't seem to have very good abilities to select which build failures you get or do not get e-mails for, unfortunately.

@mrdudz also requested that it use the cmd shell, so it does this. These two issues were cited:
* #1920 
* #1937

This workflow does successfully pass the full build and test with #2092 (#2092 =. #2098 + #2099 + this) I was also able build #2092 locally from my cmd shell. As far as I can tell this should obsolete both #1029 and #1937, but I'm not fully clear on the description of the problem in them.

### DIFF
--- a/.github/workflows/build-on-pull-request.yml
+++ b/.github/workflows/build-on-pull-request.yml
@@ -67,3 +67,7 @@ jobs:
 
       - name: Build app (release)
         run: msbuild src\cc65.sln -t:rebuild -property:Configuration=Release
+
+      # The regression tests are currently too slow to run for this Windows build,
+      # but the "Windows Test Manual" workflow (windows-test-manual.yml) can by
+      # manually dispatched from the Actions menu to test as needed.
--- a/.github/workflows/snapshot-on-push-master.yml
+++ b/.github/workflows/snapshot-on-push-master.yml
@@ -29,6 +29,10 @@ jobs:
       - name: Build app (release)
         run: msbuild src\cc65.sln -t:rebuild -property:Configuration=Release
 
+      # The regression tests are currently too slow to run for this Windows build,
+      # but the "Windows Test Manual" workflow (windows-test-manual.yml) can by
+      # manually dispatched from the Actions menu to test as needed.
+
   build_linux:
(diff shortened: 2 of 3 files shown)


bbbradsmith:
To test the workflow, I used my fork to test the equivalent of #2092. Here is the result of the successful run: [windows-test-manual.yml action](https://github.com/bbbradsmith/cc65/actions/runs/4898893536)

mrdudz:
> As far as I can tell this should obsolete both https://github.com/cc65/cc65/pull/1029 and https://github.com/cc65/cc65/pull/1937, but I'm not fully clear on the description of the problem in them.

see my explanation in #2095

In any case, this looks OK, merging - perhaps you can make another workflow that does the same, but is crontab scheduled to run once a day or so?

bbbradsmith:
> In any case, this looks OK, merging - perhaps you can make another workflow that does the same, but is crontab scheduled to run once a day or so?

I'm unwilling to make a cron action for reasons stated in the PR comment above. Mainly that I can't find any way to stop e-mail notifications from such a thing.

mrdudz:
Uh? Getting the email notification would be the whole point for having it? *shrug*

PR #2104, opened 2023-05-06
ca65: Remove .feature requirement for .addrsize

Movax12:
This will remove the need to use `.feature addrsize` to allow `.addrsize ` to be available. I am unsure if everyone agrees how to handle `.feature addrsize` that may exist in some projects (though this probably wasn't widely used). I opted to silently ignore it, but have the `.addrsize` function always available. Do we need any documentation that explains the .feature is there, but ignored?

### DIFF
--- a/src/ca65/feature.c
+++ b/src/ca65/feature.c
@@ -121,8 +121,6 @@ void SetFeature (feature_t Feature, unsigned char On)
         case FEAT_BRACKET_AS_INDIRECT:        BracketAsIndirect = On;    break;
         case FEAT_STRING_ESCAPES:             StringEscapes     = On;    break;
         case FEAT_LONG_JSR_JMP_RTS:           LongJsrJmpRts     = On;    break;
-        /* Accept, but ignore addrsize */
-        case FEAT_ADDRSIZE:                                              break;
         default:                                                         break;
     }
 }
--- a/src/ca65/feature.c
+++ b/src/ca65/feature.c
@@ -118,10 +118,11 @@ void SetFeature (feature_t Feature, unsigned char On)
         case FEAT_C_COMMENTS:                 CComments         = On;    break;
         case FEAT_FORCE_RANGE:                ForceRange        = On;    break;
         case FEAT_UNDERLINE_IN_NUMBERS:       UnderlineInNumbers= On;    break;
-        case FEAT_ADDRSIZE:                   AddrSize          = On;    break;
         case FEAT_BRACKET_AS_INDIRECT:        BracketAsIndirect = On;    break;
         case FEAT_STRING_ESCAPES:             StringEscapes     = On;    break;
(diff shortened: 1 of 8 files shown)


mrdudz:
I don't understand what the point of this change is - and why it's so many changes. Wouldn't just making it enabled by default achieve the exact same thing?

Movax12:
The goal is to make .addrsize always available. It would be simple to just enable it by default. Most of the changes remove what would become unneeded code (and documentation). Would you prefer to keep the global AddrSize and keep the ability to turn off .addrsize with .feature?

mrdudz:
I had to read up on what this actually IS :) And after doing so, i agree that it should simply be always available and that being able to enable/disable it with a .feature option is pointless.

merging...

mrdudz:
Oh wait, before i merge - i think it would be reasonable to emit a warning when the assembler sees .feature addrsize. something like ".feature addrsize is deprecated, .addrsize is always available" (or whatever better wording you come up with). Would you add that? :)

src/ca65/feature.c
```diff
@@ -118,10 +118,11 @@ void SetFeature (feature_t Feature, unsigned char On)
         case FEAT_C_COMMENTS:                 CComments         = On;    break;
         case FEAT_FORCE_RANGE:                ForceRange        = On;    break;
         case FEAT_UNDERLINE_IN_NUMBERS:       UnderlineInNumbers= On;    break;
-        case FEAT_ADDRSIZE:                   AddrSize          = On;    break;
         case FEAT_BRACKET_AS_INDIRECT:        BracketAsIndirect = On;    break;
         case FEAT_STRING_ESCAPES:             StringEscapes     = On;    break;
         case FEAT_LONG_JSR_JMP_RTS:           LongJsrJmpRts     = On;    break;
+        /* Accept, but ignore addrsize */
+        case FEAT_ADDRSIZE:                                              break;
```
Movax12, line 125: Should we leave this in the switch? Could just be removed?

mrdudz: Seems reasonable to just remove it, indeed

Movax12:
I think I am happy with this :)

src/ca65/pseudo.c
```diff
@@ -1043,6 +1043,12 @@ static void DoFeature (void)
             ErrorSkip ("Invalid feature: '%m%p'", &CurTok.SVal);
             return;
         }
+
+        if (Feature == FEAT_ADDRSIZE) {
+            /* Warn for depreciated .feature addrsize */
+            Warning (1, "Depreciated feature: '.feature addrsize'. Pseudo function .addrsize is always available.");
```
rofl0r, line 1049: typo, the right term is "deprecated". the comment above seems superfluous as the code is self-explanatory

Movax12: Thanks

Movax12:
Anything else for this PR? I think ready for merge.

mrdudz:
indeed, thanks!

PR #2105, opened 2023-05-07
Sim65 64bit cycle count

bbbradsmith:
This is a followup to #2099 where it was expressed that we would rather fix the problem by just supporting 64-bit values for the cycle count universally.

This provides that support.

It has tests to ensure the support work. (Thankfully they only take a few seconds to run a simulation for enough cycles.)

Also added some notes to the sim65 document, and moved the non-simulation error codes (argument error, timeout, etc.) to -1 and -2 where they can't conflict with sim test results, which are limited to unsigned 8-bit.

### DIFF
--- a/doc/sim65.sgml
+++ b/doc/sim65.sgml
@@ -126,9 +126,17 @@ a set of built-in paravirtualization functions (<ref id="paravirt-internal" name
 <sect>Creating a Test in Assembly<p>
 
 Assembly tests may similarly be assembled and linked with
-<tt/--target sim6502/ or <tt/--target sim65c02/,
-and the sim65 library provides an <tt/exit/ symbol that the program may <tt/JMP/
-to terminate with the current A register value as an exit code.
+<tt/--target sim6502/ or <tt/--target sim65c02/.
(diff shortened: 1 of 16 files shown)


bbbradsmith:
I'm not sure why `%llu` is causing no problem in error.c but it's creating an error in main.c?
```
sim65/6502.c
sim65/error.c
sim65/main.c
sim65/main.c: In function ‘main’:
sim65/main.c:318:78: error: unknown conversion type character ‘l’ in format [-Werror=format=]
  318 |                 ErrorCode (SIM65_ERROR_TIMEOUT, "Maximum number of cycles (%llu) reached.", MaxCycles);
      |                                                                              ^
sim65/main.c:318:49: error: too many arguments for format [-Werror=format-extra-args]
  318 |                 ErrorCode (SIM65_ERROR_TIMEOUT, "Maximum number of cycles (%llu) reached.", MaxCycles);
      |                                                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cc1: all warnings being treated as errors
```

Issue #2106, opened 2023-05-07
scheduled Windows build test


bbbradsmith:
#2101 added an action that an administrator can manually trigger to test the windows build. It was not added to the snapshot/PR workflows because of how slow it is (#2095), i.e. adds about 25 minutes to the test, which is otherwise only about 5m total.

[A scheduled version was requested](https://github.com/cc65/cc65/pull/2101#issuecomment-1537379780) but because I could not figure out a sensible way to control notifications from it.

This is different from 

* Also runs on a cron schedule.
* Create a cache storing the last used SHA so that it doesn't re-run the tests on no changes.
* Add an `if` to each build step to skip on no change.

Unlike push triggered actions, the notification seems to go to the person who created the action instead? Not exactly sure what the notification assignment is, but the lack of clarity in documentation and my own tests made me hesitant to create a PR that does it. I didn't want to have it generating e-mails all the time to myself or someone else who doesn't want then, and I couldn't see a way to properly/easily designate that.

Another downside of this caching approach is that after it fails once, if the cause isn't fixed before the next cron, the action will pass because it skips everything when the last one was successful. There's probably an easy solution to this... maybe deleting the cache on failure? I hadn't worked out how to do this.

The cron is set to every 5 days here. It seems that caches only last a maximum of 7 days normally, so it should at least be less than that.

```
name: Windows Test Schedule

on:
  schedule:
    - cron: '0 0 */5 * *'
  workflow_dispatch:
    # Allow manual dispatch.

jobs:
  build_windows:
    name: Build, Test (Windows MSVC)
    runs-on: windows-latest

    steps:
      - name: Setup
        shell: bash
        run: |
          git config --global core.autocrlf input
          mkdir cache-sha

      - name: Checkout source
        uses: actions/checkout@v3

      - name: Add msbuild to PATH
        uses: microsoft/setup-msbuild@v1.1

      - name: Cache SHA
        uses: actions/cache@v3
        id: cache-sha
        with:
          path: cache-sha
          key: cache-sha-${{ github.sha }}

      - name: Build app (MSVC debug)
        if: steps.check-sha.outputs.cache-hit != 'true'
        run: msbuild src\cc65.sln -t:rebuild -property:Configuration=Debug

      - name: Build app (MSVC release)
        if: steps.check-sha.outputs.cache-hit != 'true'
        run: msbuild src\cc65.sln -t:rebuild -property:Configuration=Release

      - name: Build utils (MinGW)
[...]

PR #2108, opened 2023-05-08
Scheduled windows build/test and cmd.exe subshell

bbbradsmith:
This makes two changes to the windows build workflow:
* Adds scheduled trigger, and caching for efficiency.
* Uses CMD shell for make.

I have been testing this [on my fork](https://github.com/bbbradsmith/cc65/actions/workflows/windows-test-manual.yml), and it appears to be successful in its goal.

# Scheduled test

Responding to [this comment](https://github.com/cc65/cc65/pull/2101#issuecomment-1537379780) on #2101, this provides a cron scheduled Github Action for the Windows build. It can also be manually triggered by an admin.

I set the schedule to once a day. Unlike on-push actions, I believe the notification will go only to the administrators, rather than the last pusher.

A cache is used to remember the last successful build hash. If there were no changes, it will skip the unnecessary re-build.

If the build fails, the cache is not saved, and it will continue to fail on schedule until fixed.

I had created issue #2106 to describe how to do this for later, but decided I should do it now.

# CMD shell for make

The make steps, despite running from `shell: cmd`, would open subshells with SH instead. I discovered that `SHELL=cmd` can override this default and provide a true build with CMD.

A very nice effect of this is that it runs `make test` about _10 minutes faster_ on the runners than with the default shell.

This should address:
* #2095 
* #1920 
* #1937 

~The only strange issue is that `-j2` causes `make lib` to fail with CMD (mysterious "Error 2"), but it's perfectly fine without it. When attempting diagnosis with `-d` it seems to go away. I suspect there is some concurrency problem with the lib make that only arises with some very specific timing, but I think it can be an issue to investigate later. It runs fine without '-j2'.~ _Edit: have included a minor fix to libsrc/Makefile to fix a concurrency issue in this which this build test exposed._

The subshell seems to be verifiable in the logs with `make -d`, on lines containing `find_and_set_shell()`. [An echo test](https://github.com/cc65/cc65/issues/2095#issuecomment-1537553828) was also suggested by mrdudz. In my testing both of these appeared to verify that CMD was used successfully.


bbbradsmith:
Just to follow up on the error with `make -j2 lib`, it also seems to occur without `QUIET=1`. Here is [one of my tests](seems to error at this point:) that uses `make -j2 lib SHELL=cmd` and fails just after archiving apple2enh.lib:
```
...
../bin/ar65 a ../lib/apple2enh.lib [... arguments omitted ...]
apple2enh - a2.auxmem.o
apple2enh - a2.stdjoy.o
apple2enh - a2.stdmou.o
apple2enh - a2.ssc.o
apple2enh - a2.hi.o
apple2enh - a2.lo.o
../bin/ld65 -o ../target/apple2enh/util/convert.system -C apple2enh-system.cfg ../libwrk/apple2enh/convert.o ../lib/apple2enh.lib
mingw32-make: *** [makefile:25: lib] Error 2
```

bbbradsmith:
Found the concurrency problem with `-j2` describe in #2109. The fix to the makefile is minor, so I've just included it in this PR.

mrdudz:
Thanks! lets see how it turns out :)

Issue #2111, opened 2023-05-10
cc65 should conflict of extern with static be an error?


bbbradsmith:
This example:
```
extern void func();
static void func(void) {}
```
Compiles only with a warning:
```
Warning: Static declaration of 'func' follows non-static declaration
```
This is an error with gcc or clang at compile time.  MSVC doesn't warn at compile time, but at link time it will error with an unresolved extern.

For exploration: [Example in Godbolt](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:___c,selection:(endColumn:1,endLineNumber:3,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:3,startColumn:1,startLineNumber:1),source:'extern+void+func()%3B%0Astatic+void+func(void)+%7B%7D%0A'),l:'5',n:'0',o:'C+source+%231',t:'0')),k:33.333333333333336,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:cc65_trunk,deviceViewOpen:'1',filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:___c,libs:!(),options:'',selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+6502+cc65+trunk+(Editor+%231)',t:'0')),k:33.333333333333336,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compilerName:'x86-64+gcc+13.1',editorid:1,fontScale:14,fontUsePx:'0',j:1,wrap:'1'),l:'5',n:'0',o:'Output+of+6502+cc65+trunk+(Compiler+%231)',t:'0')),k:33.33333333333333,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4)

Shouldn't we want an error here? My own expectation was error, since these two things are semantically incompatible. I'm not sure whether the C spec dictates how this should be handled.

kugelfuhr:
[...]

PR #2112, opened 2023-05-11
Error/warning for extern/static linkage declaration conflict

bbbradsmith:
See: #2111 

Changes:
* extern redeclared as static upgraded to error, rather than warning (C spec: undefined behaviour.)
* static redeclared as extern still permitted but now gives warning (C spec: extern is ignored)

Example C:
```
extern void a(void);
static void a(void) {}
static void b(void) {}
extern void b(void);
```
Output:
```
test.c:2: Error: Static declaration of 'a' follows non-static declaration
test.c:4: Warning: Extern declaration of 'b' follows static declaration, extern ignored
```

### DIFF
--- a/src/cc65/symtab.c
+++ b/src/cc65/symtab.c
@@ -1340,15 +1340,14 @@ SymEntry* AddGlobalSym (const char* Name, const Type* T, unsigned Flags)
                    Name);
             Entry = 0;
         } else if ((Flags & SC_ESUTYPEMASK) != SC_TYPEDEF) {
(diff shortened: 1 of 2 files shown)


bbbradsmith:
Modified [test/val/decl-static-extern.c](https://github.com/cc65/cc65/blob/e608f8a94f2a761b4c97e7fdcbea0a4fdebd9e26/test/val/decl-static-extern.c) created from #191 which was using warnings as errors and not expecting the new warning.

mrdudz:
thanks!

Issue #2110, opened 2023-05-09
cc65: NULL is not defined in stdlib.h


kugelfuhr:
The header stdlib.h should define the macro `NULL` according to the standard, but the cc65 version doesn't do this.

ISO/IEC 9899:1990 (E), 7.10 "General utilities <stdlib.h>": "The macros defined are NULL (described in 7.1.6); ..."

ISO/IEC 9899:1999 (E), 7.20p3 "General utilities <stdlib.h>": "The macros defined are NULL (described in 7.17); ..."

bbbradsmith:
Seems to be in string.h by mistake?
https://github.com/cc65/cc65/blob/55723f49770f8aac2e46243c7f61d441dbd47740/include/string.h#L39

mrdudz:
Am Freitag, 12. Mai 2023, 18:52:19 CEST schrieb Brad Smith:
> Seems to be in string.h by mistake?
> https://github.com/cc65/cc65/blob/55723f49770f8aac2e46243c7f61d441dbd47740/i
> nclude/string.h#L39

That seems to be correct

"The header <string.h> declares one type and several functions, and defines
one
macro useful for manipulating arrays of character type and other objects
treated as arrays
of character type.259) The type is size_t and the macro is NULL (both
described in
7.17). "

--

http://hitmen.eu                 http://ar.pokefinder.org
http://vice-emu.sourceforge.net  http://magicdisk.untergrund.net

A 'hacker' is any person who derives joy from discovering ways to circumvent
limitations.
<Robert Bickford>

bbbradsmith:
PR #2115. Noticed also that stddef.h chose the (void*) alternative, inconsistent with the others, so I changed it to 0 to match.

mrdudz:
see #2113

mrdudz:
test added in #2703

PR #2113, opened 2023-05-12
Unified NULL #2110

polluks2:


include/stdlib.h
```diff
@@ -44,6 +44,12 @@ typedef unsigned size_t;
 #define _HAVE_size_t
 #endif
 
+/* NULL pointer */
+#ifndef _HAVE_NULL
+#define NULL    ((void *) 0)
+#define _HAVE_NULL
+#endif
+
```
mrdudz, line 52: This test is broken - it should test for #ifndef NULL instead

kugelfuhr: > This test is broken - it should test for #ifndef NULL instead

Nope. The change is correct. The real problem is that some headers use

```C
#define NULL 0
```

while others use

```C
#define NULL ((void*)0)
```

This was working in 2.18 and messed up in #1826.

mrdudz:
> Nope. The change is correct. 

Are you saying that
```
#define NULL ((void*)0)
#include <stdlib.h>
```
should produce an error? In that case every compiler i am using is wrong.

mrdudz:
So these need to be changed accordingly
```
./locale.h:43:#define NULL    0
./time.h:43:#define NULL    0
./string.h:41:#define NULL    0
./stdio.h:42:#define NULL    0
```
also tests that check they exist and have the right value would be nice

kugelfuhr:
> Are you saying that
> 
> ```
> #define NULL ((void*)0)
> #include <stdlib.h>
> ```
> 
> should produce an error? In that case every compiler i am using is wrong. see https://godbolt.org/z/7znhe53aP

The truth is not in godbolt but in the C standard :smile:

ISO/IEC 9899:1999 (E), 7.3.1p1 "Reserved identifiers": "Each macro name in any of the following subclauses (including the future library directions) is reserved for use as specified if any of its associated headers is included; unless explicitly stated otherwise (see 7.1.4)."

ISO/IEC 9899:1999 (E), 7.3.1p2 "If the program declares or defines an identifier in a context in which it is reserved (other than as allowed by 7.1.4), or defines a reserved identifier as a macro name, the behavior is undefined."

So your code above code triggers undefined behavior and therefore any compiler behavior is correct.

mrdudz:
> The truth is not in godbolt but in the C standard

Sure. But often enough it shows the defacto standard behaviour

> So your code above code triggers undefined behavior and therefore any compiler behavior is correct.

OK. And once again i prefer the behaviour that other compilers (libraries in this case) implement.

mrdudz:
> Indeed, I would like to change the other headers to stddef.h.

Please do. And please use
```
#ifndef NULL
#define NULL    ((void *) 0)
#endif
```
in all of them.

bbbradsmith:
The definition of NULL should be consistent in all headers. stdlib.h was the only one using (void*) and the rest are just 0. They should all use the same one.

bbbradsmith:
> Indeed, I would like to change the other headers to stddef.h.
> I don't like a sys/null.h for just one define.
I do agree that a redundant guarded define is much better than requiring an extra file to be opened and parsed.

A fancier compiler would probably make such a difference unimportant with precompiled headers or other caching, but here it seems like a practical concern.

polluks2:
@mrdudz ready

mrdudz:
> @mrdudz ready

You forgot the one in gconst.h - see the failing test :)

polluks2:
`ld65: Error: 032-assert-error2.s:3: Code not at $0000`?

include/geos/gconst.h
```diff
@@ -4,14 +4,14 @@
...
 
-/* Here are constants which didn't fit into any other cathegory... */
+/* Here are constants which didn't fit into any other category... */
 
 #ifndef _GCONST_H
 #define _GCONST_H
 
-#define NULL            0
-#define FALSE           NULL
+#define NULL            ((void *) 0)
```
mrdudz, line 12: please guard this with #ifndef NULL

include/geos/gconst.h
```diff
@@ -4,14 +4,14 @@
...
 #define TRUE            0xff
+#define FALSE           ~TRUE
```
mrdudz, line 14: Please leave the definition of FALSE alone here. This is an entire different can of worms

mrdudz:
> `ld65: Error: 032-assert-error2.s:3: Code not at $0000`?

That is normal. Its val/add4.c that fails (since it includes stdlib and then redefines NULL) - that test (and perhaps others?) needs to be fixed (please remove the redefinition of NULL)

include/geos/gconst.h
polluks2: I don't want FALSE=((void *) 0)

mrdudz: As said, leave it alone for this PR. defining FALSE as ~TRUE is equally broken in the first place. Someone needs to find out why those are defined in this file at all, and whether they are GEOS specific values.

mrdudz: Oh and if (and only if) the tests still happen to fail because of the definition of FALSE, it should be defined as 0 (as it was before)

bbbradsmith:
This is the actual failing error: [test/val/add4.c](https://github.com/cc65/cc65/blob/70bf4d492be806bc43cf8c55cfe12f4de35cd4c5/test/val/add4.c#L25)
```
add4.c:25: Error: Macro redefinition is not identical
```

The only other test that defines null has a guard, so it should not need a change: [test/ref/dijkstra.c](https://github.com/cc65/cc65/blob/70bf4d492be806bc43cf8c55cfe12f4de35cd4c5/test/ref/dijkstra.c#L11)

I don't believe the FALSE defined in the geos header is used in any test or sample, so anything you do to it should not fail the checks. However I agree with polluks2 comment that FALSE should not be void* type, it should just be 0 rather than being defined as NULL (and receiving the changed semantic). There is a parallel definition in its assembly implementation which also assigns NULL to it, which should probably also be set to 0 for consistency.
[include/geos/gconst.h](https://github.com/cc65/cc65/blob/70bf4d492be806bc43cf8c55cfe12f4de35cd4c5/include/geos/gconst.h#L13)
[libsrc/geos-common/const.inc](https://github.com/cc65/cc65/blob/70bf4d492be806bc43cf8c55cfe12f4de35cd4c5/libsrc/geos-common/const.inc#L7)

mrdudz:
> I don't believe the FALSE or TRUE defined in the geos header is used anywhere in lib, test or sample, so anything you do to it should not fail the checks. However I agree with polluks2 comment that FALSE should not be void* type, it should just be 0 rather than being defined as NULL (previously assumed 0), and it should not be ~TRUE either (because TRUE is 0xff). There is a parallel definition in its assembly implementation which also assigns NULL to it, which should probably also be set to 0 for consistency.

To be honest, i lean to just *removing* all 3 (NULL, TRUE, FALSE) from that GEOS header - and then see what happens (or if someone complains and can come up with a good reason to add them back)

bbbradsmith:
If removing all 3 from include/geos/gconst.h, I'd also recommend removing TRUE/FALSE from libsrc/geos-common/const.inc. The assembly source does not use them, and it would be prudent to delete them as well so that any future maintainer doesn't accidentally think they should be used.

Also: [geos library documentation](https://cc65.github.io/doc/geos.html)

The documentation doesn't refer to TRUE or FALSE, always clarifying "non-zero" and "zero" only.

NULL in the geos lib assembly stuff is defined as 0 and should probably stay that way. Many of the assembly files in libsrc, and even some in the samples use NULL as a 0 byte.

Unfortunately in the documentation "NULL" is used throughout in an inconsistent way, sometimes referring to a pointer, sometimes referring to a byte (most often referring to a 0 byte at the end of a string). This should probably use clarification. Probably the easiest is to replace "NULL" with "0" in the docs where it refers to the non-pointer form. The lib source can probably stay as it is using NULL as a 0 byte.

mrdudz:
please change the definition of FALSE to 0 - ~0xff is broken (FALSE absolutely must be 0!)

mrdudz:
looks good now, thanks!

PR #2115, opened 2023-05-12
stdlib needs NULL, stddef has NULL inconsistend with others

bbbradsmith:
Addresses:
* #2110


mrdudz:
This patch does everything the way we don't want :)
a) we want ((void*)0)
b) we want to check #ifndef NULL
see #2113

bbbradsmith:
Oops, redundant, sorry.

PR #2124, opened 2023-05-23
Fix computed length for read and write for telestrat target

jedeoric:
The result of the length could be wrong depending of the ptr


polluks2 (commented):
sta tmp1 ᐅ pha
 lda tmp1 ᐅ pla

jedeoric:
fixed

mrdudz:
thanks!

Issue #2128, opened 2023-05-30
ca65: Unnamed labels should allow to use @


vrubleg:
I propose to allow optional `.LOCALCHAR` (`@` by default) in unnamed labels, just to make them a bit more visible. Also, the `:-` and `:+` could have `@-` and `@+` as more visible synonyms, or even `@<` and `@>` that is even more noticeable.

For example, now:

```
:	BIT PPU_STATUS
	BPL :-
```

After the change:

```
@:	BIT PPU_STATUS
	BPL @<
```

bbbradsmith:
I think the idea of allowing `@:` is not a bad one, but I had other concerns which I left on the associated PR. Quoting them here so that they aren't lost from the discussion.

From [my comment on the PR](https://github.com/cc65/cc65/pull/2132#issuecomment-1682523476):
> I don't think adding new aliases `<` and `>` for `-` and `+` have the same utility, and also they should not be special case for `@:` only. Either we should have them for both `:` and `@` or neither. (My vote is neither.)
> 
> Another question I have is whether `@-` searches the preceding local scope for _any_ label beginning with `@`, or whether it only includes `@:` or does it also include `:`?
> 
> I think intuitively I would expect `@-` to include all `@` labels, anonymous or not, and exclude the anonymous `:` that doesn't start with `@`, and correspondingly `:-` should include only `:` as it does currently, and not attach to `@:` as well. What does this PR do for these cases?

PR #2132, opened 2023-06-03
Add support of unnamed labels with @ (.localchar) prefix

vrubleg:
All local labels in ca65 use `@` prefix (or another symbol that can be selected using `.LOCALCHAR`). Unnamed symbols don't follow this rule and use inconsistent, hard to highlight, and hard to notice syntax with a mere `:` as a label (that is obviously not consistent with the `.FEATURE labels_without_colons +` setting) and weird looking `:+/:-` as a reference.

I propose to fix all these issues by introduction a new optional syntax where unnamed labels are defined by `@:`, and referenced by `@+/@-` or `@<` or `@>`. This new syntax is consistent with other local labels and follow the `.LOCALCHAR` and `.FEATURE labels_without_colons +` settings. Also it is easy to highlight in text editors, unnamed labels naturally get color of local labels.

Ticket #2128

### DIFF
--- a/doc/ca65.sgml
+++ b/doc/ca65.sgml
@@ -829,47 +829,40 @@ names like "Loop". Here is an example:
                 bne    @Loop            ; ERROR: Unknown identifier!
 </verb></tscreen>
 
+
 <sect1>Unnamed labels<p>
 
-If you really want to write messy code, there are also unnamed labels. These
-labels do not have a name (you guessed that already, didn't you?). A colon is
-used to mark the absence of the name.
+If you really want to write messy code, there are also unnamed labels. To define
+an unnamed label, use either <tt>@:</tt> (<tt>.LOCALCHAR</tt> is respected if it
+is set) or sole <tt>:</tt>.
 
-Unnamed labels may be accessed by using the colon plus several minus or plus
-characters as a label designator. Using the '-' characters will create a back
-reference (use the n'th label backwards), using '+' will create a forward
-reference (use the n'th label in forward direction). An example will help to
-understand this:
+To reference an unnamed label, use <tt>@</tt> (<tt>.LOCALCHAR</tt> is respected
+if it is set) or <tt>:</tt> with several <tt>-</tt> or <tt>+</tt> characters.
+The <tt>-</tt> characters will create a back reference (n'th label backwards),
+the <tt>+</tt> will create a forward reference (n'th label in forward direction).
+As an alternative, angle brackets <tt>&lt;</tt> and <tt>&gt;</tt> may be used
+instead of <tt>-</tt> and <tt>+</tt> with the same meaning.
+
+Example:
 
 <tscreen><verb>
-        :       lda     (ptr1),y        ; #1
-                cmp     (ptr2),y
-                bne     :+              ; -> #2
-                tax
-                beq     :+++            ; -> #4
-                iny
-                bne     :-              ; -> #1
-                inc     ptr1+1
-                inc     ptr2+1
-                bne     :-              ; -> #1
-
-        :       bcs     :+              ; #2 -> #3
-                ldx     #$FF
-                rts
-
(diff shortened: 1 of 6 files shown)


bbbradsmith:
My biggest concern with this PR is there is no test case which would both demonstrate how the feature is intended to work, and also verify that it does.

I understand the desire to want `@:` as an alias for `:`, and think this is an OK addition by itself.

However I don't think adding new aliases `<` and `>` for `-` and `+` have the same utility, and also they should not be special case for `@:` only. Either we should have them for both `:` and `@` or neither. (My vote is neither.)

Another question I have is whether `@-` searches the preceding local scope for _any_ label beginning with `@`, or whether it only includes `@:` or does it also include `:`?

I think intuitively I would expect `@-` to include all `@` labels, anonymous or not, and exclude the anonymous `:` that doesn't start with `@`, and correspondingly `:-` should include only `:` as it does currently, and not attach to `@:` as well. What does this PR do for these cases?

mrdudz:
> However I don't think adding new aliases < and > for - and + have the same utility, and also they should not be special case for @: only. Either we should have them for both : and @ or neither. (My vote is neither.)

I agree, that should be removed from this PR

vrubleg:
In the current PR, the `@:` and `:` are 100% synonyms, they just create an anonymous label.

The `:+`/`:-` and `@+`/`@-` or `@>`/`@<` are also complete synonyms and refer to a next or a previous anonymous label, so all non-anonymous `@labels` are not considered.

It is easy to add support of `<` and `>` into `:`, it is also easy to remove them from `@`.

vrubleg:
Making it a setting will complicate code somewhat, but if it is the only way to go, I can do this. In this case, I could propose two separate settings:

1. The style of an anonymous label definition (`@:` or `:`) that will also affect the first character of a reference (`@` or `:`).
```
.FEATURE localchar_anonymous_labels +
```
2. The style of reference counter (`</>` or `-/+`).
```
.FEATURE angle_bracket_anonymous_refs +
```

vrubleg:
Updated PR:
- Fixed a small issue in default VS project settings that I made previous time. `WholeProgramOptimization` was in wrong place and didn't work. Also added a few flags to generate more clean executables without [Rich headers](https://www.ntcore.com/files/richsign.htm).
- Added basic `.editorconfig` so Visual Studio follows some of conventions that are used in this project automatically.
- Fixed msbuild.bat to correctly find Visual Studio when it's not in default path.
- Added `.FEATURE line_continuations +` that fixes #2133 (moves `.LINECONT` into `.FEATURE` where all the other settings live, for consistency).
- The unnamed labels can be referenced also by `:<` and `:>`, in addition to all previous variants.

All of these features are in separate commits for convenience.

vrubleg:
Updated the docs to reflect the changes. I didn't make the feature settings to configure what style of unnamed labels should be used. All of them are accepted at the same time (with equal meaning) for better compatibility.

vrubleg:
Separated this MR into three PRs. #2204 is already merged, #2209 is pending. After it is merged, I'll rebase the current #2132 to include just changes related to the unnamed labels improvements.

mrdudz:
> Separated this PR into three PRs. #2204 is already merged, #2209 is pending. After it is merged, I'll rebase the current #2132 to include just changes related to the unnamed labels improvements.

Was that done? Just wondering if i should see this in the log here or not :)

vrubleg:
Rebased this branch. Now it includes just the unnamed label related changes =)

mrdudz:
This one really needs a test... The test should make sure the different variants actually work, and also that a mixture of the variants work (@+ and @> should be really synonymous).

Other than that i am willing to merge it :)

vrubleg:
The test was pushed a month ago =)

mrdudz:
Sorry - please always post something when a PR is ready/modified - i don't get notified when something is pushed to the branch only... merging

PR #2141, opened 2023-06-26
Update _atari5200os.h with POT shadows

baktragh:
Update the OS struct with POT shadow registers, according to the https://web.archive.org/web/20120830055323/http://www.atarimuseum.com/videogames/consoles/5200/conv_to_5200.html document

### DIFF
--- a/include/_atari5200os.h
+++ b/include/_atari5200os.h
@@ -54,14 +54,14 @@ struct __os {
     unsigned char color2;            // = $0E       PF color 2
     unsigned char color3;            // = $0F       PF color 3
     unsigned char color4;            // = $10       PF color 4
-    unsigned char pot0;              // = $11       POT0 shadow   
-    unsigned char pot1;              // = $12       POT1 shadow   
-    unsigned char pot2;              // = $13       POT2 shadow   
-    unsigned char pot3;              // = $14       POT3 shadow   
-    unsigned char pot4;              // = $15       POT4 shadow   
-    unsigned char pot5;              // = $16       POT5 shadow   
-    unsigned char pot6;              // = $17       POT6 shadow   
-    unsigned char pot7;              // = $18       POT7 shadow   
+    unsigned char pot0;              // = $11       POT0 shadow
+    unsigned char pot1;              // = $12       POT1 shadow
+    unsigned char pot2;              // = $13       POT2 shadow
+    unsigned char pot3;              // = $14       POT3 shadow
+    unsigned char pot4;              // = $15       POT4 shadow
+    unsigned char pot5;              // = $16       POT5 shadow
+    unsigned char pot6;              // = $17       POT6 shadow
+    unsigned char pot7;              // = $18       POT7 shadow
     unsigned char _free_1[0xE7];     // = $19-$FF   User space
 
     /*Stack*/
--- a/include/_atari5200os.h
+++ b/include/_atari5200os.h
@@ -54,7 +54,15 @@ struct __os {
     unsigned char color2;            // = $0E       PF color 2
     unsigned char color3;            // = $0F       PF color 3
     unsigned char color4;            // = $10       PF color 4
-    unsigned char _free_1[0xEF];     // = $11-$FF   User space
+    unsigned char pot0;              // = $11       POT0 shadow   
+    unsigned char pot1;              // = $12       POT1 shadow   
+    unsigned char pot2;              // = $13       POT2 shadow   
+    unsigned char pot3;              // = $14       POT3 shadow   
+    unsigned char pot4;              // = $15       POT4 shadow   
+    unsigned char pot5;              // = $16       POT5 shadow   
+    unsigned char pot6;              // = $17       POT6 shadow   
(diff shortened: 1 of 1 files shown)


mrdudz:
thanks!

include/_atari5200os.h
```diff
@@ -54,7 +54,15 @@ struct __os {
...
-    unsigned char _free_1[0xEF];     // = $11-$FF   User space
+    unsigned char pot0;              // = $11       POT0 shadow
+    unsigned char pot1;              // = $12       POT1 shadow
+    unsigned char pot2;              // = $13       POT2 shadow
+    unsigned char pot3;              // = $14       POT3 shadow
+    unsigned char pot4;              // = $15       POT4 shadow
+    unsigned char pot5;              // = $16       POT5 shadow
+    unsigned char pot6;              // = $17       POT6 shadow
+    unsigned char pot7;              // = $18       POT7 shadow
+    unsigned char _free_1[0xE7];     // = $19-$FF   User space
```
groessler, line 65: That's not in fact free if you use the conio screen. See in asminc/atari5200.inc.

And the runtime zero page vars are also there. (The _atarios.h file has the same problem.)

baktragh: So, what would be the best way of solving this? Is it good idea to include the conio and runtime variables in the header file? Usage of the conio is optional, the runtime is mandatory.

Perhaps it would be also good to synchronize the new shadow register names with the .asminc file - no pot0, but paddl0, and so on.

groessler: I think at least SAVMSC should be added, one might want to get a pointer to the screen memory. And then we can also add ROWCRS and COLCRS.

I would rename the _free_1 array to filler, _free sounds like one could use something there.

The runtime vars are placed by the linker. I wouldn't hardcode them here. Think of a customized linker file.

The synchronization of the names would be good.

baktragh: One more point. I see that for 5200, these SAVMSC, ROWCRS, COLCRS fields have the same names as for the Atari computers, yet they are specific to the conio library, not to the ROM OS. Is that something we are ok with, or should these be prefixed with conio_ string? I do not have a strong opinion in this matter, to be honest.

groessler: They have the same function on both machines. If the program wants to get a pointer to the screen memory, it can use SAVMSC, and doesn't need to use a different name depending whether compiling for the computers or the 5200.
For COLCRS and ROWCRS the names are different. IIRC I did it on purpose, because COLCRS is 2 bytes on the computers and 1 byte on the 5200, and to make the programmer aware of that fact. I didn't want to waste 1 byte on the ZP on the 5200 for the high byte of COLCRS. Maybe that wasted byte is not that bad, and the convenience of having the same names is worth more? I don't know...

baktragh: Yes, the sameness of the names is practical. Should I open a new PR?

groessler: Please also rename _free_1 to something else, since it is not free.
Why don't you want to add the xxxCRS and SAVMSC?

baktragh: Sure. I do not object against adding xxxCRS and SAVMSC.

groessler: savmsc is two bytes (1b/1c) and a pointer. See _atarios.h
unsigned char*  savmsc;

and xxxcrs_5200 are also only valid with conio (add a comment "; (conio)"

baktragh: Sorry for the savmsc. So rather as follows:
````
   unsigned char sdmctl;            // = $07       DMACTL shadow
   unsigned char pcolr0;            // = $08       PM color 0
   unsigned char pcolr1;            // = $09       PM color 1
   unsigned char pcolr2;            // = $0A       PM color 2
   unsigned char pcolr3;            // = $0B       PM color 3
   unsigned char color0;            // = $0C       PF color 0
   unsigned char color1;            // = $0D       PF color 1
   unsigned char color2;            // = $0E       PF color 2
   unsigned char color3;            // = $0F       PF color 3
   unsigned char color4;            // = $10       PF color 4
   unsigned char paddl0;            // = $11       POT0 Shadow
   unsigned char paddl1;            // = $12       POT1 Shadow
   unsigned char paddl2;            // = $13       POT2 Shadow
   unsigned char paddl3;            // = $14       POT3 Shadow
   unsigned char paddl4;            // = $15       POT4 Shadow
   unsigned char paddl5;            // = $16       POT5 Shadow
   unsigned char paddl6;            // = $17       POT6 Shadow
   unsigned char paddl7;            // = $18       POT7 Shadow
   
   /*cc65 runtime zero page variables*/
   unsigned char rowcrs_5200;       // = $19       Cursor row (conio)
   unsigned char colcrs_5200;       // = $1A       Cursor column (conio)
   unsigned char* savmsc;           // = $1B/$1C   Pointer to screen memory (conio)
    
   unsigned char _filler_1[0xE3];   // = $1D-$FF   Filler
````

groessler: fine for me

baktragh:
New PR #2146 opened

PR #2146, opened 2023-07-02
Update _atari5200os.h

baktragh:
Update the page 0 symbols, synchronize the names with the _atarios.h, add locations used by Atari 5200 conio. 
Continuation of the PR #2141 based on the discussion.


mrdudz:
thanks agaiN!

PR #2152, opened 2023-07-21
Allow MINGW environments to find installed config

markjfisher:
This PR allows MinGW environments in windows to build and compile from source, and have the installed directories (e.g. lib, asminc, cfg, ...) be found in the same way linux installs do.

Installing cc65 via a MinGW environment currently does not work fully without setting the following environment variables.

```shell
$ export CC65_HOME=/usr/local/cc65
$ export CA65_INC=/usr/local/cc65/share/cc65/asminc
$ export CC65_INC=/usr/local/cc65/share/cc65/include
$ export LD65_LIB=/usr/local/cc65/share/cc65/lib
```

(Or whatever prefix is preferred when building).

At the moment, the `_WIN32` variable blocks the values of CA65_INC, CC65_INC, CL65_TGT, LD65_LIB, LD65_OBJ, and LD65_CFG from being used in MINGW environments.

This PR adds a check for `__MINGW32__` definition, and if it exists, treats the path in the same way as a standard Linux build, thus allowing `/usr/local/share/cc65/` to act as a directory for configuration. This directory exists and is usable in MinGW environments.

This means the source can be compiled and used in Cygwin, MinGW64, UCRT64 environments, in the same way as Linux.

```shell
$ PREFIX=/usr/local/cc65 make
$ PREFIX=/usr/local/cc65 make install
```
will install the binaries, cfg etc under /usr/local/cc65/bin, and /usr/local/share/cc65/asminc, etc, and allow the `cl65` (etc) applications to find them in these directories without resorting to environment variables.

Note, there is no sudo for MinGW, so it is not needed for the install.

### DIFF
--- a/src/ca65/incpath.c
+++ b/src/ca65/incpath.c
@@ -75,8 +75,10 @@ void FinishIncludePaths (void)
     AddSubSearchPathFromEnv (IncSearchPath, "CC65_HOME", "asminc");
 
     /* Add some compiled-in search paths if defined at compile time. */
-#if defined(CA65_INC) && !defined(_WIN32) && !defined(_AMIGA)
+#if defined(CA65_INC) && !defined(_AMIGA)
+#if !defined(_WIN32) || defined(__MINGW32__)
     AddSearchPath (IncSearchPath, CA65_INC);
+#endif
 #endif
 
     /* Add paths relative to the parent directory of the Windows binary. */
--- a/src/cc65/incpath.c
+++ b/src/cc65/incpath.c
@@ -76,9 +76,12 @@ void FinishIncludePaths (void)
     AddSubSearchPathFromEnv (SysIncSearchPath, "CC65_HOME", "include");
 
     /* Add some compiled-in search paths if defined at compile time. */
-#if defined(CC65_INC) && !defined(_WIN32) && !defined(_AMIGA)
+#if defined(CC65_INC) && !defined(_AMIGA)
+#if !defined(_WIN32) || defined(__MINGW32__)
     AddSearchPath (SysIncSearchPath, CC65_INC);
 #endif
+#endif
+
 
     /* Add paths relative to the parent directory of the Windows binary. */
     AddSubSearchPathFromBin (SysIncSearchPath, "include");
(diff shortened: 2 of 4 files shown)


mrdudz:
I am not sure i understand what problem exactly this is trying to fix.... you build and compile in mingw (not msys, not cygwin), and then "make install" and use it on a msys shell?

markjfisher:
Yes. But without this PR, the configuration files installed are blocked from being add to the search path.

e.g.

```shell
cl65 -t atari -c --create-dep obj/atari/config.d -l obj/atari/config.lst  -o obj/atari/config.o src/config.s
cl65 -t atari -c --create-dep obj/atari/io_init.d -l obj/atari/io_init.lst  -o obj/atari/io_init.o src/atari/io_init.s
src/atari/io_init.s:6: Error: Cannot open include file 'atari.inc': No such file or directory
make: *** [Makefile.atari:150: obj/atari/io_init.o] Error 1
```

with this patch, the same makefile works fine:
```shell
cl65 -t atari -c --create-dep obj/atari/config.d -l obj/atari/config.lst  -o obj/atari/config.o src/config.s
cl65 -t atari -c --create-dep obj/atari/io_init.d -l obj/atari/io_init.lst  -o obj/atari/io_init.o src/atari/io_init.s
cl65 -t atari -c --create-dep obj/atari/setup_screen.d -l obj/atari/setup_screen.lst  -o obj/atari/setup_screen.o src/atari/setup_screen.s
cl65 -t atari -c --create-dep obj/atari/sio.d -l obj/atari/sio.lst  -o obj/atari/sio.o src/atari/sio.s
cl65 -t atari --mapfile build/config.map -l build/config.lst -o build/config.com -C src/atari/atari.cfg obj/atari/config.o obj/atari/io_init.o obj/atari/setup_screen.o obj/atari/sio.o
```

markjfisher:
Although I think you're pointing out that in MSYS (rather than the MinGW shell), if I compile it, it may work too. testing that.

mrdudz:
I have to admit that i never "install", but use the compiler in place (where the binaries are being compiled), also in msys and on linux :)

markjfisher:
And it did, because _WIN32 isn't set in MSYS. I had my environments mixed up. I'll close this PR, it isn't required.
Thanks for hinting at my error.

markjfisher:
> I have to admit that i never "install", but use the compiler in place (where the binaries are being compiled), also in msys and on linux

yeah, that works because of the flat nature of the directory, and by default it always simply looks in a sibling dir to `bin` for cfg etc and is how the windows built binaries are able to work, as the directory structure is geared to fit to the simple "remove bin, replace with asmdir" path searching.

I've learned a lesson in differences in MSYS/MinGW!

Thanks for your time.

PR #2153, opened 2023-07-23
Prevented the automatic switch to ISO mode on the CX16

cbpudding:
This is a relatively simple fix to an issue that has been bugging me for a while.

By default, the CX16 initialization causes the system to switch into ISO mode. This is not desired for many applications since it changes many of the characters on screen, the change isn't reverted when the program exits, and switching back clears the screen(which is also unwanted in a handful of cases).

The following is a screenshot of a simple "Hello world" program showing the effect of ISO mode on the platform.

image

The change is mostly apparent in the X16 logo in the top corner since all of the regular PETSCII graphics have been converted into ISO characters.

In short, if a programmer wishes to switch into ISO mode they should be able to by sending a `\x0E` character to the screen, but it should not be done automatically since it isn't suitable for every program.

### DIFF
--- a/libsrc/cx16/crt0.s
+++ b/libsrc/cx16/crt0.s
@@ -83,11 +83,6 @@ init:
         sta     sp
         stx     sp+1            ; Set argument stack ptr
 
-; Switch to the lower/UPPER PetSCII charset.
-
-        lda     #$0E
-        jsr     CHROUT
-
 ; Call the module constructors.
 
         jmp     initlib


mooinglemur:
This is a change to upper/lower mode.  ISO is not involved here. One can simply send #$8e to return to upper/petscii mode.

cc65 switches to upper/lower on other CBM platforms, including the C64.  See here https://github.com/cc65/cc65/blob/master/libsrc/c64/crt0.s#L102-L103

I think the current behavior should probably stay.

mrdudz:
Indeed, switching to lowercase is intended

cbpudding:
Assuming this *is* intended behavior, I can link with cx16-asm.cfg to prevent this at the very least. cc65 is the only compiler I'm aware of that makes this change for you. As @mooinglemur suggested, printing `\x8E` does fix the issue, however it is not apparent to beginners starting with C on the X16(or other platforms for that matter).

mrdudz:
> Assuming this is intended behavior

It sure is, check any of the other targets

> I can link with cx16-asm.cfg to prevent this at the very least.

Uhm, no. as the name suggests, this is for assembly programs. Any non trivial C program will not work that way.

> cc65 is the only compiler I'm aware of that makes this change for you.

On most other platforms, the default character set will contain upper- and lowercase - that's why. cc65 and its libraries are designed to make it possible to write and compile somewhat portable programs, which requires this.

cbpudding:
Since this is meant to preserve legacy behavior, I'm going to close this PR out. If anybody stumbles across this PR on Google at some point, the easiest way to deal with it is to print `\x8E` to the console to switch it back to PETSCII.

PR #2163, opened 2023-08-19
INIT segment definitions should be consistent, and bss is the appropriate semantic type

bbbradsmith:
From [this discussion on a recent PR](https://github.com/cc65/cc65/pull/2160#issuecomment-1683999799). Documentation stating the purpose of the INIT segment was given: [Wiki: Segment usage of constructors](https://github.com/cc65/wiki/wiki/Segment-usage-of-constructors)

INIT is uninitialized RAM to be used by constructors, separate from the BSS segment because it is expected that BSS will be cleared after constructors are initialized.

As a linker segment type, this semantically means `bss` type, because it has no initial value. I have amended the existing .cfg files to use this. It will also prevent us from accidentally trying to place initialization values in INIT.

The open question was whether the linker can emit a bss type segment to a file, and it seems there is no issue with that. (This was needed for Apple II configs which reserve bytes for INIT directly in the program data, despite them being uninitialized.)

The only outlier was creativision.cfg which treated INIT as a code segment for stuff doing at startup in one file, but I simply removed this incorrect usage (it only affected on file: creativision/cputc.s).

### DIFF
--- a/cfg/apple2-hgr.cfg
+++ b/cfg/apple2-hgr.cfg
@@ -27,7 +27,7 @@ SEGMENTS {
     CODE:     load = MAIN,           type = ro   start    = $4000;
     RODATA:   load = MAIN,           type = ro;
     DATA:     load = MAIN,           type = rw;
-    INIT:     load = MAIN,           type = rw;
+    INIT:     load = MAIN,           type = bss;
     ONCE:     load = MAIN,           type = ro,  define   = yes;
     LC:       load = MAIN, run = LC, type = ro,  optional = yes;
     BSS:      load = BSS,            type = bss, define   = yes;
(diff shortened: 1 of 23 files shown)


oliverschmidt:
> [...] separate from the BSS segment because it is expected that BSS will be cleared after constructors are initialized.

At least from my POV the point is rather, that there are linker configs where clearing BSS (potentially) overwrites (parts of) the constructors.

mrdudz:
So this is good to go? :)

bbbradsmith:
The platform libraries and samples still build, so I think that indicates we were already obeying the semantic everywhere (except the noted creativision case.) Don't know of anything else that needs testing.

mrdudz:
fine for me :)

colinleroy:
I think that's wrong on Apple2, see issue #2179...

Issue #2165, opened 2023-08-21
ser_get() issue on apple2


varna9000:
Hi, I'm struggling to get the below minimal example to work. It doesn't work both on real hardware and on Virtual ][ emulator (with file pipe serial emulation).

Using default SSC slot 2 doesn't work either. I'm able to send chars from apple2 to the other end, but I'm unable to receive chars into the apple2 i.e. `ser_get()` doesn't seem to work. Am I doing something wrong?

```
#include <stdlib.h>
#include <serial.h>
#include <apple2.h>
#include <stdio.h>
#include <conio.h>

  struct ser_params params = {
    SER_BAUD_1200,
    SER_BITS_8,
    SER_STOP_1,
    SER_PAR_NONE,
    SER_HS_HW
  };


void main(void)
{
  char Res, ch;

  ser_install(a2_ssc_ser);
  ser_apple2_slot(4);
  ser_open(&params);
 
    while (1){
      Res = ser_get(&ch);
      if (Res != SER_ERR_NO_DATA) {
        printf ("%c", ch);
        }
      if (kbhit())
        {
          ch=cgetc();
          ser_put(ch);
        }
    }
}
```

polluks2:
Did you try samples/terminal.c?

varna9000:
yes, but I just add the apple2 serial driver. i.e. I replace  `check ("ser_install", ser_install (ser_static_stddrv));`
with 
```
    check ("ser_install", ser_install (a2_ssc_ser));
    check ("ser_apple2_slot", ser_apple2_slot(4));
```
and include apple2.h on top.

It's the same result. I can send chars with ser_put() but can't receive anything with ser_get(). The serial card and connection are working fine because in BASIC after I do in#4 (card is in slot4) I can receive chars in BASIC just fine.

colinleroy:
Can you show the other side of the code?

varna9000:
@colinleroy There's no really code on the other side. In Virtual ][, I'm using [virtual serial with unix file pipes](https://www.virtualii.com/VirtualIIHelp/virtual_II_help.html). Alternatively, on real hardware, the SCC is hooked via null modem to my Macbook Pro, where I'm using [Serial2](https://www.decisivetactics.com/products/serial/) terminal to send/receive chars for testing. Both test yield same result - I can send chars from Apple2 but can't receive any.

colinleroy:
[...]

PR #2166, opened 2023-08-23
Possible serial driver optimisations

colinleroy:
Hello,

I have made some modifications to Apple2's serial driver and would like to discuss them with you. There are four different optimisations in this pull request. 

The first one is passing --cpu 65c02 when building libsrc/apple2 for the apple2enh ; we then use this to save a single cycle on SER_GET, with sta(ptr1) instead of sta(ptr1,x). 

The second one is in SER_GET too, and consists of *not* trying to send data before receiving. Is this an ACIA 6551 requirement that there's nothing to send before receving? If so, as SER_PUT is not IRQ-driven, maybe we could get rid of the SendBuf and just make sure the byte is out before returning ?
This one shaves 8 cycles if there was nothing to send (and approx 10 bytes on disk/in RAM). I could also put the patch in the other instances of the same driver for the other platforms.

The third one is in SER_IRQ, and reworks the RecvFreeCnt decrement and comparison to do it in the X register. This shaves 4 cycles in the common and best case; in the worst case (where we assert flow control), it saves no cycle as we have to reload X to Index. This one is generalizable too to the other platforms.

Finally, I've defined the most frequently accessed variables, and the ones that are in the "firehose" codepath of receiving bytes, to point to 5 unused bytes in the zero page, thus saving about one cycle per access. This last one is not clean, but could be made clean with a new segment definition and everything, I suppose.

All of these are ideas. What do you think?

Thanks!

PS: This does not save enough cycles to handle 19200 bps : (the ROM IRQ handler does about 240 cycles per call, the ProDOS IRQ handler about 250, and callirq about 42). Plus 54 in SER_IRQ, it's about 600 cycles per IRQ, but getting the data out of ser_get() (about 43 cycles) and into a large buffer still makes it too slow.

### DIFF
--- a/libsrc/apple2/ser/a2.ssc.s
+++ b/libsrc/apple2/ser/a2.ssc.s
@@ -328,20 +328,21 @@ SER_PUT:
 
         ; Try to send
         ldy     SendFreeCnt
-        iny                     ; Y = $FF?
+        cpy     #$FF            ; Nothing to flush
         beq     :+
         pha
         lda     #$00            ; TryHard = false
         jsr     TryToSend
         pla
 
-        ; Put byte into send buffer & send
-:       ldy     SendFreeCnt
+        ; Reload SendFreeCnt after TryToSend
+        ldy     SendFreeCnt
         bne     :+
         lda     #SER_ERR_OVERFLOW
         ldx     #0 ; return value is char
         rts
 
+        ; Put byte into send buffer & send
 :       ldy     SendTail
         sta     SendBuf,y
         inc     SendTail
--- a/libsrc/apple2/ser/a2.ssc.s
+++ b/libsrc/apple2/ser/a2.ssc.s
@@ -427,12 +427,13 @@ Done:   rts
 
 TryToSend:
         sta     tmp1            ; Remember tryHard flag
-Again:  lda     SendFreeCnt
+NextByte:
+        lda     SendFreeCnt
         cmp     #$FF
         beq     Quit            ; Bail out
 
         ; Check for flow stopped
-        lda     Stopped
+Again:  lda     Stopped
         bne     Quit            ; Bail out
 
         ; Check that ACIA is ready to send
@@ -449,4 +450,4 @@ Send:   ldy     SendHead
         sta     ACIA_DATA,x
         inc     SendHead
         inc     SendFreeCnt
-        jmp     Again
+        jmp     NextByte
--- a/libsrc/apple2/ser/a2.ssc.s
+++ b/libsrc/apple2/ser/a2.ssc.s
@@ -291,14 +291,9 @@ InvBaud:lda     #SER_ERR_BAUD_UNAVAIL
 
 SER_GET:
         ldx     Index
-        ldy     SendFreeCnt     ; Send data if necessary
-        iny                     ; Y == $FF?
-        beq     :+
-        lda     #$00            ; TryHard = false
-        jsr     TryToSend
 
         ; Check for buffer empty
-:       lda     RecvFreeCnt     ; (25)
+        lda     RecvFreeCnt     ; (25)
         cmp     #$FF
         bne     :+
         lda     #SER_ERR_NO_DATA
--- a/libsrc/apple2/ser/a2.ssc.s
+++ b/libsrc/apple2/ser/a2.ssc.s
@@ -26,6 +26,7 @@
         .include        "ser-error.inc"
 
         .macpack        module
+        .macpack        cpu
 
 ; ------------------------------------------------------------------------
 ; Header. Includes jump table
@@ -57,9 +58,13 @@
 ;----------------------------------------------------------------------------
 ; I/O definitions
 
+.if (.cpu .bitand CPU_ISET_65C02)
+ACIA            = $C088
+.else
 Offset          = $8F           ; Move 6502 false read out of I/O to page $BF
-
 ACIA            = $C088-Offset
+.endif
+
 ACIA_DATA       = ACIA+0        ; Data register
 ACIA_STATUS     = ACIA+1        ; Status register
 ACIA_CMD        = ACIA+2        ; Command register
@@ -200,7 +205,9 @@ SER_OPEN:
         asl
         asl
         asl
+.if .not (.cpu .bitand CPU_ISET_65C02)
         adc     #Offset                 ; Assume carry to be clear
+.endif
         tax
 
         ; Check if the handshake setting is valid
@@ -315,7 +322,11 @@ SER_GET:
         inc     RecvHead
         inc     RecvFreeCnt
         ldx     #$00            ; (59)
+.if (.cpu .bitand CPU_ISET_65C02)
+        sta     (ptr1)
+.else
         sta     (ptr1,x)
+.endif
         txa                     ; Return code = 0
         rts
 
--- a/libsrc/apple2/ser/a2.ssc.s
+++ b/libsrc/apple2/ser/a2.ssc.s
@@ -404,19 +404,19 @@ SER_IRQ:
         and     #$08
         beq     Done            ; Jump if no ACIA interrupt
         lda     ACIA_DATA,x     ; Get byte from ACIA
-        ldy     RecvFreeCnt     ; Check if we have free space left
+        ldx     RecvFreeCnt     ; Check if we have free space left
         beq     Flow            ; Jump if no space in receive buffer
         ldy     RecvTail        ; Load buffer pointer
         sta     RecvBuf,y       ; Store received byte in buffer
         inc     RecvTail        ; Increment buffer pointer
         dec     RecvFreeCnt     ; Decrement free space counter
-        ldy     RecvFreeCnt     ; Check for buffer space low
-        cpy     #33
+        cpx     #33             ; Check for buffer space low
         bcc     Flow            ; Assert flow control if buffer space low
         rts                     ; Interrupt handled (carry already set)
 
         ; Assert flow control if buffer space too low
-Flow:   lda     RtsOff
+Flow:   ldx     Index
+lda     RtsOff
         sta     ACIA_CMD,x
         sta     Stopped
         sec                     ; Interrupt handled
--- a/libsrc/atmos/ser/atmos-acia.s
+++ b/libsrc/atmos/ser/atmos-acia.s
@@ -352,12 +352,13 @@ Done:   rts
 
 TryToSend:
         sta     tmp1            ; Remember tryHard flag
-Again:  lda     SendFreeCnt
+NextByte:
+        lda     SendFreeCnt
         cmp     #$FF
         beq     Quit            ; Bail out
 
         ; Check for flow stopped
-        lda     Stopped
+Again:  lda     Stopped
         bne     Quit            ; Bail out
 
         ; Check that ACIA is ready to send
@@ -374,4 +375,4 @@ Send:   ldy     SendHead
         sta     ACIA::DATA
         inc     SendHead
         inc     SendFreeCnt
-        jmp     Again
+        jmp     NextByte
--- a/libsrc/atmos/ser/atmos-acia.s
+++ b/libsrc/atmos/ser/atmos-acia.s
@@ -227,14 +227,8 @@ InvBaud:lda     #<SER_ERR_BAUD_UNAVAIL
 ; returned.
 
 SER_GET:
-        ldy     SendFreeCnt     ; Send data if necessary
-        iny                     ; Y == $FF?
-        beq     :+
-        lda     #$00            ; TryHard = false
-        jsr     TryToSend
-
         ; Check for buffer empty
-:       lda     RecvFreeCnt     ; (25)
+        lda     RecvFreeCnt     ; (25)
         cmp     #$FF
         bne     :+
         lda     #SER_ERR_NO_DATA
--- a/libsrc/atmos/ser/atmos-acia.s
+++ b/libsrc/atmos/ser/atmos-acia.s
@@ -329,19 +329,19 @@ SER_IRQ:
         and     #$08
         beq     Done            ; Jump if no ACIA interrupt
         lda     ACIA::DATA,x    ; Get byte from ACIA
-        ldy     RecvFreeCnt     ; Check if we have free space left
+        ldx     RecvFreeCnt     ; Check if we have free space left
         beq     Flow            ; Jump if no space in receive buffer
         ldy     RecvTail        ; Load buffer pointer
         sta     RecvBuf,y       ; Store received byte in buffer
         inc     RecvTail        ; Increment buffer pointer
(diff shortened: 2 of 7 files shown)


libsrc/apple2/ser/a2.ssc.s
```diff
@@ -284,14 +288,8 @@ InvBaud:lda     #SER_ERR_BAUD_UNAVAIL
...
-        ldy     SendFreeCnt     ; Send data if necessary
-        iny                     ; Y == $FF?
-        beq     :+
-        lda     #$00            ; TryHard = false
-        jsr     TryToSend
-
         ; Check for buffer empty
-:       lda     RecvFreeCnt     ; (25)
+        lda     RecvFreeCnt     ; (25)
         cmp     #$FF
```
colinleroy, line 293: This is where we don't try to send.

libsrc/apple2/ser/a2.ssc.s
```diff
@@ -315,7 +313,11 @@ SER_GET:
         inc     RecvHead
         inc     RecvFreeCnt
         ldx     #$00            ; (59)
+.if (.cpu .bitand CPU_ISET_65C02)
+        sta     (ptr1)
+.else
         sta     (ptr1,x)
+.endif
         txa                     ; Return code = 0
```
colinleroy, line 321: This is the micro-optimization for 65C02.

libsrc/apple2/ser/a2.ssc.s
```diff
@@ -404,19 +406,20 @@ SER_IRQ:
...
+        stx     RecvFreeCnt     ; Store it
+        cpx     #33
         bcc     Flow            ; Assert flow control if buffer space low
         rts                     ; Interrupt handled (carry already set)
 
         ; Assert flow control if buffer space too low
-Flow:   lda     RtsOff
+Flow:   ldx     Index           ; Reload port
+        lda     RtsOff
         sta     ACIA_CMD,x
```
colinleroy, line 423: This is a faster way to decrement RecvFreeCnt in the nominal case

libsrc/apple2/ser/a2.ssc.s
```diff
@@ -68,23 +69,26 @@ ACIA_CTRL       = ACIA+3        ; Control register
...
+; Use free bytes in zero page for the most-often accessed variables
+; Ref: https://fadden.com/apple2/dl/zero-page.txt
+
+RecvHead   := $EB               ; Head of receive buffer
+RecvTail   := $EC               ; Tail of receive buffer
+RecvFreeCnt:= $ED               ; Number of bytes in receive buffer
+Stopped    := $EE               ; Flow-stopped flag
+Index      := $EF               ; I/O register index
+
         .bss
```
colinleroy, line 81: This is the "critical vars in ZP" optimisation

polluks2:
Good work! By the way can you help #2165?

colinleroy:
I'll try! I remember I had a hard time at the beginning, but it was due to serial shenanigans, not the driver.

libsrc/apple2/ser/a2.ssc.s
```diff
@@ -449,4 +453,4 @@ Send:   ldy     SendHead
         sta     ACIA_DATA,x
         inc     SendHead
         inc     SendFreeCnt
-        jmp     Again
+        jmp     NextByte
```
colinleroy, line 456: We don't need to reload SendFreeCnt when waiting for ACIA

libsrc/apple2/ser/a2.ssc.s
```diff
@@ -328,21 +330,21 @@ SER_PUT:
...
         lda     #SER_ERR_OVERFLOW
         ldx     #0 ; return value is char
         rts
 
-:       ldy     SendTail
+PutByte:
+        ldy     SendTail
         sta     SendBuf,y
         inc     SendTail
         dec     SendFreeCnt
```
colinleroy, line 350: Simplification of SER_PUT: cpy and iny are each two cycles, but if we have nothing to send, we can branch directly to PutByte instead of line 339

oliverschmidt:
First of all thanks for your interest in (further) contributing to cc65!

1. The changes in the Makefile are supposed to be unnecessary as the `-t apple2enh` implies setting the 65C02 CPU.

2. I don't see us accessing "some" ZP variables directly from "some" source file. You write that they are unused. When next week someone does a similar change in some other source code, he might claim the same. I think that things don't work that way. From that perspective, using "some" ZP variables from within an interrupt handler is the worst case as the usage is unpredictable.

colinleroy:
> First of all thanks for your interest in (further) contributing to cc65!

And thank you for the support in this journey!

>     1. The changes in the Makefile are supposed to be unnecessary as the `-t apple2enh` implies setting the 65C02 CPU.

Oh indeed, it does :)

>     2. I don't see us accessing "some" ZP variables directly from "some" source file. You write that they are unused. When next week someone does a similar change in some other source code, he might claim the same. I think that things don't work that way. From that perspective, using "some" ZP variables from within an interrupt handler is the worst case as the usage is unpredictable.

Yes, that's the fishy one. That's why I wrote that if it is deemed a good idea, it should be cleaned up with a new segment definition, so that they're not "some" variables anymore. But even then, we couldn't be sure whether someone, somewhere, already utilizes those 5 bytes in their own code. So, probably the best is for me to revert that part.

I hope the rest is at least somewhat good :)

oliverschmidt:
As you are working so intensively with that code, I ask you this:

My understanding is that the 6551 doesn't only generate an interrupt when there's a new byte available but also when a hardware handshake line changes. But the serial driver code filters those interrupts and thinks that the interrupt wasn't caused by the 6551.

This leads as far as I can see to an "unclaimed interrupt". The more recent ProDOS versions allow for 255 unclaimed interrupts be halting with a fatal error. So my assumption is that after >255 hardware handshake line changes, the system dies.

If you are willing and able to spend the time and effort to test this scenario, it would great to learn if my concern is correct.

colinleroy:
> My understanding is that the 6551 doesn't only generate an interrupt when there's a new byte available but also when a hardware handshake line changes. But the serial driver code filters those interrupts and thinks that the interrupt wasn't caused by the 6551.

I have noticed that we handle the interrupt only if we have a byte to fetch, yes (ACIA_STATUS & 0x08). And also that we're supposed to check the bit 7 of the status register - so I tried to do that at some point, with no success: bit 7 was always 0. According to http://csbruce.com/cbm/ftp/reference/swiftlink.txt, "bit #7 is not consistently set.  At speeds
under 9600 bps, bit #7 seems to work as designed." So we can't use this if we want to handle 9600bps (and frankly it would be so sad to stay at 4800 :sweat:).
 
> This leads as far as I can see to an "unclaimed interrupt". The more recent ProDOS versions allow for 255 unclaimed interrupts be halting with a fatal error. So my assumption is that after >255 hardware handshake line changes, the system dies.

I suppose this HW handshake change should come from the other side, then ? I'll try that.

colinleroy:
> As you are working so intensively with that code, I ask you this:
> 
> My understanding is that the 6551 doesn't only generate an interrupt when there's a new byte available but also when a hardware handshake line changes. But the serial driver code filters those interrupts and thinks that the interrupt wasn't caused by the 6551.

So, as far as my tests went, no IRQs are fired when the line status changes. I've tested both when changing flow control from the Apple 2 itself (with the attached patch, and attached test file).

When changing it from "outside", this time with a patch storing 'Z' in the ring buffer instead of returning from SER_IRQ when ACIA_STATUS&0x08 == 0, and plugging/unplugging the serial adapter, nothing is inserted in RecvBuf and SER_GET still return SER_ERR_NO_DATA.

I had to make this second test this way because my emulation chain doesn't handle CRTSCTS, so I did it on bare metal. Fro the first test, I ran it under MAME and verified the absence of SER_IRQ calls in the trace
[flow.patch.gz](https://github.com/cc65/cc65/files/12448611/flow.patch.gz)
[test.c.gz](https://github.com/cc65/cc65/files/12448612/test.c.gz)
log.

polluks2:
By the way do you know https://sites.google.com/site/drjohnbmatthews/apple2/ssc?

oliverschmidt:
Thanks for the link but our knowledge on the topic is already well above of what is provided in that article.

oliverschmidt:
> SwiftLink-232 Application Notes

Yes, I know that document and especially the section on the interrupt bit.

> So, as far as my tests went, no IRQs are fired when the line status changes.

:-( That doesn't match any documentation - incl. the one on the SwiftLink-232...

> plugging/unplugging the serial adapter

That doesn't seem like a good way to test things to me. My understanding is that you use the SSC to connect your //c to a Raspberry Pi. My assumption is that you have the HSKi pin (https://old.pinouts.ru/SerialPorts/apple_iic_serial_pinout.shtml) connected to an output pin on the Pi. To which one?

If you look at pages 4 and 5 of http://www.applelogic.org/files/IICSCHEMATIC.pdf (and additionally take https://www.onsemi.com/pdf/datasheet/mc1488-d.pdf) into account you'll see that HSKi is connected to the 6551 DCD pin (not the DSR pin as one would assume!) so that's what is supposed to see changes.

colinleroy:
> :-( That doesn't match any documentation - incl. the one on the SwiftLink-232...

Indeed. I'll try again at 1200bps to check if it's the same kind of issue as bit 7 of status register not raising to 1 when an IRQ fired.
 
> > plugging/unplugging the serial adapter
> 
> That doesn't seem like a good way to test things to me. My understanding is that you use the SSC to connect your //c to a Raspberry Pi. My assumption is that you have the HSKi pin (https://old.pinouts.ru/SerialPorts/apple_iic_serial_pinout.shtml) connected to an output pin on the Pi. To which one?
> 
> If you look at pages 4 and 5 of http://www.applelogic.org/files/IICSCHEMATIC.pdf (and additionally take https://www.onsemi.com/pdf/datasheet/mc1488-d.pdf) into account you'll see that HSKi is connected to the 6551 DCD pin (not the DSR pin as one would assume!) so that's what is supposed to see changes.

Sadly, I can't really tell you - I'm using a monster cable (that I know passes handshake status correctly, as I don't lose bytes on the Apple2 whereas I do in certain conditions on the emulator) which I'm not prepared to rip apart...
image

colinleroy:
> > 
> 
> Indeed. I'll try again at 1200bps to check if it's the same kind of issue as bit 7 of status register not raising to 1 when an IRQ fired.

No, it doesn't :( - tried both on emulator (where I grepped the trace for ser_irq afterwards), and on real hardware, where I stopped/started the flow 500 times each, in alternance.

Triying to figure out a good test for the other way.

oliverschmidt:
>> My assumption is that you have the HSKi pin (https://old.pinouts.ru/SerialPorts/apple_iic_serial_pinout.shtml) connected to an output pin on the Pi. To which one?

> Sadly, I can't really tell you - I'm using a monster cable [...] which I'm not prepared to rip apart...

1. There's no need to open a cable to learn its pinout. You can measure it. As there are different types of DB9-to-DB25 adapters, I'd be interested to know on which DB9 pin the HSKi pin ends up.

2. What chipset is your Serial-to-USB dongle using? Still presuming that you use it with a Pi, you should be able to determine that both by the USB VID (Veondor ID) and/or the loaded driver.

3. Anyhow, most likely the virtual DTR signal of the virtual serial port ends up being connected to the HSKi pin. If you connect the whole thing to a Windows PC so that a virtual COM port is created and then and have PuTTY open that COM port, then DTR gets asserted (and de-asserted when PuTTY closes the COM port). If all my assumptions are correct, then you should see this as DCD change on the //c.

oliverschmidt:
1. Cool, that you really got this all figured out!

2. My assumption that the virtual DTR on the USB side ends up as HSKi on the //c side was correct :-)

3. I have no idea, why things don't behave as expected. I know for sure that the 6551 generates interrupts on HW handshake line changes incl. DCD and the article about the SwiftLink-232 is very clear about them too. And I know for sure about the ProDOS unclaimed interrupts as I've seen them more often than I'd like ;-)

I guess we need to close this exploration for now with the preliminary result that you're unable to trigger the problem I claim to exist in the driver. I thank you very much for the time and effort spent!

oliverschmidt:
> is modern ProDOS open-source/free-software?

No, only the ProDOS 1.7 source code became available some years ago.

polluks2:
Indeed, you find inthandler on page 171 of https://mirrors.apple2.org.za/ftp.apple.asimov.net/documentation/source_code/apple2_SRC_ProDOS17_1988.pdf

oliverschmidt:
That's not the relevant code. It starts on page 276 @ irqrecev with the code in question being on page 277 @ irqdeath:

```
*************** see rev note #35 *************************
irqdeath        inc          irqcount                  ; allow 255 unclaimed interrupts before
                bne          irqdone                   ; going to system death...
**********************************************************
```

colinleroy:
Indeed. 
Well, I guess it's a good thing the ACIA doesn't raise an IRQ when the line state changes then :)

On the initial matter, what do you think of the changes in the current state of this PR?

colinleroy:
Indeed. If you folks approve of the changes, I'll be OK to report them everywhere, and if desirable, factorize the common parts in libsrc/serial.

oliverschmidt:
> [...] if desirable, factorize the common parts in libsrc/serial.

As far as I know, this would be in general rather complex because the build system presumes that drivers consist of exactly one source file. A possible workaround might sharing on source code level, see https://github.com/cc65/cc65/blob/master/libsrc/atari/tgi/atr10.s#L79 for an at least sort of similar scenario.

colinleroy:
OK there are numerous little differences that make me uncomfortable with factorization, being unable to test the other platforms. I'll go with reporting the changes :)

colinleroy:
PR updated for each driver. I have split the commits by function for legibility, and I peeled my eyes real hard for typos, but it's probably worth it that at least another pair of eyes checks for those.

colinleroy:
So what do you think, folks? I have used that branch since two weeks and it works absolutely perfectly on apple2enh. It should be the same on the other targets :)

oliverschmidt:
@colinleroy has tested the proposed code more than I could/would so there's nothing for me to add from the Apple II maintainer perspective. I'm fine with the PR From the Apple II maintainer perspective.

colinleroy:
(Sorry, force-pushed to clean accidental merge of master that makes the history less clear)

colinleroy:
> So what do you think, folks? I have used that branch since two weeks and it works absolutely perfectly on apple2enh. It should be the same on the other targets :)

@polluks2, @mrdudz, are you fine with this too?

mrdudz:
I can not test this, i don't have a swiftlink (and i don't trust emulators at all for this kind of thing). Actually even if, i wouldn't know how (that terminal example isn't terribly useful for this).

That said, the changes look OK to me. I will close eyes and click merge, we'll see what happens :)

colinleroy:
I'll be there in case of issues!

PR #2169, opened 2023-08-24
Rework debug file format to include all exports

colinleroy:
Some symbols/addresses address values are missing in the .dbg file, for imports namely. This makes mapping addresses to symbols more complicated than needed, and requires us to generate both a .dbg and a .lbl file (where those appear).

Cf next comment for a solution that allows us to get all available addresses.

### DIFF
--- a/src/dbginfo/dbginfo.c
+++ b/src/dbginfo/dbginfo.c
@@ -188,6 +188,7 @@ struct DbgInfo {
     ** actually the ones that "own" the items.
     */
     Collection          CSymInfoById;   /* C symbol infos sorted by id */
+    Collection          ExportInfoById; /* Export infos sorted by id */
     Collection          FileInfoById;   /* File infos sorted by id */
     Collection          LibInfoById;    /* Library infos sorted by id */
     Collection          LineInfoById;   /* Line infos sorted by id */
@@ -199,13 +200,15 @@ struct DbgInfo {
     Collection          TypeInfoById;   /* Type infos sorted by id */
 
     /* Collections with other sort criteria */
-    Collection          CSymFuncByName; /* C functions sorted by name */
-    Collection          FileInfoByName; /* File infos sorted by name */
-    Collection          ModInfoByName;  /* Module info sorted by name */
-    Collection          ScopeInfoByName;/* Scope infos sorted by name */
-    Collection          SegInfoByName;  /* Segment infos sorted by name */
-    Collection          SymInfoByName;  /* Symbol infos sorted by name */
-    Collection          SymInfoByVal;   /* Symbol infos sorted by value */
+    Collection          CSymFuncByName;   /* C functions sorted by name */
+    Collection          ExportInfoByName; /* Export infos sorted by name */
+    Collection          ExportInfoByVal;  /* Export infos sorted by value */
+    Collection          FileInfoByName;   /* File infos sorted by name */
+    Collection          ModInfoByName;    /* Module info sorted by name */
+    Collection          ScopeInfoByName;  /* Scope infos sorted by name */
+    Collection          SegInfoByName;    /* Segment infos sorted by name */
+    Collection          SymInfoByName;    /* Symbol infos sorted by name */
+    Collection          SymInfoByVal;     /* Symbol infos sorted by value */
 
     /* Other stuff */
     SpanInfoList        SpanInfoByAddr; /* Span infos sorted by unique address */
@@ -237,6 +240,7 @@ struct InputData {
 
 /* Typedefs for the item structures. Do also serve as forwards */
 typedef struct CSymInfo CSymInfo;
+typedef struct ExportInfo ExportInfo;
 typedef struct FileInfo FileInfo;
 typedef struct LibInfo LibInfo;
 typedef struct LineInfo LineInfo;
@@ -266,6 +270,19 @@ struct CSymInfo {
         ScopeInfo*      Info;           /* Pointer to scope */
     } Scope;
     char                Name[1];        /* Name of file with full path */
+};
+
+/* Internally used export info struct */
+struct ExportInfo {
+    unsigned            Id;             /* Id of export */
+    cc65_addr           Value;          /* Address of export */
+    cc65_size           Size;           /* Size of export */
+    union {
+        unsigned        Id;             /* Id of attached asm symbol */
+        SymInfo*        Info;           /* Pointer to attached asm symbol */
+    } Sym;
+    Collection*         ImportList;     /* List of imports */
+    char                Name[1];        /* Name of export */
 };
 
 /* Internally used file info struct */
@@ -375,7 +392,7 @@ struct SymInfo {
     cc65_size           Size;           /* Size of symbol */
     union {
         unsigned        Id;             /* Id of export if any */
-        SymInfo*        Info;           /* Pointer to export if any */
+        ExportInfo*     Info;           /* Pointer to export if any */
     } Exp;
     union {
         unsigned        Id;             /* Id of segment if any */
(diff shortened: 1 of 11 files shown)


colinleroy:
This is a much bigger pull request now :laughing: 
I wanted to update the wiki, but I can't figure how to do pull requests there, it works weirdly. You can see the diff at: 
https://github.com/cc65/wiki/compare/master...colinleroy:add-exports-to-dbgfile?expand=1

mrdudz:
> I wanted to PR the wiki, but I can't figure how to do pull requests there, it works weirdly. You can see the diff at:
https://github.com/cc65/wiki/compare/master...colinleroy:add-exports-to-dbgfile?expand=1

The goal is to move (most of) the documentation from the wiki into the sgml based docs (the PR stuff being so ackward is one of the reasons). It would be great if you'd do this in this PR.

I can't comment on the PR itself, i have not worked with or looked at the debug info myself. Waiting for someone else to comment :)

That said, it's always nice to have test cases... but i am not sure either how to make them for this case :)

colinleroy:
> The goal is to move (most of) the documentation from the wiki into the sgml based docs (the PR stuff being so ackward is one of the reasons). It would be great if you'd do this in this PR.

Thanks, I'll probably expand the SGML documentation then :)

> That said, it's always nice to have test cases... but i am not sure either how to make them for this case :)

I'm not sure either, I suppose you'll have to build something (with -g -Wl --dbgfile,"program.dbg"), with and without the patch, and compare the outputs, maybe navigate the map with `dbgsh`. 

The main difference, apart from the format, is the new presence of exports like _strcpy, _strdup, _ser_install, ptr1, etc.
It makes following the calls much easier.

bbbradsmith:
This sounds good to me, the only concern I'd have is whether it's backward compatible, i.e. all the old information is still in the same place, and the added things are all optional, and also meaning that the dbginfo tool can still parse old dbg files as well. Is that the case, or does the change require invalidation of old dbg files and parsers?

colinleroy:
It's not. The old exp field in a sym line referenced a symbol id, now an export id.
I can make it retrocompatible but that would be a bit ugly. From what I understood from the wiki, the goal of providing an interface was to be able to change the file format, so I didn't give retrocompatibilty too much thought, I preferred to go with something more logical.
Do you think that's important?

bbbradsmith:
Well, in practice you have tools like [Mesen](https://mesen.ca/) which load and parse dbg files in their current format. If you invalidate the format, it means Mesen now can't load dbg made with newer cc65, or it updates and can't load dbg made with older cc65, or it has to implement two mutually incompatible versions of dbginfo at once.

Even though we have an interface tool, that doesn't cause everything else using it to move in lock-step. Not everybody can change cc65 versions. Not everyone can change Mesen versions. Sometimes we have an old build and debug information, but no source code. Etc.

So, that's why I think it's important. I think a lot of people are using this.

In your second post, though, it looks like you add an "exp" field to "sym" without removing anything else, and then a new "exp" data type? It seems to me that this should already be backwards compatible. Can we not just ignore the new "exp" reference and get all of the old information still? What would make it ugly?

Also, at the very least your PR must change the version given in the file, whether or not it's a backward compatible change.

colinleroy:
> So, that's why I think it's important. I think a lot of people are using this.

I completely understand. Let's find a way to be compatible then! 

> In your second post, though, it looks like you add an "exp" field to "sym" without removing anything else, and then a new "exp" data type? It seems to me that this should already be backwards compatible. Can we not just ignore the new "exp" reference and get all of the old information still? What would make it ugly?

I fact I don't add it, it existed. But exp=12 in a sym type=imp line pointed to another sym ID 🙂
So what I can do is leave this as is, effectively add a field with a different name. Something like 
sym id=12,exp=10,nodbgexp=5? 
And add the Exports as 
export id=... instead of exp id=... 

If documented, that would not be too illogical, what do you think? 

> Also, at the very least your PR must change the version given in the file, whether or not it's a backward compatible change.

Good catch 🙂

colinleroy:
In fact, maybe I could even completely leave the sym lines alone, and just add the new exports category?

kugelfuhr:
Just to understand the problem correctly: What information are you missing? 

If I compile a simple test program

```C
#include <string.h>
int main()
{
    char* a;
    char* b;
    strncpy(a, b, 23);
    return 0;
}
```

with debug information, this is what I get about strncpy:

```
csym    id=0,name="strncpy",scope=0,type=0,sc=ext,sym=6
sym     id=6,name="_strncpy",addrsize=absolute,scope=0,def=7,ref=18,type=imp,exp=209
sym     id=209,name="_strncpy",addrsize=absolute,size=66,scope=26,def=414,ref=433,val=0x8EF,seg=0,type=lab
```

So the C symbol strncpy points to the assembler symbol 6, which is a 16 bit import. The import references symbol 209 as the matching export, which contains 0x8EF as value and 66 as size.

As far as I can say, all the information is there, it just needs to get linked together. An even this is done for you by the routines in dbginfo.c:

```C
/* Notes:
**  - If the symbol is segment relative, the segment id gives segment
**    information, otherwise it contains CC65_INV_ID.
**  - If the type is CC65_SYM_IMPORT, export_id may contain the id of the
**    export. This is not the case if the module contaiing the export doesn't
**    have debug information.
**  - For an import, the fields symbol_value and segment_id are taken from
**    the export, if it is available, since imports have no value or segments
**    by itself.
**  - For an import symbol_size doesn't have a meaning.
**  - For normal symbols (not cheap locals) parent_id contains CC65_INV_ID,
**    for cheap locals it contains the symbol id of the parent symbol.
*/
```

So the module already places the segment and value of the export into the information passed back for an import. It doesn't pass back the size, which is understandable, because _strncpy being 66 bytes in size doesn't mean anything for a module that imports it. And if you need the information, just lookup the linked export.

For me it looks like you're trying to solve a non-problem. Is this a misunderstanding on my side?

colinleroy:
Hmm, this is quite weird. I don't have those. Do I compile wrongly ?
```
$ grep sym.*strncpy prog.dbg 
csym	id=223,name="strncpy",scope=56,type=0,sc=ext,sym=1624
csym	id=458,name="strncpy",scope=110,type=0,sc=ext,sym=2197
sym	id=1624,name="_strncpy",addrsize=absolute,scope=56,def=7663,ref=4309+4964+6969+8333,type=imp
sym	id=2197,name="_strncpy",addrsize=absolute,scope=110,def=12943,ref=12674,type=imp
```
I compile with `cl65 --debug-info -Wl --dbgfile,"prog.dbg" -o prog.bin <list of c files>`

kugelfuhr:
Did you compile the libraries with debug information enabled? Obviously the linker cannot add information to the debug info file it doesn't have. This is true for every other compiler out there.

For some unknown reason the cc65 developers decided that the libraries are built and delivered without such information. If you want debug information, you have to rebuild them using the `-g` switch.

colinleroy:
Indeed, this reaches exactly the goal I wanted:
```
--- a/libsrc/Makefile
+++ b/libsrc/Makefile
@@ -131,8 +131,8 @@ $(TARGETS): | ../lib
 
 else # TARGET
 
-CA65FLAGS =
-CC65FLAGS = -Or -W error
+CA65FLAGS = -g
+CC65FLAGS = -g -Or -W error
 CPUFLAGS =
 
 EXTZP = cbm510 \
```
Shall I close this PR ? Or does someone think it's still good to list exports with no debug symbol associated?

kugelfuhr:
Not having debug information if debug information is not enabled is "works as expected", so don't see a reason for any changes.

My suggestion would therefore be to close this PR but add another that builds the library sources with debug information.

colinleroy:
Indeed. I think I got side tracked by the fact that lbl files have exports without debug info, but adding the relevant flag is much better (especially since it even adds internal asm labels).

colinleroy:
(@kugelfuhr oh thanks, the correct way (-g) is so much better :heart_eyes: 

image

kugelfuhr:
You're welcome :smile:

PR #2170, opened 2023-08-26
Build libs with debug symbols

colinleroy:
Following PR #2169, I have been made aware that I wrote a thousand lines of code for a suboptimal result, where obviously the better solution was to build platform libs with debug info.
So I'm opening this PR to suggest building them with debug info. Maybe it has drawbacks I'm not aware of, but if there isn't any, I think it would be a good thing!


mrdudz:
certainly a good idea :)

PR #2174, opened 2023-09-02
Conio: 65c02 optimisations

colinleroy:
cputc: -2 cycles per char, -2 cycles per carriage return
cputs: -5 cycles per char
vcprintf: -6 cycles per char

### DIFF
--- a/libsrc/apple2/cputc.s
+++ b/libsrc/apple2/cputc.s
@@ -14,6 +14,8 @@
 
         .include        "apple2.inc"
 
+        .macpack        cpu
+
         .segment        "ONCE"
 
         .ifdef  __APPLE2ENH__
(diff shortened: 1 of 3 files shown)


colinleroy:
BTW I have noticed some of those optimizations in existing sources use .cpu .bitand CPU_ISET_65SC02 instead of 65C02. Is there any difference and should I update this for 65SC02 instead of 65C02 ?

polluks2:
The 65SC02, S like simplified, is a variant of the WDC 65C02 without bit instructions.

colinleroy:
> The 65SC02, S like simplified, is a variant of the WDC 65C02 without bit instructions.

Thanks - I've updated the PR to check for 65SC02.

mrdudz:
Pardon my ignorance... is the Apple lib used for different targets with different CPUs?

colinleroy:
It's used for two generations of apple 2, the old ones with a 6502 and the newer ones with a 65c02 (and more hardware capabilities like 80 columns), yes.

PR #2175, opened 2023-09-02
RFC: Apple2: Handle IRQs without ProDOS

colinleroy:
This makes IRQ handling very much faster on apple2enh.

**Tests:**
Tested with serial and mouse drivers:
- target `-t apple2` on emulated apple2e, apple2ee, apple2ep, apple2c0, apple2c4, apple2gs* and real apple2c0
- target `-t apple2enh` on emulated apple2ee, apple2ep, apple2c0, apple2c4, apple2gs* and real apple2c0

I couldn't test on apple2 and apple2p due to MAME having issues with the ROMs.
*(serial not tested on apple2gs as it's not the same chip)

**Numbers:**
Previous system had about:
- 248 cycles per ProDOS IRQ handler ($BFEB)
- 42 cycles per callirq call
- 62 cycles per ser_irq call

For a total of 352 cycles per ACIA interrupt (606 if triggered while in ROM, as $C803 has about 254 cycles before returning).

New system has:
- 37 cycles for handle_ram_irq
- no change in either callirq or ser_irq, as this isn't the scope of this patch

For a total of 141 cycles per ACIA interrupt (395 if triggered while in ROM).

**Callgrind profile before:**
image

**Callgrind profile after:**
image

**Why:**
I'm on a quest to be able to use the serial port at 19200bps, and the existing IRQ handling is much too slow for this to succeed: another byte arrives before the IRQ for the previous one has returned. (This page shows how long it takes per byte to arrive: https://lucidar.me/en/serialib/most-used-baud-rates-table/)

**Results of this patch:**
With this patch, I am able to read/write in a sustained manner at 19200bps, **as long as RAM is banked in** when an IRQ arrives. If ROM is banked in, the added cost of the ROM IRQ handler, before execution is passed to the $03fe/$03ff vector, is still too much to sustain 19200 bps. But this means if one is cautious about how/when they receive bytes, double the speed can be achieved.

**Caveats:**
I am quite sure this patch is kind of naive, and there are probably corner cases that should be taken into account when entering the new IRQ handlers. I'd gladly improve the patch if it is necessary and it is deemed interesting to consider merging.

Thanks!

### DIFF
--- a/asminc/apple2.inc
+++ b/asminc/apple2.inc
@@ -25,6 +25,11 @@ BRKVec  :=      $03F0   ; Break vector
 SOFTEV  :=      $03F2   ; Vector for warm start
 PWREDUP :=      $03F4   ; This must be = EOR #$A5 of SOFTEV+1
 
+ROMIRQVEC :=      $03FE   ; IRQ handler vector, called when an IRQ is fired
+                          ; and we're banked on ROM
+
+RAMIRQVEC :=      $FFFE   ; IRQ handler vector, called when an IRQ is fired
+                          ; and we're banked on RAM
 ;-----------------------------------------------------------------------------
 ; Hardware
 
--- a/libsrc/apple2/irq.s
+++ b/libsrc/apple2/irq.s
@@ -1,5 +1,6 @@
 ;
 ; Oliver Schmidt, 2012-11-17
+; Colin Leroy-Mira, 2023-09-02 - Handle IRQs without ProDOS
 ;
 ; IRQ handling (Apple2 version)
 ;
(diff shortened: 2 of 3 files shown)


colinleroy:
Let me add a comment about the different situations we can find ourselves in when an IRQ fires, because memory banking has been a concern on Apple2Infinitum's slack:

**Standard setup:**

"The vector at $FFFE in the ROM points to Apple lie’s built-in interrupt handling routine. When you initialize the mouse or serial communication firmware, copies of 
the ROM’s interrupt vector are placed in the interrupt vector addresses in both main and auxiliary bank-switched memory. [...]
"The built-in interrupt handling routine records the current memory configuration, then sets up its own standard memory configuration"
"passes them to a user’s interrupt handling routine whose address is stored at $03FE and $03FF of main memory".

On the Apple 2c for example, FFFE/FFFF point to $C803.

ProDOS then registers its handler into $3FE/3FF and it points to $BFEB.

so when an interrupt fires in the standard configuration, we have:
jmp ($fffe) => jmp $C803 => jmp ($3FE) => jmp $BFEB => callirq => our IRQ functions

and at the time callirq is called, we're setup to a known memory banking configuration that is triggered by ProDOS RR $c08b: read from RAM, write to RAM, LC bank 1. (it's the first thing it does, at $BFEB and $BFEE)

After this we go back to where we were, which in a cc65 program, is either c080 (read from RAM, write to none, LC bank 2) or c082 (read from ROM, write to none, LC bank 2).

**What this patches changes:**

1) it removes BFEB from 03FE and patches in handle_rom_irq instead, which only jsr to callirq.
2) it removes C803 from FFFE, __in the main RAM only__, and patches in handle_ram_irq instead, which saves registers, checks for BRK and jsr to callirq.

a) What happens when an IRQ fires and we're in ROM:
jmp ($fffe) => jmp $C803 => jmp ($3FE) => jmp handle_rom_irq => callirq => our IRQ functions
Is it bad that we're in ROM? No, because $C803 banks back to main RAM.

b) What happens when an IRQ fires and we're in main RAM:
jmp ($fffe) => jmp handle_ram_irq => callirq => our IRQ functions
Is it bad that we could be in LC1? No, because handle_rom_irq, callirq and our IRQ functions are all in LOWCODE or CODE, so always banked in.

c) What happens when an IRQ fires and we're in aux RAM:
jmp ($fffe) => jmp $C803 => jmp ($3FE) => jmp handle_rom_irq => callirq => our IRQ functions
Is it bad that we're in AUX RAM? No, because $C803 banks back to main RAM.

Finally, AppleIIgs always jumps to main RAM's ($fffe), so in case of a IIgs, we don't patch $fffe at all and we're always in case A or C. It's slower, but safe.

colinleroy:
Closing this PR after discussion with @oliverschmidt : It is a bit too hacky to be deemed safe, and only applies to a small subset of Apple 2s.

Issue #2179, opened 2023-09-04
Commit 2431fb7d85b8aa74153e6e8343bb577e28cc8832 breaks AppleSingle binaries


colinleroy:
Maybe this is an issue for AppleCommander... But 2431fb7d85b8aa74153e6e8343bb577e28cc8832 from PR #2163 breaks copying AppleSingle binaries to disk images:
```
java.lang.ArrayIndexOutOfBoundsException: arraycopy: last source index 10928 out of bounds for byte[10899]
	at java.base/java.lang.System.arraycopy(Native Method)
	at io.github.applecommander.applesingle.AppleSingleReader.readAt(AppleSingleReader.java:32)
	at io.github.applecommander.applesingle.Entry.create(Entry.java:29)
	at io.github.applecommander.applesingle.AppleSingle.asEntries(AppleSingle.java:188)
	at io.github.applecommander.applesingle.AppleSingle.asEntries(AppleSingle.java:176)
	at io.github.applecommander.applesingle.AppleSingle.read(AppleSingle.java:159)
	at io.github.applecommander.applesingle.AppleSingle.read(AppleSingle.java:147)
	at com.webcodepro.applecommander.ui.ac.putAppleSingle(ac.java:374)
	at com.webcodepro.applecommander.ui.ac.putAppleSingle(ac.java:366)
	at com.webcodepro.applecommander.ui.ac.main(ac.java:168)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:104)
	at java.base/java.lang.reflect.Method.invoke(Method.java:577)
	at org.springframework.boot.loader.MainMethodRunner.run(MainMethodRunner.java:49)
	at org.springframework.boot.loader.Launcher.launch(Launcher.java:108)
	at org.springframework.boot.loader.Launcher.launch(Launcher.java:58)
	at org.springframework.boot.loader.JarLauncher.main(JarLauncher.java:88)
```
@oliverschmidt, what do you think? is #2163 wrong for Apple2 or is AppleCommander expecting things it shouldn't ?
Thanks! 

PS: The binaries generated with 2431fb7d85b8aa74153e6e8343bb577e28cc8832 crash on real hardware after being transferred over serial (a tried and working process for me). They're also 156 bytes bigger.

bbbradsmith:
Could you give more information about what command you are running here on what inputs?

colinleroy:
[...]

PR #2181, opened 2023-09-05
change INIT back to rw for configurations that needed it

bbbradsmith:
This is a partial revert of #2163 based on the issue raised in #2179.

I had believed incorrectly that placing a `bss` segment into a memory region that outputs to file would either emit fill bytes, or give an error (or at least a warning). However, what it actually does is not emit those bytes, and not give any error or warning about it.

After reading the example given in the [ld65 documentation](https://cc65.github.io/doc/ld65.html#ss5.2), I realized that this behaviour was explicitly intended.

Because we apparently want to be able to assign `bss` segments to memory regions that output to a file, even though they won't be allowed to emit bytes to that file, it cannot give an error, or even a warning, because it is valid input. This was unexpected and unintuitive to me, but nevertheless it was my mistake for not catching this before submitting #2163. I am used to explicitly using `file=""` for memory regions like `RAM`, and hadn't thought it was possible or desirable to just implicitly keep it empty in this manner.

Only the cases where INIT appeared to be reserving space for a file were reverted, and I have added a comment to each to clarify this is the reason they are `rw` and not `bss`.

### DIFF
--- a/cfg/apple2-hgr.cfg
+++ b/cfg/apple2-hgr.cfg
@@ -27,7 +27,7 @@ SEGMENTS {
     CODE:     load = MAIN,           type = ro   start    = $4000;
     RODATA:   load = MAIN,           type = ro;
     DATA:     load = MAIN,           type = rw;
-    INIT:     load = MAIN,           type = bss;
+    INIT:     load = MAIN,           type = rw;  # uninitialized, but reserves output space
     ONCE:     load = MAIN,           type = ro,  define   = yes;
     LC:       load = MAIN, run = LC, type = ro,  optional = yes;
     BSS:      load = BSS,            type = bss, define   = yes;
--- a/cfg/apple2-overlay.cfg
+++ b/cfg/apple2-overlay.cfg
@@ -43,7 +43,7 @@ SEGMENTS {
     CODE:     load = MAIN,           type = ro;
     RODATA:   load = MAIN,           type = ro;
     DATA:     load = MAIN,           type = rw;
-    INIT:     load = MAIN,           type = bss;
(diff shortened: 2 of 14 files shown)


bbbradsmith:
The check failure appears to be some unrelated github service outage?

colinleroy:
Thank you @bbbradsmith!
for the check failure I believe you can trigger the jobs again with git push  - -force.

mrdudz:
or you can click on "actions", pick the one that failed, and then "rerun jobs" on the top/right

mrdudz:
@bbbradsmith btw, it would be great if you could enhance the documentation on the special segments a bit. Ie while you still have it in your head, read over the descriptions once more and add whatever info you were missing before :)

bbbradsmith:
The relevant documentation was here: https://github.com/cc65/wiki/wiki/Segment-usage-of-constructors It should probably be transferred to cc65-intern, though probably a note about why it's not bss should be incorporated as well.

> or you can click on "actions", pick the one that failed, and then "rerun jobs" on the top/right
You can, but I can't. That option does not appear unless you have higher permissions than an outside contributor.

A force push does work... or rather two force pushes would work, because I'd have to make a change to be able to push, and then change it back.

mrdudz:
> The relevant documentation was here: https://github.com/cc65/wiki/wiki/Segment-usage-of-constructors It should probably be transferred to cc65-intern, though probably a note about why it's not bss should be incorporated as well.

Urks, yes. As said before, a lot of stuff that is now hiding in the wiki should get moved to the documentation.

> You can, but I can't. That option does not appear unless you have higher permissions than an outside contributor.

Oh, ok. too bad. Close and reopen the ticket should also do the trick, i think

PR #2182, opened 2023-09-06
Add ntohs/htons and ntohl/htonl

colinleroy:


mrdudz:
32 bit values are passed in A/X/sreg/sreg+1

I usually just make an empty function in C with the signature i want, and then compile it and have a look :)

colinleroy:
> I usually just make an empty function in C with the signature i want, and then compile it and have a look :)

Thank you, that's exactly what I did :)

oliverschmidt:
* The prototypes belong into arpa/inet.h
* The two implementations should go into separate files. There's no point in having users of the 'short' variant link all the code of the 'long' variant.
* I think it would be more elegant to use linker aliases than C macros for "the other way round". See https://github.com/cc65/cc65/blob/master/libsrc/apple2/libref.s for examples.

colinleroy:
Thanks, I addressed your comments. Also I stopped brainfarting and made the long swap much cleaner/faster 🙂

doc/funcref.sgml
```diff
@@ -4388,6 +4398,45 @@ to undefined behaviour.
 </descrip>
 </quote>
 
+<sect1>htonl<label id="htonl"><p>
+
+<quote>
+<descrip>
+<tag/Function/Swaps byte order in a 32 bit word.
+<tag/Header/<tt/<ref id="stdlib.h" name="stdlib.h">/
```
oliverschmidt, line 4406: Missing header adjustment.

doc/funcref.sgml
```diff
@@ -4388,6 +4398,45 @@ to undefined behaviour.
...
+</descrip>
+</quote>
+
+
+<sect1>htons<label id="htons"><p>
+
+<quote>
+<descrip>
+<tag/Function/Swaps byte order in a 16 bit word.
+<tag/Header/<tt/<ref id="stdlib.h" name="stdlib.h">/
```
oliverschmidt, line 4426: Missing header adjustment.

doc/funcref.sgml
```diff
@@ -5757,6 +5806,44 @@ memory allocated for the driver.
 </descrip>
 </quote>
 
+<sect1>ntohl<label id="ntohl"><p>
+
+<quote>
+<descrip>
+<tag/Function/Swaps byte order in a 32 bit word.
+<tag/Header/<tt/<ref id="stdlib.h" name="stdlib.h">/
```
oliverschmidt, line 5814: Missing header adjustment.

doc/funcref.sgml
```diff
@@ -5757,6 +5806,44 @@ memory allocated for the driver.
...
+<tag/Availability/cc65
+</descrip>
+</quote>
+
+<sect1>ntohs<label id="ntohs"><p>
+
+<quote>
+<descrip>
+<tag/Function/Swaps byte order in a 16 bit word.
+<tag/Header/<tt/<ref id="stdlib.h" name="stdlib.h">/
```
oliverschmidt, line 5833: Missing header adjustment.

libsrc/common/ntohl.s
```diff
@@ -0,0 +1,34 @@
...
+;
+; int __fastcall__ ntohl (long val);
+;
+
+.export     _ntohl, _htonl
+.import     popa
+.importzp   tmp1, tmp2, sreg
+
+_htonl := _ntohl
+.code
```
oliverschmidt, line 12: The predominant style is to leverage the fact that the assembler defaults to `.code`.

libsrc/common/ntohs.s
```diff
@@ -0,0 +1,18 @@
...
+;
+; int __fastcall__ ntohs (int val);
+;
+
+.export     _ntohs, _htons
+.importzp   tmp1
+
+_htons := _ntohs
+
+.code
```
oliverschmidt, line 12: The predominant style is to leverage the fact that the assembler defaults to `.code`.

colinleroy:
Thanks, should be better now.

kugelfuhr:
At least for ntohs/htons, it might be an option to use a macro:

```C
#define ntohs(x)                \
    (                           \
        __AX__=(x),             \
        asm("sta tmp1"),        \
        asm("txa"),             \
        asm("ldx tmp1"),        \
        __AX__                  \
    )
```

The code above is 5 bytes. A call is three bytes plus ntohs as a subroutine (6 bytes), making 9 bytes total.

One might also use `__OPT_i__` to either use the macro or call a subroutine.

colinleroy:
Something like that? :)

include/arpa/inet.h
```diff
@@ -31,8 +31,22 @@
 #ifndef _ARPA_INET_H
 #define _ARPA_INET_H
 
+#ifndef __OPT_i__
```
kugelfuhr, line 34: `#if (__OPT_i__ < 200)`

`__OPT_i__` contains the argument to `--codesize` or just 200 if `-Oi` was given.

colinleroy:
Thanks!

include/arpa/inet.h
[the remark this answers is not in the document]
oliverschmidt: Sorry for the nitpicking... but I was actually referring to the whole thing:
* Number of new lines
* "Code" header
* Comment on the final `#endif`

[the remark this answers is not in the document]
colinleroy: Nitpicking is OK, it makes a code base legible. I'll adapt :)

include/arpa/inet.h
```diff
@@ -0,0 +1,65 @@
...
+        asm("ldx tmp1"),        \
+        __AX__                  \
+    )
+#define htons(x) ntohs(x)
+
+#endif
+
+long __fastcall__ ntohl (long val);
+long __fastcall__ htonl (long val);
+
```
oliverschmidt, line 63: Two more new lines and we're there...

colinleroy: Right! sorry. Should be good now :)

oliverschmidt (approved):

mrdudz:
OK :)

PR #2185, opened 2023-09-07
Apple2: fail opening serial port on IIgs

colinleroy:
The Pascal Firmware Protocol Bytes ID is not enough to differentiate an SSC card from a IIgs serial firmware:
http://www.1000bit.it/support/manuali/apple/technotes/misc/tn.misc.08.html

Loading a2(e).ssc.ser on a IIgs succeeds, then goes to limbo when one tries to use the serial port.

Check first byte on the slot's firmware, as it's not supposed to be $EF (65C816 sep instruction) on an SSC card, and it is on the IIgs' serial firmware (at least ROM revisions 0, 1, 3).

### DIFF
--- a/libsrc/apple2/ser/a2.ssc.s
+++ b/libsrc/apple2/ser/a2.ssc.s
@@ -132,16 +132,33 @@ ParityTable:
         .byte   $60             ; SER_PAR_EVEN
         .byte   $A0             ; SER_PAR_MARK
         .byte   $E0             ; SER_PAR_SPACE
+
+        ; (*): The first byte we'll check at offset 0 is
+        ; expected to be a BIT instruction on any Apple2
+        ; machine with an ACIA 6551, including the //c
+        ; and //c+.
+        ; The IIgs, on the other hand, has a
+        ; Zilog Z8530 chip and its firmware starts with
+        ; a SEP instruction. We don't want to load this
+        ; driver on the IIgs' serial port.
+        ;
+        ; The next four bytes we check are the Pascal
+        ; Firmware Protocol Bytes that identify a
+        ; serial card.
+
 IdOfsTable:
+        .byte   $00             ; First instruction
         .byte   $05             ; Pascal 1.0 ID byte
         .byte   $07             ; Pascal 1.0 ID byte
         .byte   $0B             ; Pascal 1.1 generic signature byte
         .byte   $0C             ; Device signature byte
 IdValTable:
-        .byte   $38             ; Fixed
-        .byte   $18             ; Fixed
-        .byte   $01             ; Fixed
-        .byte   $31             ; Serial or parallel I/O card type 1
+        .byte   $2C             ; BIT
+        .byte   $38             ; ID Byte 0 (from Pascal 1.0), fixed
+        .byte   $18             ; ID Byte 1 (from Pascal 1.0), fixed
+        .byte   $01             ; Generic signature for Pascal 1.1, fixed
+        .byte   $31             ; Device signature byte (serial or
+                                ; parallel I/O card type 1)
 
 IdTableLen      = * - IdValTable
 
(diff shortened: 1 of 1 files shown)


colinleroy:
The relevant ROM dumps:
```
apple2c0
C200  2c 89 c1 70 14 38 90 18  b8 50 0e 01 31 11        |,..p.8...P..1.|

apple2c3
C200  2c 89 c1 70 14 38 90 18  b8 50 0e 01 31 11        |,..p.8...P..1.|

apple2c4
C200  2c 89 c1 70 14 38 90 18  b8 50 0e 01 31 11        |,..p.8...P..1.|

apple2cp
C200  2c 89 c1 70 14 38 90 18  b8 50 0e 01 31 11        |,..p.8...P..1.|

apple2ee
C200  2c 58 ff 70 0c 38 90 18  b8 50 06 01 31 8e        |,X.p.8...P..1.|

apple2ep
C200  2c 58 ff 70 0c 38 90 18  b8 50 06 01 31 8e        |,X.p.8...P..1.|

apple2gsr0
C200  e2 40 70 25 ea 38 90 18  b8 50 1e 01 31 45        |.@p%.8...P..1E|

apple2gsr1
C200  e2 40 70 25 ea 38 90 18  b8 50 1e 01 31 45        |.@p%.8...P..1E|

apple2gsr3
C200  e2 40 70 27 ea 38 90 18  b8 50 20 01 31 47        |.@p'.8...P .1G|

```

colinleroy:
(this one would be for @oliverschmidt :) )

oliverschmidt:
I don't think that there will be so much new things coming our way. So there's no need to do the "negative check" for $E2. We can as well do the "positive check" for $2C, can't we? The benefit from that would be that there's no need for that "ugly" additional code. Rather it's enough to just insert one value each at beginning of `IdOfsTable` and `IdValTable`. Please do so with a comment explaining that the $2C is present in both the SSC firmware and the //c firmware (-> exactly where the 6551 is present) and not present in the IIgs firmware (-> where the 6551 is not present).

colinleroy:
> I don't think that there will be so much new things coming our way. So there's no need to do the "negative check" for $E2. We can as well do the "positive check" for $2C, can't we? 

You're absolutely right, it's much simpler, cleaner, and also just 2 bytes more. :)

libsrc/apple2/ser/a2.ssc.s
```diff
@@ -132,16 +132,33 @@ ParityTable:
         .byte   $60             ; SER_PAR_EVEN
         .byte   $A0             ; SER_PAR_MARK
         .byte   $E0             ; SER_PAR_SPACE
+
+        ; (*): The first byte we'll check at offset 0 is
+        ; expected to be a BIT instruction on any Apple2
+        ; machine with an ACIA 6551, including the //c
+        ; and //c+.
```
oliverschmidt, line 139: Me nitpicking again...
The BIT instruction isn't an attribute of Apple II machines in general. It's an attribute of the SSC firmware (and the //c firmware).

colinleroy: Yeah, my comment is weirdly worded indeed.

colinleroy:
Thank you! Now I'm ready for my next challenge: try and write a driver for the IIgs' Zilog.
Expect never to hear about this project ever again :laughing:

oliverschmidt:
I thank you for the contribution!

Regarding the IIgs SCC:
https://github.com/ADTPro/adtpro/blob/main/src/client/prodos/serial/iigsscc.asm might be something to have a look at.
https://github.com/ADTPro/adtpro/blob/main/src/client/prodos/serial/iigsscc.asm#L22 doesn't sound that nice.

colinleroy:
> I thank you for the contribution!
> 
> Regarding the IIgs SCC: https://github.com/ADTPro/adtpro/blob/main/src/client/prodos/serial/iigsscc.asm might be something to have a look at. https://github.com/ADTPro/adtpro/blob/main/src/client/prodos/serial/iigsscc.asm#L22 doesn't sound that nice.

I'm definitely NOT looking at ADTPro for this, it's GPL code and I don't think we want GPL in cc65 :)

I'll refer to https://gswv.apple2.org.za/a2zine/Utils/Z8530_SCCsamples_info.txt (public domain) and http://www.applelogic.org/files/Z8530UM.pdf (tech ref).

PR #2188, opened 2023-09-12
Serial kernel: Fix driver lifecycle

colinleroy:
- Make sure we don't install a driver on top of another one
- Make sure we don't uninstall a driver twice
- Make sure we uninstall a driver if needed at exit


mrdudz:
looks ok, thanks!

PR #2196, opened 2023-09-17
Added FX registers to cx16.h

ToboterXP:
Added the fx registers to the VERA struct in cx16.h . No comments explaining them, because there's way too much going on to explain in a brief and precise way.


mrdudz:
thanks!

PR #2198, opened 2023-09-18
[cc65] Fixed check for conflicting extern vs no-linkage/static declarations in functions

### LINKED ISSUE #2162: cc65: Compiler accepts invalid variable declarations (regression)
The following - invalid - code is accepted by the compiler:

```C
int main()
{
    extern  int i;
    int i = 42;
    return i;
}
```

First, there should be a diagnostic because the declaration is invalid. Next, as can be seen from the assembly, the compiler omits the initialization of i completely and treats i as an imported variable, so the compiler will not even create code that corresponds to what the programmer might expect.

Please note that this is a regression. Version 2.18 of the compiler didn't accept the input:

```
test.c(4): Error: Multiple definition for 'i'
``` 

BTW, this is what gcc outputs:

```
test.c: In function ‘main’:
test.c:4:9: error: declaration of ‘i’ with no linkage follows extern declaration
    4 |     int i = 42;
      |         ^
test.c:3:17: note: previous declaration of ‘i’ was here
    3 |     extern  int i;
      |                 ^
```

acqn:
Fixed #2162.

### DIFF
--- a/src/cc65/symtab.c
+++ b/src/cc65/symtab.c
@@ -1225,6 +1225,15 @@ SymEntry* AddLocalSym (const char* Name, const Type* T, unsigned Flags, int Offs
     /* Do we have an entry with this name already? */
     SymEntry* Entry = FindSymInTable (Tab, Name, HashStr (Name));
     if (Entry) {
+        int CheckExtern = 0;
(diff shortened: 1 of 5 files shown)


mrdudz:
Thanks!

PR #2199, opened 2023-09-22
[cc65] Forbidden struct itself with flexible array member as struct member or array element

### LINKED ISSUE #2016: cc65: Compiler accepts structs with flexible members as struct fields
cc65 allows struct fields that are structs with flexible array members:

```c      
typedef struct x {
    int a;
    int b[];    /* Ok: Flexible array member can be last */
} x;
    
struct y {
    x   x;      /* Not ok but accepted: Contains flexible array member */
    int a;
};
```

### LINKED ISSUE #2017: cc65: Compiler allows arrays of structs with flexible array members
The following is illegal according to the standard but accepted by cc65:

```c
struct x {
    int a;
    int c;
    int b[];
};

struct x y[3];          /* Should be an error */
````

acqn:
Fixed #2016.
Fixed #2017.
A structure with a flexible array member shall not be a member of a structure or an element of an array according to the ISO C Standard.

### DIFF
--- a/src/cc65/declare.c
+++ b/src/cc65/declare.c
@@ -515,6 +515,13 @@ static void CheckArrayElementType (Type* DataType)
                 if (!IsTypeVoid (T) || IS_Get (&Standard) != STD_CC65) {
                     Error ("Array of 0-size element type '%s'", GetFullTypeName (T));
                 }
+            } else {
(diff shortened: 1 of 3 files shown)


mrdudz:
nice, thanks!

PR #2200, opened 2023-09-22
[cc65] Fixed compound initialization with certain omitted enclosing curly braces

### LINKED ISSUE #2135: cc65: Error in initialization of compound data
The following legal code causes the compiler to emit invalid data:

```C
struct s { char a[1]; int b; };
struct s x = { 1, 2 };
```

When using fully bracketed initializers, the code would be identical to 

```C
struct s { char a[1]; int b; };
struct s x = { { 1 }, 2 };
```

cc65 version 2.18 tries to read both values as array elements and outputs an error, because it sees two initializers for an array of size one:

```
test.c(2): Error: '{' expected
test.c(2): Error: Too many initializers
test.c(2): Error: '}' expected
```

The current git version translates the code without an error, so it looks like the problem has been fixed but this is wrong - as the assembly shows:

```
_x:
        .byte   $01     ; a[0]
        .byte   $02     ; a[1]
        .res    2,$00   ; b
```

(comments by me).

So while the old version complained about too many initializers, the current version silently reads and outputs all of them, even if there are no corresponding array elements.

acqn:
Fixed #2135.
Fixed compound initialization with omitted enclosing curly braces when an array/struct/union to initialize is nested.

### DIFF
--- a/src/cc65/initdata.c
+++ b/src/cc65/initdata.c
@@ -343,8 +343,8 @@ static unsigned ParseArrayInit (Type* T, int* Braces, int AllowFlexibleMembers)
         /* Char array initialized by string constant */
         int NeedParen;
 
-        /* If we initializer is enclosed in brackets, remember this fact and
-        ** skip the opening bracket.
+        /* If the initializer is enclosed in curly braces, remember this fact
+        ** and skip the opening one.
         */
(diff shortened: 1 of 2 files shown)


mrdudz:
looks ok, thanks!

PR #2203, opened 2023-09-26
Apple2: Always copy cmdline

colinleroy:
Otherwise, the last exec() cmdline lingers around for the next exec("file", NULL).

Example:
- program A does exec("B", "parameter1 parameter2")
- program B does exec("C", NULL)
- program C starts with argc == 3, argv[1] == "parameter1" and argv[2] == "parameter2"


oliverschmidt:
Thanks :-)

PR #2204, opened 2023-09-26
General project fixes and quality of life improvements

vrubleg:
- Fixed a small issue in default VS project settings that I made previous time. `WholeProgramOptimization` was in wrong place and didn't work. Also added a few flags to generate more clean executables without [Rich headers](https://www.ntcore.com/files/richsign.htm).
- Added basic `.editorconfig` so Visual Studio follows some of conventions that are used in this project automatically. Otherwise it's painful to work with this project when code editor constantly wants to follow some other coding style.
- Fixed msbuild.bat to correctly find Visual Studio when it's not in default path.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,16 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = crlf
+indent_style = space
+indent_size = 4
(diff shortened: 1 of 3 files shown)


mrdudz:
thanks!

PR #2205, opened 2023-09-30
Apple2 serial: Cleanup coding style a bit

colinleroy:
@oliverschmidt, you mentioned you'd like the coding style reviewed on the SSC driver :)

### DIFF
--- a/libsrc/apple2/ser/a2.ssc.s
+++ b/libsrc/apple2/ser/a2.ssc.s
@@ -38,8 +38,8 @@
         .endif
 
         ; Driver signature
-        .byte   $73, $65, $72           ; "ser"
-        .byte   SER_API_VERSION         ; Serial API version number
+        .byte   $73, $65, $72   ; "ser"
+        .byte   SER_API_VERSION ; Serial API version number
 
         ; Library reference
         .addr   $0000
@@ -59,16 +59,16 @@
 ; I/O definitions
 
 .if (.cpu .bitand CPU_ISET_65C02)
-ACIA            = $C088
+ACIA           := $C088
 .else
 Offset          = $8F           ; Move 6502 false read out of I/O to page $BF
-ACIA            = $C088-Offset
+ACIA           := $C088-Offset
 .endif
 
-ACIA_DATA       = ACIA+0        ; Data register
-ACIA_STATUS     = ACIA+1        ; Status register
-ACIA_CMD        = ACIA+2        ; Command register
-ACIA_CTRL       = ACIA+3        ; Control register
+ACIA_DATA      := ACIA+0        ; Data register
+ACIA_STATUS    := ACIA+1        ; Status register
+ACIA_CMD       := ACIA+2        ; Command register
+ACIA_CTRL      := ACIA+3        ; Control register
 
 ;----------------------------------------------------------------------------
 ; Global variables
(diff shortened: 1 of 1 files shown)


oliverschmidt:
Thanks :-)

mrdudz:
> Great! Thanks :-)
> 
> Just one minor thing: The `return value is char` doesn't start with an upper case letter. However, I don't see the point in those comments anyhow. So I suggest to remove them altogether.

Sorry to be an ass - but sticking to the codestyle described in contributing.md is not optional - the comment is mandatory. Please add it again.

oliverschmidt:
So the rule really includes the comment? I'm not the maintainer of this project anymore, but nevertheless I'd like to state that I consider the comment close to the worst possible one. The fact that the return value is char would be a reason to **_not_** set X. The comment needs to read something like:

* Although return value is char
* Promote char return value

Just my two cents...

@colinleroy: Sorry for asking you to do a change that conflicts with the current coding style :-( I hope that it is acceptable to have a consistent first letter case ;-)

colinleroy:
Not a problem. I'll make another pr to put them back if needed. I don't have a strong opinion on the matter so I'll wait to see what comorimise/solution you and @mrdudz come with 🙂

oliverschmidt:
1. I don't think that https://github.com/cc65/cc65/blob/master/Contributing.md makes it clear that the comment is obligatory.
2. Please advise on using `Return value is char` instead of `return value is char`.
3. After all I'm still a core team member. Whatever discussion happened when where, excluded me.
4. I'm neither interested in a discussion here now nor in creating a PR.

mrdudz:
> I don't think that https://github.com/cc65/cc65/blob/master/Contributing.md makes it clear that the comment is obligatory.

Its in the first paragraph

oliverschmidt:
It seems you are really keen on creating a problem here :-(

> Its in the first paragraph

What is in which first paragraph?

mrdudz:
> What is in which first paragraph?

The first paragraph of the file:

_"The word "must" indicates a requirement. The word "should" indicates a recomendation."_

oliverschmidt:
That's not the point. The point is that it isn't clear that
```
    lda #RETURN_VALUE
    ldx #0
 ```
doesn't already fulfill the rule. And if the above isn't enough, then it's unclear if
```
    lda #RETURN_VALUE
    ldx #0 ; Return value is char
 ```
fulfills the rule, or if 
```
    lda #RETURN_VALUE
    ldx #$00 ; return value is char
 ```
fulfills the rule.

If you come up with these type of rules, you need to be much more precise. I.e. using `#$00` instead of `#0` is a style use in quite some places in the code base.

mrdudz:
_This must be done in one of the following ways:_, followed by examples, makes it pretty clear to me at least. The first of your examples does not qualify, the others do.

Please don't start nitpicking like in those fruitless discussions with Greg, that basically were the reason for creating the file. It surely isn't complete, nor perfect, we will never agree on all of it, and no doubt it can be improved - hence the "if something isn't clear, please ask" also in the first paragraph. However, so far the number of questions i received regarding these rules is exactly zero, so there was very little reason to do so. I'm not not planning to proactively work those out to be bulletproof to all nitpicking either, as that lowers the chance anyone would read - or even respect - them even more.

To come back to this "problem" - the important part there is that the comment (or the assert) exists. No more no less. I'm not insisting on a particular wording or spelling either for that matter, as long as there is a hint that tells why this lonely ldx exists.

oliverschmidt:
So maybe there's just a misunderstanding. Are you saying that there only needs to be a suitable comment? So would the comment `Promote char return value` instead of `return value is char` be acceptable?

mrdudz:
Definitely!

oliverschmidt:
I see. Then I'm sorry for my part in the misunderstanding. I can totally see the point in obligating a suitable comment.

Thanks for the clarification :-)

oliverschmidt:
@colinleroy Would you please add the comment `Promote char return value` to all places in question?

colinleroy:
I will, and will use the same in the IIgs pr too. Thanks for the clarification guys!

mrdudz:
Perhaps also in codestyle.md then? :)

mrdudz:
BTW i just skimmed over the code, and noticed there are hardware registers being accessed by indexed instructions.... i am not familiar with the ACIA, but are you aware of the dummy accesses of the 6502? This can produce really weird and hard to debug problems with hardware registers sometimes. (Just thought i'd mention this, it might not be a problem in this case)

oliverschmidt:
> Perhaps also in contributing.md then? :)

Sorry, can you please elaborate?

> ... are you aware of the dummy accesses of the 6502?

The ACIA in question was only used in a machine with a 65816. However, thanks for pointing out :-)

colinleroy:
> BTW i just skimmed over the code, and noticed there are hardware registers being accessed by indexed instructions.... i am not familiar with the ACIA, but are you aware of the dummy accesses of the 6502? This can produce really weird and hard to debug problems with hardware registers sometimes. (Just thought i'd mention this, it might not be a problem in this case)

Yes, we should be in the clear on the IIgs 🙂

mrdudz:
> > Perhaps also in contributing.md then? :)
> 
> Sorry, can you please elaborate?

i meant, change that comment to what you suggested, in the example in contributing.md

colinleroy:
> i meant, change that comment to what you suggested, in the example in contributing.md

Will do.

PR #2206, opened 2023-10-03
Apple2: Re-introduce necessary comments, wit a clearer wording

colinleroy:
Following discussion on #2205


oliverschmidt:
Thanks for putting up with the - as it turned out at the end mostly unnecessary - internal discussion ;-)

PR #2209, opened 2023-10-05
ca65: `.LINECONT` as a part of `.FEATURE`

### LINKED ISSUE #2133: ca65: `.LINECONT` as a part of `.FEATURE`
`.FEATURE` is used for tweaking how the code is parsed a lot, but for some reason the `.LINECONT` is a separate control command, even though it changes a setting that is stored and semantically close to the other `.FEATURE` settings. I propose add support of changing `.LINECONT` through `.FEATURE` for consistency with other settings. I can make a PR with this change.

vrubleg:
Added `.FEATURE line_continuations +` that fixes #2133.

### DIFF
--- a/doc/ca65.sgml
+++ b/doc/ca65.sgml
@@ -2867,6 +2867,26 @@ See: <tt><ref id=".ASCIIZ" name=".ASCIIZ"></tt>,<tt><ref id=".CHARMAP" name=".CH
     overridden. When using this feature, you may also get into trouble if
     later versions of the assembler define new keywords starting with a dot.
 
+  <tag><tt>line_continuations</tt><label id="line_continuations"></tag>
+
+    Switch on or off line continuations using the backslash character
+    before a newline. The option is off by default.
+    Note: Line continuations do not work in a comment. A backslash at the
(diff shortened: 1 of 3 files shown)


vrubleg:
It's rater a feature than a bug. All these settings affecting syntax are defined together and it was just inconsistent that all of them except `.LINECONT` were controlled by `.FEATURE`.

mrdudz:
OK

In any case - is this something that is often used? Somehow i'd prefer to remove the .linecont directive (and we are historically not afraid of breaking changes) instead of having both.

mrdudz:
PS: or at least print a warning (".linecont is deprecated, use .feature line_continuations") when .linecont is seen

vrubleg:
Analyzed a bit more. Seems like some other similar settings also can be changed like this (`.CASE`, `.AUTOIMPORT`, `.SMART`). So most of them can be changed through `.FEATURE` in a single universal way (that allows to change a few settings in one line if desired), but others have a dedicated directive or can be changed through command line options.

Later when I have more free time I'll make a table of settings and how they can be changed, so we could figure out how to make them more consistent or decide to keep it as is :)

PR #2212, opened 2023-10-07
Remove unused abc_score_ptr* zp entries for lynx lib

BrianPeek:
This wastes 8 bytes in the zp for no reason...


mrdudz:
Mmmh, why was this even there?

BrianPeek:
I would wager it's left over from the ABC music implementation:
https://atarilynxdev.net/doku.php/lynx:abcmusic

I'm assuming it was part of the lynx.lib at one point in the past?

mrdudz:
I have no idea....

anyway, it's not used anywhere in the tree, so can be safely removed. thanks!

PR #2213, opened 2023-10-07
Apple2: Change byte 0 test to "different than $E2"

colinleroy:
Remember PR #2185 ? We decided to do a positive check because it was cleaner and shorter... But we didn't think of the clones. Some of those don't have their "SSC" firmware starting with a BIT instructions apparently - cf https://github.com/colinleroy/a2tools/issues/1

### DIFF
--- a/libsrc/apple2/ser/a2.ssc.s
+++ b/libsrc/apple2/ser/a2.ssc.s
@@ -141,35 +141,14 @@ ParityTable:                    ; Table used to translate RS232 parity param
         .byte   $A0             ; SER_PAR_MARK
         .byte   $E0             ; SER_PAR_SPACE
 
-IdOfsTable:                     ; Table of bytes positions, used to check five
-                                ; specific bytes on the slot's firmware to make
-                                ; sure this is an SSC (or Apple //c comm port)
-                                ; firmware that drives an ACIA 6551 chip.
-                                ;
-                                ; The SSC firmware and the Apple //c(+) comm
-                                ; port firmware all begin with a BIT instruction.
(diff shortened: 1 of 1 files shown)


mrdudz:
One for @oliverschmidt - i'm not touching it :)

oliverschmidt:
I see your point, but I guess that if you search long enough, you'll also find other ROMs that claim to be a SSC, but don't have a 6551. I imagine that the Apple II Workstation card could fall into that category.

I'm not sure if it's wise to only rule out the IIgs and "declare" every thing else as 6551-based. Maybe a third driver that doesn't check that byte at all (like the SSC did originally) and that is "use at your own risk" is more appropriate (?)

oliverschmidt:
Or maybe it's most appropriate to say: If all SSC signature bytes are there, then it's appropriate to access the $C0n0 I/O space and try to see if there's a 6551. If that approach doesn't hurt the IIgs, then it may be preferable...

colinleroy:
It could be, but there's not much to check. From what I see, ACIA_STATUS is 0x10 (data register empty) upon startup of Apple 2's with SSC (but 0x70 on Laser 128EX).

oliverschmidt:
I'm thinking along the lines of something like this:
1. Set the command register bit 0 to 0.
2. Read the command register and check, if bit 0 is 0.
3. Set the command register bit 0 to 1.
4. Read the command register and check, if bit 0 is 1.
5. Write to the status register to trigger a 'program reset'.
6. Read the command register and check, if bit 0 is 0.

PR #2216, opened 2023-10-07
[cc65] Type category in a function definition cannot be inherited from a typedef

### LINKED ISSUE #2020: cc65: Compiler allows function definitions via a typedef
The following should give a syntax error but does not:

```c                
typedef void F(void);
F c { }
```
                
It is taken from ISO/IEC 9899:1999 (E), 6.9.1 footnote 137 which explicitly says "The intent is that the type category in a function definition cannot be inherited from a typedef".

acqn:
Fixed #2020.

### DIFF
--- a/src/cc65/compile.c
+++ b/src/cc65/compile.c
@@ -65,6 +65,7 @@
 #include "preproc.h"
 #include "standard.h"
 #include "staticassert.h"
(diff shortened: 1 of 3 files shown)


mrdudz:
looks ok to me, thanks!

PR #2219, opened 2023-10-10
Apple2: Better check for 6551

colinleroy:
Replaces #2213.
I compare the full register because just checking bit 0 sometimes work, sometimes fail. It doesn't seem to hurt the IIgs.

### DIFF
--- a/libsrc/apple2/ser/a2.ssc.s
+++ b/libsrc/apple2/ser/a2.ssc.s
@@ -25,6 +25,8 @@
         .include        "ser-kernel.inc"
         .include        "ser-error.inc"
 
+        .importzp       tmp1, tmp2
+
         .macpack        module
         .macpack        cpu
 
@@ -141,35 +143,16 @@ ParityTable:                    ; Table used to translate RS232 parity param
         .byte   $A0             ; SER_PAR_MARK
         .byte   $E0             ; SER_PAR_SPACE
 
-IdOfsTable:                     ; Table of bytes positions, used to check five
+IdOfsTable:                     ; Table of bytes positions, used to check four
                                 ; specific bytes on the slot's firmware to make
-                                ; sure this is an SSC (or Apple //c comm port)
-                                ; firmware that drives an ACIA 6551 chip.
-                                ;
-                                ; The SSC firmware and the Apple //c(+) comm
-                                ; port firmware all begin with a BIT instruction.
-                                ; The IIgs, on the other hand, has a
-                                ; Zilog Z8530 chip and its firmware starts with
-                                ; a SEP instruction. We don't want to load this
-                                ; driver on the IIgs' serial port. We'll
-                                ; differentiate the firmware on this byte.
-                                ;
-                                ; The next four bytes we check are the Pascal
-                                ; Firmware Protocol Bytes that identify a
-                                ; serial card. Those are the same bytes for
-                                ; SSC firmwares, Apple //c firmwares and IIgs
-                                ; Zilog Z8530 firmwares - which is the reason
-                                ; we have to check for the firmware's first
-                                ; instruction too.
-        .byte   $00             ; First instruction
+                                ; sure this is an SSC (or Apple //c comm port).
         .byte   $05             ; Pascal 1.0 ID byte
         .byte   $07             ; Pascal 1.0 ID byte
         .byte   $0B             ; Pascal 1.1 generic signature byte
         .byte   $0C             ; Device signature byte
 
-IdValTable:                     ; Table of expected values for the five checked
+IdValTable:                     ; Table of expected values for the four checked
                                 ; bytes
-        .byte   $2C             ; BIT
         .byte   $38             ; ID Byte 0 (from Pascal 1.0), fixed
         .byte   $18             ; ID Byte 1 (from Pascal 1.0), fixed
         .byte   $01             ; Generic signature for Pascal 1.1, fixed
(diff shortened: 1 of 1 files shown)


libsrc/apple2/ser/a2.ssc.s
```diff
@@ -210,6 +193,48 @@ SER_CLOSE:
...
+; Output: carry set if this is very probably a 6551, clear if not
+
+Verify6551:
+        lda     ACIA_STATUS,x   ; Save current values in what we expect to be
+        pha                     ; the ACIA status register
+        lda     ACIA_CMD,x      ; and command register. So we can restore them
+        sta     tmp1            ; if this isn't a 6551.
+
+        and     #$01
+        bne     NotAcia         ; We expect command register bit 0 to be 0
```
oliverschmidt, line 209: I don't think it's valid to assume that the 6551 is correctly reset when the driver gets loaded. I'd certainly start setting bit 0 before assuming any value.

libsrc/apple2/ser/a2.ssc.s
```diff
@@ -210,6 +193,48 @@ SER_CLOSE:
...
+; Input: X= Index
+; Output: carry set if this is very probably a 6551, clear if not
+
+Verify6551:
+        lda     ACIA_STATUS,x   ; Save current values in what we expect to be
+        pha                     ; the ACIA status register
+        lda     ACIA_CMD,x      ; and command register. So we can restore them
+        sta     tmp1            ; if this isn't a 6551.
+
+        and     #$01
```
oliverschmidt, line 208: Maybe #%00000001 describes your intention better?

libsrc/apple2/ser/a2.ssc.s
```diff
@@ -210,6 +193,48 @@ SER_CLOSE:
         stx     Index           ; Mark port as closed
         rts
 
+;----------------------------------------------------------------------------
+; Internal function to test whether we're really talking with a 6551,
+; or at least it really looks like it.
+; Input: X= Index
+; Output: carry set if this is very probably a 6551, clear if not
```
oliverschmidt, line 200: In 6502 asm it's pretty common to use the carry flag as an error indicator. So I suggest to revert the logic here.

libsrc/apple2/ser/a2.ssc.s
```diff
@@ -210,6 +193,48 @@ SER_CLOSE:
...
+        clc
+        rts
+
+NotAcia:
+        lda     tmp1            ; Restore saved bytes, they're probably needed
+        sta     ACIA_CMD,x
+        pla
+        sta     ACIA_STATUS,x
+        sec
+        rts
```
oliverschmidt, line 236: I'd suggest to restructure this function a bit...

* After line 222, you can do an LSR to directly have the desired information in the carry flag.
* You can just always execute the restore of the saved values in lines 231-234 as we're not optimizing for cycles, but for bytes. Those instructions all don't change the carry flag.
* As far as I can see, when taking my comment above into account, it's shorter (and clearer) to save both original values on the stack.

colinleroy:
I can't really use LSR to only check bit 0, because the sta at ACIA_STATUS changes ACIA_CMD on the IIgs and its bit 0 is sometimes 1. I suppose these bytes do something else on the IIgs. (which means maybe that check is dangerous, even if it doesn't seem to hurt?)

oliverschmidt:
Regarding the IIgs I see two ways:

1. Drop support for a SSC in a IIgs. In other words: Do the same IIgs check as in the IIgs driver and bail out right away.

2. If a IIgs is detected than check if slot in question is set to "Your Card". I don't know exactly how it is done, but as far as I remember it's not a complex thing to do. Only if the slot in question is set to "Your Card", start checking for the 6551, otherwise bail out. Maybe you want to get other opinions/hints/help on this approach in the Slack channel.

colinleroy:
Oh, I think that's a very good idea. At worst, if checking for "your card" involves 65816 opcodes, I can maybe reinstate the E2 check, but only do it on IIgs?

mrdudz:
as usual, i will not touch this :) @oliverschmidt go ahead :)

oliverschmidt:
The "Your Card" looks exactly I imagined it and it is placed at the very point I expected it.

The thing that I don't understand is why that 6551 detection code is gone again. I imagine the flow like this:

1. Check for IIgs, if not then skip to 3.)
2. Check Your Slot, if not then error out
3. Check ROM ID bytes, if not then error out
4. Check 6551 (as discussed), if not then error out

BTW: The very last statement of the driver can be a BRA instead of JMP for the 65C02. For the dynamically loadable driver, this saves even more than one byte ;-) as there's then no need for load time fixup metadata.

colinleroy:
Oh, I misunderstood you. Will do that, then :)

colinleroy:
It's back and working well :)

libsrc/apple2/ser/a2.ssc.s
```diff
@@ -239,12 +245,53 @@ SER_OPEN:
...
+        lda     tmp2
+        ora     #%00000001      ; Enable receiver/transmitter
+        sta     ACIA_CMD,x
+        sta     tmp3            ; Store it for comparison
+
+        lda     ACIA_CMD,x      ; Is command register what we wrote?
+        cmp     tmp3
+        bne     NotAcia
+
+        sta     ACIA_STATUS,x   ; Reset Acia (value written is not important)
```
oliverschmidt, line 271: ACIA

libsrc/apple2/ser/a2.ssc.s
```diff
@@ -239,12 +245,53 @@ SER_OPEN:
...
+
+        lda     ACIA_CMD,x      ; Is the Acia disabled now?
```
oliverschmidt, line 273: receiver/transmitter

libsrc/apple2/ser/a2.ssc.s
```diff
@@ -239,12 +245,53 @@ SER_OPEN:
...
+        and     #%00000001
+        bne     NotAcia         ; We expect command register bit 0 to be 0
+
+        lda     tmp2
+        ora     #%00000001      ; Enable receiver/transmitter
+        sta     ACIA_CMD,x
+        sta     tmp3            ; Store it for comparison
+
+        lda     ACIA_CMD,x      ; Is command register what we wrote?
+        cmp     tmp3
```
oliverschmidt, line 268: Instead of those 4 instruction just
`cmp ACIA_CMD,x`

libsrc/apple2/ser/a2.ssc.s
```diff
@@ -239,12 +245,53 @@ SER_OPEN:
...
+        bne     NotAcia
+
+        sta     ACIA_STATUS,x   ; Reset Acia (value written is not important)
+
+        lda     ACIA_CMD,x      ; Is the Acia disabled now?
+        and     #%00000001
+        beq     AciaOK
```
oliverschmidt, line 275: I wrote before that we want to only modify bit 0. But now, I'm unsure about that:
* There are 6551 builds, which - of course totally stupidly - reset to interrrupts being **_enabled_**. Keeping interrupts enabled and then enabling the receiver/transmitter might easily result in (unhandled) interrupts.
* When we completely save and restore the two I/O port values in case it's not a 6551 then it's supposed to be not really a difference if we temporarily set other bits beside bit 0.

Therefore my proposal would be now something like this (of course with useful comments):
```
  ldy #%00000010
: tya
  sta ACIA_CMD,x
  cmp ACIA_CMD,x
  bne NotAcia
  iny
  cpy #%00000100
  bne :-
  sta ACIA_STATUS,x
  lda ACIA_CMD,x
  lsr
  bcc AciaOK
```

colinleroy: Oh, that's right, I didn't think about that either. Ok then!

oliverschmidt: Great :-) However, I haven't tested the my code proposal at all. Am I correct to presume that you did?

colinleroy: I did of course! and it was correct. It does exactly what I've put in the comments. 
Driver loads on any II (c, e, gs) with an SSC in slot 2. 
Driver refuses to load on a IIgs with "Modem" instead of "Your card" in slot 2.
Driver refuses to load on a II with a non-6551 card in slot 2. I tested that on emulation, with the Pascal bytes check commented, and a CCS Model 7710 Asynchronous Serial Interface on the slot (it's based on an ACIA 6850).

oliverschmidt: Thanks for the comprehensive testing :-)

colinleroy: At that level it's better to test! 
To be sure it just doesn't happen to work, every test I do in emulation, I trace the execution to verify that it fails for the reason it should fail. For example with the simulated 6850 card, it fails at 

sta ACIA_CMD,x
cmp ACIA_CMD,x
bne NotAcia

PR #2220, opened 2023-10-10
Apple IIgs serial: Real-life fixes

colinleroy:
After having some real hardware tests and continuing on working on this driver, I introduced a regression when I added both channels handling.

Apparently, the real hardware is more finicky than the emulator about addressing mode and/or timing when writing index/value to the control register. So the macros are gone and replaced with functions.

So here's the (fixed) driver, tested on real hardware :) https://www.youtube.com/watch?v=NIU4-WJLx-Y

Also fixed a missing sei at line354, a missing channel load at 679, and a logic bug while waiting to send at 722.

### DIFF
--- a/libsrc/apple2/ser/a2.gs.s
+++ b/libsrc/apple2/ser/a2.gs.s
@@ -314,20 +314,19 @@ SER_CLOSE:
 
 SER_OPEN:
         bit     $C082                   ; Check if this is a IIgs
-        lda     $FE1F                   ; https://prodos8.com/docs/technote/misc/07/
-        cmp     #$60                    ; Everything but the IIgs has an RTS there
-        bne     HardwareFound
+        sec
+        jsr     $FE1F                   ; https://prodos8.com/docs/technote/misc/07/
+        bcc     IIgs
 
-        ; Device (hardware) not found
         bit     $C080
-        lda     #SER_ERR_NO_DEVICE
+        lda     #SER_ERR_NO_DEVICE      ; Not a IIgs
 SetupErrOut:
         cli
         ldx     #$00                    ; Promote char return value
         stx     Opened                  ; Mark port closed
         rts
 
-HardwareFound:
+IIgs:
         bit     $C080
         sei                             ; Disable interrupts
 
--- a/libsrc/apple2/ser/a2.gs.s
+++ b/libsrc/apple2/ser/a2.gs.s
@@ -305,7 +305,7 @@ SER_CLOSE:
         stx     Opened                  ; Mark port as closed
 
         cli
-:       txa
+:       txa                             ; Promote char return value
         rts
 
 ;----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


libsrc/apple2/ser/a2.gs.s
```diff
@@ -314,20 +314,19 @@ SER_CLOSE:
 
 SER_OPEN:
         bit     $C082                   ; Check if this is a IIgs
-        lda     $FE1F                   ; https://prodos8.com/docs/technote/misc/07/
-        cmp     #$60                    ; Everything but the IIgs has an RTS there
-        bne     HardwareFound
+        sec
+        jsr     $FE1F                   ; https://prodos8.com/docs/technote/misc/07/
```
oliverschmidt, line 318: Please don't use a URL here. URLs get invalid at some point in time. Rather use https://github.com/cc65/cc65/blob/master/libsrc/apple2/get_ostype.s#L11

libsrc/apple2/ser/a2.gs.s
```diff
@@ -314,20 +314,19 @@ SER_CLOSE:
...
+        bcc     IIgs
 
-        ; Device (hardware) not found
         bit     $C080
```
oliverschmidt, line 321: BIT doesn't set the carry flag. So you can put the BIT $C080 before the BCC - and thus reuire it only once. Maybe the most elegant looking code would be:
```
sec
bit $C082
jsr $FE1F
bit $c080
bcc IIgs
```

libsrc/apple2/ser/a2.gs.s
```diff
@@ -314,20 +314,19 @@ SER_CLOSE:
 
 SER_OPEN:
         bit     $C082                   ; Check if this is a IIgs
```
oliverschmidt, line 316: I'd say, this comment is better placed on its own above this line as it refers rather to the section than this instruction.

libsrc/apple2/ser/a2.gs.s
```diff
@@ -314,20 +314,19 @@ SER_CLOSE:
```
oliverschmidt, line 314: GitHub doesn't allow me to comment on line 308... There "the" comment is missing.

libsrc/apple2/ser/a2.gs.s
oliverschmidt, line 316: GitHub doesn't allow me to comment on line 308... There "the" comment is missing.

libsrc/apple2/ser/a2.gs.s
oliverschmidt, line 314: The line 308 above should have the `; Promote char return value` comment.

colinleroy:
Thanks! updated :)

libsrc/apple2/ser/a2.gs.s
```diff
@@ -370,7 +357,12 @@ SetupBufs:
```
oliverschmidt, line 357: In line 344 you can replace the JMP with a BRA and thus save a byte.

oliverschmidt:
I think there's one a bit more general thing to improve on this driver that I only noticed now:

In general the hardware detection is supposed to happen in SER_INSTALL. But the SSC driver does it in SER_OPEN. Why? Because the user needs to be able to set the slot before the hardware detection happens.

But this problem isn't relevant for the IIgs driver. Its hardware detection is independent from the port setting. Therefore this driver should do its hardware detection in SER_INSTALL as designed.

Maybe it would be a good idea to add some comment to the SSC driver explaining why its hardware detection is postponed.

PR #2223, opened 2023-10-12
Reverted EOL settings (LF vs CRLF) for Visual Studio

acqn:
WIth the line `end_of_line = crlf` Visual Studio inserts every new line ending as CRLF in all source files despite they all have LF already in them, resulting in mixed line endings saved in source files edited with VS. 

This PR fixes the problem by letting Visual Studio decide which EOL to use by itself.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -2,7 +2,6 @@ root = true
 
 [*]
 charset = utf-8
-end_of_line = crlf
 indent_style = space
 indent_size = 4
 trim_trailing_whitespace = true


spiro-trikaliotis:
Shouldn't git take care of it? That is, text files (this includes sources) should get checked out with CRLF on Windows and checkedin with LF again to the repository?

It least, it was the last time I used git on Windows (admittedly, that was many years ago).

If git is configured that way, your patch might break this workflow.

mrdudz:
IMHO version control should not alter files that way at all. And the editor should leave the line endings alone :)

acqn:
We could just remove this config line and let the editor decide the approprite EOL by itself.

vrubleg:
Support the idea that Git shouldn't alter any files at all and should commit all files as is. So it's better to change `.gitattributes` from `* text=auto` (that can cause unexpected issues because of applied conversions) to `* text=false` (that disables auto conversion and keeps files as is).

acqn:
The current config (#2204) broke my workflow and resulted in mixed EOLs in (almost) every text file edited with VS.

I am proposing to revert that change.

If there is a better way to help with this EOL issue universally, I think it can be applied later.

mrdudz:
> The current config (#2204) broke my workflow and resulted in mixed EOLs in (almost) every text file edited with VS.
> 
> I am proposing to revert that piece of change.
> 
> If there is a better way to help with this EOL issue universally, I think it can be applied later.

so be it :)

PR #2224, opened 2023-10-12
[cc65] Made functions in src/cc65/declare.c less confusing

acqn:
By renaming some parameters and re-hiding a function that was intended to be used internally but got exposed.

### DIFF
--- a/src/cc65/declare.c
+++ b/src/cc65/declare.c
@@ -72,7 +72,7 @@
 
 
 
-static void ParseTypeSpec (DeclSpec* D, typespec_t TSFlags, int* SignednessSpecified);
+static void ParseTypeSpec (DeclSpec* Spec, typespec_t TSFlags, int* SignednessSpecified);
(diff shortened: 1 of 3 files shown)


src/cc65/declare.c
```diff
@@ -1806,7 +1805,7 @@ static FuncDesc* ParseFuncDecl (void)
         /* If the identifier is a typedef, we have a new-style parameter list;
         ** if it's some other identifier, it's an old-style parameter list.
         */
-        Sym = FindSym (CurTok.Ident);
+        SymEntry* Sym = FindSym (CurTok.Ident);
```
spiro-trikaliotis, line 1808: Do we target c99 now? I thought the cc65 sources were still c89/c90?

spiro-trikaliotis: Ok, forget it, this is the first statement in a block.
Nevermind...

mrdudz:
looks good, thanks!

PR #2226, opened 2023-10-13
[cc65] Pragma fixes

### LINKED ISSUE #2151: cc65: #pragma causes errors when used within functions
The compiler seems to get confused and prints errors when `#pragma` is used within functions:

```C
int main()
{
#pragma bss-name("BSS1")
    static int x;
    x = 0;
    return 0;
}
```

produces:

```
test.c:4: Error: Mixed declarations and code are not supported in cc65
test.c:5: Error: Undefined symbol: 'x'
```

It's not only that it ignores the pragma, it will also skip the declaration of 'x'. Older compiler versions (2.18) had different error messages but the same result:

```
test.c(4): Error: Expression expected
test.c(4): Warning: Statement has no effect
test.c(4): Error: `;' expected
test.c(4): Error: Expression expected
test.c(4): Warning: Statement has no effect
test.c(4): Error: `;' expected
test.c(4): Error: Undefined symbol: `x'
test.c(4): Warning: Statement has no effect
```

The problem seems to occurs with all pragmas.

acqn:
- Fixed #2151.
  - As a bug/feature, string literals between the parentheses of a `_Pragma()` operation can be concatenated. A warning will be given if this happens.
- `#pragma charmap` will immediately change the character mapping and affect any character/string literals after it with a few exceptions:
  - String literals as the message of a static assertion or inline assembler code (only the required one, not any optional formatted arguments) in an asm() expression are not translated with either #pragma charmap or target presets.
  - String literals used for preprocessor directives or as the result of stringized macro arguments are never translated.

### DIFF
--- a/doc/cc65.sgml
+++ b/doc/cc65.sgml
@@ -837,21 +837,21 @@ This cc65 version has some extensions to the ISO C standard.
 
 <itemize>
 
-<item>  The compiler allows to insert assembler expressions into the output
-        file. The syntax is
+<item>  The compiler allows to insert inline assembler code in the form of the
+        <tt/asm/ expression into the output file. The syntax is
 
         <tscreen><verb>
-        asm [optional volatile] (&lt;string literal&gt;[, optional parameters]) ;
+        asm [optional volatile] (&lt;string literal&gt;[, optional parameters])
         </verb></tscreen>
         or
         <tscreen><verb>
-        __asm__ [optional volatile] (&lt;string literal&gt;[, optional parameters]) ;
+        __asm__ [optional volatile] (&lt;string literal&gt;[, optional parameters])
         </verb></tscreen>
 
         The first form is in the user namespace; and, is disabled if the <tt/-A/
         switch is given.
 
-        There is a whole section covering inline assembler expressions,
+        There is a whole section covering the inline assembler,
         <ref id="inline-asm" name="see there">.
         <p>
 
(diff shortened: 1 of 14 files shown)


acqn:
Apparently the strings in `asm` should be protected from `#pragma charmap`...

So, should string literals used for _Static_assert be protected from `#pragma charmap`?

mrdudz:
In the assert the string always needs to be in host encoding, no?

Not sure about the asm, but i think the same is true *shrug*

acqn:
Well, `asm` is a non-standard cc65 extension, so we could take special treatment on its content without being too counter-intuitive.
>`asm` is just a cc65 extension that could be executed whenever as we design, for example, in the same translation phase when `#pragma` is executed (just like a `_Pragma` unary operator).
----
On the contrary, `_Static_assert` is a standard C facility that apparently takes just an optional C-standard string literal, so if we treat it differently (from other string literals), it could feel somewhat inconsistent.
>`_Static_assert` is required to be evaluated in a translation phase later than when all `#pragma`s have been executed and removed according to the ISO C standard.

mrdudz:
> whilst the string literal for _Static_assrt should be affected by charmaps....

That makes no sense to me... the additional string literal is going to be printed on the host - that could result in random gibberish and basically be completely useless when the charmap doesn't match whatever the host uses.

acqn:
Agreed. So:

```c
#pragma charmap('a', 'A')
/* ... */
asm("lda #%b", 'a');                             // The user may intend to load 'A' here
```
```c
#pragma charmap('c', 'C')
/* ... */
_Static_assert('c'=='a'+2, "'c' is translated"); // Fail and output "'c' is translated"
```

mrdudz:
That seems the expected behaviour, yes :)

Ready for merge? :)

acqn:
Sure.

PR #2233, opened 2023-10-17
[cc65] Fixed array subscript with a bit-field with patch by kugelfuhr

### LINKED ISSUE #2186: Wrong array indexing when index comes from bit-field
When debugging some weird behavior on a NES game I was making, I found this issue; then I've written this short `sim65`-targeted code which demonstrates the problem:

```c
#include <stdio.h>

typedef struct {
  char flag : 1;
  char index : 7;
} weird_type;

const char array[] = { '6', '5', '0', '2' };

weird_type data;

int main() {
  data.flag = 1;

  data.index = 0;
  putchar(array[data.index]);

  data.index = 1;
  putchar(array[data.index]);
  
  return 0;
}

```

Running this code I get:

```
$ cl65 test.c -t sim6502 -o test; sim65 test
52
```

... when the expected output is `65`.

Looking at the assembly code, apparently almost everything dealing with the bit-fields work correctly (although strangely it places the `flag` at the end of the byte instead of at the beginning).

But when the `.index` field is used as an array index, the assembly code is just loading the whole `data` byte on `Y` and using it to index on `array` (i.e. `ldy _data; lda _array, Y`). Because of the flag at the end of the byte, this is basically the same as accessing the position `2 * data.index + data.flag`. 

I tried this with and without optimizations, both on the version I had before ( `cc65 V2.19 - Git c2e593772` ) and the current master I just installed (`cc65 V2.19 - Git 043590c97`)

acqn:
Fixed #2186.


acqn:
>@mrdudz
[add newline at the end](https://github.com/cc65/cc65/pull/2233/commits/3e602682438c6916d6ae0e50d833d026f8201ef7)

Nice catch.

mrdudz:
This is ok, i guess :)

PR #2234, opened 2023-10-21
[cc65] Minor fix for error recovery from duplicated struct members

acqn:
Moved the check for duplicated members from `HandleSymRedefinition` to `AddLocalSym` and fixed the return value and behavior of the two functions when this check fails. Plus some newlines in code formatting.


mrdudz:
Thanks!

PR #2238, opened 2023-10-22
[cc65] Bool opt fixes

acqn:
The optimization steps on bool tranformers (`boolne`, `boolugt`, `bnegax` etc.) had some obscure bugs (demostrated with the new testcase) that they could neglect usage of the Z flag that was set/cleared by the removed bool tranformers. This PR fixes that and do optimization even further.


polluks:
excellent!

mrdudz:
Nice, thanks!

PR #2241, opened 2023-10-24
Added clarification to inline assembly format specifiers interaction with integers

ToboterXP:
I just spent way too long debugging an error that happened because I assumed an integer passed to the inline assembler would be treated as one by the inline assembly, so this clarification should hopefully prevent any future cases of that.

### DIFF
--- a/doc/cc65.sgml
+++ b/doc/cc65.sgml
@@ -1774,15 +1774,17 @@ as the arguments for replacing the format specifiers will be translated as
 usual.
 
 <itemize>
-  <item><tt/%b/ - Numerical 8-bit value
-  <item><tt/%w/ - Numerical 16-bit value
-  <item><tt/%l/ - Numerical 32-bit value
+  <item><tt/%b/ - Numerical 8-bit value ($xx)
+  <item><tt/%w/ - Numerical 16-bit value ($xxxx)
+  <item><tt/%l/ - Numerical 32-bit value ($xxxxxxxx)
   <item><tt/%v/ - Assembler name of a global variable or function
   <item><tt/%o/ - Stack offset of a local variable
   <item><tt/%g/ - Assembler name of a C label
   <item><tt/%s/ - The argument is converted to a string
   <item><tt/%%/ - The % sign itself
 </itemize><p>
+
+Note: For a numerical value to be treated as an immediate value by the assembler, a # has to be added in front of the format specifier.
 
 Using those format specifiers, you can access C <tt/#defines/, variables, or
 similar stuff from the inline assembler. For example, to load the value of


mrdudz:
This actually confuses me more than it clears up anything. What are you trying to say there?

ToboterXP:
Well, I assumed that if I put an integer into the inline assembler it would be treated as a number, not an adress like it currently is when doing for instance lda %b.
And I want to make sure noone else can get that false impression.

Am 24. Oktober 2023 21:58:53 MESZ schrieb Bob Andrews ***@***.***>:
>This actually confuses me more than it clears up anything. What are you trying to say there?
>
>-- 
>Reply to this email directly or view it on GitHub:
>https://github.com/cc65/cc65/pull/2241#issuecomment-1777940772
>You are receiving this because you authored the thread.
>
>Message ID: ***@***.***>
-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

mrdudz:
> Well, I assumed that if I put an integer into the inline assembler it would be treated as a number, not an adress like it currently is when doing for instance lda %b.
And I want to make sure noone else can get that false impression.

Well, this just has nothing to do with the inline assembler - of course you have to use immediate mode if you want to use an immediate value. That's how 6502 assembly works.

ToboterXP:
Well, I was under the false impression that "numerical 8bit value" automatically meant immediate mode, since that's what is required for a value to be treated as a number rather than an address in Assembly. Of course, that might be clear to someone with a lot of Assembly experience, but it seems like an easy pitfall for people that don't have that much Assembly experience (like me).

 So a line of clarification might be a good idea, since a lot of people that use C might not have in depth knowledge of Assembly, and only occasionally use it to supplement things C can't do.

Am 24. Oktober 2023 22:11:31 MESZ schrieb Bob Andrews ***@***.***>:
>> Well, I assumed that if I put an integer into the inline assembler it would be treated as a number, not an adress like it currently is when doing for instance lda %b.
>And I want to make sure noone else can get that false impression.
>
>Well, this just has nothing to do with the inline assembler - of course you have to use immediate mode if you want to use an immediate value. That's how 6502 assembly works.
>
>
>-- 
>Reply to this email directly or view it on GitHub:
>https://github.com/cc65/cc65/pull/2241#issuecomment-1777963160
>You are receiving this because you authored the thread.
>
>Message ID: ***@***.***>
-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

Marrin:
> Well, I was under the false impression that "numerical 8bit value" automatically meant immediate mode, since that's what is required for a value to be treated as a number rather than an address in Assembly.

But an address can also be an 8 bit value, so why should an 8 bit value be automagically choose immediate mode instructions?  Try to look at it from the other angle: if `%b` automagically would choose immediate mode, how is the user supposed to indicate an 8 bit address then?  You are aware that this is a different addressing mode than 16 bit addresses‽

See:
```
.> c000 a9 2a     lda #$2a
.> c002 a5 2a     lda $2a
.> c004 ad 2a 00  lda $002a
```

How would the second instruction be expressed if 8 bit values would automagically be assembled as the first instruction?

> Of course, that might be clear to someone with a lot of Assembly experience, but it seems like an easy pitfall for people that don't have that much Assembly experience (like me).

That the `#` or the absence of it chooses between immediate values and address doesn't need a lot of 6502 assembly experience.  That's basic knowledge needed to program 6502 assembly.  As is zero page addressing mode.

mrdudz:
> That the # or the absence of it chooses between immediate values and address doesn't need a lot of 6502 assembly experience. That's basic knowledge needed to program 6502 assembly. As is zero page addressing mode.

This. And the inline assembly documentation is the wrong place to explain basic 6502 assembly knowledge for that matter.

ToboterXP:
I mean, it is an important place to explain which bits of the assembly syntax the inline assembler takes care off, and which bits not. Since the # is part of the parameter in all the immediate functions, you could assume it handles it. 

Am 25. Oktober 2023 08:41:40 MESZ schrieb Bob Andrews ***@***.***>:
>> That the # or the absence of it chooses between immediate values and address doesn't need a lot of 6502 assembly experience. That's basic knowledge needed to program 6502 assembly. As is zero page addressing mode.
>
>This. And the inline assembly documentation is the wrong place to explain basic 6502 assembly knowledge for that matter.
>
>-- 
>Reply to this email directly or view it on GitHub:
>https://github.com/cc65/cc65/pull/2241#issuecomment-1778611900
>You are receiving this because you authored the thread.
>
>Message ID: ***@***.***>
-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

mrdudz:
> I mean, it is an important place to explain which bits of the assembly syntax the inline assembler takes care off, and which bits not. Since the # is part of the parameter in all the immediate functions, you could assume it handles it.

Again this confuses me more than it should, sorry. I don't see how you could assume you can omit the # when it is needed. ```asm("lda %b", $02)``` is clearly loading from an address in zeropage, and ```asm("lda #%b", $02)``` is loading an immediate value. What makes you think otherwise? As said, this is basic 6502 knowledge.

ToboterXP:
Well, if there's the statements
lda #$05
And lda $05,
The docs don't really explain which part if the parameter the %b covers. So I assumed it'd take care of the entire #$xx parameter, not just the $xx part. And some other formatter, or the type passed in, would handle the #, since it's also part of the parameter. 
This isn't basic 6502 assembly at all, it very much requires you to know the design behind the language, and how different statements are interpreted, not just what the different commands do.

Am 25. Oktober 2023 10:24:35 MESZ schrieb Bob Andrews ***@***.***>:
>> I mean, it is an important place to explain which bits of the assembly syntax the inline assembler takes care off, and which bits not. Since the # is part of the parameter in all the immediate functions, you could assume it handles it.
>
>Again this confuses me more than it should, sorry. I don't see how you could assume you can omit the # when it is needed. ```asm("lda %b", $02)``` is clearly loading from an address in zeropage, and ```asm("lda #%b", $02)``` is loading an immediate value. What makes you think otherwise? As said, this is basic 6502 knowledge.
>
>
>
>
>
>-- 
>Reply to this email directly or view it on GitHub:
>https://github.com/cc65/cc65/pull/2241#issuecomment-1778763040
>You are receiving this because you authored the thread.
>
>Message ID: ***@***.***>
-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

mrdudz:
> The docs don't really explain which part if the parameter the %b covers.

It clearly says "numerical value". I still don't see how you could interpret "a # is automatically added in front of it" into that.

mrdudz:
> Because for assembly to treat it as a numeric value (that you can do maths on and such), rather than an address (which you can't do maths on), you need the #.

But "numeric value" does in no way imply that is used as an immediate value. $02 is a numeric value. It can be used as an immediate or as an address.

Compyx:
> Because for assembly to treat it as a numeric value (that you can do maths on and such), rather than an address (which you can't do maths on), you need the #.

You can do math on an address: e.g. `lsr $1000`, `asl $10`.

Marrin:
@ToboterXP There is a section [about inline assembly in the cc65 documentation](https://cc65.github.io/doc/cc65.html#inline-asm) where the very first example reads like this:
```
        #define OFFS  23
        __asm__ ("ldy #%b", OFFS);
```

See the use of `#` there.

Also you can of course do math on addresses in the assembler.  Example:
```
    #define ZP_BASE     42
    __asm__ ("lda %b + 23", ZP_BASE);
```

In the assembled code one might want to do math to addresses too.  Calculating pointers from addresses or addresses and offsets for instance.  So neither before assembling nor after there is a real distinction between numeric values that are data and numeric values that are addresses in this regard.

You see the `#` as part of the operand where it actually is part of the instruction, because it selects the addressing mode.  And more importantly it disambiguates immediate addressing mode from zero page addressing mode.  I repeat my question: How would you express zero page addressing mode if the `#` would be implicit in `%b`?  That wasn't a rhetorical question because that would be a real problem that needs to be solved then.

Marrin:
@ToboterXP So far we where addressing the added *Note:*, but I think the other change, the `$xx`, `$xxxx` in parenthesis isn't a good idea either because this is an implementation detail IMHO.  Currently the compiler inserts those values in hexadecimal notation, but there's no reason why they couldn't be in decimal notation instead.  The user shouldn't make any assumptions here other than the number is inserted into the assembly source in any format understood by the assembler.

mrdudz:
I am closing this, for all the reasons already mentioned.

PR #2242, opened 2023-10-25
[cc65] Fixed type promotion of switch case values

### LINKED ISSUE #2019: cc65: Compiler has problems with switch statements
Type promotion in switch statements seems to be broken.

```c
unsigned char c = 0xFF;
void func(void)
{
    switch (c) {
        case (signed char)0xFF: ;
        case (unsigned char)0xFF: ;
    }
}
```
Gives:
```
test.c:5: Error: Range error
test.c:6: Error: Duplicate case label
```

But both errors are wrong. The first one is simply a label that is never jumped to and the second one is not a duplicate of the first.

ISO/IEC 9899:1999 (E) 6.8.4.2p5 says: "The integer promotions are performed on the controlling expression. The constant
expression in each case label is converted to the promoted type of the controlling expression. If a converted value matches that of the promoted controlling expression, control jumps to the statement following the matched case label."

It seems that this is not done by the compiler. The controlling expression is not promoted to int, and in turn the case label expressions are kept as is, in which case the are identical. If they were correctly promoted to int, they would be different.

This is a problem even without any casts. The following does also give a range error:

```c
signed char c = 0xFF;
void func(void)
{
    switch (c) {
        case 0x80: ;
    }
}
```

acqn:
Fixed #2019.

### DIFF
--- a/src/cc65/swstmt.c
+++ b/src/cc65/swstmt.c
@@ -64,7 +64,7 @@
 typedef struct SwitchCtrl SwitchCtrl;
 struct SwitchCtrl {
     Collection* Nodes;          /* CaseNode tree */
-    TypeCode    ExprType;       /* Basic switch expression type */
+    const Type* ExprType;       /* Switch controlling expression type */
     unsigned    Depth;          /* Number of bytes the selector type has */
     unsigned    DefaultLabel;   /* Label for the default branch */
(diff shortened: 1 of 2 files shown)


mrdudz:
Thanks!

Issue #2207, opened 2023-10-04
memcpy optimization buggy when dst ptr is cast


clbr:
Today's master. When memcpy's destination is a pointer, and there's a cast (even to its own type), the fixed-dest-and-src optimization gets wrongly applied. It's a pointer, not an array, so this overwrites some random memory.

The two memcpy lines should be identical and create identical code. The cast somehow confuses the optimization.

```c
#include <string.h>

extern void *dst;
extern unsigned char src[16];

void foo() {
        memcpy(dst, src, 16);
        memcpy((void *) dst, src, 16);
}
```

```asm
;
; File generated by cc65 v 2.19 - Git 9667a5f99
;
        .fopt           compiler,"cc65 v 2.19 - Git 9667a5f99"
        .setcpu         "6502"
        .smart          on
        .autoimport     on
        .case           on
        .debuginfo      off
        .importzp       sp, sreg, regsave, regbank
        .importzp       tmp1, tmp2, tmp3, tmp4, ptr1, ptr2, ptr3, ptr4
        .macpack        longbranch
        .import         _memcpy
        .import         _dst
        .import         _src
        .export         _foo

; ---------------------------------------------------------------
; void __near__ foo (void)
; ---------------------------------------------------------------

.segment        "CODE"

.proc   _foo: near

.segment        "CODE"

;
; memcpy(dst, src, 16);
;
        lda     _dst
        ldx     _dst+1
        jsr     pushax
        lda     #<(_src)
        ldx     #>(_src)
        jsr     pushax
        ldx     #$00
        lda     #$10
        jsr     _memcpy
;
; memcpy((void *) dst, src, 16);
;
        ldy     #$0F
L0002:  lda     _src,y
        sta     _dst,y
        dey
        bpl     L0002
;
; }
;
        rts

.endproc
```

acqn:
It seems to be caused by the change of meaning of "rvalueness" without correcting cc65/stdfunc.c accordingly.

Generally, `ED_IsRVal` in that file should be replaced with `ED_IsAddrExpr` and `ED_IsLVal` should be replaced with `ED_IsIndExpr`.

mrdudz:
I am always amazed by how much we can still break without any tests failing...

Could you have a look? I think you are the only one who understands what you have been saying :)

acqn:
[...]

PR #2243, opened 2023-10-26
[cc65] Inlined std functions fixes

acqn:
Fixed the OP issue and the `strlen` issue in #2207. The missing argment evaluation and `(void)call` bugs are not fixed.


mrdudz:
Nice, thanks!

PR #2244, opened 2023-10-27
Apple2 SSC: Implement no flow control

colinleroy:


oliverschmidt:
I must admit that I don't like the IRQ routine to become slower - especially in "default" code path.

My proposal is to "just" have RtsOff the bit %00001000 set when no handshake is desired. Then both the code asserting flow control in the IRQ routine and the code de-asserting flow control in the get routine just always "unnecessarily" write the same value into the register.

This allows you to remove the HsType variable again and use tmp2 until you get to the point where RtsOff is set.

colinleroy:
Very good idea :)

oliverschmidt:
You are right. I originally thought that the additional cycles would be spent on every interrupt. But you did it right in the first place: Only check for the handshake type type _after_ checking that flow control is desired.

colinleroy:
Thank you @oliverschmidt!

PR #2246, opened 2023-10-28
[cc65] Fixes for nested unspecified-length/flexible arrays

acqn:
```c
int a[][] = { 0 }; /* Should be an error */
```


mrdudz:
hooray :)

PR #2248, opened 2023-10-30
[cc65] Improved function return type check

### LINKED ISSUE #2024: cc65: Inconsistent warnings for qualified function return values
The compiler warns if there are type qualifiers in a function return type, but not always. This generates a warning:

```c
const char f1(void);
```
while this does not:
```c
char* const f2(void);
```

Both functions return a const qualified value, so one would expect to get warnings in both cases.

acqn:
Resolved #2024.

### DIFF
--- a/src/cc65/declare.c
+++ b/src/cc65/declare.c
@@ -493,8 +493,44 @@ static void FixQualifiers (Type* DataType)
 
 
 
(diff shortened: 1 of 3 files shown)


mrdudz:
Thanks!

PR #2252, opened 2023-11-06
Optimize multiplication by zero

colinleroy:
Before, without -O:
```
; c = width * 0;
;
	lda     M0004
	ldx     M0004+1
	jsr     pushax
	ldx     #$00
	lda     #$00
	jsr     tosmulax
	sta     M0003
	stx     M0003+1
```
Before, with -O:
```
; c = width * 0;
;
	lda     M0004
	ldx     M0004+1
	jsr     pushax
	lda     #$00
	jsr     tosmula0
	sta     M0003
	stx     M0003+1
```

After, without -O:
```
; c = width * 0;
;
	lda     M0004
	ldx     M0004+1
	lda     #$00
	tax
	sta     M0003
	stx     M0003+1

```

After, with -O:
```
; c = width * 0;
;
	stz     M0003
	stz     M0003+1
```

### DIFF
--- a/src/cc65/codegen.c
+++ b/src/cc65/codegen.c
@@ -2726,7 +2726,12 @@ void g_mul (unsigned flags, unsigned long val)
                 if (flags & CF_FORCECHAR) {
                     /* Handle some special cases */
                     switch (val) {
-
+                        case 0:
+                            AddCodeLine ("lda #$00");
+                            return;
+                        case 1:
+                            /* Nothing to do */
+                            return;
                         case 3:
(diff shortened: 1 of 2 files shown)


colinleroy:
I admit it looks stupid because who would do an `a = b * 0`, but in some cases such constructs can be used for code legibility, like in

```
uint8 pixel[640*480];
#define PIXEL(x, y) pixel[(640*y)+x]

start = PIXEL(0, 0);
```

mrdudz:
nice one, thanks!

PR #2253, opened 2023-11-06
Implement SER_HS_NONE on Apple IIgs serial driver

colinleroy:
Same as previous one, for the IIgs SCC this time :)


colinleroy:
And poke @oliverschmidt :)

colinleroy:
Oh indeed! Will do.

PR #2254, opened 2023-11-07
Upload 64-bit Windows snapshot to Sourceforge

brian6932:
It would be nice to have a persistent snapshot binary upload for 64bit as well

### DIFF
--- a/.github/workflows/snapshot-on-push-master.yml
+++ b/.github/workflows/snapshot-on-push-master.yml
@@ -135,6 +135,15 @@ jobs:
           port: ${{ secrets.SSH_PORT }}
           user: ${{ secrets.SSH_USER }}
           key: ${{ secrets.SSH_KEY }}
+      - name: Upload 64-bit Windows snapshot to sourceforge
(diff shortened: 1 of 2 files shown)


brian6932:
Since Travis CI isn't used anymore, shouldn't the `Makefile.travis`, and `.travis.yml` files be deleted?

mrdudz:
Since i prefer things being kept simple - what is the advantage of using the 64bit build over the 32bit build?

brian6932:
Everyone has 64bit computers nowadays, so if one should exist, I think it should be 64bit. As someone who wants to make a persistent package for both 64bit and 32bit, it kinda breaks my ability to do so, if for whatever reason the repo doesn't get a commit in the last 30 days (That's the actions artifact expiration time). Since both snapshots are made anyway, I don't really see why not, not like it adds any complexity, and just improves the performance of the compiler for it's users.

mrdudz:
> Since both snapshots are made anyway, I don't really see why not, not like it adds any complexity

Yes, i didn't see this. Added, thx :)

PR #2260, opened 2023-11-13
Optimize integer decrements by 1

colinleroy:
Decrementing integers could be a little bit faster, esp. on 65C02. Comparisons attached. I added an optimisation pass because the change left a 
``` 
lda M0001
sta M0002
lda M0001
```
That was hurting my eyes.

Without -O:
sub1-no-opt-comparison

With -O:
sub1-opt-comparison


mrdudz:
Nice one, thanks!

PR #2266, opened 2023-11-15
[cc65] Fixes for structs/unions

### LINKED ISSUE #2018: cc65: Compiler has problems with const struct fields
Handling of const struct fields is broken in several respects:

```c
struct x {
    const int a;
    int b;
};
    
struct x f(void)
{
    struct x x = { 1, 2 };
    return x;
}
    
int main(void)
{
    struct x x;
    x = f();            /* Should be an error "assignment to read only variable" */
    return 0;
}
```

```c
struct y {
    const union {
        int a, b;
    };
} y;
        
void f(void)
{
    y.a = 3;            /* Should be an error "assignment of member ‘a’ in read-only object" */
}
```

```c
struct x {
    const a;            /* Should default to int but gives error */
};
```

acqn:
- Fixed #2018. Fixed const qualifiers on named structs/unions members that should prevent assignments to the whole structs/unions.
  Added warning on ignored qualifiers on anonymous structs/unions.
- Fixed empty declarations in structs/unions.
  Note: Empty complete struct/union declarations with tag names but no identifiers as fields names inside a struct/union were wrongly treated as anonymous structs/unions. This is fixed now.

### DIFF
--- a/src/cc65/assignment.c
+++ b/src/cc65/assignment.c
@@ -44,6 +44,7 @@
 #include "scanner.h"
 #include "stackptr.h"
 #include "stdnames.h"
+#include "symentry.h"
 #include "typecmp.h"
 #include "typeconv.h"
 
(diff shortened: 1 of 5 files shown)


mrdudz:
Thanks!

PR #2268, opened 2023-11-17
add rp6502 target

rumbledethumps:
New target for the [Picocomputer](https://picocomputer.github.io/).

### DIFF
--- a/libsrc/rp6502/gettimezone.c
+++ b/libsrc/rp6502/gettimezone.c
@@ -1,0 +1,24 @@
+/*
+ * Copyright (c) 2023 Rumbledethumps
+ *
+ * SPDX-License-Identifier: Zlib
+ * SPDX-License-Identifier: BSD-3-Clause
+ * SPDX-License-Identifier: Unlicense
+ */
+
+#include <rp6502.h>
+#include <time.h>
+
+int clock_gettimezone(clockid_t clock_id, struct _timezone *tz)
+{
+    int ax;
+    ria_set_ax(clock_id);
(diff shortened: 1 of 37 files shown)


mrdudz:
Some nitpicking from me:
- please use braces on the same line as if/for/while etc (see codestyle)
- i am not too keen on having those license headers in every file (the existing library files do not do this either)
- a line or two in the documentation would be nice (see other targets, perhaps a hint on the target itself, and how to emulate it)

WojciechGw (approved):

libsrc/rp6502/gettimezone.c
```diff
@@ -0,0 +1,15 @@
...
+#include <time.h>
+
+int clock_gettimezone (clockid_t clock_id, struct _timezone* tz)
+{
+    int ax;
+    ria_set_ax (clock_id);
+    ax = ria_call_int_errno (RIA_OP_CLOCK_GETTIMEZONE);
+    if (ax >= 0) {
+        char i;
+        for (i = 0; i < sizeof (struct _timezone); i++)
```
mrdudz, line 11: please use braces also for single-line for/while/if etc

libsrc/rp6502/randomize.c
```diff
@@ -0,0 +1,8 @@
+#include <rp6502.h>
+#include <stdlib.h>
+
+// Non-standard cc65
```
mrdudz, line 4: no C++ comments please :)

mrdudz:
looks good now, thanks!

PR #2270, opened 2023-11-24
Remove all indexed accesses to SCC registers

colinleroy:
Some IIgs models had issues sending bytes out due to phantom reads. (I wonder why mine didn't!)
Also make channel tests clearer.


colinleroy:
This one's for you @oliverschmidt . (Sorry, I should have addressed those with #2270 even if it worked on *my* IIgs.)

PR #2271, opened 2023-11-27
[cc65] Fixed assertion failure when there is an undeclared symbol used in a parameter list

acqn:
Fixed the issue below:
```c
void ddf(int a[c], int b[d()]); /* Wrongly inserted 'c' and 'd' to the parameter symbol table before 'a' and 'b' */
void ddf(int a[], int b[]);     /* Assertion failure as 'c' (or 'd' if there weren't 'c') is not a parameter symbol */
```

Note: the diagnostic message complains about "undefined symbol" whilst it is actually an "undeclared symbol". See #2272.

### DIFF
--- a/src/cc65/declare.c
+++ b/src/cc65/declare.c
@@ -1850,6 +1850,7 @@ static FuncDesc* ParseFuncDecl (void)
     }
 
     /* Parse params */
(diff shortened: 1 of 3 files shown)


mrdudz:
Thanks!

PR #2272, opened 2023-11-27
[cc65] Improved diagnostics

### LINKED ISSUE #2023: cc65: Regression in diagnostic output
The latest compiler version doesn't handle error recovery as gracefully as version 2.18. For this one liner

```c
void f(_Bool b);
```

the version 2.18 compiler recovers from the error after a few error messages. The current git version repeatedly outputs "Error: Mixed declarations and code are not supported in cc65" until the maximum error count is reached.

acqn:
- [x] Resolved #2023.
- [x] Corrected "undefined symbol" with "undeclared symbol".
- [x] Fixed some rare cases when a single file-scope error could get reapeated endlessly until the maximum total count of errors allowed is reached.
- [X] Maximum total count of errors before the compiler unconditionally bails out is now 200.
- [X] If more than 20 errors occur on the same source line, the compiler will immediately bail out.

### DIFF
--- a/src/cc65/asmstmt.c
+++ b/src/cc65/asmstmt.c
@@ -102,7 +102,7 @@ static SymEntry* AsmGetSym (unsigned Arg, unsigned Type)
 
     /* Did we find a symbol with this name? */
     if (Sym == 0) {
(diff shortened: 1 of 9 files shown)


mrdudz:
thanks!

PR #2279, opened 2023-12-02
Expand macros in listing

pm100:
This code expands macros in the listing. This is extremely useful for macros that include segment switches (its what made me do this) but is also generally useful.

adds command line switches
 - -x
 - --expand_macros

It does not try to save to macro source code, instead the macro is 'disassembled' when it is expanded. This was, I felt, less invasive on the internal data structures. This does mean that the expansion is not 100% like the input but is certainly sufficient to work out what line generated what code, and importantly what segment it went into.

Here is sample from msbasic.s (from https://github.com/mist64/msbasic) with and without expansion, it target 3 different segments

without

```
000000r 2  rr rr 45 4E  		keyword_rts "END", END
000004r 2  C4 xx 
```

with expansion

```
000000r 2               		keyword_rts "END", END
000000r 0               >  .SEGMENT VECTORS
000000r 0  rr rr        >  .WORD vec-1
000002r 0               >  keyword key, token
000002r 0               >>  .SEGMENT KEYWORDS
000000r 0               >>  htasc key
000000r 0               >>>  .REPEAT .STRLEN(str)-1,I
000000r 0               >>>  .BYTE .STRAT(str,I)
000001r 0  45 4E        >>>  .ENDREP
000002r 0  C4           >>>  .BYTE .STRAT(str,.STRLEN(str)-1) | 128
000003r 0               >>  define_token token
000003r 0               >>>  .SEGMENT DUMMY
000000r 0               >>>  .IFNBLANK token
000000r 0               >>>  token := <(*-DUMMY_START)+128
000000r 0               >>>  .ENDIF
000000r 0  xx           >>>  .RES 1
```

these macros are nested 3 deep (thats what the > characters show)

### DIFF
--- a/src/ca65/main.c
+++ b/src/ca65/main.c
@@ -702,9 +702,11 @@ static void OneLine (void)
     int           Instr = -1;
 
     /* Initialize the new listing line if we are actually reading from file
-    ** and not from internally pushed input, unless expanding macros
+    ** and not from internally pushed input
     */
-    if (!HavePushedInput () || ExpandMacros) {
+
+    
+    if (!HavePushedInput () ) {
         InitListingLine ();
     }
 
--- a/src/ca65/main.c
+++ b/src/ca65/main.c
@@ -676,7 +676,7 @@ static void OptExpandMacros (const char* Opt attribute ((unused)),
     const char* Arg attribute ((unused)))
     /* Expand macros in listing
     ** one -m means short listing
-    ** two means full listing 
+    ** two means full listing
     */
 {
 
@@ -709,7 +709,6 @@ static void OneLine (void)
     ** and not from internally pushed input
     */
 
-    
     if (!HavePushedInput () ) {
         InitListingLine ();
     }
@@ -899,7 +898,7 @@ static void Assemble (void)
     while (CurTok.Tok != TOK_EOF) {
         OneLine ();
     }
- 
+
 }
 
 
--- a/src/ca65/main.c
+++ b/src/ca65/main.c
@@ -131,7 +131,7 @@ static void Usage (void)
             "  --target sys\t\t\tSet the target system\n"
             "  --verbose\t\t\tIncrease verbosity\n"
             "  --version\t\t\tPrint the assembler version\n"
-            "  --expand_macros\t\tExpand macros in the listing\n",
+            "  --expand-macros\t\tExpand macros in the listing\n",
             ProgName);
 }
 
@@ -975,7 +975,7 @@ int main (int argc, char* argv [])
         { "--verbose",             0,      OptVerbose              },
         { "--version",             0,      OptVersion              },
         { "--warnings-as-errors",  0,      OptWarningsAsErrors     },
-        { "--expand_macros",       0,      OptExpandMacros         },
+        { "--expand-macros",       0,      OptExpandMacros         },
     };
 
     /* Name of the global name space */
--- a/src/ca65/main.c
+++ b/src/ca65/main.c
@@ -674,9 +674,13 @@ static void OptWarningsAsErrors (const char* Opt attribute ((unused)),
 
 static void OptExpandMacros (const char* Opt attribute ((unused)),
     const char* Arg attribute ((unused)))
-    /* Exapnd macros in listing */
+    /* Expand macros in listing
+    ** one -m means short listing
+    ** two means full listing 
+    */
 {
-    ExpandMacros = 1;
+
+    ExpandMacros++;
 }
 
 static void DoPCAssign (void)
@@ -895,6 +899,7 @@ static void Assemble (void)
     while (CurTok.Tok != TOK_EOF) {
         OneLine ();
     }
+ 
 }
 
 
@@ -1080,7 +1085,7 @@ int main (int argc, char* argv [])
                     WarnLevel = atoi (GetArg (&I, 2));
                     break;
                 case 'x':
-                    ExpandMacros = 1;
+                    ExpandMacros++;
                     break;
 
 
--- a/src/ca65/main.c
+++ b/src/ca65/main.c
@@ -107,6 +107,7 @@ static void Usage (void)
             "  -s\t\t\t\tEnable smart mode\n"
             "  -t sys\t\t\tSet the target system\n"
             "  -v\t\t\t\tIncrease verbosity\n"
+            "  -x\t\t\t\tExpand macros\n"
             "\n"
             "Long options:\n"
             "  --auto-import\t\t\tMark unresolved symbols as import\n"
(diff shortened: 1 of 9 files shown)


spiro-trikaliotis:
But you see the disadvantage?
```
000000r 0               >>>  .REPEAT .STRLEN(str)-1,I
000000r 0               >>>  .BYTE .STRAT(str,I)
000001r 0  45 4E        >>>  .ENDREP
```
The bytes are emmited in the line with .BYTE, not with the .ENDREP. Note also that the address at the beginning has already progressed (to 000001r), and thus, it is wrong for this line.

If you have more than a simple .BYTE in .REPEAT .ENDREP (think of segment changes!), it will not work, either.

pm100:
@spiro-trikaliotis - ty for pointing this out. I will fix it

pm100:
A big rework of original PR - thanks to @spiro-trikaliotis for pointing out deficiencies.

Now 2 modes of macro expansion , short and long.
- -x is short, only shows code generating lines and .segment statements
- -x -x is long, shows full expansion

Samples from msbasic (nested 3 deep, with repeats, and segment switches)

Original output

```
000000r 2  rr rr 45 4E  		keyword_rts "END", END
000004r 2  C4 xx   
```  

Short expansion. With code lines, segments statements and the macro invocation lines

```
000000r 2               		keyword_rts "END", END
000000r 0               >  .SEGMENT VECTORS
000000r 0  rr rr        >  .WORD vec-1
000002r 0               >  keyword key, token
000002r 0               >>  .SEGMENT KEYWORDS
000000r 0               >>  htasc key
000000r 0  45           >>>>  .BYTE .STRAT( END,I)
000001r 0  4E           >>>>  .BYTE .STRAT( END,I)
000002r 0  C4           >>>  .BYTE .STRAT(str,.STRLEN(str)-1) | 128
000003r 0               >>  define_token token
000003r 0               >>>  .SEGMENT DUMMY
000000r 0  xx           >>>  .RES 1
```

Full expansion
```
000000r 2               		keyword_rts "END", END
000000r 0               >  .SEGMENT VECTORS
000000r 0  rr rr        >  .WORD vec-1
000002r 0               >  keyword key, token
000002r 0               >>  .SEGMENT KEYWORDS
000000r 0               >>  htasc key
000000r 0               >>>  .REPEAT .STRLEN(str)-1,I
000000r 0               >>>  .BYTE .STRAT(str,I)
000000r 0               >>>  .ENDREP
000000r 0  45           >>>>  .BYTE .STRAT( END,I)
000001r 0  4E           >>>>  .BYTE .STRAT( END,I)
000002r 0  C4           >>>  .BYTE .STRAT(str,.STRLEN(str)-1) | 128
000003r 0               >>  define_token token
000003r 0               >>>  .SEGMENT DUMMY
000000r 0               >>>  .IFNBLANK token
000000r 0               >>>  token := <(*-DUMMY_START)+128
000000r 0               >>>  .ENDIF
000000r 0  xx           >>>  .RES 1
```

mrdudz:
Do we still see the non-expanded macro somewhere? Could you perhaps make it output the non expanded one before the first expansion or something like that?

Also, please use --expand-macros instead of --expand_macros

pm100:
@mrdudz  - not sure what you mean. The macro definition is still output like before. Is that what you meant. This patch doesnt remove anything from the listing, only adds stuff

Will fix the long option name

mrdudz:
I see, sorry - i thought for a moment that the above excerpt is a complete listing

mrdudz:
Why close this?

pm100:
bacause for my needs I have discovered a better way of doing things, and it was just sitting here

pm100:
closed for the same reason as the macro one - certainly this one is better served by a different solution (I now know my way around the dbgfile output of the linker - it provides all the info I need, I am using it to do a huge upgrade of db65 ), and it just sat  for 2 months, I dont like to leave PRs dangling

mrdudz:
> it just sat for 2 months, I don't like to leave PRs dangling

2 months is nothing really :) Especially when it comes to the assembler (compiler and library has priority) - i don't have a lot of time for these things, so please be patient - i will look at it at some point.

Obviously, if you have a better solution, then make a PR for that instead :)

src/ca65/main.c
```diff
@@ -962,6 +975,7 @@ int main (int argc, char* argv [])
         { "--verbose",             0,      OptVerbose              },
         { "--version",             0,      OptVersion              },
         { "--warnings-as-errors",  0,      OptWarningsAsErrors     },
+        { "--expand-macros",       0,      OptExpandMacros         },
```
mrdudz, line 978: The added line should go after "--debug-info" (options are alphabetically sorted

mrdudz, line 978: The added line should go after "--debug-info" (options are alphabetically sorted

src/ca65/main.c
```diff
@@ -670,7 +672,16 @@ static void OptWarningsAsErrors (const char* Opt attribute ((unused)),
     WarningsAsErrors = 1;
 }
 
+static void OptExpandMacros (const char* Opt attribute ((unused)),
+    const char* Arg attribute ((unused)))
+    /* Expand macros in listing
+    ** one -m means short listing
```
mrdudz, line 678: shouldn't this be -x (and not -m)? :)

src/ca65/main.c
```diff
@@ -129,7 +130,8 @@ static void Usage (void)
             "  --smart\t\t\tEnable smart mode\n"
             "  --target sys\t\t\tSet the target system\n"
             "  --verbose\t\t\tIncrease verbosity\n"
-            "  --version\t\t\tPrint the assembler version\n",
+            "  --version\t\t\tPrint the assembler version\n"
+            "  --expand-macros\t\tExpand macros in the listing\n",
```
mrdudz, line 134: The added line should go after "--debug-info" (options are alphabetically sorted

mrdudz:
The added options should also be added to the documentation.

pm100:
@mrdudz Not sure If i have fixed the sgml correctly. I dont know how to render it. Especially as I needed two lines

dominicbeesley:
Upvote for this - do you need help with the SGML?

pm100:
@dominicbeesley - I just copied what was already there, but I do not know how to render the sgml to make sure that I did it right.

dominicbeesley:
I think you just do a 

`make html`

at the top level then check the ca65.html file.

D

mrdudz:
diff looks ok to me, does it still work? :)

mrdudz:
Finally got the time to test this - its really one of those features i have been missing myself for a long time.

So yeah, thanks - merging :)

(Mental note: the command line options details is missing the docs, i will fix that)

Issue #2277, opened 2023-11-29
Please use -std=gnu89 for reference builds


fweimer-rh:
The testsuite is mostly written in C89.  By default, future compilers will not support C89-only language features that were in C99, leading to build failures.

```diff
diff --git a/test/ref/Makefile b/test/ref/Makefile
index abd3e9bc0..7c01bcfbb 100644
--- a/test/ref/Makefile
+++ b/test/ref/Makefile
@@ -40,7 +40,7 @@ OPTIONS = g O Os Osi Osir Osr Oi Oir Or
 ISEQUAL = ..$S..$Stestwrk$Sisequal$(EXE)
 
 CC = gcc
-CFLAGS = -O2 -Wall -W -Wextra -funsigned-char -fwrapv -fno-strict-overflow
+CFLAGS = -std=gnu89 -O2 -Wall -W -Wextra -funsigned-char -fwrapv -fno-strict-overflow
 
 .PHONY: all clean
 
```

Related to:

* https://fedoraproject.org/wiki/Changes/PortingToModernC
* https://fedoraproject.org/wiki/Toolchain/PortingToModernC

mrdudz:
I'd be curios to see what would actually break :)

Could you just make a PR for this?

mrdudz:
Sorry, i don't understand. How is that a problem when you are making a fork on github in order to make a PR? *shrug*

kugelfuhr:
He is working for a company that creates (and maybe sells) a Linux distribution. If he would fork the repository, he would distribute the code, but since he has a commercial background, this is not allowed according to the license cited.

It's a strict interpretation of the license but probably correct. Just because nothing serious would happen does not mean it is allowed.

The real problem is with cc65 which includes sources with a non-free license. The LCC license that covers some of the tests says:

> Permission to use, copy, modify, and distribute this software for any
purpose, subject to the provisions described below, without fee is
hereby granted, provided that this entire notice is included in all
copies of any software that is or includes a copy or modification of
this software and in all copies of the supporting documentation for
such software.

Distributing LCC code as part of the cc65 sources without pointing to the LCC license is actually a license violation.

The tests contain other problematic code. I think it contains also GPLed code from SDCC without mentioning this anywhere but the sources.

BTW: SDCC is not better in this respect. Some time ago, it contained pieces of the cc65 runtime with the GPL attached to them. Looking now, references to cc65 have been removed, but it's still the code from cc65 and GPL licensed. Go figure.

[...]

PR #2283, opened 2023-12-05
Use C89 for reference builds (Fixes: #2277)

sur5r:


mrdudz:
as said in the other ticket, it doesnt work with c89. not with gnu89 either. its not the right thing to do either, since the tests may have C99 or even later stuff in them,

Issue #1919, opened 2022-11-18
Building 64bit windows binaries is broken


mrdudz:
So now that finally building and testing using cmd.exe is supposed to be working again - the regular build just fails. This seems to be another crazy issue with mingw and the windows crt
```
cc65/codegen.c: In function ‘g_defdata’:
cc65/codegen.c:4432:30: error: unknown conversion type character ‘l’ in format [-Werror=format=]
 4432 |                 AddDataLine ("\t.byte\t$%02"PRIXPTR, val & 0xFF);
      |                              ^~~~~~~~~~~~~~~
In file included from cc65/codegen.c:36:
/usr/share/mingw-w64/include/inttypes.h:38:18: note: format string is defined here
   38 | #define PRIX64 "llX"
      |                  ^
cc65/codegen.c:4432:30: error: too many arguments for format [-Werror=format-extra-args]
 4432 |                 AddDataLine ("\t.byte\t$%02"PRIXPTR, val & 0xFF);
      |                              ^~~~~~~~~~~~~~~
cc65/codegen.c:4436:30: error: unknown conversion type character ‘l’ in format [-Werror=format=]
 4436 |                 AddDataLine ("\t.word\t$%04"PRIXPTR, val & 0xFFFF);
      |                              ^~~~~~~~~~~~~~~
In file included from cc65/codegen.c:36:
/usr/share/mingw-w64/include/inttypes.h:38:18: note: format string is defined here
   38 | #define PRIX64 "llX"
      |                  ^
cc65/codegen.c:4436:30: error: too many arguments for format [-Werror=format-extra-args]
 4436 |                 AddDataLine ("\t.word\t$%04"PRIXPTR, val & 0xFFFF);
      |                              ^~~~~~~~~~~~~~~
cc65/codegen.c:4440:30: error: unknown conversion type character ‘l’ in format [-Werror=format=]
 4440 |                 AddDataLine ("\t.dword\t$%08"PRIXPTR, val & 0xFFFFFFFF);
      |                              ^~~~~~~~~~~~~~~~
In file included from cc65/codegen.c:36:
/usr/share/mingw-w64/include/inttypes.h:38:18: note: format string is defined here
   38 | #define PRIX64 "llX"
      |                  ^
cc65/codegen.c:4440:30: error: too many arguments for format [-Werror=format-extra-args]
 4440 |                 AddDataLine ("\t.dword\t$%08"PRIXPTR, val & 0xFFFFFFFF);
[...]

PR #2284, opened 2023-12-05
Adds 64 bit Visual Studio build and then fixes the errors with it

### LINKED ISSUE #2282: 64bit build using MSVC (Visual Studio 2022) fails
Many fatals. Two primary causes
- treating the difference of two pointers as an int `12>C:\work\forks\cc65\src\ca65\instr.c(1892,32): warning C4244: 'return': conversion from '__int64' to 'int', possible loss of data
`
- treating size_t as an int `12>C:\work\forks\cc65\src\ca65\objfile.c(291,1): warning C4267: 'initializing': conversion from 'size_t' to 'unsigned int', possible loss of data
`

These are fatals due to the 'treat warnings as errors' flags.

I have not yet tried building the libraries

I am happy to fix

pm100:
fixes #2282


mrdudz:
To be honest - I don't like "fixing" errors just by casting to the right type. It's almost always the wrong thing to do, and more often than not it masks errors instead of fixing them. If anything, the code should be fixed to use the correct types instead.

pm100:
@mrdudz I thought a lot about that. But I did not want to launch off on huge changes to a large codebase.

Every change I made is either:
 - casting the return of strlen from size_t to an unsigned or int. Since these worked before and the result fit into a 32 bit value on 32 bit builds they must fit now. The strings in the app have not suddenly got longer
 - casting the result of subtracting 2 pointers. This determines the length of something based on looking at a buffer start and a pointer to bytes further down the buffer. These also fit into a 32 bit value on the 32 bit builds. I cannot imagine that the length of things placed in those  buffers suddenly got so big that their length will not fit in a 32 bit result.

I did think about this. In other cases I would have fixed it differently

Considering how clean this code is (and it is astonishingly clean) I am surprised that it does not use explicit lengths everywhere (int32_t, uint32_t,....). But it does not. That would be the clean solution, and a lot of work.

Compyx:
The correct type for the result of subtracting two pointers is `ptrdiff_t`, that's been part of the standard since C89. For the issue mentioned in #1919, the line `int Type = CD - ConDes;` could indeed be fixed with `ptrdiff_t Type = CD - ConDes;` since that variable is only used inside the function and not used as an argument to other functions. However if the result of subtracting two pointers is used to determine an index, or ID of sorts, for a function then changing the signature of the function to accept `ptrdiff_t index` instead of `int index` or `unsigned int index` would look "unnatural" to me, and casting would be the better solution.

An example I saw of where casting away the warning probably isn't the correct solution is in src/ld65, where the `Size` argument to `WriteData (FILE* F, const void* Data, unsigned Size)` is cast from `size_t` to `unsigned int` in the PR. The correct approach, but a bit more work, would be to change the signature of `WriteData()` to `WriteData (FILE* F, const void* Data, size_t Size)`, since `size_t` is the natural type for sizes.

As for using `<stdint.h>` for the fixed-size types introduced in C99, I was involved in doing that in the VICE codebase, and although it was a lot of work, it did expose some bugs. The code guidelines of CC65 state that although the code must generally assume a C89-compliant compiler, an exception is explicitly made for C99's `<stdint.h>` and `<inttypes.h>`.

mrdudz:
> I am surprised that it does not use explicit lengths everywhere (int32_t, uint32_t,....). But it does not. That would be the clean solution, and a lot of work.

Using them *everywhere* would still be wrong and not "clean" at all. stdint types should be used when a certain size is explicitly needed AND a proper std type does not exist - this is not the case here.

pm100:
@mrdudz Taking the strlen one as the main culprit. I am not sure what you would prefer (assuming this win 64 bit thing should be fixed at all).

For example

```
void ObjWriteStr (const char* S)
/* Write a string to the object file */
{
    unsigned Len = (unsigned)strlen (S);

    /* Write the string with the length preceded (this is easier for
    ** the reading routine than the C format since the length is known in
    ** advance).
    */
    ObjWriteVar (Len);
    ObjWriteData (S, Len);
}
```

I added the unsigned cast. The other solutions are
- write an `xstrlen` that returns unsigned
- write `ObjWriteVarSizet` that accepts `size_t` that duplicates `ObjWriteVar`
- change `ObjWriteVar` to accept `size_t` instead of `unsigned`

(Would need to fix `ObjWriteData` too)
The first 2 are not very attractive. The last one seems the most natural but

```
void ObjWriteBuf (const StrBuf* S)
/* Write a string to the object file */
{
    /* Write the string with the length preceeded (this is easier for
    ** the reading routine than the C format since the length is known in
    ** advance).
    */
    ObjWriteVar (SB_GetLen (S));
    ObjWriteData (SB_GetConstBuf (S), SB_GetLen (S));
}
```

calls `ObjWriteVar` with the result of `SB_GetLen`. So this would mean changing `StrBuf` to have `size_t` for `Len` instead of `unsigned`. Thats a huge change

mrdudz:
> The first 2 are not very attractive. The last one seems the most natural but calls ObjWriteVar with the result of SB_GetLen. So this would mean changing StrBuf to have size_t for Len instead of unsigned. Thats a huge change

Yes. The proper fix for this is a huge change - but would only change types here and there, not introduce any new code, nor use any type casting. And of course should compile without warnings afterwards, on GCC and MSVC, 32 and 64bit.

pm100:
@mrdudz - will do, I also note this change committed 6 years ago https://github.com/cc65/cc65/pull/571 that addresses the same issue. It uses the wrong type as far as I am concerned. uintptr_t suggests that you need a size that is the same size as a pointer , it is typically used in very specific circumstances (that dont apply here). I came accross it when working through the whole code base (git blame is a wonderful thing). I intend to change these to size_t to be consistent with the changes I am making. I will close this PR and open a new one since its a totally different fix now

PR #2288, opened 2023-12-09
[cc65] Fixed an iteration bug in type composition

### LINKED ISSUE #2285: [cc65] Regression in type composition
1de1855b13cae4cdd4b75db9b153520f57aea418 is correct when
```c
void foo(int (*)(int));
void foo(int (*)(long)); /* Error: Conflicting function types for 'foo' */
```
but
```c
void foo();              /* OK */
void foo(int (*)(int));  /* OK */
void foo(int (*)(long)); /* WRONG: Should be an error */
```
The cause of the issue is easy to find out and a quick fix is done already.

acqn:
Fixed #2285.


mrdudz:
thanks!

PR #2289, opened 2023-12-09
[cc65] Fixed the issue that qualifiers of pointees of function parameters were ignored for type compatibility check

### LINKED ISSUE #2286: [cc65] Qualifiers of pointees of function parameters ignored for type compatibility check
```c
void woo(int* p);
void woo(const int* p); /* WRONG: Should be an error */
```

acqn:
Fixed #2286.


mrdudz:
thanks!

PR #2291, opened 2023-12-09
Fix bug introduced in #2260

colinleroy:
bne should have applied to A, not X, but adding a cmp #$00 before would make that change less optimized than the existing: revert this part.
The bug is visible with the added unit test. (256-1 = 511)


colinleroy:
cc @mrdudz

mrdudz:
good catch, thanks!

Issue #2292, opened 2023-12-09
UB in xvsnprintf when using %n


pm100:
Spotted during work on making msvc x64 work.
```
static void StoreOffset (PrintfCtrl* P)
/* Store the current output offset (%n format spec) */
{
    switch (P->LengthMod) {
        case lmChar:     *va_arg (P->ap, int*)       = P->BufFill; break;
```

this code treats a pointer to a char as a pointer to an int

demo test failure
```
    StrBuf sb = AUTO_STRBUF_INITIALIZER;
    char x[10];
    x[1] = 42;
    SB_Printf (&sb, "%d %hhn\n", 42 , &x[0]);
    assert (x[1] == 42);
```
x[1] [2] and [3] get overwritten by 0

Same error for short

This code is never used but still should be fixed

mrdudz:
fixed via [9985ee7](https://github.com/cc65/cc65/commit/9985ee7f61f9c0e1b1256d16e3fb901c87531019)

PR #2293, opened 2023-12-09
fix UB in xvsnprintf

pm100:
fix for #2292


mrdudz:
Whats the deal with all those typecasts? When they were not necessary before, why are they now?

pm100:
Needed for char and short and probably for some win64 weirdness for others, did not test. I put them in for all since it makes it very clear and symmetrical.

PR #2262, opened 2023-11-13
Optimize static long assignment a bit

colinleroy:
I've worked a bit on optimizing the basics about longs: assignments.
It gives something a little larger without -O (because I've simplified that part in codegen.c to ease matching the pattern in coptlong.c), but I think it's worth it as the results with -O are quite good ?

Before/after without -O
image

Before/after with -O
image

PR #2295, opened 2023-12-10
Fix #2262: Make sure there's no branching after the sequence

colinleroy:
Also better check that arguments match.

### DIFF
--- a/src/cc65/coptlong.c
+++ b/src/cc65/coptlong.c
@@ -49,24 +49,30 @@
 /*                      Remove unused loads and stores                       */
 /*****************************************************************************/
 
-
-
 unsigned OptLongAssign (CodeSeg* S)
 /* Simplify long assignments.
 ** Recognize
-**      lda     ...       0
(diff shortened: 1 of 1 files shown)


mrdudz:
Any idea how to craft a proper test for that bug? Would be really nice to have that :)

colinleroy:
> Any idea how to craft a proper test for that bug? Would be really nice to have that :)

I couldn't reproduce it with any construct on the main branch, so not really :/ You could add one in the float branch, though?

mrdudz:
I have found it because a test in the float branch broke :)

Working on a test using that inline asm snippet - will commit later

mrdudz:
https://github.com/cc65/cc65/commit/b31a1c7c0c25de087e272c045eb2370a03856c93

cheers :)

colinleroy:
Oh yeah that works too! Thanks!

mrdudz:
There should probably more of this kind of tests for the optimizer ... :)

PR #2297, opened 2023-12-11
[cc65] Fixed wrapped call when the function was defined before its later declaration gets wrapped with the pragma

acqn:


mrdudz:
Thanks!

PR #2299, opened 2023-12-11
Apple2: Fix BSS obliteration by mliparam during exec()

colinleroy:
Hello @oliverschmidt,

It's me again, the fucked-up-bug-finder. (This time it only took me a whole Sunday to figure it out, so there's progress!).

When exec()'ing a new program, we set the IO buffer to copy it from disk at $BB00, as seen in exec.s lines 112 to 126:

```
        ; No BASIC.SYSTEM so use addr of ProDOS system global page
prodos: ldx     #>$BF00         ; High byte

        ; The I/O buffer needs to be page aligned
setbuf: lda     #$00            ; Low byte

        ; The I/O buffer needs four pages
        dex
        dex
        dex
        dex

        ; Set I/O buffer
        sta     mliparam + MLI::OPEN::IO_BUFFER
        stx     mliparam + MLI::OPEN::IO_BUFFER+1
```

However, we're not yet completely done with BSS at that point, and if the BSS segment ends after $BB00, problems ensue, especially if... mliparam is over there.

(My BSS segment ends after $BB00 in some rather fat programs, where I specify HIMEM to $BEFF in the linker cfg file, because I only use statics inside, want to spare the malloc/free footprint, and want to make sure that if it builds, it'll run without an out-of-memory).

In this reproducer, a.bin's BSS ends at $BB9C, and mliparam is at $BB8A:
```
$ grep ^BSS a.map
BSS                   001D52  00BB9C  009E4B  00001
$ grep mliparam a.map
ldaxysp                   00178B RLA    mliparam                  00BB8A RLA    
```

$BB8A is a good value to see a beautiful brk-based explosion of b.bin, dropping to the monitor. Other values may cause less visible issues.

There are two BSS variables used and potentially destroyed after setting the ProDOS IO buffer. First is `level`, backing up the system LEVEL variable. Here is an extract of what happens to it between lines 134, where it is saved, and 144, where it is restored:

```
00244446 A=00,X=BB,Y=FF 1604: ldx $bf94 ; adr: RAM.LC1.1604     ldx LEVEL     (apple2/exec.s:133)
00244447 A=00,X=00,Y=FF 1607: stx $bb87 ; adr: RAM.LC1.1607     stx level     (apple2/exec.s:134)

[...] ; Here we do the OPEN MLI call

00475496 A=00,X=00,Y=FF 1616: ldx $bb87 ; adr: RAM.LC1.1616     ldx level     (apple2/exec.s:144)
00475497 A=00,X=20,Y=FF 1619: stx $bf94 ; adr: RAM.LC1.1619     stx LEVEL     (apple2/exec.s:145)
```
`level` in X is now $20 instead of $00.

The other one is mliparam itself.
We now go through crt0's `_exit`, `_donelib`, etc and jump to `done`. `done` has been set in exec.s:191 to `target`, which has been copied from `source` at exec.s:65. `source` is our loader stub, and it copies the whole new program using the buffer at $BB00, to $1000 (the start address), via a READ MLI call.
However, I think that the ProDOS OPEN mli call has already read and cached the first 256 bytes into the buffer; then, the mliparam struct has been modified during that call, and it is in that buffer; So, either the first block is corrupted when it is copied, or the mliparam struct is corrupted by the copy. I have not *exactly* understood what happens, it's deep in ProDOS and it's a terrible headache to try to understand it. But it breaks, and bytes are wrong around $1090 after b.bin is loaded into memory and starts.

For `level`, so solution is quite obvious and I replaced the backup `level` variable with a pha/pla.

For `mliparam`, it is more complicated, and I have replaced the OPEN callmli with a raw jsr $BF00. 
Another option would be to move mliparam to the DATA segment. I remember this was "an option, but not the preferred one" from #2160's discussion, because for #2160, we wanted to avoid making every ProDOS-using program 18 bytes larger, and only do it for getcwd()-using program. Here it is the same, only exec()-using programs will be affected.

I have avoided saving and restoring the random counter, as it would take more room, and anyway it's not saved/restored in the `target` loader stuff that is going to run just after.

This costs 11 bytes, both on disk and in RAM.

### DIFF
--- a/libsrc/apple2/exec.s
+++ b/libsrc/apple2/exec.s
@@ -121,35 +121,9 @@ setbuf: lda     #$00            ; Low byte
         dex
         dex
 
-        ; Set I/O buffer
-        sta     mliparam + MLI::OPEN::IO_BUFFER
-        stx     mliparam + MLI::OPEN::IO_BUFFER+1
-
-        ; PATHNAME already set
-        .assert MLI::OPEN::PATHNAME = MLI::INFO::PATHNAME, error
-
-        ; Lower file level to avoid program file
-        ; being closed by C library shutdown code
-        ldx     LEVEL
-        stx     level
-        beq     :+
-        dec     LEVEL
-
-        ; Open file
-:       lda     #OPEN_CALL
-        ldx     #OPEN_COUNT
-        jsr     callmli
-
-        ; Restore file level
-        ldx     level
-        stx     LEVEL
-        bcc     :+
-        jmp     oserr
-
-        ; Get and save fd
-:       lda     mliparam + MLI::OPEN::REF_NUM
-        sta     read_ref
-        sta     close_ref
+        ; Set OPEN MLI call I/O buffer parameter
+        sta     io_buffer
+        stx     io_buffer+1
 
         .ifdef  __APPLE2ENH__
         ; Calling the 80 column firmware needs the ROM switched
@@ -194,14 +168,25 @@ setbuf: lda     #$00            ; Low byte
         ; Initiate C library shutdown
         jmp     _exit
 
-        .bss
-
-level : .res    1
-
         .rodata
 
+source:
+        ; Open program file
+        ; PATHNAME parameter is already set (we reuse
+        ; the copy at $0280); IO_BUFFER has been setup
+        ; before shutting down the C library
+        jsr     $BF00
+        .byte   OPEN_CALL
+        .word   open_param
+        bcs     error
+
+        ; Copy REF_NUM to MLI READ and CLOSE parameters
+        lda     open_ref
+        sta     read_ref
+        sta     close_ref
+
         ; Read whole program file
-source: jsr     $BF00
+        jsr     $BF00
         .byte   READ_CALL
         .word   read_param
         bcs     error
@@ -254,6 +239,14 @@ jump:   jmp     (data_buffer)
 file_type       = * - source + target
         .byte   $00
 
+open_param      = * - source + target
+        .byte   $03             ; PARAM_COUNT
+        .addr   $0280           ; PATHNAME
+io_buffer       = * - source + target
+        .addr   $0000           ; IO_BUFFER
+open_ref        = * - source + target
+        .byte   $00             ; REF_NUM
+
 read_param      = * - source + target
         .byte   $04             ; PARAM_COUNT
 read_ref        = * - source + target
@@ -284,5 +277,9 @@ quit_param      = * - source + target
 size            = * - source
 
 target          = DOSWARM - size
+
+        ; Make sure that the loader isn't too big, and
+        ; fits in $300-$3D0
+        .assert target >= $300, error
 
 dosvec: jmp     quit


colinleroy:
[test-mli-bss-obliteration.tar.gz](https://github.com/cc65/cc65/files/13638724/test-mli-bss-obliteration.tar.gz)
This archive provides a reproducer. There are three large char buffers defined.
- The first one's size pushes mliparam to $BB8A when compiled with cc65 revision 9634cc4d1ec970f2073086768a983c029f7371da. It provides an monitor-based in-flight explosion of b.bin right after exec().
- The second one pushes mliparam and the end of BSS back to the safe zone, under $BB00, and everything functions correctly with the current cc65 master branch.
- The third one pushes mliparam back to $BB8A when built using cc65 from this branch, and everything functions correctly.

Hope this helps!

colinleroy:
Thanks for the quick reply! 
That's an alternative I've considered, but didn't try at first because it was more complicated, and I wondered about the loader size : is there a maximum?

oliverschmidt:
Yeah, it's more complex but it's the only correct way.

Regarding the loader size. Maybe you have already wondered about those strange `* - source + target` constructs. The reason is that the loader is not loaded _above_ a certain address but _below_ a certain address. So the load address has to be determined by the size of the loader.

Why that effort? Small routines extending the system are sometimes loaded at $300. My idea was/is to make the loader as compatible as possible with those routines. Therefore it gets placed right below the end of that free $300 area. This is $3D0.

It would be a good idea to add an ASSERT that it is placed (the symbol is `target`) now lower than $300.

So to finally answer your actual question: The maximum is $D0. If if gets (very) close to the maximum, one can consider to remove those `* - source + target` constructs and place it at $300 in the first place.

colinleroy:
OK, thanks! Are you OK if I do some size optimizations to compensate the extra weight, that may impact legibility a little bit, as long as I comment?

oliverschmidt:
1. The cc65 C library is always primarily about size. So in general, size reduction is always (more than) welcome!

2. I thought more about what I wrote above 'Patch (aka copy) the file name into page 3.'. This can't work that way: A Prodos filename is up to 64 bytes (plus a length byte) - and of course we can't spare that space in page 3.

The pushname routine stores the file name on the cc65 "software" stack. I don't see why it should be destroyed there by the C library shutdown. So the approach is to copy it from there after the C library shutdown but before opening the file - which usually destroys the cc65 "software" stack as you learned the hard way.

So I see the loader code start with a simple 65 byte copy routine. The source address is patched by the exec() code running before the C library shutdown. The target address is fixed to let's say $800 because the "software" stack is never located there meaning that source and target of the copy operation never overlap.

The file open operation has then the file name at $800 and the io-buffer at $BB00 (or similar). By the time the file read operation might overwrite the file name, it isn't necessary anymore.

colinleroy:
I was thinking of reusing the filename from $0280 in fact, I suppose it should work?

oliverschmidt:
As far as I can see, you're right. And of course it's preferrable if it works!

colinleroy:
Here you are! The little bonus is that the size "cost" of this is a *gain* of 15 bytes !
Your solution is more satisfying and everythings falls in place :)

oliverschmidt:
> [...] and everythings falls in place :)

Yes, this looks indeed very nice - thanks for the contribution :-)

Merging...

PR #2305, opened 2023-12-13
[cc65] Fixed missing diagnosis on function parameter lists with trailing commas

### LINKED ISSUE #2301: cc65: Invalid parameter handling for functions
The following code should not compile but does:

```
int foo(int a,);
```

acqn:
#Fixed #2301.


mrdudz:
thanks!

PR #2307, opened 2023-12-13
[cc65] Fixed function declarator parser when a parameter has a function type

### LINKED ISSUE #2302: [cc65] Parameters of function types not parsed correctly
```c
void f(int (int));   /* Should be OK: void f(int (*)(int)) but cc65 fails */

typedef int A;
void g(int (*)(A));  /* Correct: void g(int (*)(int)) */
void g(int (A));     /* Wrong:   void g(int A), should be of the same type as above */
```

### LINKED ISSUE #2303: [cc65] Parameters with no type specified in K&R-style function declarations
> An identifier list in a function declarator that is not part of a function definition shall be empty.

```
int f(a); /* Should be an error */
```
Clarification: this currently doesn't compile so there is not such a "bug". However, a fix for some other issues (eg. #2302) could probably "enable" this wrong code. Moreover the current diagnostic message doesn't reflect what is actually wrong with this code. So this is a "TODO" issue.

acqn:
Fixed #2302.
Fixed #2303.

### DIFF
--- a/src/cc65/compile.c
+++ b/src/cc65/compile.c
@@ -206,6 +206,14 @@ static void Parse (void)
                 } else {
                     /* Just a declaration */
                     Decl.StorageClass |= SC_DECL;
+
+                    FuncDef = GetFuncDesc (Decl.Type);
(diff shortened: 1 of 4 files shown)


mrdudz:
Thanks!

PR #2309, opened 2023-12-14
[cc65] Added warning on static functions that are used but not defined

acqn:
Also shortened the error message text on tentative declarations of incomplete types.
```c
struct S s;
static void f(void); /* No warning if 'f' were never used */

int main(void)
{
    f();
}
```
Before:
```
test.c:8: Error: Definition of 's' has type 'struct S' that is never completed
```
After:
```
test.c:8: Error: Definition of 's' never has its type 'struct S' completed
test.c:8: Warning: Static function 'f' used but never defined
```


mrdudz:
Thanks!

PR #2310, opened 2023-12-14
second try at fixing win64 build

pm100:
This is step a) No code changes, just suppress the warnings that are also suppressed by gcc. No code changes

It also fixes a truly hard to resolve issue. GetFileType duplicates the name of a Window API. The 64 bit linker was linking the Windows API not the entry in common. The Win32 call has sufficiently similar arguments so it worked (did not crash), but returned the wrong answer. I changed the name to GetTypeOfFile. This change has to be part of this fix, otherwise all the tests fail.

I also added 64 bit builds to the github action


mrdudz:
Thanks!

PR #2311, opened 2023-12-17
ca65 struct and union documentation improvement and feature testing.

bbbradsmith:
This adds a test for documented behaviours of .struct and .union.

I've tried to add clarification and extra examples to the documentation about these features. Nested structures especially needed some clarification.

`.DBYT` was incorrectly documented as `.DBYTE`.


mrdudz:
Thanks!

PR #2313, opened 2023-12-18
initial commit of printf rewrite

pm100:
Note, I do not expect this PR to be accepted as is.

This is a rewrite of _printf.s. I decided to fix all the bugs in the todo test directory, started by tweaking the existing code but decided a rewrite was needed. I kept the housekeeping code (prologue, exit, handle varargs, strtol and a few more), but the code format parse and execute is new.

It passes all tests in the todo directory including the ones excluded by the NOCRASH define. I also found another huge printf test here https://github.com/BartMassey/printf-tests. It passes a total of 700 tests.

It is a little bloated at the moment. I will trim it down. Also at the moment its called newprint.s, I will replace the original file eventually. For the moment I just stopped _printf.s from exporting __printf

I have questions

a) are you ok with the basic idea of this rewrite 

b) I created a new macpack to do u16 operations. I wonder if it would be better to use the runtime functions instead. I would welcome pointers to what functions I should use. Big ones are u16_add, u16_sub, u16_gt

c) I see that PCE has its own copy of _printf.s. I think it would be better to have only one implementation and to have #defines controlling the platform specifics. I have no way to test this code though. DO you have pointers to how I could test it on PCE

d) I have a test set for the macpack. I can add that. Or I could not make a macpack and just use uin16.mac as a private .inc

### DIFF
--- a/asminc/uint16.mac
+++ b/asminc/uint16.mac
@@ -5,13 +5,13 @@
 
 .macro  u16_add arg1, arg2
     .local @out
-        clc        
-        lda     arg1   
+        clc
+        lda     arg1
     .if (.match (.left (1, {arg2}), #))
       .if >(.right (.tcount ({arg2})-1, {arg2}))=0
         ; arg2 = single byte immediate
         adc     arg2
-        sta     arg1   
+        sta     arg1
         bcc     @out
         inc     arg1+1
       .else
@@ -25,13 +25,13 @@
     .else
         ; arg2 = word
         adc     arg2
-        sta     arg1   
+        sta     arg1
         lda     arg1+1
         adc     arg2+2
-        sta     arg1+1 
+        sta     arg1+1
     .endif
     @out:
-.endmacro     
+.endmacro
 
 ;    arg1 -= arg2
 ;    arg1 is u16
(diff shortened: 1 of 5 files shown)


pm100:
I assume the RODATA overflow of RAM in some samples (which I did not try before) is due to printf being too large. I know its big, this now gives me something to shoot for

pm100:
dang - still too big for some tests

mrdudz:
You can run the tests locally too, you know... :)

I am not a big friend of all those macros to be honest, they usually just make the code harder to read and optimize.

That said, we might need different versions of printf... a full featured one is probably too big and not needed either for many practical purposes (and it will get even worse when throwing floats into it...)

polluks:
About c) https://cc65.github.io/doc/intro.html#ss6.10

pm100:
@polluks thanks i will check that out

@mrdudz Ironically I agree about the macros disrupting the reading flow. But the same code appears over and over again , let me stew on it for a while see if i can come up with better macro names (like the existing macpack ones, jeq, add,....)

Yes I was thinking that a small stdio (I assume scanf is either incomplete or huge - I havent looked) or something like that. 

ie `cl65 myproc.c stdio_small.lib` to use the small version, or maybe small is the default one and you have to ask for big

would need to decide whats left out for the small version
- certainly %n
- precision (thats the main missing piece today)
- float support

A good starting point tho is a printf implementation thats 100% feature complete.

on xmas vac - resume in jan. Merry xmas all

PR #2315, opened 2023-12-20
Apple2: Implement statvfs() and stat() 

colinleroy:
Edit of the description: following below comments, this PR is now implementing stat() and statvfs() for the APPLE2 platform.

### DIFF
--- a/doc/funcref.sgml
+++ b/doc/funcref.sgml
@@ -93,6 +93,8 @@ function.
 <itemize>
 <item>_dos_type
 <item><ref id="get_ostype" name="get_ostype">
+<item><ref id="gmtime_dt" name="gmtime_dt">
+<item><ref id="mktime_dt" name="mktime_dt">
 <item>rebootafterexit
 </itemize>
 
@@ -844,6 +846,20 @@ communication, see also <tt>testcode/lib/ser-test.c</tt>.
 </itemize>
 
 (incomplete)
+
+
+<sect1><tt/stat.h/<label id="sys/stat.h"><p>
+
+<itemize>
+<item><ref id="stat" name="stat">
+</itemize>
+
+
+<sect1><tt/statvfs.h/<label id="sys/statvfs.h"><p>
+
+<itemize>
+<item><ref id="statvfs" name="statvfs">
+</itemize>
 
 
 <sect1><tt/vic20.h/<label id="vic20.h"><p>
@@ -2847,6 +2863,79 @@ setting the time may not work. See also the platform-specific information.
 <ref id="clock_getres" name="clock_getres">,
 <ref id="clock_gettime" name="clock_gettime">
 <tag/Example/None.
+</descrip>
+</quote>
+
+
+<sect1>gmtime_dt<label id="gmtime_dt"><p>
+
+<quote>
+<descrip>
+<tag/Function/Converts a ProDOS date to a struct tm.
+<tag/Header/<tt/<ref id="apple2.h" name="apple2.h">/
+<tag/Declaration/<tt/struct tm* __fastcall__ gmtime_dt (const struct datetime* dt);/
+<tag/Description/The <tt/gmtime_dt/ function converts the given
+proDOS date/time to a struct tm. On error, NULL is returned and <tt/errno/ is set
+to an error code describing the reason for the failure.
+<tag/Notes/<itemize>
+<item>The function is only available as fastcall function, so it may only
+be used in presence of a prototype.
+<item>This function is only available on Apple II.
+<item>On Apple II, you can't stat() an opened file. stat() before opening.
+</itemize>
+<tag/Availability/cc65
+<tag/Example/
+<verb>
+
+#include <stdio.h>
+#include <time.h>
+#include <sys/stat.h>
+int main(void)
+{
+    struct stat st;
+    struct tm* tm;
+    if (stat ("/disk/file", &amp;st) == 0) {
+        tm = gmtime_dt (&amp;st.st_ctime);
+        if (tm)
+            printf ("File created on %s\n", asctime(tm));
+    }
+}
+</verb>
+</descrip>
+</quote>
+
+
+<sect1>mktime_dt<label id="mktime_dt"><p>
+
+<quote>
+<descrip>
+<tag/Function/Converts a ProDOS date to a time_t.
+<tag/Header/<tt/<ref id="apple2.h" name="apple2.h">/
+<tag/Declaration/<tt/time_t __fastcall__ mktime_dt (const struct datetime* dt);/
+<tag/Description/The <tt/mktime_dt/ function parses the given
+proDOS date/time and returns a time_t timestamp. On error, 0 is returned,
+and errno is set.
+<tag/Notes/<itemize>
+<item>The function is only available as fastcall function, so it may only
+be used in presence of a prototype.
+<item>This function is only available on Apple II.
+</itemize>
+<tag/Availability/cc65
+<tag/Example/
+<verb>
+
+#include <stdio.h>
+#include <time.h>
+#include <sys/stat.h>
+int main(void)
+{
+    struct stat st;
+    if (stat ("/disk/file", &amp;st) == 0) {
+        printf ("File created on %s\n",
+          localtime (mktime_dt (&amp;st.st_ctime)));
+    }
+}
+</verb>
 </descrip>
 </quote>
 
@@ -6229,6 +6318,9 @@ be used in presence of a prototype.
 <item>The returned pointer may point to a statically allocated instance of
 <tt/struct dirent/, so it may get overwritten by subsequent calls to
 <tt/readdir/.
+<item>On the Apple II platform, the d_ctime and d_mtime returned are in the
+ProDOS format. You can convert them to more portable time representations using
+the ProDOS datetime conversion functions.
 <item>On several platforms, namely the CBMs and the Atari, the disk drives get
 confused when opening/closing files between directory reads. So for example a
 program that reads the list of files on a disk, and after each call to
@@ -7071,6 +7163,85 @@ be used in presence of a prototype.
 <ref id="_randomize" name="_randomize">,
 <ref id="rand" name="rand">
 <tag/Example/None.
+</descrip>
+</quote>
+
+
+<sect1>stat<label id="stat"><p>
+
+<quote>
+<descrip>
+<tag/Function/Get file status.
+<tag/Header/<tt/<ref id="sys/stat.h" name="sys/stat.h">/
+<tag/Declaration/<tt/int __fastcall__ stat (const char* pathname, struct stat* statbuf);/
+<tag/Description/<tt/stat/ gets information for the file with the given name. On success,
+zero is returned. On error, -1 is returned and <tt/errno/ is set to an error
+code describing the reason for the failure.
+<tag/Notes/<itemize>
+<item>The function is only available as fastcall function, so it may only
+be used in presence of a prototype.
+<item>On the Apple II platform, the st_ctim, st_mtim and st_atim members are left
+to zero, for size and performance reasons. The ProDOS creation and modification dates
+are returned in the ProDOS format in st_ctime and st_mtime. The access date does
+not exist. You can convert them to POSIX-style time representations using
+the <url url="apple2.html#ss9.3" name="ProDOS datetime conversion functions">.
+</itemize>
+<tag/Availability/POSIX 1003.1
+<tag/See also/
+<ref id="statvfs" name="statvfs">
+<tag/Example/
+<verb>
+#include &lt;sys/stat.h&gt;
+
+#define FILENAME "helloworld"
+struct stat stbuf;
+if (stat (FILENAME, &amp;stbuf) == 0) {
+    printf ("%s size is %lu bytes (created on %s)\n", FILENAME, stbuf.st_size,
+#ifndef __APPLE2__
+    localtime (&amp;stbuf.st_ctim.tv_sec)
+#else
+    localtime (mktime_dt (&amp;stbuf.st_ctime))
+#endif
+    );
+} else {
+    printf ("There was a problem stat'ing %s: %d\n", FILENAME, errno);
+}
+</verb>
+</descrip>
+</quote>
+
+
+<sect1>statvfs<label id="statvfs"><p>
+
+<quote>
+<descrip>
+<tag/Function/Get filesystem statistics.
+<tag/Header/<tt/<ref id="sys/statvfs.h" name="sys/statvfs.h">/
+<tag/Declaration/<tt/int __fastcall__ statvfs (const char* pathname, struct statvfs* buf);/
+<tag/Description/<tt/statvfs/ gets information for the filesytem on which the given file
+resides. On success,
+zero is returned. On error, -1 is returned and <tt/errno/ is set to an error
+code describing the reason for the failure.
+<tag/Notes/<itemize>
+<item>The function is only available as fastcall function, so it may only
+be used in presence of a prototype.
+<item>The function requires an absolute pathname.
+</itemize>
+<tag/Availability/POSIX 1003.1
+<tag/See also/
+<ref id="stat" name="stat">
+<tag/Example/
+<verb>
+#include &lt;sys/statvfs.h&gt;
+
+#define FILENAME "/disk/helloworld"
+struct statvfs stvbuf;
+if (statvfs (FILENAME, &amp;stvbuf) == 0) {
+    printf ("%s filesystem has %u blocks of %u size, %u of them free.\n", FILENAME, stvbuf.f_blocks, stvbuf.f_bsize, stvbuf.f_bfree);
+} else {
+    printf ("There was a problem statvfs'ing %s: %d\n", FILENAME, errno);
+}
+</verb>
 </descrip>
 </quote>
 
--- a/include/apple2.h
+++ b/include/apple2.h
@@ -41,6 +41,7 @@
 #  error This module may only be used when compiling for the Apple ][!
 #endif
 
+#include <time.h>
 #include <apple2_filetype.h>
 
 
@@ -142,6 +143,27 @@ extern unsigned char _dos_type;
 ** ProDOS 8 2.4.x - 0x24
 */
 
+/* struct stat.st_mode values */
+#define S_IFDIR  0x01
+#define S_IFREG  0x02
+#define S_IFBLK  0xFF
+#define S_IFCHR  0xFF
+#define S_IFIFO  0xFF
+#define S_IFLNK  0xFF
+#define S_IFSOCK 0xFF
+
+struct datetime {
+    struct {
+        unsigned day  :5;
+        unsigned mon  :4;
+        unsigned year :7;
+    }                 date;
+    struct {
+        unsigned char min;
+        unsigned char hour;
+    }                 time;
+};
+
 
 
 /*****************************************************************************/
@@ -151,20 +173,10 @@ extern unsigned char _dos_type;
 
 
 /* The file stream implementation and the POSIX I/O functions will use the
-** following struct to set the date and time stamp on files. This specificially
+** following struct to set the date and time stamp on files. This specifically
 ** applies to the open and fopen functions.
 */
-extern struct {
-    struct {
-        unsigned day  :5;
-        unsigned mon  :4;
-        unsigned year :7;
-    }             createdate;    /* Current date: 0 */
-    struct {
-        unsigned char min;
-        unsigned char hour;
-    }             createtime;    /* Current time: 0 */
-} _datetime;
+extern struct datetime _datetime;
 
 /* The addresses of the static drivers */
 #if !defined(__APPLE2ENH__)
@@ -210,6 +222,12 @@ void rebootafterexit (void);
 #define _bordercolor(color)     COLOR_BLACK
 #define _cpeekcolor()           COLOR_WHITE
 #define _cpeekrevers()          0
+
+struct tm* __fastcall__ gmtime_dt (const struct datetime* dt);
+/* Converts a ProDOS date/time structure to a struct tm */
+
+time_t __fastcall__ mktime_dt (const struct datetime* dt);
+/* Converts a ProDOS date/time structure to a time_t UNIX timestamp */
 
 
 
--- a/include/time.h
+++ b/include/time.h
@@ -37,6 +37,15 @@
 #define _TIME_H
 
 
+/* Forward declaration for target.h */
+typedef unsigned long time_t;
+typedef unsigned long clock_t;
+
+
+
+#include <target.h>
+
+
 
 /* NULL pointer */
 #ifndef NULL
@@ -48,9 +57,6 @@
 #define _HAVE_size_t
 typedef unsigned size_t;
 #endif
-
-typedef unsigned long time_t;
-typedef unsigned long clock_t;
 
 /* Structure for broken down time */
 struct tm {
--- a/libsrc/apple2/filename.s
+++ b/libsrc/apple2/filename.s
@@ -8,6 +8,7 @@
         .import         subysp, addysp, decsp1
 
         .include        "zeropage.inc"
+        .include        "apple2.inc"
         .include        "mli.inc"
 
 pushname:
@@ -15,7 +16,7 @@ pushname:
         stx     ptr1+1
 
         ; Alloc pathname buffer
-        ldy     #64+1           ; Max pathname length + zero
+        ldy     #FILENAME_MAX
         jsr     subysp
 
         ; Check for full pathname
@@ -71,14 +72,14 @@ copy:   lda     (ptr1),y
         sta     (sp),y
         beq     setlen
         iny
-        cpy     #64+1           ; Max pathname length + zero
+        cpy     #FILENAME_MAX
         bcc     copy
 
         ; Load oserror code
         lda     #$40            ; "Invalid pathname"
 
         ; Free pathname buffer
-addsp65:ldy     #64+1
+addsp65:ldy     #FILENAME_MAX
         bne     addsp           ; Branch always
 
         ; Alloc and set length byte
@@ -93,5 +94,5 @@ setlen: tya
 
 popname:
         ; Cleanup stack
-        ldy     #1 + 64+1       ; Length byte + max pathname length + zero
-addsp:  jmp     addysp          ; Preserves A
+        ldy     #1 + FILENAME_MAX
+addsp:  jmp     addysp          ; Preserves A and X
--- a/libsrc/apple2/mli.inc
+++ b/libsrc/apple2/mli.inc
@@ -83,8 +83,8 @@ EOF_COUNT       = 2
                                 AUX_TYPE        .word
                                 STORAGE_TYPE    .byte
                                 BLOCKS          .word
-                                MODE_DATE       .word
-                                MODE_TIME       .word
+                                MOD_DATE        .word
+                                MOD_TIME        .word
                                 CREATE_DATE     .word
                                 CREATE_TIME     .word
                         .endstruct
@@ -139,3 +139,6 @@ LEVEL   :=      $BF94           ; File level: used in open, flush, close
 MACHID  :=      $BF98           ; Machine identification
 PFIXPTR :=      $BF9A           ; If = 0, no prefix active
 KVERSION:=      $BFFF           ; Kernel version number
+
+; Max filename length
+FILENAME_MAX =  64+1
--- a/libsrc/apple2/mli_file_info.s
+++ b/libsrc/apple2/mli_file_info.s
@@ -1,0 +1,33 @@
+;
+; Colin Leroy-Mira, 2023 <colin@colino.net>
+;
+
+        .export         mli_file_info
+        .import         pushname, popname, mli_file_info_direct
+        .import         popax
+        .include        "zeropage.inc"
+        .include        "errno.inc"
+        .include        "mli.inc"
+
+        ; Calls ProDOS MLI GET_FILE_INFO on the filename
+        ; stored as C string in AX at top of stack
+        ; Returns with carry set on error, and sets errno
+mli_file_info:
+        ; Get pathname
+        jsr     popax
+        jsr     pushname
+        bne     oserr
+
+        jsr     mli_file_info_direct
+        php                     ; Save return status
+
+        jsr     popname         ; Preserves A
+
+        plp
+        bcs     oserr
+        rts
+
+oserr:
+        jsr     ___mappederrno
+        sec
+        rts
--- a/libsrc/apple2/open.s
+++ b/libsrc/apple2/open.s
@@ -18,6 +18,7 @@
         .include        "fcntl.inc"
         .include        "mli.inc"
         .include        "filedes.inc"
+        .include        "time.inc"
 
         .segment        "ONCE"
 
--- a/libsrc/apple2/stat.s
+++ b/libsrc/apple2/stat.s
@@ -1,0 +1,129 @@
+;
+; Colin Leroy-Mira, 2023 <colin@colino.net>
+;
+; int __fastcall__ stat(const char *pathname, struct stat *statbuf);
+;
+
+        .export         _stat
+        .import         __errno, _open,_close
+        .import         mli_file_info
+        .import         popax, pushax, pusha0, incsp2
+        .include        "zeropage.inc"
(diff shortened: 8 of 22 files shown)


oliverschmidt:
Hi,

First of all sorry for the delayed response, but the time before xmas way (too) busy.

I - of course - like the idea to have this functionality made available. However, I must admit that I'm rather unhappy about the way it is made available...

The DIO API is about "media". I refer to this term in the context of this layering - ordered from low layer to high layer:

1. Device/Drive
2. Media
3. File System

It's not about "device" as things like sector count (and even sector size) can be different for different media types which work in the same drive. But it's also not about file system as sectors are in general below the file system layer.

What might be confusing is that the dio_query_sectcount() implementation on the Apple II uses the ProDOS file system. But this is just an implementation detail. I've done it this way because any other approach would have caused quite some issues without providing tangible benefits.

But in contrast to the sector count, the number of free sectors is totally clearly a file system property - and therefore doesn't belong to DIO.

Instead of inventing a new function, I suggest to implement statvfs(). My proposal is to set all fields of `struct statvfs` to zero but:

* f_bsize: $256
* f_frsize: $256
* f_blocks: same as dio_query_sectcount()
* f_bfree: "the" thing
* f_namemax: same as FILENAME_MAX

Furthermore I propose to deviate from https://pubs.opengroup.org/onlinepubs/009695299/functions/statvfs.html in that `path` doesn't work for any file name on a volume (which might be relative and would require fiddling with PREFIX then etc.), but only for the volume name itself with a prepended slash as it is i.e. returned from getdevicedir(). Or - if you prefer - for any file name as long as it is absolute.

The idea is to call `ON_LINE` with `unit_num` set to 0 and compare the results with (a part of) `path`.

For course this is way more heavyweight than your original idea. But beside being a standard it also has the benefit that a program like the one that made you implement the feature doesn't need to know anything about volumes / units or alike.

In case the user want's to get an overview on the free sectors on all volumes - or a program seeks for a volume with certain amount of free sectors, the approach would be combine those functions:

1. getfirstdevice()/getnextdevice()
2. getdevicedir()
3. statvfs()

Regards,
Oliver

colinleroy:
Hi Oliver!
seems like a very good idea, thanks!

oliverschmidt:
> I didn't add ifdefs around the .h declaration because the other dio_* functions don't have them even though they're not implemented for each platform.

That's totally normal for all kinds of functions. It's up to the cc65 user to interpret the linker error accordingly.

BTW: If you haven't signed up for https://cc65.github.io/mailing-lists.html#cc65-devel yet I'd suggest that you do so and - when this PR is merged - post there about you having contributed a statvfs() for the Apple II and that others might want to do the same for e.g. CBM or Atari machines.

colinleroy:
Will do! Thanks.
what do you think about implementing stat() too while I'm at it, or at least a subset of it?

colinleroy:
Another question, would it be good to use iobuf_alloc/free instead of malloc/free for those and dio_querysectcount? They'd return too big buffers, but users of iobuf0800 wouldn't have to worry about extra allocation?
or would you rather not?

oliverschmidt:
I haven't looked at your actual free space logic yet but I want to make sure that you are aware of the following statement regarding the GET_FILE_INFO call:

_When file information about a volume directory is requested, the total number of blocks on the volume is returned in the aux_type field and the total blocks for all files is returned in blocks_used._

colinleroy:
> I haven't looked at your actual free space logic yet but I want to make sure that you are aware of the following statement in the ProDOS TechRefMan regarding the GET_FILE_INFO call:
> 
> _When file information about a volume directory is requested, the total number of blocks on the volume is returned in the aux_type field and the total blocks for all files is returned in blocks_used._

I was not, and it seems so much easier (and faster) than iterating over the bits of the bytes of the bitmap blocks 😅

colinleroy:
By the way, this thing about GET_FILE_INFO makes the ON_LINE (unitnum=0) call you mention in your first comment unnecessary, doesn't it? Using GET_FILE_INFO will allow implementing both statvfs and stat (with an extra GET_EOF call to get a file's precise size in bytes) easy.

oliverschmidt:
> By the way, this thing about GET_FILE_INFO makes the ON_LINE (unitnum=0) call you mention in your first comment unnecessary, doesn't it?

You're of course right. I got things mixed up. My proposal to have statvfs() either take a slash-prefixed volume name (or an absolute path containing such a slash-prefixed volume name) fits perfectly the GET_FILE_INFO parameter. At some point I knew that but at some point I "lost>" it :confused:

And in case you wonder: Yes, it might make sense to re-implement dio_query_sectcount() based on getdevicedir() and GET_FILE_INFO. I leave it up to you if you want to try that or leave dio_query_sectcount() alone.

> Using GET_FILE_INFO will allow implementing both statvfs and stat (with an extra GET_EOF call to get a file's precise size in bytes) easy, without even bothering with the device number.

Yes, I'd appreciate a stat() implementation based on GET_FILE_INFO and GET_EOF. However, please be aware that GET_EOF requires to open/close the file. Nevertheless it seems that the byte size is worth it.

colinleroy:
> By the way, this thing about GET_FILE_INFO makes the ON_LINE (unitnum=0) call you mention in your first comment unnecessary, doesn't it? Using GET_FILE_INFO will allow implementing both statvfs and stat (with an extra GET_EOF call to get a file's precise size in bytes) easy.

Thanks for confirmations!
concerning stat(), do you think I should stick to the posix definition of struct stat or take liberties with it, as is done with dirent? I fear doing mktime for the crezte/modify dates is going to bloat code size quite a bit.

colinleroy:
> > By the way, this thing about GET_FILE_INFO makes the ON_LINE (unitnum=0) call you mention in your first comment unnecessary, doesn't it?
> 
> You're of course right. I got things mixed up. My proposal to have statvfs() either take a slash-prefixed volume name (or an absolute path containing such a slash-prefixed volume name) fits perfectly the GET_FILE_INFO parameter. At some point I knew that but at some point I "lost>" it 😕
> 
> And in case you wonder: Yes, it might make sense to re-implement dio_query_sectcount() based on getdevicedir() and GET_FILE_INFO. I leave it up to you if you want to try that or leave dio_query_sectcount() alone.
> 
> > Using GET_FILE_INFO will allow implementing both statvfs and stat (with an extra GET_EOF call to get a file's precise size in bytes) easy, without even bothering with the device number.
> 
> Yes, I'd appreciate a stat() implementation based on GET_FILE_INFO and GET_EOF. However, please be aware that GET_EOF requires to open/close the file. Nevertheless it seems that the byte size is worth it.

Thanks for the confirmations!
concerning stat(), do you think I should stick to the posix definition of struct stat or take liberties with it, as is done with dirent? I fear doing mktime for the crezte/modify dates is going to bloat code size quite a bit.

oliverschmidt:
I was in fact already about to write something about `struct dirent` :wink:

Yes, I propose to have custom fields:
* `st_type` (identical to `d_type`)
* `st_auxtype` (identical to `d_auxtype`)
* `st_access` (identical to `d_access`)

However, the situation regarding the time format isn't comparable to the `struct dirent`. https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/dirent.h.html describes only `d_ino` and `d_name`. Everything else is proprietary. In contrast https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_stat.h.html clearly defines the timestamps to be of type `struct timespec`. It seems to me we don't have much room for maneuver here...

oliverschmidt:
> Another question, would it be good to use iobuf_alloc/free instead of malloc/free for those and dio_querysectcount, opendir and friends? They'd return too big buffers, but users of iobuf0800 wouldn't have to worry about extra allocation?
or would you rather not?

Thanks for the proposal :smile:

I introduced the iobuf mechanism primarily because the alignment requirements and the size of the posix_memalign() code itself (written in C) added a LOT of overhead. I've thought about the pros and cons of that approach for some time and it seems to me we should rather stick with malloc/free.

colinleroy:
Hi @oliverschmidt, Here you are :)

The stat() function weighs quite a bit, and pulls in a lot of things (mli, open, close, mktime, etc etc). In an otherwise empty program, linking stat() in costs 4.1kB. (429 bytes if open/malloc/free/mktime are already linked in). The time conversion are heavy...

statvfs is much lighter, 523 bytes.

libsrc/apple2/filename.s
```diff
@@ -94,4 +95,4 @@ setlen: tya
 popname:
         ; Cleanup stack
         ldy     #1 + 64+1       ; Length byte + max pathname length + zero
```
oliverschmidt, line 97: Please use FILENAME_MAX here too.

asminc/apple2.inc
```diff
@@ -85,3 +85,6 @@ IOUDISOFF   :=  $C07F   ; Enable IOU
 
 ; Control Your Apple
 CYAREG  :=      $C036   ; Bits 0-3=disk detect 4=shadow all banks 7=fast
+
+; Max filename length
+FILENAME_MAX := 64
```
oliverschmidt, line 90: * I think FILENAME_MAX doesn't belong here. This file is about Apple II hardware attributes, not about Apple II software attributes. It fits much better in libsrc/apple2/mli.inc.
* The definition should use `=`, not `:=` as it doesn't denote an address.
* Given that the C definition is `64+1` I'm wondering if it might be preferrable to use `64+1` here too.

doc/funcref.sgml
```diff
@@ -7075,6 +7089,74 @@ be used in presence of a prototype.
...
+<item>The function requires an absolute pathname.
+</itemize>
+<tag/Availability/POSIX 1003.1
+<tag/See also/
+<ref id="stat" name="stat">
+<tag/Example/
+<verb>
+#include &lt;sys/statvfs.h&gt;
+
+#define FILENAME "/DISK/helloworld"
```
oliverschmidt, line 7148: I suggest to use lowercase for `DISK` too.

include/sys/stat.h
```diff
@@ -47,6 +53,39 @@
 #define S_IREAD  0x01
 #define S_IWRITE 0x02
 
+#define S_IFMT   0x03
+
+
+
```
oliverschmidt, line 59: Please remove two empty lines.

include/sys/stat.h
```diff
@@ -47,6 +53,39 @@
...
+#define S_IFDIR  0x01
+#define S_IFREG  0x02
+#define S_IFBLK  0xFF
+#define S_IFCHR  0xFF
+#define S_IFIFO  0xFF
+#define S_IFLNK  0xFF
+#define S_IFSOCK 0xFF
+#endif
+
+
```
oliverschmidt, line 70: Please remove one empty line.

libsrc/apple2/mli.inc
```diff
@@ -83,8 +83,8 @@ EOF_COUNT       = 2
                                 AUX_TYPE        .word
                                 STORAGE_TYPE    .byte
                                 BLOCKS          .word
-                                MODE_DATE       .word
-                                MODE_TIME       .word
+                                MOD_DATE        .word
+                                MOD_TIME        .word
```
oliverschmidt, line 87: Thanks 👍🏻

libsrc/apple2/stat.s
```diff
@@ -0,0 +1,163 @@
...
+is_reg: lda     #S_IFREG
+
+set_st_mode:
+        sta     (ptr4),y
+
+        ; st_auxtype
+        lda     mliparam + MLI::INFO::AUX_TYPE
+        ldy     #stat::st_auxtype
+        sta     (ptr4),y
+        lda     mliparam + MLI::INFO::AUX_TYPE + 1
```
oliverschmidt, line 87: It's amazing how well you adapted to the existing coding style !!!

Just one minor thing here: When two "actual" values are added, then the `+` is separated with blanks. But if it's "just the usual + 1" (especially for 16-bit high-byte related stuff), then the blanks are omitted.

libsrc/apple2/stat.s
```diff
@@ -0,0 +1,163 @@
...
+
+        ; Store size
+        lda     stbuf
+        ldx     stbuf+1
+        sta     ptr4
+        stx     ptr4+1
+        ldy     #stat::st_size
+        lda     mliparam + MLI::EOF::EOF
+        sta     (ptr4),y
+        lda     mliparam + MLI::EOF::EOF + 1
```
oliverschmidt, line 138: Blanks (see above).

libsrc/apple2/stat.s
```diff
@@ -0,0 +1,163 @@
...
+        iny
+        sta     (ptr4),y
+        lda     mliparam + MLI::EOF::EOF + 2
```
oliverschmidt, line 141: Blanks (see above).

libsrc/apple2/statvfs.s
```diff
@@ -0,0 +1,128 @@
...
+        lsr
+        lsr
+        ldy     #statvfs::f_fsid
+        sta     (ptr4),y
+
+        ; total number of blocks
+        lda     mliparam + MLI::INFO::AUX_TYPE
+        ldy     #statvfs::f_blocks
+        sta     (ptr4),y
+        lda     mliparam + MLI::INFO::AUX_TYPE + 1
```
oliverschmidt, line 79: Blanks (see above).

libsrc/apple2/statvfs.s
```diff
@@ -0,0 +1,128 @@
...
+        ; blocks free & avail
+        sec
+        lda     mliparam + MLI::INFO::AUX_TYPE
+        sbc     mliparam + MLI::INFO::BLOCKS
+        ldy     #statvfs::f_bfree
+        sta     (ptr4),y
+        ldy     #statvfs::f_bavail
+        sta     (ptr4),y
+
+        lda     mliparam + MLI::INFO::AUX_TYPE + 1
```
oliverschmidt, line 92: Blanks (see above).

libsrc/apple2/statvfs.s
```diff
@@ -0,0 +1,128 @@
...
+        sbc     mliparam + MLI::INFO::BLOCKS + 1
```
oliverschmidt, line 93: Blanks (see above).

libsrc/apple2/statvfs.s
```diff
@@ -0,0 +1,128 @@
...
+        iny
+        sta     (ptr4),y
+        ldy     #statvfs::f_bfree + 1
```
oliverschmidt, line 96: Blanks (see above).

libsrc/apple2/statvfs.s
```diff
@@ -0,0 +1,128 @@
...
+        ldy     #statvfs::f_bsize
+        sta     (ptr4),y
+        ldy     #statvfs::f_frsize
+        sta     (ptr4),y
+        ; f_frsize high byte
+        iny
+        txa
+        sta     (ptr4),y
+        ; f_bsize high byte
+        ldy     #statvfs::f_bsize + 1
```
oliverschmidt, line 111: Blanks (see above).

libsrc/apple2/statvfs.s
```diff
@@ -0,0 +1,128 @@
...
+        sta     (ptr4),y
+
+        ; f_namemax
+        lda     #FILENAME_MAX
+        ldy     #statvfs::f_namemax
```
oliverschmidt, line 116: I believe that this value needs to end up in one way or another being `65`, not `64`.

oliverschmidt (changes requested):

oliverschmidt:
> In an otherwise empty program, linking stat() in costs 4.1kB.

That's indeed a lot :-(

I'm wondering if we could/should replace
```
struct stat {
    dev_t             st_dev;
    ino_t             st_ino;
    mode_t            st_mode;
    nlink_t           st_nlink;
    uid_t             st_uid;
    gid_t             st_gid;
    off_t             st_size;
    struct timespec   st_atim;
    struct timespec   st_ctim;
    struct timespec   st_mtim;
    #ifdef __APPLE2__
    unsigned char     st_access;
    unsigned char     st_type;
    unsigned int      st_auxtype;
    #endif
};
```
with
```
struct stat {
    dev_t             st_dev;
    ino_t             st_ino;
    mode_t            st_mode;
    nlink_t           st_nlink;
    uid_t             st_uid;
    gid_t             st_gid;
    off_t             st_size;
    struct timespec   st_atim;
    struct timespec   st_ctim;
    struct timespec   st_mtim;
    #ifdef __APPLE2__
    struct {
        unsigned day  :5;
        unsigned mon  :4;
        unsigned year :7;
    }                 st_cdate;
    struct {
        unsigned char min;
        unsigned char hour;
    }                 st_ctime;
    struct {
        unsigned day  :5;
        unsigned mon  :4;
        unsigned year :7;
    }                 st_mdate;
    struct {
        unsigned char min;
        unsigned char hour;
    }                 st_mtime;
    unsigned char     st_access;
    unsigned char     st_type;
    unsigned int      st_auxtype;
    #endif
};
```

The idea is to keep
```
    struct timespec   st_atim;
    struct timespec   st_ctim;
    struct timespec   st_mtim;
```
in place to stay compliant but just not fill them. After all, `st_atim` can't be filled anyway. Instead fill proprietary fields.

@mrdudz: The point here is that mktime() is pretty large. And it's almost certain that the user doesn't benefit from a timespec    at all. He won't try to subtract file timestamps or alike. Rather he only wants to display them, which adds additional cost for localtime(). So do you consider the approach described above as acceptable/preferable?

Note1: It's likely desirable to have a struct declared for the ProDOS time/date representation and use it both above and in dirent.h instead of repeating identical definitions.

Note2: It might be desirable to add some proprietary function that converts such a ProDOS time/date representation (as it is exposed both be readdir() and stat()) into a `struct tm`.

colinleroy:
@oliverschmidt, thanks for your review! I will address those.
Concerning stat()'s size, mktime is part of it, but not only. There's also the requirement to open() to get the file size, which pulls in malloc,free,posix_memalign etc.
But I think using prodos "extensions" is a good idea too, and that will indeed shrink stat() a bit.

oliverschmidt:
> There's also the requirement to open() to get the file size, which pulls in malloc,free,posix_memalign etc.

I see your point. However, it's very likely that a program using stat() will also want to open() at least one file. So  all that code is pulled in anyhow. But that's not true at all for mktime(), especially as there are likely (many) use cases where stat() is only called to get the file size and the timestamps are completely ingored.

colinleroy:
> > There's also the requirement to open() to get the file size, which pulls in malloc,free,posix_memalign etc.
> 
> I see your point. However, it's very likely that a program using stat() will also want to open() at least one file. So all that code is pulled in anyhow. But that's not true at all for mktime(), especially as there are likely (many) use cases where stat() is only called to get the file size and the timestamps are completely ingored.

I wasn't disagreeing, and am 100% aligned with your view - I was just managing your expectations :wink: 
Latest push fixes the coding style issues, and implements what you suggested. stat() now weighs 2392 bytes in an otherwise empty program, 314 bytes in a program already using open().

colinleroy:
Added some cleanup for pushname/popname, the "ownership" of it was unclear, now it is.

Concerning dio_querysectcount(): it could be rewritten with something along the lines of
```
... prepare a buffer for the C-style filename
jsr get_devicedir
jsr pushax
jsr mli_info_call
lda mliparam + MLI::INFO::AUX_TYPE
ldx mliparam + MLI::INFO::AUX_TYPE+1
... cleanup C-style filename buffer
```
but I don't think it's worth it: the size gain on dio_querysectcount() would be completely cancelled by the pulling in of get_devicedir plus the buffer handling.

colinleroy:
If/when you think it's good to merge, you can ping if you want me to squash commits to keep a clean history on the main branch :)

libsrc/apple2/mli_file_info.s
```diff
@@ -0,0 +1,54 @@
...
+        jsr     popname         ; Preserves A
+
+        plp
+        bcs     oserr
+        rts
+
+oserr:
+        jsr     ___mappederrno
+        sec
+        rts
```
oliverschmidt, line 54: I don't see a technical reason why these two functions need to be in a single file. Therefore I ask you to separate them in order to have i.e. exec() callers not link the unused code of mli_file_info.

asminc/time.inc
```diff
@@ -58,6 +58,15 @@
...
 
+;------------------------------------------------------------------------------
+; ProDOS date/time format
+.ifdef __APPLE2__
+.struct prodos_datetime
+        date        .word
+        time        .word
+.endstruct
+.endif
+
```
oliverschmidt, line 69: I don't like this section here. Maybe mli.inc or <do you have a better idea?>

libsrc/apple2/mli_file_info.s
colinleroy: Good idea!

libsrc/apple2/stat.s
```diff
@@ -0,0 +1,172 @@
...
+        sta     (ptr4),y
+        lda     mliparam + MLI::INFO::CREATE_DATE+1
+        iny
+        sta     (ptr4),y
+        lda     mliparam + MLI::INFO::CREATE_TIME
+        iny
+        sta     (ptr4),y
+        lda     mliparam + MLI::INFO::CREATE_TIME+1
+        iny
+        sta     (ptr4),y
```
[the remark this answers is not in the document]
colinleroy: That was the plan, but I forgot about it midway !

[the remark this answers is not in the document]
colinleroy: I've put st_storagetype and st_blocks instead of st_reserved, they could be useful to someone and there's no reason to hide them IMHO

include/time.h
```diff
@@ -65,6 +65,27 @@ struct tm {
...
+    }                 time;
+};
+
+int __fastcall__ prodos_datetime_to_posix_tm(struct tm *tm, struct prodos_datetime *prodos_dt);
+/* Converts a ProDOS date/time structure to a struct tm */
+
+time_t __fastcall__ prodos_datetime_to_posix_time_t(void *prodos_date_ptr);
+/* Converts a ProDOS date/time structure to a time_t UNIX timestamp */
+
+#endif
```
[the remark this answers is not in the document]
colinleroy: I wanted to originally, but then we'd have to include <apple2.h> in stat.h, dirent.h etc, and <time.h> in apple2.h, that seemed more gross in the end?

asminc/time.inc
colinleroy: Same as for the .h files: putting it elsewhere could be possible, but not in apple2.inc either, or we'd have to include it there, too. Maybe a specific file asminc/prodos_datetime.inc ?

include/time.h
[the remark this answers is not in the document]
oliverschmidt: 1. I don't think so. Headers including other headers seems more normal to me than putting "all into one" header :smirk: I propose to `#include <target.h>` where necessary.
2. https://github.com/cc65/cc65/blob/master/include/apple2.h#L153-L167 was **NO** struct type declaration. It was a variable declaration. You can't just remove it.

asminc/time.inc
oliverschmidt: I don't agree that it is the same as for the .h files: The .h files include other .h files "where necessary", but .inc files never include other .inc files - at least afaik. It's up to the user to know what to include in which order.

Given that there's no code using this .struct at all I propose to "simply" go without. In stat.inc we just write:
```
.struct stat
        ...
        st_blocks         .word
        st_mod_date       .word
        st_mod_time       .word
        st_create_date    .word
        st_create_time    .word
```

include/apple2.h
```diff
@@ -150,22 +150,6 @@ extern unsigned char _dos_type;
...
-        unsigned day  :5;
-        unsigned mon  :4;
-        unsigned year :7;
-    }             createdate;    /* Current date: 0 */
-    struct {
-        unsigned char min;
-        unsigned char hour;
-    }             createtime;    /* Current time: 0 */
-} _datetime;
-
```
oliverschmidt, line 168: As discussed, restore the declaration of the variable `_datetime`.

libsrc/apple2/open.s
```diff
@@ -247,5 +248,4 @@ __auxtype:
         .word   $0000           ; AUX_TYPE:     Load address N/A
         .byte   $01             ; STORAGE_TYPE: Standard seedling file
 __datetime:
-        .word   $0000           ; CREATE_DATE:  Current date
-        .word   $0000           ; CREATE_TIME:  Current time
+        .tag prodos_datetime
```
oliverschmidt, line 251: As discussed, please restore this to the original state.

libsrc/apple2/prodos_datetime_to_posix_tm.s
```diff
@@ -0,0 +1,99 @@
...
+
+        ; Load errno code
+erange: jsr     zero_tm_buf
+        jsr     fill_result
+        lda     #ERANGE
+        ; Set __errno
+        jmp     ___directerrno
+
+prodos_tm_buf:
+        .tag    tm
```
oliverschmidt, line 99: You moved this from .BSS to .CODE - this seems undesirable to me.

libsrc/apple2/prodos_datetime_to_posix_tm.s
```diff
@@ -0,0 +1,99 @@
+;
+; Oliver Schmidt, 14.08.2018
+; Colin Leroy-Mira, 2023 <colin@colino.net>
+;
+; int __fastcall__ prodos_datetime_to_posix_tm(struct tm *tm, struct prodos_datetime *prodos_dt)
```
oliverschmidt, line 5: I think it's preferable to have this function return a value to the "internal" struct tm, just like localtime() and gmtime() do.

oliverschmidt:
Sorry for coming up with this so late :flushed: but...

I don't think that the `prodos_` prefix is desirable. We only support ProDOS on the Apple II and that won't chnage. And after all we call it `open()` and not `prodos_open()` :smirk:

So `prodos_datetime` should become `datetime`.

In addition, we rather prefer short, cryptic names over long, descriptive names. Therefore I propose:

`time_t __fastcall__ mktime_dt(const struct datetime* dt);`
`struct tm* __fastcall__ gmtime_dt(const struct datetime* dt);`

libsrc/apple2/prodos_datetime_to_posix_tm.s
colinleroy: nice catch!

include/time.h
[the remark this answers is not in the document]
colinleroy: 1. Addressed :)
2. 3.: Indeed! I read that too fast.

include/sys/stat.h
```diff
@@ -47,6 +54,40 @@
...
+#ifdef __APPLE2__
+#define S_IFDIR  0x01
+#define S_IFREG  0x02
+#define S_IFBLK  0xFF
+#define S_IFCHR  0xFF
+#define S_IFIFO  0xFF
+#define S_IFLNK  0xFF
+#define S_IFSOCK 0xFF
+#endif
```
oliverschmidt, line 67: Please move this to apple2.h too.

include/time.h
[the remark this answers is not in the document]
colinleroy: About the `extern struct datetime _datetime`: I have put it back. It references __datetime from open.s and could indeed be used by cc65 users; however, it's not used internally and not mentioned in the docs. 
The fact I've put it back using the struct datetime definition instead of the "one shot" definition means its fields changed, from `_datetime.createdate/_datetime.createtime` to `_datetime.date/_datetime.time`.
Is this acceptable?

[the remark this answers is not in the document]
oliverschmidt: Thanks! That's exactly how I envisioned it. I believe this incompatible change is to be accepted. I don't see it being used a lot because of the missing documentation. Maybe you would want to add some documentation in another change(?).

oliverschmidt:
And finally for the **REAL** nitpicking :stuck_out_tongue_winking_eye:

* The cc65 coding style is to put the dereference `*` at the type, not at the variable. See my proposed prototypes above.
* You sometimes write 'POSIX', 'UNIX' and 'POSIX-style'. I haven't checked if there a clear distinction. If there is, then I'm of course fine. Otherweise it might be preferable to consistently use one term.

colinleroy:
> And finally for the **REAL** nitpicking 😜

:sweat_smile: 

>     * The cc65 coding style is to put the dereference `*` at the type, not at the variable. See my proposed prototypes above.

Fixed! Sorry.
 
>     * You sometimes write 'POSIX', 'UNIX' and 'POSIX-style'. I haven't checked if there a clear distinction. If there is, then I'm of course fine. Otherweise it might be preferable to consistently use one term.

I'm unsure too. I sneakily reworded it to "more portable" to avoid hurting my brain with reference documentations, apart for the time_t which everybody knows is a "UNIX timestamp" :-)

include/apple2.h
```diff
@@ -142,29 +143,42 @@ extern unsigned char _dos_type;
...
-/*                                 Variables                                 */
-/*****************************************************************************/
+/* struct stat.st_mode values */
+#define S_IFDIR  0x01
+#define S_IFREG  0x02
+#define S_IFBLK  0xFF
+#define S_IFCHR  0xFF
+#define S_IFIFO  0xFF
+#define S_IFLNK  0xFF
+#define S_IFSOCK 0xFF
```
oliverschmidt, line 156: Inside the "sections", we only have only linefeed.

oliverschmidt:
> Fixed! Sorry.

Certainly no reason to say "sorry"! Please adjust the documentation too.

And I noticed that (parts of) the example code doesn't adhere to the "strange" cc65 C code style with the black in front of function parameters.

colinleroy:
> > Fixed! Sorry.
> 
> Certainly no reason to say "sorry"! Please adjust the documentation too.
> 
> And I noticed that (parts of) the example code doesn't adhere to the "strange" cc65 C code style with the blank in front of function parameters.

Fixed (and in apple2.h too)

include/apple2.h
```diff
@@ -151,20 +173,10 @@ extern unsigned char _dos_type;
 
 
 /* The file stream implementation and the POSIX I/O functions will use the
-** following struct to set the date and time stamp on files. This specificially
-** applies to the open and fopen functions.
+** following struct to set/get the date and time stamp on files. This specifically
+** applies to the open, fopen, stat and readdir functions.
```
oliverschmidt, line 177: Is this a correct comment here? I think _datetime only applies to open and fopen.

colinleroy: Uh indeed, that's a leftover from when I had put it on the struct.

oliverschmidt:
Believe it or not :grin: I've nothing to complain about anymore :thumbsup:

Do you want to squash something or do you want me to merge as-is?

BTW: I've not checked anything regarding the actual functional correctness. But I have reason to believe that you'll take care of issues if they should show up :smile:

colinleroy:
> Believe it or not 😁 I've nothing to complain about anymore 👍

Yay!

> Do you want to squash something or do you want me to merge as-is?

Squashing's done !

> BTW: I've not checked anything regarding the actual functional correctness. But I have reason to believe that you'll take care of issues if they should show up 😄

I have tested everything rather extensively, on volumes, directories, files, non-existent entities etc. And I will indeed be there in case of issues !

colinleroy:
Oh wait just one info I want to add to the doc.

oliverschmidt:
That was a "tiny bit" more work than you expected originally, I guess :stuck_out_tongue_winking_eye:

The branch name `apple2-dio-freesectcount` is a silent witness...

colinleroy:
> That was a "tiny bit" more work than you expected originally, I guess 😜
> 
> The branch name `apple2-dio-freesectcount` is a silent witness...

Haha, yes, but it's better work :) 
Now I'm gonna remove my #ifdefs __CC65__ that were using stat() on linux and an opendir/readdir-based loop on 6502, cleanup my code, and regain a few hundred bytes, so I'm happy!

oliverschmidt:
That reminds me on my work on Contiki. There too, it was a ping-pong of adding capabilities I needed to cc65 and then adding their usage to Contiki. I.e. exec() was born by the need to run wget from the the web browser. I remember that approach being very rewarding - so I'm with you :grinning:

colinleroy:
> That reminds me on my work on Contiki. There too, it was a ping-pong of adding capabilities I needed to cc65 and then adding their usage to Contiki. I.e. exec() was born by the need to run wget from the the web browser. I remember that approach being very rewarding - so I'm with you 😀

This commit was very satisfying - and makes for a 221 bytes gain on my binary!
https://github.com/colinleroy/a2tools/commit/a6a8cfcc0b8e261b2424cc6c5589c9acd669d7ef

PR #2316, opened 2023-12-21
Add free space info in relevant segments

colinleroy:
I'm tired of guessing how many bytes I got left in BSS, so I've made this patch. It outputs free space in relevant segments. I consider them relevant if 
a) the next segment is in the same memory area, but doesn't start right after this one. That's the case for example of the LOWCODE segment in apple2, when the HGR page from 0x2000-0x4000 is used.
b) or it is the BSS segment.

The output in the map file with that patch is, for example:
```
Segment list:
-------------
Name                   Start     End    Size  Align    Free  Memory area
------------------------------------------------------------------------
NULL                  000000  000000  000000  00001
ZEROPAGE              000080  000099  00001A  00001          ZP
EXEHDR                000BC6  000BFF  00003A  00001          HEADER
STARTUP               000C00  000C33  000034  00001          MAIN
LOWCODE               000C34  001FC2  00138F  00001  00003D  MAIN
HGR                   002000  003FFF  002000  00001          MAIN
CODE                  004000  008D6B  004D6C  00001          MAIN
RODATA                008D6C  009CBB  000F50  00001          MAIN
DATA                  009CBC  009EBC  000201  00001          MAIN
INIT                  009EBD  009F99  0000DD  00001          MAIN
RT_ONCE               009F9A  00A56E  0005D5  00001          MAIN
BSS                   00A56F  00B724  0011B6  00001  0005DA  BSS
ONCE                  00A56F  00A6FA  00018C  00001          MAIN
LC                    00D400  00DF7C  000B7D  00001          LC

```
I hope you'll consider it for inclusion. If the conditions are not perfect, I can fix them :)

### DIFF
--- a/src/ld65/segments.c
+++ b/src/ld65/segments.c
@@ -590,8 +590,8 @@ void PrintSegmentMap (FILE* F)
     qsort (SegPool, CollCount (&SegmentList), sizeof (Segment*), CmpSegStart);
 
     /* Print a header */
-    fprintf (F, "Name                   Start     End    Size  Align\n"
-                "----------------------------------------------------\n");
+    fprintf (F, "Name                   Start     End    Size  Align    Free  Memory area\n"
+                "------------------------------------------------------------------------\n");
 
     /* Print the segments */
     for (I = 0; I < CollCount (&SegmentList); ++I) {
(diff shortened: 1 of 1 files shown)


mrdudz:
Interesting addition - of course it breaks all the script magic i am using :)

colinleroy:
Oh, I should have thought of that. Maybe it would break less scripts if I added the new info as the last columns?

mrdudz:
Oh generally don't be afraid of breaking changes - it makes sense this way (and relying on the format of the file is totally my fault).

That said, it would be nice if that extra column would always exist (ie for all segments) - that makes parsing it in a script much less error prone

colinleroy:
Updated the commit to print 000000 instead of leaving an empty column.

mrdudz:
Just checked the patch in detail - i don't understand why there have to be hardcoded section names, that seems a bit backwards... I see the problem with different sections going into the same memory area, so you often cant really say how much is free in them - so shouldnt the table actually contain the free space per memory area instead?

colinleroy:
Showing free space per memory area is kind of misleading:
```
Name                   Start     End    Size  Align    Free  Memory area
------------------------------------------------------------------------
NULL                  000000  000000  000000  00001  000000
ZEROPAGE              000080  000099  00001A  00001  000000  ZP
EXEHDR                000BC6  000BFF  00003A  00001  000000  HEADER
STARTUP               000C00  000C36  000037  00001  000000  MAIN
LOWCODE               000C37  001FE4  0013AE  00001  00001B  MAIN
HGR                   002000  003FFF  002000  00001  000000  MAIN
CODE                  004000  008946  004947  00001  000000  MAIN
RODATA                008947  009961  00101B  00001  000000  MAIN
DATA                  009962  009B61  000200  00001  000000  MAIN
INIT                  009B62  009C3F  0000DE  00001  000000  MAIN
RT_ONCE               009C40  00A457  000818  00001  000000  MAIN
BSS                   00A458  00B626  0011CF  00001  0006D8  BSS
ONCE                  00A458  00A658  000201  00001  000000  MAIN
LC                    00D400  00DFE5  000BE6  00001  00001A  LC

Memory Area           Size    Fill    Free
--------------------------------------------
ZP                    00001A  00001A  000000
HEADER                00003A  00003A  000000
MAIN                  00B2FF  00A63F  000CC0
BSS                   0018A7  0011CF  0006D8
LC                    000C00  000BE6  00001A
```
Notice MAIN has loads of free space (because it could indeed be bigger, which would only decrease free space in BSS); and we cannot easily notice anymore that with the reserved 0x2000-0x4000 HGR page having LOWCODE before and CODE above causes a 0x1B gap between the end of LOWCODE and the start of HGR.

mrdudz:
> Showing free space per memory area is kind of misleading

Well - i find showing it per section even more misleading. And whats more of an issue: those hardcoded section names makes it rely on certain preconditions - which is really a bad idea. This kind of thing must work always, even when none of those sections exist at all (eg an assembler program with a custom linker config). That means hardcoded section names to determine whatever is a no go.

> and we cannot easily notice anymore that with the reserved 0x2000-0x4000 HGR page having LOWCODE before and CODE above causes a 0x1B gap between the end of LOWCODE and the start of HGR.

I'm not familiar with the apple2 config at all, so that doesn't really tell me anything :) However, when more than one section goes into a memory area, then its simply not possible to tell how much is free in any of those sections. And unfortunately - the same is true for memory areas that live in the same RAM space.

I am not sure how to proceed here - however, what we need is a generic solution that always works. The current solution sems to be too much of a tweak for a specific situation to me.

colinleroy:
Hi,
I understand. I was hoping it would be generic enough for inclusion, but I now get that it's quite tailored to my use (and maybe even to the Apple 2 platform), and I guess I'll retire that PR and replace it with my own very little shell script 🙂

mrdudz:
It's unfortunate really - the idea is good of course (and i also wish i wouldn't need those scripts myself) - but i don't see how a proper generic solution could work :/

PR #2318, opened 2023-12-29
Added SPRCTL0 and SPRCTL1  bit definitions for ca65

alexthissen:
This adds bit definitions for the SPRCTL0 and SPRCTL1 sprite control block addresses. It allows the use of the bit definitions in .asm files for ca65. 
The definitions are mostly consistent with the /includes/_suzy.h defines for C files in cc65.


mrdudz:
Thanks!

PR #2320, opened 2023-12-30
bug 2319 - dont allow -t after input files given

### LINKED ISSUE #2319: cl65 confusing behavior due to c64 defaulting
This took me a long time to unearth.

compiling like this

`cl64 test.c -t sim6502`

produces a file that runs under sim6502 but has been compiled and assembled for c64. This in particular means that printfs dont work because
- the format string gets upper cased (%d -> %D) and so are ignored
- the \n gets converted to a \r. On sim65 this results in no output being produced at all

this is due to cl65 doing this

```
PS C:\work\db65> ..\forks\cc65\bin\cl65 -d argtest.c -t sim6502
Executing: C:\work\forks\cc65\bin\cc65 -d -t c64 argtest.c
Executing: C:\work\forks\cc65\bin\ca65 -t c64 argtest.s
Executing: C:\work\forks\cc65\bin\ld65 -o argtest -t sim6502 argtest.o sim6502.lib
```

compile and assemble is done for c64, link is done for sim65

I think this should be treated as an error. If the -t flag is present after the source file names it should be rejected. I have spent a huge amount of time trying to work out why things just stopped working, seemingly at random (this was when I was working on a rewrite of __printf.s so I assumed it was my printf code)

pm100:
fixes #2319 cl65 now fatals if a command like this is given

```
cl65 mycode.c -t sim6502
cl65: Target must be specified before input files
```

### DIFF
--- a/doc/cl65.sgml
+++ b/doc/cl65.sgml
@@ -261,6 +261,9 @@ different options for different files on the command line. As an example.
 translates main.c with full optimization and module.c with less optimization
 and debug info enabled.
 
+Note that the target system (-t , --target) must be specified before any file
+unless using the default target of c64
+
 The type of an input file is derived from its extension:
 
 <itemize>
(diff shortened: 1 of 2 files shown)


polluks:
What about `cl65 mycode.c -t c64`? Just nitpicking.

mrdudz:
Would you also add a line or two in the docs?

mrdudz:
Thanks!

PR #2328, opened 2024-01-01
gmtime() / mktime(): work on unsigned 

colinleroy:
Edit: wrong analysis of the problem, see https://github.com/cc65/cc65/pull/2328#issuecomment-1880017105

This is only useful in case of weird (aka wrong) values in the input struct tm, but may be considered useful enough to spend 8 more bytes on strftime: on Apple II, trying to
```
time_t t = (time_t)-1;
printf("%s", ctime(&t));
```
outputs garbage (as expected), but also somehow clears the screen.

image

This patch makes it print garbage, but in a safe way:
image

### DIFF
--- a/libsrc/common/gmtime.c
+++ b/libsrc/common/gmtime.c
@@ -48,11 +48,7 @@ struct tm* __fastcall__ gmtime (const time_t* timep)
     static struct tm timebuf;
     time_t t;
 
-    /* Check the argument */
-    if (timep == 0 || (long) (t = *timep) < 0) {
-        /* Invalid arg */
-        return 0;
-    }
+    t = *timep;
 
     /* Since our ints are just 16 bits, split the given time into seconds,
     ** hours and days. Each of the values will fit in a 16 bit variable.
--- a/libsrc/common/mktime.c
+++ b/libsrc/common/mktime.c
@@ -70,7 +70,12 @@ static const unsigned MonthDays [] = {
 static unsigned char __fastcall__ IsLeapYear (unsigned Year)
 /* Returns 1 if the given year is a leap year */
 {
-    return (((Year % 4) == 0) && ((Year % 100) != 0 || (Year % 400) == 0));
+    /* Originally the famous
+    ** return (((Year % 4) == 0) && ((Year % 100) != 0 || (Year % 400) == 0))
+    ** with a real year, but simplified for size in the range that we handle,
+    ** where only 200 (2100) is a special case.
+    */
+    return (Year != 200 && Year % 4 == 0);
 }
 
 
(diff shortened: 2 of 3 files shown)


mrdudz:
I'm not sure if this kind of "fix" is what we want... i can't seem to find a proper description on what exactly this function is expected to do in this case either though

polluks:
Indeed, seems to be UB.

Am Sa., 6. Jan. 2024 um 17:46 Uhr schrieb Bob Andrews <
***@***.***>:

colinleroy:
Hi,
I have given this a second look and do feel stupid: strftime()'s failure was only due to gmtime() checking whether the passed time_t was 0 or negative.

Changing gmtime() to handle unsigned time_t (and mktime to work on unsigned, too) fixes it (_and moves the 2038 bug to 2106, we'll be long dead, not our problem, yay!_)

colinleroy:
Also while I'm at it: I have a performance improvement (by a factor > 12) to mktime(), which costs 31 bytes though.
Are you folks interested?

mrdudz:
Is this really a performance criticial function? IMHO if you call it so often that performance matters, something is wrong with the code :)

colinleroy:
> Is this really a performance criticial function? IMHO if you call it so often that performance matters, something is wrong with the code :)

Yeah no, it's not 🙂 that's why I asked!
Want a bit of size optimisation instead?

mrdudz:
> Want a bit of size optimisation instead?
That again is almost always desired (exception being a few very critical functions, mostly from the runtime) - as size is usually much more of a problem than performance.

colinleroy:
> > Want a bit of size optimisation instead?
> That again is almost always desired (exception being a few very critical functions, mostly from the runtime) - as size is usually much more of a problem than performance.

Done!

And concerning the performance side. I do know that we're there for size, not speed. But. Still, I did some cycle counting with MAME and my profiler tool, and mktime() on the dataset of the unit test (https://github.com/cc65/cc65/pull/2328/files#diff-3597e873056dbe3352108abd228b70db64387c13e0dab36f4d368f045422386eR7)  takes 8.1 millions cycles, almost 1 sec per mktime() in average (the farther the date is, the longer it is to iterate over the months.) 
With the patch I left out of this PR, it's down to 436k cycles (less that 50ms per mktime()).
It costs 99 bytes, but this PR gains up 83. 
If you think the non-negligible performance boost warrants the extra 16 bytes, I'll submit a PR :)

The commit is there if you want to look at it: https://github.com/cc65/cc65/commit/ebf94e696d9eb3282d95f3d3133a9818945254ff

colinleroy:
Closing this PR, it's a mess. I'll do a consolidated one this evening which will
- be smaller than master
- be faster than master
- Fix the Y2038 bug and make it an Y2106 bug instead

PR #2334, opened 2024-01-04
[SIM65] Support undocumented opcodes for 6502

carlo-bramini:
This PR is mostly a complete rewrite of the emulator for 6502/65c02 opcodes. It provides an easier to maintain implementation of the instructions, by using few macros rather than having hand-written code for each function. All undocumented, previously missing opcodes for 6502 are also implemented. The patch also includes a detailed documentation of those opcodes, for reference to developers. This PR should fix one of the milestones listed here for the next version of CC65:

https://github.com/cc65/wiki/wiki/Before-the-next-release


mrdudz:
Cool, this kind of thing was on my todo list :)

groessler:
At first glance it seems that the undoc codes are always enabled. Could you please make them optional?

groessler:
I don't have a strong opinion about them being opt-in or opt-out.

mrdudz:
I think to be consistent with the rest of the toolchain, the illegal opcodes in sim65 should be added like this:

- when ca65 writes the file for sim65, it puts the CPU type into the header (currently 0 = 6502, 1 = 65C02). This should be extended to indicate 2 = 6502X, if that cpu was selected in the assembler
- sim65 reads the type of cpu from the binary. it should recognize 2 = 6502X, and only then use the opcode table that includes the illegal opcodes

(stretch goal: add a --cpu option to sim65 which overrides the type given in the binary)

carlo-bramini:
I think it could be quite easy to make illegal opcodes optional.
Actually, adding a `OP6502LegacyTable[]` near to `OP65C02Table[]` and `OP6502Table[]` with a new CPU type should be enough.
I can include it here or I can do a new PR with that change, if you want.
Of course, I just called it `CPU_6502Legacy`, but if you prefer different names, just let me know that.

rofl0r:
we have ricoh 2a03, which doesn't support BCD, but otherwise "supports" undocumented opcodes, the mos 6502 which "supports" them too, and then the 65c02 which turns them to NOPs. adding a "virtual" 6502 cpu which doesn't support them seems contrived, unless you want it to trap on illegal opcodes for debugging/code cleanness purposes.

in either case it seems 6502 setting should support the illegal opcodes, and a potential cpu addition that traps should use an invented name, imo.

mrdudz:
> in either case it seems 6502 setting should support the illegal opcodes, and a potential cpu addition that traps should use an invented name, imo.

That'd be different than in the rest of the toolchain though, where "6502" means "no illegals" and "6502X" means "6502+illegals" - which seems really odd and confusing to me. Or in other words: the names/"virtual cpu" have already been invented and are used all over the place.

rofl0r:
> That'd be different than in the rest of the toolchain though, where "6502" means "no illegals" and "6502X" means "6502+illegals" - which seems really odd and confusing to me. 

indeed confusing. maybe that could be rectified for the next release instead of continuing to use the confusing nomenclature.

mrdudz:
> indeed confusing. maybe that could be rectified for the next release instead of continuing to use the confusing nomenclature.

Hu? No. The naming per se isn't confusing at all. Not to me anyway  (and other assemblers do it in similar ways). Especially in the context of cc65, its also kinda handy "6502" does NOT include the illegals

mrdudz:
> I think it could be quite easy to make illegal opcodes optional. Actually, adding a `OP6502LegacyTable[]` near to `OP65C02Table[]` and `OP6502Table[]` with a new CPU type should be enough. I can include it here or I can do a new PR with that change, if you want. Of course, I just called it `CPU_6502Legacy`, but if you prefer different names, just let me know that.

Please do the following:
- add another table for 6502 without illegals
- use the new table when 6502X target is used (determined from the binary header)
- i haven't looked or tried, maybe this means ld65 needs to be updated to write the right identifier into the header. maybe it already does :)

carlo-bramini:
I added the code for handling illegal opcodes into another table, with the additional code for handling this new feature and this was the easy part.
About adding the new 6502X CPU type into sim65, I studied a bit the code and it seems to me that we have to look into this file:
https://github.com/cc65/cc65/blob/master/libsrc/sim6502/exehdr.s
This line:
https://github.com/cc65/cc65/blob/8aa59e4af309ff58fde6ab98e98490462c0e19a3/libsrc/sim6502/exehdr.s#L16
seems the point to touch for adding the feature.
It looks like that there is not something about 6502X into the target macros like `__SIM65C02__`, at least I have not found it, so by looking other asm source, perhaps another change could be done.
By using the macros provided by CPU.MAC:
https://github.com/cc65/cc65/blob/master/asminc/cpu.mac
perhaps we can replace that line with something like this:

```
.if (.cpu .bitand ::CPU_ISET_6502X)
    .byte   2
.elif (.cpu .bitand ::CPU_ISET_65C02)
    .byte   1
.elif (.cpu .bitand ::CPU_ISET_6502)
    .byte   0
.else
    .error Unknow CPU type.
.endif
```

However, I'm not expert of asm syntax of ca65 and CPU.MAC also lists several extra CPU capabilities.
For example, there are `CPU_ISET_65SC02` and `CPU_ISET_65C02` and I'm a bit unsure of their difference or if they need to be both tested into the above correction (how?)
So, I need a bit of support for writing this piece of code properly, I hope that someone could give some help.
Thank you very much for your time.

mrdudz:
That looks ok (even more elegant than i thought). I'd use the macros below, ie not the bitfield but the absolute CPU_ values - else the order of the checks matters and the result might be not obvious

mrdudz:
Just noticed, there is also a fatal bug in cpu.mac

```
CPU_6502DTV        = CPU_ISET_6502|CPU_ISET_6502X|CPU_ISET_6502DTV
```

This implies the DTV CPU has the exact same illegal opcodes as a 6502/6510 does - but this is not true. Some work the same, some others work differently (and iirc two of them don't work at all)

IMHO the entire bitfield stuff should go, and whatever checks the CPU should check explicitly for each CPU by equality.

So, please start with that in exeheader.s - so we can push this PR further :) I'll try to set up the vice testbench for sim6502 then, so we can confirm it all works OK

mrdudz:
Interesting that the malloc test fails now - wonder what triggers that :)

carlo-bramini:
> Interesting that the malloc test fails now - wonder what triggers that :)

I have to say that I tried to run `make test QUIET=1` also on my PC, after I have seen that one of the tests failed.
I have used CYGWIN 64bit, because I have create a (unreleased) package for this platform, but unfortunately I was not able to reproduce the same result. Then, I tried to run the same tests in a virtual machine with linux Debian and it still worked fine.
After looking the error here into the log:

```
val/lib_common_htons.g.6502.prg
val/lib_common_malloc.g.6502.prg
lib_common_malloc.c: Fail at line 27:
malloc returned something but shouldn't have
Expected status should be true but wasn't!
```

I tried to search the same point into output on the console of machine, but I was not able to find any reference to this test.
After `val/lib_common_htons.g.6502.prg`, on my console there is `val/lib_common_memmove.g.6502.prg`.
I have no idea from where this error comes from, at the moment.

carlo-bramini:
> > Interesting that the malloc test fails now - wonder what triggers that :)
> 
> I have to say that I tried to run `make test QUIET=1` also on my PC, after I have seen that one of the tests failed. I have used CYGWIN 64bit, because I have create a (unreleased) package for this platform, but unfortunately I was not able to reproduce the same result. Then, I tried to run the same tests in a virtual machine with linux Debian and it still worked fine. After looking the error here into the log:
> 
> ```
> val/lib_common_htons.g.6502.prg
> val/lib_common_malloc.g.6502.prg
> lib_common_malloc.c: Fail at line 27:
> malloc returned something but shouldn't have
> Expected status should be true but wasn't!
> ```
> 
> I tried to search the same point into output on the console of machine, but I was not able to find any reference to this test. After `val/lib_common_htons.g.6502.prg`, on my console there is `val/lib_common_memmove.g.6502.prg`. I have no idea from where this error comes from, at the moment.

EDIT: it seems that this test on malloc is a recent addition, I will try to check what's wrong with it.

mrdudz:
my time is running out again... i hope i can look at this soon. still want to check how we can run some of the CPU tests on it, we really cant afford breaking anything in the simulator (even if right now the tests pass, that doesn't mean its all ok, unfortunately)

carlo-bramini:
Hello,
I gave a look to https://sourceforge.net/p/vice-emu/code/HEAD/tree/testprogs/ while thinking a way for doing a good test of sim65.
At first sight, it doesn't seem to be easy to add support for these tests to sim65 as it is now.
Everything can be done but it requires some work, with the risk to introduce bugs into the testsuite
However, since the purpose would be "testing if sim65 is working fine", I was thinking something and I wanted to know your opinion. Instead of changing the assembly sources of the test suite and recompiling them, we could add an application that will test the opcodes of our virtual CPU.
In other words, we have these souces into sim65:

```
6502.c
error.c
main.c
memory.c
paravirt.c
```

They are compiled and linked with other common objects to make sim65.
Now, let's think to add an extra target into the `Makefile` and a `test.c` file that will be linked with `6502.o`:

```
6502.c -> 6502.o ---+
                    +---> test_sim65.exe
test.c -> test.o ---+
```

This `test.c` file will provide a very minimal support for running the testsuite: it will setup a custom "paravirt" for emulating things like key input from the keyboard and character output on the screen, as CBM computers did, and just ignore other stuff that it is not strictly required.
Then, it can loop for loading all `.prg` files from a given directory (that's an example) and run the virtual CPU on them.
At the end, on the console we can read all the results.
Loading these binary files is quite easy, all we need to to is to get the first 2 bytes for building the start address (typically `$801`).
Of course, we won't include support for other peripherals like CIA, VIA, VIC and so on, the simulator will just run the tests for checking the CPU.

Compiling this `test_sim65` executable would be easy for the `Makefile`, but in my opinion it would be better to leave to the user the choice if building it or not. By default, it is not created, but the user can generate it if he types `make test_sim65` for example.

What do you think?

mrdudz:
That only shifts the problem to: now you have to somehow prove the testsuite is OK :)

It shouldn't be too much work to port the CPU tests of the testbench (i have already ported them to most CBM machines anyway) - not the entire testbench of course, that is pretty pointless :) The big advantage with that is, that all the tests are proven to work and cross-checked on real CPUs, so there wont be any errors on that side.

carlo-bramini:
After some work, I created a test application for validating the opcodes of the simulator, as I have described in my previous post.
I used the precompiled Lorenz-2.15, provided by the vice-testprogs repository.
The SIM65 program with the new support for 6502X passed the tests successfully.
It executed all tests except the ones for the access to the I/O port of the 6510 and the hardware interrupts, because the testsuite uses the CIAs for generating them and they didn't exist in my quickly written test application.

Now the emulation is working fine, at least in my opinion, and the next task may be the addition of a testsuite, like the Lorenz-2.15, with native support for SIM65.
This will make available a tool to be used for discovering regressions after changes to the simulator.

mrdudz:
Just for curiosity - can you post those extra files somewhere?

Also, can you check the extra tests in the CPU directory too? some will not work (because they are very C64 specific and need cycle exact interrupts and/or VICII), but about half of them are generic enough that they should work - in particular the various decimal mode tests, jsrselfmod, and some of the more elaborate undocs tests are interesting here (as these things are underrepresented in the lorenz suite).

Also, did you test both regular (6502) and with undocs (6502X)?

carlo-bramini:
> Just for curiosity - can you post those extra files somewhere?

This is the source file that I have written for doing the test with the lorenz suite:

```
#include <stdio.h>
#include <stdlib.h>
#include <io.h>
#include <string.h>

#include "6502.h"
#include "memory.h"
#include "error.h"
#include "paravirt.h"

/* Search path, change it according your needs */
static const char search_path[] =
    "C:\\cygwin64\\home\\carlo\\vice-testprogrs\\general\\Lorenz-2.15\\src\\";

/* count of total cycles executed */
unsigned long long TotalCycles = 0;

/* countdown from MaxCycles */
unsigned long long RemainCycles;

extern CPURegs Regs;

int load_prg(const char *fname)
{
    FILE *fp = fopen(fname, "rb");
    int   adr_h, adr_l, adr, size;

    if (fp == NULL)
        return 0;

    adr_l = fgetc(fp) & 0xFF;
    adr_h = fgetc(fp) & 0xFF;

    adr = adr_l + adr_h * 256;

    fseek(fp, 0, SEEK_END);
    size = ftell(fp) - 2;
    fseek(fp, 2, SEEK_SET);

    fread(Mem+adr, 1, size, fp);
    fclose(fp);

    return 1;
}

void run_prg(int address)
{
    Reset ();

    Regs.PC = Mem[address + 5] - '0';
    Regs.PC = Regs.PC * 10 + Mem[address + 6] - '0';
    Regs.PC = Regs.PC * 10 + Mem[address + 7] - '0';
    Regs.PC = Regs.PC * 10 + Mem[address + 8] - '0';
    Regs.PC = Regs.PC * 10 + Mem[address + 9] - '0';

    for (;;) {
        int Cycles = ExecuteInsn ();
        TotalCycles += Cycles;
        if (Regs.PC == 0xA474) // returns to Basic start -> finish
            break;
    }
}

static unsigned char Pop (CPURegs* Regs)
{
    return MemReadByte (0x0100 + (++Regs->SP & 0xFF));
}

void build_tables (void)
{
    Mem[0xFFFE] = 0x48; // Vector address
    Mem[0xFFFF] = 0xFF;

    Mem[0xFF48] = 0x48; // PHA
    Mem[0xFF49] = 0x8A; // TXA
    Mem[0xFF4A] = 0x48; // PHA
    Mem[0xFF4B] = 0x98; // TYA
    Mem[0xFF4C] = 0x48; // PHA
    Mem[0xFF4D] = 0xBA; // TSX
    Mem[0xFF4E] = 0xBD;
    Mem[0xFF4F] = 0x04;
    Mem[0xFF50] = 0x01; // LDA $0104,X
    Mem[0xFF51] = 0x29;
    Mem[0xFF52] = 0x10; // AND #$10
    Mem[0xFF53] = 0xF0;
    Mem[0xFF54] = 0x03; // BEQ $FF58
    Mem[0xFF55] = 0x6C;
    Mem[0xFF56] = 0x16;
    Mem[0xFF57] = 0x03; // JMP ($0316) ; BRK vector
    Mem[0xFF58] = 0x6C;
    Mem[0xFF59] = 0x14;
    Mem[0xFF5A] = 0x03; // JMP ($0314) ; IRQ vector

    Mem[0x0314] = 0x81;
    Mem[0x0315] = 0xEA;

    Mem[0x0316] = 0x81;
    Mem[0x0317] = 0xEA;

    Mem[0xEA81] = 0x68; // PLA
    Mem[0xEA82] = 0xA8; // TAY
    Mem[0xEA83] = 0x68; // PLA
    Mem[0xEA84] = 0xAA; // TAX
    Mem[0xEA85] = 0x68; // PLA
    Mem[0xEA86] = 0x40; // RTI
}

void ParaVirtHooks(CPURegs* Regs)
{
    static char fname[64] = {0};
    char tmp[_MAX_PATH];
    int handled;
    FILE *fp;
    int   adr_h, adr_l, size, adr;

    switch (Regs->PC) {
    case 0xff84: // cbmk_ioinit
    case 0xff8a: // cbmk_restor
    case 0xffba: // cbmk_setlfs
        handled = 1;
        break;
    case 0xffe4: // cbmk_getin
        Regs->AC = 'A'; // like getchar();
        handled = 1;
        break;
    case 0xffd2: // cbmk_bsout
        if (Regs->AC & 0x80) { // Handle reverse chars
            int ch = Regs->AC & ~0x80;
            putchar((ch >= 'A' && ch <= 'Z') ? ch + 'a' - 'A' : ch);
        } else
            putchar(Regs->AC);
        if (Regs->AC == 0xD)
            putchar(0xA);
        handled = 1;
        break;
    case 0xffd5: // cbmk_load
        handled = 1;

        strcpy(tmp, search_path);
        strcat(tmp, fname);
        strcat(tmp, ".prg");

        fp = fopen(tmp, "rb");
        if (fp == NULL) {
            Regs->SR |= CF;
            break;
        }

        adr_l = fgetc(fp) & 0xFF;
        adr_h = fgetc(fp) & 0xFF;
        adr = adr_l + adr_h * 256;

        fseek(fp, 0, SEEK_END);
        size = ftell(fp) - 2;
        fseek(fp, 2, SEEK_SET);

        fread(Mem+adr, 1, size, fp);
        fclose(fp);
        Regs->SR &= ~CF;
        build_tables();
        break;
    case 0xffbd: // cbmk_setnam
        memcpy(fname, Mem + (int)Regs->XR + (int)Regs->YR*256, Regs->AC);
        fname[Regs->AC] = '\0';
        handled = 1;
        break;
    default:
        handled = 0;
        break;
    }

    if (handled) { /* Simulate RTS */
        unsigned lo = Pop (Regs);
        Regs->PC = lo + (Pop (Regs) << 8) + 1;
    }
}

int main()
{
    char test_start[_MAX_PATH];

    strcpy(test_start, search_path);
    strcat(test_start, "start.prg");

    if (load_prg(test_start)) {
        CPU = CPU_6502X;
        run_prg(0x801);
    }
    return 0;
}
```
Hopefully, as you can see, it is quite small and somehow not too difficult to write.
It was more complicated to make all the tests working, actually.
However, it needs that you temporally remove the `static` keyword from the declaration of the `Regs` structure into `6502.c`, otherwise the test application won't find it.
Then, I created a project with MSVC with these 4 files:
```
test.c
6502.c
error.c
memory.c
```
Adjust the value of `search_path[]` on top of `test.c`, build, link and run.
The program will execute all tests and it will hang when it executes TRAP1 test.
You can bypass this blocking point by changing the first name into `main.c` from `start.prg` to another file and continue with other tests, if you want.
Keep in mind that, probably, this test may not work on a case-sensitive file system, like ext2/3/4 on linux, because the testsuite loads the next test by using a function directly from the kernal. Obviously, it can be fixed if someone wants, but I left this out for simplicity.

> Also, can you check the extra tests in the CPU directory too? some will not work (because they are very C64 specific and need cycle exact interrupts and/or VICII), but about half of them are generic enough that they should work - in particular the various decimal mode tests, jsrselfmod, and some of the more elaborate undocs tests are interesting here (as these things are underrepresented in the lorenz suite).

After using some time for experiments, this looks a bit more complicated... ^^

> Also, did you test both regular (6502) and with undocs (6502X)?

Yes, I did.

mrdudz:
OK well, i will merge this as is - it's probably now better tested than the old implementation ever was...

Would still be great to have some kind of support for automatic testing, obviously.

carlo-bramini:
Thank you very much!
Perhaps, the only thing left to do is striking the issue at wiki page:

https://github.com/cc65/wiki/wiki/Before-the-next-release

immagine

PR #2336, opened 2024-01-05
Optimize strdup

colinleroy:
Changing _strlen to work with ptr4 (and updating its users) allows for simpler code in _strdup.
Results are :

    -12 bytes on disk
    -20 cycles per strdup (independant from the string length, it's before the memcpy)
    -6 cycles per strlen called from strdup (same remark)

Hope you'll like it?


mrdudz:
Thanks!

PR #2338, opened 2024-01-07
Remove non-standard (and useless) include

### LINKED ISSUE #2337: cc65: time.h header contains non standard identifiers in C89/C99 mode
It includes `target.h` which in turn includes all the target specific headers.

This was broken in #2315.

See #1670 for previous problems and reasoning.

I would like to repeat my suggestion from #1670: 

> Since some of the problems have been introduced in recent PRs you might think about an additional "standard compliance check" step whenever standard headers are changed.

I agree that this is difficult to automate, but there's a list of rules to be checked when reviewing a PR. How about adding something like "If standard C headers are changed, make sure non standard identifiers are hidden in standard compliant mode" or similar?

colinleroy:
Fixes #2337

### DIFF
--- a/include/time.h
+++ b/include/time.h
@@ -43,10 +43,6 @@ typedef unsigned long clock_t;
 
 
 
-#include <target.h>
-
-
-
 /* NULL pointer */
 #ifndef NULL
 #define NULL ((void *) 0)


mrdudz:
thanks!

PR #2344, opened 2024-01-09
[cc65] Cleanup for symbol types and flags

acqn:
This cleans up the definition and application of `SC_*` flags. Now the symbol types and storage classes are exclusive values instead of bits flags. To check for them you need bitmask the bits first and then compare the result to them rather than bitwise-and the bits with them directly.

These changes make the storage duration, linkage and visibility of symbols clearer as well as remove some hurdles for some bug fixes and future enhancements.

Note: `SC_` might mean "**S**torage **C**lass" in the beginning, but nowadays you may take it as "**S**ymbol **C**ategory".

### DIFF
--- a/src/cc65/asmstmt.c
+++ b/src/cc65/asmstmt.c
@@ -80,9 +80,9 @@ static void AsmErrorSkip (void)
 
 
 
-static SymEntry* AsmGetSym (unsigned Arg, unsigned Type)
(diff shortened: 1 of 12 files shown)


mrdudz:
Thanks! Lets see how bad that breaks my fp branch :/

PR #2345, opened 2024-01-09
[cc65] Fixed regression on comparison to null pointer

acqn:
Whilst the case could be very obscure, it is a stupid regression nonetheless.


mrdudz:
Thanks!

PR #2346, opened 2024-01-09
[cc65] Diagnosis fixes

acqn:
Note: the incorrect "Mixed declarations and code are not supported in cc65" diagnosis was once fixed in a previous PR but got "reverted" in a later PR by mistake. Now it's fixed again and checked in a testcase.


mrdudz:
thanks!

Issue #2137, opened 2023-06-15
Support C compilation to 65816? From discussion: So is SNES supported?


SeanRamey:
### Discussed in https://github.com/cc65/cc65/discussions/1886

><div type='discussions-op-text'>
>
><sup>Originally posted by **MrVeink** October 29, 2022</sup>
>Was trying to find an answer for this and I did find that ca65 can assemble for SNES but can you do C and have SNES as >the target? If not, is anyone into that?
>
>From [georgjz's guide](https://georgjz.github.io/snesaa01/):
>```
>MEMORY
>{
>    ROM0: start = $008000, size = $8000, fill = yes;
>}
>
>SEGMENTS
>{
>    CODE:   load = ROM0, align = $100;
>    VECTOR: load = ROM0, start = $00ffe4;
>}
>```
>Was looking at the [config files](https://github.com/cc65/cc65/tree/master/cfg) in the source and didn't find an SNES file >so this above could help.</div>

 So it seems like there really isn't any active C compilers for the 65816 cpu. The builds of tcc (supposedly from the unnofficial snes-sdk) that were linked to me all fail to build. There is an lcc compiler that also seems to be unmaintained, not working, and questionable whether it actually supports 65816 code. 

The only compiler that may actually work is the wdc compiler, and it's not open source, legally restricted to generating code for wdc cpus, windows only, and quite old. The only one I've found that actually is active is Orca C written in assembly and pascal for running on the AppleIIGS.

So, I have a request to add C compiler support for the 65816 and the SNES specifically.

mrdudz:
someone could start taking https://github.com/bauen1/cc65 and cleaning it up/fix it for the current compiler

polluks:
Yeah, this branch https://github.com/bauen1/cc65/tree/65816.

bbbradsmith:
I've been using cc65 to run C code on SNES as-is. Here is a public example with full source code: [SNESert Golfing](https://github.com/bbbradsmith/NESertGolfing/tree/snes)

The short version of how is just to use LoROM so that we can rely on memory available in all banks, and keep 8-bit register size by default. The 6502 code produced by the compiler will be compatible with 65816.

Some finer details:
* You can use HiROM too, but just run the C code from LoROM-like regions of the memory map.
* You can use `#pragma wrapped-call` to support far-calls.
[...]

PR #2348, opened 2024-01-09
Convert more time functions to asm

colinleroy:
_time_t_to_tm: 46 bytes size gain, -8% cycles on the unit tests
mktime: 415 bytes gain, -39% cycles on the unit tests
asctime: 7 bytes gain


libsrc/common/_time_t_to_tm.s
```diff
@@ -0,0 +1,121 @@
...
+        sta     ptr4
+        lda     #$00
+        sta     ptr4+1
+
+        ; Clear TM buf while we have zero in A
+        ldx     #.sizeof(tm)-1
+:       sta     TM,x
+        dex
+        bne     :-
+
```
Compyx, line 43: I assume you mean to clear all of `TM` here, in which case you'll need `bpl`, not `bne` here.

colinleroy: Indeed!

rofl0r:
i wonder if it's such a good idea to simply convert C functions to asm, as support for other CPUs might be upcoming (e.g. 65c816 for SNES), and then those will lack these routines. in musl libc, we have a .c version for all functions, but if a .s file of the same name exists in the selected arch directory, that file is used instead via some makefile magic.

mrdudz:
> i wonder if it's such a good idea to simply convert C functions to asm, as support for other CPUs might be upcoming (e.g. 65c816 for SNES), and then those will lack these routines. in musl libc, we have a .c version for all functions, but if a .s file of the same name exists in the selected arch directory, that file is used instead via some makefile magic.

First of all the chance of that ever happening is small. Very small in fact. And if it ever happened, you'll have to rewrite large parts of the runtime and library anyway to make proper use of the 65816 - if it really can't be implemented in conditional assembly, like it is done now for the 6502 variants. And if you really need the C versions back, you can always fish them from the repo history. Not that it will ever happen :)

rofl0r:
> Not that it will ever happen :)

ah, i was looking the other day on the SNES thread, and it appeared that the 65816 support was already (almost) complete in a forked repo.

mrdudz:
> > Not that it will ever happen :)
> 
> ah, i was looking the other day on the SNES thread, and it appeared that the 65816 support was already (almost) complete in a forked repo.

I am not aware of anything like that (what "SNES thread"?)

colinleroy:
The 65816 is a moot point, isn't it? It'll ingest 6502 assembly with no problem.

rofl0r:
> I am not aware of anything like that (what "SNES thread"?)

https://github.com/cc65/cc65/discussions/1886#discussioncomment-6178521
https://github.com/cc65/cc65/issues/2137

> The 65816 is a moot point, isn't it? It'll ingest 6502 assembly with no problem.

i never coded for it, but i recall that the manual mentioned something that it needs to be explicitly switched to/from 16 to 8 bit mode. how much of a problem that is in practice, *shrug*

mrdudz:
> > I am not aware of anything like that (what "SNES thread"?)
> 
> [#1886 (reply in thread)](https://github.com/cc65/cc65/discussions/1886#discussioncomment-6178521) #2137

Ok, five year old discussion, and a dead fork that noone even bothered to try since then. :)

> > The 65816 is a moot point, isn't it? It'll ingest 6502 assembly with no problem.
> 
> i never coded for it, but i recall that the manual mentioned something that it needs to be explicitly switched to/from 16 to 8 bit mode. how much of a problem that is in practice, _shrug_

Its no problem at all. IIRC it even starts in emulation mode, and will basically work like a 65C02 then.

colinleroy:
Hi,
Is anything missing for this PR to be merged?

mrdudz:
This one seems fine to merge, thanks!

PR #2351, opened 2024-01-11
mention John Dunning as original author

groessler:
I think we should give him credit.


mrdudz:
perhaps it should also say that Uz rewrote large parts of the compiler? AFAIK there is almost no original code left by now.

groessler:
How about the new version?

mrdudz:
thats great, thanks!

PR #2352, opened 2024-01-11
Rewrite more functions in asm

colinleroy:
fgetc : -82 bytes, -20% cycles
fgets : -104 bytes, -1% cycles
gets: +19 bytes if used alone (because it pulls in fgets), but -128 if the program also links fgets
(pmemalign: a low-hanging -36 bytes)


libsrc/runtime/returnFFFF.s
```diff
@@ -0,0 +1,20 @@
...
+        lda     #$FF
+        tax
+        rts
+
+.endproc
+
+
+
+
+
```
polluks, line 20: Hmm, no Uz memorial space :-)

rofl0r:
while rewriting simple functions like gets in asm is basically a no-brainer, rewriting complex logic like realloc() for the sake of 30% speedup (maybe introducing new bugs along the way) but sacrificing readability in the process isn't that great imo. the effort would be better spent improving the codegen...

colinleroy:
Hi,
Is anything missing for this PR to be merged?

mrdudz:
Yes, time for me to look :) I need to make my project work again first...

colinleroy:
I've found the bug in realloc.s, a simple wrong variable, nevertheless I'll push it in a separate PR once my stack of PRs will have flushed a bit :)

mrdudz:
Took a brief look - somehow this PR changes a lot of files, can't it be splitted somehow? That'd make it easier to pinpoint bugs later, should there be any.

Oh and btw

> while rewriting simple functions like gets in asm is basically a no-brainer, rewriting complex logic like realloc() for the sake of 30% speedup (maybe introducing new bugs along the way) but sacrificing readability in the process isn't that great imo. the effort would be better spent improving the codegen...

I don't agree with this at all - the codegen will never beat handwritten assembler - not even close, thats just wishful thinking. And having large parts of the library written in optimized assembly is one of the key points with cc65, it's why it works so well even with all the constraints.

colinleroy:
Hi!
Let me close this one and split it clean 😊

colinleroy:
Split into #2378 to #2384, they should be applied in order.

Issue #1428, opened 2021-03-15
Misleading warning


mrdudz:
not sure if this is a bug... at least it seems a bit odd

```
typedef unsigned long u32;
static u32 addr;

int main(void)
{
    addr = (32*1024*1024) - 1;
    while (addr < (32*1024*1024)) {      <- warning here
        return 0;
    }
    return -1;
}
```
```
$ cc65 -Osir -o main.s main.c
main.c(8): Warning: Comparison of unsigned type < 0 is always false
```
now i understand why it emits a warning, and even why it thinks its comparing with 0.... but - its odd

Bonus:

a stupid error:
```
typedef u32 unsigned long;  <- error here
```
makes it output:
```
$ cc65 -Osir -o main.s main.c
main.c(2): Error: ';' expected
main.c(2): Warning: Useless declaration
```
yeah i know, its wrong, but... say what? :=)

groessler:
Interestingly the warning goes away with this change:

```
--- x.c.org     2021-03-15 23:13:24.293581934 +0100
+++ x.c 2021-03-15 23:12:16.557295542 +0100
@@ -4,7 +4,7 @@
 int main(void)
 {
     addr = (32*1024*1024) - 1;
-    while (addr < (32*1024*1024)) {
+    while (addr < (32ul*1024*1024)) {
         return 0;
     }
     return -1;
```

groessler:
Maybe it's this: a constant without "L" or "UL" designator starts as 16-bit (signed) int, and then after the compiler detect that it doesn't fit into 16 bits, it sign-extends it to 32 bit. Which will result in a negative number.

mrdudz:
aargs, and it gets more nasty. Now, i admit i dont know what the standard says about it, and it might be technically correct what happens - but this worked before and chances are more code like this exists in my project and i cant identify it without getting a proper warning :(

```
#include <stdio.h>

void mmuset(unsigned long dest, unsigned char val, unsigned long len)
{
    printf("%08lx %02x %08lx\n", dest, val, len);
}

int main(void)
{
    mmuset(0x01000000, 0x1e, 0x2000 * 70);  /* len=0xffffc000 (original, previously working, code) */
    mmuset(0x01000000, 0x1e, 0x2000L * 70); /* len=0x0008c000 (ok) */
    mmuset(0x01000000, 0x1e, 0x2000 * 70L); /* len=0x0008c000 (ok) */
    return 0;
}
```
[...]

PR #2354, opened 2024-01-14
[cc65] Improved diagnosis

acqn:
- Added a warning on implicit `int` type in `typedef`s in post-C89 mode. See https://github.com/cc65/cc65/issues/1428#issuecomment-1890588844
- Removed the extra "unused parameter" warning when the parameter had an duplicated identifier error.
- Removed unnecessary warning on tagless enum/struct/unions that would be invisible out of a function declaration.
- Skipped anonymous tag names in diagnosis on empty structs/unions.
- Fixed repeated diagnosis when reading EOF in certain cases:
  - In unfinished declaration list of a K&R-style function definition.
  - In unfinished member declaration in a struct/union.
- Added warning on some code patterns of faulty attempt to declare anonymous structs/unions.
  - Using a _tagged_ struct/union to form an empty declaration (that declares no instances) - the standard requires a _tagless_ struct/union for the empty declaration.
  - Declaring outside a struct/union - the standard only supports anonymous structs/unions nested inside a struct/union.

### DIFF
--- a/src/cc65/datatype.c
+++ b/src/cc65/datatype.c
@@ -989,6 +989,24 @@ int IsIncompleteESUType (const Type* T)
 
 
 
+int IsAnonESUType (const Type* T)
+/* Return true if this is an anonymous ESU type */
+{
+    SymEntry* TagSym = GetESUTagSym (T);
+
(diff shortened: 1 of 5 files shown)


mrdudz:
Thanks!

PR #2356, opened 2024-01-14
Apple2: implement sleep using MONWAIT

colinleroy:
Also publish apple2_slowdown() and apple2_speedup() that could be useful to users (I do use those in some projects), and refactor to only store one apple2_ostype variable.
The sleep time is not precise to the millisecond of course, but close enough that it matches the chronometer on sleep(300) and sleep(600), and on
```
  while(1) {
    sleep(1);
    __asm__("bit $C030");
  }
```

### DIFF
--- a/doc/apple2.sgml
+++ b/doc/apple2.sgml
@@ -331,9 +331,25 @@ usage.
 <item>_filetype
 <item>_datetime
 <item>get_ostype
+<item>gmtime_dt
+<item>mktime_dt
 <item>rebootafterexit
 <item>ser_apple2_slot
 <item>tgi_apple2_mix
+</itemize>
+
+
+<sect1>Apple IIgs specific functions in accelerator.h<p>
+
+In addition to those, the <tt/accelerator.h/ header file contains three functions
+to help determine whether the program is running on a IIgs, and change the IIgs
+CPU speed. See the <url url="funcref.html" name="function reference"> for declaration and
+usage.
+
+<itemize>
+<item>detect_iigs
+<item>get_iigs_speed
+<item>set_iigs_speed
 </itemize>
 
 
--- a/libsrc/apple2/get_ostype.s
+++ b/libsrc/apple2/get_ostype.s
@@ -5,7 +5,7 @@
 ;
 
         .constructor    initostype, 9
-        .export         _get_ostype
+        .export         _get_ostype, ostype
 
 ; Identify machine according to:
 ; Apple II Miscellaneous TechNote #7, Apple II Family Identification
--- a/libsrc/apple2/sleep.s
+++ b/libsrc/apple2/sleep.s
@@ -1,0 +1,54 @@
+;
+; Colin Leroy-Mira <colin@colino.net>, 2024
+;
+; void __fastcall__ sleep(unsigned s)
+;
+;
+
+        .export         _sleep
+        .import         _get_iigs_speed
+        .import         _set_iigs_speed
+        .import         WAIT
+        .importzp       tmp1
+
+        .include        "accelerator.inc"
+
+        ; This functions uses the Apple2 WAIT ROM routine to waste a certain
+        ; amount of cycles and returns approximately after the numbers of
+        ; seconds passed in AX.
+        ;
+        ; It takes 1023730 cycles when called with AX=1 (1,0007s),
+        ; 10236364 cycles when called with AX=10 (10,006 seconds),
+        ; 306064298 cycles with AX=300 (299.2 seconds).
+        ;
+        ; Caveat: IRQs firing during calls to sleep will make the sleep longer
+        ; by the amount of cycles it takes to handle the IRQ.
+        ;
+_sleep:
+        stx     tmp1            ; High byte of s in X
+        tay                     ; Low byte in A
+        ora     tmp1
+        bne     :+
+        rts
+:       jsr     _get_iigs_speed ; Save current CPU speed
+        pha
+        lda     #SPEED_SLOW     ; Down to 1MHz for consistency around WAIT
+        jsr     _set_iigs_speed
+sleep_1s:
+        ldx     #$0A            ; Loop 10 times
+sleep_100ms:
+        lda     #$C7            ; Sleep about 99ms
+        jsr     WAIT
+        lda     #$0D             ; About 1ms
+        jsr     WAIT
+        dex
+        bne     sleep_100ms
+        dey
+        bne     sleep_1s
+        dec     tmp1
+        bmi     done
+        dey                     ; Down to #$FF
+        bne     sleep_1s
+done:
+        pla                     ; Restore CPU speed
+        jmp     _set_iigs_speed
--- a/libsrc/apple2/waitvsync.s
+++ b/libsrc/apple2/waitvsync.s
@@ -5,20 +5,10 @@
 ;
         .ifdef  __APPLE2ENH__
 
-        .constructor    initvsync
         .export         _waitvsync
-        .import         _get_ostype
+        .import         ostype
 
         .include        "apple2.inc"
-
-        .segment        "ONCE"
-
-initvsync:
-        jsr     _get_ostype
-        sta     ostype
-        rts
-
-        .code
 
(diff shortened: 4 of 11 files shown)


colinleroy:
This one's for @oliverschmidt  :)

libsrc/apple2/cpuspeed.s
```diff
@@ -0,0 +1,40 @@
...
+        .import         apple2_ostype
+
+        .include        "apple2.inc"
+
+initcyareg:
+        lda     apple2_ostype
+        bpl     :+
+        lda     CYAREG
+        sta     cyareg_orig
+:       rts
```
oliverschmidt, line 20: The user can change the system speed at any time using the Control Panel. It doesn't seem desirable to me to get the current speed at the start of the program.

libsrc/apple2/cpuspeed.s
```diff
@@ -0,0 +1,40 @@
...
+        and     #%01111111
+        sta     CYAREG
+:       rts
+
+_apple2_speedup:
+        lda     apple2_ostype
+        bpl     :+
+        lda     cyareg_orig
+        sta     CYAREG
+:       rts
```
oliverschmidt, line 35: Given that you restore the original speed, the name "speedup" seems to be a misnomer to me.

libsrc/apple2/sleep.s
```diff
@@ -0,0 +1,41 @@
...
+; void __fastcall__ sleep(unsigned s)
+;
+;
+
+        .export         _sleep
+        .import         _apple2_slowdown
+        .import         _apple2_speedup
+        .importzp       tmp1
+
+MONWAIT         := $FCA8
```
oliverschmidt, line 13: Are you aware that there's a pattern regarding ROM calls in the Apple II library? See i.e.:
https://github.com/cc65/cc65/blob/master/libsrc/apple2/cout.s
https://github.com/cc65/cc65/blob/master/libsrc/apple2/home.s
I'd appreciate if you would follow that pattern here. BTW: The ROM listing in the original Apple II reference manual calls the routing just `WAIT`. I propose to stick to that name.

libsrc/apple2/cpuspeed.s
```diff
@@ -0,0 +1,40 @@
+;
+; Colin Leroy-Mira <colin@colino.net>, 2024
+;
+; void __fastcall__ apple2_slowdown(void)
+; void __fastcall__ apple2_speedup(void)
```
oliverschmidt, line 5: Are you aware of https://github.com/cc65/cc65/blob/master/include/accelerator.h?
You may be surprised to find quite some references to Apple II in https://github.com/cc65/cc65/blob/master/include/accelerator.h#L45-L55

Some background: Someone needed to implement tests for the accelerator support in VICE. And there's was a momentum to implement those tests as cc65 library code in order to have a synergy effect. It was planned to extend the cc65 library to support accelerators on other targets as well. However, for "some reason or another" this didn't happen. And what too didn't happen, was to implement some umbrella interface that allows the user to interact with the accelerator speed independently from the actual accelerator.

So my proposal is to implement three functions
`set_iigs_speed()`
`get_iigs_speed()`
`detect_iigs()`
and add them to https://github.com/cc65/cc65/blob/master/include/accelerator.h

libsrc/apple2/get_ostype.s
```diff
@@ -5,7 +5,7 @@
 ;
 
         .constructor    initostype, 9
-        .export         _get_ostype
+        .export         _get_ostype, apple2_ostype
```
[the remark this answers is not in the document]
colinleroy: I didn't touch stdjoy as it's a module and I was not sure I wouldn't break it. I'll investigate it more :)

libsrc/apple2/cpuspeed.s
colinleroy: Oh, indeed. I thought it was at boot time only.

libsrc/apple2/cpuspeed.s
colinleroy: I was not! I'll investigate that and rework the PR.
Thanks !

libsrc/apple2/cpuspeed.s
oliverschmidt: But actual speed can certainly be changed at any time. But I seem to remember that I had problems with CYAREG not reflecting that actual speed...

libsrc/apple2/get_ostype.s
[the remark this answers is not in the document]
oliverschmidt: I don't have a good overview, but concerned that building and using the stdjoy as static module might lead to a conflict of the global and local `ostype` symbols. But I'd really like to have the new global symbol be called `ostype`. It's in general no cc65 pattern to apply a target prefix to "all kinds of" symbols.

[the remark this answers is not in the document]
colinleroy: It works fine (verified under MAME's debugger) with the same ostype name. In fact, as it's a module, even if I wanted to import ostype, it's not recognized.

colinleroy:
Thanks for the pointers, everything should be updated following your advice.

libsrc/apple2/get_ostype.s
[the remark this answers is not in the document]
oliverschmidt: > In fact, as it's a module, even if I wanted to import ostype, it's not recognized.

The static variant is no module.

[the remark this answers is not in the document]
colinleroy: ah, indeed :) but still there's no clash of the global one over the other.

doc/apple2.sgml
```diff
@@ -334,6 +334,9 @@ usage.
 <item>rebootafterexit
 <item>ser_apple2_slot
 <item>tgi_apple2_mix
+<item>detect_iigs
+<item>get_iigs_speed
+<item>set_iigs_speed
```
oliverschmidt, line 339: I'm not sure, if desirable to list those functions here. Doesn't it imply that they become available by including `apple2[enh].h`?

oliverschmidt (changes requested):

libsrc/apple2/accelerator.s
```diff
@@ -0,0 +1,52 @@
...
+        ; Returns SPEED_FAST or SPEED_SLOW
+_get_iigs_speed:
+        lda     ostype          ; Return SLOW if not IIgs
+        bpl     :+
+        lda     CYAREG          ; Check current setting
+        bpl     :+
+        lda     #SPEED_FAST
+        ldx     #$00
+        rts
+        .assert SPEED_SLOW = 0, error
```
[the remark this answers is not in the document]
colinleroy: Me too! Makes sure we don't break stuff by mistake.

libsrc/apple2/waitvsync.s
```diff
@@ -5,21 +5,11 @@
...
 
-        .segment        "ONCE"
-
-initvsync:
-        jsr     _get_ostype
-        sta     ostype
-        rts
-
-        .code
-
```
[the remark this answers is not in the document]
colinleroy: Thanks!

libsrc/apple2/accelerator.s
```diff
@@ -0,0 +1,52 @@
...
+        cpx     #SPEED_SLOW
+        beq     :+
+        ora     #%10000000
+        bne     set_speed
+:       and     #%01111111
+set_speed:
+        sta     CYAREG
+        txa
+        ldx     #$00
+        rts
```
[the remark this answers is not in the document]
colinleroy: Indeed.

doc/apple2.sgml
colinleroy: Indeed, kind of. And on the other hand, I should have added mktime_dt and gmtime_dt there.

colinleroy: What do you think of that change?
I feel like it's good to mention those in the all-things-apple2 page, but it's, indeed, better to split those three from the ones defined in apple2.h.

oliverschmidt (approved):

Issue #2357, opened 2024-01-15
Compiler produces invalid code


mrdudz:
I came back to a project of mine and noticed cc65 produces code that does not assemble (branch too far). Bisecting shows this happened after commit d8a3938f2b8d6c53c2f75b5a07964e8beef746a5

@colinleroy would you have a look please? It looks like the optimizer replaces some JNEs by BNEs when it shouldn't do that

Attached is a somewhat stripped down test case

```
cl65 --add-source -Osir --codesize 180 -Wc --local-strings -t none -o bug.prg bug.c
```

```
#include <string.h>

typedef unsigned char FRESULT;
typedef unsigned char	BYTE;
typedef unsigned long	DWORD;
typedef unsigned short	WORD;

typedef BYTE	DRESULT;

typedef struct _FATFS_ {
	BYTE	fs_type;	/*!< FAT sub type */
	BYTE	csize;		/*!< Number of sectors per cluster */
	BYTE	n_fats;		/*!< Number of FAT copies */

	BYTE	wflag;		/*!< win[] dirty flag (1:must be written back) */
	BYTE	fsi_flag;	/*!< fsinfo dirty flag (1:must be written back) */
	WORD	n_rootdir;	/*!< Number of root directory entries (0 on FAT32) */
	DWORD	last_clust;	/*!< Last allocated cluster */
	DWORD	free_clust;	/*!< Number of free clusters */
	DWORD	fsi_sector;	/*!< fsinfo sector */
	DWORD	cdir;		/*!< Current directory (0:root)*/
	DWORD	sects_fat;	/*!< Sectors per fat */
	DWORD	max_clust;	/*!< Maximum cluster# + 1. Number of clusters is max_clust - 2 */
	DWORD	fatbase;	/*!< FAT start sector */
	DWORD	dirbase;	/*!< Root directory start sector (Cluster# on FAT32) */
	DWORD	database;	/*!< Data start sector */
	DWORD	winsect;	/*!< Current sector appearing in the win[] */
} FATFS;

FATFS FatFsObj; /* actually extern, but both produces the error */
#pragma zpsym ("FatFsObj")          /* this is the problem */

#define FR_OK 0
#define FS_FAT32	3
#define BS_55AA			510

#define	FSI_LeadSig		0
#define	FSI_StrucSig		484
#define	FSI_Free_Count		488
#define	FSI_Nxt_Free		492

BYTE FatFsObjwin[512];

[...]

PR #2361, opened 2024-01-16
Cleaner instruction updating for long opt

colinleroy:
As part of #2357 investigation.


mrdudz:
Ok i am merging this, its obviously needed anyway

PR #2367, opened 2024-01-18
[cc65] Fixed segname pragmas right after a function definition

### LINKED ISSUE #2365: data goes into the wrong section/segment when segment is changed more than once
Sorry for odd description :) I am just done bisecting.... 20c3e994c6 is the first non working commit. 

It breaks so odd and weird, i have no idea what i am even looking for....

@acqn perhaps you can give a hint on what kind of thing could be a problem (yes i am using pragmas a lot, mostly to put things into various segments/banks)

acqn:
Fixed #2365?


mrdudz:
Thank you - with this my code is back to normal (as far as i can see)

PR #2368, opened 2024-01-18
Add beep for apple2

colinleroy:


polluks:
What about `putchar('\a');`? :-)

colinleroy:
> What about `putchar('\a');`? :-)

Good question :) We talked about it with Oliver and thought it would be good to have a 16-byte standalone version for people not wanting to link putchar + fwrite + write.

Plus,
```
        lda     CH              ; Bell scrambles CH in 80col mode on IIgs, storing
        pha                     ; it in OURCH and resetting CH to 0. Save it.
```

polluks:
I know >600 bytes... the bug is in every GS ROM?

colinleroy:
> I know >600 bytes... the bug is in every GS ROM?

No, actually it's a bug in our implementation of conio. We should use different addresses (OURCH and OURCV instead of CH and CV) in 80 columns mode. But that's a much bigger fix, and one costing bytes and cycles too.

libsrc/apple2/beep.s
```diff
@@ -0,0 +1,25 @@
...
+
+        lda     CH              ; Bell scrambles CH in 80col mode on IIgs, storing
+        pha                     ; it in OURCH and resetting CH to 0. Save it.
+        jsr     $FF3A           ; BELL
+        pla
+        sta     CH              ; Restore CH
+
+        ; Switch in LC bank 2 for R/O and return
+        bit     $C080
+        rts
```
oliverschmidt, line 25: Sorry to nitpick once more, but I'd like to have this done like in:
https://github.com/cc65/cc65/blob/master/libsrc/apple2/clrscr.s
https://github.com/cc65/cc65/blob/master/libsrc/apple2/home.s

In other words: There's always the ROM banking wrapper named like the ROM symbol (here `BELL`) and then building on this there's any library code (even it's only an alias like clrscr()).

libsrc/apple2/beep.s
```diff
@@ -0,0 +1,20 @@
...
+
+        .export         _beep
+        .import         BELL
+
+        .include        "apple2.inc"
+
+        .segment        "LOWCODE"
+
+_beep:
+        lda     CH              ; Bell scrambles CH in 80col mode on IIgs, storing
```
oliverschmidt, line 15: I didn't know that. Does this mean that on the IIgs, BELL is more or less a wrapper around COUT with A set to 7 ?

PR #2371, opened 2024-01-19
Lynx upload fixes for IRQ and configuration

### LINKED ISSUE #2369: Lynx:: Uploader implementation and configuration does not work
The current implementation of the uploader for the Atari Lynx is non-functional and misconfigured.
Details:
- The `uploader.s` implementation does not reset the IRQ bit for timer 4. This is a hardware bug that an implementation needs to compensate. When the IRQ bit is set to indicate the serial interrupt, the bit must be reset, because the hardware is level-sensitive. This requires a write of $10 to address INTRST.
- The current lynx-uploader configuration loads the UPCODE and UPDATA segments into a separate memory area, This requires the program to load the segments from a file. The __DEFDIR__ only caters for a single file.

There are two possible fixes for the configuration: 
1. Use the MAIN memory area for UPCODE and UPDATA, effectively loading it in the same main executable file. No separate loading of a file and a new directory structure is needed.
2. Create a directory with at least two entries, one for the main executable and one for the UPLDR file based on UPCODE and UPDATA.

The downside of option 1 is that the UPCODE will be low in memory and most likely be overwritten by the uploaded code. 
Opting for option 2 is a better solution, as this is the most common scenario, with the UPLDR file being loaded on demand, right before uploading starts. This will require the definition of a custom directory and changes to the directory size definition in the config. 

Bottom line: the configuration lynx-uploader.cfg cannot be used as is, because specific code needs to be used: activating timer 4 and loading file (in case of option 2).

Proposing a change to `uploader.s` to reset the IRQ bit for timer 4 and an adjustment of the lynx-uploader.cfg` to reflect the new uploader code size and a directory with at least two entries.

alexthissen:
Fix #2369
Changes for the uploader implementation for the Atari Lynx library.

- The interrupt handler logic is adjusted to clear timer 4 bit for serial IRQs. This IRQ bit is level-sensitive instead of edge-sensitive due to a hardware bug and must reset to avoid firing the interrupt continuously as long as the bit is set. 
- The feedback is now given from palette color green on pen 1 instead of 0. This works better for the default cc65 Lynx color palette
- Uploader configuration `lynx-uploader.cfg` is adjusted. Encourages a second directory entry for a separate file to be loaded with the uploader logic in `UPCODE` and `UPDATA`. Size of the `UPCODE` code is taken into account for RAM area.
- Removed redundant `__BLOCKSIZE__` and changed it to __BANK0BLOCKSIZE__ where it was still being referenced in the `libsrc\lynx` files (`lseek.s` and `lynx-cart.s`)

### DIFF
--- a/cfg/lynx-uploader.cfg
+++ b/cfg/lynx-uploader.cfg
@@ -5,16 +5,17 @@ SYMBOLS {
     __BANK1BLOCKSIZE__:   type = weak, value = $0000; # bank 1 block size
     __EXEHDR__:           type = import;
     __BOOTLDR__:          type = import;
-    __DEFDIR__:           type = import;
     __UPLOADER__:         type = import;
+    __UPLOADERSIZE__:     type = export, value = $61;
+    __HEADERSIZE__:       type = export, value = 64;
 }
 MEMORY {
     ZP:     file = "", define = yes, start = $0000, size = $0100;
-    HEADER: file = %O,               start = $0000, size = $0040;
+    HEADER: file = %O,               start = $0000, size = __HEADERSIZE__;
     BOOT:   file = %O,               start = $0200, size = __STARTOFDIRECTORY__;
-    DIR:    file = %O,               start = $0000, size = 8;
-    MAIN:   file = %O, define = yes, start = $0200, size = $BD38 - __STACKSIZE__;
-    UPLDR:  file = %O, define = yes, start = $BFDC, size = $005C;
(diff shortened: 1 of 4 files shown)


karrika:
Is there a reason why we would support two roms with different block sizes? I know that the lnx header has two fields. But I fail to see the advantage in it. Perhaps a better fix would be to change the exehdr to state the same blocksize for both ROMs?

mrdudz:
I'll have to trust you that this is the right thing to do :) Thanks!

PR #2372, opened 2024-01-19
[cc65] Added primitive support for the ISO C99 inline feature as well as the __inline__ extension

acqn:
No inlining is actually done but that part is not required by the standard.


mrdudz:
Thanks!

PR #2374, opened 2024-01-19
Save two bytes in pushax and popptr1

colinleroy:
It's not because Y must equal zero on rts that we should'nt spare one byte and one cycle.

### DIFF
--- a/libsrc/runtime/popptr1.s
+++ b/libsrc/runtime/popptr1.s
@@ -8,12 +8,18 @@
         .import         incsp2
         .importzp       sp, ptr1
 
+        .macpack        cpu
+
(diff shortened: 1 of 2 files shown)


mrdudz:
Well spotted - thanks!

colinleroy:
Yay, one less in my queue :)
Do you have an idea on an ETA for the other ones? (I don't want to pressure you, tell me if I'm annoying, it's just that I'm starting to have just a bit too many branches locally :rofl:)

mrdudz:
No idea, sorry - my time budget for this is very limited - and i have to carefully look at those things. You can speed it up by adding more tests (which admittedly isn't easy for some things)

colinleroy:
Thanks :) I'll see if I can think of more tests.

PR #2375, opened 2024-01-21
Apple2: Swap in ROM before exec()

colinleroy:
Constructors expect ROM swapped in at start. If we don't swap it in right before exec()'ing the next program, and the next program uses constructors that need the ROM swapped in, the program explodes if it was not in.

This does not always happen: _exit() swaps in ROM fairly late, [at line 44](https://github.com/cc65/cc65/blob/a173428fabfd8c362ac072ac680c61be45e880c4/libsrc/apple2/crt0.s#L44). 
However, the modules destructors are called right after that. If you had ser_install(&a2e_gs_ser), it will be ser_uninstall()'d, and there will [switch to ROM, jsr $FE1F, switch back to RAM](https://github.com/cc65/cc65/blob/a173428fabfd8c362ac072ac680c61be45e880c4/libsrc/apple2/ser/a2.gs.s#L282). 

Then exec() will start the new program with RAM switched in, and initostype() will [call $FE1F](https://github.com/cc65/cc65/blob/a173428fabfd8c362ac072ac680c61be45e880c4/libsrc/apple2/get_ostype.s#L17), and fail :)

as ser_install() and ser_uninstall() share the same code path, I'm not sure it's the best place to fix that problem.

Rather, I suggest with swap in ROM right before we jump into the new program. We could also swap in ROM in initostype(), but I feel that's safer to do it right where we want to be sure it's in.

Minimal reproducer, this TEST program:
```
#include <conio.h>
#include <serial.h>
#include <stdio.h>
#include <unistd.h>

int main(int argc, char *argv[]) {
  ser_install(&a2e_gs_ser);
  printf("OS: %d\n", get_ostype());

  cgetc();
  exec("TEST", NULL);
}
```

### DIFF
--- a/doc/apple2.sgml
+++ b/doc/apple2.sgml
@@ -62,7 +62,7 @@ Special locations:
 </descrip><p>
 
 While running <tt/main()/ the Language Card bank 2 is enabled for read access.
-However while running module constructors/destructors the Language Card is disabled.
+However while running module constructors the Language Card is disabled.
 
 Enabling the Language Card allows to use it as additional memory for cc65
 generated code. However code is never automatically placed there. Rather code
--- a/doc/apple2enh.sgml
+++ b/doc/apple2enh.sgml
@@ -63,7 +63,7 @@ Special locations:
 </descrip><p>
 
 While running <tt/main()/ the Language Card bank 2 is enabled for read access.
-However while running module constructors/destructors the Language Card is disabled.
+However while running module constructors the Language Card is disabled.
 
 Enabling the Language Card allows to use it as additional memory for cc65
 generated code. However code is never automatically placed there. Rather code
--- a/libsrc/apple2/crt0.s
+++ b/libsrc/apple2/crt0.s
@@ -40,11 +40,14 @@ _exit:  ldx     #<exit
         lda     #>exit
         jsr     reset           ; Setup RESET vector
 
-        ; Switch in ROM, in case it wasn't already switched in by a RESET.
-        bit     $C082
+        ; Switch in LC bank 2 for R/O in case it was switched out by a RESET.
+        bit     $C080
 
         ; Call the module destructors.
         jsr     donelib
+
+        ; Switch in ROM.
+        bit     $C082
 
         ; Restore the original RESET vector.
 exit:   ldx     #$02


oliverschmidt:
Is it possible that there's something missing in your description? Where's the relationship between `ser_uninstall` and destructors? Either I'm missing the point or there's some `atexit(ser_uninstall)`.

https://cc65.github.io/doc/apple2.html#s3 states:

_While running main() the Language Card bank 2 is enabled for read access. However while running module constructors/destructors the Language Card is disabled._

It seems to me that the original problem is that the doatexit destructor calls user code with the language card disabled. I think, user code running as part of doatexit should run with the language card enabled.

Maybe we should review the decision to run destructors with the language card disabled in general. Maybe no destructor really benefits from it. And then doatexit wouldn't require special handling in the first place.

colinleroy:
Hi Oliver!
Yes, ser-kernel.s defines
```
        .destructor     _ser_uninstall
```

Whether we run the destructors with LC enabled or not wouldn't change much: the IIgs driver does swap to ROM briefly to verify it's running on a IIgs, both at ser_install and ser_uninstall time, and then back to LC... Which means that "the decision to run destructors with the language card disabled" is already compromised.

oliverschmidt:
I see, I got this wrong because `_ser_uninstall` was missing in https://github.com/cc65/wiki/wiki/Priority-of-constructors,-destructors,-and-interruptors. I added it now.

But then I'm surprised by your proposals. As of today https://github.com/cc65/cc65/blob/master/libsrc/apple2/ser/a2.gs.s#L284 is a bug that needs to be fixed. Simply because it doesn't take...

_While running main() the Language Card bank 2 is enabled for read access. However while running module constructors/destructors the Language Card is disabled._

...into account. There's no point in just breaking this rule and then thinking about hacks to make the rest of the code get along with the broken system.

colinleroy:
Another possible fix would be to use _get_ostype as libref in a2.gs.ser and not use $FE1F there, if you would prefer that.

oliverschmidt:
However, as I pointed out above, it might be desirable to change the rule in question. There are only two Apple II specific destructors and both seem to not rely on the ROM at all. So changing in crt0.s ...
```
        ; Switch in ROM, in case it wasn't already switched in by a RESET.
        bit     $C082

        ; Call the module destructors.
        jsr     donelib
```
... to ...
```
        ; Switch in Switch in LC bank 2 for R/O in case it was switched out by a RESET.
        bit     $C080

        ; Call the module destructors.
        jsr     donelib

        ; Switch in ROM.
        bit     $C082
```
... and changing ...

_While running main() the Language Card bank 2 is enabled for read access. However while running module constructors/destructors the Language Card is disabled._

... to ...

_While running main() the Language Card bank 2 is enabled for read access. However while running module constructors the Language Card is disabled._

... changes the rule in a way that makes both doatexit and your SER_UNINSTALL work. I'd appreciate if you would try that.

colinleroy:
Hi Oliver,
It does fix the problem and seems indeed like a good solution. Sorry I missed that rule when implementing the IIgs serial driver!

oliverschmidt:
I didn't have _any_ problem with the IIgs serial driver introducing the issue! It was very far from obvious and it should have rather been me noticing it. I was _only_ surprised by your ideas in this PR to solve the issue.

PR #2377, opened 2024-01-23
[cc65] Diagnosis improvements

acqn:
- Added missing checks for forward declarations of the main() function.
- More accurate diagnosis on implicit 'int' type specifiers.


mrdudz:
Thanks!

PR #2378, opened 2024-01-24
Fix malloc and realloc overflow

### LINKED ISSUE #2358: malloc(65535UL) to malloc(65532UL) succeed while they should not
FYI, found while working on #2352. When we add #HEAP_ADMIN_SPACE to the user requested size, no carry check is done, and if user asks for a 65535 byte buffer, we'll "add" .sizeof(usedblock) (4) to it, get a new size of 3, and happily allocate #HEAP_MIN_BLOCKSIZE (6 bytes).
Same for realloc.

colinleroy:
If user requests a size >= 65532, adding the heap admin size overflows size, both on malloc and realloc.
Fix it by checking carry in malloc.s, fix it with a simple test in realloc.s, and add malloc and realloc tests.

Fixes #2358.

### DIFF
--- a/libsrc/common/malloc.s
+++ b/libsrc/common/malloc.s
@@ -131,6 +131,7 @@ _malloc:
         sta     ptr1
         bcc     @L1
         inc     ptr1+1
+        beq     OutOfHeapSpace          ; if high byte's 0, we overflowed!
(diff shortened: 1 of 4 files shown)


mrdudz:
This looks OK, thanks!

PR #2379, opened 2024-01-24
add a return -1 helper

colinleroy:
@mrdudz hi :) this one I made because I wanted to have it in the fgetc/fgets/gets rewrites, so it should be merged before the other ones I'm going to push!


mrdudz:
This one looks fine, thanks!

PR #2380, opened 2024-01-24
Rewrite fgetc in asm

colinleroy:
-82 bytes, -20% cycles

This one depends on #2379 and will not pass tests before #2379 is merged :)


mrdudz:
mmh somehow i dont like the test output being generated in the test/ref/ directory... cant we make it go into the testwrk dir where the binaries are? (Are there other tests that do the same thing?)

colinleroy:
> mmh somehow i dont like the test output being generated in the test/ref/ directory... cant we make it go into the testwrk dir where the binaries are? (Are there other tests that do the same thing?)

I wanted to, but didn't manage. I could try again with some more effort if you want. I also tried to unlink the file but it's not implemented in sim65.
And no, no other test do that.

colinleroy:
Hi @mrdudz,
I must have been tired: I found an elegant solution to make sure the file created is right next to the test binary.
HTH!

mrdudz:
Looks OK, thanks!

PR #2381, opened 2024-01-24
Rewrite fgets in asm

colinleroy:
-104 bytes, -1% cycles

Does not depend on #2380, but is tested by the unit test added in #2380


mrdudz:
Seems OK - Thanks!

PR #2382, opened 2024-01-24
Rewrite gets in asm

colinleroy:
+19 bytes if used alone, because it pulls in fgets, but as code is factorized, -128 bytes in programs using both fgets and gets.

Depends on  #2381, test will fail before it's merged.


mrdudz:
Mmmmh #2381 is merged, but re-running the check still fails *shrug*

colinleroy:
Indeed, of course, it still needed a rebase :) Pushed!

mrdudz:
thanks!

PR #2383, opened 2024-01-24
Optimize -36 bytes out of posix_memalign

colinleroy:
And add a unit test


mrdudz:
This looks OK, thanks!

PR #2387, opened 2024-01-26
Implement __sysremove for sim65

colinleroy:
This will allow using unlink()/remove() in sim65 programs
Use it to unlink fgets' test output file

### DIFF
--- a/libsrc/sim6502/paravirt.s
+++ b/libsrc/sim6502/paravirt.s
@@ -8,10 +8,15 @@
 ;
 
         .export         exit, args, _open, _close, _read, _write
+        .export         __sysremove, ___osmaperrno
 
+__sysremove     := $FFF2
+___osmaperrno   := $FFF3
 _open           := $FFF4
 _close          := $FFF5
 _read           := $FFF6
 _write          := $FFF7
 args            := $FFF8
 exit            := $FFF9
+
+                ; $FFFA-FFFF are hardware vectors, extend before not after!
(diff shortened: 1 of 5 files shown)


mrdudz:
Good idea - but this patch is broken :) the "jumptable" ends at $fff9 for a reason - $fffa is the NMI vector. So you'll have to put new functions at the top of the table (and adjust the base address of the table), not at the bottom :)

mrdudz:
Oops wrong click :)

Also, perhaps add some comment(s) at strategic places that point out this ("$fffa-$ffff are the hardware vectors")

polluks:
Who likes this proprietary jump table at all? How about using the well-defined DOS/65 for sim65's I/O?
http://www.z80.eu/dos65/DOS-65_System_Interface_Guide_A.pdf

colinleroy:
> Oops wrong click :)
> 
> Also, perhaps add some comment(s) at strategic places that point out this ("$fffa-$ffff are the hardware vectors")

I fell in the trap, didn't notice that!
PR updated.

polluks:
Maybe we can add some kind of segment for protection?
`VECTORS:  load = ROM,            type = ro,  start    = $FFFA;`

mrdudz:
> Maybe we can add some kind of segment for protection? `VECTORS: load = ROM, type = ro, start = $FFFA;`

Makes sense - feel free to make a PR :)

mrdudz:
Merging... thanks!

PR #2389, opened 2024-01-28
Fix #2388

### LINKED ISSUE #2388: test_gets.c fails on windows
what the topic says. must be a LF vs CRLF problem - not sure how to fix it :)

@colinleroy could you have a look?

colinleroy:
@mrdudz, can you test if this works?

(If it does, it contradicts https://learn.microsoft.com/en-us/cpp/c-runtime-library/gets-getws?view=msvc-170 that mention nothing of \r).
Also, cf.in has Unix file format with \n line endings...

### DIFF
--- a/test/ref/test_gets.c
+++ b/test/ref/test_gets.c
@@ -30,9 +30,7 @@ char *argv[];
 #endif
 {
     /* Fake stdin with the reference file */
-    fclose(stdin);
-    stdin = fopen(INFILE, "r");
-    if (stdin == NULL) {
+    if (freopen(INFILE, "rb", stdin) == NULL) {
         return EXIT_FAILURE;
     }
 


mrdudz:
> If it does, it contradicts https://learn.microsoft.com/en-us/cpp/c-runtime-library/gets-getws?view=msvc-170 that mention nothing of \r

Actually... no. When you are opening the file in textmode ("r"), then the line endings will be translated before fgets even sees them. and when there is no CRLF, then it wont see any line ending - which is what makes it fail.

So this should work, i'll have to try later on my windows box - no idea how to re-trigger that check here including your PR (really should add the windows test to the regular PR checks)

colinleroy:
It was failing at build time in fact. (but also reopening binary is needed). 
I've read https://learn.microsoft.com/en-us/cpp/c-runtime-library/stdin-stdout-stderr?view=msvc-170, "These pointers are constants, and can't be assigned new values. The [freopen](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/freopen-wfreopen?view=msvc-170) function can be used to redirect the streams to disk files or to other devices", and updated accordingly.

mrdudz:
thanks, merging this - lets see what happens :)

colinleroy:
> thanks, merging this - lets see what happens :)

According to my tests on my fork, where I temporarily enabled Windows tests on push, it should work.
By the way, they're not that slow, I think it wouldn't be too much of an annoyance to activate.

mrdudz:
> By the way, they're not that slow, I think it wouldn't be too much of an annoyance to activate.

I'd happily merge a PR that does this, go ahead :)

PR #2390, opened 2024-01-28
Implement aslax7/shlax7/asrax7/shrax7

colinleroy:
Size stats:
every call to 
```
jsr shlax4
jsr shlax3
```
are replaced by
```
jsr shlax7
```
So that's -3 bytes per call.

_Numbers updated following Miłosław's optimisation of the optimisation:_

When optimized (for char only), nothing changes, the optimizer inserts seven `asl` (instead of inserting 4 `asl` then 3 `asl`).
Of course if one program's include shifts by 4, 3, and 7, more code is linked (+10 or +9 bytes). It could be offset by the number of calls, or maybe not. That's the very little drawback of this PR.

In terms of cycles, 87 cycles before for shrax4+shrax3, 17.5 cycles after (average over 105k calls) (counting rts)


libsrc/runtime/aslax7.s
```diff
@@ -0,0 +1,21 @@
+;
+; Miloslaw Smyk, 2024
+;
+; CC65 runtime: Scale the primary register by 128, unsigned
+;
+
+        .export         shlax7, aslax7
+
+aslax7:
```
mrdudz, line 9: are you sure aslax7 shouldn't preserve the sign?

test/val/lib_runtime_shlax7.c
```diff
@@ -0,0 +1,44 @@
...
+#include <stdlib.h>
+#include <string.h>
+#include <limits.h>
+#include <errno.h>
+
+int main (void)
+{
+    unsigned int ai = 0, ti, refi;
+    unsigned char ac = 0, tc, refc;
+
```
mrdudz, line 18: This seems broken - how can the test yield useful results when the bits are all 0 ?

test/val/lib_runtime_shrax7.c
```diff
@@ -0,0 +1,44 @@
...
+#include <stdlib.h>
+#include <string.h>
+#include <limits.h>
+#include <errno.h>
+
+int main (void)
+{
+    unsigned int ai = 0, ti, refi;
+    unsigned char ac = 0, tc, refc;
+
```
mrdudz, line 18: This seems broken - how can the test yield useful results when the bits are all 0 ?

test/val/lib_runtime_shlax7.c
colinleroy: The idea was to test every possible number, but of course I forgot the ++ :sweat_smile:

test/val/lib_runtime_shrax7.c
colinleroy: Same :)

libsrc/runtime/aslax7.s
colinleroy: Hi!
Yes. I was surprised about it too at first when looking at the other aslax*, but that's normal: https://en.wikipedia.org/wiki/Arithmetic_shift

mrdudz:
looks ok, lets see what happens :)

PR #2391, opened 2024-01-30
helloworld-c64

netpipe:
hello world demo for c64 in assembly


mrdudz:
This PR has so many problems, i don't even know where to start...
- why does it define a macro for which we have a include file? (it's even mentioned in the thread linked in the comments!)
- it first uses a kernal call to clear the screen (in a way that is technically broken), then copies the text to screen memory, in a non portable way - why not use a kernal call for that as well?
- why does it use a custom linker config?
- why a shell script (that doesnt work on windows) instead of a makefile? and why is none of it hooked into the buildsystem?

That said, the samples directory IMHO is for examples that show how to use the C compiler - and perhaps for some more advanced assembler things that are ca65 specific. This code however is neither a good example on how to write C64 code, nor how to use ca65.

netpipe:
cl65 -t c64 -C c64-asm.cfg  can be used instead i guess did not know it was there. i'll try and get a make file but it could take a while. as for the include macropack im not sure how to include cbm.h properly then. hello world in assembly is important because its the gateway to other benchmarks in pure assembly would be very practical to have a working demo.

netpipe:
[.MACPACK cbm](https://www.cc65.org/doc/ca65.html#toc13.3)

netpipe:
https://www.lemon64.com/forum/viewtopic.php?t=33082 this does it better i guess but its still not able to do both upper and lower case displayed.

Marrin:
@netpipe Just using the `scrcode` macro without any argument and at the top of the source code does nothing.  That macro is for inserting the given values as screen codes, instead of using `.byte`.

Untested:
```
.macpack cbm

start:
    lda #14
    jsr $ffd2

    ldy #0
loop:
    lda message,y
    beq done
    sta $0400,y
    iny
    bne loop
done:
    rts

message:
    scrcode "Hello, World!@"
```

netpipe:
works great, the upper/lowercase are reversed though.

netpipe:
https://github.com/cc65/cc65/pull/2405

Issue #2395, opened 2024-02-01
Bitwise operation with boolean operand wrong when optimised


colinleroy:
Hi,
Found in https://github.com/StewBC/cc65-Chess/issues/10, introduced by 8b6d78a075dafde8583c4992d37995d9b4c9db97,
the bitwise operations optimisation mess up when done with a boolean subexpression:

Reproducer: 
```
#include <stdio.h>
#ifdef __CC65__
#include <conio.h>
#endif

unsigned char a, b;
unsigned char c = 255;
unsigned char d = 100;
void main() {

    a = c ^ (d != 0);
    b = c ^ 1;

    if (a != b) {
      printf("error! a %d, b %d\n", a, b);
    }

#ifdef __CC65__
    cgetc();
#endif
}
```
Result is a=155, b=254. (gcc's result is a=b=254)

The generated code shows:
```
; a = c ^ (d != 0);
;
        lda     _d
        cmp     #$00
        ; there's something missing there :) 
        eor     _c
        sta     _a
;
```

mrdudz:
For a start, it would be great if you could add a test into the todo dir :)

@acqn please have a look :)

colinleroy:
@mrdudz done :)

acqn:
`Opt_a_tosxor`, as well as `Opt_a_tosor` and `Opt_a_tosand` are all missing the same opimization guard: `OP_RHS_REMOVE_DIRECT | OP_RHS_LOAD_DIRECT`.

PR #2397, opened 2024-02-01
Add test case for issue #2395

colinleroy:


mrdudz:
Mmmmh the forced failure is a bit unfortunate (this way whoever fixes it won't recognize the test must be changed/moved).

Perhaps you can make it a special case and move into misc?

colinleroy:
> Mmmmh the forced failure is a bit unfortunate (this way whoever fixes it won't recognize the test must be changed/moved).

How about that? It'll stop failing (and break the build) when the -O build will work.

mrdudz:
Thats better, i guess :) thanks!

PR #2401, opened 2024-02-02
Enable Windows tests on pull requests

colinleroy:
This should work :crossed_fingers: 
We're going to see the result right on this one.

So yeah indeed that's a bit longer, it takes 14 minutes. I'll let you be the judge of whether this is better than merging a PR, then having to go back to it because Windows scheduled tests failed!

### DIFF
--- a/.github/workflows/build-on-pull-request.yml
+++ b/.github/workflows/build-on-pull-request.yml
@@ -54,7 +54,7 @@ jobs:
           make -j2 bin USER_CFLAGS=-Werror CROSS_COMPILE=x86_64-w64-mingw32-
 
   build_windows:
-    name: Build (Windows)
+    name: Build and Test (Windows)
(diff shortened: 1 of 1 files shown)


mrdudz:
are you sure this actually uses cmd as sub shell, like the crontab check does? (see https://github.com/cc65/cc65/pull/2108 and https://github.com/cc65/cc65/pull/1937 ) ?

colinleroy:
> are you sure this actually uses cmd as sub shell, like the crontab check does? (see #2108 and #1937 ) ?
dd
I *think* so, I copied the bits from the crontab check. I wouldn't bet my life on it though, I'm Not Good ™ at Windows.

mrdudz:
I'll just have the crontab one running, so next time something fails, both tests should fail :)

PR #2402, opened 2024-02-02
a BIT of 65C02 optimisations

colinleroy:
Use BIT immediate instead of AND when reloading A would be required afterwards.
Add an fread unit test as the optimisation touches fread.
Sprinkle a few zero page indexed while we're at it.

### DIFF
--- a/libsrc/common/_printf.s
+++ b/libsrc/common/_printf.s
@@ -13,6 +13,7 @@
         .import         _strlower, _strlen
 
         .macpack        generic
+        .macpack        cpu
 
 ; ----------------------------------------------------------------------------
 ; We will store variables into the register bank in the zeropage. Define
@@ -37,7 +38,11 @@ FCount          = ptr2
 
 GetFormatChar:
         ldy     #0
+        .if (.cpu .bitand ::CPU_ISET_65SC02)
+        lda     (Format)
+        .else
         lda     (Format),y
+        .endif
 IncFormatPtr:
         inc     Format
         bne     @L1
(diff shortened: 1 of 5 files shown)


colinleroy:
(I hope I'm not tiring you with all these PRs :sweat_smile: )

libsrc/common/_printf.s
```diff
@@ -135,9 +144,9 @@ ReadInt:
         pha                             ; Save digit value
         lda     ptr1
         ldx     ptr1+1
-        asl     ptr1
+        asl     a
         rol     ptr1+1                  ; * 2
-        asl     ptr1
+        asl     a
```
mrdudz, line 149: sure this is the right thing to do? the original code will add the shifted value below - after the change its the non shifted value - or am i missing something? :)

colinleroy: I was quite sue you'd say something, because I had to think twice about this one, but yes:
```
        lda     ptr1
        ldx     ptr1+1
        asl     ptr1
        rol     ptr1+1                  ; * 2
        asl     ptr1
        rol     ptr1+1                  ; * 4, assume carry clear
        adc     ptr1
        sta     ptr1
```
Let's say you LDA with ptr1=2, so now A=2
ASL ptr1 ; ptr1=4, A=2
ASL ptr1 ; ptr1=8, A=2
adc ptr1 ; A = 8+2 = 10
sta ptr1 ; ptr1 = 10
Now instead you you LDA with ptr1=2, so now A=2
ASL A ; ptr1=2, A=4
ASL A ; ptr1=2, A=8
adc ptr1 ; A = 2+8 = 10
sta ptr1 ; ptr1 = 10

PR #2410, opened 2024-02-11
Improvements and fixes in serial support for Atari Lynx

alexthissen:
This improves the codebase for the serial library and its support for the Atari Lynx.

Functionality changes:
- On close the serial timer is stopped and serial errors are cleared
- Parity bit is now checked correctly for each of the even/odd/mark/space settings
- Baud rates are now calculated and set correctly
- Baud rates of 150 bit/s and below are regarded as invalid
- Using SER_HS_SW software handshake will check for connected redeye (comlynx cable) on ser_open

Improvements and other changes:
- Added bit definition values for IODAT
- Renamed bit definitions for SERCTL read and write to be same as Epyx harddefs.i defined names
- Changed hardcoded bit values for SERCTL to use new names
- Removed duplication for resetting interrupt flag and return value

### DIFF
--- a/libsrc/lynx/ser/lynx-comlynx.s
+++ b/libsrc/lynx/ser/lynx-comlynx.s
@@ -200,7 +200,18 @@ checkhs:
         ldy     #SER_PARAMS::HANDSHAKE  ; Handshake
         lda     (ptr1),y
         cmp     #SER_HS_NONE
+        beq     redeye_ok
+        cmp     #SER_HS_SW              ; Software handshake will check for connected redeye
         bne     invparameter
+
+        lda     IODAT
+        and     #NOEXP                  ; Check if redeye bit flag is unset
+        beq     redeye_ok
+        lda     #SER_ERR_NO_DEVICE      ; ComLynx cable is not inserted
+        ldx     #0
+        rts
+
+redeye_ok:
         lda     SERDAT
         lda     contrl
         ora     #RXINTEN|RESETERR       ; Turn on interrupts for receive
@@ -209,6 +220,7 @@ checkhs:
         .assert SER_ERR_OK = 0, error
         tax
         rts
+
 invparameter:
         lda     #SER_ERR_INIT_FAILED
         ldx     #0 ; return value is char
--- a/libsrc/lynx/ser/lynx-comlynx.s
+++ b/libsrc/lynx/ser/lynx-comlynx.s
@@ -76,8 +76,7 @@ SER_CLOSE:
         ; Disable interrupts and stop timer 4 (serial)
         lda     #TXOPEN|RESETERR
         sta     SERCTL
-        lda     #$00  ; Disable count and no reload
-        sta     TIM4CTLA
+        stz     TIM4CTLA  ; Disable count and no reload
 
         ; Done, return an error code
         lda     #SER_ERR_OK
--- a/libsrc/lynx/ser/lynx-comlynx.s
+++ b/libsrc/lynx/ser/lynx-comlynx.s
@@ -76,7 +76,8 @@ SER_CLOSE:
         ; Disable interrupts and stop timer 4 (serial)
         lda     #TXOPEN|RESETERR
         sta     SERCTL
-        stz     TIM4CTLA  ; Disable count and no reload
+        stz     TIM4CTLA    ; Disable count and no reload
+        stz     SerialStat  ; Reset status
 
         ; Done, return an error code
         lda     #SER_ERR_OK
@@ -241,8 +242,8 @@ GetByte:
         ldy     RxPtrOut
         lda     RxBuffer,y
         inc     RxPtrOut
+        sta     (ptr1)
         ldx     #$00
-        sta     (ptr1,x)
         txa                     ; Return code = 0
         rts
 
@@ -288,8 +289,8 @@ PutByte:
 
 SER_STATUS:
         lda     SerialStat
+        sta     (ptr1)
         ldx     #$00
-        sta     (ptr1,x)
         txa                     ; Return code = 0
         rts
 
@@ -327,7 +328,7 @@ SER_IRQ:
         and     #PAREN      ; Parity enabled implies SER_PAR_EVEN or SER_PAR_ODD
         tay
         ora     #OVERRUN|FRAMERR|RXBRK
-        bit     SERCTL      ; Check presence of relevant error flags in SERCTL
+        and     SERCTL      ; Check presence of relevant error flags in SERCTL
 
         beq     @rx_irq     ; No errors so far
 
--- a/libsrc/lynx/ser/lynx-comlynx.s
+++ b/libsrc/lynx/ser/lynx-comlynx.s
@@ -158,33 +158,6 @@ SER_OPEN:
         cmp     #SER_BAUD_300
         beq     setbaudrate
 
-        ldx     #103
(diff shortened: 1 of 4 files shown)


mrdudz:
So who can check this? :)

alexthissen:
Hoping that @karrika or @42Bastian can check this, as these are two active Lynx developers that wrote the original libraries. Will ask them to review.

libsrc/lynx/ser/lynx-comlynx.s
```diff
@@ -73,7 +73,12 @@ SER_UNINSTALL:
 ; Must return an SER_ERR_xx code in a/x.
 
 SER_CLOSE:
-        ; Disable interrupts
+        ; Disable interrupts and stop timer 4 (serial)
+        lda     #TXOPEN|RESETERR
+        sta     SERCTL
+        lda     #$00  ; Disable count and no reload
```
42Bastian, line 79: Rather use stz here

libsrc/lynx/ser/lynx-comlynx.s
```diff
@@ -108,17 +113,17 @@ SER_OPEN:
...
         stz     TxPtrOut
 
-        ; clock = 8 * 15625
-        lda     #%00011000
-        sta     TIM4CTLA
         ldy     #SER_PARAMS::BAUDRATE
         lda     (ptr1),y
 
+        ; Source period is 1 us
+        ldy     #%00011000  ; ENABLE_RELOAD|ENABLE_COUNT|AUD_1
```
42Bastian, line 120: Why use here the direct value and not the symols?

libsrc/lynx/ser/lynx-comlynx.s
```diff
@@ -218,31 +184,43 @@ baudsuccess:
...
+
+redeye_ok:
         lda     SERDAT
         lda     contrl
-        ora     #RxIntEnable|ResetErr
+        ora     #RXINTEN|RESETERR       ; Turn on interrupts for receive
         sta     SERCTL
         lda     #SER_ERR_OK
         .assert SER_ERR_OK = 0, error
         tax
```
42Bastian, line 221: move tax up directly after redeye_ok, no assert needed

libsrc/lynx/ser/lynx-comlynx.s
```diff
@@ -308,7 +288,7 @@ PutByte:
 ; Must return an SER_ERR_xx code in a/x.
 
 SER_STATUS:
-        ldy     SerialStat
+        lda     SerialStat
         ldx     #$00
         sta     (ptr1,x)
```
42Bastian, line 293: This opcode seems to be weird, it works, but rather use sta (ptr1)

42Bastian (commented):
So far it looks ok, not sure if the line comments are visible.

libsrc/lynx/ser/lynx-comlynx.s
alexthissen: I will add and change more symbols in another pull request. Didn't want to introduce these yet, and instead do that when working on the timers. The comment is for later when the symbols are introduced.

libsrc/lynx/ser/lynx-comlynx.s
alexthissen: The return code for the ser_open is a char, so the accumulator needs to contain the error code (SER_ERR_OK when there is no error), and x should be 0. At this point, a ldx #$00 would be okay, but since SER_ERR_OK is asserted to be zero (first number in enum values of SER_ERR_*) it saves a byte by doing tax. 
This code is similar to other serial driver implementations.

libsrc/lynx/ser/lynx-comlynx.s
alexthissen: I assume that this is done to express that we are setting member zero of structure ptr (which is an address). It seems to be for clarity rather than brevity.
Agree to leave?

libsrc/lynx/ser/lynx-comlynx.s
alexthissen: Changed it.

libsrc/lynx/ser/lynx-comlynx.s
42Bastian: sta (nn,x) != sta (nn),y !!!

First: Use address at nn+x.
So if nn == 10 and x and y are 2, first will use the address at 12 to store to, and second will store to address in 12.

So, no, better remove.

libsrc/lynx/ser/lynx-comlynx.s
42Bastian: yes, but A is zero before "lda SERDAT", so it makes sense to move it there. and get rid of the assert.

libsrc/lynx/ser/lynx-comlynx.s
mrdudz: sta (ptr1) also saves two bytes, no? :)

alexthissen: Here is the prototype for ser_status:
```
unsigned char __fastcall__ ser_status (unsigned char* status);
```
As far as I knew the calling convention assumes that the return value is in A and X should be 0. 
ptr1 that corresponds to the argument list. The first argument in the list is `status` (of type `unsigned char*`). By using Absolute Indexed Indirect addressing `sta (ptr1,x)` we essentialy save to the same location as sta (ptr1) does. 
No problem for me to change it into sta(ptr1). I left the original code there as all implementations of ser_status use it like this. If it saves two bytes, why not? ;)

libsrc/lynx/ser/lynx-comlynx.s
alexthissen: Okay to resolve?

libsrc/lynx/ser/lynx-comlynx.s
42Bastian: All three version use 2 bytes only. But "sta (ptr1,x)" is simply wrong here.
And if you gonna change code, you should remove such.

libsrc/lynx/ser/lynx-comlynx.s
42Bastian: well, I see no sense in polluting code with asserts which are not needed. No dead code if you can avoid it.
But just my 2cts.

alexthissen: Agree. 
In this case I would follow @mrdudz recommendation of the contribution guidelines and leave it.

42Bastian: but why not
```
.if SER_ERR_OK != 0
  ldx #0
 .endif
```

libsrc/lynx/ser/lynx-comlynx.s
mrdudz: 6502 has no sta(zp) - so i guess that is why other drivers use it. its a common pattern when X is 0 and Y can't be used for whatever reason.

42Bastian: :-)
I see.

Luckily the Lynx is an 65C02.

alexthissen: Made the change to sta (ptr1) in two places. The routine set_get also used the same construct.

mrdudz:
ok then :) thanks!

PR #2414, opened 2024-02-14
vic-asm.cfg: Fix size of MAIN to end at $1E00.

SvenMichaelKlose:
Caused negative size of MAIN in cc65-contrib/quikmans2k8.


test/ref/strtok.c
```diff
@@ -0,0 +1,43 @@
+// 2024-02-14 Sven Michael Klose <pixel@hugbox.org>
```
mrdudz, line 1: While extra tests are always appreciated - please put this one into a separate PR

cfg/vic20-asm.cfg
```diff
@@ -7,7 +7,7 @@ SYMBOLS {
 MEMORY {
     ZP:       file = "", start = $0002,  size = $001A, define = yes;
     LOADADDR: file = %O, start = %S - 2, size = $0002;
-    MAIN:     file = %O, start = %S,     size = $0DF3 - %S;
+    MAIN:     file = %O, start = %S,     size = $1E00 - %S;
```
mrdudz, line 10: It would be great if you could add some comment to this file that
- tell what vic20 memory config it is meant to be used with (it should be "unexpanded")
- how to modify this for other configs (+8k, full)

mrdudz:
So what about that strtok test?

mrdudz:
*ahem* why close this? :)

mrdudz:
Thanks!

PR #2416, opened 2024-02-19
Allow choosing 115200bps as the card allows it

colinleroy:
Of course, that won't work full speed with the standard IRQ-based RX. But that will allow users to setup the port at this speed without duplicating the setup part of the code. Up to them to add hooks to disable IRQs and read directly in a tight asm loop.

### DIFF
--- a/doc/apple2.sgml
+++ b/doc/apple2.sgml
@@ -452,9 +452,14 @@ The names in the parentheses denote the symbols to be used for static linking of
   (RTS/CTS) and does interrupt driven receives. Speeds faster than 9600 baud
   aren't reachable because the ROM and ProDOS IRQ handlers are too slow.
   Software flow control (XON/XOFF) is not supported.
+
   Note that because of the peculiarities of the 6551 chip transmits are not
   interrupt driven, and the transceiver blocks if the receiver asserts
   flow control because of a full buffer.
+
+  Note that using the driver at SER_BAUD_115200 will disable IRQs. It will be up
+  to the users to use the serial port, either by re-enabling IRQs themselves,
+  or by directly poll-reading the ACIA DATA register without the help of ser_get().
 
   The driver defaults to slot 2. Call <tt/ser_apple2_slot()/ prior to
   <tt/ser_open()/ in order to select a different slot. <tt/ser_apple2_slot()/
--- a/doc/apple2enh.sgml
+++ b/doc/apple2enh.sgml
@@ -453,9 +453,14 @@ The names in the parentheses denote the symbols to be used for static linking of
   (RTS/CTS) and does interrupt driven receives. Speeds faster than 9600 baud
   aren't reachable because the ROM and ProDOS IRQ handlers are too slow.
   Software flow control (XON/XOFF) is not supported.
+
   Note that because of the peculiarities of the 6551 chip transmits are not
   interrupt driven, and the transceiver blocks if the receiver asserts
   flow control because of a full buffer.
+
+  Note that using the driver at SER_BAUD_115200 will disable IRQs. It will be up
+  to the users to use the serial port, either by re-enabling IRQs themselves,
+  or by directly poll-reading the ACIA DATA register without the help of ser_get().
 
   The driver defaults to slot 2. Call <tt/ser_apple2_slot()/ prior to
   <tt/ser_open()/ in order to select a different slot. <tt/ser_apple2_slot()/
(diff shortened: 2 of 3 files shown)


oliverschmidt:
I see the point. I think it is desirable to have it just a bit less of a hack. My thoughts:

1. Do I know at all where the ACIA is? Yes, the driver doesn't scan slots. It's documented to work with slot 2 or the chosen slot :-)
2. What happens if data flows in after ser_open() was called but before ACIA receive IRQ is disabled. I'd guess the machine seems to hang as the implicit CLI of RTI makes IRQ asserted again right away. This seems pretty undesirable :-(

So 2.) makes me believe that there should be a way to not activate receive IRQ on the ACIA in the first place. I guess the simplest approach is to leverage the fact that the baud rate in question happens to have the value 0. So I'd "just" hardcode that with baud rate 0, there's no receive IRQ activated. Still pretty much of a hack, bit a little less than your proposal.

What do you think?

colinleroy:
Ah, right, you raise a good point.

But, I fear if we disable IRQs, we'd be sort-of expected to extend the driver - and driver interface - to provide a non-IRQ read method? and/or a way to toggle IRQs? Which would mean a bigger driver for everybody.

After all, the documentation specifically mentions "Speeds faster than 9600 baud aren't reachable because the ROM and ProDOS IRQ handlers are too slow." ; and SER_BAUD_19200 is already supported in this way: you can use it, but beware how you do or you'll lose bytes!

In my use-case, I do the contrary : ser_open() with IRQ enabled, so that I get the comfort of not needing to race to get bytes in the general usage - the other end's careful to time sending slow enough; and when needed, I disable IRQs just before big transfers, read, and re-enable it. This, of course, requires me to implement ser_enable_irq() and ser_read_direct().

oliverschmidt:
> But, I fear if we disable IRQs, we'd be sort-of expected to extend the driver - and driver interface - to provide a non-IRQ read method? and/or a way to toggle IRQs?

I sort of see your point.

> After all, the documentation specifically mentions "Speeds faster than 9600 baud aren't reachable because the ROM and ProDOS IRQ handlers are too slow." ; and SER_BAUD_19200 is already supported in this way: you can use it, but beware how you do or you'll lose bytes!

You're conceptually right, but 19200 becomes easily reachable with some acceleration. So it "makes sense" to offer that. In contrast, I don't see an Apple II work with firmware->ProDOS->cc65 framework->serial handler IRQs (especially with the //c & enhanced //e firmware) at 115200.

> In my use-case, I do the contrary : ser_open() with IRQ enabled, so that I get the comfort of not needing to race to get bytes in the general usage - the other end's careful to time sending slow enough; and when needed, I disable IRQs just before big transfers, read, and re-enable it. This, of course, requires me to implement ser_enable_irq() and ser_read_direct().

That seems like a pretty special use case...

The other option I saw when I proposed the "value-0-hardcoding" was some cheap way to influence the setting from the outside. Like setting the MSB of the baudrate or the handshake in ser_params.

You see, I'm willing to have something built in (be it documented or not) that doesn't make any sense without implying the user to bypass the whole ser_get() / ser_put(). But I think it needs to allow to start out with receive IRQs disabled.

colinleroy:
Yes, that makes sense. I've pushed a commit to implement that :)

oliverschmidt:
> Thanks! The "that" in "implement that" refers to the "value-0-hardcoding", not to setting the MSB of the baudrate or the handshake in ser_params, right?

colinleroy:
> Thanks! The "that" in "implement that" refers to the "value-0-hardcoding", not to setting the MSB of the baudrate or the handshake in ser_params, right?

Right :)

doc/apple2.sgml
```diff
@@ -452,10 +452,15 @@ The names in the parentheses denote the symbols to be used for static linking of
   (RTS/CTS) and does interrupt driven receives. Speeds faster than 9600 baud
   aren't reachable because the ROM and ProDOS IRQ handlers are too slow.
   Software flow control (XON/XOFF) is not supported.
+
   Note that because of the peculiarities of the 6551 chip transmits are not
   interrupt driven, and the transceiver blocks if the receiver asserts
   flow control because of a full buffer.
 
+  Note that using the driver at SER_BAUD_115200 will disable IRQs. It will be up
+  to the user to use the serial port, either by re-enabling IRQs themselves,
```
oliverschmidt, line 461: I'm no native speaker but shouldn't it either be "users" ... "themselves" or "user" ... "himself" ?

colinleroy: I'm no native speaker either, but I was aiming for the non-gendered singular they (https://en.wikipedia.org/wiki/Singular_they).
I switched to user**s** so that other non-native speakers don't think we can't english properly ;)

Compyx: Not a native English speaker here either, but I think the proper way would be "user ... themself"

oliverschmidt: Thanks. At least I would have understood this is to be related to that thing.

PR #2418, opened 2024-02-19
helloworld ASM CBM Demo

netpipe:
Hello World for C64 CBM Assembly with makefile

### DIFF
--- a/samples/cbm/Makefile
+++ b/samples/cbm/Makefile
@@ -135,6 +135,8 @@ plasma: plasma.c
 	$(CL) -t $(SYS) -O -o plasma -m plasma.map plasma.c
 nachtm: nachtm.c
 	$(CL) -t $(SYS) -O -o nachtm -m nachtm.map nachtm.c
+hello: hello.s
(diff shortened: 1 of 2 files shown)


mrdudz:
*sigh* This is now the 3rd incarnation of PR #2391 - please stop making new PRs for the same thing. If you need to change/update a PR, commit into that very branch.

That said, while some issues of the original PR have been fixed, some others still stand:

* it first uses a kernal call to clear the screen (in a way that is technically broken), this needs to be fixed
* then copies the text to screen memory, in a non portable (c64 specific) way - why not use a kernal call for that as well?
* it right now lives in the cbm directory, but it is c64 specific
* its not added to the samples disk image

And last not least, as said before, the samples directory IMHO is for examples that show how to use the C compiler - and perhaps for some more advanced assembler things that are ca65 specific. I am not convinced that this kind of program even belongs there at all.

@oliverschmidt @groessler et al - what do you think?

oliverschmidt:
I have no desire to have this program be part of cc65.

groessler:
I also don't see the need for it

mrdudz:
Agreed, case closed

PR #2422, opened 2024-02-29
[cc65] Fixed potential errors with subtraction evaluation of identifiers at different memory locations

acqn:
Just in case.


mrdudz:
Thanks!

PR #2434, opened 2024-03-18
Implement strcasestr

colinleroy:
This makes strstr gain 40 bytes (82 to 122), but I think the functionality is worth it?


colinleroy:
PS if you'd rather I implement it in a different file, in order not to touch strstr.s, I can, strcasestr should be about 100-110 bytes.

mrdudz:
> PS if you'd rather I implement it in a different file, in order not to touch strstr.s, I can, strcasestr should be about 100-110 bytes.

I wonder if we can have both... ie use some linker magic to pull in the combined one when both are used, else the separate versions. mmmh

colinleroy:
> > PS if you'd rather I implement it in a different file, in order not to touch strstr.s, I can, strcasestr should be about 100-110 bytes.
> 
> I wonder if we can have both... ie use some linker magic to pull in the combined one when both are used, else the separate versions. mmmh

I don't think we can do that :/

colinleroy:
What I can do, though, is factorize (at the expense of cycles) and shave twenty bytes - cf last commit.

colinleroy:
What do you think, @mrdudz ? :)

colinleroy:
Hi! Any news on this one?
Thanks 🙂

mrdudz:
Sorry, i am a bit short of time right now...

I don't know what to do with this... generally size is more important than speed, but in this case it doesn't really make the decision easier :)

IMHO both strstr and strcasestr are so rarely used, that it doesn't justify a size increase when only one of the two is used.  Also strcasestr is not a standard function. So overall, perhaps its a better idea to have this one standalone in a separate file.

colinleroy:
Hi @mrdudz ,
Yes, that seems OK. Pushed that and also I had forgotten to update the doc.

colinleroy:
Ping @mrdudz :)

mrdudz:
Thanks, merging!

PR #2436, opened 2024-03-26
Alternative Atmos args

polluks:
Extended BASIC support

### DIFF
--- a/doc/atmos.sgml
+++ b/doc/atmos.sgml
@@ -235,6 +235,7 @@ supported directly by BASIC, the following syntax was chosen:
 
 <tscreen><verb>
     RUN:REM arg1 " ARG2 IS QUOTED" ARG3 "" ARG5
+    RUN'arg1
 </verb></tscreen>
 
 <enum>
--- a/libsrc/atmos/mainargs.s
+++ b/libsrc/atmos/mainargs.s
@@ -40,15 +40,17 @@ L1:     sta     name,y
 
         ldx     #0
 L2:     lda     BASIC_BUF,x
-        beq     done            ; No "rem", no args.
+        beq     done            ; No remark, no args.
         inx
         cmp     #REM
+        beq     L4
+        cmp     #'''
         bne     L2
 
 ; The arguments must be copied to a safe place because BASIC's input buffer
 ; might be re-used by the stdin console.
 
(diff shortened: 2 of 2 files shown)


mrdudz:
I am not convinced we want this kind of patch - what's wrong with the REM approach? It's not like people produce a lot of "commandline" programs with cc65 afterall - which may justify increasing code size to save a little bit of typing.

polluks:
The idea came from the Defence Force forum. You only have a copy cursor (not full screen like CBM) and typing `RUN'` multiple times is quite handy, you don't even need a colon.
Is this feature not worth four bytes?

mrdudz:
> typing RUN' multiple times is quite handy, you don't even need a colon.

But does the architecture even allow to run a (non trivial) program multiple times? (On most cc65 targets that is not the case)

> Is this feature not worth four bytes?

As i wrote before - is this something people are doing all the time? then sure, yes. I highly doubt that is the case though. 

I have yet to see one non trivial program, for any cc65 target, that uses the "commandline" arguments for that matter.

mrdudz:
BTW wouldn't it make sense to change it so support *only* the second (shorter) form? That would be a perfectly acceptable patch to me.

polluks:
Hmm, that would break compatibility. The suggestion came from the Defence Force forum.

mrdudz:
> that would break compatibility of course.

how? compatibility with what?

polluks:
Existing Atoms projects. Ok, I suppose there are not that many cc65 projects using args so far.

mrdudz:
So can this be closed or will you fix it?

polluks:
> I have yet to see one non trivial program, for any cc65 target, that uses the "commandline" arguments for that matter.

Let me see this weekend...

mrdudz:
Ar you planning to actually do this?

mrdudz:
@polluks hello? :)

mrdudz:
I am closing this due to no feedback/apparent interest.

I'm still willing to merge a PR that **replaces** the REM approach by the shorter `RUN'` approach. Having both makes no sense however.

polluks:
FYI https://retrocomputing.stackexchange.com/questions/30548/which-basic-dialect-first-featured-a-single-character-comment-introducer/

PR #2437, opened 2024-03-28
provide simple examples for using sim65 with C and assembly code

bbbradsmith:
Adding two minimal examples to the documentation for sim65 to explain how to write, build and run a working sim65 program in both C and assembly.

### DIFF
--- a/doc/sim65.sgml
+++ b/doc/sim65.sgml
@@ -124,11 +124,28 @@ Exit codes are limited to 8 bits.
 
 The standard C library high level file input and output is functional.
 A sim65 application can be written like a command line application,
-providing arguments to <tt/main/ and using the <tt/stdio.h/ interfaces.
+providing command line arguments to <tt/main/ and using the <tt/stdio.h/ interfaces
+to interact with the console or access files.
 
 Internally, file input and output is provided at a lower level by
 a set of built-in paravirtualization functions (<ref id="paravirt-internal" name="see below">).
 
+Example:
+
+<tscreen><verb>
+#include <stdio.h>
+int main()
+{
+    printf("Hello!\n");
+    return 5;
+}
+
+// Build and run:
+//   cc65 -o example.s example.c
+//   ca65 -o example.o example.s
+//   ld65 -t sim6502 -o example.prg example.o sim6502.lib
+//   sim65 example.prg
+</verb></tscreen>
 
 <sect>Creating a Test in Assembly<p>
 
@@ -141,11 +158,25 @@ and the sim65 library provides two ways to return an 8-bit exit code:
 
 <item>Return from <tt/_main/ with the exit code in <tt/A/.
 
-<item><tt/jmp exit/ with the code in <tt/A/.
+<item><tt/jmp exit/ with the code in <tt/A/. (<tt/.import exit/ from the sim65 library.)
 
 </itemize>
 
-The binary file has a 12 byte header:
+Example:
+
+<tscreen><verb>
+.export _main
+_main:
+    lda #5
+    rts
+
+; Build and run:
+;   ca65 -o example.o example.s
+;   ld65 -t sim6502 -o example.prg example.o sim6502.lib
+;   sim65 example.prg
+</verb></tscreen>
+
+Internally, the binary program file has a 12 byte header provided by the library:
 
 <itemize>
 
@@ -181,6 +212,9 @@ These use cc65 calling conventions, and are intended for use with the sim65 targ
 
 <item><tt/IRQ/ and <tt/NMI/ events will not be generated, though <tt/BRK/
 can be used if the IRQ vector at <tt/$FFFE/ is manually prepared by the test code.
+
+<item>The sim6502 or sim65c02 targets provide a default configuration,
+but if customization is needed, <tt/sim6502.cfg/ or <tt/sim65c02.cfg/ might be used as a template.
 
 </itemize>
 


mrdudz:
Looks ok - except that the example commandlines should either use cl65, or provide -t sim6502 to all tools - else you'll get a mixup of the default (c64) and sim6502 things (eg character conversion)

doc/sim65.sgml
```diff
@@ -115,37 +115,70 @@ PVExit ($01)
...
+<tscreen><verb>
+#include <stdio.h>
+int main()
+{
+    printf("Hello!\n");
+    return 5;
+}
+
+// Build and run:
+//   cc65 -o example.s example.c
```
mrdudz, line 144: -t sim6502 should be provided to cc65 and ca65 too

doc/sim65.sgml
```diff
@@ -115,37 +115,70 @@ PVExit ($01)
...
+Example:
+
+<tscreen><verb>
+.export _main
+_main:
+    lda #5
+    rts
+
+; Build and run:
+;   ca65 -o example.o example.s
```
mrdudz, line 176: -t sim6502 should be provided to cc65 and ca65 too

bbbradsmith:
> Looks ok - except that the example commandlines should either use cl65, or provide -t sim6502 to all tools - else you'll get a mixup of the default (c64) and sim6502 things (eg character conversion)

I can add -t sim6502 to all commands if that's appropriate, but I believe the c64 default _only_ applies to cl65, actually. cc65 and ca65 do not default to c64 character conversion.

bbbradsmith:
Okay, I've added -t 6502 to the cc65 and ca65 example commands.

mrdudz:
> cc65 and ca65 do not default to c64 character conversion. (This has long been something I've found confusing about cl65.)

Hu? Of course they do. What else would do the character conversion? The linker certainly does not :) If you use -t only for cl65, you will get CBM character translation, combined with the startup- and runtime code for sim65

bbbradsmith:
> > cc65 and ca65 do not default to c64 character conversion. (This has long been something I've found confusing about cl65.)
> 
> Hu? Of course they do. What else would do the character conversion? The linker certainly does not :) (i believe ca65 has no use for the -t arg though) If you use -t only for ld65, you will get CBM character translation, combined with the startup- and runtime code for sim65.

No, they do not.

cc65 and ca65 have no target by default. Using -t will add a #define like `__SIM6502__` and set the character translation table, though I think the translation table is only used when creating an object, so maybe doesn't do anything directly in cc65. (See: SetSys). The translation table (Tab) is initialized by default to a direct copy of the input (i.e. ASCII).

ld65 also has no target by default. Using -t will select a CFG, and set a default binary output format. Without -t the linker needs at least -C to give it a config file. The default binary output format is BINFMT_BINARY, i.e. a direct copy. (See: OptTarget)

cl65 is the only one that behaves differently. In main it overrides the static default `Target = TGT_NONE` with `Target = TGT_C64`. No other utility in the cc65 suite does this. In addition to all the things -t does for cc/ca/ld65 it also adds the target library to the link step, which ld65 alone does not do.

Here is an example that demonstrates all of these things: [default_target_test.zip](https://github.com/cc65/cc65/files/15002004/default_target_test.zip)

```
ca65 -o test.o test.s
ld65 -o ca-ld.bin -t sim6502 test.o sim6502.lib

REM this isn't able to produce a binary
cl65 -o cl-s-default.bin test.s

cl65 -o cl-s.bin -t sim6502 test.s


cc65 -o test.c.s test.c
ca65 -o test.c.o test.c.s
ld65 -o cc-ca-ld.bin -t sim6502 test.c.o sim6502.lib

REM note this deletes the existing test.s
cl65 -o cl-c-default.bin test.c

cl65 -o cl-c.bin -t sim6502 test.c
```

> In any case, why split those commands and not use cl65 instead though?

I find cl65 confusing in a lot of ways and avoid using it. The c64 default target is just one of them. It also will silently create an .s file and delete it, even if a conflicting file already exists (see note in my example). It makes it harder to generate debug and diagnostic information as well. The separate steps allow better understanding of the process and greater control over the results, so I more or less just never have any reason to use it, unless it's someone else's project. It's not something I think makes a useful example, unless for a small c64 program maybe.

The reason I left out the -t in the examples I gave is that I was trying to reduce it to the most minimal command line. In both cc65 and ca65 the default target is already what sim6502 needs, so there's not really any function for -t sim6502 to perform. I added it at your request, but this use is merely a semantic aid, and doesn't have any current function.

mrdudz:
You got me there with -t, you are right. It is confusing indeed. Guess that's why i made it a habit to just use -t :)

That said, i think using cl65 in this particular case is still the way to go - as using the separate tools contributes nothing to the given case (how to make a test propram). And calling the separate tools is supposed to be something that you shouldn't have to do for a simple program like this :)

bbbradsmith:
Hmm. Well, if I was writing a test program I would do it in separate steps. Inspecting the intermediate assembly output is often part of of a test, as is custom linker configurations, etc. In a case like this it would help to know the minimal steps needed for that. For someone who is learning, I think it is much better to show all the steps.

To me, the purpose of cl65 is to make a quick build convenient, not to make it easier to understand. Adding an example to the cc65 unit tests, sure, it's good for that, but understanding their build process is not for the faint of heart.

I've checked in a compromise that gives both alternatives.

mrdudz:
I will merge it like this. Because my opinion on that topic is not strong enough - and i have to test if this damned access token works :=)

mrdudz:
(yay, it worked)

bbbradsmith:
woot!

PR #2450, opened 2024-06-16
Add asm config for C16 and Plus/4

ops:
Added asm configs for C16 and Plus/4. Additionally, build asm example in samples/cbm.


mrdudz:
thanks!

PR #2452, opened 2024-06-18
task: wrote script to regex functions and docs from header files

bobbyjim:
I commented on the X16 Discord channel that the documentation lags implementation a bit, and they said "we always just look at the headers; that's what they're there for after all."

I thought this sounded just slightly cynical, so I looked at the headers and found that they were very nicely standardized in how they defined functions and documented them.

So I wrote a short Perl script to try to create documentation straight from the headers, and it looks like it's a good first stab.  I think there are edge cases to worry about, but for now this isn't bad.


mrdudz:
Uhm, judging from that hdr.html it seems a LOT more incomplete than the docs.

I'd rather see a script that checks the docs for missing stuff (and perhaps the libs for missing implementations too), so the docs (and libs) can be completed easier.

bobbyjim:
That's an excellent suggestion, thank you.  I'll see what I can do, and I think I'll close this MR.

PR #2462, opened 2024-07-06
Keep gcc-14 from aborting with errors due to new defaults.

SvenMichaelKlose:
Adds -Wno-error=implicit-int -Wno-error=int-conversion to CFLAGS.

Tested with gcc-12.4 and gcc-14.1.


mrdudz:
Thanks!

PR #2463, opened 2024-07-06
Test strtok().

SvenMichaelKlose:


mrdudz:
Any reason for why this test was added to test/ref/ instead of test/val/ ?

mrdudz:
thanks!

PR #2464, opened 2024-07-07
Add stpcpy().

SvenMichaelKlose:
Like strcpy() but returning pointer to ending zero of copied string.

### DIFF
--- a/include/string.h
+++ b/include/string.h
@@ -52,7 +52,6 @@ char* __fastcall__ strchr (const char* s, int c);
 int __fastcall__ strcmp (const char* s1, const char* s2);
 int __fastcall__ strcoll (const char* s1, const char* s2);
 char* __fastcall__ strcpy (char* dest, const char* src);
-char* __fastcall__ stpcpy (char* dest, const char* src);
 size_t __fastcall__ strcspn (const char* s1, const char* s2);
 char* __fastcall__ strerror (int errcode);
 size_t __fastcall__ strlen (const char* s);
@@ -91,6 +90,7 @@ char* __fastcall__ strlower (char* s);
 char* __fastcall__ strupr (char* s);
 char* __fastcall__ strupper (char* s);
 char* __fastcall__ strqtok (char* s1, const char* s2);
+char* __fastcall__ stpcpy (char* dest, const char* src);
 #endif
 
 const char* __fastcall__ __stroserror (unsigned char errcode);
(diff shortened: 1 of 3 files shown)


include/string.h
```diff
@@ -52,6 +52,7 @@ char* __fastcall__ strchr (const char* s, int c);
 int __fastcall__ strcmp (const char* s1, const char* s2);
 int __fastcall__ strcoll (const char* s1, const char* s2);
 char* __fastcall__ strcpy (char* dest, const char* src);
+char* __fastcall__ stpcpy (char* dest, const char* src);
```
mrdudz, line 55: The prototype should be moved below to the non standard ones

mrdudz:
Also please add a test for stpcpy

mrdudz:
Oh and... why not write this function in assembly? Should be easy enough - and almost certainly saves a couple bytes :)

include/string.h
SvenMichaelKlose: Oops! True. Sorted.

SvenMichaelKlose:
Test added but I kindly refuse to re-implement this in assembly. That's a job for the optimizer department. If the optimizer cannot handle stpcpy(), improving it instead must result in a huge overall improvement of the compiler.

mrdudz:
> Test added but I kindly refuse to re-implement this in assembly. That's a job for the optimizer department. If the optimizer cannot handle stpcpy(), improving it instead must result in a huge overall improvement of the compiler's output.

Go ahead :)

SvenMichaelKlose:
Sorry. Having too much on my plate already to dig into that. Am sure there are experts who could do that in a jiff. Would take me ages.

mrdudz:
> Sorry. Having too much on my plate already to dig into that. Am sure there are experts who could do that in a jiff. Would take me ages.

The current number of active developers is zero - which is exactly why writing library functions in assembler makes a lot of sense.

SvenMichaelKlose:
Oops... right... what a pity. Since no-one was missing stpcpy() really, please just have this pulled in and I'll take a look at the optimizer by next week end.

libsrc/common/stpcpy.c
```diff
@@ -0,0 +1,8 @@
+#include <string.h>
+
+char * __fastcall__
+stpcpy (char * dst, const char * src)
+{
```
mrdudz, line 5: this should adhere to the codestyle of the rest of the library, ie use
`char* __fastcall__ stpcpy (char* dest, const char* src)`

SvenMichaelKlose:
I've heard they've invented computers to do such things (e.g. GNU indent). :p There you go.

mrdudz:
> I've heard they've invented computers to do such things (e.g. GNU indent). :p There you go.

I haven't seen any such tool yet that would not royally mess up existing formatting - else i would have integrated it long ago. Feel free to provide an indent rule that matches the existing cc65 codestyle for that matter.

merged, thanks!

PR #2468, opened 2024-07-20
Fixed #2459

### LINKED ISSUE #2459: [atari7800] joystick driver is not safe for the console hardware
When I wrote this driver originally I did not understand that adding a single button joystick and a dual button joystick at the same time may overheat a transistor and damage the console in the long run. The heating happens only when you keep the single button joystick pressed.

There is code for handling this case. But it turns both joysticks to single button joysticks if either stick is detected to be a single button joystick. The way it does this is to set the RIOT output pin from output to input. It may be enough to turn off the transistors. But I would rather drive to pins to state "1".

In any case I would like to update the driver to be safe for any mix of single or double joysticks.
7800joysticks

The problematic case is you have a zero on both transistors. In this case 5V will go directly through both transistors via the single button joystick to the ground. There is no current limiting resistor so the transistors may break.

karrika:
This fix allows operating single and double button joysticks at the same time (in different ports, of course). I will still test the operations with measuring the logic on the motherboard. But this is open for comments.

### DIFF
--- a/libsrc/atari7800/joy/atari7800-stdjoy.s
+++ b/libsrc/atari7800/joy/atari7800-stdjoy.s
@@ -53,13 +53,20 @@ JOY_COUNT       = 2             ; Number of joysticks we support
 ; Must return an JOY_ERR_xx code in a/x.
 ;
 
+PB2 = $04 ; Joystick 0
+PB4 = $10 ; Joystick 1
+
 INSTALL:
         ; Assume 7800 2-button controller, can change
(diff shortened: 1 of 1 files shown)


mrdudz:
If noone has objections, i will merge this.... :)

karrika:
I tried it on a real 7800 and it works afaik.

mrdudz:
Ok merge - thanks!

PR #2471, opened 2024-08-02
let clrscr() make use of _bzero() (-17 bytes + 20% speed)

IrgendwerA8:
Optimization for Atari-"clrscr()"

### DIFF
--- a/libsrc/atari/clrscr.s
+++ b/libsrc/atari/clrscr.s
@@ -1,36 +1,29 @@
 ;
-; Christian Groessler, Apr-2000
-;
 ; void clrscr (void);
 ;
 
         .export         _clrscr
         .include        "atari.inc"
-        .importzp       ptr1
+        .import         _bzero
+        .import         pushax
         .import         setcursor
 
 SCRSIZE =       960             ; 40x24: size of default atari screen
 
-_clrscr:lda     SAVMSC          ; screen memory
-        sta     ptr1
-        lda     SAVMSC+1
-        clc
-        adc     #>(SCRSIZE-1)
-        sta     ptr1+1
-        lda     #0              ; screen code of space char
-        sta     OLDCHR
-        ldy     #<(SCRSIZE-1)
-        ldx     #>(SCRSIZE-1)
-_clr1:  sta     (ptr1),y
-        dey
-        bne     _clr1
-        sta     (ptr1),y
-        dex
-        bmi     done
-        dec     ptr1+1
-        dey
-        jmp     _clr1
+_clrscr:
+        lda     SAVMSC          ; screen memory
+        ldx     SAVMSC+1
 
-done:   sta     COLCRS
+        jsr     pushax          ; pointer to memory to clear
+
+        lda     #<SCRSIZE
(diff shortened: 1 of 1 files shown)


groessler:
Hmm, I don't see the "-17bytes" point.

```
#include <conio.h>
#include <atari.h>

int main(void)
{
  clrscr();
  return(0);
}
```

Old cc65:

`-rw-r--r-- 1 chris chris   660 Aug  3 01:56 clrscrtest.com`

Your branch:

`-rw-r--r-- 1 chris chris   758 Aug  3 01:59 clrscrtest.com`

Is "+98" bytes....

polluks:
At least you gain space if you also use bzero, memset etc.

IrgendwerA8:
Yes, the "-17 bytes" refers to an already used "bzero"/"memset".
Somehow I thought "bzero()" is already in charge for clearing BSS.
If I "fix" that in "zerobss.s" too, the 758 bytes will go down to 733.

Personally the number of programs I created without using "bzero"/"memset" tend to zero - so for me the changes would result in a gain, but I could understand if the mileage of others vary.
Opinions / comments?

mrdudz:
I'm not familiar with the atari library... but how likely is it to write a non trivial program that would not pull in bzero?

groessler:
@mrdudz, I cannot really assess this. bzero/memset might be used, or might not be used. I think it's not directly related to the triviality of the program.

groessler:
I was a bit reluctant for this PR since the Atari lib is already quite large (runtime memory consuming). But with PR2484 I'm fine with the change. Please also change it for atari5200.

Marrin:
This depends on what you mean by `' '` and `'\0'`.  The video memory doesn't expect ASCII or ATASCII but an [internal encoding for the characters](https://www.atariarchives.org/mapping/appendix10.php).  Similar to Commodore machines with the difference between PETSCII and screen codes.

mrdudz:
I am leaning towards not merging this - is speed really an issue with clrscr() at all?

IrgendwerA8:
For me it's not - but AFAIR for @groessler it is.
Otherwise the code would just consist in an output of CHR$(125) (clear screen) and would also work with 80 column drivers etc.

Since size is always an issue for me, I thought reducing the double/triple implementations is a good idea. (To make 'big' programs with 'bzero' not unnecessary even bigger.)
But of course I understand different motivations, too.

groessler:
clrscr() in its current form is not there because of speed, but because it's part of conio. Mixing conio and rom functions is officially not supported.
You can always "putc" or "write" the 125 char on your program to clear the screen without linking in clrscr.

IrgendwerA8:
Thanks for pointing out!

groessler:
Interesting. I don't find it in the docs that conio and "regular" screen output shouldn't be mixed. But I remember it was for all targets, not just Atari.

mrdudz:
It also applies to C64 (and all other CBM machines)... yes some conio functions use the kernal... however, not all operations work the same and it will start to behave strange sooner or later.

Someone might want to add a note to the docs if its not there :)

Issue #2431, opened 2024-03-06
compiler warning on cc65/scanner.c


icepic:
As reported on the maillist, this occurs on recent MacOS with clang 15, but also on OpenBSD with clang 16.0.6:

BUILD_ID: Git 5c3ff714a
cc65/scanner.c
cc65/scanner.c:1270:27: warning: cast to smaller integer type 'token_t' (aka 'enum token_t') from 'void *' [-Wvoid-pointer-to-enum-cast]
        token_t LastTok = (token_t)CollLast (C);
                          ^~~~~~~~~~~~~~~~~~~~~
1 warning generated.

PR #2473, opened 2024-08-05
Fixed warning #2431

polluks:
cast splitted

### DIFF
--- a/src/cc65/scanner.c
+++ b/src/cc65/scanner.c
@@ -1267,7 +1267,8 @@ static int CloseBrace (Collection* C, token_t Tok)
 */
 {
     if (CollCount (C) > 0) {
-        token_t LastTok = (token_t)CollLast (C);
+        token_t* LastTok_ptr = (token_t*)CollLast (C);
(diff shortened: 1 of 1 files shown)


src/cc65/scanner.c
```diff
@@ -1267,7 +1267,8 @@ static int CloseBrace (Collection* C, token_t Tok)
 */
 {
     if (CollCount (C) > 0) {
-        token_t LastTok = (token_t)CollLast (C);
+        token_t* LastTok_ptr = (token_t*)CollLast (C);
+        token_t LastTok = *LastTok_ptr;
```
spiro-trikaliotis, line 1271: Have you checked that this code actually works? You don't just calmed down the warning, you changed the logic:

1. Old code: reinterprets the CollLast(c) as token_t LastTok
2. New code: dereferences (!) the CollLast(c), which is now considered to be a pointer to token_t

This is a very essential change. If the change is right, the old code was actually wrong. Otherwise, you risk accessing an invalid pointer.

polluks:
I suppose my fix is nonsense.

PR #2474, opened 2024-08-06
Added option to disable the force-to-uppercase behavior of the apple2 target.

oliverschmidt:
When a program built for the apple2 target is run on i.e. an Apple //e, it could actually display lowercase characters. Usually we don't do such runtime checks because we want a single program to behave identically on different machines. Otherwise the author would need to test a program on all possible machines in order to see how it behaves.

However, there's a really significant desire to have a single program that runs both on the ][+ and the //e - and does lowercase output on the //e. Therefore this PR adds a function allow_lowercase() that gives a program author the option to enable lowercase output. Two things are to be noted:

1. It's totally up to the program author to call allow_lowercase(1) only when he has made sure in one way or another that the current machine is actually capable of lowercase display. Otherwise the program will output garbage.

2. The function allow_lowercase() is only available for the apple2 target, not for the apple2enh target. If a source code is to be built both for apple2 and apple2enh, it's up to the program author to use the preprocessor or other means to not call allow_lowercase()  for apple2enh.

### DIFF
--- a/doc/apple2.sgml
+++ b/doc/apple2.sgml
@@ -330,6 +330,7 @@ usage.
 <item>_dos_type
 <item>_filetype
 <item>_datetime
+<item>allow_lowercase
 <item>beep
 <item>get_ostype
 <item>gmtime_dt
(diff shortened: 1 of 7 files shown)


oliverschmidt:
Nobody seems to oppose to this PR...

PR #2484, opened 2024-08-15
Use bzero() for clearing BSS.

IrgendwerA8:
I know this change may be a bit "controversy".
Let's gather some feedback and comments.

If its common sense, that "bzero/memset" is usually a part of a "real-life™"-application, this change would mean a size/speed gain (and my other "clrscr"-PR shouldn't be an issue at all anymore).

### DIFF
--- a/libsrc/common/zerobss.s
+++ b/libsrc/common/zerobss.s
@@ -1,43 +1,21 @@
 ;
-; Ullrich von Bassewitz, 1998-09-17, 2005-02-26.
-;
 ; Zero the bss segment.
 ;
 
         .export         zerobss
         .import         __BSS_RUN__, __BSS_SIZE__
-        .importzp       ptr1
-
+        .import         _bzero
+        .import         pushax
 
 .code
 
 zerobss:
         lda     #<__BSS_RUN__
-        sta     ptr1
-        lda     #>__BSS_RUN__
-        sta     ptr1+1
-        lda     #0
-        tay
+        ldx     #>__BSS_RUN__
 
-; Clear full pages
+        jsr     pushax
 
-L1:     ldx     #>__BSS_SIZE__
-        beq     L3
-L2:     sta     (ptr1),y
-        iny
-        bne     L2
-        inc     ptr1+1
-        dex
-        bne     L2
+        lda     #<__BSS_SIZE__
+        ldx     #>__BSS_SIZE__
 
-; Clear remaining page (y is zero on entry)
-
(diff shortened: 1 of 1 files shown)


mrdudz:
using bzero in crt0 seems a quite sane choice to me - how many extra bytes are we talking about (for something that would not pull in bzero by other means)?

IrgendwerA8:
To give you some numbers:
Current zerobss(): 35 bytes
PR zerobss(): 14 bytes

pushax: 22 bytes
memset: 93 bytes

So in worst case, the program will be 14-35+22+93 = 94 bytes longer after merging this PR.
I think everybody can agree that normally "pushax" is part of a program.
That would mean 72 bytes more than before.
(Or 21 bytes less, if bzero/memset is part of it.)

groessler:
I can live with that. One think to think about maybe is if zerobss might be used by assembler programs. Then using bzero for it is bad, since the C SP hasn't been setup. zerobss() can currently be called by asm programs which put uninitilialized vars into the BSS segment.

IrgendwerA8:
If that's really an issue (thanks for the valuable comment!) the latest commit doesn't use the c-SP anymore (by spending 5 additional bytes).

groessler:
That'd be also an issue with copydata -> memcpy. Maybe we can create zerobss_asm (the current function) and zerobss_c (the new function using memset/bzero). To let people using zerobss currently notice that something has changed.
Similar with copydata (in case you want to change that, too).

mrdudz:
Almost 100 bytes are much more than i'd be willing to sacrifice for sth like this

This is another great example for how useful it would be to be able to link different sets of lib functions depending on certain conditions/options...

That said, couldnt this alternative zerobss be provided in an "overload" file, like we do for the conio80 stuff?

IrgendwerA8:
Like I wrote, without accounting  "pushax" (I really regard this obligatory) it's 'only' 72 bytes.
Another approach (I even suggested once to Uz) is to offer two different libraries and a linker flag to choose speed vs. size.
A memory conservative 'memset()' (currently optimized for speed)  could be a lot smaller.

mrdudz:
After thinking it over again... i think we should leave zerobss alone. It's just a handful bytes being saved, and only if bzero is actually used in a program.

I am leaning towards not merging this. Comments?

IrgendwerA8:
Ok. See my other PR for my motivation.
Maybe someone else likes to create a small/'slow' "bzero" to merge at least "zerobss()" and "clrscr()" for the Atari...

PR #2485, opened 2024-08-18
Add Plus/4 wrapper for KERNAL's GETIN.

SvenMichaelKlose:


libsrc/plus4/kgetin.s
```diff
@@ -0,0 +1,14 @@
+.export GETIN
+
+.include "plus4.inc"
+
+KERNAL_GETIN := $FFE4
```
MonteCarlos, line 5: Already available here: `master/asminc/cbm_kernal.inc`.

SvenMichaelKlose: Yes, it is.

groessler: I guess because it should use the definition from cbm_kernal.inc.

mrdudz: obviously :)

mrdudz: please look at libsrc/c64/kplot.s as an example

PR #2486, opened 2024-08-22
Apple II: Move _exit out of STARTUP segment

colinleroy:
_exit() is in the STARTUP segment, because it's called after returning from callmain.

But I'll argue that this is counter-intuitive, and it prevents us to spare a few hundred bytes of memory in certain situations. 

For example, I want to use both HGR pages, my program's size is near the maximum, so instead of using `--start-addr 0x6000`, I want to use `--start-addr 0x5D00`, with the following segments:

```
ZEROPAGE              000080  000099  00001A  00001
EXEHDR                005CC6  005CFF  00003A  00001
STARTUP               005D00  005D0E  00000F  00001
RT_ONCE               005D0F  005F0E  000200  00001
LOWCODE               006000  007322  001323  00001
...
```

(This allows me to put my own init code into the RT_ONCE segment, inside the HGR2 page, as it's not needed anymore by the time HGR page2 is overwritten).

But if STARTUP contains _exit, this is not possible.

### DIFF
--- a/libsrc/apple2/callmain.s
+++ b/libsrc/apple2/callmain.s
@@ -1,0 +1,75 @@
+;
+; Ullrich von Bassewitz, 2003-03-07
+;
+; Push arguments and call main()
+;
+
+
+        .export         callmain, _exit
+        .export         __argc, __argv
+
+        .import         _main, pushax, done, donelib
+        .import         zpsave, rvsave, reset
+
+        .include        "zeropage.inc"
+        .include        "apple2.inc"
+
+
+;---------------------------------------------------------------------------
+; Setup the stack for main(), then jump to it
+
+callmain:
+        lda     __argc
+        ldx     __argc+1
+        jsr     pushax          ; Push argc
+
+        lda     __argv
+        ldx     __argv+1
+        jsr     pushax          ; Push argv
+
+        ldy     #4              ; Argument size
+        jsr     _main
+
+        ; Avoid a re-entrance of donelib. This is also the exit() entry.
+_exit:  ldx     #<exit
+        lda     #>exit
+        jsr     reset           ; Setup RESET vector
+
+        ; Switch in LC bank 2 for R/O in case it was switched out by a RESET.
+        bit     $C080
+
+        ; Call the module destructors.
+        jsr     donelib
+
+        ; Switch in ROM.
+        bit     $C082
+
+        ; Restore the original RESET vector.
+exit:   ldx     #$02
+:       lda     rvsave,x
+        sta     SOFTEV,x
+        dex
+        bpl     :-
+
+        ; Copy back the zero-page stuff.
+        ldx     #zpspace-1
+:       lda     zpsave,x
+        sta     sp,x
+        dex
+        bpl     :-
+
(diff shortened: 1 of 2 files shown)


oliverschmidt:
I must admit that I'm not convinced.
- The change adds 3 bytes for everybody in order to save bytes in a pretty specific scenario.
- The goal is only reached if exit() is explicitly called. Returning from main() instead is undefined behavior.

colinleroy:
I agree with the first point, it's an annoyance. (I tried to `.if * != __LOWCODE_START__`, but of course one can't do that, it's not constant).

For the second one, `_exec` also does `jmp _exit`, that's how I noticed it. As for returning from main(), it is maybe undefined, but it's how it works now: crt0 does `jsr callmain` which does `jmp _main`, so when main `rts` we come back to crt0 after `jsr callmain`)

Would you have ideas how to reach the goal of being able to trust that STARTUP is unneeded after startup, that would be smarter? (for now, I memcpy it in a temp buffer before writing over it, and restore it after I'm done with HGR, but that's ugly)

colinleroy:
It would be possible to 

- replace `jsr callmain` with `jmp callmain` in crt0.s, 
- replace `jmp _main` with `jsr _main` in callmain.s
- and stick the _exit code right after that in callmain.s

... But it would require refactoring all targets and move callmain.s to libsrc/{targets} instead of runtime.

colinleroy:
@polluks your thumbs up means you're fine with that plan?
@oliverschmidt, what do you think of it?

Thanks!

oliverschmidt:
> and stick the _exit code right after that in callmain.s
 
 I'm not sure I really get what you have in mind. I suggest to change this PR in the way you have in mind - but just for the Apple II (breaking the other targets). Then at least I can better understand and provide feedback.

colinleroy:
@oliverschmidt : Something like that. (It doesn't even break the other targets if duplicating the callmain.s file to libsrc/apple2/, it takes precedence over the one in libsrc/runtime).

polluks:
I suppose we can change other targets later, just without benefit but unity.

oliverschmidt:
@colinleroy Thanks for showing. Now even I understand :relaxed:

I like the approach from the control flow perspective :+1:

Regarding the source file layout it feels strange after so many years with cc65 to have _exit not it ctr0.s. But taking a step back it seems a viable approach even if it means that zpsave, rvsave and reset now become global symbols.

colinleroy:
@oliverschmidt thanks for your feedback!
regarding your last comment : we could put back the contents of apple2/callmain.s into crt0 so we don't globalize symbols (and .ifndef __APPLE2__ in runtime/callmain.s)?

colinleroy:
I can push that to this PR if you find it better: https://github.com/cc65/cc65/compare/master...colinleroy:cc65:apple2-fix-exit-STARTUP-2?expand=1

colinleroy:
It could also be better to have a comment-only callmain.s in apple2, so we don't touch the existing one at all and explain the difference in there.

oliverschmidt:
Thanks and done.

PR #2489, opened 2024-08-25
Add fnc to change working screen addr

MonteCarlos:
Attempt to implement functionality to switch the screen memory area used for output by BSOUT based code.

The following was implemented as a starting point:

1. Switch screen memory area used for output
1. Switch viewport (sets the vic and cia registers to display alternate screen)
1. Check the given screen address if it is plausible (Deny shadowing areas from $1000-$1fff, $9000-$9fff and register area from $d000-$dfff)
1. Check if given screen and charset are in the same vic bank when setting viewport
1. Backup and restore cursor position
1. Testprogram


MonteCarlos:
The original intend was to enable the screen output at another memory location but 0x400. 
This is implemented, already.
However, several things have to be considered:
- Should we add a function to reset the screen to the original location, too?
- Should we save and restore cursor position? In the simple first attempt, there were gaps left, when printing, switching, printing and switching back. The reason was that the cursor position advanced independently of the selected screen memory.
- Should we allow only original and alternate screen or should it be possible to switch screen memory to several different locations? If allowing several different screen locations, howto deal with cursor positioning?
- Should the (alternate) screen be cleared after switching?
- Should we add a function to switch the viewport, not only the screen where the writes go?
- ...

mrdudz:
This looks very overengineered to me... way too much code for something thats basically a single "poke".

IMHO there should be a single function that sets the screen location - and only that. no clrscr, no checking, no cursor things(*) . Instead of error checking, provide a bunch of constants for all legal screen locations. (That also solves the problem of the valid locations being different on different cbm targets). That code also seems to assume exactly two screens will be used, which is strange, and an unnecessary limitation.

(*) Maintaining the cursor position when switching forth and back between two screens will probably not even work properly, due to how "logical" lines work in the cbm kernal. It's fine IMHO when switching screens requires doing clrscr() and cursor positioning manually - in most cases i guess the screen will be switched exactly once anyway, and then we have a lot of overhead for this general case.

MonteCarlos:
I could implement independent functions each for one and only one thing and remove the checks. 

- One function to switch the screen
- One function to set registers given screen and charset addr
- Two functions to backup / restore the cursor pos

If they are implemented per file, then the linker will only include the used code.

mrdudz:
@MonteCarlos are you still working on this?

IMHO we should really start with two simple functions. Something like this:

```csetscreen(unsigned char addrhi)```
and
```csetcharset(unsigned char addrhi)```

Those should set up the respective registers ($dd00 and $d018 for the C64) as needed, plus some runtime variables perhaps (like the vramhi for kernal), and not do anything else. No checking, no screen clearing, no save/restore no nothing. That will already cover many usecases, and not add a lot of overhead to programs that just want to use a custom charset (for example). This API can also work easily on many other targets (although it is a bit target specific, obviously)

That said, perhaps the argument to the function shouldnt be the addr hi, but instead some constants/enum/whatever that a target can define as needed - that might save a bit of bitfiddling inside those two functions.

eg sth like:
```
#define CSCREEN04 0x10
#define CSCREEN08 0x20
#define CSCREEN0C 0x30
#define CSCREEN20 0x80
#define CSCREEN24 0x90
...
#define CCHARSET10 0x04 /* ROM */
#define CCHARSET18 0x06 /* ROM */
#define CCHARSET20 0x08
...
```

mrdudz:
@MonteCarlos ping?

I am tempted to close this - the existing PR is way "too much"

PR #2492, opened 2024-09-01
Allow alternative names for pragmas that contain underlines instead of dashes

### LINKED ISSUE #2470: Minus sign in identifiers following #pragma violate ANSI-C token rules
According to [ISO C standard (ISO/IEC 9899)](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n2310.pdf), section 6.10.6.

Perhaps use underscores instead?

kugelfuhr:
This fixes #2470


mrdudz:
Thanks!

PR #2494, opened 2024-09-01
Fix issue #1744 for structs. Allow it for enums.

### LINKED ISSUE #1744: ca65: The assembler allows anonymous structs/unions/enums on top level
The assembler allows anonymous structs or enums on top level. Such constructs are both useless and unusable. What actually happens is that symbols are introduced in the global scope:
```Assembly
.struct
        foo     .word
.endstruct
.union
        bar     .word
.endunion
.enum
        baz
.endenum

foo:
bar:
baz:
```
Gives:
    
```bash
$ ca65 test.s
test.s:11: Error: Symbol 'foo' is already defined
test.s:11: Error: Symbol '.size' is already defined
test.s:12: Error: Symbol 'bar' is already defined
test.s:12: Error: Symbol '.size' is already defined
test.s:13: Error: Symbol 'baz' is already defined
```
Funny about it is that the source code explicitly says "Outside of other structs, we need a name" but then fails to enforce this.

kugelfuhr:
Fixes #1744

### DIFF
--- a/src/ca65/enum.c
+++ b/src/ca65/enum.c
@@ -63,7 +63,9 @@ void DoEnum (void)
     long      Offs     = 0;
     ExprNode* BaseExpr = GenLiteral0 ();
 
-    /* Check for a name */
+    /* Check for a name. We allow anon structs even on top level. This will
+    ** generate a set of constants in the current scope.
+    */
     int Anon = (CurTok.Tok != TOK_IDENT);
     if (!Anon) {
         /* Enter a new scope, then skip the name */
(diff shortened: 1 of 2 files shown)


kugelfuhr:
Ok. This doesn't work with the ugly `.org` within structs. Using `.org` within structs requires the ability to define anonymous structs on top level because they are abused to generate constants.

This means that this PR and #1744 can be closed.

kugelfuhr:
The addition of `.org` within structs (added without PR/discussion, at least I didn't find one) made anonymous structs on top level a requirement so it cannot be considered a bug and doesn't need to be fixed. Closing.

PR #2495, opened 2024-09-01
Generate shorter code for main() in some cases

kugelfuhr:
With this change, the generated code for `main()` ...
* ... will no longer cleanup the stack on exit since the program terminates anyway, and
* ... will not save the old contents of register variables on entry since they do not contain anything useful.

This reduces code size for all programs where `main()` has parameters or declares variables; even more if there are register variables.

### DIFF
--- a/src/cc65/codegen.c
+++ b/src/cc65/codegen.c
@@ -507,34 +507,39 @@ void g_enter (unsigned flags, unsigned argsize)
 
 
 
-void g_leave (void)
+void g_leave (int IsMainFunc)
 /* Function epilogue */
 {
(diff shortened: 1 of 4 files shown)


mrdudz:
Sounds good, thanks!

kugelfuhr:
> Hm, then `main()` can't be called recursively within the program, right‽ Never thought about this before, but it _should_ be possible, or is `main()` more special than I thought?

Yes, recursive calls should be possible - at least in standard compliant mode. I've realized that shortly after the change has been merged. A change fixing this is already prepared and I will publish a PR soon.

PR #2496, opened 2024-09-01
Fix macro preprocessing for #include

### LINKED ISSUE #2458: Preprocessor ignores context: GEOS string fails because of header order
```c
#include <geos/gsym.h>
#include <string.h>

void main()
{
}
```
```
$ cl65 -tgeos /tmp/a.c
/tmp/a.c(2): Error: Include file '(*(unsigned*)0x24).h' not found
```
See https://github.com/cc65/cc65/blob/0541b65aa48ea2508b218ff7097911e741b4bc12/include/geos/gsym.h#L73

kugelfuhr:
Fix macro preprocessing for #include. Arguments enclosed in "" or <> must not be preprocessed. See ISO/IEC 9899 1990 (E) section 6.8.2.

Please note that this change doesn't make the compiler completely standard compliant. To do that, token look ahead for many tokens would be required which is difficult to do. But it's a much better approximation than before :smile: 

This change fixes #2458.


mrdudz:
Cool, thanks!

Issue #2044, opened 2023-03-30
imports in o65 format config lead to "ld65: Internal Error: Invalid address size: 0"


ghost:
ld65 returns an internal error when using the "import" directive in the o65 format section in the config file.

Here's a tarball with the repro case.
[repro.tar.gz](https://github.com/cc65/cc65/files/11109936/repro.tar.gz)

```
$ make
ca65 -o test.o test.s
ld65 -m test.map -C firmware.cfg -o test.o65 test.o
ld65: Internal Error: Invalid address size: 0
make: *** [Makefile:7: test.o65] Error 1
```

mrdudz:
closed via #2497

PR #2497, opened 2024-09-01
Fix issue #2044

kugelfuhr:
Assign an address size for imported and exported o65 symbols. Do not try to output exports to the map file that do not have a value because they are actually o65 externals.

This change fixes issue #2044.


mrdudz:
Thanks!

Issue #2159, opened 2023-08-11
ca65: .INCLUDE is also broken within .REPEAT


kugelfuhr:
It is known that .INCLUDE is broken when used within a macro. The behavior is that all include statements within the macro are not executed when the macro is expanded, but when macro expansion is over. So actually they are executed out of order.

I found that this is also true for .REPEAT. This code:

```
$ cat test.inc
        lda     #sym
$ cat test.s
.repeat  10, i
        sym     .set    i
        .include        "test.inc"
.endrep
```

does not generate 10 `lda` instructions with an increasing immediate operand, but 10 `lda` instructions that load the immediate value 9.

The behavior of .INCLUDE within macros was changed in #1477 in an attempt to "fix" it. It was found that this change breaks existing code in #1503 and #1655 and then reverted. If it will ever get fixed, this has to be done for .REPEAT also.

bbbradsmith:
Related to #2155

mrdudz:
This never worked, right?

kugelfuhr:
> This never worked, right?

Since it seems to be related to .INCLUDE handling which has always been broken, I would say yes. But I don't know for sure.

mrdudz:
closed via #2498

PR #2498, opened 2024-09-01
Fix .include within .macro/.repeat

kugelfuhr:
Fix behavior of .INCLUDE within a macro or .REPEAT. In the original code .INCLUDE was executed after expansion of the macro or .REPEAT - which was wrong and caused all sorts of unexpected behavior. Related issues/PRs are #231, #1473, #2159 and maybe others.
    
Note: After this change, diagnosed locations for errors in files included within .MACRO/.REPEAT are wrong. This is an unrelated bug that was always there but wasn't triggered since .include didn't work within .MACRO/.REPEAT. It gets exposed by this fix and needs to be addressed in a separate PR.


mrdudz:
Great, thanks!

PR #2499, opened 2024-09-02
Disallow recursive calls to main() in cc65 mode

kugelfuhr:
The change from #2495 didn't take into account that recursive calls to `main()` are legal in C. With the changes from #2495, such calls will generate unexpected results because no stack cleanup is done on function return.

But recursive calls to main() are rare and on the 6502 every byte is precious. So this change limits the effect of #2495 to cc65 mode and at the same time disallows recursive calls to `main()` in this mode.

So with `--standard cc65` (the default) recursive calls to `main()` aren't possible and the compiler omits the cleanup code in the function epilogue of `main()`, while in all other modes recursive calls of `main()` are allowed and the compiler generates the standard epilogue code for `main()`.

### DIFF
--- a/doc/cc65.sgml
+++ b/doc/cc65.sgml
@@ -823,6 +823,11 @@ and the one defined by the ISO standard:
         as it sounds, since the 6502 has so few registers that it isn't
         possible to keep values in registers anyway.
         <p>
+<item>  In <tt/cc65/ mode, <tt/main()/ cannot be called recursively. If this
+        is necessary, the program must be compiled in <tt/c89/ or <tt/c99/ mode
+        using the <tt><ref id="option--standard" name="--standard"></tt>
+        command line option.
+        <p>
 </itemize>
 
 There may be some more minor differences I'm currently not aware of. The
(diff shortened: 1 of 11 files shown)


kugelfuhr:
By the way, is there an easy possibility to add tests that are successful with one `--standard` argument and fail for some others? I would have added tests but didn't see an easy way to do so.

mrdudz:
> By the way, is there an easy possibility to add tests that are successful with one `--standard` argument and fail for some others? I would have added tests but didn't see an easy way to do so.

Unfortunately not - the one and only test for which that is true lives in "misc" right now, with a hardcoded recipe in the makefile.

mrdudz:
I'll have to trust you that this is fine - thanks :)

groessler:
On 9/7/24 14:39, Bob Andrews wrote:
>
> I'll have to trust you that this is fine - thanks :)
>

I just now only looked at the diff.

Why was targettest/atari/mem.c changed? Can't the address of main() be 
used anymore?


regards,
chris

kugelfuhr:
> Why was targettest/atari/mem.c changed? Can't the address of main() be used anymore?

No, in cc65 mode the address cannot be taken. Otherwise the following would be possible, crashing the machine:

```C
int main(void)
{
    int (*mainfunc(void) = main;
    return mainfunc();
}
```

PR #2502, opened 2024-09-03
Fix issue #2461

### LINKED ISSUE #2461: Probable miscompilation involving complex bit manipulation code and a specific subfunction coincidence
The compiler does not seem to handle complex bitfield manipulations when it coincides with parameter clobbering.
See  [this godbolt snippet](https://godbolt.org/z/8rhz1PboP) for the minimal offending example.
This seems to be a bug in cc65, but is unclear. Other compilers (gcc,armcc,etc) work fine with the code.

kugelfuhr:
This change fixes #2461. Not only the '+' operator was affected, but also '&', '|' and '^' (all commutative operators).

The funny thing is that the transformed code was always wrong if the necessary preconditions were met. So even when it seemed to work (see function `original()` in file `opt-bug-2463.c` in the linked zip) this was just out of coincidence.

### DIFF
--- a/src/cc65/codeoptutil.c
+++ b/src/cc65/codeoptutil.c
@@ -1128,8 +1128,10 @@ void AddOpHigh (StackOpData* D, opc_t OPC, LoadInfo* LI, int KeepResult)
             InsertEntry (D, X, D->IP++);
         }
 
-        /* In both cases, we can remove the load */
-        LI->X.Flags |= LI_REMOVE;
+        /* If this is the right hand side, we can remove the load. */
+        if (LI == &D->Rhs) {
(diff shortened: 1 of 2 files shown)


kugelfuhr:
You might want to run a few tests on your own before merging.

mrdudz:
More tests would be nice indeed (why the heck did i name the original test opt-bug-2463.c ? *shrug*)

mrdudz:
I'll wait a bit for @EDI-Systems to test this (hopefully)

EDI-Systems:
I think this works.

mrdudz:
OK, lets see what happens then :)

PR #2503, opened 2024-09-04
Optimize stpcpy's size and speed

colinleroy:


mrdudz:
please add a comment to strcpy that tells how stpcpy relies on the values in a/x/y - else we have a timebomb that goes off when someone changes strcpy :)

colinleroy:
Hi @mrdudz , good idea. I added a comment, and reworked stpcpy.s just to make clear it depends on ptr2 and Y (not on A or X)

mrdudz:
Thanks, this is fine :)

PR #2508, opened 2024-09-07
[ATMOS] waitvsync

polluks:


libsrc/atmos/waitvsync.s
```diff
@@ -0,0 +1,18 @@
...
+
+        .include "atmos.inc"
+
+.proc   _waitvsync
+
+wait:   lda     VIA::PRA2
+        and     #%00010000      ; CB1
+        bne     wait
+        rts
+
```
mrdudz, line 17: what about swapping the arguments, that might save a cycle or two at runtime

```
lda #%00010000
wait:
and VIA::PRA2
bne wait
```

PR #2509, opened 2024-09-07
Remove explicit keyword

polluks:
Better code consistency


groessler:
I actually prefer the "extern" keyword for extern functions.

polluks:
Right now we have 122 headers:
* 2 "groessler style"
* 120 implicit style

mrdudz:
using extern for functions in header files is redundant, i also prefer not to have it there for that matter. I will update the contributing.md :)

PR #2511, opened 2024-09-08
In time(), set the clock id to CLOCK_REALTIME when calling clock_gettime.

coronax:
While implementing time functions for my homebrew, I noticed that time() allocates stack space for the clock id argument when it calls clock_gettime(), but doesn't actually set a value. So the system-specific clock_gettime() implementation receives some arbitrary clock id value. I set it to CLOCK_REALTIME so that at least it's doing something consistent.


mrdudz:
thanks!

PR #2512, opened 2024-09-08
Warn for braces around a pointer initializer

### LINKED ISSUE #2134: cc65: Warning about braces around scalar initializer is not output for all scalars
This code line:

```C
int i = { 0 };
```

generates 

```
test.c:1: Warning: Braces around scalar initializer
```

but this code line:

```C
char* p = { 0 };
```

compiles without a warning. Both variables are scalars.

kugelfuhr:
This fixes #2134 by adding a warning similar to that used for other scalars.


mrdudz:
thanks!

Issue #2514, opened 2024-09-10
lastline.sh complains about empty files


kugelfuhr:
It should be possible to add empty files (for example for tests) but it is not:

```
$ cd test/val
$ echo "#include \"foo.h\"" > foo.c
$ touch foo.h
$ make -C ../.. --no-print-directory checkstyle
error: found following files that have no newline at the end:
./foo.h
./test/val/foo.h
make[1]: *** [Makefile:26: lastline] Error 255
make: *** [Makefile:44: checkstyle] Error 2
```

The problem is causes by the lastline.sh file that doesn't exclude empty files before checking that the last line contains a linefeed.

mrdudz:
It should probably exclude the test directory instead (i don't see a reason to add empty files elsewhere - and in the test dir anything should be possible, i guess)

kugelfuhr:
I don't think its a good idea to allow anything in the test directory. You don't want source files with cr/lf even in this directory and also checking for tabs and dangling white space does no harm.

mrdudz:
Well, it doesn't. BUT the compiler also should be able to deal with it :)

PR #2516, opened 2024-09-12
Make lastline.sh ignore empty files

kugelfuhr:
This fixes issue  #2514.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,4 +9,4 @@
 /cc65.zip
 /util/atari/*.exe
 /util/gamate/*.exe
-
+targettest/cbm/cbmread.prg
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -9,4 +9,4 @@
 /cc65.zip
 /util/atari/*.exe
 /util/gamate/*.exe
-
+targettest/cbm/cbmread.prg
```
mrdudz, line 12: would you remove this from the PR please?

kugelfuhr: Why? What's wrong with it?

mrdudz: Well. IF we start adding the test binaries to the git ignores... then all of them, using proper wildcards. a single one is just weird :) (and i prefer different things in separate commits)

kugelfuhr: Sounds a bit like bike shedding looking at the current file contents, but ok.

mrdudz:
thanks!

PR #2517, opened 2024-09-12
Improved/fixed the time() function

kugelfuhr:
Looking at #2511 I noticed that the `time()` function could be improved and that it doesn't always work as it should. 

When the underlying `clock_gettime()` function returns an error, the existing implementation would return `(time_t)-1` but didn't store this value into `*timep` as is required by the standard.

So this change ...
* ... fixes the error and will causes `time()` to always return identical numbers for the return value and `*timep`.
* ... reduces the code size by 7 bytes.
* ... doesn't use `ldeaxi` which may lead to additional size decreases depending on the program.
* ... adds a test.

### DIFF
--- a/libsrc/common/time.s
+++ b/libsrc/common/time.s
@@ -22,55 +22,50 @@
 
 ; Get the time (machine dependent)
 
+        .assert timespec::tv_sec = 0, error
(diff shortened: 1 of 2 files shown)


mrdudz:
thanks!

Issue #2515, opened 2024-09-11
Missing / excessive diagnostics for preprocessing directives


pmor13:
Compiling this code:
```
#line 13"x"
```
leads to:
```
Error: #line directive requires a simple decimal digit sequence
```
while it does not violate any syntax rule or constraint.

Compiling this code:
```
#define X"y"
```
leads to:
```
<nothing>
```
while it violates constraint "There shall be white-space between the identifier and the replacement list in the definition
of an object-like macro." A conforming implementation shall produce at least one diagnostic message.

mrdudz:
> while it violates constraint "There shall be white-space between the identifier and the replacement list in the definition
of an object-like macro."

To be exact: in C99 mode

mrdudz:
fixed via #2518

PR #2518, opened 2024-09-12
Fix minor preprocessor problems

kugelfuhr:
This PR fixes both issues from #2515.

I would have also added a test which was already written, but there seems to be no simple way to execute it in a way that works with the current test infrastructure.

### DIFF
--- a/src/cc65/preproc.c
+++ b/src/cc65/preproc.c
@@ -2913,7 +2913,7 @@ static unsigned GetLineDirectiveNum (void)
 
     /* Ensure the buffer is terminated with a '\0' */
     SB_Terminate (&Buf);
-    if (SkipWhitespace (0) != 0 || CurC == '\0') {
+    if (SB_GetLen (&Buf) > 0) {
         const char* Str = SB_GetConstBuf (&Buf);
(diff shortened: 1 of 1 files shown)


mrdudz:
Thanks!

would you post the test anyway? I'll try adding it in misc then

kugelfuhr:
This test file

```c
#line 13"x"
#define X"y"
int main() { foo; }
```
generates 
```
test.c:1: Error: #line directive requires a simple decimal digit sequence
test.c:3: Error: Undeclared identifier 'foo'
```

before the fix. After the fix, the output depends on the `--standard` setting. With no standard or `--standard c89` it generates

```
x:14: Error: Undeclared identifier 'foo'
```

while it says 

```
x:13: Warning: ISO C99 requires whitespace after the macro name
x:14: Error: Undeclared identifier 'foo'
```

using `--standard c99`.

mrdudz:
Oh this is a tricky one :)

mrdudz:
added the test in https://github.com/cc65/cc65/commit/7dc09fdb05051de5273abb83679924b7b265a680

PR #2528, opened 2024-10-21
checkversion.c fix missing shifts

lcvgit:
Added missing shifts for checking major version.


mrdudz:
Wth, how could this even sneak in... some odd compiler bug at the time i added it? *shrug*

thanks!

PR #2534, opened 2024-11-10
Apple2: Provide a way to get directory file count

colinleroy:
The information is available in the directory key block. Providing it to the user as soon as opendir() is done can save them costly code.

### DIFF
--- a/include/apple2.h
+++ b/include/apple2.h
@@ -232,6 +232,11 @@ struct tm* __fastcall__ gmtime_dt (const struct datetime* dt);
 time_t __fastcall__ mktime_dt (const struct datetime* dt);
 /* Converts a ProDOS date/time structure to a time_t UNIX timestamp */
 
+typedef struct DIR DIR;
+
+unsigned int __fastcall__ dir_file_count(DIR *dir);
+/* Returns the number of active files in a ProDOS directory */
+
 #if !defined(__APPLE2ENH__)
 unsigned char __fastcall__ allow_lowercase (unsigned char onoff);
 /* If onoff is 0, lowercase characters printed to the screen via STDIO and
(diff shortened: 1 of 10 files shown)


oliverschmidt:
I totally see the point and I like the idea to approach it via a function taking a DIR*.

However, I don't think that function declaration should go into dirent.h. I rather see it similiar to https://github.com/cc65/cc65/blob/master/include/apple2.h#L209-L227 meaning that target specific extension to standard interfaces go into into the target specific header. That's at least how it is done so far for the Apple II.

Regarding the function name: If it would be in dirent.h, it would have an apple2_ prefix, no prodos_ prefix. ProDOS is an implementation detail that the cc65 user is supposed to be shielded from! A prodos_ prefix would only have to be considered if there were an opendir() for DOS3.3.

However, when the function declaration moves to apple2.h then I don't see a prefix at all. Conceptually it's very well possible that the same functionality is added for another target as well. I.e. videomode() was originally C128 specific. Then I implemented it for the Apple //e too.

colinleroy:
Thanks! This makes sense. Addressed in latest push, along with the doc update.

include/apple2.h
```diff
@@ -232,6 +232,11 @@ struct tm* __fastcall__ gmtime_dt (const struct datetime* dt);
 time_t __fastcall__ mktime_dt (const struct datetime* dt);
 /* Converts a ProDOS date/time structure to a time_t UNIX timestamp */
 
+typedef struct DIR DIR;
```
oliverschmidt, line 235: I'm unsure, but I believe that the usual cc65 style is to rather include dirent.h than to duplicate a typedef (?)

oliverschmidt:
I assume that the function also counts subdirectories. So the name should rather be dir_entry_count, shouldn't?

oliverschmidt:
Thanks :smile:

PR #2547, opened 2024-11-30
Fixed the behavior of JMP (ind) in sim65 when it runs with the 6502X CPU type.

sidneycadot:
The JMP (ind) bug is present in the 6502 which is emulated by both the "6502" and "6502X" emulation targets of sim65; specifically, the OPC_6502_6C handler. In the old code, the bug-exhibiting code was not executed when the target was set to 6502X, which is incorrect. The patch removes the (CPU == CPU_6502) check, which made no sense.

The JMP (ind) bug was actually fixed in the 65c02. Indeed, the OPC_65C02_6C opcode handler has code that implements the 'right' behavior.

### DIFF
--- a/src/sim65/6502.c
+++ b/src/sim65/6502.c
@@ -1964,25 +1964,17 @@ static void OPC_6502_6C (void)
     PC = Regs.PC;
     Lo = MemReadWord (PC+1);
 
-    if (CPU == CPU_6502)
-    {
-         /* Emulate the 6502 bug */
-        Cycles = 5;
-        Regs.PC = MemReadByte (Lo);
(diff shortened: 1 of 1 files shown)


kugelfuhr:
When touching the code anyway, I would suggest to fix the brace style in the piece of code you changed. It doesn't match the one [described in the wiki](https://github.com/cc65/cc65/blob/master/Contributing.md).

mrdudz:
> If making the coding style consistent is a goal, that warrants a separate issue, with separate patches to fix precisely and only those coding style consistency issues.

yes indeed. don't mix style fixes with code fixes

mrdudz:
Thanks!

PR #2548, opened 2024-11-30
Fixed clock-cycle timing of branch (Bxx) instructions.

sidneycadot:
Branch instructions, when taken, take three or four cycles, depending on whether a page is crossed by the branch.

The proper check to determine whether the extra cycle must be added is the target address of the branch vs the address immediately following the branch.

In the former version of the BRANCH instruction handler, the target address was incorrectly checked vs the address of the branch instruction itself.

The corrected behavior was verified against a real 6502 (Atari) and the 65x02 testsuite.


mrdudz:
Thanks!

PR #2549, opened 2024-11-30
Add section to ld65 doc about debug info

clydeshaffer:
Ported the information from https://github.com/cc65/wiki/wiki/Debug-info-data to the ld65 linuxdoc page.

The goal of this PR is solely to _port_ the content rather than edit or add to it.

### DIFF
--- a/doc/ld65.sgml
+++ b/doc/ld65.sgml
@@ -1180,6 +1180,202 @@ The ZPSAVE segment contains the original values of the zeropage locations used
 by the ZEROPAGE segment. It is placed in its own segment because it must not be
 initialized.
 
+<sect>Debug Info<p>
+
+The debug info and the API mirrors closely the items available in the sources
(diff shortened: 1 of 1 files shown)


clydeshaffer:
One small edit to content: the "Segments" subsection from the DebugInfo page was changed to "Segment Info" because there is already a "Segments" section on the ld65 page.

Undrafting the PR now that it's been verified to build with `linuxdoc-tools`.

(Incidentally, I wasn't able to add this package from my MSYS2 dev environment, and had to switch to another machine where I use WSL Ubuntu)

mrdudz:
Cool, thanks!

mrdudz:
That said, it would be super nice if you could add https://github.com/cc65/wiki/wiki/Debug-info-overview before (this can be shortened a bit, i guess) and https://github.com/cc65/wiki/wiki/Debug-info-module (this can probably also be reworded/shortened) after this info

PR #2550, opened 2024-11-30
Fixed behavior of the 65C02 "BIT #imm" instruction in sim65.

sidneycadot:
The BIT #imm instruction behaves differently from the BIT instruction with other addressing modes, in that it does /not/ set the N and V flags according to the value of its operand. It only sets the Z flag, in accordance to the value of (A & operand).

This is corroborated in two ways:

- The 65x02 test suite;
- Documentation about BIT #imm such as http://www.6502.org/tutorials/65c02opcodes.html

This patch implements the correct behavior for BIT with immediate addressing. The patched version passes the 65x02 test suite for 65C02 opcode 0x89.


mrdudz:
thanks!

PR #2551, opened 2024-12-01
Fixed wrong clearing of D-flag on interrupts for sim65 with 6502X CPU.

sidneycadot:
The 65C02 clears the D flag on interrupts while the 6502 does not.

The old sim65 code erroneously cleared the D flag also for the 6502X CPU type, which was incorrect.


mrdudz:
Thanks!

PR #2553, opened 2024-12-03
Sim65: removed ZR register from CPURegs type.

sidneycadot:
The sim65 simulator has a struct type "CPURegs" for holding the values of the CPU registers while executing instructions.

This type currently has a ZR field, which is not a register available in the 6502 or 65C02. This field is also not used anywhere.

This PR removes this field.


mrdudz:
Thanks!

Issue #2539, opened 2024-11-16
Comprehensive 6502 testsuite run on sim65 exposes a number of implementation bugs.


sidneycadot:
Hello,

After an initial report on a simulator bug (#2536), and following a suggestion from user @kugelfuhr in that issue's discussion, I implemented a program to test the "sim65" simulator against the comprehensive "65x02" suite of 6502 instruction behavior tests (https://github.com/SingleStepTests/65x02).

The repository containing the sim65-test program can be found here:

https://github.com/sidneycadot/sim65-test-prototype

This test exposes several bugs in sim65, among which the original bug as reported in #2536, but also several overflow bugs in the simulator code, as well as several issues with the emulation of other 6502 instructions. For example, the tests indicate that the cycle counts of branch instructions that do page-crossing are not correct in some circumstances.

I'd like to discuss several issues here:

* Is someone in particular the go-to person for issues with the sim65 simulator at this point in time?

* What is the best way to handle the issues exposed by these tests? Should we make a single issue-tracker issue for each type of behavioral discrepancy, or should someone make a dedicated effort to minimize or eliminate the discrepancies between sim65 behavior and the testsuite behavior?

* To what extent do we trust the testsuite? At first sight it looks well done, but there is no real traceability on how the test cases were created. We can either decide to trust the testsuite as is, and strive for compliance; or, in every instance where the behavior of sim65 deviates from the behavior described in the testsuite, investigate what happens on a real (physical) 6502. The latter is potentially quite time-consuming.

* It would be possible to integrate the sim65-test program (or something like it) into the cc65 project; perhaps as a stand-alone test program, or as a part of the standard test procedure, or as part of the sim65 program (an idea by user @kugelfuhr, not necessarily my preference). Would that be desirable? Keeping sim65-test as a separate project is also certainly an option, I think.

* The sim65 simulator code (6502.c, in particular the 'ExecuteInsn' function) is pretty nice; however, it is also obvious that it is not written to be used in a context other than running the sim65 program. I had to deal with / cut several implicit dependencies of the 6502.c code to the sim65 code around it (see the "How it works" section in https://github.com/sidneycadot/sim65-test-prototype/blob/main/README.md). It would perhaps be possible to make slight changes to the sim65 code to make it easier to test its core 'ExecuteInstn' routine.

sidneycadot:
To make all this a bit more concrete, here's an overview of the test results of sim65 vs the 65x02 testsuite.

[...]

Issue #2355, opened 2024-01-14
[Feature Request] Add performance counter readonly register to sim65


jroweboy:
A feature request for sim65, right now it only supports printing out the cycle count after program execution, but for performance testing, it would be much more handy if you could access the performance counter directly in the executing application.

### Motivation
The challenge with performance testing with sim65 right now is there's no way to filter out any of the driver and setup code. Additionally, even if you have a minimal test driver, there's not a good way to do average performance metrics, or finding the minimum cycle count/maximum cycle count for code that will change over multiple runs.

### Implementation
One way to achieve this without affecting any existing code would be to add a virtual file that the application can open with `open` with a name that is illegal/unlikely for a file to contain. For Linux, its not really possible to have such a filename, as the only illegal characters in filenames are `/` and `\0` which wouldn't work, but its close enough to just use something illegal in windows like `:`.

The cycle counter virtual register can be an unsigned long long, and reading this will return the number of cycles since the start of the application. The bytes returned can be in little endian order, so if the user only needs 16 or 32 bits of counter, they will be able to just read 2 or 4 bytes from the virtual register instead of the full 8 bytes. 

The file descriptor returned can be a negative number. Since this usually indicates error, on the very very very small chance the user has a file named `:cyclecount` that they actually intended to open, the negative return value should tip them off that something is up. The actual number for the FD could be a constant like `-2` to disambiguate with the OS provided error. This is admittedly a bit counter intuitive to return an error, but this way it shouldn't conflict with any real file descriptors on the OS.

Attempting to write to the virtual register will return an error. I don't think its necessary to prevent opening the file with write mode enabled (as it appears to be the default for the open call).

### Example
So with that in mind here's how I could see it working with a small example.

```c
static unsigned long start_count;
static unsigned long end_count;
static unsigned long cycles;
static unsigned long min_cycles, max_cycles, avg_cycles, total_cycles;

int main() {
  // Open the virtual cycle count register
  int perf = open(":cyclecount");
  if (perf >= 0) {
    // Error! Expected virtual register
  }
  
  // Find out how much overhead the reads add.
  read(perf, (void*)&start_count, sizeof(start_count));
  read(perf, (void*)&end_count, sizeof(end_count));
  unsigned int overhead = end_count - start_count;
  
  // Run 100 samples of the task that we are measuring
  for (char i=0; i < 100; i++) {
    read(perf, (void*)&start_count, sizeof(start_count));
    // Do Task
[...]

PR #2555, opened 2024-12-03
sim65: changing memory access types to uint8_t and uint16_t.

sidneycadot:
In sim65, simulator memory access to a 64 KB array is implemented via functions defined in src/sim65/memory.h and src/sim65/memory.c.

In the old version, the types for both content bytes (8 bits), content words (16 bits), regular addresses (16 bits), and zero-page addresses (8 bits) were all given as bare 'unsigned'.

This lead to several cases of address overrun (e.g., when an instruction wraps around from address 0xffff to 0x0000) when running the simulator against a stress test (specifically, the 65x02 test suite). To protect from this, and to more properly express the bit width of the types involved which is a good idea anyway, we start using the fixed-width types provided by 'stdint.h'.

In the process, we also change the MemReadByte macro to a full function call. This may impact performance (by a small amount), but it improves memory safety, as cases where the address is accidentally expressed as a value exceeding 0xffff are handled by wrap-around (as it is in the actual hardware), rather than causing access outside of the Mem[] array where the 64 KB of simulated RAM resides.

The reason for this patch is twofold.

(1) It is a partial patch for issue #2539.

Several issues brought to the surface by running the 65x02 testsuite are eliminated by these changes. In the discussion about this issue, it was concluded that it is a Good Idea to use the stdint-types, both for the simulated CPU registers and for the memory. This patch addresses the memory-part of that change.

(2) It is a precursor patch for issue #2355.

For that issue, we will implement a memory-mapped timer register. This will make handling of memory access in the simulator a bit more complex.

Having proper functions with the proper types in place will help to make the timer register patch easier.

### DIFF
--- a/src/sim65/memory.c
+++ b/src/sim65/memory.c
@@ -45,8 +45,8 @@
 
 
 
-/* THE memory */
-unsigned char Mem[0x10000];
+/* The memory */
+uint8_t Mem[0x10000];
 
 
 
@@ -56,7 +56,7 @@ unsigned char Mem[0x10000];
 
 
 
-void MemWriteByte (unsigned Addr, unsigned char Val)
+void MemWriteByte (uint16_t Addr, uint8_t Val)
 /* Write a byte to a memory location */
 {
     Mem[Addr] = Val;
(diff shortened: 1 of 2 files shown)


mrdudz:
thanks

PR #2556, opened 2024-12-03
Sim65 : improve implementation of ROL and ROR operations

sidneycadot:
Issue #2539 demonstrates a number of issues in the sim65 simulator.

Several issues can be traced back to undesirable side effects of the
use of bare 'unregistered' types for the CPU registers in the CPURegs
type defined in '6502.h'.

The intention is to tighten the types of the registers defined there
to uint8_t and uint16_t, in accordance with the actual number of bits
that those registers have in the 6502. However, it turns out that about
10 opcode implementations make use of the fact that the register types
have more bits than the actual 6502 registers themselves.

In preparation of fixing the CPURegs types, we will first make sure
that those opcode implementations are changed in such a way that they
still work if the underlying register types are tightened to their
actual bit width.

Today's opcodes that get this upgrade are ROL and ROR.

### DIFF
--- a/src/sim65/6502.c
+++ b/src/sim65/6502.c
@@ -737,23 +737,29 @@ static unsigned HaveIRQRequest;
 
 /* ROL */
 #define ROL(Val)                                                \
-    Val <<= 1;                                                  \
-    if (GET_CF ()) {                                            \
-        Val |= 0x01;                                            \
-    }                                                           \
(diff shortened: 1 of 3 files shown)


sidneycadot:
Oops, I mixed up some branches. Retracting this PR and preparing a better one.

PR #2558, opened 2024-12-09
Clean-up preprocessor syntax

polluks:
Did you ever see `#include <foo.h>;`?
The C preprocessor has its own syntax.


mrdudz:
thanks!

PR #2560, opened 2024-12-15
Improve description of namespace access in ca65

binary-sequence:
This avoid confusion with referencing global scope with the namespace token.


mrdudz:
Thanks!

PR #2563, opened 2024-12-16
sim65: fixed ADC/SBC for the 65C02.

sidneycadot:
The current (before-this-patch) version of sim65.c does not correctly implement the ADC and SBC instructions in 65C02 mode. This PR fixes that.

The 6502 and 65C02 behave identically in binary mode; in decimal behavior however they diverge, both in the handling of inputs that are not BCD values, and in the handling of processor flags.

This fix restructures the original "ADC" and "SBC" macros in versions that are specific for the 6502 and the 65C02, and updates the opcode tables to ensure that they point to the correct implementations.

Considering the ADC instruction for a moment, the original "ADC" macro was changed to two macros ADC_6502 and ADC_65C02. These check the D (decimal mode) bit, and defer their implementation to any of three macros ADC_BINARY_MODE, ADC_DECIMAL_MODE_6502, and ADC_DECIMAL_MODE_65C02. This is a bit verbose but it makes it very clear what's going on.

(For the SBC changes, the analogous changes were made.)

The correctness of the changes made is ensured as follows:

First, an in-depth study was made how ADC and SBC work, both in the original 6502 and the later 65C02 processor. The actual behavior of both processors was captured on hardware (an Atari 800 XL with a 6502 and a Neo6502 equipped with a WDC 65C02 processor), and was analyzed. The results were cross-referenced with internet sources, leading to a C implementation that reproduces the exact result of the hardware processors. See:

https://github.com/sidneycadot/6502-test/blob/main/functional_test/adc_sbc/c_and_python_implementations/6502_adc_sbc.c

Next, these C implementations of ADC and SBC were fitted into sim65's macro-based implementation scheme, replacing the existing 6502-only implementation.

Finally, the new sim65 implementation was tested against the 65x02 testsuite, showing that (1) the 6502 implementation was still correct; and (2) that the 65C02 implementation is now also correct.

As a bonus, this new implementation of ADC/SBC no longer relies on a dirty implementation detail in sim65: the previous implementation relied on the fact that currently, the A register in the simulator is implemented as an "unsigned", with more bits than the actual A register (8 bits). In the future we want to change the register width to 8 bits, and this updated ADC/SBC is a necessary precursor to that change.

### DIFF
--- a/src/sim65/6502.c
+++ b/src/sim65/6502.c
@@ -556,15 +556,15 @@ static unsigned HaveIRQRequest;
 /* #imm */
 #define ALU_OP_IMM(op)                                          \
     unsigned char immediate;                                    \
-        MEM_AD_OP_IMM(immediate);                               \
-        Cycles = 2;                                             \
+    MEM_AD_OP_IMM(immediate);                                   \
+    Cycles = 2;                                                 \
     op (immediate)
 
 /* zp / zp,x / zp,y / abs / abs,x / abs,y / (zp,x) / (zp),y / (zp) */
 #define ALU_OP(mode, op)                                        \
     unsigned address, operand;                                  \
-        Cycles = ALU_CY_##mode;                                 \
-        MEM_AD_OP (mode, address, operand);                     \
+    Cycles = ALU_CY_##mode;                                     \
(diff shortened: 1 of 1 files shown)


mrdudz:
Thanks!

PR #2564, opened 2024-12-16
sim65: simulated CPU registers can be accessed from outside 6502.c.

sidneycadot:
The linkage of the 'Regs' variable in 6502.c was changed from static to extern. This makes the Regs struct visible (and even alterable) from the outside.

This change helps tools inspect the CPU state. In particular, it was implemented to facilitate a tool that verifies opcode functionality using the '65x02' test suite. But the change is also potentially useful for, e.g., an online debugger that wants to inspect the CPU state while the 6502 is being simulated.

### DIFF
--- a/src/sim65/6502.c
+++ b/src/sim65/6502.c
@@ -391,7 +391,7 @@ CPUType CPU;
 typedef void (*OPFunc) (void);
 
 /* The CPU registers */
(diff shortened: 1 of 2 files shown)


mrdudz:
Thanks!

Issue #2562, opened 2024-12-16
Atari: reading from stdin using gets/fgets is broken


sidneycadot:
The program below compiles but doesn't currently work as expected on the atari target. We never get to the second printf().

I also tried fgets(), same issue.

In the git log I see some work has been done on this early 2024, which seems to correspond to the time when this stopped working. But I still need to do a proper bisect.

```
#include <stdio.h>

int main(void)
{
    char s[100];
    for (;;)
    {
        printf("type a string.\n");
	gets(s);
        printf("you typed: %s", s);
    }
}
```

sidneycadot:
Okay I looked into this a bit more.

It seems the root cause of the problem is in libsrc/common/fgets.s

The fgets read loop reads characters until it sees a newline character (0x0a).

However, on the Atari, the end-of-line character when reading from the console is oddly represented as 0x9b, and no routine in-between seems to be doing character translation.

So, what would be the best place to solve this?

polluks:
Hmm, maybe Atari should define \n as \x9B?
C defines "\n", this is already an abstraction, we don't use "\x0A".

sidneycadot:
> Hmm, maybe Atari should define \n as \x9B?

I just checked -- they already do. '\n' is 155.

> C defines "\n", this is already an abstraction, we don't use "\x0A".

Currently libsrc/common/fgets.s does, so that must have slipped by code review:

```
        inc     ptr4
        bne     :+
        inc     ptr4+1

:       cmp     #$0A            ; Stop at \n
        beq     done
        bne     read_loop
```
I'll test to make sure that comparing to '\n' fixes the problem and if so, I'll submit a PR.

Thanks for the suggestion!

sidneycadot:
Question: what is the correct, target-independent way to compare to newline in ca65? "cmp EOL" perhaps?

polluks:
The assembler should use the same rules for simplicity.

sidneycadot:
I'm not going to touch ca65. Too much of a time investment, I have other dishes in the oven ...

polluks:
I see, how about a macro?

mrdudz:
shouldnt '\n' do what we want here (provided the right target with -t is used)?

polluks:
At least in the past this was C
https://www.cs.bu.edu/~jappavoo/Resources/psml/RHCOLAB-LINUXCC/6502/ext/cc65/libsrc/common/gets.c

sidneycadot:
[...]

PR #2565, opened 2024-12-17
Temporary fix for fgets() not using target-specific newline.

sidneycadot:
This patch provides a temporary fix for the issue where the fgets() function did not use the target-specific newline character to decide if it has reached the end of the line. It defaulted to the value $0a, which is the newline character on only some targets. The Atari, for example, has newline character $9b instead.

This issue was discovered while investigating #2562. 

This patch is ugly, because the ca65 assembler that is used for fgets doesn't currently accept C-type character escape sequences as values. Ideally we'd be able to write:

		cmp #'\n'

And this would end up being translated to a compare-immediate to the target-specific newline character.

Since that is impossible, this patch substitutes the equivalent, but ugly, code:

		.byte $c9, "\n"

This works because $c9 is the opcode for cmp #imm, and the "\n" string /is/ translated to the platform-specific newline character, at least when the 'string_escapes' feature is enabled.

### DIFF
--- a/libsrc/common/fgets.s
+++ b/libsrc/common/fgets.s
@@ -8,6 +8,8 @@
         .import         _fgetc, popptr1, pushptr1, popax, pushax, return0, ___errno
         .importzp       ptr1, ptr4
 
+        .feature        string_escapes
+
         .include        "errno.inc"
         .include        "stdio.inc"
         .include        "_file.inc"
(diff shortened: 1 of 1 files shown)


rofl0r:
ftr, this PR contains one unrelated commit, which you already submitted in another PR

sidneycadot:
Sigh, I'm a bit of a dumbass when it comes to git usage. (To be more precise: I think git sucks donkey balls in terms of usability.)

In my fork I accidentally committed that change to master, and it got into this development branch when I created it. It's hard to untangle this, for me at least. Did I mention what I think of git yet?

rofl0r:
you can use `git rebase -i master` in the terminal and get an interactive editor, remove the line containing the unwanted commit, and save. then you can force-push the fixed branch. btw i'd recommend to use the tool gitk to look at the commits if you're more the visual type.

sidneycadot:
I am definitely not the visual type - more of a command line type. Just not the git command line type.

sidneycadot:

colinleroy:
Thanks for fixing this and sorry, I'm the one who introduced the problem. I had no idea \n could be a different hex value accross platforms!

sidneycadot:
Yep the different character sets are a bit of a trap, and the end-of-line character is a problem in and of itself. Even today Windows has the \r \n combo, and some devices I encounter in my daytime job use a lone '\r' (13).

Great that it's fixed now, hopefully somebody will find time to fix the root issue at some point (the lack of character translation in the assembler).

GorillaSapiens:
#2562 #2565 #2571 #2581 all related.

GorillaSapiens:
see #2642

PR #2568, opened 2024-12-19
sim65: changes constant of the unstable "ANE" instruction.

sidneycadot:
This PR changes the constant of the unstable "ANE" instruction to comply with the 65x02 test suite.

ANE (0x8b) is an unstable illegal opcode that depends on a "constant" value that isn't really constant. It varies between machines, with temperature, and so on.

The original sim65 behavior was to use the constant value 0xEF. To get the behavior in line with the 65x02 test suite, we now use the value 0xEE instead, which is also a reasonable choice that can be observed in practice.

This eases the testing of sim65.

### DIFF
--- a/src/sim65/6502.c
+++ b/src/sim65/6502.c
@@ -969,8 +969,14 @@ static unsigned HaveIRQRequest;
     } while (0);
 
 /* ANE */
+/* An "unstable" illegal opcode that depends on a "constant" value that isn't
+ * really constant. It varies between machines, with temperature, and so on.
+ * Original sim65 behavior was to use the constant 0xEF here. To get behavior
+ * in line with the 65x02 testsuite, we now use the value 0xEE instead,
+ * which is also a reasonable choice that can be observed in practice.
+ */
 #define ANE(Val)                                                \
-    Val = (Regs.AC | 0xEF) & Regs.XR & Val;                     \
+    Val = (Regs.AC | 0xEE) & Regs.XR & Val;                     \
     Regs.AC = Val;                                              \
     TEST_SF (Val);                                              \
     TEST_ZF (Val)


sidneycadot:

rofl0r:
btw you might be interested in https://github.com/rofl0r/blargg-6502-cpu-test which exercises the "illegal" opcodes. in my emulator i also had to switch to 0xee to make it work.

sidneycadot:
Interesting... From my understanding, the "constant" (which isn't) depends on analog effects in the 6502 -- something like two drivers onto the same net, and which one wins depends on circumstances.

In related work, I'm writing a tool to test clock cycles of all different 6502 and 65C02 instructions, currently it runs on Ataris, the C64, and a Neo6502 board with a 65C02 board. Myself I'm not familiar with the NES, but I'm curious: do you know if there is a way on the NES to do cycle-exact timing of short instruction sequences?

One thing that tool is showing is that at least five undocumented instructions ($93, $9b, $9c, $9e, $9f) show behavior that I haven't seen described anywhere. In particular, the effective address where they store their result appears to behave non-deterministically, leading to situations where they clobber my (rather big) program, leading to crashes. I may want to look into your code to see if there's any hint there on what could be going on.

(With "non-deterministically", I really mean it: on real hardware, what these opcodes actually do can change from one run to the next, using the same program.)

rofl0r:
> do you know if there is a way on the NES to do cycle-exact timing of short instruction sequences?

nope, sorry. i suppose you can use similar tricks as the ones that were already described in earlier issues.

> In particular, the effective address where they store their result appears to behave non-deterministically.

i never had the luxury of exercising these on real hw, but when writing my cpu emulator all non-determinism disappeared once i stepped often enough through visual6502 to figure out what's going on, and after [implementing](https://github.com/rofl0r/jg6502) these behaviours, the blargg suite was happy - and i suppose it has been tested on the real NES by its author.

sidneycadot:
Ok. As good as visual6502 is, it is a strictly digital simulation, and they have to make choices when the behavior is "non-digital". Their FAQ says:

> In our simulation, if a transient short occurs from power to ground, ground wins.  This may happen in our iterative solver as the simulation state is converging.  Also, when two floating regions are switched together, their final voltage is taken from the region with the most components.  This is a simple but completely non-physical way to account for capacitance.

Stuff like this is why I like to run test programs running on real hardware, if at all possible. Fortunately, my trusty old Atari 800 XL is still going strong :)

mrdudz:
BTW when i tested those undocumented opcodes... you'll see differences to the real CPU exactly with those so called "unstable" ones :) In most other cases its an excellent reference though - but anything that relies on "floating" or "non driven" bus behaviour will not work correctly.

PR #2569, opened 2024-12-19
sim65: tighten 6502 register types

sidneycadot:
After a lot of preparatory work, we are now in position to finally tighten the types of the 6502 registers defined in the CPURegs struct of sim65.

All registers were previously defined as bare 'unsigned', leading to subtle bugs where the bits beyond the 8 or 16 "true" bits in the register could become non-zero. Tightening the types of the registers to uint8_t and uint16_t as appropriate gets rid of these subtle bugs once and for all, assisted by the semantics of C when assigning an unsigned value to an unsigned type with less bits: the high-order bits are simply discarded, which is precisely what we'd want to happen.

This change cleans up a lot of spurious failures of sim65 against the 65x02 test-set. For the 6502 and 65C02, we're now *functionally* compliant. For timing (i.e., clock cycle counts for each instruction), some work remains.

### DIFF
--- a/src/sim65/6502.c
+++ b/src/sim65/6502.c
@@ -664,15 +664,12 @@ static unsigned HaveIRQRequest;
 
 
 /* ADC, binary mode (6502 and 65C02) */
-/* TODO: once the Regs fields are properly sized, get rid of the
- * "& 0xff" in the Regs.AC asignment.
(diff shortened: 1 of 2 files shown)


sidneycadot:

Issue #2566, opened 2024-12-19
cc65: field alignment bug


sidneycadot:
Hi,

I encountered a strange problem in CC65 that appears to be an interplay between structs, pointers,  and unsigned longs.

The smallest case I can come up with that demonstrates the problem is below:

```

#include <stdio.h>

typedef struct {
    unsigned long Field1;
    unsigned long Field2;
} RecordType;

typedef struct {
    char dummy; // -- Offsets the Rec field by 1 byte.
    RecordType Rec;
} StructTypeA;

typedef struct {
    RecordType Rec;
} StructTypeB;

int main(void)
{
    StructTypeA A;
    StructTypeB B;
    StructTypeA * Aptr;

    A.Rec.Field1 = 0x11111111;
    A.Rec.Field2 = 0x22222222;

    Aptr = &A;
    B.Rec = Aptr->Rec;

    printf("%lx, %lx\n", B.Rec.Field1, B.Rec.Field2);

    return 0;
}
```

Expected output (also what GCC shows with the same code):

```
11111111, 22222222
```

Actual output with CC65 (using sim65, which initializes unused memory to 0xff):

```
111111ff, 22222211
```

This is some offset error introduced by the 'dummy' field in StructTypeA. I tried to make the program smaller but it's not so easy, with the most obvious attempts the issue disappears. Perhaps someone else has more luck to come up with a smaller demonstration?

sidneycadot:
Ok, the issue was introduced 4 years ago with a pretty small commit: https://github.com/cc65/cc65/commit/c3a6b399456937093eda9994f19b7f722731528d

That's the good news. The bad news is that I don't understand the compiler internals enough to dig any further.

Pinging @acqn and @oliverschmidt who authored resp merged that specific commit.

oliverschmidt:
I'm sorry if merged a regression :flushed:. However, I'm not maintaining cc65 anymore so I can't be of any help here.

sidneycadot:
Okay thanks for the heads-up.

Last commit activity from @acqn was in February 2024, let's hope he shows up. Otherwise, we'll need to find something willing and able to dig into this.

mrdudz:
unfortunately @acqn seems to be MIA :(

mrdudz:
[...]

PR #2570, opened 2024-12-20
Test to demonstrate issue #2566.

sidneycadot:


sidneycadot:
Self-merging, as a review doesn't seem to be incoming.

PR #2573, opened 2024-12-22
sim65: properly implement 5 'illegal' 6502X opcodes.

sidneycadot:
This PR fixes the implementation of 5 illegal opcodes in the 6502, which the 6502X supports:

* $93   SHA (zp),y
* $9B   TAS abs,y
* $9C   SHY abs,x
* $9E   SHX abs,x
* $9F   SHA abs,y

The common denominator of the previous implementation was that it didn't correctly handle the case when the Y or X indexing induced a page crossing. In those cases, the effective address calculation of the instructions becomes truly messed up (with the high byte of the address equal to the value being written).

The correctness of the implementations in this PR was verified using the 65x02 test suite, and corresponds to a (detailed) reading of the "No More Secrets" document.

Stylistically, there is room for improvement in these implementations, specifically in factoring out common behavior into macros. However, for now, the "explicit" coding style will suffice. It is clear enough, and we want to reach a situation soon where the sim65 code is able to pass the full '65x02' testsuite. Once we get to that point, we can refactor this code with a lot more confidence, since we will have the benefit of a working exhaustive test to make sure we don't break stuff.

### DIFF
--- a/src/sim65/6502.c
+++ b/src/sim65/6502.c
@@ -2534,7 +2534,20 @@ static void OPC_65SC02_92 (void)
 static void OPC_6502_93 (void)
 /* Opcode $93: SHA (zp),y */
 {
-    STO_CB (ZPINDY, SHA);
+    ++Regs.PC;
+    uint8_t zp_ptr_lo = MemReadByte(Regs.PC);
+    ++Regs.PC;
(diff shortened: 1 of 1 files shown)


sidneycadot:
Note: I plan to merge these changes to master in two days if I get nobody's feedback before that time.

PR #2576, opened 2024-12-24
sim65: Fix instruction timings for 6502 and 65C02.

sidneycadot:
This PR fixes all discrepancies of sim65 instruction timings, for both the 6502 and the 65C02 processors.

The timings as implemented in this PR have been verified against actual hardware (Atari 800 XL for 6502; and WDC 65C02 for 65C02).

These timings can also be verified against the 65x02 test suite. However, in this case, a single discrepancy arises; the 65x02 testsuite suggests that the 65C02 opcode 0x5c should take 4 clocks. However, tests on a hardware 65C02 have conclusively shown that this instruction takes 8 clock cycles. The 8 clock cycles duration for the 65C02 0xfc opcode is also confirmed by other sources, e.g. Section 9 of http://www.6502.org/tutorials/65c02opcodes.html.

This test makes sim65 correct both in terms of functionality (all opcodes now do what they do on hardware) and in terms of timing (all instructions take as long as they would on real hardware).

The one discrepancy that remains, is that on a real 6502/65C02, some instructions issue R or W cycles on the bus while the instruction processing is being done. Those spurious bus cycles are not replicated in sim65. Sim65 is thus an instruction-level simulator, rather than a bus-cycle level simulator. In other words, while the clock cycle counts for each instruction are now correct, not all clock cycles are individually simulated.

### DIFF
--- a/src/sim65/6502.c
+++ b/src/sim65/6502.c
@@ -579,6 +579,10 @@ static unsigned HaveIRQRequest;
 #define STO_CY_ZPINDY   6
 #define STO_CY_ZPIND    5
 
+#define STO_CY_ABSX_NP    STO_CY_ABSX
+#define STO_CY_ABSY_NP    STO_CY_ABSY
+#define STO_CY_ZPINDY_NP  STO_CY_ZPINDY
+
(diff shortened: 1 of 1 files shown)


sidneycadot:
I intend to merge this PR somewhere in the next few days into master, unless objections are raised.

PR #2577, opened 2024-12-24
sim65: add memory-mapped cycle/instruction/wallclock timer.

sidneycadot:
This PR adds timer functionality to sim65. It originates from feature request #2355.

The PR implements a functional memory-mapped timer, but not yet ready to be merged to master. Before doing that, we will need:

- [ ] tests
- [x] updated documentation
- [x] ensure that it builds on Windows.
- [ ] review

This is reviewed as a draft PR at this time to tickle the CI build for Windows and see what happens. The most obvious thing that may trip up the Windows build is the use of the 'clock_gettime' function in peripherals.c.

### DIFF
--- a/src/sim65.vcxproj
+++ b/src/sim65.vcxproj
@@ -86,6 +86,7 @@
     <ClInclude Include="sim65\error.h" />
     <ClInclude Include="sim65\memory.h" />
     <ClInclude Include="sim65\paravirt.h" />
+    <ClInclude Include="sim65\peripherals.h" />
   </ItemGroup>
   <ItemGroup>
     <ClCompile Include="sim65\6502.c" />
@@ -93,6 +94,7 @@
     <ClCompile Include="sim65\main.c" />
     <ClCompile Include="sim65\memory.c" />
     <ClCompile Include="sim65\paravirt.c" />
+    <ClInclude Include="sim65\peripherals.c" />
   </ItemGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
   <ImportGroup Label="ExtensionTargets">
--- a/src/sim65.vcxproj
+++ b/src/sim65.vcxproj
@@ -82,19 +82,19 @@
     </Link>
   </ItemDefinitionGroup>
   <ItemGroup>
-    <ClInclude Include="sim65\peripherals.h" />
     <ClInclude Include="sim65\6502.h" />
     <ClInclude Include="sim65\error.h" />
     <ClInclude Include="sim65\memory.h" />
     <ClInclude Include="sim65\paravirt.h" />
+    <ClInclude Include="sim65\peripherals.h" />
   </ItemGroup>
   <ItemGroup>
-    <ClInclude Include="sim65\peripherals.c" />
     <ClCompile Include="sim65\6502.c" />
     <ClCompile Include="sim65\error.c" />
     <ClCompile Include="sim65\main.c" />
     <ClCompile Include="sim65\memory.c" />
     <ClCompile Include="sim65\paravirt.c" />
+    <ClInclude Include="sim65\peripherals.c" />
   </ItemGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
   <ImportGroup Label="ExtensionTargets">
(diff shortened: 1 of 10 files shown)


sidneycadot:
Okay, useful information: both the Linux and the Windows builds fail.

The 'clock_gettime' is indeed the curlpit. It makes the build fail on Windows, and interestingly it turns out that the Linux CI build process also includes a build step for windows, via cross-compilation. So the Linux CI build fails as well.

I will now (temporarily) remove the use of clock_gettime() from the PR, to check if this is the only problem.

sidneycadot:
Okay disabling the reference to 'clock_gettime' fixes the Linux build, but the Windows build still fails.

It turns out that the Windows CI build depends on the vcxproj files that are littering the src directory, and they list C files and header files that need to be processed in the build.

sidneycadot:
Right, I hit a brick wall. The Windows CI builds appear to depend on opaque "vcxproj" files, which somehow leads to the linker not finding the new "peripherals.c" file.

Visual Studio project files have been added to the CC65 source tree at some point in the past -- files that are not intended for humans to inspect or change. I think that is a bad idea, and I added an issue about it https://github.com/cc65/cc65/issues/2578

This is a blocking issue for making progress on this PR.

Further work will have to wait until somebody shows up who knows how to deal with these Windows-specific build system issues. Specifically, how to add a file to be compiled and linked to one of the cc65 tools. It's pretty sucky that it isn't immediately obvious how to do that.

src/sim65.vcxproj
```diff
@@ -86,15 +86,17 @@
...
     <ClInclude Include="sim65\paravirt.h" />
+    <ClInclude Include="sim65\peripherals.h" />
   </ItemGroup>
   <ItemGroup>
     <ClCompile Include="sim65\6502.c" />
     <ClCompile Include="sim65\error.c" />
     <ClCompile Include="sim65\main.c" />
     <ClCompile Include="sim65\memory.c" />
     <ClCompile Include="sim65\paravirt.c" />
+    <ClInclude Include="sim65\peripherals.c" />
```
matpompili, line 97: Perhaps `ClCompile` instead of `ClInclude`?

```suggestion
    <ClCompile Include="sim65\peripherals.c" />
```

👋  @sidneycadot you appeared in my GitHub news feed, couldn't help but poke around :)

sidneycadot: If this works I owe you a beer :-)

sidneycadot: All hail the power of a fresh pair of eyes! 🍺🍺🍺

sidneycadot:
Thanks to a sharp eye the strange lack of inclusion of peripherals.c in the Visual Studio build was resolved, yay! Thanks @matpompili  :-) Turns out I made a silly mistake editing sim65.vcxproj. The code compiles once again under windows.

Next up is an attempt to compile the code with clock_gettime() re-enabled. This is expected to fail on the Windows build, but the error messages will hopefully teach us something.

sidneycadot:
Ok we now have working code that pleases both the Linux and Windows workflow builds. Gotta love CI/CD, good grief.

sidneycadot:
As per the discussion in #2579, I will not be providing tests for the timer infrastructure that fit neatly in the cc65 test infrastructure.

Note that the functionality in the PR is currently working just fine. I have a pretty good test implemented; it would just take work to fit it into CC65's test philosophy that I am not willing to do for reasons outlined in #2579.

An important question is if this PR will be accepted without an accompanying test fitting in the cc65 testing infrastructure. I myself would be fine with it, because I know it works; but I'll leave this as a policy decision to @mrdudz and perhaps others.

If that's a firm 'no', let's conclude that now, because then we can simply reject the PR rather than continue working on it.

sidneycadot:
I added documentation for the counter peripheral to doc/sim65.html.

As far as I am concerned, the PR is now ready for review. It could even be accepted as-is, without tests, if that is deemed acceptable. So I'm removing the draft status, and defer to @mrdudz and others to see if any more work is necessary to merge this feature.

mrdudz:
Looks OK to me

sidneycadot:
Ok, then I merge. In the nick of time in 2024 :-)

PR #2587, opened 2025-01-01
Rewrite rewind() in assembly,add lseek() to paravirt

colinleroy:


sidneycadot:
Looks like a useful bit of functionality.

Personally I'm not very fond of the "paravirtualization" machinery in sim65 and perhaps some day I will think of a better way, but this addition will not make that more difficult :)

colinleroy:
ping @mrdudz (it seems like this one did not get anyone's attention) :)

PR #2589, opened 2025-01-02
sim65: Add execution tracing and sim65 control peripheral

sidneycadot:
This PR adds three related features to sim65:

* Execution tracing (--trace option for sim65 executable).
* CPU model override (--cpu <type> option for sim65 executable).
* Inspectability/controlability of SIM65 behavior from within a running sim65 program, through memory mapped addresses.

Currently, the live SIM65 behavior monitoring and control concerns the aforementioned CPU mode and the tracing mode.

To assess the PR, you are cordially invited to read the updated documentation, temporary hosted at: http://data.jigsaw.nl/sim65.html (Specifically: see updated sim65 options, and the new Section 7).

PR #2590, opened 2025-01-02
sim65: add target-specific C header file and asm include file.

sidneycadot:
This PR adds three related features to sim65:

* Execution tracing (--trace option for sim65 executable).
* CPU model override (--cpu option for sim65 executable).
* Inspectability/controlability of SIM65 behavior from within a running sim65 program, through memory mapped addresses.

To assess the PR, you are cordially invited to read the updated documentation, temporary hosted at: http://data.jigsaw.nl/sim65.html (Specifically: see updated sim65 options, and the new Section 7).

This PR also adds a C header file (sim65.h) and an assembly include file (sim65.inc) that define convenient access to the new sim65 peripherals from C and assembly.

To implement these, a new export symbol "_peripherals" was added to the linker configuration files sim6502.cfg and sim65c02.cfg, corresponding to the base address of the peripheral memory aperture.

The C header file defines a structure rooted at _peripherals, meaning it is possible to write things like

```
    peripherals.counter.latch = 0;
```

... and have it translated to the most efficient code possible.

The assembly include file defines a number of symbols to access registers. In assembly, the above statement would correspond to

```
    lda #0
    sta peripheral_counter_latch
```

Apart from the addresses of the peripheral registers, both the C header file and assembly include file also provide a bunch of constants that are useful for readability.

Lastly, the C header file provides TRACE_ON() and TRACE_OFF() macros that make it very easy to enable processor instruction tracing from C.

### DIFF
--- a/src/sim65/6502.c
+++ b/src/sim65/6502.c
@@ -42,14 +42,16 @@
  * the WAI ($CB) and STP ($DB) instructions are unsupported.
  */
 
-#include <stdbool.h>
 #include <stdint.h>
+#include <stdbool.h>
 
 #include "memory.h"
 #include "peripherals.h"
 #include "error.h"
-#include "6502.h"
 #include "paravirt.h"
+#include "trace.h"
+
+#include "6502.h"
 
 /*
 
@@ -4485,7 +4487,7 @@ static const OPFunc OP65C02Table[256] = {
     OPC_6502_41,
     OPC_65C02_NOP22,    // $42
     OPC_65C02_NOP11,    // $43
-    OPC_6502X_44,        // $44
+    OPC_6502X_44,       // $44
     OPC_6502_45,
     OPC_6502_46,
     OPC_65C02_47,
@@ -4730,6 +4732,10 @@ unsigned ExecuteInsn (void)
     /* If we have an NMI request, handle it */
     if (HaveNMIRequest) {
 
+        if (TraceMode != TRACE_DISABLED) {
+            PrintTraceNMI ();
+        }
+
         HaveNMIRequest = false;
         Peripherals.Counter.NmiEvents += 1;
 
@@ -4745,6 +4751,10 @@ unsigned ExecuteInsn (void)
         Cycles = 7;
 
     } else if (HaveIRQRequest && GET_IF () == 0) {
+
+        if (TraceMode != TRACE_DISABLED) {
+            PrintTraceIRQ ();
+        }
 
         HaveIRQRequest = false;
         Peripherals.Counter.IrqEvents += 1;
@@ -4765,11 +4775,16 @@ unsigned ExecuteInsn (void)
         /* Normal instruction - read the next opcode */
         uint8_t OPC = MemReadByte (Regs.PC);
 
-        /* Execute it */
-        Handlers[CPU][OPC] ();
+        /* Print a trace line, if trace mode is enabled. */
+        if (TraceMode != TRACE_DISABLED) {
+            PrintTraceInstruction ();
+        }
 
-        /* Increment the instruction counter by one.NMIs and IRQs are counted separately. */
+        /* Increment the instruction counter by one. */
         Peripherals.Counter.CpuInstructions += 1;
+
+        /* Execute the instruction. The handler sets the 'Cycles' variable. */
+        Handlers[CPU][OPC] ();
     }
 
     /* Increment the 64-bit clock cycle counter with the cycle count for the instruction that we just executed. */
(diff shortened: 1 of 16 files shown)


sidneycadot:
This PR is technically independent from #2589, but in reality it depends on it, since it presumes that the functionality introduced there is, in fact, available as memory-mapped registers.

I propose to merge this PR shortly after #2589. (I will do so myself unless feedback is received that will require attention.)

When #2589 and this PR are accepted, the sim65 documentation will be updated to reflect the new way of using the peripherals afforded by these changes.

src/sim65/6502.c
```diff
@@ -4485,7 +4487,7 @@ static const OPFunc OP65C02Table[256] = {
     OPC_6502_41,
     OPC_65C02_NOP22,    // $42
     OPC_65C02_NOP11,    // $43
-    OPC_6502X_44,        // $44
+    OPC_6502X_44,       // $44
```
mrdudz, line 4490: no c++ comments please :)

mrdudz:
Would be nice to also have a sample that uses the header file

src/sim65/6502.c
sidneycadot: These were already there for a long long time. I'm fine with changing them to comply with the style convention, but should be done in a different sweep.

I do think it's a curious convention. Calling them "C++ comments" 30 years after they were added to C is also somewhat, uhm, ideosyncratic :-)

sidneycadot:
> Would be nice to also have a sample that uses the header file

Yep. The docs have a few examples. As soon as the dust settles I will put a few examples in samples/sim65/ , no worries.

sidneycadot:
@mrdudz 

Seems I accidentally botched a git command. I branched this PR off of the branch I prepared as #PR2589, rather than master as intended. 2589 was intended to implement the functionality, with this PR (2590) adding support from C and asm.

I think the most pragmatic thing to do would be to merge this PR if you like what you see, and ditch #2589. Otherwise I will have to do git carpentry that is not really compliant with my level of git mastery (so nicely demonstrated by this little mistake). I would like to avoid that.

src/sim65/6502.c
groessler: It's not c89, fwiw.

sidneycadot: I think a conversation needs to be had at some point about the coding style conventions, specifically the odd choice to aim for some weird mix of C89 and C99. It seems to be borne from taste rather than rationality.

sidneycadot:
I added three samples, in samples/sim65, that show the new functionality: fancy tracing under runtime control, runtime CPU model switching, and using the timer peripheral.

As of now, no Makefile. I did look at some sibling directories, but (like with the tests), the sample makefiles are neither trivial nor documented.

sidneycadot:
I will retract PR #2589 in favor of this one, which combines the peripheral feature (including the timer) and its support files for using the features from C or assembly.

mrdudz:
> I think the most pragmatic thing to do would be to merge this PR if you like what you see, and ditch #2589. Otherwise I will have to do git carpentry that is not really compliant with my level of git mastery (so nicely demonstrated by this little mistake). I would like to avoid that.

please fix it - like this various unrelated things are mixed in the PR, and it will cause headache in the future

(please also don't forget to tag your PRs - for the same reason)

sidneycadot:
The only way I am able  fix it is to throw away everything and prepare new PRs.

As I said, I am not good with git.

mrdudz:
> The only way I am able fix it is to throw away everything and prepare new PRs.
> 
> As I said, I am not good with git.

Same here - i have to look up how stuff works every time i do something that isn't covered by git gui.

However, the git history is funny enough already due to the sloppy way we deal with merging in this project, so please lets not make it worse than it already is :)

sidneycadot:
I never understand people's fascination with the revision history of their repositories. It's a tool, not a novel.

But okay, if you insist. I will retract this PR and re-submit it in a few parts.

mrdudz:
> I never understand people's fascination with the revision history of their repositories. It's a tool, not a novel.

Oh sure. Which is why the history in this project is what it is :) Its no reason to mess it up even more though :)

rofl0r:
your best tools for the task at hand are git rebase -i to squash relevant commits together (i usually use the fixup option), and then either export the cleaned up history as patches (git format-patch) and reapply them in a fresh branch using git am, or use git cherry-pick.

sidneycadot:
I have given up on ever going becoming proficient in git. I am sure the magic incantations exists that can get you into and out of any situation, but as far as I am concerned life is too short.... Version management, while somewhat important, is too uninteresting for me to spend the time to dive in.

sidneycadot:
_wince_

PR #2592, opened 2025-01-05
Fix a field alignment bug

### LINKED ISSUE #2566: cc65: field alignment bug

kugelfuhr:
This fixes #2566. No test added since #2570 is already waiting for a merge.


sidneycadot:
Excellent, and a very small fix at that!

I don't understand the compiler internals, can you explain in simple terms what the bug was and how your patch fixes it?

sidneycadot:
Just now I merged #2570, which was a bit overdue.

If all goes well that merge should break the workflow of this PR if you rebase its branch to master. If my understanding of the test structure is correct (from test/readme.txt), that should be solved by moving test/todo/bug2566.c to test/val/bug2566.c.

kugelfuhr:
> I don't understand the compiler internals, can you explain in simple terms what the bug was and how your patch fixes it?

The error is in the part that calls `memcpy()` to copy structs. `memcpy()` needs two addresses and a size.

The `ExprDesc` struct contains all information about an expression. Depending on its contents, different steps have to be taken to generate the address of the expression.

One flag in the `ExprDesc` structure says that the primary register contains the address of the expression. This is tested with `ED_IsLocExpr()`. But there's a caveat: There may also be a fixed offset to the address in the primary. So
part of the address is in the primary and the other one is in the `ExprDesc` structure. The offset is kept separate since there are runtime routines that are able to handle this offset in an efficient manner, for example by loading it into Y
before accessing the data. This saves an addition in many cases.

The problem in the code was that it didn't handle the fixed offset when `ED_IsLocExpr()` returned true. In case of your sample program, the primary contained the address of the struct, and the offset field was 1. Since the
offset wasn't handled, `memcpy()` was effectively called with an address off by one.

The solution is to adjust the address in the primary if there is an additional fixed offset. I've changed the code to make that more explicit even if it is a few lines larger now. The first version was a bit hackish and used a side effect of `ED_AddrExpr()`.

sidneycadot:
Sounds good to me, thanks for the explanation.

sidneycadot:
Not sure if anybody is around who is good at judging the fix. I suppose if nobody shows up in the next few days, we could merge to master.

mrdudz:
I'm merging it now, it looks fine and doesnt break any tests so what could possibly happen... :)

PR #2591, opened 2025-01-03
sim65: add tracing, and a sim65 control peripheral for sim65 runtime control

sidneycadot:
This PR is the first of two PRs that replaces earlier PRs #2589 and #2590. Due to a git branching mishap it was decided to re-partition the new functionality in two sequential PRs that offer self-contained, new functionality to sim65.

The functionality in this first PR extends the sim65 simulator in the following ways:

* It provides tracing functionality, i.e., the possibility of printing one line of simulator state information per instruction executed.
* It provides a memory mapped "sim65 control" peripheral that allows control of (a) the tracing functionality, and (b) the cpu mode.
* It provides command-line options to sim65 to enable the tracing, and to override the CPU mode as specified in the program file header.

More detailed information and some discussion can be found in the discussions with the (now retracted) PRs #2589 and #2590.

This PR provides the technical infrastructure inside the sim65 simulator program itself. Once this PR is accepted, a follow-up PR will be posted that adds C and assembly-language support for the new tracing and peripheral features so they can be easily accessed from the CC65 compiler and the CA65 assembler; some examples; and the documentation for these features. The lack of the latter, in this pull request, will be addressed then.

Issue #2582, opened 2025-01-01
Insufficient documentation for the build system


sidneycadot:
CC65's build system comprises a number of interlocking parts:

* The Makefiles, mostly intended to run on Linux systems. This is the original build system for CC65, and it used to be all there was. They work on MacOS as well, and they also been patched to detect/work in Windows (mingw environment I think).
* A bunch of Microsoft compiler files in src/. These are useful for people who want to use Visual Studio I suppose, and they are also  used by the Windows-hosted build in Guthub Actions.
* Github actions: these comprise a bunch of yaml files in .github/workflow. They trigger Windows-hosted and Linux-hosted builds, with also some cross-compilation in place (eg a MinGW64 build on Linux for a Windows target).

The entire setup is not trivial. The problem is that, when adding new stuff that's more involved than a single testcase, you have to make sure to not break anything, and to conform to conventions that are not written down.

The build system (Makefiles, workflow files, and what else there is) are _code_, and as such have its goals, assumptions, and design written down somewhere. This is not currently available.

I think it would be useful if someone sat down to figure out precisely how everything works, and wrote a README or a chapter in the documentation that explains all assumptions and steps in each part of the build system; and also, how to add non-trivial new stuff.

mrdudz:
It would certainly be great if all of that happened. I wouldn't count on it ever happening though - unless you are the one who will do it :)

(All of what you mentioned is almost self-documenting compared to the mess that is VICE :D)

sidneycadot:
> I wouldn't count on it ever happening though

Well that just means it can go on the pile of 200 things that won't ever happen, right. And it will also mean that I won't be contributing non-trivial tests.

> All of what you mentioned is almost self-documenting compared to the mess that is VICE

I think "other projects do it worse" is not a great angle.

These changes were merged at some point, without documentation in place. If anything, it would be useful to consider what went wrong there.

mrdudz:
> I think "other projects do it worse" is not a great angle.

The point is... i don't think its THAT much of a problem. Sure, in an ideal world everything would be nicely documented - but it is not. The original cc65 site even used to have an explicit note saying something like "if in doubt i prefer programming over documenting" :) We have improved since then (hey we even have some sort of automatic regression testing now...), but there is certainly still some way to go until everything is nice and complete.

PR #2593, opened 2025-01-05
sim65: add C/assembly support, docs, and samples for the new peripheral functionality.

sidneycadot:
This PR is the second of two PRs that replaces earlier PRs #2589 and #2590. Due to a git branching mishap it was decided to re-partition the new functionality in two sequential PRs that offer self-contained, new functionality to sim65. The first one, #2591, was accepted earlier.

The functionality in this second and last PR provides the following things in relation to the new peripheral support:

* C support: there is now an include/sim65.h that can be included from C. It provides access to the memory-mapped peripheral addresses.

* Assembly support: there is now an asminc/sim65.inc that can be included from assembly. It provides symbolic labels for the memory-mapped peripheral addresses.

Note: the two items above are implemented by adding a "_peripherals" symbol to cfg/sim6502.cfg and cfg/sim65c02.cfg, with the fixed base address of the peripherals memory aperture (0xffc0).

* Updated the sim65 documentation to describe the peripherals in some detail, with examples that show to use the new features from within C.

* Some examples in the new samples/sim65/ directory. These are currently not integrated in the build system (in other words, there's no Makefile there), because I don't know how to do that. I will happily implement that after #2582 is taken care of.

If that is not acceptable, the next best thing will be for somebody else (who understands how the Makefiles are set up) to take care of this.

If that's not going to happen, and we don't want examples that are not properly integrated with the build system, there's always the option of removing these samples from the PR.

### DIFF
--- a/asminc/sim65.inc
+++ b/asminc/sim65.inc
@@ -1,0 +1,75 @@
+
+; *******************************************************************************
+; **                                                                           **
+; **  sim65.inc : assembler definitions for the sim6502 and sim65c02 targets.  **
+; **                                                                           **
+; **  Sidney Cadot, January 2025                                               **
+; **                                                                           **
+; *******************************************************************************
+
+            ; The '_peripherals' symbol is defined in the linker configuration
+            ; file to correspond to the first address in the periperal memory
+            ; aparture.
+            ;
+            ; We use it here as a base address for all peripheral addresses.
+
+            .import _peripherals
+
+; **************************************************************
+; **                                                          **
+; **  Define assembler symbols for the "counter" peripheral.  **
+; **                                                          **
+; **************************************************************
+
+peripheral_counter_base   := _peripherals + 0
+
+peripheral_counter_latch  := peripheral_counter_base + 0
+peripheral_counter_select := peripheral_counter_base + 1
+peripheral_counter_value  := peripheral_counter_base + 2
+
+; Values for the peripheral_counter_select register.
+
+COUNTER_SELECT_CLOCKCYCLE_COUNTER   = $00
+COUNTER_SELECT_INSTRUCTION_COUNTER  = $01
+COUNTER_SELECT_IRQ_COUNTER          = $02
+COUNTER_SELECT_NMI_COUNTER          = $03
+COUNTER_SELECT_WALLCLOCK_TIME       = $80
+COUNTER_SELECT_WALLCLOCK_TIME_SPLIT = $81
+
+; ********************************************************************
+; **                                                                **
+; **  Define assembler symbols for the "sim65 control" peripheral.  **
+; **                                                                **
+; ********************************************************************
+
+peripheral_sim65_base       := _peripherals + 10
+
+peripheral_sim65_cpu_mode   := peripheral_sim65_base + 0
+peripheral_sim65_trace_mode := peripheral_sim65_base + 1
+
+; Values for the peripheral_sim65_cpu_mode register.
+
+SIM65_CPU_MODE_6502  = $00
+SIM65_CPU_MODE_65C02 = $01
+SIM65_CPU_MODE_6502X = $02
+
+; Bitfield values for the peripheral_sim65_trace_mode field.
+
+SIM65_TRACE_MODE_FIELD_INSTR_COUNTER  = $40
+SIM65_TRACE_MODE_FIELD_CLOCK_COUNTER  = $20
+SIM65_TRACE_MODE_FIELD_PC             = $10
+SIM65_TRACE_MODE_FIELD_INSTR_BYTES    = $08
+SIM65_TRACE_MODE_FIELD_INSTR_ASSEMBLY = $04
+SIM65_TRACE_MODE_FIELD_CPU_REGISTERS  = $02
+SIM65_TRACE_MODE_FIELD_CC65_SP        = $01
+
+; Values for the peripheral_sim65_trace_mode field that fully disable / enable tracing.
+
+SIM65_TRACE_MODE_DISABLE              = $00
+SIM65_TRACE_MODE_ENABLE_FULL          = $7F
+
+; ************************
+; **                    **
+; **  End of sim65.inc  **
+; **                    **
+; ************************
--- a/cfg/sim6502.cfg
+++ b/cfg/sim6502.cfg
@@ -1,12 +1,15 @@
 SYMBOLS {
     __EXEHDR__:    type = import;
     __STACKSIZE__: type = weak, value = $0800; # 2k stack
+    _peripherals:  type = export, value = $FFC0;
 }
+
 MEMORY {
     ZP:     file = "",               start = $0000, size = $0100;
     HEADER: file = %O,               start = $0000, size = $000C;
-    MAIN:   file = %O, define = yes, start = $0200, size = $FDC0 - __STACKSIZE__;
+    MAIN:   file = %O, define = yes, start = $0200, size = $FFC0 - $0200 - __STACKSIZE__;
 }
+
 SEGMENTS {
     ZEROPAGE: load = ZP,     type = zp;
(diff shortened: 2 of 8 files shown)


sidneycadot:
Unless discussion ensues, I will self-merge this PR in three days.

kugelfuhr:
My suggestion would be to add tracing based on debug information instead of having to modify the code. Modifying the code has several disadvantages:
* It needs a recompile each time tracing is changed.
* When changing the code, the addresses in the output change, so tracing output is no longer comparable.
* When looking for a bug, the bug may disappear when the address layout changes.

Compare that to using debug info:
* No recompile necessary. Each run uses the same binary.
* No address layout changes.
* Change tracing output from the command line by specifying which function(s), module(s) or source line(s) you want to be traced.

When using the debug info for tracing, it will also open the door for further improvements like profiling based on functions or source lines.

sidneycadot:
Hi,

> My suggestion would be to add tracing based on debug information instead of having to modify the code.

I don't understand what you have in mind. Could you give an example?

FWIW: it is already possible to do tracing without changing the code, in fact I expect that to be the most usual way of using it, i.e, just do "sim65 --trace program". Enabling/disabling it from inside the 6502 program itself is mostly intended for rather niche cases.

In general, the sim65 executable doesn't have access to debugging symbols etc -- it just sees the flat memory image with a 12 byte header. There is currently no functionality in the simulator to use information outside of that. Doing that would be more of a job I'd associate with a debugger. That could certainly be build on top of sim65, or perhaps as a separate tool.

kugelfuhr:
> I don't understand what you have in mind. Could you give an example?
> 
> FWIW: it is already possible to do tracing without changing the code, in fact I expect that to be the most usual way of using it, i.e, just do "sim65 --trace program". Enabling/disabling it from inside the 6502 program itself is mostly intended for rather niche cases.

The use case for traces is in most cases debugging. But for debugging, the usefulness of "--trace" is actually very limited since you will get drowned in data that has nothing to do with the problem you are looking for. Remember that this is an 
execution trace, so if the program loops 100 times, you will get 100 times an almost identical output.

Say I've written a new library function and it doesn't work as expected. Using "--trace" will give me lots of output that has nothing to do with the actual problem. Using TRACE_ON/TRACE_OFF changes the program and may hide the bug I'm searching for.

The solution is to read the debug information file and allow the simulator to output a trace for exactly the code that I want to look at. It would look something like

```
sim65 --dbg-syms myprogram.dbg --trace myfunction myprogram.bin
```

This is much better than the approach implemented in this PR and allows tracing based on functions, modules, source lines or whatever. You can also support something like

```
--trace-dump my_variable
```

to output variables with each trace line. 

And, when support for debug symbols was added, you can also add profiling based on functions or source lines with little more effort.

Please note that I'm not saying that the stuff in the PR will not work. But I think it's a step in the wrong direction. Using the (already available!) debug information is superior in every respect and it opens the door for many more improvements in the future. So my suggestion would be to add "--trace" (as you did) but to remove the code for controlling tracing from within the program. It can be replaced in another step by code that reads the debug file as described above.

sidneycadot:
> The use case for traces is in most cases debugging.

I don't necessarily agree with that statement, but it's certainly a use-case.

> But for debugging, the usefulness of "--trace" is actually very limited since you will get drowned in data that has
> nothing to do with the problem you are looking for. Remember that this is an
> execution trace, so if the program loops 100 times, you will get 100 times an almost identical output.

Sure. As a Unix-minded person I tend to think that a few strategic seds and greps will quickly prune that down to something manageable, so from that perspective the usefulness would not "actually be very limited", as you put it. But I can certainly agree that there are situations where the ability to trim down / preselect on the command line would be useful.

> The solution is to read the debug information file and allow the simulator to output a trace for exactly the code that I want to look at. It would look something like
> 
> sim65 --dbg-syms myprogram.dbg --trace myfunction myprogram.bin

What would be the semantics of this? For one thing, "myfunction" doesn't have a size, it's just a symbol.

Also I personally wouldn't use such an interface at all, probably. I'm more of a printf-debugging person anyway. We're having a discussion that's pretty much hinging on personal taste, I think.

> This is much better than the approach implemented in this PR and allows tracing based on functions, modules, source lines or whatever.

It would be powerful for sure. It's also a huge amount of work that I am not going to do.

> You can also support something like [...]

And I can think of many other things you could implement, yes.

I think we're looking at this differently. To me, sim65 is first and foremost a CC65 target and a virtual machine to run 6502 code on.

The features you're describing would normally be associated with a debugger. If you want to implement a proper, full-fledged debugger, you could implement that (either inside of sim65, or outside, as a program that somehow hooks into sim65 as it is running). Myself, I don't have the need for that, but it could be done. The "--trace" facility, and specifically the runtime control-ability, is a limited subset of what you have in mind. You may think it's not very useful, and that's fine -- different strokes for different blokes. For me it addresses close to 100% of my needs.

> And, when support for debug symbols was added, you can also add profiling based on functions or source lines with little more effort.

I think it is important to first have a discussion on whether you want those kinds of features in sim65. One of the charming things of the program at this point is that it's pretty simple and straightforward. And the second thing is that adding debugging symbol support is most certainly non-trivial.

> Please note that I'm not saying that the stuff in the PR will not work.

Well good, because it does. Note that the feature itself is already in master (that was PR #2592). This PR is about documentation, support files for C and assembly, and a few examples.

> But I think it's a step in the wrong direction.

Well there we disagree. Your vision for sim65 appears to be that it should gain a lot of capabilities, such as debugging and profiling. Previously, you advocated to put support for the 65x02 testsuite tests inside the executable, if memory serves.

I, on the other hand, would prefer to keep sim65 lean and mean. And to provide features like testing, a debugger, and a profiler separately, to the maximum extent possible. Pretty much analogous to gprof and gperf being separate, dedicated tools in the unix world.

> Using the (already available!) debug information

That's a weird thing to say. The information is most certainly not available in sim65 at the moment. It would require a significant effort to make sim65 capable of ingesting and using debug information in a useful way. It's not something that I am going to do, because I don't personally see a net positive benefit compared to the effort it would take.

> So my suggestion would be to add "--trace" (as you did) but to remove the code for controlling tracing from within the program. It can be replaced in another step by code that reads the debug file as described above.

Well, I'm not going along with that, because I'm not going to implement it. In time, you may want to implement these features yourself. And also, I do think that having the ability to enable/disable tracing from within the 6502 program is pretty useful and powerful -- given that you can, for example, enable/disable tracing based on runtime information / conditions. Sure, full fledged debuggers can also do that, but I'd wager a serious amount of money that cc65 will never get a debugger at that level of functionality.

Having said all that, the fact that we disagree on this makes me curious about the opinion of other people on this matter, so I won't self-merge within the next few days, until other people have had their say.

The weird thing is, we are now discussing a feature that's already merged, in a PR that is mostly about documentation and examples of that feature. So it's a bit late to decide to retract this feature, but if other people chime in and dislike it (or specific subsets of it, like the TRACE_ON/TRACE_OFF functionality) as much as you do, it won't be difficult to rip out the feature. If a consensus appears that would want that to happen, I will be happy to do the necessary surgery.

mrdudz:
> Having said all that, the fact that we disagree on this makes me curious about the opinion of other people on this matter, so I won't self-merge within the next few days, until other people have had their say.

I don't have a strong opinion on this - mostly because, to me, the main purpose and use case for sim65 is the regression testing of the compiler (and i have not, and probably will never, used it for anything besides that)

But i agree about keeping it simple

sidneycadot:
Okay for now I'm merging this PR. Apart from @kugelfuhr and @mrdudz I don't expect more comments, and this is essentially about documentation for an already-present feature anyway.

If anywone wants to yank this feature or turn it into something much more powerful as @kugelfuhr suggests, the tracing stuff is pretty self-contained and could be augmented or replaced at some time in the future. But that would be the subject of another PR or issue, as far as I am concerned.

Issue #2609, opened 2025-02-24
charmap translation violates C specification 6.4.4.4 Character constant


GorillaSapiens:
note that for both octal and hex escapes, the specification says this should be the literal value of the desired character.
this is independent and separate from the single character escapes (like "\n") outlined in "5.2.2 Character display semantics"

cc65 on a PETSCII machine will map both "\n" and "\x0A" to "\x0D".  this is incorrect.  ONLY the "\n" should get mapped.

====

6.4.4.4 Character constant

867 The octal digits that follow the backslash in an octal escape sequence are taken to be part of the construction of a single character for an integer character constant or of a single wide character for a wide character constant.

868 The numerical value of the octal integer so formed specifies the value of the desired character or wide character.

869 The hexadecimal digits that follow the backslash and the letter x in a hexadecimal escape sequence are taken to be part of the construction of a single character for an integer character constant or of a single wide character for a wide character constant.

870 The numerical value of the hexadecimal integer so formed specifies the value of the desired character or wide character.

====

// compile with "cc65 -t c64 test.c"

char backslash_n = '\n';   // should be PETSCII newline character 0x0D
char hex_0A      = '\x0a'; // should be literal byte 0x0A
char octal_012   = '\012'; // should also be 0x0A
int  int_hex_0A      = '\x0a'; // should be literal int 0x000A
int  int_octal_012   = '\012'; // should also be 0x000A

;
; File generated by cc65 v 2.19 - N/A
;
        .fopt           compiler,"cc65 v 2.19 - N/A"
        .setcpu         "6502"
        .smart          on
        .autoimport     on
        .case           on
        .debuginfo      off
        .importzp       sp, sreg, regsave, regbank
        .importzp       tmp1, tmp2, tmp3, tmp4, ptr1, ptr2, ptr3, ptr4
        .macpack        longbranch
        .export         _backslash_n
        .export         _hex_0A
        .export         _octal_012
        .export         _int_hex_0A
        .export         _int_octal_012

.segment        "DATA"

_backslash_n:
        .byte   $0D
_hex_0A:
        .byte   $0D
_octal_012:
        .byte   $0D
_int_hex_0A:
        .word   $000D
_int_octal_012:
        .word   $000D

sidneycadot:
I agree that cc65 isn't compliant here, and that fixing this will be difficult.

@GorillaSapiens : for reference, which version of "the specification" you are citing?

GorillaSapiens:
[...]

PR #2610, opened 2025-02-25
conform to 6.4.4.4 for hex and octal escapes

GorillaSapiens:
fixes issue #2609 

// compile with "cc65 -t c64 test.c"

char backslash_n = '\n';   // should be PETSCII newline character 0x0D
char hex_0A      = '\x0a'; // should be literal byte 0x0A
char octal_012   = '\012'; // should also be 0x0A
int  int_hex_0A      = '\x0a'; // should be literal int 0x000A
int  int_octal_012   = '\012'; // should also be 0x000A
char char_ten        = 10; // should be 0x0A
int  int_ten         = 10; // should be 0x000A

const char doubled[] = "\n\x0a"; // should be 0x0D 0x0A

;
; File generated by cc65 v 2.19 - N/A
;
        .fopt           compiler,"cc65 v 2.19 - N/A"
        .setcpu         "6502"
        .smart          on
        .autoimport     on
        .case           on
        .debuginfo      off
        .importzp       sp, sreg, regsave, regbank
        .importzp       tmp1, tmp2, tmp3, tmp4, ptr1, ptr2, ptr3, ptr4
        .macpack        longbranch
        .export         _backslash_n
        .export         _hex_0A
        .export         _octal_012
        .export         _int_hex_0A
        .export         _int_octal_012
        .export         _char_ten
        .export         _int_ten
        .export         _doubled

.segment        "DATA"

_backslash_n:
        .byte   $0D
_hex_0A:
        .byte   $0A
_octal_012:
        .byte   $0A
_int_hex_0A:
        .word   $000A
_int_octal_012:
        .word   $000A
_char_ten:
        .byte   $0A
_int_ten:
        .word   $000A

.segment        "RODATA"

_doubled:
        .byte   $0D,$0A,$00

### DIFF
--- a/src/cc65/scanner.c
+++ b/src/cc65/scanner.c
@@ -163,6 +163,12 @@ static const struct Keyword {
 typedef uint32_t scan_t;
 
 
+/* ParseChar return values */
+typedef struct {
+    int Val;
+    int Cooked;
+} parsedchar_t;
+
 /*****************************************************************************/
 /*                                   code                                    */
 /*****************************************************************************/
@@ -326,12 +332,15 @@ static void SetTok (int tok)
 
 
 
-static int ParseChar (void)
+static parsedchar_t ParseChar (void)
 /* Parse a character token. Converts escape chars into character codes. */
 {
+    parsedchar_t Result;
     int C;
     int HadError;
     int Count;
+
+    Result.Cooked = 1;
 
     /* Check for escape chars */
     if (CurC == '\\') {
@@ -373,6 +382,7 @@ static int ParseChar (void)
             case 'x':
             case 'X':
                 /* Hex character constant */
+                Result.Cooked = 0;
                 if (!IsXDigit (NextC)) {
                     Error ("\\x used with no following hex digits");
                     C = ' ';
@@ -401,6 +411,7 @@ static int ParseChar (void)
             case '6':
             case '7':
                 /* Octal constant */
+                Result.Cooked = 0;
                 Count = 1;
                 C = HexVal (CurC);
                 while (IsODigit (NextC) && Count++ < 3) {
@@ -423,7 +434,12 @@ static int ParseChar (void)
     NextChar ();
 
     /* Do correct sign extension */
-    return SignExtendChar (C);
+    Result.Val = SignExtendChar(C);
+    if (Result.Cooked) {
+        Result.Cooked = Result.Val;
+    }
+
+    return Result;
 }
 
 
@@ -431,7 +447,7 @@ static int ParseChar (void)
 static void CharConst (void)
 /* Parse a character constant token */
 {
-    int C;
+    parsedchar_t C;
 
     if (CurC == 'L') {
         /* Wide character constant */
@@ -457,7 +473,8 @@ static void CharConst (void)
     }
 
     /* Translate into target charset */
-    NextTok.IVal = SignExtendChar (C);
+    NextTok.IVal = SignExtendChar (C.Val);
+    NextTok.Cooked = C.Cooked;
 
     /* Character constants have type int */
     NextTok.Type = type_int;
@@ -468,6 +485,9 @@ static void CharConst (void)
 static void StringConst (void)
 /* Parse a quoted string token */
 {
+    /* result from ParseChar */
+    parsedchar_t ParsedChar;
+
     /* String buffer */
     StrBuf S = AUTO_STRBUF_INITIALIZER;
 
@@ -494,7 +514,8 @@ static void StringConst (void)
             Error ("Unexpected newline");
             break;
         }
-        SB_AppendChar (&S, ParseChar ());
+        ParsedChar = ParseChar ();
+        SB_AppendCharCooked(&S, ParsedChar.Val, ParsedChar.Cooked);
     }
 
     /* Skip closing quote char if there was one */
@@ -689,6 +710,7 @@ static void NumericConst (void)
 
         /* Set the value and the token */
         NextTok.IVal = IVal;
+        NextTok.Cooked = 0;
         NextTok.Tok  = TOK_ICONST;
 
     } else {
@@ -805,7 +827,12 @@ static void GetNextInputToken (void)
         if (NextTok.Tok == TOK_SCONST || NextTok.Tok == TOK_WCSCONST) {
             TranslateLiteral (NextTok.SVal);
         } else if (NextTok.Tok == TOK_CCONST || NextTok.Tok == TOK_WCCONST) {
-            NextTok.IVal = SignExtendChar (TgtTranslateChar (NextTok.IVal));
+            if (NextTok.Cooked) {
+                NextTok.IVal = SignExtendChar (TgtTranslateChar (NextTok.IVal));
+            }
+            else {
+                NextTok.IVal = SignExtendChar (NextTok.IVal);
+            }
         }
     }
 
--- a/test/val/bug2610.c
+++ b/test/val/bug2610.c
@@ -1,0 +1,15 @@
+#include <stdio.h>
+#if '\x0A' != 0x0A
+#error "Suspicious character set translation"
+#endif
+int main()
+{
+    char c = '\x0A';
+    if (c == 0x0A) {
+        printf("Ok\n");
+        return 0;
+    } else {
+        printf("Failed\n");
+        return 1;
+    }
+}
--- a/src/cc65/litpool.c
+++ b/src/cc65/litpool.c
@@ -92,7 +92,7 @@ static Collection       LPStack  = STATIC_COLLECTION_INITIALIZER;
 
 
 
-static Literal* NewLiteral (const void* Buf, unsigned Len)
+static Literal* NewLiteral (const StrBuf* S)
 /* Create a new literal and return it */
 {
     /* Allocate memory */
@@ -103,7 +103,7 @@ static Literal* NewLiteral (const void* Buf, unsigned Len)
     L->RefCount = 0;
     L->Output   = 0;
     SB_Init (&L->Data);
-    SB_AppendBuf (&L->Data, Buf, Len);
+    SB_Append (&L->Data, S);
 
     /* Return the new literal */
     return L;
@@ -162,7 +162,7 @@ void ReleaseLiteral (Literal* L)
 void TranslateLiteral (Literal* L)
 /* Translate a literal into the target charset */
 {
-    TgtTranslateBuf (SB_GetBuf (&L->Data), SB_GetLen (&L->Data));
+    TgtTranslateStrBuf (&L->Data);
 }
 
 
@@ -468,18 +468,18 @@ void OutputGlobalLiteralPool (void)
 Literal* AddLiteral (const char* S)
 /* Add a literal string to the literal pool. Return the literal. */
 {
-    return AddLiteralBuf (S, strlen (S) + 1);
+    StrBuf SB;
+    SB_InitFromString(&SB, S);
+    return AddLiteralStr(&SB);
 }
 
 
 
-Literal* AddLiteralBuf (const void* Buf, unsigned Len)
-/* Add a buffer containing a literal string to the literal pool. Return the
-** literal.
-*/
+Literal* AddLiteralStr (const StrBuf* S)
+/* Add a literal string to the literal pool. Return the literal. */
 {
     /* Create a new literal */
-    Literal* L = NewLiteral (Buf, Len);
+    Literal* L = NewLiteral (S);
 
     /* Add the literal to the correct pool */
     if (IS_Get (&WritableStrings)) {
@@ -490,12 +490,4 @@ Literal* AddLiteralBuf (const void* Buf, unsigned Len)
 
     /* Return the new literal */
     return L;
-}
-
-
-
-Literal* AddLiteralStr (const StrBuf* S)
-/* Add a literal string to the literal pool. Return the literal. */
-{
-    return AddLiteralBuf (SB_GetConstBuf (S), SB_GetLen (S));
 }
--- a/src/cc65/litpool.h
+++ b/src/cc65/litpool.h
@@ -125,11 +125,6 @@ void OutputGlobalLiteralPool (void);
 Literal* AddLiteral (const char* S);
 /* Add a literal string to the literal pool. Return the literal. */
 
-Literal* AddLiteralBuf (const void* Buf, unsigned Len);
-/* Add a buffer containing a literal string to the literal pool. Return the
-** literal.
-*/
-
 Literal* AddLiteralStr (const StrBuf* S);
 /* Add a literal string to the literal pool. Return the literal. */
 
--- a/src/cc65/scanner.h
+++ b/src/cc65/scanner.h
@@ -213,6 +213,7 @@ typedef struct Token Token;
 struct Token {
     token_t         Tok;        /* The token itself */
     long            IVal;       /* The integer attribute */
+    int             Cooked;     /* The "cooked" flag for char constants */
     Double          FVal;       /* The float attribute */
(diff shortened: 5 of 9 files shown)


mrdudz:
a) Please use proper markdown in your posts (for eg code)
b) only one patch per issue. the added \e stuff does not belong into this
c) please add test cases for char constants, string literals, and combinations of those (for the testbench, this will require using those charmap includes and header files, i guess)

I'd also like to hear from @spiro-trikaliotis or @kugelfuhr on how they interpret the part of the standard cited in #2609 (i believe we discussed this before - i don't have time now to dig through past issues/discussions though)

spiro-trikaliotis:
Let's see how I interpret the standard:

For me, 5.2.1 in N1256 ("C99") seems to be more important:

> 5.2 Environmental considerations
>
> 5.2.1 Character sets
>
>1 Two sets of characters and their associated collating sequences shall be defined: the set in which source files are written (the source character set), and the set interpreted in the execution environment (the execution character set). Each set is further divided into a basic character set, whose contents are given by this subclause, and a set of zero or more locale-specific members (which are not members of the basic character set) called extended characters. The combined set is also called the extended character set. The values of the members of the execution character set are implementation-defined.

So, we have two character sets: The source character set, in which the C source code is written, and the execution character set, which is how the characters are used in the target environment.

Then, there is following in the same section:

>2 In a character constant or string literal, members of the execution character set shall be represented by corresponding members of the source character set or by escape sequences consisting of the backslash \ followed by one or more characters. A byte with all bits set to 0, called the null character, shall exist in the basic execution character set; it
is used to terminate a character string.


For me, this demands a translation, because the characters of the execution character set should be represented by corresponding members of the source character set in character constants and string literals. Note also that this paragraph makes it clear that escape sequences with a backslash are also part of this translation. However, it doesn't specifically demands that the escape sequence and the source character with the same numerical  value are translated into the same character in the execution character set.

The parts that are cited in #2609 tell that the octal and hexadecimal escape sequences should correspond to the same literal that can be entered directly.

Thus, from my understand, '\x40' and '@' must be translated to the same, while '\x0d', '\x0a', '\n' and '\r' can be translated to 4 different things.

So, for the examples above, if we assume that \n is translated to 0x0D, then I would expect the same results as they are given above. Note, however, that the character literals for '\x0A' (and the other represesantions) could be translated to something else, while  char_ten and int_ten could not, because these are not character literals.

Of course, this all is only true for strings because 6.4.5 essentially says that string literals should be treated "as if" it were a concatenation of character literals (with some small exceptions):

> 6.4.5 String literals
> [...]
> 3 The same considerations apply to each element of the sequence in a character string
literal or a wide string literal as if it were in an integer character constant or a wide
character constant, except that the single-quote ' is representable either by itself or by the
escape sequence \', but the double-quote " shall be represented by the escape sequence
\".

This is only my interpretation, though.

Note that this discussion came up before in the cc65 mailing list with Uz back then. I found references to it back in 2004.

mrdudz:
> This is only my interpretation, though.

Thanks you. That is also how i remember it....

> Note that this discussion came up before in the cc65 mailing list with Uz back then. I found references to it back in 2004.

... from this. IIRC there was a more recent iteration on github as well (which came to the same conclusion)

groessler:
My 2 Euro-ct:
IMO, if I write for example '\x0a' instead of '\n', I normally want a 0x0a byte, and not think reverse that it's '\n' and that '\n' converts to something different on the target charset. If '\x0a' is converted as well, how would I insert a 0x0a then?

mrdudz:
> IMO, if I write for example '\x0a' instead of '\n', I normally want a 0x0a byte, and not think reverse that it's '\n' and that '\n' converts to something different on the target charset. If '\x0a' is converted as well, how would I insert a 0x0a then?

You use a char array, not a string. Strings are for text, not "bytes" :) It only "works" and people got used to do this, because most of the time they used a native compiler, and no character translation was involved.

GorillaSapiens:
> Thus, from my understand, '\x40' and '@' must be translated to the same, while '\x0d', '\x0a', '\n' and '\r' can be translated to 4 different things.

@spiro-trikaliotis  you can't just ignore 6.4.4.4   '\x40' and '@' do not necessarily translate to the same thing.  specifically, per 6.4.4.4, '\x40' MUST be whatever character 0x40 is in whatever character set you are targeting.  this may not be '@', it could be any other character.  '@' should translate to whatever the at-sign is in whatever character set you are targeting, which may not be the the character 0x40.

spiro-trikaliotis:
> > Thus, from my understand, '\x40' and '@' must be translated to the same, while '\x0d', '\x0a', '\n' and '\r' can be translated to 4 different things.
> 
> @spiro-trikaliotis you can't just ignore 6.4.4.4 '\x40' and '@' do not necessarily translate to the same thing. specifically, per 6.4.4.4, '\x40' MUST be whatever character 0x40 is in whatever character set you are targeting. this may not be '@', it could be any other character. '@' should translate to whatever the at-sign is in whatever character set you are targeting, which may not be the the character 0x40.

I just wanted to explain why I think you are wrong, but looking at it more thoroughly, I think you are actually right.

6.4.4.4 states:
> 2 An integer character constant is a sequence of one or more multibyte characters enclosed
in single-quotes, as in 'x'. A wide character constant is the same, except prefixed by the
letter L. With a few exceptions detailed later, *the elements of the sequence are any
members of the source character set*; *they are mapped in an implementation-defined
manner to members of the execution character set*

Emphasis added be me.

Thus, for 'literal' character constants, it is clear that they are in source character set and have to be mapped.

For octal and hexadecimal constants, the part for the mapping is missing in the spec:

> 5 The octal digits that follow the backslash in an octal escape sequence are taken to be part
of the construction of a single character for an integer character constant or of a single
wide character for a wide character constant. The numerical value of the octal integer so
formed specifies the value of the desired character or wide character.

and

> 6 The hexadecimal digits that follow the backslash and the letter x in a hexadecimal escape
sequence are taken to be part of the construction of a single character for an integer
character constant or of a single wide character for a wide character constant. The
numerical value of the hexadecimal integer so formed specifies the value of the desired
character or wide character.


I *wanted* to say that:
1. I assume a source character set of ASCII, at least for 0x00 - 0x7F;
2. part 6 of the spec speaks about the sources, thus has to be in source character set. However, as there is this distinction between literal characters (which must be specifically converted; so, the conversion is not implicitly, but explicitly mentioned) and octal or hexadecimal constants (where no conversion is mentioned), I think these must be handled differently.

I believe I was wrong: Thus, char c = '\x40'; must be the same as if char c = 0x40; was written, regardless of the translation, while char c = '@'; sets c to the value that the @ has in the execution character set.

mrdudz:
Interesting (as said i'd also prefer it working this way). At this point i'm curious if other compilers do the same - but i actually can't think of one that has characterset translation :) (Not one that aims to be C-Standard compliant anyway.)

I'll leave this dangling for some days, hoping @kugelfuhr will also comment.

In the meantime, this patch needs to be fixed so test/val/bug2151.c does not break (note how it doesn't even use \x, just pragma charmap). We also need another test, which checks if the translation works as intended (something like the code in the OP, but more elaborate, and checking more edge cases - eg using that pragma in various ways)

kugelfuhr:
The original issue (#2609) is correct, the implemented behavior is not standard compliant. The following code should print "Ok" but it prints "Suspicious character set translation" when compiled and "Failed" when run:

```c
#include <stdio.h>
#if '\x0A' != 0x0A
#warning "Suspicious character set translation"
#endif
int main()
{
    char c = '\x0A';
    if (c == 0x0A) {
        printf("Ok\n");
        return 0;
    } else {
        printf("Failed\n");
        return 1;
    }
}
```

I haven't looked at the code for this PR closely so I cannot say if it will work in all cases and has no side effects. Above test program prints "Ok" with the changes which looks good. 

Style issues:
* The passages inside `#if 0` should probably get removed. No need to keep all sorts of old code around. This is what revisions control systems are for.
* Mixing extensions and bug fixes is not optimal, so it might be better to have one PR that adds `\e` and another one that changes character translation.

mrdudz:
OK so - i guess we are back to where we already were many years ago - problem acknowledged, but it's not trivial to fix :)

as a reminder for the OP:

> In the meantime, this patch needs to be fixed so test/val/bug2151.c does not break (note how it doesn't even use \x, just pragma charmap). We also need another test, which checks if the translation works as intended (something like the code in the OP, but more elaborate, and checking more edge cases - eg using that pragma in various ways)

src/cc65/scanner.c
```diff
@@ -346,6 +354,14 @@ static int ParseChar (void)
             case 'b':
                 C = '\b';
                 break;
+            case 'e':
```
mrdudz, line 357: this should be in a separate PR

GorillaSapiens:
> In the meantime, this patch needs to be fixed so test/val/bug2151.c does not break (note how it doesn't even use \x, just pragma charmap). We also need another test, which checks if the translation works as intended (something like the code in the OP, but more elaborate, and checking more edge cases - eg using that pragma in various ways)

i've been working in cygwin, and a lot of the tests just don't work for me, i didn't want to get bogged down in environment configuration issues, and i kind of ran out of steam trying.

i have more time now.

my guess is you're working in Linux?  i can spin up a VM, no big deal, but i'd like to know what the preferred / sanctioned / supported development actually is before i go through the effort.

mrdudz:
Some are using Linux, some macOS, some Windows. The tests should also work in Windows for that matter (the github actions also run them on Windows). The "trick" is to use cmd as a shell on Windows, cygwin or msys will not do the right thing - they will be detected incorrectly as being *nix. Eg @oliverschmidt might be able to tell details on what kind of environment he uses.

GorillaSapiens:
> The original issue (#2609) is correct, the implemented behavior is not standard compliant. The following code should print "Ok" but it prints "Suspicious character set translation" when compiled and "Failed" when run:

code added as val/test/bug2610.c and problem fixed

GorillaSapiens:
> Thus, from my understand, '\x40' and '@' must be translated to the same, while '\x0d', '\x0a', '\n' and '\r' can be translated to 4 different things.

no. '@' may be anything.  for example:
include/atari_screen_charmap.h:#pragma charmap (0x40, 0x20)

'\x40' must ALWAYS be 0x40, regardless of charmap  refer to 6.4.4.4
ditto, '\0x0d' must ALWAYS be 0x0d and '\x0a' must ALWAYS be 0x0A.  again, refer to 6.4.4.4

test/val/bug2610.c
```diff
@@ -0,0 +1,15 @@
+#include <stdio.h>
+#if '\x0A' != 0x0A
+#warning "Suspicious character set translation"
```
mrdudz, line 3: this should be #error

mrdudz:
Two more things:
- as said before, please remove the \e related stuff from this PR, this should be separate
- please extend bug2151.c to test against constants wrapped into the same pragma charmap(s) as used in that test, ie test if multiple redefinitions in the same file work correctly (this is a particular nasty case that broke before - which is why this test exists)

test/val/bug2610.c
GorillaSapiens: fixed in my next push

GorillaSapiens:
> * please extend bug2151.c to test against constants wrapped into the same pragma charmap(s) as used in that test, ie test if multiple redefinitions in the same file work correctly (this is a particular nasty case that broke before - which is why this test exists)

ah 2151.  yeah, this made me rip out my original effort and spend a good 8+ hours yesterday reworking everything.  specifically to pass this test.

bug2151.c already has "multiple redefinitions in the same file", so i'm not sure what exactly you want me to do here.

if you have specific ideas with code examples you want tested i'm happy to incorporate them.

but since i don't understand exactly what additional tests you want done here, it is unlikely the time i spend trying to guess what you are thinking will be fruitful.

mrdudz:
> > * as said before, please remove the \e related stuff from this PR, this should be separate
> 
> aside from organization, do you see anything wrong with what's there now? if you're just going to approve it later, ripping it out just so i can put it back in later is busy work, and i don't like busy work. if you're not going to approve it, i want to fix it to be in line with other changes in this PR

There should be checks that only enable this feature when "cc65" standard is active (since this is a non standard feature).

And yes, please please with sugar on top - make this stuff a separate PR.

GorillaSapiens:
> There should be checks that only enable this feature when "cc65" standard is active (since this is a non standard feature).

is this sufficient, or do i need to do more?
<pre>
            case 'e':
                if (IS_Get(&Standard) != STD_CC65) {
                    goto IllegalEscape;
                }
                /* we'd like to use \e here, but */
                /* not all build systems support it */
                C = '\x1B';
                break;
</pre>

GorillaSapiens:
> Right now that test checks character literals vs character literals (mapped or not remapped). What it does not check, is Rharacter literals (mapped or not remapped) against escaped constants - so that would be a nice addition.

Currently we have the code below.  does that not meet "character literals against escaped constants" ?
if it doesn't i'm afraid i still don't understand what you're asking for

<pre>
    assert('\a' == 0x61);
    assert('\07' == 0x07);
    assert('\x07' == 0x07);
</pre>

and

<pre>
char *s = "\07\a\x07";
char t[] = { 7, 0x61, 7, 0 };

    if (strcmp(s,t) || s[0] == s[1]) {
        printf("\n\n !FAIL! strcmp\n");
        for (i = 0; i < 4; i++) {
            printf("%02x ", s[i]);
        }
        printf("\n");
        for (i = 0; i < 4; i++) {
            printf("%02x ", t[i]);
        }
        printf("\n");
        printf("\n");
        return EXIT_FAILURE;
    }
</pre>

mrdudz:
> > There should be checks that only enable this feature when "cc65" standard is active (since this is a non standard feature).
> 
> is this sufficient, or do i need to do more?

Not sure what you are asking - if that makes them work wth "cc65" standard and produce an error in the other standards - then yes, sure. the easiest way to confirm it works correctly would be adding a test that checks this :) (although i admit, this isnt as trivial as the regular tests, and requires understanding those makefiles)

mrdudz:
> > Right now that test checks character literals vs character literals (mapped or not remapped). What it does not check, is Rharacter literals (mapped or not remapped) against escaped constants - so that would be a nice addition.
> 
> Currently we have the code below. does that not meet "character literals against escaped constants" ? if it doesn't i'm afraid i still don't understand what you're asking for

The key point is "multiple remapping" - like what bug2151.c does, at various places, in different contexts. bug2609.c does check it once, but only remaps once at top level (and at the top of the file), which is the trivial case.

GorillaSapiens:
> The key point is "multiple remapping" - like what bug2151.c does, at various places, in different contexts. bug2609.c does check it once, but only remaps once at top level (and at the top of the file), which is the trivial case.

i've added more to bug2609.c

please note that bug1373.c and bug2151.c also do multiple remappings

GorillaSapiens:
> And yes, please please with sugar on top - make this stuff a separate PR.

Removed.  expect a PR after this merges.

GorillaSapiens:
i believe i have addressed everything now

rofl0r:
As I understand it, the strbuf primitive is used all over the place in the codebase to build strings; does it make sense to add a second string "cooked" to its guts that's only used for one purpose, but adds memory overhead everywhere?

GorillaSapiens:
> As I understand it, the strbuf primitive is used all over the place in the codebase to build strings; does it make sense to add a second string "cooked" to its guts that's only used for one purpose, but adds memory overhead everywhere?

There needs to be a way to track, at the character level, whether something is "raw" (should not be translated) or "cooked" (should be translated).  i thought of modifying the type for the concept of a character as used by StrBuf to be a struct that included a flag, or a larger integer.  as you can imagine, this was DEEPLY intrusive and quickly exploded.

mrdudz:
PR looks good to me now - merging. Thanks for patience :)

PR #2612, opened 2025-03-05
Fix RANGE ERROR STOPPED AT 8214

colinleroy:
On a 48k Apple II, the BLTU2 call throws an error, even when there is nothing to copy in the LC segment.

Checking it's of a non-zero size avoids that.

Screenshot:
image

Execution trace:
```
7EA5: bit ROMIN         ; From crt0.s, starting at "Switch in LC bank 2 for W/O."/
7EA8: bit ROMIN
7EAB: lda #$a7          ; #<(__ONCE_LOAD__ + __ONCE_SIZE__)
7EAD: ldy #$7f          ; #>(__ONCE_LOAD__ + __ONCE_SIZE__)
7EAF: sta $9b
7EB1: sty $9c
7EB3: lda #$a7          ; The source last address, equal to the first
7EB5: ldy #$7f          ; so it should be a 0-byte copy
7EB7: sta $96
7EB9: sty $97
7EBB: lda #$00          ; Destination address (the non-existent LC)
7EBD: ldy #$d4
7EBF: sta $94
7EC1: sty $95
7EC3: jsr $d39a         ; Jumping to BLTU2
D39A: stx $0323
D39D: ldy #$05
D39F: lda ($4a), y
D3A1: tax
D3A2: iny
D3A3: lda ($4a), y
D3A5: tay
D3A6: cpx #$18
D3A8: sbc #$01
D3AA: bcc $d399
D3AC: jmp $ee68
EE68: ldy #$7b
EE6A: bne $ee65
EE65: jmp $e3e0
E3E0: jsr $eed3
EED3: tya
EED4: tax
EED5: ldy #$6e
EED7: jsr $e3c4
E3C4: lda $eb00, y
E3C7: bmi $e3c0
E3C0: jsr F8ROM:COUT    ; The first * echoed to the screen
```


oliverschmidt:
I like the idea to do something like iobuf-800. But I'd go one step further. What about making the $D39A itself an external reference. Then a file like iobuf-800 could contain either the actual Applesoft routine or code compatible with it.
Then we'd in general stay with documenting the apple2 target requiring an Apple ][+. But we'd document the new file as extending the compatibility to include the "original Apple ][ (with Integer BASIC ROM)".

colinleroy:
> I like the idea to do something like iobuf-800. But I'd go one step further. What about making the $D39A itself an external reference. Then a file like iobuf-800 could contain either the actual Applesoft routine or code compatible with it. Then we'd in general stay with documenting the apple2 target requiring an Apple ][+. But we'd document the new file as extending the compatibility to include the "original Apple ][ (with Integer BASIC ROM)".

Hi Oliver,
I don't understand "a file like iobuf-800 could contain either the actual Applesoft routine or code compatible with it."

Should I make two differents files and the user would need to choose from one or the other if they want to use LC?

colinleroy:
I think I understood what you meant and I've pushed a commit to do that. If that's fine for you, I'll update the docs.

oliverschmidt:
Maybe there's a - from the user perspective - btter name for `lc-copy-memcpy.s`. Something like `integer-basic-compat.s` or alike. Something that helps the reader of a Makefile linking it to have a clue. Just an idea...

colinleroy:
Hi @oliverschmidt, here's the doc PR (and renaming) : #2614

PR #2616, opened 2025-03-12
Optimize LZ4 decompressor

colinleroy:
Use a walking out pointer instead of &out[written]. This simplifies the code by 27 bytes and spares 15% cycles. Tested with both the unit test and code uncompressing 10kB of data.

Renamed the labels for legibility.

### DIFF
--- a/libsrc/common/lz4.s
+++ b/libsrc/common/lz4.s
@@ -14,12 +14,11 @@
         .export         _decompress_lz4
 
 out = regsave
-written = regsave + 2
+end = regsave + 2
 tmp = tmp1
 token = tmp2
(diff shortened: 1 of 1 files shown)


colinleroy:
Tagging @clbr as the original code author and @oliverschmidt who reviewed the initial PR #446

oliverschmidt:
I have nothing smart to add to the picture. The new code seems for sure shorter and I guess it's faster. So if it still behaves correctly (which Colin states he has checked) then I see no reason to not merge the PR. As usual I'd like to mention that Colin has proven to be around if anything he contributes should cause a regression.

PR #2626, opened 2025-04-07
Fix a typo in g_typeadjust()

### LINKED ISSUE #2611: g_typeadjust problem?
in src/cc65/codegen.c g_typeadjust(), we see

```
    /* Otherwise, if one operand has type long int and the other has type unsigned int,
    ** if a long int can represent all values of an unsigned int, the operand of type unsigned int
    ** is converted to long int ; if a long int cannot represent all the values of an unsigned int
    ** both operands are converted to unsigned long int.
    */
    if ((ltype == CF_LONG && rtype == CF_INT && (rhs & CF_UNSIGNED)) ||
        (rtype == CF_LONG && ltype == CF_INT && (rhs & CF_UNSIGNED))) {
        /* long can represent all unsigneds, so we are in the first sub-case. */
        return const_flag | CF_LONG;
    }
```

the code doesn't seem to match the comment.  shouldn't it be

```
    if ((ltype == CF_LONG && rtype == CF_INT && (rhs & CF_UNSIGNED)) ||
        (rtype == CF_LONG && ltype == CF_INT && (lhs & CF_UNSIGNED))) {   // <--- note change here!
        /* long can represent all unsigneds, so we are in the first sub-case. */
        return const_flag | CF_LONG;
    }
```

???

kugelfuhr:
This fixes #2611. I tried to create a test program but that's not possible since the result of `g_typeadjust ()` is the same with and without the change. It takes another branch without the fix but has the same outcome. Nevertheless it's still an error since the code tried to behave as described in the standard, but did not.

### DIFF
--- a/src/cc65/codegen.c
+++ b/src/cc65/codegen.c
@@ -1489,7 +1489,7 @@ unsigned g_typeadjust (unsigned lhs, unsigned rhs)
     ** both operands are converted to unsigned long int.
     */
     if ((ltype == CF_LONG && rtype == CF_INT && (rhs & CF_UNSIGNED)) ||
-        (rtype == CF_LONG && ltype == CF_INT && (rhs & CF_UNSIGNED))) {
+        (rtype == CF_LONG && ltype == CF_INT && (lhs & CF_UNSIGNED))) {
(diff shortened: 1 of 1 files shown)


mrdudz:
thanks!

PR #2627, opened 2025-04-07
Add decompress_lz4 C implementation as comment

colinleroy:
For reference.
I suspected a bug in the decompressor and had to write it in C to help debugging. (It was not a bug in the decompressor).

As I could not find the "official C source" where it came from, I suppose it's good to have it there, in the same way malloc() has its C equivalent in a comment.

Feel free to reject if we don't care!


mrdudz:
Does the comment above the lines you added still apply though? :)

colinleroy:
Approximately :)

PR #2628, opened 2025-04-11
Added search path relative to the binary on POSIX systems

zmetzing:
On POSIX systems, including FreeBSD, a ".macpack longbranch" will fail to find the include file in $(DESTDIR)/share/cc65/asminc, as that is not in the include paths. This small patch corrects that oversight.

### DIFF
--- a/src/ca65/incpath.c
+++ b/src/ca65/incpath.c
@@ -79,6 +79,11 @@ void FinishIncludePaths (void)
     AddSearchPath (IncSearchPath, CA65_INC);
 #endif
 
+#if defined(_WIN32)
     /* Add paths relative to the parent directory of the Windows binary. */
     AddSubSearchPathFromBin (IncSearchPath, "asminc");
+#else
+    /* Add paths relative to the parent directory of the POSIX binary. */
+    AddSubSearchPathFromBin (IncSearchPath, "share/cc65/asminc");
+#endif
 }
--- a/src/ca65/incpath.c
+++ b/src/ca65/incpath.c
@@ -79,10 +79,11 @@ void FinishIncludePaths (void)
     AddSearchPath (IncSearchPath, CA65_INC);
 #endif
 
-#if defined(_WIN32)
     /* Add paths relative to the parent directory of the Windows binary. */
+    /* Also needed for "make test" to work. */
     AddSubSearchPathFromBin (IncSearchPath, "asminc");
-#else
+
+#if !defined(_WIN32)
     /* Add paths relative to the parent directory of the POSIX binary. */
     AddSubSearchPathFromBin (IncSearchPath, "share/cc65/asminc");
 #endif


rofl0r:
are you sure you're using latest code ? i fixed precisely that not too long ago:
https://github.com/cc65/cc65/issues/1726
https://github.com/cc65/cc65/commit/6a5375ace4e34553f14b929a059d75c914ef291b

rofl0r:
btw:

> On POSIX systems, including FreeBSD, a ".macpack longbranch" will fail to find the include file in $(DESTDIR)/share/cc65/asminc

that's not supported use anyway. DESTDIR is a staging mechanism for packagers, and not to be hardcoded into binaries. instead $prefix/share/cc65/asminc is where the file is looked for.

zmetzing:
> that's not supported use anyway. DESTDIR is a staging mechanism for packagers, and not to be hardcoded into binaries. instead $prefix/share/cc65/asminc is where the file is looked for.

Yes, the patch is against HEAD. No, $(DESTDIR) isn't hard-coded into the binary.

If installed into /opt/local, for instance, the binaries ends up in /opt/local/bin/... and the assembly includes end up in /opt/local/share/cc65. In my specific case, DESTDIR is /opt/local/install/cc65 and a tree of symlinks point back to that from /opt/local/bin, etc. The resulting failure can be seen by a ktrace of the build process:
```
79186 ca65     NAMI  "asm/longbranch.mac"
 79186 ca65     RET   access -1 errno 2 No such file or directory
 79186 ca65     CALL  access(0x823814040,0<F_OK>)
 79186 ca65     NAMI  "/share/cc65/asminc/longbranch.mac"
 79186 ca65     RET   access -1 errno 2 No such file or directory
 79186 ca65     CALL  access(0x823814040,0<F_OK>)
 79186 ca65     NAMI  "/opt/local/install/cc65/asminc/longbranch.mac"
 79186 ca65     RET   access -1 errno 2 No such file or directory
```
This patch fixes the issue by working backwards from the ca65 binary path and adding the proper share/cc65/asminc directory to the search path.

rofl0r:
if your DESTDIR is /opt/local/install/cc65, then what's your prefix ? if it's empty, and you're using symlinks to anchor your install into the main filesystem (like my distro sabotage linux does - i'm curious what solution you're using here), then /share/cc65/asminc/longbranch.mac is precisely where cc65 should be looking - and apparently does. and also were you should have your symlink to your DESTDIR. the solution we've come up to make POSIX binaries find components relative to the binary is meant for the purpose of running the binary from the build dir, not to climb arbitrary directory structures of weird stow setups (in case that's what you're using).

zmetzing:
> then what's your prefix ?

The source never uses the GNU-autoconf-life $prefix, as far as I can tell. The variables CA54_INC and CC65_HOME from the user's environment are used, and some "compiled-in search paths".

> the solution we've come up to make POSIX binaries find components relative to the binary is meant for the purpose of running the binary from the build dir

According to the comments in the source, the "solution" does the following:
```
/* Add paths relative to the parent directory of the Windows binary. */
```
Anyway, I find the change useful for my setup, which keeps each package in its own /opt/local/install/[package] tree and then symlinks that into /opt/local. The patch improves the ability of ca65 to find include files and has a negligible impact on execution time. 

Feel free to merge, or not.

rofl0r:
> The source never uses the GNU-autoconf-life $prefix, as far as I can tell. The variables CA54_INC and CC65_HOME from the user's environment are used, and some "compiled-in search paths".

wrong. src/Makefile is the place where the "compiled-in search path" is passed to the compiler to be baked in as e.g. CA65_INC: `-DCA65_INC="\"$(CA65_INC)\""` which is composed earlier from PREFIX via datadir: `CA65_INC = $(datadir)/asminc` and `datadir := $(PREFIX)/share/cc65`.

that's the standard way of doing things in unix programs that require to find datafiles at runtime.
what's not standard is the "lookup stuff relative to binary" scenario, which was added to the unix code because that's how things used to work for windows users, and before my changes the mechanism to "detect" whether the binary-relative lookup should be used was to check whether prefix is empty, which is obviously bogus- because an empty prefix is a valid choice for a system-wide install, to have e.g. ca65 end up in /bin.

> keeps each package in its own /opt/local/install/[package] tree and then symlinks that into /opt/local

that's quite similar to how i do things in sabotage linux: each package that's compiled uses --prefix= (empty), and then DESTDIR=/opt/packagename. after install it's then symlinked to prefix:
in cc65's case we end up with the following structure:

    /bin/cc65 -> ../opt/cc65/bin/cc65
    /share/cc65/asminc/longbranch.mac -> ../../../opt/cc65/share/cc65/asminc/longbranch.mac

this works, because the files are actually available in the prefix that was configured.
in your case, you'd need to set --prefix to /opt/local, however with DESTDIR /opt/local/install/PKG your files would end up in DESTDIR+prefix+filename, which results in e.g. /opt/local/install/PKG/opt/local/bin/cc65, pointing to /opt/local/bin/cc65 which is rather awkward, yet it would work as expected.
with your current setup, any package that uses datafiles won't find those if you use --prefix= instead of --prefix=/opt/local. so i'm curious where the setup you're using comes from.

PR #2631, opened 2025-04-26
Apple2: implement get_tv()

colinleroy:
get_tv() will return TV_NTSC or TV_PAL for any Apple II model with a way of checking vblank. For Apple ][ and ][+ it will return TV_OTHER and let the user figure it out in another way.

Takes 186 bytes on disk, of which 179 are in ONCE, so 6 bytes in memory.

Contrary to our previous discussion on the subject, @oliverschmidt, I have chosen to make the function available to __APPLE2__ platform, as binaries built for __APPLE2__ may well be used on apple2enh-capable machines. This allows game developers to be able to rely on this function with a single APPLE2 build of their games. (like me :wink:)

From the user perspective, it's now very simple:
```
sync = get_tv();
if (sync == TV_OTHER) {
  ask_user_for_freq();
}
```

or
```
        ldx     _tv
        beq     :+
        rts

        ; We don't know. Let's ask.
        ...
        stx     _tv
        rts
```


asminc/apple2.inc
```diff
@@ -91,3 +91,8 @@ IOUDISOFF   :=  $C07F   ; Enable IOU
 
 ; Control Your Apple
 CYAREG  :=      $C036   ; Bits 0-3=disk detect 4=shadow all banks 7=fast
+
+; PAL/NTSC
+TV_NTSC  :=     0
```
oliverschmidt, line 96: := is only for addresses. For non-address concstants = is to be used. See https://github.com/cc65/cc65/blob/master/asminc/c64.inc for and example getting this right.

doc/funcref.sgml
```diff
@@ -4108,6 +4110,28 @@ be used in presence of a prototype.
 </quote>
 
 
+<sect1>get_tv<label id="get_tv"><p>
+
+<quote>
+<descrip>
+<tag/Function/The function returns the system's vertical blank frequency.
+<tag/Header/<tt/<ref id="apple2.h" name="apple2.h">/
```
oliverschmidt, line 4118: There are quite some other headers containing this prototype. Yes, you are supposed to list them all, see i.e. https://cc65.github.io/doc/funcref.html#ss3.295

libsrc/apple2/get_tv.s
```diff
@@ -0,0 +1,151 @@
...
+.endproc
+
+.proc calibrate_iigs
+        ; Backup speed and slow down
+        jsr     _get_iigs_speed
+        pha
+        lda     #SPEED_SLOW
+        jsr     _set_iigs_speed
+
+        ; The same as IIe, but reverted, because fuck you that's why
```
oliverschmidt, line 111: * I personally dislike any swearing in source code comments.
* Most of the time, something considered stupid turns out to have some valid reason leaving the one calling it stupid behind looking stupid ;-)

libsrc/apple2/get_tv.s
```diff
@@ -0,0 +1,151 @@
...
+
+; Cycle wasters
+waste_72:
+        jsr     waste_36
+waste_36:
+        jsr     waste_12
+waste_24:
+        jsr     waste_12
+waste_12:
+        rts
```
oliverschmidt, line 24: As discussed, cycle wasting should include a reference to the paddle circuit in order to (likely) cause accelerators to slow down.

libsrc/apple2/get_tv.s
```diff
@@ -0,0 +1,151 @@
...
+.proc vbl_calibrate_done
+        lda     #TV_NTSC      ; Consider we're at 60Hz
+        sta     _get_tv+1
+        ; we'll now have X ~= 205 at 60Hz, ~= 240 at 50Hz:
+        ; 205*85 = 17340, 240*85 = 20400.
+        ; Consider any X >= 220 means 50Hz
+        cpx     #220
+        bcc     :+
+        lda     #TV_PAL       ; we're at 50Hz
+        sta     _get_tv+1
```
oliverschmidt, line 140: You can't assume to be able to write to the CODE segment.

libsrc/apple2/get_tv.s
```diff
@@ -0,0 +1,151 @@
...
+waste_72:
+        jsr     waste_36
+waste_36:
+        jsr     waste_12
+waste_24:
+        jsr     waste_12
+waste_12:
+        rts
+
+.proc calibrate_tv
```
oliverschmidt, line 26: A general remark: You may have noticed that there was (nearly) no .proc inside the apple2 directory before you started to contribute to it. That was because you have not too seldomly two functions share code by one falling through into the other. So for consistency reasons I personally stayed clear from .proc altogether.

libsrc/apple2/get_tv.s
colinleroy: Sorry! copy-pasted it from my code without thinking twice :-)

libsrc/apple2/get_tv.s
```diff
@@ -0,0 +1,151 @@
...
+        lda     #TV_PAL       ; we're at 50Hz
+        sta     _get_tv+1
+:       rts
+.endproc
+
+; The only thing remaining from that code after init
+.segment "CODE"
+
+.proc _get_tv
+        lda     #TV_OTHER
```
oliverschmidt: Hm, not sure if self-modifying code is a bit over-engineered here. I.e. https://github.com/cc65/cc65/blob/master/libsrc/apple2/get_ostype.s doesn't use it which causes two more bytes. If you feel that it is desirable I ask you to modify get_ostype (in another PR) as well for consistency reasons.

libsrc/apple2/get_tv.s
colinleroy: Is there any other softswitch I can read that would slow the accelerators down? If I bit PTRIG in the loop, it resets the //c's VBL.

libsrc/apple2/get_tv.s
colinleroy: Ah, that's why. I like the .procs because it gives my debugger the size of the function. In my own code I have some fallthroughs that I comment this way:

```
.proc extra_things
...
        ; Fallthrough to things
.endproc

.proc things
...
```

I can remove them, though.

libsrc/apple2/get_tv.s
colinleroy: I don't think that's important, it's better if it's clean. Having _get_tv or _get_ostype in DATA feels weird :) I've pushed a change to mimic ostype (with tv in INIT).

colinleroy:
- Tested on IIgs with both 50Hz and 60Hz, both booting slow and fast
- Tested on PAL //c
- Tested on NTSC //c+ (by Dennis)
- Tested on emulation for ][+, IIe, IIee, IIep.

libsrc/apple2/get_tv.s
```diff
@@ -0,0 +1,187 @@
...
+        .constructor    calibrate_tv, 2
+
+        .include    "accelerator.inc"
+        .include    "apple2.inc"
+        .include    "get_tv.inc"
+
+; The only thing remaining from that code after init
+.proc _get_tv
+        lda     tv
+        ldx     #$00
```
oliverschmidt, line 20: Please think about the cc65 policy to mark this as integer promotion. In the apple2 specific code base this is consistently doene by writing ldx >#$0000

libsrc/apple2/get_tv.s
```diff
@@ -0,0 +1,187 @@
...
+        .import         ostype
+
+        .constructor    calibrate_tv, 2
+
+        .include    "accelerator.inc"
+        .include    "apple2.inc"
+        .include    "get_tv.inc"
+
+; The only thing remaining from that code after init
+.proc _get_tv
```
oliverschmidt, line 18: Just a matter of style: So far my style was to either have "just some code" without any segment assignment in the source file or to have a more complex with (several) segment assignment. The latter case I explicitly wrote .code and didn't rely on it being the default. Additionally I'm someone how puts the main() at then end of a C source. In the same way I'd put this "final ourward facing result" code at the end of this source file. And then you need .code anyhow. But as I wrote, just a question of style...

libsrc/apple2/get_tv.s
```diff
@@ -0,0 +1,187 @@
...
+.endproc
+
+.proc calibrate_iie
+:       bit     RDVBLBAR      ; Wait for bit 7 to be off (VBL start)
+        bmi     :-
+:       bit     RDVBLBAR      ; Wait for bit 7 to be on (VBL end)
+        bpl     :-
+
+        ; Wait and count during a full cycle
+        ldx     #0
```
oliverschmidt, line 130: Please be consistent about #$00 vs #0.

oliverschmidt:
Thank you 👍🏻

PR #2634, opened 2025-05-01
Move IRQ handler into CODE

colinleroy:
We could have IRQs firing before LOWCODE is loaded if the user linked LOWCODE to a separate binary they load themselves.

This is a suggestion, as I saw no comment on the subject, but there may be a good reason to have it in LOWCODE?

My use case is Shufflepuck where I init the mouse first thing (because it clears HGR on IIp), then load the file containing  LOWCODE+splashscreen. 

For now I php/sei before mouse init and plp after LOWCODE load, but that breaks loading Shufflepuck over Appletalk, as far as I understand.

### DIFF
--- a/libsrc/apple2/irq.s
+++ b/libsrc/apple2/irq.s
@@ -60,7 +60,7 @@ doneirq:
         .addr   i_param
 :       rts
 
-        .segment        "LOWCODE"
+        .segment        "CODE"
 
 intptr:
         ; ProDOS TechRefMan, chapter 6.2:


oliverschmidt:
The IRQ handler is deliberately located in LOWCODE for most (?) targets. The idea is that there might be address ranges that can be banked. Those address ranges tend to be towards the higher addresses. So stuff that needs to be reentrant is better placed in low addresses. This idea doesn't fit the Apple II perfectly well but there's at least the option on 128kB machines to "hide" code "behind" the HIRES screen. In that scenario, the IRQ handler mustn't be located in $2000-$4000. I see that this is a rather obscure scenario, but at least to be it seems less obscure than LOWCODE not being present on startup.

Issue #2637, opened 2025-05-02
include/* headers give warnings when compiled with --standard c99


GorillaSapiens:
the headers in include/ should NOT issue warnings

these warnings do not happen with --standard c89 or --standard cc65

it's not just stdio.h either, many (maybe all) generate warnings

<pre>
#include &lt;stdio.h&gt;

// compile with --standard c99
int main()
{
   return 0;
}
</pre>

<pre>
$ bin/cc65 --standard c99 foo.c
In file included from foo.c:1:
/home/gorilla/cc65/include/stdio.h:37: Warning: ISO C99 requires whitespace after the macro name
/home/gorilla/cc65/include/stdio.h:47: Warning: ISO C99 requires whitespace after the macro name
</pre>

GorillaSapiens:
fixed by #2639

PR #2639, opened 2025-05-03
fixes issue #2637

GorillaSapiens:
in c99, whitespace is not required if there is no replacement list

we were producing warnings for things like
<pre>
#define FOO
</pre>
when we really shouldn't.


mrdudz:
looks good to me, thanks!

PR #2641, opened 2025-05-05
Apple2: Make 80-columns support dynamic

colinleroy:
This allows supporting 80-column mode on every Apple II with an 80-column card installed, enhanced or not.


asminc/apple2.inc
```diff
@@ -44,6 +44,9 @@ SET80COL:=      $C001   ; Enable 80 column store
 RD80COL :=      $C018   ; >127 if 80 column store enabled
 RD80VID :=      $C01F   ; >127 if 80 column video enabled
 
+VIDEOMODE_40x24 = $15
+VIDEOMODE_80x24 = $00
```
oliverschmidt, line 48: It somehow feels to me that this doesn't belong here. This file is supposed to be "universal Apple II truth", not about cc65 specific things.

libsrc/apple2/videomode.s
```diff
@@ -1,17 +1,51 @@
...
+; signed char __fastcall__ videomode (unsigned mode);
 ;
-        .ifdef  __APPLE2ENH__
-
         .export         _videomode
 
+        .import         has_80cols_card, returnFFFF
+
+        .ifndef __APPLE2ENH__
+        .constructor    detect_80cols_card, 25
```
oliverschmidt, line 11: I think this is a misnomer. You're not detecting anything. You're just relying on the detection performend by ProDOS.

libsrc/apple2/videomode.s
```diff
@@ -1,17 +1,51 @@
...
+        .include        "mli.inc"
+
+
+        .ifndef __APPLE2ENH__
+
+        .segment        "ONCE"
+
+; has_80cols_card = $80 if a 80 column card is present, $00 otherwise
+.proc detect_80cols_card
+        lda     MACHID
```
oliverschmidt, line 24: That's (in contrast to the name) a real problem. The cc65 C library is supposed to be usable without ProDOS. If you check the exsisting occurances of access to MACHID, they are all behind a check of ProDOS being there in the first place.
Apart from that aspect, I think this check might not be what we want in the first place:
* If we're running on a ][+ with a Videx 80 column card, then ProDOS will set this flag. But all our 80 column code will fail misserably.
* We're only supporting the //e 80 column functionality. So why don't we just check for the //e? Yes, there might be //e's without the 80 column board. But isn't that the 100% same question with today's apple2enh code?
What I'm trying to express is that it seems to me that you don't really want a flag that says "yes, I have an 80 column card" but rather a flag that says "yes, I'm an //e (and therefore I can rely on all the //e softswitches)".

libsrc/apple2/tgi/a2.hi.s
```diff
@@ -124,6 +125,10 @@ pages:  .byte   2               ; Number of screens available
 
 ERROR:  .res    1               ; Error code
 
+.ifndef __APPLE2ENH__
+has_80cols_card:                ; Our own copy of the standard variable,
+        .res    1               ; as we don't have access to it from here.
```
oliverschmidt, line 130: Not really. That's exactly what the "library references" are for. Think about i.e. the joystick driver that wants to know what type of machine it runs on. Does it replicate the machine detection? No, it uses the library reference mechanism:
For every machine type and for every driver class (not driver instance!) there's a symbol defined in the static C library that is handed to the drivers in question. The (usually unused) default is an alias to the _exit symbol:
https://github.com/cc65/cc65/blob/master/libsrc/apple2/libref.s
But if on a certain machine a certain driver class needs access to something from the static C library, it can refer to it. The Apple II joystick driver class does so:
https://github.com/cc65/cc65/blob/master/libsrc/apple2/joyref.s#L8
https://github.com/cc65/cc65/blob/master/libsrc/apple2/joy/a2.stdjoy.s#L68-L73
If more than one thing is needed, one defines a struct containing the things in question and accesses them via offsets from the reference.

colinleroy: I always forget about libref :)

asminc/apple2.inc
colinleroy: Understood.
I'll move them away to videomode.s for now; but on that subject, I often miss having asm defines in my programs for values (like CH_CURS_*, APPLE_IIE, etc). Should I introduce a file specific to those (in another PR)?

colinleroy:
Hi Oliver,
I addressed your comments. I've also opened #2645 in advance, that's where I'd like to go from this one. We can either do this one first then 2645 or close that one and go directly to #2645, as you prefer!

include/apple2.h
```diff
@@ -202,6 +208,11 @@ extern void a2_lo_tgi[];
 
 
 
+signed char __fastcall__ videomode (unsigned mode);
```
oliverschmidt, line 211: A general remark: In former times, this secton was sorted alphabetically.

libsrc/apple2/80cols.s
```diff
@@ -0,0 +1,11 @@
+.ifndef __APPLE2ENH__
+
+        .export         has_80cols_card
```
oliverschmidt, line 3: I'm a bit surprised that you're still with the "has_80cols_card" terminology after our discussion. I thought we came up with the approach to have a single, general byte named i.e. "is_2e". That byte isn't restricted to anything video. It's in general used everywhere, where you want prior static behavior to be dynamic.
As discussed: There could very well //e machines without 80col support. So far, this wasn't taking into account in the apple2enh target. This should be improved both for the apple2enh and now the apple2 target - in the _same_ way. And given that the apple2enh target doesn't have or get a "has_80cols_card", the apple2 target doesn't get one too.
Apart from all this there's another aspect: The so far static 80cols support could just assume to have the MouseText chars avail in the alternative charset - because is requires a 65C02 which is supposed to always be combined with MouseText. But now with the dynamic approach, we might very well run on an unenhanced //e. I don't have an overview but I believe that there are areas like i.e. chline/cvline where where an unenhanced //e needs to be rather treated like "some" Apple ][ and not like an Apple //e.
Coming from there I assume it would be great to have the "is_2e" byte rather have two significant bits. One for all //e (meaning all the additional softswitches are in place) and one for the enhanced //e (meaining the MouseText charset is in place). I propose to use bit 7 for all //e and bit 6 for the enhanced //e.  The latter allows you to use `bit` followed by `bvc` / `bvs` to branch depending on MouseText presence. Maybe "is_2e" should then rather be named something like i.e. "machine_type".

libsrc/apple2/cputc.s
```diff
@@ -5,13 +5,15 @@
 ; void __fastcall__ cputc (char c);
 ;
 
-        .ifdef  __APPLE2ENH__
-        .constructor    initconio
-        .endif
+        ; Call constructor *after* the 80-columns card detection
+        ; in videomode.s
+        .constructor    initconio, 7
```
oliverschmidt, line 10: As far as I know, 7 is the default priority. It's rather normal that default priority constructors implicitly rely on higher priority contructors having run first. There's no need/desire to exolictly mention this and/or explicitly set the default priority. If there's a need/desire to clarify things then they should be mentioned at the higher priority constructor explaining why the explicitly given priority needs to be higher there.

libsrc/apple2/videomode.s
```diff
@@ -1,17 +1,53 @@
 ;
 ; Oliver Schmidt, 07.09.2009
 ;
-; unsigned __fastcall__ videomode (unsigned mode);
+; signed char __fastcall__ videomode (unsigned mode);
```
oliverschmidt, line 4: To clarify: I consider the current apple2enh videomode() bogus. It just assumes that every enhanced //e has an 80 column card. Now that we have defined that videomode() can fail, it's clear how the apple2enh videomode() needs to be fixed: It needs to check if `RD80VID` has the expected value after having executed the switch logic.
Background knowledge (as far as I know): The //e 80 col board just contains RAM, nothing else. So it's the task of the -always present - 80 col firmware to check if the RAM is present. If it isn't, it doesn't switch to 80 cols. That can be detected by looking at `RD80VID` after calling the firmware.
All that needs to be done independently from static or dynamic 80 col support. The only thing the dynamic support needs to do addtionally is to check if it runs on a //e in the first place, because otherwise there's no `RD80VID` and no col80 firmware to begin with.

colinleroy:
Thanks for the review Oliver. I'll address those this week-end, I think!

oliverschmidt:
If the dynamic 80 col functionality is to be identical to the static functionality - which I assume - then the best I can come up with for the conceptual issue I pointed out above is that the user needs to explcitly enable dynamic box drawing chars (by defining `DYN_BOX_DRAW` before including header files). This way we stay compatible with all exsisting code (which easily may rely on the boax drawing chars to be static) while still alowing new code to take full advantage of the new dynamic 80 col functionality.

In apple2.h we have:

```
#if defined(__APPLE2ENH__)
#define CH_HLINE        0x5F
#define CH_VLINE        0xDF
#define CH_ULCORNER     0x5F
#define CH_URCORNER     0x20
#define CH_LLCORNER     0xD4
#define CH_LRCORNER     0xDF
#define CH_TTEE         0x5F
#define CH_BTEE         0xD4
#define CH_LTEE         0xD4
#define CH_RTEE         0xDF
#define CH_CROSS        0xD4
#else
#if defined(DYN_BOX_DRAW)
#define CH_HLINE        ch_box(0)
#define CH_VLINE        ch_box(1)
#define CH_ULCORNER     ch_box(2)
#define CH_URCORNER     ch_box(3)
#define CH_LLCORNER     ch_box(4)
#define CH_LRCORNER     ch_box(5)
#define CH_TTEE         ch_box(2)
#define CH_BTEE         ch_box(4)
#define CH_LTEE         ch_box(4)
#define CH_RTEE         ch_box(5)
#define CH_CROSS        ch_box(4)
#else
#define CH_HLINE        '-'
#define CH_VLINE        '!'
#define CH_ULCORNER     '+'
#define CH_URCORNER     '+'
#define CH_LLCORNER     '+'
#define CH_LRCORNER     '+'
#define CH_TTEE         '+'
#define CH_BTEE         '+'
#define CH_LTEE         '+'
#define CH_RTEE         '+'
#define CH_CROSS        '+'
#endif
#endif
```

Then we have the ch_box() implementation which returns for...
0: 0x5F or '-'
1: 0xDF or '!'
2: 0x5F or '+'
3: 0x20 or '+'
4: 0xD4 or '+'
5: 0xDF or '+'
based on the availability of MouseText.

So far so good, but then chline() and cvline() don't necessarily match. Therefore we need to make sure they use whatever the user chose via `DYN_BOX_DRAW`. So we need in conio.h something like https://github.com/cc65/cc65/blob/master/include/conio.h#L203-L217 for chline()/chlinexy()/cvline()/cvlinexy()

And in apple2.h we need:

```
#if defined(__APPLE2ENH__)
#define _chline(length)       chline(length)
#define _chlinexy(x,y,length) chlinexy(x,y,length)
#define _cvline(length)       cvline(length)
#define _cvlinexy(x,y,length) cvlinexy(x,y,length)
#else
#define _chline(length)       dyn_chline(CH_HLINE,length)
#define _chlinexy(x,y,length) dyn_chlinexy(CH_HLINE,x,y,length)
#define _cvline(length)       dyn_cvline(CH_VLINE,length)
#define _cvlinexy(x,y,length) dyn_cvlinexy(CH_VLINE,x,y,length)
#endif
```

I propose to extend https://github.com/cc65/cc65/blob/master/targettest/conio.c to replace https://github.com/cc65/cc65/blob/master/targettest/conio.c#L27-L33 with an array programatically filled in main() if `DYN_BOX_DRAW` is defined. Then this program can serve both as test for you and as demo on how to use `DYN_BOX_DRAW`. I leave it up to you in which detail (if at all) you want to document `DYN_BOX_DRAW`.

include/apple2.h
```diff
@@ -83,7 +83,63 @@
...
+extern char CH_VLINE_SCRCODE;
+extern char CH_ULCORNER;
+extern char CH_URCORNER;
+extern char CH_LLCORNER;
+extern char CH_LRCORNER;
+extern char CH_TTEE;
+extern char CH_BTEE;
+extern char CH_LTEE;
+extern char CH_RTEE;
+extern char CH_CROSS;
```
oliverschmidt, line 135: I don't know why we're back to 13 individual chars. I seem to remember we had agreed that we need only 5. In a comment to the old PR I had already shown how the 13 macros map to the 5 values.

colinleroy: I've done it this way but they point to 5 different bytes (see libsrc/apple2/boxchars.s)

include/apple2.h
```diff
@@ -95,7 +151,12 @@
 #define CH_LTEE         '+'
 #define CH_RTEE         '+'
 #define CH_CROSS        '+'
-#endif
+#define _chline(length)         dyn_chline(CH_HLINE|0x80, length)
+#define _chlinexy(x, y, length) dyn_chlinexy(CH_HLINE|0x80, x, y, length)
+#define _cvline(length)         dyn_cvline(CH_VLINE|0x80, length)
+#define _cvlinexy(x, y, length) dyn_cvlinexy(CH_VLINE|0x80, x, y, length)
```
oliverschmidt, line 157: I can't follow you with the two sets of _c[h|v]line<xy> macros. No matter if DYN_BOX_DRAW is defined or not, the user is suppoed to be able to write i.e. cputc(CH_HLINE). He's not supposed to have to write cputc(CH_HLINE_SCRCODE) or cputc(CH_HLINE|0x80). In the end, _c[h|v]line<xy>() call internally something like cputc(). So I don't see why what works for cputc() doesn't also work as-is for _c[h|v]line<xy>().

libsrc/apple2/boxchars.s
```diff
@@ -0,0 +1,72 @@
...
+
+VERT:          .byte      $5F | $80
+VERT_SCRCODE:  .byte      $5F
+
+ULCORNER:      .byte      $5F
+URCORNER:      .byte      $20
+LLCORNER:      .byte      $D4
+LRCORNER:      .byte      $DF
+
+.endif                ; not __APPLE2ENH__
```
oliverschmidt, line 72: It seems to me we have a different idea how this function works. In may mind there are two arrays of 5 chars. And this function loops just copies the one arry over the other array on a loop over 5 items.

libsrc/apple2/dynchline.s
```diff
@@ -0,0 +1,39 @@
+;
+; Ullrich von Bassewitz, 08.08.1998
+;
+; void __fastcall__ chlinexy (unsigned char x, unsigned char y, unsigned char length);
+; void __fastcall__ chline (unsigned char length);
```
oliverschmidt, line 5: Please adjust this comment.

libsrc/apple2/dyncvline.s
```diff
@@ -0,0 +1,38 @@
+;
+; Ullrich von Bassewitz, 08.08.1998
+;
+; void __fastcall__ cvlinexy (unsigned char x, unsigned char y, unsigned char length);
+; void __fastcall__ cvline (unsigned char length);
```
oliverschmidt, line 5: Please adjust this comment.

include/apple2.h
oliverschmidt: I see. CH_HLINE isn't a constant, so the compiler has to generate code for the things you write in the macros - for each and every call site! Wouldn't it be much better to rather have a
`eor     #$80            ; Invert high bit`
inside the dyn_...() functions?

libsrc/apple2/boxchars.s
oliverschmidt: Now I understand better. Taking HORIZ_SCRCODE and VERT_SCRCODE aside you have the same 6 values (not 5 as I wrote) that I see.
Are you sure that 6 bytes of .RODATA plus a copy loop from 5 down to 0 in .CODE aren't less bytes than what you have now?

targettest/conio.c
```diff
@@ -41,6 +49,35 @@ static char grid[5][5] = {
     {CH_VLINE,    ' ',      CH_VLINE, ' ',      CH_VLINE   },
     {CH_LLCORNER, CH_HLINE, CH_BTEE,  CH_HLINE, CH_LRCORNER}
 };
+#endif
+
+#if defined(__APPLE2__) && !defined(__APPLE2ENH__)
```
oliverschmidt, line 54: This needs to depend on DYN_BOX_DRAW.

targettest/conio.c
```diff
@@ -9,6 +9,10 @@
  *
  */
 
+#if defined(__APPLE2__) && !defined(__APPLE2ENH__)
```
oliverschmidt, line 12: I don't think that it is desirable to make this depend on any target define.

targettest/conio.c
```diff
@@ -9,6 +9,10 @@
...
+/* Have the grid characters adapt to the availability of MouseText. */
```
oliverschmidt: I propose a comment along these lines:
_Box drawing characters are usually constant expressions. However, there are scenarios where this is not the case. To ensure compatibility with code that assumes they are constant expressions, the scenarios in question must be explicitly enabled by defining DYN_BOX_DRAW. Currently, the only such scenario is the apple2 target. There, DYN_BOX_DRAW can be used to enable the use of MouseText characters on exactly those machines that support them._

include/apple2.h
```diff
@@ -83,7 +83,57 @@
...
 #define CH_CURS_RIGHT   0x15
 
-#if !defined(__APPLE2ENH__)
+#if defined(__APPLE2ENH__)
+
+/* MouseText-based functions for boxes and lines drawing */
+void mt_chline(unsigned char length);
+void mt_cvline(unsigned char length);
+void mt_chlinexy(unsigned char x, unsigned char y, unsigned char length);
+void mt_cvlinexy(unsigned char x, unsigned char y, unsigned char length);
```
oliverschmidt, line 92: Some nit picking: The (unusual) cc65 style for function prototypes is to have a space before the parameter list.

include/apple2.h
```diff
@@ -83,7 +83,57 @@
...
+#define _cvline(length)       mt_cvline(length)
+#define _cvlinexy(x,y,length) mt_cvlinexy(x,y,length)
+
+#else
+
+/* Functions that don't depend on MouseText to draw boxes and lines */
+void dyn_chline(unsigned char h, unsigned char length);
+void dyn_cvline(unsigned char v, unsigned char length);
+void dyn_chlinexy(unsigned char h, unsigned char x, unsigned char y, unsigned char length);
+void dyn_cvlinexy(unsigned char v, unsigned char x, unsigned char y, unsigned char length);
```
oliverschmidt, line 117: Some nit picking: The (unusual) cc65 style for function prototypes is to have a space before the parameter list.

colinleroy:
Thanks, fixed. I also added a .gitignore in targettest to quit adding binaries :)

targettest/.gitignore
```diff
@@ -0,0 +1,5 @@
+*.map
+*.o
+*.s
+cbm/*.prg
+seek
```
oliverschmidt, line 5: I assume this file was not intentionally made part of the PR.

PR #2644, opened 2025-05-06
Add additional video modes that were added in ROM Pre Release 43 on CX16

JimmyDansbo:
This is a small change, it just adds the additional video modes that came in prerelease 43 of the Commander X16 ROM.


mrdudz:
"Unable to read response from the server. Please try again later." *shrug*

PR #2645, opened 2025-05-10
Apple2: add dynamic lowercase support

colinleroy:
Hi,

I'm pushing that now before #2641 is merged, so that you can see where I'd like to go with dynamic feature support in conio. 
We can either leave this one alone until 2641 is merged and then come back to this one, or if you're fine with the idea, we can close 2641 and work on this one :-)

Thanks!
Colin


libsrc/apple2/get_ostype.s
```diff
@@ -67,4 +83,13 @@ _get_ostype:
 
         .segment        "INIT"
 
-ostype: .res    1
+ostype:         .res    1
+iie_or_newer:   .res    1
```
oliverschmidt, line 87: I think this - with a different name - want's to be inside an `.ifndef __APPLE2ENH__`.

libsrc/apple2/get_ostype.s
```diff
@@ -19,6 +24,17 @@ initostype:
...
 done:   sta     ostype
+
+        .ifndef __APPLE2ENH__
+        cmp     #$30          ; Apple //e?
+        ror                   ; Carry to high bit
+        sta     iie_or_newer
+
+        bpl     :+            ; Set uppercase mask
+        lda     #$FF
+        sta     uppercasemask
```
oliverschmidt, line 35: It seems to me it shouldn't be the task of get_ostype to set uppercasemask. I don't want get_ostype to become the "initialize everything we likely need somewhere" function.
I believe iie_or_newer - with a different name - should get it's own file. It should get its own constructor that runs after the get_ostype constructor. That constructor checks ostype and sets the - renamed - iie_or_newer accordingly.
In the same way I believe uppercasemask should stay in its own file. It should get its own constructor that runs after the get_ostype constructor. That constructor checks ostype and sets uppercasemask accordingly.

oliverschmidt:
I totally see how we automagically initialize uppercasemask. This means that calling allow_lowercase() isn't necessary anymore when running on a //e. However, there are more machines with lowercase support. Almost all Apple ][+ clones have char ROMs with lowercase chars. And all machines with Videx 80 column cards have lowercase chars in 80 col mode. You may easily argue that the Videx 80 column cards aren't supported by the current CONIO in the first place. But the uppercasemask is also used for STDIO - and STDIO works just fine with the Videx 80 column cards already initialized when the program starts.
Especially the Apple ][+'s "just" equiped with lowercase char ROMs aren't detectable at all. But a program could ask the user if he has lowercase output. Therefore I suggest to keep the allow_lowercase() function just there as it doesn't hurt anybody. Additionally this approach doesn't "unnecessarily" break the builds of programs I added that function call to in the last weeks.
I guess it would be nice to update https://github.com/cc65/cc65/blob/master/include/apple2.h#L252-L254 to say:
"By default lowercase characters are forced to uppercase on an Apple ][+ because a stock Apple ][+ doesn't support lowercase display."

colinleroy:
Closing PR, will open a much shorter oneafter #2641

PR #2650, opened 2025-05-13
Possible segfault in coptcmp.c

ascheepe:
On debian testing cc65 2.19-2 segfaulted while optimizing a recursive function. I'm not familiar with the codebase but I believe L should be checked for NULL here also. This was the output of gdb on debian:

(gdb) run -Oirs /tmp/s64.c
Starting program: /usr/bin/cc65 -Oirs /tmp/s64.c
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".

Program received signal SIGSEGV, Segmentation fault. OptCmp8 (S=0x5555555f88a0) at cc65/coptcmp.c:887
887	                    if (
(gdb) bt
    Max=1) at cc65/codeopt.c:1205
    Max=<optimized out>) at cc65/codeopt.c:1188
    at cc65/codeopt.c:1343
    at cc65/main.c:1069
(gdb) p L
$1 = (CodeLabel *) 0x0

This segfault doesn't occur on the master branch.

### DIFF
--- a/src/cc65/coptcmp.c
+++ b/src/cc65/coptcmp.c
@@ -637,12 +637,12 @@ unsigned OptCmp8 (CodeSeg* S)
                     ** shared branch operation. This prevents the unsafe
                     ** removal of the compare for known problem cases.
                     */
-                    if (
+                    if (L &&
                         /* Jump to branch that relies on the comparison. */
-                        (L->Owner->Info & (OF_CBRA | OF_ZBRA)) ||
+                        ((L->Owner->Info & (OF_CBRA | OF_ZBRA)) ||
                         /* Jump to boolean transformer that relies on the comparison. */
                         (L->Owner->OPC == OP65_JSR &&
-                         (FindBoolCmpCond (L->Owner->Arg)) != CMP_INV)
+                         (FindBoolCmpCond (L->Owner->Arg)) != CMP_INV))
                        )
                     {
                         ++ProtectCompare;


kugelfuhr:
Do you have some source code to reproduce the issue?

And, what does "This segfault doesn't occur on the master branch." mean? Not sure how old 2.19 is (probably very old). The last change to coptcmp.c was in october 2023.

Unfortunately the developers run a "we don't do any releases" strategy which makes it very difficult for ordinary users to get an up-to-date version.

ascheepe:
> And, what does "This segfault doesn't occur on the master branch." mean? Not sure how old 2.19 is (probably very old). The last change to coptcmp.c was in october 2023.

I did a checkout of the master branch to check if the most up-to-date version also segfaulted which it did not. It looks like the debian 2.19 version is from 2020;
`axel@teal:/tmp/src/cc65-2.19/src/cc65$ stat coptcmp.c
  File: coptcmp.c
  Size: 30578           Blocks: 64         IO Block: 4096   regular file
Device: 0,47    Inode: 121851      Links: 1
Access: (0644/-rw-r--r--)  Uid: ( 1000/    axel)   Gid: ( 1000/    axel)
Access: 2025-05-13 18:05:25.589591369 +0200
Modify: 2020-05-30 21:03:15.000000000 +0200
Change: 2025-05-13 18:00:10.393186565 +0200
 Birth: 2025-05-13 18:00:10.369186534 +0200
`
The reason I suspect it might still be an issue is because in the declaration of `LabelName` there's a check for L being NULL while it is accessed later on without checks.

`
629                     } else {
630                         CodeLabel* L = N->JumpTo;
631                         const char* LabelName = L? L->Name : N->Arg;
`
(taken from the latest version)

This is the sample code which segfaults on debian testing:

`#include <stdio.h>
#include <stdlib.h>

#define CLEAR putchar(147)
#define HOME  putchar(19)

unsigned char board[9][9] = {
	{ 5, 3, 0, 0, 7, 0, 0, 0, 0 },
	{ 6, 0, 0, 1, 9, 5, 0, 0, 0 },
	{ 0, 9, 8, 0, 0, 0, 0, 6, 0 },
	{ 8, 0, 0, 0, 6, 0, 0, 0, 3 },
	{ 4, 0, 0, 8, 0, 3, 0, 0, 1 },
	{ 7, 0, 0, 0, 2, 0, 0, 0, 6 },
	{ 0, 6, 0, 0, 0, 0, 2, 8, 0 },
	{ 0, 0, 0, 4, 1, 9, 0, 0, 5 },
	{ 0, 0, 0, 0, 8, 0, 0, 7, 9 },
};

void
writeboard(FILE *f)
{
	unsigned char x, y;

	for (y = 0; y < 9; ++y) {
		if (y > 0 && y % 3 == 0) {
			for (x = 0; x < 9; ++x) {
				if (x > 0 && x % 3 == 0)
					fprintf(f, " + ");
				fprintf(f, "---");
			}
			fprintf(f, "\n");
		}

		for (x = 0; x < 9; ++x) {
			if (x > 0 && x % 3 == 0)
				fprintf(f, " | ");
			fprintf(f, "%2d ", board[y][x]);
		}

		fprintf(f, "\n");
	}
}

unsigned char
isvalid(unsigned char row, unsigned char col, unsigned char n)
{
	unsigned char xoff, yoff, x, y;

	/* A row can't contain the same number more than once. */
	for (x = 0; x < 9; ++x) {
		if (board[row][x] == n)
			return 0;
	}

	/* A column as well. */
	for (y = 0; y < 9; ++y) {
		if (board[y][col] == n)
			return 0;
	}

	/* And a 3x3 box also. */
	xoff = col - (col % 3);
	yoff = row - (row % 3);
	for (y = 0; y < 3; ++y) {
		for (x = 0; x < 3; ++x) {
			if (board[y + yoff][x + xoff] == n)
				return 0;
		}
	}

	return 1;
}

unsigned char
solve(unsigned char y, unsigned char x)
{
	unsigned char n;

	/* Are we done processing the board? */
	if (y == 8 && x == 9)
		return 1;

	/* Next row if at end of one. */
	if (x == 9) {
		++y;
		x = 0;
	}

	/* If the current position isn't free try the next one. */
	if (board[y][x] != 0)
		return solve(y, x + 1);

	/* Valid numbers range from 1..9 */
	for (n = 1; n < 10; ++n) {
		/* And should satisfy sudoku's constrains */
		if (isvalid(y, x, n)) {
			/* Try the number. */
			board[y][x] = n;
			if (solve(y, x + 1))
				return 1;

			/* It didn't work out. */
			board[y][x] = 0;
		}
	}

	return 0;
}

int
main(void)
{
	if (solve(0, 0))
		writeboard(stdout);

	return 0;
}
`

kugelfuhr:
While I agree that the code looks somewhat suspicious, I doubt it's a good idea to fix something without understanding the problem (if there is one).

I think that - as a first step - the reason for the access violation should be determined and why it is no longer present in the current version. Next, there should be a test that shows that the current code still has a problem. Then, it should be fixed.

There's a chance that the (maybe) problematic code you're fixing in this PR cannot be exploited. I have some ideas why this might be the case but an analysis would be in order. It might be a good idea to apply the change from this PR anyway, but not without understanding what's going on.

Just my 2 cents - and I'm not in charge.

ascheepe:
I agree but I'm not sure if I can help with that. What I can see is that the label in question is incsp7 but otherwise I'm clueless about the code :)

PR #2652, opened 2025-05-14
zlib: Make crc32 8 bytes shorter and 3 cycles/byte faster

pfusik:


mrdudz:
Thanks!

(Would be great to have some regression tests for those zlib things :))

pfusik:
> (Would be great to have some regression tests for those zlib things :))

Good point! In 2001, regression tests weren't industry standard. I'll add them when I find time.
I tested my change the good old way :) - manually with samples/gunzip65.

sidneycadot:
Looking at the crc32 implementation, it consumes 1024 bytes of table memory in the bss segment. An alternative is to calculate the crc32 without tables, going bit by bit, which would make the code smaller (getting rid of the table initialization code) and the BSS segment tables. It would of course be a lot slower.

To what extent is the crc32 time-critical for zlib? If it isn't, it may be worth considering switching to a bitwise CRC implementation.

PR #2653, opened 2025-05-14
Fix Atari buffer overrun on keyboard EOF

frotzer:
I was writing a program and wanted it to allow entering multiple lines to `stdin` on the keyboard and finish by typing `<CTRL-3>`, Atari's EOF character. But I discovered this causes the screen to go haywire. I found the bug in `libsrc/atari/read.s`. On `stdin` EOF, the number of characters read (stored in `ICBLL` by `CIOV`) is zero. So the loop at line 155 incorrectly copies 256 bytes instead, overwriting the display list and messing up the screen. My fix is to just jump over the loop if the length is zero. A simple test program is:
```
#include <stdio.h>

void main(void)
{
    printf("\nPress <CTRL-3>: ");
    getchar();
    puts("If you can see this, it worked!");
    for(;;);
}
```

### DIFF
--- a/libsrc/atari/read.s
+++ b/libsrc/atari/read.s
@@ -147,6 +147,7 @@ icbll_copy:
         sta     dataptr+1
         lda     ICBLL,x
         sta     copylen
+        beq     copied          ; length = 0 if EOF
         pha                     ; remember for return value
         ldy     #0
         ldx     index
@@ -159,7 +160,7 @@ copy:   lda     linebuf,x
         bne     copy
 
         pla                     ; length
-        pha                     ; save length to return at okdone
+copied: pha                     ; save length to return at okdone
 
         clc
         adc     index


frotzer:
Whoops, I just realized my test program works fine with  `master` even without my patch. But it definitely goes wrong on  V2.19.

frotzer:
OK, my problem was fixed in 476591e8b7f28bd5145c4e81750a969ce83b97ff so I was about to just say never mind. But that buffer overrun may cause trouble sometime somewhere, so I'll rename this PR and leave it for review anyway.

libsrc/atari/read.s
```diff
@@ -148,6 +148,7 @@ icbll_copy:
         lda     ICBLL,x
         sta     copylen
         pha                     ; remember for return value
+        beq     copied          ; length = 0 if EOF
```
groessler, line 151: You could move the beq in front of the pha, and the label 'copied' past the pla.

frotzer:
Sure, whatever you say.

libsrc/atari/read.s
frotzer: OK, how's this?

frotzer:
Here we go. This program exposes the bug. The `fread(buf, ...)` call overwrites `msg.` But with my patch, `msg` remains untouched.
```
#include <stdio.h>
#include <string.h>

char buf[80], msg[256];

void main(void)
{
    strcpy(msg, "Important data!");
    printf("\nBefore: %s\n", msg);
    printf("Now press <CTRL-3>:");
    fread(buf, 1, 1, stdin);
    printf("After: %s\n", msg);
    for(;;);
}
```

groessler:
Thanks for your contribution!

PR #2657, opened 2025-05-20
Remove the unused function ClearDiagnosticStrBufs()

kugelfuhr:
This change removes a function that is never called.


mrdudz:
thanks!

PR #2658, opened 2025-05-21
Fix -W -unreachable-code not working in all cases

### LINKED ISSUE #2655: cc65: -W -unreachable-code doesn't work in all cases
Switching off the "unreachable-code" warning doesn't work in all cases. This one works:

```bash
$ cat test.c
void f(void)
{
    if (1) {
        f();
    } else {
        f();
    }
}
$ cc65 -W -unreachable-code test.c
$
```

While in this case it doesn't:
```bash
$ cat test.c
void f(void)
{
    if (0) {
        f();
    } else {
        f();
    }
}
$ cc65 -W -unreachable-code test.c
test.c:3: Warning: Unreachable code
$
```

kugelfuhr:
Make sure, the command line setting for the "Unreachable code" warning is checked in all cases before outputting such a warning. 

I've also added a test which took most of the time since the test infrastructure is awkward in some respects.

Fixes #2655.

The diff is against a master with #2657 applied, so it might be necessary to merge #2657 before this one - don't know if it compiles without.


mrdudz:
Thanks!

PR #2661, opened 2025-05-27
Apple2: Safely check for 80-columns card

colinleroy:
Check the Pascal ID bytes for an 80-columns card
to avoid blindly jumping into a $C300 that could be totally unrelated.


libsrc/apple2/detect80cols.s
```diff
@@ -0,0 +1,56 @@
+;
+; Colin Leroy-Mira, 27/05/2025
+;
+; Verify the presence of a 80 columns card in slot 3,
+; and publish a flag accordingly.
+;
+        .export         ext80_present
```
oliverschmidt, line 7: I'm wondering what "ext80" stands for. It reminds me all the time of "extended 80 column card". But that's not what this here is about.

colinleroy: Indeed it's not necessarily extended... I have a hard time coming up with a good variable name (it can't start with a number!) I don't know if c300_is_80col is better?

oliverschmidt: I see. What about 'aux80col'? We're only interested in an 80 column board in the auxilary slot of Apple //e.

oliverschmidt:
Thanks for the fix for this long standing issue 😀

PR #2663, opened 2025-05-28
Apple2: Don't depend on apple2enh definition for keys the //e has

colinleroy:
Up, Down and Del, as well as Open-Apple, exist on
non-enhanced Apple //e.

### DIFF
--- a/include/apple2.h
+++ b/include/apple2.h
@@ -83,6 +83,26 @@
 #define CH_CURS_LEFT    0x08
 #define CH_CURS_RIGHT   0x15
 
+/* These characters are not available on the ][+, but
+ * are on the //e. */
+#if defined(__APPLE2ENH__) || defined(APPLE2_INCLUDE_IIE_CHARS)
+#define CH_DEL          0x7F
+#define CH_CURS_UP      0x0B
+#define CH_CURS_DOWN    0x0A
+
+/* These are defined to be OpenApple + NumberKey */
+#define CH_F1   0xB1
+#define CH_F2   0xB2
+#define CH_F3   0xB3
+#define CH_F4   0xB4
+#define CH_F5   0xB5
+#define CH_F6   0xB6
+#define CH_F7   0xB7
+#define CH_F8   0xB8
+#define CH_F9   0xB9
+#define CH_F10  0xB0
+#endif
+
 #if defined(__APPLE2ENH__)
 
 /* MouseText-based functions for boxes and lines drawing */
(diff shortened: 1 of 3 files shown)


include/apple2.h
```diff
@@ -83,6 +83,24 @@
...
+#define CH_F2   0xB2
+#define CH_F3   0xB3
+#define CH_F4   0xB4
+#define CH_F5   0xB5
+#define CH_F6   0xB6
+#define CH_F7   0xB7
+#define CH_F8   0xB8
+#define CH_F9   0xB9
+#define CH_F10  0xB0
+
```
oliverschmidt, line 103: This isn't the no-brainer it seems to be. You're supposed to be able to write somehting like https://github.com/cc65/cc65/blob/master/libsrc/dbg/dbg.c#L106-L108 where you explicitly define replacements for missing keys. But if the defines for the keys exist then you don't know that anymore that replacements are necessary anymore.
Additionally there's code like https://github.com/FujiNetWIFI/fujinet-contiki-webbrowser/blob/main/src/apple2/contiki-conf.h#L40-L44 around that just assumes the current situation and will simply fail with duplicate defines.
Anyhow, I have right now no good idea what to do better 😒.

libsrc/apple2/cgetc.s
```diff
@@ -51,10 +51,14 @@ put_caret:
 
         ; At this time, the high bit of the key pressed is set.
 :       bit     KBDSTRB         ; Clear keyboard strobe
-        .ifdef __APPLE2ENH__
+
+        .ifndef __APPLE2ENH__
+        bit     machinetype     ; Apple //e or more recent?
+        bpl     clear
+        .endif
```
oliverschmidt, line 58: I remember that many years back when CH_F1 - CHF10 were introduced for the Apple II, there was a community member arguing that it isn't necessary to limit them to the apple2enh targetr because the Apple ][+ user can push his joystick button isntead. "Of course" this wasn't considered reasonable because the user maybe/likely...
* doesn't know
* doesn't have a joystick connetec in the first place
But _IF_ we now decide to move CH_F1 - CHF10 to the apple2.h _THEN_ there's no reason to keep the ][+ user who a) knows and b) has a joystock to successfully type them.

include/apple2.h
colinleroy: Indeed... For the first example, Ctrl-H is equal to CH_DEL in fact. For the second one it would indeed require change in the target program, so I'm indeed not sure it's desirable. In my usecases I do most of my key inputs in cases, so I just do:

```
switch(c) {
    case CH_CURS_UP:
    case 'U'-'A'+1:
      ....
```

But yeah there must be a lot of programs that would break, so I suppose we can drop it and I'll just do it in my .h too :)

colinleroy: Seems like a good idea!

PR #2673, opened 2025-06-03
fixes #2666, double charmap of char literals

### LINKED ISSUE #2666: CX16 – fgets not stopping at line feed
This simple program should take input from the user and print it. Note `fgets` should return after the first Enter, but you usually have to press Enter several times before `fgets` returns. Confirmed in hardware and emulator (R48). 

```C
#include <stdlib.h>
#include <stdio.h>

int main()
{
	char delay[20];

	printf("\nDelay in clock ticks (60/s): ");
	fgets(delay, sizeof(delay) * sizeof(char), stdin);
	printf(delay);

	return EXIT_SUCCESS;
}
```

I cloned the latest master without any pre-existing files and did `make all`. The version reported is cc65 v 2.19 - Git e444f4c40. There are also problems with other stdio functions, will get to those later!

GorillaSapiens:
i think this fixes issue #2666.
removes unnecessary extra translation of character literals that leads to errors on c64.

### DIFF
--- a/src/ca65/scanner.c
+++ b/src/ca65/scanner.c
@@ -1550,7 +1550,6 @@ CharAgain:
                 ** string later.
                 */
                 ReadStringConst ('\'');
-                TgtTranslateStrBuf(&CurTok.SVal);
                 if (SB_GetLen (&CurTok.SVal) == 1) {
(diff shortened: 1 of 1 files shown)


mrdudz:
Thanks, lets see what happens :)

PR #2674, opened 2025-06-04
rename "sp" to "c_sp", avoid conflict with 4510 opcodes

GorillaSapiens:
[edit, i found the problem, see comments after this one]

you should be VERY skeptical of this.  see "PROBLEMS" below.

WHAT I DID

from the cc65 root:
<pre>
find . \( -path './.git' -o -path './.github' -o -path './node_modules' -o -path './venv' -o -path './build' -o -path './dist' -o -path './src/da65' \) -prune \
  -o -type f ! -name '*4510*' -exec sed -i 's/\(^\|[^a-zA-Z0-9_]\)sp\([^a-zA-Z0-9_]\|$\)/\1spc\2/g' {} +
</pre>
then applied this patch, because "\tsp" didn't get picked up by that find/sed magic:
<pre>
diff --git a/src/cc65/codegen.c b/src/cc65/codegen.c
index e55a318d6..a986b4244 100644
--- a/src/cc65/codegen.c
+++ b/src/cc65/codegen.c
@@ -218,7 +218,7 @@ void g_preamble (void)
     AddTextLine ("\t.debuginfo\t%s", (DebugInfo != 0)? "on" : "off");
 
     /* Import zero page variables */
-    AddTextLine ("\t.importzp\tsp, sreg, regsave, regbank");
+    AddTextLine ("\t.importzp\tspc, sreg, regsave, regbank");
     AddTextLine ("\t.importzp\ttmp1, tmp2, tmp3, tmp4, ptr1, ptr2, ptr3, ptr4");
 
     /* Define long branch macros */
</pre>

PROBLEMS

this was not my first attempt at renaming.  at first i tried to rename it to "c_sp" (per pull request #1792),
which failed tests

the optimized version of test/val/add5 failed.  when i compared the assembly output to an unmodified build, it looked like some optimizations weren't being applied, but i could not figure out why.

at first i thought there was some kind of weird problem with the underscore in "c_sp", so i tried "csp" and it also failed, in the exact same way.

on a dumb wild hunch i tried "spc", and that passed all tests.  comparing the optimized test/val/add5 assembly output to an unmodified copy, it looked exactly as i expected, only the "sp" and been renamed to "spc".

TL;DR this works, but i believe there is some kind of weird bug in the optimizer that is affected by the NAME of the variable.

also, of course, this is a TON of change, and should be inspected carefully.


GorillaSapiens:
this is the diff from the failed build when i tried "c_sp", in case it gives anyone any insight
<pre>
--- a/testwrk/val/add5.O.6502.s	2025-06-04 01:11:36.975616369 +0000
+++ b/testwrk/val/add5.O.6502.s	2025-06-04 04:00:40.020729727 +0000
@@ -1,13 +1,13 @@
 ;
-; File generated by cc65 v 2.19 - Git 58171691d
+; File generated by cc65 v 2.19 - Git 3d118dc6e
 ;
-	.fopt		compiler,"cc65 v 2.19 - Git 58171691d"
+	.fopt		compiler,"cc65 v 2.19 - Git 3d118dc6e"
 	.setcpu		"6502"
 	.smart		on
 	.autoimport	on
 	.case		on
 	.debuginfo	off
-	.importzp	sp, sreg, regsave, regbank
+	.importzp	c_sp, sreg, regsave, regbank
 	.importzp	tmp1, tmp2, tmp3, tmp4, ptr1, ptr2, ptr3, ptr4
 	.macpack	longbranch
 	.forceimport	__STARTUP__
@@ -60,15 +60,15 @@ _lPtr:
 ;
 ; cPtr = cP + offset;
 ;
-	ldx     #$00
-	lda     (sp,x)
+	ldy     #$02
+	lda     (c_sp),y
 	clc
-	ldy     #$01
-	adc     (sp),y
+	dey
+	adc     (c_sp),y
 	pha
-	txa
+	lda     #$00
 	iny
-	adc     (sp),y
+	adc     (c_sp),y
 	tax
 	pla
 	sta     _cPtr
@@ -118,11 +118,11 @@ L0002:	jmp     incsp3
 	jsr     ldax0sp
 	clc
 	ldy     #$02
-	adc     (sp),y
+	adc     (c_sp),y
 	pha
 	txa
 	iny
-	adc     (sp),y
+	adc     (c_sp),y
 	tax
 	pla
 	sta     _cPtr
@@ -173,11 +173,11 @@ L0002:	jmp     incsp4
 	jsr     ldeax0sp
 	clc
 	ldy     #$04
-	adc     (sp),y
+	adc     (c_sp),y
 	pha
 	txa
 	iny
-	adc     (sp),y
+	adc     (c_sp),y
 	tax
 	pla
 	sta     _cPtr
@@ -224,18 +224,19 @@ L0002:	jmp     incsp6
 ;
 ; iPtr = iP + offset;
 ;
+	ldy     #$02
 	ldx     #$00
-	lda     (sp,x)
+	lda     (c_sp),y
 	asl     a
 	bcc     L0005
 	inx
 	clc
-L0005:	ldy     #$01
-	adc     (sp),y
+L0005:	dey
+	adc     (c_sp),y
 	pha
 	txa
 	iny
-	adc     (sp),y
+	adc     (c_sp),y
 	tax
 	pla
 	sta     _iPtr
@@ -286,11 +287,11 @@ L0002:	jmp     incsp3
 	jsr     aslax1
 	clc
 	ldy     #$02
-	adc     (sp),y
+	adc     (c_sp),y
 	pha
 	txa
 	iny
-	adc     (sp),y
+	adc     (c_sp),y
 	tax
 	pla
 	sta     _iPtr
@@ -342,11 +343,11 @@ L0002:	jmp     incsp4
 	jsr     aslax1
 	clc
 	ldy     #$04
-	adc     (sp),y
+	adc     (c_sp),y
 	pha
 	txa
 	iny
-	adc     (sp),y
+	adc     (c_sp),y
 	tax
 	pla
 	sta     _iPtr
@@ -393,16 +394,17 @@ L0002:	jmp     incsp6
 ;
 ; lPtr = lP + offset;
 ;
+	ldy     #$02
 	ldx     #$00
-	lda     (sp,x)
+	lda     (c_sp),y
 	jsr     aslax2
 	clc
-	ldy     #$01
-	adc     (sp),y
+	dey
+	adc     (c_sp),y
 	pha
 	txa
 	iny
-	adc     (sp),y
+	adc     (c_sp),y
 	tax
 	pla
 	sta     _lPtr
@@ -453,11 +455,11 @@ L0002:	jmp     incsp3
 	jsr     aslax2
 	clc
 	ldy     #$02
-	adc     (sp),y
+	adc     (c_sp),y
 	pha
 	txa
 	iny
-	adc     (sp),y
+	adc     (c_sp),y
 	tax
 	pla
 	sta     _lPtr
@@ -509,11 +511,11 @@ L0002:	jmp     incsp4
 	jsr     aslax2
 	clc
 	ldy     #$04
-	adc     (sp),y
+	adc     (c_sp),y
 	pha
 	txa
 	iny
-	adc     (sp),y
+	adc     (c_sp),y
 	tax
 	pla
 	sta     _lPtr
@@ -561,7 +563,7 @@ L0002:	jmp     incsp6
 ; cPtr = offset + cP;
 ;
 	ldy     #$00
-	lda     (sp),y
+	lda     (c_sp),y
 	sta     ptr1
 	ldy     #$02
 	jsr     ldaxysp
@@ -617,11 +619,11 @@ L0002:	jmp     incsp3
 	jsr     ldaxysp
 	clc
 	ldy     #$00
-	adc     (sp),y
+	adc     (c_sp),y
 	pha
 	txa
 	iny
-	adc     (sp),y
+	adc     (c_sp),y
 	tax
 	pla
 	sta     _cPtr
@@ -727,7 +729,7 @@ L0002:	jmp     incsp6
 ; iPtr = offset + iP;
 ;
 	ldx     #$00
-	lda     (sp,x)
+	lda     (c_sp,x)
 	asl     a
 	bcc     L0003
 	inx
@@ -903,7 +905,7 @@ L0002:	jmp     incsp6
 ; lPtr = offset + lP;
 ;
 	ldx     #$00
-	lda     (sp,x)
+	lda     (c_sp,x)
 	jsr     aslax2
 	sta     ptr1
 	stx     ptr1+1
</pre>

GorillaSapiens:
ok, i found the issue.

in addition to a monstrous number of changes, the register names appear in src/cc65/codeinfo.c in a structure called ZPInfoTable.

in that same file, this table is bsearch'd.  meaning it must be in order by name.  so changing the name to "c_sp" put it out of order, which confuzzled everything.

fixed now, and using "c_sp" as requested in #1792

GorillaSapiens:
also wanted to note, the reason i was keen on doing this change, is that i found it horribly confusing to see a "sp" in the code that was NOT the 6502 hardware stack pointer.  although technically i guess that's just "S". i still found it confusing.  i'm not overly fond of "c_sp", the underscore is a bit jarring, although i like it more than "sp".  "paramstack" might be better, now that we know how to change it, but that's not really a hill i want to die on, it's rather long.  your call at this point @mrdudz

GorillaSapiens:
it's also worth noting, i did NOT touch anything in da65; i assume a disassembler doesn't know what we named our variables in our *.s files.  i also didn't touch anything named &ast;4510&ast; ; i assume those use whatever opcodes have "sp" in them legitimately.

mrdudz:
> i also didn't touch anything named *4510* ; i assume those use whatever opcodes have "sp" in them legitimately.

Not sure what that means - but the "sp" symbol - ie the software stack pointer - must be renamed all and everywhere, else all hell breaks loose (and its confusing as hell too)

GorillaSapiens:
> > i also didn't touch anything named _4510_ ; i assume those use whatever opcodes have "sp" in them legitimately.
> 
> Not sure what that means - but the "sp" symbol - ie the software stack pointer - must be renamed all and everywhere, else all hell breaks loose (and its confusing as hell too)

if you recall, the whole reason for renaming "sp" is @dillof comment:
> I've run into a problem: cc65 uses the symbol sp, but sp is a register name on the 4510 and 45GS02 CPUs. 

when making this change (replace "sp" with "c_sp"), i skipped the following files:
<pre>
./src/da65/opc4510.c
./src/da65/opc4510.h
./wrk/da65/opc4510.o
./wrk/da65/opc4510.d
./test/asm/cpudetect/4510-cpudetect.ref
./test/asm/opcodes/4510-opcodes.s
./test/asm/opcodes/4510-opcodes.ref
./test/dasm/4510-disass.s
./testwrk/asm/cpudetect/4510-cpudetect.bin
./testwrk/asm/cpudetect/4510-cpudetect.lst
./testwrk/asm/cpudetect/4510-cpudetect.o
./testwrk/asm/opcodes/4510-opcodes.o
./testwrk/asm/opcodes/4510-opcodes.lst
./testwrk/asm/opcodes/4510-opcodes.bin
./testwrk/dasm/4510-reass.s
./testwrk/dasm/4510-reass.bin
./testwrk/dasm/4510-disass.bin
</pre>

under the assumption that any appearances of "sp" here refered to the 4510/45GS02 "sp", and not the cc65 "sp".

i also skipped everything under the src/da65 directory, under the assumption that the disassembler would not know names of zero page variables used by the compiler.  (btw, this seems to be a correct assumption, as da65 only seems to have a single line of code with "sp" in it, and that line is related to the 4510)

mrdudz:
AAAAhh yes ok, i misunderstood you then :) Obviously when it refers to the 4510 Stackpointer, then it should be just "sp" - thats the whole point of this renaming stuff afterall :)

kugelfuhr:
It seems that automatically changing "sp" to "c_sp" destroys the alignment of the comments on the right side. Is it possible to run something like `sed -i -e 's/\(c_sp\),y  /\(c_sp\),y/' -e 's/c_sp  /c_sp/'` to get the comments back in line?

colinleroy:
Hi,
Just a word to point out the obvious, not that I'm against this change, I'll do the work, but this will potentially break a lot of client code using sp ditectly from some asm implementations of c callable functions.

kugelfuhr:
> Hi, Just a word to point out the obvious, not that I'm against this change, I'll do the work, but this will potentially break a lot of client code using sp ditectly from some asm implementations of c callable functions.

How about aliasing `sp` to `c_sp` for all platforms besides 4510? For example by adding a file containing
```
.include "zeropage.inc"
.export sp := c_sp
```
for all other platforms. This would remove the problem with breaking old code.

GorillaSapiens:
> Hi, Just a word to point out the obvious, not that I'm against this change, I'll do the work, but this will potentially break a lot of client code using sp ditectly from some asm implementations of c callable functions.

the find/sed incantation i used is in the first comment.  just replace "spc" with "c_sp".

alternately, there may be a solution where you do some kind of "sp = c_sp" wrapped in a ".if" somewhere in asminc/, i won't claim to be an expert on as65.

mrdudz:
> > Hi, Just a word to point out the obvious, not that I'm against this change, I'll do the work, but this will potentially break a lot of client code using sp ditectly from some asm implementations of c callable functions.
> 
> How about aliasing `sp` to `c_sp` for all platforms besides 4510? For example by adding a file containing
> 
> ```
> .include "zeropage.inc"
> .export sp := c_sp
> ```
> 
> for all other platforms. This would remove the problem with breaking old code.

I don't like doing things like this by default... but its something that should be mentioned prominently in the docs at some strategic place(s) :) Ie i'd rather tell ppl what happened and how to fix it.

kugelfuhr:
There's no harm done by adding it, and it saves a lot (!) of hassle. Or do you expect any problems with such an alias?

kugelfuhr:
Here's an improved proposal. Use
```
.include "zeropage.inc"
.export sp := c_sp
.assert 1, ldwarning, "Symbol sp is deprecated - use c_sp instead"
```
This means everybody using the `sp` symbol will get a warning when linking such a module.

I promise that you will otherwise get flooded with issues since everybody with existing assembler modules that use `sp` (for example to access passed variables) will get errors when compiling.

mrdudz:
> There's no harm done by adding it, and it saves a lot (!) of hassle. Or do you expect any problems with such an alias?

First of all i find it confusing, i wouldn't want this to happen in our library code, for example. I don't like the idea of having to remember what CPU this file is compiled for, only to find out what sp means. And i'd totally expect c_sp being used everywhere, no exceptions.

Providing such module that users can link if needed is another thing (i'd still not make it default though, because it hides errors - ie sp being used in the library when it should be c_sp instead)

kugelfuhr:
You don't have to remember anything. Just use `c_sp` and everything is fine. Supplying `sp` as a fallback in addition to `c_sp` helps users to make the transition. And it doesn't hide any errors within cc65 - a test that checks for imports of `sp` in all library modules is rather easy to write.

You can of course do what you like, but @colinleroy has a point here. You're definitely alienating your user base if you break their code in regular intervals.

mrdudz:
> a test that checks for imports of `sp` in all library modules is rather easy to write.

I'd totally be willing to merge a PR that contains the module(s) as you proposed, plus such a test

It should happen after (and outside) this one in any case - its such a massive patch, i'd like it to address exactly this one thing (renaming the symbol)

mrdudz:
#2730 merged - i'll have a look at some more PRs that you guys swamped me with in the past days, then we can do this one :)

libsrc/runtime/zeropage.s
```diff
@@ -10,7 +10,7 @@
 
 .zeropage
 
-sp:             .res    2       ; Stack pointer
+c_sp:             .res    2     ; Stack pointer
```
mrdudz, line 13: please fix indention

src/dbginfo/dbgsh.c
```diff
@@ -458,7 +458,7 @@ static unsigned FindIdType (const char* TypeName)
         {   "segment",  SegmentId       },
         {   "source",   SourceId        },
         {   "src",      SourceId        },
-        {   "sp",       SpanId          },
+        {   "c_sp",       SpanId          },
```
mrdudz, line 461: please fix the alignment

mrdudz:
Scrolled over it, just some minor "issues" - please fix. Will do some local tests now and then merge

mrdudz:
So lets try what happens :) Got C somewhat working for C65/MEGA65 the other night, so it can't be terribly broken afterall

PR #2675, opened 2025-06-04
added information to Makefile output

GorillaSapiens:
this just adds extra output to the make process.  with so many nested makefiles, i was getting confused by what was going on.  the extra output looks like this:
<pre>
~~~~~~~~~~~~~~~~~~~~~~~~~~~~
~~~ Using Makefile: /home/gorilla/cc65/test/asm/misc/Makefile
~~~ Invoked target: all
~~~~~~~~~~~~~~~~~~~~~~~~~~~~
</pre>


rofl0r:
maybe this could be put into a separate .mk file and included rather than duplicating it everywhere?

GorillaSapiens:
> maybe this could be put into a separate .mk file and included rather than duplicating it everywhere?

that was my initial thought, but doing that becomes MUCH more complex because of how make resolves variables.
<pre>
$(info ~~~ Using Makefile: $(realpath $(firstword $(MAKEFILE_LIST))))
</pre>
becomes a nontrivial monster in the "included file" case.

spiro-trikaliotis:
Can this be made configurable? There has been much work by Oliver "back then" to make the makefiles quiet, and thus patch does the opposite.
Yes, it has a value, but please do it only if it is actually needed.

mrdudz:
> Can this be made configurable? There has been much work by Oliver "back then" to make the makefiles quiet, and thus patch does the opposite. Yes, it has a value, but please do it only if it is actually needed.

I'd rather see the makefiles being even more quiet for that matter (somehow not everything is suppressed that could be - no idea why, i don't see a common pattern). i don't see a difference between using QUIET=1 or =0 on the cmdline either (but the makefiles suggest there should be) - its just always "quiet" :) (i'd prefer a guarded .silent: too, instead of using @ for every command)

But yes, please make it configurable (perhaps via VERBOSE=1 - so it doesn't clash with the apparently not really working QUIET)

rofl0r:
> $(info ~~~ Using Makefile: $(realpath $(firstword $(MAKEFILE_LIST))))
> becomes a nontrivial monster in the "included file" case.

maybe that just needs `$(secondword)` instead of `$(firstword)` - or equivalent ?

GorillaSapiens:
@spiro-trikaliotis @mrdudz 

if you want a quiet build, just "make -s".
<pre>
$ man make
[...]
-s, --silent, --quiet
Silent operation; do not print the commands as they are executed.
[...]
</pre>

there is no reason to add anything to Makefiles to make them more quiet.  why ask for "quiet" when you as a human are just ignoring the output anyway?

however, developer sanity is an excellent reason to add a little extra output to make it easier to interpret what's going on with multiple nested makefiles, as we have in our automated testing.

mrdudz:
-s wouldnt suppress that extra output at all - you'd get all that output, all the time, for every makefile, even if nothing was done at all.

GorillaSapiens:
> -s wouldnt suppress that extra output at all - you'd get all that output, all the time, for every makefile, even if nothing was done at all.

how would you feel about it only in the test directories?

GorillaSapiens:
btw, who is "travis" and why is there a Makefile.travis and a .travis.yml ???

mrdudz:
travis is yet another CI provider, which we used before github actions. Its not used right now

mrdudz:
oh and, bzw - there is another reason for why this is a questionable thing to do - it doesn't really work. Have a look at the log output of the test run here:

https://github.com/cc65/cc65/actions/runs/15432587980/job/43432998675?pr=2675

GorillaSapiens:
> oh and, bzw - there is another reason for why this is a questionable thing to do - it doesn't really work. Have a look at the log output of the test run here:
> 
> https://github.com/cc65/cc65/actions/runs/15432587980/job/43432998675?pr=2675

that's a lot to look at.  can you point out specifically what it is that concerns you?

mrdudz:
nevermind, brainfart - i didn't see its all "make clean" (i meant the output at the beginning of "run the regression tests")

In any case, that output shows why i don't really like this.... its a lot of output with no practical use.

What problem are you trying to solve here for that matter? Lets say the tests error out here:

https://github.com/cc65/cc65/actions/runs/15432587980/job/43432998675?pr=2675#step:9:6051

The last output was
```
val/remove.Or.65c02.prg
```
That pretty much tells you its in test/val/ and the Makefile is in this directory. Do we really need to be able to scroll up over 5000 (!) lines to read
```
~~~ Using Makefile: /home/runner/work/cc65/cc65/test/val/Makefile
```
which tells the same?

GorillaSapiens:
how would you feel about making it less obtrusive?  maybe just printing
<pre>
make -f /home/gorilla/cc65/test/asm/misc/Makefile all
</pre>
i realize you can get this information with "make -d", but that spews a lot of other stuff too

mrdudz:
What about something like
```
SILENT:=$(findstring s,$(word 1, $(MAKEFLAGS)))
ifneq ($(SILENT),s)
    $(info Using Makefile: $(realpath $(firstword $(MAKEFILE_LIST))) $(MAKECMDGOALS))
endif
```

That will add one line to the regular output, and remove it when using -s

($(SILENT) could perhaps be reused elsewhere too)

mrdudz:
That way its acceptable, i guess. thanks!

Issue #2681, opened 2025-06-08
src/dbginfo is hopelessly broken


GorillaSapiens:
src/dbginfo doesn't compile "as is", with or without -I../common

bare minimum, this needs a working Makefile
it should also be included in the build.

if it is intended to work some other way than what i have here,
then that also needs to be documented,
and additional code / Makefiles should be included to demonstrate.

if we're NOT going to give that minimum level of support,
it should simply be removed altogether.

<pre>
gorilla@guest:~/cc65/src/dbginfo (master)$ gcc -g  *.c -o dbgsh
dbgsh.c:44:10: fatal error: attrib.h: No such file or directory
   44 | #include "attrib.h"
      |          ^~~~~~~~~~
compilation terminated.
</pre>

<pre>
gorilla@guest:~/cc65/src/dbginfo (master)$ gcc -g -I../common *.c -o dbgsh
dbgsh.c:60:22: error: unknown type name ‘Collection’
   60 | static void CmdHelp (Collection* Args attribute ((unused)));
      |                      ^~~~~~~~~~
dbgsh.c:63:22: error: unknown type name ‘Collection’
   63 | static void CmdLoad (Collection* Args);
      |                      ^~~~~~~~~~
dbgsh.c:66:22: error: unknown type name ‘Collection’
   66 | static void CmdQuit (Collection* Args attribute ((unused)));
      |                      ^~~~~~~~~~
dbgsh.c:69:22: error: unknown type name ‘Collection’
   69 | static void CmdShow (Collection* Args);
      |                      ^~~~~~~~~~
dbgsh.c:72:26: error: unknown type name ‘Collection’
   72 | static void CmdShowHelp (Collection* Args);
      |                          ^~~~~~~~~~
dbgsh.c:75:33: error: unknown type name ‘Collection’
   75 | static void CmdShowChildScopes (Collection* Args);
      |                                 ^~~~~~~~~~
dbgsh.c:78:29: error: unknown type name ‘Collection’
   78 | static void CmdShowCSymbol (Collection* Args);
      |                             ^~~~~~~~~~
dbgsh.c:81:30: error: unknown type name ‘Collection’
   81 | static void CmdShowFunction (Collection* Args);
      |                              ^~~~~~~~~~
dbgsh.c:84:29: error: unknown type name ‘Collection’
   84 | static void CmdShowLibrary (Collection* Args);
      |                             ^~~~~~~~~~
dbgsh.c:87:26: error: unknown type name ‘Collection’
[...]

PR #2682, opened 2025-06-08
Hook dbginfo into build by compiling dbgsh

GorillaSapiens:
Adds a Makefile for dbgsh, to ensure this gets built, at least on the linux side.
Calls this Makefile from src/Makefile

this does NOT include a dbginfo.vcxproj ; someone with that build environment will need to do that.
partially addresses issue #2681 
at least now linux developers will know if they do something that breaks dbginfo.

i'm sure someone from team mediocrity will complain about the Makefile contents not being a horrible lumbering mess like other Makefiles in the build system.  i encourage you merge this PR anyway, because it is better than nothing.  i'm sure someone who cares can modify it to their liking in the future.  but for right now the primary goal is to get dbginfo hooked in so we can tell when/if someone breaks it.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -15,6 +15,7 @@ PROGS = ar65     \
         cl65     \
         co65     \
         da65     \
+        dbginfo  \
         grc65    \
         ld65     \
         od65     \
--- a/src/dbginfo/Makefile
+++ b/src/dbginfo/Makefile
@@ -1,0 +1,21 @@
+CC=gcc
+CFLAGS=-g -Wall
+IFLAGS=-I../common
+
+SRCS=\
+    ../common/check.c \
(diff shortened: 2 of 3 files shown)


mrdudz:
> i'm sure someone from team mediocrity will complain about the Makefile contents not being a horrible lumbering mess like other Makefiles in the build system.

If you mean that they must work on both Windows (cmd.exe) and Linux - yes indeed (i'm surprised that the CI passes, somehow the Windows runner must understand mkdir -p now)

mrdudz:
Oh and it shouldn't be part of the "all" target - it should be triggered via "check" (this isn't some tool that belongs to the toolchain) and then perhaps needs to be added to the GHA

mrdudz:
superseded by #2697

PR #2683, opened 2025-06-09
fix bsearch tables

mrdudz:
- fix bsearch tables that must be sorted
- add comment to all tables that must be sorted


GorillaSapiens (commented):
sorting the instantiations at line 106-220 has no affect on the order for bsearch.

mrdudz:
> sorting the instantiations at line codeopt.c:106-220 has no affect on the order for bsearch.

Obviously. But why wouldn't you want both sorted the exact same way in the first place...

PR #2684, opened 2025-06-09
simple script(s) to check if bsearch tables are sorted correctly

mrdudz:
So this would be my proposal for how to test if those tables are sorted correctly.

Yes, this is more "fragile" if you want to call it that - however, i don't think it's a big deal. It will break when someone refactors certain things - however, this will most likely happen exactly once at most in the future - and it is trivial to update the test to make it pass again in that case as well.

On the other hand, it keeps the source clean, has no penalties whatsoever, and even tests that one table in dbginfo :)


GorillaSapiens:
fragile.  what happens when someone renames an array?  what happens when someone creates a new array?  what happens when someone moves an array into a different file? what happens when someone removes an array entirely?

do this instead:
before each array add a comment "/&ast; BEGIN BSEARCH CHECKING -- do not remove this comment &ast;/"
at the end of each array, add a comment "/&ast; END BSEARCH CHECKING -- do not remove this comment &ast;/"
modify the scripts to scan ALL files for pairs of those BEGIN / END comments (error if you find a BEGIN w/o END or vice versa)
scripts should ensure "sortedness" between pairs of BEGIN / END comments

this makes it much easier to add / move / change / whatever in the future, and you won't have to revisit your scripts again.

GorillaSapiens:
while you're in there, can you add "checkstyle" and "sorted" to "all" ?  it'll save developers time when doing PRs

.github/checks/sorted.sh
```diff
@@ -0,0 +1,117 @@
...
+    awk '/'"$2"'/{flag=1;next}/};/{flag=0}flag' "$CHECK_FILE" | \
+        sed -e 's:.*\"\(.*\)\".*:\1:g' | \
+        sed '/^\s*$/d' > .a.tmp
+
+    if [[ -z $(grep '[^[:space:]]' .a.tmp) ]] ; then
+        echo "error: "$2" table is empty"
+        exit -1
+    fi
+
+    LC_COLLATE=C sort $SORT_OPT .a.tmp > .b.tmp
```
kugelfuhr, line 20: Not sure if this is true for the `sort` command on github, but the one used on common Linux distributions has a `-c` command line switch that checks if the input is sorted. This would simplify the script and remove the need for a second temporary file.

kugelfuhr:
> while you're in there, can you add "checkstyle" and "sorted" to "all" ? it'll save developers time when doing PRs

These targets require bash as far as I know and would not run on Windows.

GorillaSapiens:
in codeopt.c, if you move Name out of OptFunc and into the array, you can make the array like all the others and eliminate the special case CODEOPT.SH altogether.

GorillaSapiens:
you'll be more robust if you just match on "BEGIN SORTED.SH" and not the whole ="\\/\\* BEGIN SORTED.SH \\*\\/"; someone in the future may decide that "//" style line comments are OK after all.

this would also allow you to add additional text, like "do not remove this comment" or "comment used to mark bsearch'd arrays".

GorillaSapiens:
> in codeopt.c, if you move Name out of OptFunc and into the array, you can make the array like all the others and eliminate the special case CODEOPT.SH altogether.

for the record, i tried this, twice, two different ways.  it's a TERRIBLE idea. it blows up incredibly quickly in ways that are not easy to resolve.

PR #2686, opened 2025-06-10
Fix temporary file name creation.

### LINKED ISSUE #2685: Hard-to-read .map file
Hi,

PR #2660 makes the .map file rather unusable if one wants to check the size of a built file:
```
Modules list:             
-------------
fileHRNxzx.o:             
    CODE              Offs=000000  Size=000003  Align=00001  Fill=0000
    RODATA            Offs=000000  Size=000000  Align=00001  Fill=0000
    BSS               Offs=000000  Size=000000  Align=00001  Fill=0000
    DATA              Offs=000000  Size=000000  Align=00001  Fill=0000
    ZEROPAGE          Offs=000000  Size=000000  Align=00001  Fill=0000
    NULL              Offs=000000  Size=000000  Align=00001  Fill=0000
fileyoUnKU.o:             
    CODE              Offs=000003  Size=0000A2  Align=00001  Fill=0000
    RODATA            Offs=000000  Size=000000  Align=00001  Fill=0000
    BSS               Offs=000000  Size=00038C  Align=00001  Fill=0000
    DATA              Offs=000000  Size=000008  Align=00001  Fill=0000
    ZEROPAGE          Offs=000000  Size=000000  Align=00001  Fill=0000
    NULL              Offs=000000  Size=000000  Align=00001  Fill=0000
filexmE6p1.o:             
...
```

colinleroy:
Use the original name as template for readability in the map file, and use mkstemps, even if it means we have to unlink the created file.

I hope Windows has mkstemps.
Would fix #2685


mrdudz:
won't work in windows (but the idea is good...)

BUT: isn't the name of the source file(s) also baked into the object file(s)? so you can make the linker print something like

47543w695.o (mysource.s):

by reading the .o file (which it does anyway, i guess)?

(sorry if that is a dumb assertion, i have actually never looked at the object format yet :) but i'd expect this kind of thing to be there - because why not)

colinleroy:
> won't work in windows (but the idea is good...)

Realizing that now and trying to work around it :)

> BUT: isn't the name of the source file(s) also baked into the object file(s)? so you can make the linker print something like
> 47543w695.o (mysource.s):
> by reading the .o file (which it does anyway, i guess)?

Possibly, but we also use that MakeTmpFilename() to assemble from .c to .s!

colinleroy:
I think I got the build part right for Windows, but I'll need a volonteer to runtime-test it, I don't have a Windows here...

src/common/fname.c
```diff
@@ -119,35 +129,49 @@ char* MakeFilename (const char* Origin, const char* Ext)
...
+#endif
+
+    /* Allocate template */
+    Out = xmalloc (strlen (Origin) + strlen (".XXXXXX") + strlen (Ext) + 1);
+    strcpy (Out, Origin);
+
+    /* Add template markers */
+    strcat (Out, ".XXXXXX");
+
+#if defined(_MSC_VER)
```
GorillaSapiens, line 152: having different code paths for different systems is fragile.  if you don't mind using the current directory, you could just construct a file name with sprintf() and lrand48().  be sure to srand48() with a process id or the current time, and you should be fine.

colinleroy: You're right, at that point, it's simpler. Updated commit.

mrdudz: > using the same directory was only a problem before because the file names were also the same. now that the file names are different, i think using the same directory will be less of a problem.

What is "same directory" at this point anyway - directory of the source- or the destination file? same dir as destination is no problem, i agree. same dir as source is quirky though

colinleroy:
True, working directory is better for out of tree builds for example. I'll see how to strip the path part later.

mrdudz:
Another nitpick: what is the seeding with time+pid trying to achive? ie why is it important to get "unique" random numbers (which this doesn't really achieve either - only by chance). Wouldn't it be smarter to eg just append/prepend the literal pid to the generated name, for example?

rofl0r:
> what is the seeding with time+pid trying to achive? ie why is it important to get "unique" random numbers (which this doesn't really achieve either - only by chance). Wouldn't it be smarter to eg just append/prepend the literal pid to the generated name, for example?

indeed, using a unix timestamp with second granularity is useless, one can probably run several thousand instances of cc65 during that second, and all would get the same seed. the pid is a bit better, but e.g. on linux a pid is a 16bit integer, so if one would start 64K instances of cc65 during a make invocation, a collision is very likely. in a [similar case](https://github.com/rofl0r/proxychains-ng/commit/092d7042e092a033ac0c33a238927050c2cc7de0) where i needed a unique seed for srand, i chose to use clock_gettime with nanosecond granularity where available.

GorillaSapiens:
From a security perspective, PID is sequential and predictable. If you know
you're on a Linux machine, /dev/random is available. The important thing
for us is that two processes not pick the same "random" number. Which they
absolutely would if they didn't seed the RNG. I think you're right though,
for our purposes PID is probably sufficient. Two processes running at the
same time cannot have the same PID. And I doubt a 6502 cross compiler is
really going to be targeted by a bad actor.

rofl0r:
> From a security perspective, PID is sequential and predictable.

that's only relevant though if you use /tmp where everyone has access. and with this PR we use the current working directory.

> If you know you're on a Linux machine, /dev/random is available.

that's correct, but clock_gettime should be supported across all POSIX platforms, and is therfore more portable. however that's irrelevant because:

> Two processes running at the same time cannot have the same PID.

indeed (unless cc65 becomes multi-threaded, and that will never happen...), so a simple getpid() should be a sufficient "PRNG" to make a unique temp filename for our usecase.

GorillaSapiens:
Agree. 💯

colinleroy:
Hi,
Took these last comments into account and modified the function:
- strip the path part of the original filename so the output is in the working directory
- use only the PID to make the temporary filenames unique.

groessler:
Working directory is bad. Better use output directory.

mrdudz:
> Which they absolutely would if they didn't seed the RNG. I think you're right though, for our purposes PID is probably sufficient.

My point is that, if the idea is to _guarantee_ unique names - then using random numbers this way is not sufficient either. It would only work most of the time, by chance - which is terrible design.

Making the pid part of the name should combat this problem - at least for about 2^15 processes. If we want to improve this further, adding the timestamp as well should do the trick and work 100% (unless you can spawn 2^15 processes really really quick - which isnt going to happen anytime soon). No idea if we want to go this far :)

In any case, we don't really need randomness at all UNLESS we care about security (whatever that means in this context) as well.

> Working directory is bad. Better use output directory.

Yes! The source directory must be considered "read only" (it may actually be)

colinleroy:
> > Working directory is bad. Better use output directory.
> 
> Yes! The source directory must be considered "read only" (it may actually be)

I can add that, however, have no idea how to get output directory from fname.c::MakeTmpFilename(). Can anyone point me in the right direction?

I'd also like to point out that at this point, this PR outputs temporary files in the same place as previously, and its aim is to fix the unrecognizable filenames in the .map files (issue #2685).

mrdudz:
> I can add that, however, have no idea how to get output directory from fname.c::MakeTmpFilename(). Can anyone point me in the right direction?

not me.

> I'd also like to point out that at this point, this PR outputs temporary files in the same place as previously, and its aim is to fix the unrecognizable filenames in the .map files (issue https://github.com/cc65/cc65/issues/2685).

Right. If there is no easy way to get that path, we could merge as is. Perhaps put a FIXME comment in there though, which says that we ideally want to put the file into either temp- or destination directory

colinleroy:
> Right. If there is no easy way to get that path, we could merge as is. Perhaps put a FIXME comment in there though, which says that we ideally want to put the file into either temp- or destination directory

I found the way :)

mrdudz:
nice, case closed then :) It'd still be cleaner to produce those files in the temp dir - but that'd be mostly cosmetics (and i don't know an OS agnostic way for this offhand either)

cheers!

cleroymira-unabiz:
Uuuugh, @mrdudz we're not done yet :)
Moving from the source directory to the target directory has a side-effect I didn't catch (because I didn't rebuild that particular project):

```
cl65 ... api/compose.c ... cli/compose.c ...

cl65: Duplicate file in argument list: 'compose.c.256412.o'
cl65: Duplicate file in argument list: 'compose.c.256412.o'
```

We can't have files with the same name in different directories anymore. How should we fix it?
One suggestion: keep the path, replacing the separators with _, so that temp files are still created in the output directory, but in the above case, with names like api_compose.c.256412.o and cli_compose.c.256412.o

rofl0r:
maybe if of more than one file is processed by one cc65 instance, append a running counter value ?

// edit that was a reply to colin's example of cc64 foo/bar.c bar/bar.c which somehow disappeared?

colinleroy:
> maybe if of more than one file is processed by one cc65 instance, append a running counter value ?
> 
> // edit that was a reply to colin's example of cc64 foo/bar.c bar/bar.c which somehow disappeared?

Sorry for that comment-flash, rofl0r. I dropped it because I made it a PR :) (#2699)

Issue #2363, opened 2024-01-16
Wishlist: a __CPU__ macro 


colinleroy:
It would be nice to have a __CPU__ macro that does the same as `.if (.cpu .bitand ::CPU_ISET_65SC02)` for C files with inline __asm__.

For exemple, we could cut down on two bytes of ntohs:
```
#ifdef __CPU__65SC02__
#define ntohs(x)                \
    (                           \
        __AX__=(x),             \
        asm("pha"),        \
        asm("txa"),             \
        asm("plx"),        \
        __AX__                  \
    )#else
#define ntohs(x)                \
    (                           \
        __AX__=(x),             \
        asm("sta tmp1"),        \
        asm("txa"),             \
        asm("ldx tmp1"),        \
        __AX__                  \
    )
#endif
```

rofl0r:
> `#ifdef __CPU__65SC02__`

maybe a form like this would be more versatile: `#if __CPU__ >= __CPU_65C02__`
or even `#if (__CPU__ < __CPU_65C02__) && (__CPU__ & __CPU_ILLEGAL_OPCODES_ENABLED__ != 0)`

mrdudz:
> maybe a form like this would be more versatile: #if __CPU__ >= __CPU_65C02__
or even #if (__CPU__ < __CPU_65C02__) && (__CPU__ & __CPU_ILLEGAL_OPCODES_ENABLED__ != 0)

That makes no sense at all. (Neither does the bitfield)

rofl0r:
maybe it makes sense if you view it in ascending order of capabilities, such as in my 6502 emulator: https://github.com/rofl0r/jg6502/blob/master/cpu65.c#L15-L19
0 = 2a03, base 6502 without BCD, 1 = 6502, base 6502 with illegal opcodes, and so forth.

mrdudz:
> maybe it makes sense if you view it in ascending order of capabilities

thats exactly what doesn't make sense

mrdudz:
That said, a macro that tells you what CPU the target uses makes a lot of sense. I am surprised it doesn't exist :)

colinleroy:
I may add it myself at some point :) (but not right now, right now I'm having too much pending branches, it's getting difficult to remember which version I last built)

PR #2687, opened 2025-06-10
issue #2363, add __CPU_*__ defines to preprocessor

GorillaSapiens:
what it says on the tin. ;)

### DIFF
--- a/src/cc65/main.c
+++ b/src/cc65/main.c
@@ -317,6 +317,64 @@ static void SetSys (const char* Sys)
 
 
 
+static void SetCpuMacros (void)
+/* Defines for CPU */
+{
+    switch (CPU) {
+
+        case CPU_NONE:
+            DefineNumericMacro ("__CPU_NONE__", 1);
+            break;
+
+        case CPU_6502:
+            DefineNumericMacro ("__CPU_6502__", 1);
+            break;
+
+        case CPU_6502X:
+            DefineNumericMacro ("__CPU_6502X__", 1);
+            break;
+
+        case CPU_6502DTV:
+            DefineNumericMacro ("__CPU_6502DTV__", 1);
+            break;
+
+        case CPU_65SC02:
+            DefineNumericMacro ("__CPU_65SC02__", 1);
+            break;
+
+        case CPU_65C02:
+            DefineNumericMacro ("__CPU_65C02__", 1);
+            break;
+
+        case CPU_65816:
+            DefineNumericMacro ("__CPU_65816__", 1);
+            break;
+
+        case CPU_SWEET16:
+            DefineNumericMacro ("__CPU_SWEET16__", 1);
+            break;
+
+        case CPU_HUC6280:
+            DefineNumericMacro ("__CPU_HUC6280__", 1);
+            break;
+
+        case CPU_M740:
+            DefineNumericMacro ("__CPU_M740__", 1);
+            break;
+
+        case CPU_4510:
+            DefineNumericMacro ("__CPU_4510__", 1);
+            break;
+
+        case CPU_UNKNOWN:
+        default:
+            DefineNumericMacro ("__CPU_UNKNOWN__", 1);
+            break;
+    }
+}
+
+
+
 static void FileNameOption (const char* Opt, const char* Arg, StrBuf* Name)
 /* Handle an option that remembers a file name for later */
 {
@@ -1071,6 +1129,8 @@ int main (int argc, char* argv[])
             CPU = CPU_6502;
         }
     }
+
+    SetCpuMacros();
 
     /* If no memory model was given, use the default */
     if (MemoryModel == MMODEL_UNKNOWN) {


kugelfuhr:
I would suggest to also add something like the `.cpu` and `.cpu_iset_xxx` symbols that are available in ca65. If the CPU macros are really used in C code then in most cases to check if certain instructions are available - which is easier using something similar to what is implemented in the assembler.

GorillaSapiens:
> I would suggest to also add something like the `.cpu` and `.cpu_iset_xxx` symbols that are available in ca65. If the CPU macros are really used in C code then in most cases to check if certain instructions are available - which is easier using something similar to what is implemented in the assembler.

No.  this patch meets the ask of #2363.  specifically, it mirrors the value of the "Cpu" global in cc65, and enables the code example given in #2363. if you wanted something more / different, perhaps you should have commented on #2363.   your desire for more / different features should not prevent this from being merged.  once it is merged, you are perfectly able to create a new PR adding whatever additional #defines you want.

kugelfuhr:
The problem with cpu specific defines as they are implemented in this PR is that using them is getting messy in many cases. A common usage is to check if certain instructions are available. When using the defines as they are implemented in this PR, you will have to do something like

```asm
#if defined(__CPU_65SC02__) || defined(__CPU_65C02__) || defined(__CPU_65816__) || defined(__CPU_4510__) || defined(__CPU_HUC6280__)
    asm("phx");
#else
    asm("txa");
    asm("pha");
#endif
````

And this will fail if more cpus are added that support `phx`. That's the reason for the assembler using bitsets to describe cpu capabilities. So using

```asm
#if (__CPU__ & __CPU_ISET_65SC02__)
    asm("phx");
#else
    asm("txa");
    asm("pha");
#endif
````
is not only shorter, it is also more future proof. Such macros will also match nicely the ones available in the assembler making the use more orthogonal.

@mrdudz Let me know if you want an improved PR.

mrdudz:
> @mrdudz Let me know if you want an improved PR.

I'd indeed prefer a PR that:
- defines the ```__CPU__``` and ```__CPU_ISSET_``` macros
- additionally define a ```__CPU_xy__``` macro for the current CPU, using the same value(s) used for the former (ie ```__CPU_xy__``` for the current CPU is the same as ```__CPU__```)

That should give us the best of the two worlds and cover all usecases

GorillaSapiens:
Look. I'm tired of you trying to scope creep everything I do. This meets
the ask. Do you want something better? Do it yourself. I'm not your trained
monkey. Can it be improved? Yes, absolutely. Everything can be improved.
Merging this PR will not prevent anyone from changing or adding something
later. "Perfect is the enemy of good." If you want something specific,
maybe you should put it in an issue, so people can do what you want, rather
than waiting until someone does something and then pissing on it just to
mark your territory. I'm not psychic. You're not helping. You're just
sucking all of the air out of the room.

mrdudz:
> Look. I'm tired of you trying to scope creep everything I do. This meets the ask. Do you want something better? Do it yourself. I'm not your trained monkey. Can it be improved? Yes, absolutely. Everything can be improved. Merging this PR will not prevent anyone from changing or adding something later. "Perfect is the enemy of good." If you want something specific, maybe you should put it in an issue, so people can do what you want, rather than waiting until someone does something and then pissing on it just to mark your territory. I'm not psychic. You're not helping. You're just sucking all of the air out of the room.

Let me quote contributing.md - which i am sure you have read:
> Also, before you put a lot of work into implementing something you want to contribute, please get in touch with one of the developers and ask if what you are going to do is actually wanted and has a chance of being merged. Perhaps someone else is already working on it, or perhaps what you have in mind is not how we'd expect it to be - talking to us before you start might save you a lot of work in those cases.

We are not going to merge stuff 1:1 if it doesn't meet our expectations. We are going to discuss things and fix them up until they do, and then merge eventually. This is how we work. And no, those tickets are often not perfectly describing what is wanted and what we expect. No point in ranting about it every time.

mrdudz:
Oh and - this PR should add the added macros to the documentation as well, of course

GorillaSapiens:
> We are not going to merge stuff 1:1 if it doesn't meet our expectations. We are going to discuss things and fix them up until they do, and then merge eventually. This is how we work. And no, those tickets are often not perfectly describing what is wanted and what we expect. No point in ranting about it every time.

This particular issue had been sitting idle for over a year. I would think that, if someone wanted to say something, they would have done so by now. 

Reading your response, I now understand why you have over 200 issues spanning back over 10 years, and why you have pull requests that have been sitting idle for 9 years. 

I apologize for thinking that you actually wanted to get things done. This will certainly change how I approach this project in the future.

kugelfuhr:
@GorillaSapiens A problem I see is that you constantly refuse to pick up suggestions for improvements. Not all of my suggestions are good, and there are different opinions about others - which is ok. I do not expect anything I suggest to be implemented, but it's usually worth discussing.

In the given case what I suggested has many advantages and wasn't much effort to implement. And - if you read #2363 - it was actually explicitly requested:

> It would be nice to have a CPU macro that does the same as ```.if (.cpu .bitand ::CPU_ISET_65SC02)```

In C with the changes from #2695 you can now write ```#if (__CPU__ & __CPU_ISET__65SC02__)```. That's as similar to the request as possible.

I'm not trying to work against you. But I think, if it's worth doing something, it's also worth doing it well. This means discussing changes to find the best solution that can be implemented with reasonable effort. I would be glad if we could work together next time.

mrdudz:
Its clear now that you never worked on a OSS project :)

superseeded by #2695

mrdudz:
> I'm not trying to work against you. But I think, if it's worth doing something, it's also worth doing it well. This means discussing changes to find the best solution that can be implemented with reasonable effort. I would be glad if we could work together next time.

This. We are discussing things to make them better - and *usually* it is not terribly hard to do stuff - like the Makefile in #2682 - it's trivial to make it work the same as the others. And thats why we want it to be like that from the start. Maybe if you can't do something, for whatever reason, just ask if someone else can do it - instead of telling everyone how shit they are.

colinleroy:
> Look. I'm tired of you trying to scope creep everything I do. This meets the ask. Do you want something better? Do it yourself. I'm not your trained monkey. Can it be improved? Yes, absolutely. Everything can be improved. Merging this PR will not prevent anyone from changing or adding something later. "Perfect is the enemy of good." If you want something specific, maybe you should put it in an issue, so people can do what you want, rather than waiting until someone does something and then pissing on it just to mark your territory. I'm not psychic. You're not helping. You're just sucking all of the air out of the room.

I have pull requests here with 74, 43, dozens of comments. When you arrive in a project and want to contribute, it's good to listen to the long-term maintainers. They're the ones that know everything about the thing you're just starting to scratch the surface of. That's how you get better. (That's also how you get code merged instead of having to maintain a subpar fork).

GorillaSapiens:
Perfectly reasonable PR that meets the issue as stated. Closed because you don't like me. Pathetic.

oliverschmidt:
@GorillaSapiens
The first comment starts with `I would suggest [...]`. Your answer starts with `No.` At this point you already disqualified yourself as member of this community. Thank you in advance for moving on and not further wasting anybody's time here.

PR #2689, opened 2025-06-10
issue #1640, print macros

GorillaSapiens:
add printing of macros.  removed debugging info about hash table statistics as it isn't really interesting or useful.  i think we're pretty sure the hash table works by now.

### DIFF
--- a/src/cc65/compile.c
+++ b/src/cc65/compile.c
@@ -573,7 +573,7 @@ void Compile (const char* FileName)
     DonePreprocess ();
 
     if (Debug) {
-        PrintMacroStats (stdout);
+        PrintMacroDetails (stdout);
     }
 
     /* Print an error report */
--- a/src/cc65/macrotab.c
+++ b/src/cc65/macrotab.c
@@ -329,24 +329,42 @@ int MacroCmp (const Macro* M1, const Macro* M2)
 
 
 
-void PrintMacroStats (FILE* F)
-/* Print macro statistics to the given text file. */
+void PrintMacroDetails (FILE* F)
+/* Print macro details to the given text file. */
 {
     unsigned I;
+    int J;
     Macro* M;
 
-    fprintf (F, "\n\nMacro Hash Table Summary\n");
     for (I = 0; I < MACRO_TAB_SIZE; ++I) {
-        fprintf (F, "%3u : ", I);
         M = MacroTab [I];
         if (M) {
             while (M) {
-                fprintf (F, "%s ", M->Name);
+                if (M->ParamCount == -1) {
+                    fprintf (F, "#define %s", M->Name);
+                    if (M->Replacement.Len) {
+                        fprintf(F," %.*s", M->Replacement.Len, M->Replacement.Buf);
+                    }
+                    fprintf (F, "\n");
+                }
+                else {
+                    fprintf (F, "#define %s(", M->Name);
+                    for (J = 0; J < M->ParamCount; ++J) {
+                        if (J > 0) {
+                            fprintf(F, ",");
+                        }
+                        fprintf(F, "%s", (char *) CollAtUnchecked (&M->Params, J));
+                    }
+                    fprintf (F, ")");
+                    if (M->Replacement.Len) {
+                        fprintf(F," %.*s", M->Replacement.Len, M->Replacement.Buf);
+                    }
+                    fprintf (F, "\n");
+                }
+
                 M = M->Next;
             }
-            fprintf (F, "\n");
-        } else {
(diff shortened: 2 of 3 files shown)


GorillaSapiens:
example output
<pre>
#define _poserror __poserror
#define EOF -1
#define _HAVE_size_t
#define __STDC_NO_VLA__ 1
#define FOOBIZ(s,t) 12
#define __FILE__
#define L_tmpnam FILENAME_MAX
#define __DATE__ "Jun 10 2025"
#define _IONBF 2
#define SEEK_CUR 0
#define __CC65_STD_C99__ 1
#define SEEK_END 1
#define __CC65__ 531
#define __CC65_STD__ 2
#define _STDIO_H
#define __TIME__ "22:42:52"
#define FOPEN_MAX 8
#define FOOD(x) x
#define __CC65_STD_CC65__ 2
#define __COUNTER__
#define NULL ((void *) 0)
#define FILENAME_MAX (16+1)
#define _IOLBF 1
#define putc(c,f) fputc (c, f)
#define _IOFBF 0
#define __CC65_STD_C89__ 0
#define getc(f) fgetc (f)
#define BUFSIZ 256
#define __STDC_NO_THREADS__ 1
#define FOO(x,y,z) (x+y+z)
#define TMP_MAX 256
#define __STDC_NO_COMPLEX__ 1
#define SEEK_SET 2
#define __STDC_NO_ATOMICS__ 1
#define __LINE__
#define FOOBAR() 10
#define __STDC_HOSTED__ 1
</pre>

kugelfuhr:
The macro listing is currently coupled to the `-d` switch which is labeled "debug mode". `-d` enables lots of other things like outputting internal information to the assembler file. So my suggestion would be to leave it as it is.

Instead, new options `-dM`, `-dD`, and maybe `-dN` might be added, similarly as they are available in gcc. This would leave `-d` available for its original purpose and on the other side, use commonly used options for the same purpose as they are known from gcc and clang.

GorillaSapiens:
> The macro listing is currently coupled to the `-d` switch which is labeled "debug mode". `-d` enables lots of other things like outputting internal information to the assembler file. So my suggestion would be to leave it as it is.
> 
> Instead, new options `-dM`, `-dD`, and maybe `-dN` might be added, similarly as they are available in gcc. This would leave `-d` available for its original purpose and on the other side, use commonly used options for the same purpose as they are known from gcc and clang.

No.  #1640 does not ask for that level of functionality.  this patch meets the request in #1640, "Add possibility to print the pre-defined macros in compile time".  if someone wants to expand "-d" to more closely match gcc, they are welcome to do so.

kugelfuhr:
The problem with this approach is that `-d` was never thought to be a "print preprocessor macros" option. The macro list was just part of the debug output enabled with `-d`.

@mrdudz Please let me know if you plan to merge this change or prefer a more solid solution adding a `-dM` command line option. If you prefer the later, I will prepare a PR.

GorillaSapiens:
> The problem with this approach is that `-d` was never thought to be a "print preprocessor macros" option. The macro list was just part of the debug output enabled with `-d`.
> 
> @mrdudz Please let me know if you plan to merge this change or prefer a more solid solution adding a `-dM` command line option. If you prefer the later, I will prepare a PR.

and it still is part of the debug output enabled with '-d'.  i merely changed the format. ;)

"-d" currently sets a global "Debug" variable.  i looked at adding "-dM" briefly, and decided that adding this would quickly snowball into searching for ALL uses of "Debug" and deciding what letter to give each one, which i was certain would be a long protracted discussion that i really didn't care to get into, and would touch things that have nothing to do with #1640.

this patch was meant only to meet the ask in #1640.

mrdudz:
I don't have an opinion about this - since i don't use this feature. I usually favour solutions that are more like gcc/clang for that matter.

Someone who actually uses this should comment

kugelfuhr:
Another problem with this PR is that it doesn't work with variadic macros.

GorillaSapiens:
That's a legitimate point, and I'd be happy to work on that if the PR
weren't already closed and I didn't think y'all's just piss over it all
anyway.

mrdudz:
It's not closed.

In any case, by now i think i'd also prefer a proper -dM option ... -d is too much of a moving target IMHO

GorillaSapiens:
If this were approved, it would be trivial to move the command line flag
that triggers it. I'm not opposed to that. All I'm saying is that it's out
of scope to do a complete overhaul of all the -d* handling. This PR was
intended to be just about printing the macro information

mrdudz:
replaced by #2698

PR #2693, opened 2025-06-12
Apple2: Setup IRQ/RST vectors in LC if needed

### LINKED ISSUE #2691: Missing 6502 vectors W/O ProDOS
The address space used by the Apple II Language Card (LC) includes the 6502 vectors.

On the Apple II, C programs run with the LC activated. This does by the way not preclude to run on a 48kB machine without LC as long as the LC isn't actively used by the program.

ProDOS temporarily activates the LC too. Therefore it makes sure to have valid 6502 vectors set up. But C programs are supposed to generally also work without ProDOS.

* The NMI vector isn't relevant at all on the Apple II.
* The Reset vector in the LC is only relevant on the ][. The //e (and newer) implicitly disables the LC on Reset.
* The IRQ vector is only relevant for BRK as the cc65 interruptor support is based on ProDOS.

The proposal is therefore:

* Add code to the crt0.s "ONCE" segment (https://github.com/cc65/cc65/blob/master/libsrc/apple2/crt0.s#L42) that checks if ProDOS is loaded (the usual $BF00 == $4C check) and if it is not loaded, sets the Reset and IRQ vector in the LC.

* Add two functions to have the vectors point to each consisting of
```
sta $C082
jmp ($FFF?)
```

This approach adds 12 bytes of RAM usage for super-specific scenarios. But I don't see an alternative if we stick with the approach of a single binary both for ProDOS and non-ProDOS.

colinleroy:
Programs running under DOS3.3 need to setup correct reset and IRQ vectors in the language card.
Fixes #2691

### DIFF
--- a/libsrc/apple2/crt0.s
+++ b/libsrc/apple2/crt0.s
@@ -58,6 +58,7 @@ init:   ldx     #zpspace-1
         ; Check for ProDOS.
         ldy     $BF00           ; MLI call entry point
         cpy     #$4C            ; Is MLI present? (JMP opcode)
+        php                     ; Remember whether we're running ProDOS
         bne     basic
 
         ; Check the ProDOS system bit map.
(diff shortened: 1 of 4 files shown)


oliverschmidt:
Thanks for the heads-up. I'll review asap.

libsrc/apple2/crt0.s
```diff
@@ -76,6 +76,27 @@ init:   ldx     #zpspace-1
         ldx     #>$BF00
         bne     :+              ; Branch always
 
+set_lc_vectors:
```
oliverschmidt, line 79: I know that this is really silly, but I'd like to at least mention that so far the style was to have code labels always on the line of the first actual instruction - and find names that are short enough to do so.

libsrc/apple2/crt0.s
```diff
@@ -58,7 +58,7 @@ init:   ldx     #zpspace-1
         ; Check for ProDOS.
         ldy     $BF00           ; MLI call entry point
         cpy     #$4C            ; Is MLI present? (JMP opcode)
```
oliverschmidt, line 60: I'm not perfectly sure but could it be possible to save the result of this check via PHP and restore it later via PLP so that we can get it cheaply further down where we have the LC write-enabled anyway?

oliverschmidt:
Thanks 😀

PR #2695, opened 2025-06-12
Add macros to check for CPU type and supported instruction set.

### LINKED ISSUE #2363: Wishlist: a __CPU__ macro

kugelfuhr:
Supersedes #2687, fixes #2363. I haven't made the value of the `__CPU_6502__` etc. macros the same as that of the `__CPU__` macro (as @mrdudz requested) since I couldn't see how it could be useful. But this is easily changed if you show me a use case.

The patch
* ... adds feature test macros `__CPU_xxx__`
* ... adds a `__CPU__`  macro that contains a bit set similar to `.CPU` in the ca65 assembler.
* ... adds several `__CPU_ISET_xxx__` macros that define numeric constants to be tested against `__CPU__`.
* ... and adds all this to the docs.

### DIFF
--- a/doc/cc65.sgml
+++ b/doc/cc65.sgml
@@ -1154,6 +1154,96 @@ The compiler defines several macros at startup:
   <item><tt/__CC65_STD_CC65__/
   </itemize>
 
+  <label id="macro-CPU">
+  <tag><tt>__CPU__</tt></tag>
+
+  This macro contains a bitset that allows to check if a specific instruction
+  set is supported. For example, the 65C02 CPU supports all instructions of the
+  65SC02. So testing for the instruction set of the 65SC02 using the following
+  check will succeed for both CPUs (and also for the 65816 and HUC6280).
+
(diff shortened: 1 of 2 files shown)


mrdudz:
Thanks!

colinleroy:
Nice, thanks!

PR #2696, opened 2025-06-12
Use new CPU macro to optimize ntohs on 65c02

colinleroy:


mrdudz:
good catch :)

colinleroy:
> good catch :)

That use case was the reason I thought about C accessible cpu macros, after all 🙂

PR #2698, opened 2025-06-12
Implement -dD, -dM and -dN command line switches to output macro definitions

### LINKED ISSUE #1640: Add possibility to print the pre-defined macros in compile time
Hi.

Is it possible to dump the pre-defined macros, e.g. similar to as it does in GCC?

e.g. if I try something like this: `cc65 -E -dM main.c` then I got the following output:

`
Macro Hash Table Summary
  0 : empty
  1 : empty
  2 : empty
  3 : empty
  4 : empty
  5 : empty
  6 : empty
  7 : empty
  8 : empty
  9 : empty
 10 : empty
 11 : empty
 12 : empty
 13 : empty
 14 : empty
 15 : empty
 16 : empty
 17 : empty
 18 : empty
 19 : empty
 20 : empty
 21 : empty
 22 : empty
 23 : empty
 24 : empty
 25 : empty
 26 : empty
 27 : empty
 28 : empty
 29 : empty
 30 : empty
 31 : empty
 32 : empty
 33 : empty
 34 : empty
 35 : empty
 36 : empty
 37 : empty
 38 : empty
 39 : empty
 40 : empty
 41 : empty
 42 : empty
 43 : empty
 44 : empty
 45 : empty
 46 : empty
 47 : empty
 48 : VAR_NAME_VALUE
 49 : empty
 50 : empty
 51 : empty
 52 : empty
 53 : empty
 54 : empty
 55 : empty
 56 : empty
 57 : empty
 58 : empty
 59 : VALUE
 60 : empty
 61 : empty
 62 : empty
 63 : empty
 64 : empty
 65 : empty
 66 : __DATE__
 67 : empty
 68 : empty
 69 : empty
 70 : empty
 71 : __CC65_STD_C99__
 72 : empty
 73 : empty
 74 : empty
 75 : empty
 76 : __CC65__
 77 : empty
 78 : empty
 79 : empty
 80 : empty
 81 : empty
 82 : empty
 83 : empty
 84 : empty
 85 : empty
 86 : empty
 87 : empty
 88 : empty
 89 : empty
 90 : empty
 91 : empty
 92 : empty
 93 : empty
 94 : empty
 95 : __CC65_STD__
 96 : empty
 97 : empty
 98 : empty
 99 : empty
100 : empty
101 : empty
102 : __TIME__
103 : empty
104 : empty
105 : empty
106 : empty
107 : empty
108 : empty
109 : empty
110 : empty
111 : empty
112 : empty
113 : empty
114 : __CC65_STD_CC65__
115 : empty
116 : empty
117 : empty
118 : empty
119 : empty
120 : empty
121 : empty
122 : empty
123 : empty
124 : empty
125 : empty
126 : empty
127 : empty
128 : empty
129 : empty
130 : __CC65_STD_C89__
131 : empty
132 : empty
133 : empty
134 : empty
135 : empty
136 : empty
137 : empty
138 : empty
139 : empty
140 : empty
141 : empty
142 : empty
143 : empty
144 : empty
145 : empty
146 : empty
147 : empty
148 : empty
149 : empty
150 : empty
151 : empty
152 : empty
153 : empty
154 : empty
155 : empty
156 : empty
157 : empty
158 : empty
159 : empty
160 : empty
161 : empty
162 : empty
163 : empty
164 : empty
165 : empty
166 : empty
167 : empty
168 : empty
169 : empty
170 : VALUE_TO_STRING
171 : empty
172 : empty
173 : empty
174 : empty
175 : empty
176 : empty
177 : empty
178 : empty
179 : empty
180 : empty
181 : empty
182 : empty
183 : empty
184 : empty
185 : empty
186 : empty
187 : empty
188 : empty
189 : empty
190 : empty
191 : empty
192 : empty
193 : empty
194 : empty
195 : empty
196 : empty
197 : empty
198 : empty
199 : empty
200 : empty
201 : empty
202 : empty
203 : empty
204 : empty
205 : empty
206 : __STDC_HOSTED__
207 : empty
208 : empty
209 : empty
210 : empty
`

As you can see, there are no any values for that macros.

Also I tried the following way in my `main.c` file:

`
#define XSTR(x) STR(x)
#define STR(x) #x

#pragma message "The value of __CC65__: " XSTR(__CC65__)
`

but I got then the following error:
`
cc65 main.c
main.c:4: Error: '(' expected
`

Is it possible somehow print the macros?

kugelfuhr:
This is a much improved version of #2689 and fixes #1640. It adds three command line options that work similar to the ones in gcc:

* `-dD` outputs all user macro definitions to the output file.
* `-dM` outputs all macro definitions to the output file. Includes predefined macros.
* `-dN` outputs all user macros to the output file. Does not include the replacement.

All three require the `-E` switch and do not touch the existing `-d` command line option.

I do not consider this PR really finished, since the options are mostly gcc compatible, but gcc is itself not consistent between the options, so we might want to change the behavior and discuss the following:

* Currently the output is to the file that contains the preprocessed source. This is consistent with gcc but gcc uses stdout by default while cc65 outputs to a file instead. Would it be better to have cc65 output the macros to stdout?

* gcc is inconsistent in suppressing normal preprocessed output when `-dM` is given, but not for `-dD` and `dN`. I haven't followed this behavior so cc65 will always output the preprocessed source followed by the macros. Should this be changed to be more gcc compatible?

* Would it be better to suppress normal output completely and just output the macro lists when any of the switches are given?

* Does having `-dN` actually make sense? I have added it since gcc has it, but I cannot really see a use case, so it might also be dropped.

I will be away over the weekend starting tomorrow, so if you come to any conclusions until I'm back, I can make the necessary changes (if any) next week.

### DIFF
--- a/doc/cc65.sgml
+++ b/doc/cc65.sgml
@@ -63,6 +63,9 @@ Short options:
   -V                            Print the compiler version number
   -W [-+]warning[,...]          Control warnings ('-' disables, '+' enables)
   -d                            Debug mode
+  -dD                           Output all macro definitions (needs -E)
+  -dM                           Output user defined macros (needs -E)
+  -dN                           Output user defined macro names (needs -E)
   -g                            Add debug info to object file
   -h                            Help (this text)
   -j                            Default characters are signed
@@ -197,6 +200,26 @@ Here is a description of all the command line options:
   <tag><tt>-d, --debug</tt></tag>
 
   Enables debug mode, for debugging the behavior of cc65.
+
+
+  <label id="option-dD">
+  <tag><tt>-dD</tt></tag>
+
+  Like <tt/<ref id="option-dM" name="-dM">/ but does not include the predefined
+  macros.
+
+
+  <label id="option-dM">
+  <tag><tt>-dM</tt></tag>
+
+  When used with -E, will output <tt>#define</tt> directives for all the macros
+  defined during execution of the preprocessor, including predefined macros.
+
+
+  <tag><tt>-dN</tt></tag>
+
+  Like <tt/<ref id="option-dD" name="-dD">/ but will only output the macro names,
+  not their definitions.
 
 
   <tag><tt>--debug-tables name</tt></tag>
--- a/src/cc65/compile.c
+++ b/src/cc65/compile.c
@@ -496,6 +496,17 @@ void Compile (const char* FileName)
         while (PreprocessNextLine ())
         { /* Nothing */ }
 
+        /* Output macros if requested by the user */
+        if (DumpAllMacrosFull) {
+            OutputAllMacrosFull ();
+        }
+        if (DumpUserMacros) {
+            OutputUserMacros ();
+        }
+        if (DumpUserMacrosFull) {
+            OutputUserMacrosFull ();
+        }
+
         /* Close the output file */
         CloseOutputFile ();
 
--- a/src/cc65/global.c
+++ b/src/cc65/global.c
@@ -44,12 +44,15 @@
 
 
 unsigned char AddSource         = 0;    /* Add source lines as comments */
+unsigned char AllowNewComments  = 0;    /* Allow new style comments in C89 mode */
 unsigned char AutoCDecl         = 0;    /* Make functions default to __cdecl__ */
 unsigned char DebugInfo         = 0;    /* Add debug info to the obj */
+unsigned char DumpAllMacrosFull = 0;    /* Output all macro defs */
+unsigned char DumpUserMacros    = 0;    /* Output user macro names */
+unsigned char DumpUserMacrosFull= 0;    /* Output user macro defs */
 unsigned char PreprocessOnly    = 0;    /* Just preprocess the input */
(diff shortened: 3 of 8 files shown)


mrdudz:
I haven't tried yet, but this makes sense to me :) perhaps this weekend...

kugelfuhr:
After sleeping over it, I would suggest to change the behavior slightly:

* `-dN` should be removed - I cannot see where it could be useful.
* `-dM` and `-dD` should be replaced by `-dM` and `-dP` where the former outputs user macros only and the latter prints predefined macros only.
* Allow specifying more than one suboption letter, so `-dMP` or `-dPM` will print all macros.

This would differ from gcc but make the behavior more orthogonal and allows to also print compiler defined macros only, which is not possible with the options as implemented now (apart from giving an empty input file).

kugelfuhr:
Since no one had any comments, I have changed the implementation as described in my last post. As far as I'm concerned this PR is now done.

mrdudz:
Sorry, i was still busy with brushing up that m740 patch...

mrdudz:
looks good to me, thanks!

PR #2699, opened 2025-06-13
Fix temporary filenames again.

colinleroy:
Outputting temp files in the output directory means we have to distinguish source files in different source directories that happen to have the same name:

```
cl65 ... api/compose.c ... cli/compose.c ...

cl65: Duplicate file in argument list: 'compose.c.256412.o'
cl65: Duplicate file in argument list: 'compose.c.256412.o'
```
We can't have input files with the same name in different directories anymore. How should we fix it?
One suggestion: this PR. Keep the path, replacing the separators with _, so that temp files are still created in the output directory, but in the above case, with names like api_compose.c.256412.o and cli_compose.c.256412.o

### DIFF
--- a/src/common/fname.c
+++ b/src/common/fname.c
@@ -156,15 +156,22 @@ char* MakeTmpFilename (const char *Directory, const char *Origin, const char* Ex
 {
     char* Out;
     size_t Len = 0;
+    static unsigned int Counter = 0;
 
-    /* Allocate template */
+    /* Allocate enough for the directory, ... */
     if (Directory != NULL) {
       Len = strlen (Directory);
     }
-    Len += strlen (Origin) + strlen (".2147483648") + strlen (Ext) + 1;
+
+    /* ... plus the the original name, the maximum length of the PID, the
+     * maximum length of the counter, the extension, and the terminator.
+     */
+    Len += strlen (Origin) + (strlen (".2147483648") * 2) + strlen (Ext) + 1;
     Out = xmalloc (Len);
 
-    snprintf (Out, Len, "%s%s.%u%s", (Directory != NULL ? Directory : ""),
-                                     FindName(Origin), getpid(), Ext);
+    snprintf (Out, Len, "%s%s.%u%u%s", (Directory != NULL ? Directory : ""),
+                                     FindName(Origin), getpid(), Counter, Ext);
+    Counter++;
+
     return Out;
 }


colinleroy:
(ping @mrdudz)

colinleroy:
Another possibility suggested by @rofl0r would be to add a running counter to the filename (like orig_file.c.PID.COUNT.s)

mrdudz:
Somehow a counter seems more desireable... else you may end up with VERY long names, which might trigger new problems

colinleroy:
> Somehow a counter seems more desireable... else you may end up with VERY long names, which might trigger new problems

I'll do that. You're OK with a single static incremented at each call regardless of the uniqueness of the source filename? It seems like a huge bother otherwise.

mrdudz:
> You're OK with a single static incremented at each call regardless of the uniqueness of the source filename? It seems like a huge bother otherwise.

Fine with me. pid+counter should give a very much unique identifier - the source filename is only decoration at this point :)

colinleroy:
> Fine with me. pid+counter should give a very much unique identifier - 

Done!

> the source filename is only decoration at this point :)

Decoration that we want so we can parse our .map files, though!

mrdudz:
wouldnt strlen (".2147483648") have to be added twice now (once for pid, once for counter)?

colinleroy:
> wouldnt strlen (".2147483648") have to be added twice now (once for pid, once for counter)?

Good catch! Fixed.

mrdudz:
cheers, looks good. (hopefully all quirks found now? :))

colinleroy:
> cheers, looks good. (hopefully all quirks found now? :))

Hope so! At worst we go back to when cc65 trashed your hand-crafted test.s file when you built test.c :sweat_smile: (jk. At worst I'll come back to it because I want it to work but I also want to read my .maps :-)

src/common/fname.c
```diff
@@ -156,15 +156,22 @@ char* MakeTmpFilename (const char *Directory, const char *Origin, const char* Ex
...
+
+    /* ... plus the the original name, the maximum length of the PID, the
+     * maximum length of the counter, the extension, and the terminator.
+     */
+    Len += strlen (Origin) + (strlen (".2147483648") * 2) + strlen (Ext) + 1;
     Out = xmalloc (Len);
 
-    snprintf (Out, Len, "%s%s.%u%s", (Directory != NULL ? Directory : ""),
-                                     FindName(Origin), getpid(), Ext);
+    snprintf (Out, Len, "%s%s.%u%u%s", (Directory != NULL ? Directory : ""),
```
rofl0r, line 172: having no separator between pid and counter creates a situation where somepid+counter could be a valid otherpid

PR #2701, opened 2025-06-14
Make sure pid and counter don't interfere

colinleroy:


mrdudz:
another good catch!

(how many quirks can you have in 3 lines of code? :))

PR #2704, opened 2025-06-14
initial commit, 2400 baud userport driver

mrdudz:
Another thing that was laying around here for years. Completely untested right now. Needs cleanup too. Not ready for merge

### DIFF
--- a/libsrc/c64/ser/c64-up2400.s
+++ b/libsrc/c64/ser/c64-up2400.s
@@ -1,0 +1,498 @@
+; This is an implementation of a reliable 2400b serial driver for the
+; userport of a C64.
+;
+; Johan Van den Brande, (c) 2014
+; 
+; Based on George Hug's 'Towards 2400' article in the Transactor Magazine volume 9 issue 3.
+; (https://archive.org/details/transactor-magazines-v9-i03) 
+;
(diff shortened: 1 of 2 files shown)


groessler:
For this to use one would need a userport -- RS232 driver to adapt the voltage, correct?

mrdudz:
Yes, that, and the control lines lines have to be inverted, see eg:

https://www.zimmers.net/anonftp/pub/cbm/documents/projects/rs232/rs232-userport.txt

(Today you'd use a MAX2323 and an 7404, i guess, but the principle remains)

The inverted control lines are a bit annoying, since that means you cant just connect a random cheap rs232->ttl dongle thingy

libsrc/c64/ser/c64-up2400.s
```diff
@@ -0,0 +1,498 @@
...
+        ldx #>SER_ERR_NO_DATA
+        rts
+
+;----------------------------------------------------------------------------
+; OURBASIN: This is a minimised call to get the character from the buffer.
+; The Kernal code does not allow zero bytes (0x00)... this does.
+;
+
+SER_OURBASIN:
+        jsr $F14E
```
groessler: Calling directly into ROM?
I think at least a comment is missing what these function are supposed to do (also below $F1B4).
Will this work on e.g. VIC20?

PR #2707, opened 2025-06-15
m740 Support

mrdudz:
took #986 and made all opcodes work

Basic roundtrip (disassemble/assemble) works now. Some docs are missing. Some more cleanup is needed perhaps :)


mrdudz:
So, it passes the regression test (that i added/updated too).

One odd thing is that "inc a" and "dec a" are right now swapped (in both the assembler and the disassembler) - ie the documentation (that pdf) has them the other way around than currently implemented. This seems to follow the general logic of 6502 clones - however, we need to confirm whether this is really correct for the m740. @roberto314 do you remember anything about that?

mrdudz:
```
da65 -i IMO100_2.in
ca65 -t none --cpu m740 -o rebuild.o IMO100_2.asm
ld65 -C imo100.cfg -o rebuild.bin rebuild.o
```

works too, and gives an identical file... now about the inc a/dec a....

mrdudz:
So, for me this is mostly fine... let the dogs out, please comment - perhaps something can be done more elegant, please tell :)

src/da65/handler.c
```diff
@@ -719,33 +736,38 @@ void OH_DirectImmediate (const OpcDesc* D)
 
 
 
+/* <bit> zp */
+/* NOTE: currently <bit> is part of the instruction */
 void OH_ZeroPageBit (const OpcDesc* D)
 {
-    unsigned Bit = GetCodeByte (PC) >> 5;
+    /* unsigned Bit = GetCodeByte (PC) >> 5; */
```
oliverschmidt, line 743: I personally - actually strongly - dislike commented-out code. Especially without a comment why it was commented-out. If there's nothing to say in such a comment, then the code should just be removed and it should be relied upon that the Git history provides better context/knowledge.

src/da65/handler.c
```diff
@@ -719,33 +736,38 @@ void OH_DirectImmediate (const OpcDesc* D)
...
 }
 
 
 
+/* <bit> A */
+/* NOTE: currently <bit> is part of the instruction */
 void OH_AccumulatorBit (const OpcDesc* D)
 {
-    unsigned Bit = GetCodeByte (PC) >> 5;
+    /* unsigned Bit = GetCodeByte (PC) >> 5; */
```
oliverschmidt, line 759: I personally - actually strongly - dislike commented-out code. Especially without a comment why it was commented-out. If there's nothing to say in such a comment, then the code should just be removed and it should be relied upon that the Git history provides better context/knowledge.

src/da65/handler.c
```diff
@@ -719,33 +736,38 @@ void OH_DirectImmediate (const OpcDesc* D)
...
 }
 
 
-
+/* <bit> A, rel */
+/* NOTE: currently <bit> is part of the instruction */
 void OH_AccumulatorBitBranch (const OpcDesc* D)
 {
-    unsigned Bit = GetCodeByte (PC) >> 5;
+    /* unsigned Bit = GetCodeByte (PC) >> 5; */
```
oliverschmidt, line 770: I personally - actually strongly - dislike commented-out code. Especially without a comment why it was commented-out. If there's nothing to say in such a comment, then the code should just be removed and it should be relied upon that the Git history provides better context/knowledge.

oliverschmidt (changes requested):

src/da65/handler.c
mrdudz: I agree (it is only still there for that case that someone comes up with the idea that "bit" should be handled differently). Will remove before merge

sidneycadot:
I am a bit surprised that I was asked to review -- my expertise is the simulator part of the codebase, and I know very little if anything about the other tools.

Please proceed without my input :)

cfg/m740.cfg
```diff
@@ -0,0 +1,37 @@
+FEATURES {
```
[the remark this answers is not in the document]
mrdudz: done

mrdudz:
removed some more unrelated stuff from the patch. Already spotted a few more little things - more for tomorrow :)

mrdudz:
@kugelfuhr perhaps you can skik over it as well? :) (Why cant i add you as reviewer? bleh)

asminc/cpu.mac
```diff
@@ -22,3 +22,4 @@ CPU_65816          = CPU_ISET_6502|CPU_ISET_65SC02|CPU_ISET_65816
 CPU_SWEET16        = CPU_ISET_SWEET16
 CPU_HUC6280        = CPU_ISET_6502|CPU_ISET_65SC02|CPU_ISET_65C02|CPU_ISET_HUC6280
 CPU_4510           = CPU_ISET_6502|CPU_ISET_65SC02|CPU_ISET_65C02|CPU_ISET_4510
+CPU_M740           = CPU_ISET_6502|CPU_ISET_M740
```
kugelfuhr, line 25: Is it correct that the `CPU_ISET_65SC02` and `CPU_ISET_65C02` bits aren't set?

doc/ca65.sgml
```diff
@@ -152,7 +152,7 @@ Here is a description of all the command line options:
   Set the default for the CPU type. The option takes a parameter, which
   may be one of
 
-        6502, 6502X, 6502DTV, 65SC02, 65C02, 65816, sweet16, HuC6280, 4510
+        6502, 6502X, 6502DTV, 65SC02, 65C02, 65816, sweet16, HuC6280, 4510, m740
```
kugelfuhr, line 155: Are you sure this should be lower case? Most other names are upper case.

kugelfuhr (commented):
I haven't found much to complain about, but I have no knowledge of this cpu, so this was to be expected.

doc/ca65.sgml
mrdudz: good catch :)

asminc/cpu.mac
mrdudz: Ah right, forgot... i'm still roaming around the docs and tests to find out which instructions those flags refer to exactly. The M740 is a bit strange for that matter :)

mrdudz: yes i know, i am making a test now that checks all instructions per bit in that flag, and then update that line accordingly :)

kugelfuhr: Having a closer look shows that there's no support for the complete instruction set, so the code is correct as is.

mrdudz:
so there is the test. it got a bit out of hand :) A few more quirks show up, but they should be fixed in another PR (and the test updated)

mrdudz:
ok thats it for me, will merge soon if noone steps in :)

GorillaSapiens:
> One odd thing is that "inc a" and "dec a" are right now swapped (in both the assembler and the disassembler) - ie the documentation (that pdf) has them the other way around than currently implemented. This seems to follow the general logic of 6502 clones - however, we need to confirm whether this is really correct for the m740.

not sure which PDF you're referring to.  i found this, if it is helpful:
https://bitsavers.org/components/mitsubishi/melps/1989_Series_740_Software_Users_Manual.pdf

mrdudz:
> > One odd thing is that "inc a" and "dec a" are right now swapped (in both the assembler and the disassembler) - ie the documentation (that pdf) has them the other way around than currently implemented. This seems to follow the general logic of 6502 clones - however, we need to confirm whether this is really correct for the m740.
> 
> not sure which PDF you're referring to. i found this, if it is helpful: https://bitsavers.org/components/mitsubishi/melps/1989_Series_740_Software_Users_Manual.pdf

Thats the one, indeed. (@roberto314 linked to it in #986 ) It seems weird the INC A/DEC A doesn't follow the usual pattern.

@roberto314 did you find any other software we could disassemble?

roberto314:
Sorry, no luck. The code from the programmable logic unit is inside the
m50747, not in a ROM.
Also the chip is in some chess computers, but there is also no external ROM.


mit freundlichen Grüßen.
Ing. Robert Offner MSc.


Bob Andrews ***@***.***> schrieb am Fr., 20. Juni 2025, 17:13:

mrdudz:
> Sorry, no luck. The code from the programmable logic unit is inside the m50747, not in a ROM. Also the chip is in some chess computers, but there is also no external ROM. 

OK, thanks anyway.... too bad :)

So i'll merge it as is - we'll see

roberto314:
Which opcode or opcodes are we talking about?


mit freundlichen Grüßen.
Ing. Robert Offner MSc.


Bob Andrews ***@***.***> schrieb am Fr., 20. Juni 2025, 19:39:

mrdudz:
> Which opcode or opcodes are we talking about?

in the pdf $1a is "dec a", and $3a is "inc a" - however, that does not match the usual encoding scheme, which is why the assembler (and disassembler) currently uses $3a for "dec a" and $1a for "inc a". unfortunately the software you dumped does not contain either, so no way to even guess what is correct.

can you actually run software on that thing for a test? should be easy to find out then :)

roberto314:
I' ll think about it and get back to you.

mit freundlichen Grüßen.
Ing. Robert Offner MSc.


Bob Andrews ***@***.***> schrieb am Fr., 20. Juni 2025, 19:59:

roberto314:
Did you see this?
image

mrdudz:
yes, the pdf in itself is consistent

PR #2709, opened 2025-06-16
added reporting of fatal/error/warning/note location generation with …

GorillaSapiens:
…"-d"

debugging some other work, i found there are many Fatal() / Error() / Warning() / Note() calls with identical or substantially similar verbiage.  tracing down where exactly the error messages were coming from was a bit tedious, especially since printf style formatting frustrates simple grep in some cases.  i thought "wouldn't it be nice if i could turn on a debug flag, and get the location in the compiler where the message was generated."

here it is.

without "-d", business as usual, nothing changed
<pre>
gorilla@guest:~/cc65 (gps4bugs)$ bin/cc65 foo.c
foo.c:18: Fatal: Floating point type is currently unsupported
</pre>

with "-d", also prints the file and line in the compiler where the message was generated.
<pre>
gorilla@guest:~/cc65 (gps4bugs)$ bin/cc65 -d foo.c
[cc65/codegen.c:90] foo.c:18: Fatal: Floating point type is currently unsupported
</pre>

### DIFF
--- a/src/cc65/error.c
+++ b/src/cc65/error.c
@@ -39,6 +39,7 @@
 
 /* common */
 #include "coll.h"
+#include "debugflag.h"
 #include "print.h"
 #include "strbuf.h"
 
@@ -183,10 +184,14 @@ static unsigned GetDiagnosticLineNum (void)
 
 
 
-void Fatal (const char* Format, ...)
+void _Fatal (const char *file, int line, const char* Format, ...)
 /* Print a message about a fatal error and die */
 {
     va_list ap;
+
+    if (Debug) {
+        fprintf(stderr, "[%s:%d] ", file, line);
+    }
 
     fprintf (stderr, "%s:%u: Fatal: ", GetDiagnosticFileName (), GetDiagnosticLineNum ());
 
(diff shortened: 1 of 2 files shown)


kugelfuhr:
I didn't have problems locating simple errors, but I can see that this change might be helpful. Adding the location only when `-d` is given is a good idea.

A problem with the PR is the use of reserved identifiers (see ISO/IEC 9899:1999 (E), 7.2.1 "Reserved identifiers"). I would suggest to change the function names.

GorillaSapiens:
> I didn't have problems locating simple errors, but I can see that this change might be helpful. Adding the location only when `-d` is given is a good idea.
> 
> A problem with the PR is the use of reserved identifiers (see ISO/IEC 9899:1999 (E), 7.2.1 "Reserved identifiers"). I would suggest to change the function names.

suggest an alternate prefix or other naming scheme that you would approve of.

kugelfuhr:
I do not know what the developers prefer. You could for example place the underline at the end of the identifiers: `Error_`, `Fatal_` etc. would be unproblematic. But some people don't like that, so others might speak up here.

mrdudz:
appending the underscore is usually what i do myself (after a learned prepending is not always quite correct)

GorillaSapiens:
> appending the underscore is usually what i do myself (after a learned prepending is not always quite correct)

 (E), 7.2.1 aside, i don't mind prepending on internal symbols.  it almost never causes problems, and when it does it is simple to change.

appending is fine, i don't care.  i just don't want to do it and then have someone whine about it being wrong.  that's busy work.

anecdote: working on a commercial project, marketing couldn't come up with a name.  so i just put "GENERIC PRODUCT NAME" anywhere i needed a name.  huge meeting, everyone told me to change it, but nobody could agree on what it should be.  told them it was in a lot of places, and i didn't want to do this twice, because it's more than just "change a name", you've got to tweak UI label placements, etc, etc...  they said "we don't care, just change it".  i said alright...  the next day when they got the product with the name "LIVE NUDE GIRLS", you better believe they cared.  they got me a new correct name immediately...

so i ask in all seriousness, are we ok with appending an underscore?  or do you want something else?

mrdudz:
seems ok, thanks!

PR #2710, opened 2025-06-16
Fix wrong file offset handling for a memory area with only run segments written to a file

### LINKED ISSUE #2694: unexpected ld65 error when load=ROM and run=RAM
### Discussed in https://github.com/cc65/cc65/discussions/2692

<div type='discussions-op-text'>

<sup>Originally posted by **rcook** June 11, 2025</sup>
Here's my program `simple.s`:

```asm
.code
    lda #<str
    sta $80
    lda #>str
    sta $81
    rts

.data
str:
    .byte "hello world"
```

And my config `simple.cfg`:

```
MEMORY {
    RAM: start = $0000, size = $8000, fill = yes;
    ROM: start = $8000, size = $8000, fill = yes;
}

SEGMENTS {
    CODE: type = ro, load = ROM;
    DATA: type = ro, load = ROM, run = RAM;
}
```

So, I'm trying to use `load` and `run` to assemble the data segment into ROM but with run addresses in RAM (with the intention of eventually using `copydata` to copy the data from ROM to RAM at startup). I build this as follows:

```bash
ca65 -o simple.o simple.s
ld65 -C simple.cfg -o simple.bin simple.o
```

Unfortunately, this fails with the following error message:

```
ld65: Internal Error: Invalid file offset for memory area ROM: 32768/32757
```

What does this error mean and how do I fix it? Perhaps I am misunderstanding how `load` and `run` work. I believe that this is a consequence of the `fill = yes`.
</div>

kugelfuhr:
This should fix #2694. I've also reformatted two long comments and slightly refactored a short piece of code that was hard to read.

Not sure about a test. Do you want me to add the problematic input from #2694? If so, where should it go? Is it ok to check in binary files for good/bad comparison?

### DIFF
--- a/src/ld65/config.c
+++ b/src/ld65/config.c
@@ -2107,7 +2107,9 @@ unsigned CfgProcess (void)
                             FillLevel - M->Size, (FillLevel - M->Size == 1) ? "" : "s");
             }
             if (FillLevel > M->FillLevel) {
-                /* Regular segments increase FillLevel. Overwrite segments may increase but not decrease FillLevel. */
+                /* Regular segments increase FillLevel. Overwrite segments may
+                ** increase but not decrease FillLevel.
+                */
                 FillAdded = FillLevel - M->FillLevel;
                 M->FillLevel = FillLevel;
             }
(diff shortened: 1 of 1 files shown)


kugelfuhr:
A second look at the change might also be a good idea ...

mrdudz:
mmmh i cant judge if that change makes sense

mrdudz:
It would really be nice to add a test for this (and prepare something for future ld problems) - but no idea how to start. The heat doesn't help either :=)

kugelfuhr:
The whole test infrastructure is a bit inflexible. And, there are no linker tests so far. The "asm/misc" directory is probably the most appropriate place. Or you can add a new "ld" test directory.

mrdudz:
Thanks!

PR #2711, opened 2025-06-17
TGI Documentation and Comments fix

mrdudz:
promised follow up to #967

### DIFF
--- a/doc/tgi.sgml
+++ b/doc/tgi.sgml
@@ -43,7 +43,7 @@ of range.
 <ref id="tgi_setcolor" name="tgi_setcolor">
 <tag/Example/<tscreen><verb>
 /* Draw the upper half of an ellipse */
-tgi_setcolor(TGI_COLOR_BLUE);
+tgi_setcolor(1);
 tgi_arc (50, 50, 40, 20, 0, 180);
 </verb></tscreen>
 </descrip>
@@ -67,7 +67,7 @@ be used in presence of a prototype.
 <tag/Availability/cc65
 <tag/See also/Other tgi function
 <tag/Example/<tscreen><verb>
-tgi_setcolor(TGI_COLOR_GREEN);
+tgi_setcolor(1);
 tgi_bar(10, 10, 100, 60);
 </verb></tscreen>
 </descrip>
@@ -94,7 +94,7 @@ be used in presence of a prototype.
 <ref id="tgi_pieslice" name="tgi_pieslice">,
 <ref id="tgi_setcolor" name="tgi_setcolor">
 <tag/Example/<tscreen><verb>
-tgi_setcolor(TGI_COLOR_BLACK);
+tgi_setcolor(1);
 tgi_circle(50, 40, 40);
 </verb></tscreen>
 </descrip>
@@ -154,7 +154,7 @@ be used in presence of a prototype.
 <ref id="tgi_pieslice" name="tgi_pieslice">,
 <ref id="tgi_setcolor" name="tgi_setcolor">
 <tag/Example/<tscreen><verb>
-tgi_setcolor(TGI_COLOR_RED);
+tgi_setcolor(1);
 tgi_ellipse (50, 40, 40, 20);
 </verb></tscreen>
 </descrip>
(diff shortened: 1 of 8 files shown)


mrdudz:
So this hopefully clears up some misunderstandings in the future

@oliverschmidt please check and tell me if i am on the right track there, or if i messed up again :/

mrdudz:
Oh and is it really true (by design) that all platforms should have TGI_COLOR_BLACK and TGI_COLOR_WHITE? (i removed that line for now, because it seems it doesn't have to be true). If the answer is yes though, does that imply every monochrome platform defines *only* those two? (lets say we have a green-on-black display, would it not define _GREEN and _BLACK instead? Or _GREEN, _BLACK AND _WHITE, although it really has 2 colors only?)

(rereading that now... don't take it too seriously :) - the point is whether the docs should even mention that:
- TGI_COLOR_BLACK and TGI_COLOR_WHITE exists (imho not necessarily, but perhaps by convention that is the case)
- TGI_COLOR_BLACK and TGI_COLOR_WHITE are guaranteed to be 0 and 1 (NO - this is just wrong)
- "colors" (as in palette indices) 0 and 1 are guaranteed to exist (goes without saying, imho)

polluks:
In 2025 I suppose the monochrome colours just define how to do something like "dark mode" :-)

oliverschmidt:
Not sure if I miss the point, but I believe...
TGI_COLOR_BLACK and TGI_COLOR_WHITE come colloquially calling monochrome tech black&white tech. Even renaming them to something like "light" and "dark" doesn't necessarily fit. On a printer i.e. "setting a pixel" turns it from white (light) to black (dark). Therefore software for graphics screen dumps always needs the "invert" option. So to name them technically correct they would have to be named something like "on" and "off". Nobody wants that, so

* TGI_COLOR_BLACK means the pixel is off
* TGI_COLOR_WHITE means the pixel is on

and I believe that even if not perfectly correct I think it's 100% in line with user expectations.

oliverschmidt:
I believe the "rules" are like this:
* A palette has at least two entries
* A default palette has always two entries with values equal to TGI_COLOR_BLACK and TGI_COLOR_WHITE
* Which default palette entries have those two values is target specific

oliverschmidt:
> @oliverschmidt please check and tell me if i am on the right track there, or if i messed up again :/

As I wrote before it's too along ago to be sure, but as far as I can tell, these changes make sense. However, there are other team menbers who are supposed to contribute more, because they can check the actual TGI code on "their" target to match what is discussed / changed here. "My" Apple II TGI code is as boring as it gets as palettes aren't supported at all on the Apple II. There's just the immutable default palette.

mrdudz:
> * A palette has at least two entries

i hope that goes without saying (although black on black is nice, of course) :)

> * A default palette has always two entries with values equal to TGI_COLOR_BLACK and TGI_COLOR_WHITE
> * Which default palette entries have those two values is target specific

Yes, those are the important bits, i guess

mrdudz:
> As I wrote before it's too along ago to be sure, but as far as I can tell, these changes make sense. However, there are other team menbers who are supposed to contribute more, because they can check the actual TGI code on "their" target to match what is discussed / changed here. 

Somehow i only remember you, gregg and myself from that dicsussion though :) Lets ping everyone who has contributed something to TGI drivers in the past decade:

c128: @oliverschmidt @greg-king5 @xlar54 @polluks
vic20: @greg-king5 @ops 
c64: @oliverschmidt @greg-king5 @polluks @rhalkyard
plus4: @rhalkyard @mrdudz
cx16: @greg-king5  @xlar54 @polluks
lynx: @oliverschmidt @greg-king5 @Fabrizio-Caruso @karrika @polluks
nes : @oliverschmidt @polluks
atari: : @oliverschmidt @groessler @pfusik 
apple2 : @oliverschmidt @polluks @colinleroy 
telestrat: @oliverschmidt @mrdudz ( @jefftranter @jedeoric )
geos-cbm: @oliverschmidt
atmos: @oliverschmidt @mrdudz @greg-king5 

Everyone: Please check this PR and speak up if you think something is wrong.

Please also try if samples/tgidemo.c works on "your" target. If it does, that is already a good hint at that the driver is matching the (now) documented behaviour :)

I will keep it dangling for some days and then merge

karrika:
I am happy with having black. The old palette with a transparent black caused a lot of confusion. This approach is ok. Besides... For games you can always override the palette with your own.

So these document changes are ok.

mrdudz:
PS: if you notice "your" target does NOT match this (new) documented behaviour - don't hesitate to fix it :)

mrdudz:
Merging now - no hot topics left, it seems :)

Issue #2527, opened 2024-10-17
Missing optimization


polluks:
Hi!
x ^= x;
should be equivalent to
x = 0;

GorillaSapiens:
does anyone really do that on purpose though?  there are probably a ton of things like
x ^ 0 = x
x | 0 = x
x | x = x
x & 0xff = x
x & x = x
that could be checked and optimized.

mrdudz:
> that could be checked and optimized, but you'd never actually see in real code.

BUT things like this could very well happen in intermediate versions of generated code. They should get optimized out either way :)

Marrin:
> Well, did you ever see "+0"?

Yes. Can't recall the exact code but there was something like `registers+0`, `registers+1`, `registers+2` and also just `registers` to make it clear to the human reader when single register 0 was meant and when a pointer to registers that just happens to be also the pointer to the first register.

mrdudz:
so after #2715, are there still simple cases that justifies keeping this ticket open?

PR #2715, opened 2025-06-18
Added a few simple optimizations

kugelfuhr:
The steps added should resolve most of #2527.

To see the effect, you can translate this file with and without the changes:

```c
int main()
{
    unsigned char x = 3;
    unsigned y = 2;

    x = x;
    y = y;

    x ^= x;
    y ^= y;

    x |= x;
    y |= y;

    x &= x;
    y &= y;

    x ^= 0;
    y ^= 0;

    x &= 0;
    y &= 0;

    x |= 0;
    y |= 0;

    x ^= 0xFF;
    y ^= 0xFFFF;

    x &= 0xFF;
    y &= 0xFFFF;

    x |= 0xFF;
    y |= 0xFFFF;

    return 0xAA55;
}
```


mrdudz:
I tried to come up with a test.... using labels after every second operation, and then the && operator to compute the difference between them.... oh well. that only opened another can of works :)

Thanks, looks good!

PR #2716, opened 2025-06-18
Fix an internal error in da65 when there are certain labels in a skipped section

### LINKED ISSUE #2649: Internal error: GetGraularity called for style = 10 when using "skip"
I try to skip a segment of `0`s in an rom file however I hit the error when calling `da65 -i Config/proms.info`

```
Internal error: GetGraularity called for style = 10
```

Fails:
```
RANGE { START $1800; END $1fff; TYPE Skip; };
```

Works:
```
RANGE { START $1800; END $1fff; TYPE ByteTable; };
```

kugelfuhr:
Fixes #2649.


mrdudz:
nice, thanks!

Issue #938, opened 2019-09-20
ca65: cheap local label scope vs. variable definition


fadden:
A question about "cheap local labels"...

This is not expected to work:

    LAB1:   nop
            beq     @LAB2
    FOO     nop
    @LAB2:  nop

because an instruction with a non-local label sits between the definition of and reference to `@LAB2`.

However, I was surprised to discover that defining a constant also ends the scope:

    LAB1:   nop
            beq     @LAB2
    FOO     =       $1234
    @LAB2:  nop

as does defining a numeric variable:

    LAB1:   nop
            beq     @LAB2
    FOO     .set    $1234
    @LAB2:  nop

The latter case is of interest to me because SourceGen now allows the definition of local variable tables, which allow the user to redefine zero-page address labels and 65816 stack-relative constants to match what a subroutine is doing.  The local variable tables are interfering with the automatic label localization.

I was wondering if this was intentional.

greg-king5:
> I was wondering if this was intentional.

It's intentional.  This:
```asm
LAB1:   nop
        beq     @LAB2
FOO := *
        nop
@LAB2:  nop
```
is the same as:
> ```asm
> LAB1:   nop
>         beq     @LAB2
> FOO:    nop
> @LAB2:  nop
> ```

mrdudz:
the programmcounter assignment should be handled as a special case, imho - its the one and only case where breaking the scope is necessary

greg-king5:
> the programmcounter assignment should be handled as a special case, imho -- it's the one and only case where breaking the scope is necessary.

OK, I will mark it as a bug.  I have no idea how easy or hard it will be to fix.

mlauke:
can confirm the cheap local label scope is "closed" or ends with a variable definition inside a macro or without the macro at all.

```
@lp:
    sys_delay_ms 1
    dec 
    bne @lp
```
with macro sys_delay_ms defined as
```
.macro sys_delay_us micros
    .local @0,@1,@2
    .local cnt
    cnt = (micros * clockspeed / 6)
    ....
```
just the local variable assignment will end the label scope. we end up with assembly error regarding the cheap local label **@lp**
``` 
delay.s(21): Error: Symbol '@lp' is undefined
make: *** [Makefile:92: ../libwrk/steckschwein-clib/delay.o] Error 1
````

bbbradsmith:
This issue is strongly related to: #505

adiee5:
[...]

PR #2718, opened 2025-06-19
Fix the behavior of variable symbols in regard to cheap locals

### LINKED ISSUE #505: ca65: cheap local scopes never truly end
I've noticed it's possible to "revisit" the same cheap local scope multiple times like this:

```
;this starts a new scope for cheap local lables
SomeSymbol .set 4

	jmp @CheapLocal1

@CheapLocal0:

	.byte $8b

CheapLocalScopeBreaker0:

CheapLocalScopeBreaker1:

CheapLocalScopeBreaker2:

CheapLocalScopeBreaker3:

;this continues the same cheap scope as before, regardless of the many global labels in between
SomeSymbol .set 5

@CheapLocal1:

	lda @CheapLocal0
```
This looks like a bug to me, because we may need to assign values to the same symbol using **.set** in various places of the program, and it looks dangerous to have seemingly unrelated scopes being "merged" like this.

In fact, having assignments through ".set", "=" and ":=" reset the scope for cheap local labels at all seems like a bad decision to me, because it severely restricts the kind of symbol manipulation that can be done around cheap local labels.

kugelfuhr:
Previously every assignment to a variable symbol opened the same scope for cheap locals. So when redefining a variable symbol, an old cheap local scope was reopened which was unexpected and confusing. With this change only the first definition of a variable symbol opens a new scope for cheap locals, but redefinitions of the same symbol do not.

Fixes #505.

### DIFF
--- a/test/asm/err/bug505.s
+++ b/test/asm/err/bug505.s
@@ -1,0 +1,28 @@
+; Test for #525 taken from the issue
+; Redefining a variable symbol "reopens" the old name space for cheap locals
+; Behavior should be: First definition of a variable symbol opens a new 
+; scope for cheap locals, redefinitions of the same symbols do not.
+
+;this starts a new scope for cheap local lables
+SomeSymbol .set 4
+
+        jmp @CheapLocal1
+
+@CheapLocal0:
+
+        .byte $8b
+
+CheapLocalScopeBreaker0:
+
+CheapLocalScopeBreaker1:
+
+CheapLocalScopeBreaker2:
+
+CheapLocalScopeBreaker3:
+
+;this continues the same cheap scope as before, regardless of the many global labels in between
+SomeSymbol .set 5
+
+@CheapLocal1:
+
+        lda @CheapLocal0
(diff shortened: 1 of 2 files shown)


test/asm/err/bug505.s
```diff
@@ -0,0 +1,28 @@
+; Test for #525 taken from the issue
```
rofl0r, line 1: shouldnt this read 505 ?

mrdudz:
nice one. how does this affect #938 ?

kugelfuhr:
> nice one. how does this affect #938 ?

The rule - as I understand it - is "every time a new non cheap symbol is introduced, a new scope for cheap locals starts", and this is of course true for symbols brought into existence by `=` and `:=`. Consequently this PR doesn't affect the behavior discussed in #938. Actually, in my eyes the behavior complained about in #938 is not a bug. 

The only behavior that was wrong was reopening an old cheap local space when redefining a variable symbol using `.set`. If you define it the first time, you get a new cheap local scope as expected since you're introducing a new non cheap symbol. Using another `.set` for the same symbol is not defining it again, it is only changing its value, so the current cheap local scope shouldn't change. This wasn't true before and is fixed in this PR.

mrdudz:
> Actually, in my eyes the behavior complained about in https://github.com/cc65/cc65/issues/938 is not a bug.

right, its more of a feature request, or something.

mrdudz:
Anyway, thanks - merging :)

PR #2719, opened 2025-06-19
Remove useless code

colinleroy:
LOSCR is a valid and safe softswitch on any Apple II. Thanks Oliver S!


oliverschmidt:
Thanks 😀

PR #2720, opened 2025-06-19
fixes #2714

### LINKED ISSUE #2714: unexpected leftover temp files on failed compiles
i suspect this is related to recent changed wrt tmp file creation by cl65

i'm seeing unexpected leftover temp files.  i suspect they are from failed compiles / linking

<pre>
0.s            compiler  nlib                       nrt0.s.53237.0.o
0.s.53212.0.o  foo.bin   nlib_zeropage.s.53237.1.o  samples
assembler      map.txt   notes.txt
</pre>

GorillaSapiens:
install RemoveTempFiles as atexit() handler in case Link() fails and exits early.

### DIFF
--- a/src/cl65/main.c
+++ b/src/cl65/main.c
@@ -1719,7 +1719,7 @@ int main (int argc, char* argv [])
         ** Link() may not return if there's an error, so we install
         ** RemoveTempFiles() as an atexit() handler.
         */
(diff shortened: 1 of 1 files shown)


mrdudz:
Good catch, thanks!

Issue #2717, opened 2025-06-18
da65: Disassembler Generates Nonexistent Branch Targets


gilmorem560:
Wanted to report this one because I have ways to work around it but it's also a routinely annoying behavior. So a problem I'm running into is when the diassembler hits data segments, without a config file, it treats everything as code. This isn't too big of a deal, when I identify what segments are data I can then go back and draft a config file to reprocess it. However, in the event I don't do this, I would still like the resulting file to immediately go back through ca65 and produce a 1:1 binary image.

The problem that arises when processing arbitrary data as code is you wind up often times with completely random branches. Here's what I think is a reproducible example:

```
        .org    $E0CA
        .res    35, 0

label_a:
        sta     label_b

label_b:
        bcs     label_a+1
        lda     #$00
```

Obviously this is nonsense that nobody would write this way, but the assembly is meant to be as if label_b really represented a chunk of data. Anywho, I can assemble this into an example NES ROM like so:

```
ca65 -l test.lst test.s
ld65 -c nes -o test.bin test.o
```

And taking a look at the listing:

```
ca65 V2.18 - N/A
Main file   : test.s
Current file: test.s

000000r 1                       .org    $E0CA
00E0CA  1               
00E0CA  1  00 00 00 00          .res    35, 0
00E0CE  1  00 00 00 00  
00E0D2  1  00 00 00 00  
00E0ED  1               
00E0ED  1               label_a:
00E0ED  1  8D F0 E0             sta     label_b
00E0F0  1               
00E0F0  1               label_b:
00E0F0  1  B0 FC                bcs     label_a+1
00E0F2  1  A9 00                lda     #$00
00E0F4  1               
00E0F4  1               
00E0F4  1               
```

So the end result is that there is a branch to address $E0EE from the branch at $E0F0. Since this puts it in the middle of the absolute sta operation, in reality this winds up branching to bytes "$F0 $E0" which is then interpreted as a branch backwards. So far so good. Well, if I then take this result and immediately put it back through da65 with the -S as 0xE0BA (accounting for the 16-byte iNES header), I get basically this:

```
LE0CA:
        .res    35, 0

        .byte   $8D
LE0EE:  beq     $E0D0
LE0F0:  bcs     LE0EE
        lda     #$00
```

[...]

PR #2723, opened 2025-06-20
Allow multiple passes for da65 to resolve labels

kugelfuhr:
This adds `-d`/`--debug` and `-m`/`--multi-pass` options to the disassembler. The latter will make the disassembler run multiple preparation passes to find all addresses where labels must be placed. Without `-m` some label addresses are found in the final pass, where the disassembler cannot make use of them.

As in many other cases I would have added test but the test infrastructure is lacking for things like this one.

This should fix what is left from #2717.

### DIFF
--- a/doc/da65.sgml
+++ b/doc/da65.sgml
@@ -46,14 +46,16 @@ The assembler accepts the following options:
 ---------------------------------------------------------------------------
 Usage: da65 [options] [inputfile]
 Short options:
+  -d                    Debug mode
   -g                    Add debug info to object file
   -h                    Help (this text)
   -i name               Specify an info file
+  -m                    Run multiple passes to resolve labels
   -o name               Name the output file
   -v                    Increase verbosity
   -F                    Add formfeeds to the output
-  -s                    Accept line markers in the info file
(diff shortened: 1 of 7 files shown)


mrdudz:
> As in many other cases I would have added test but the test infrastructure is lacking for things like this one.

In such cases it would help if you could propose such test - the test infrastructure can be updated :)

kugelfuhr:
> In such cases it would help if you could propose such test - the test infrastructure can be updated :)

This file:

```
        .org    $E0ED
label_a:
        sta     label_b
label_b:
        bcs     label_a+1
        lda     #$00
```
as taken mostly from #2717 when assembled, linked and disassembled with `-m` should contain
```
LE0CB:  beq     LE0AD
```
and the label `LE0AD` of course defined instead of
```
LE0CB:  beq     * + (-30)
```

It should be possible to create an input file that needs more than just 3 preparation passes (as the one above does) by carefully constructing instructions where every new label breaks up another 3-byte instruction to form `.byte` plus a branch, but without any chance to get it into the tests this is probably wasted time.

mrdudz:
Thanks!

PR #2730, opened 2025-06-22
Mega65 target

mrdudz:
So this is the rest of #1792 - i decided against splitting it further, as it is not much anyway


mrdudz:
merging this now, no idea what else to do :)

PR #2734, opened 2025-06-24
C65 and Mega65 initial C support

mrdudz:
still WIP, just making the PR early to get the CI :)


mrdudz:
So well, this provides simple C support... stdio seems to work, for example. Lots of other things will (obviously) not work (and likely produce linker errors).

I will play around some more and see how much of conio i can make work without much effort.... and then merge soon :)

mrdudz:
heaptest fails for some reason... mmh

cpeek test is broken, but that was expected

i'm slowly getting to the point where someone with actual c65/mega65 knowledge would be helpful :) @dillof

mrdudz:
- allchars() in conio test makes the test crash for some reason? *shrug*
- the kernal reports screensize as 79x24 for some reason, so we need to wrap this one
- cursor does not work

mrdudz:
- conio mostly works
- joystick works
- stdio (printf) works

I guess i am mostly done with this

mrdudz:
Merging this now. Pinged some C65 people which may or may not show up to contribute :)

PR #2735, opened 2025-06-24
Add an optimization that removes compare instructions preceeding RTS or function calls

### LINKED ISSUE #2025: cc65: Dead code not removed by optimizer
The following code snippet does actually nothing:

```c
unsigned long c = 0xFF;
void func(void)
{
    switch (c) {
        case 0xFF: ;
    }
}
```
But it generates lots of assembler code:

```
.proc   _func: near

.segment        "CODE"

;
; switch (c) {
;
        lda     _c+3
        sta     sreg+1
        lda     _c+2
        sta     sreg
        ldx     _c+1
        lda     _c
;
; }
;
        ldy     sreg+1
        cpy     #$00
        bne     L0003
        ldy     sreg
        cpy     #$00
        bne     L0003
        cpx     #$00
        bne     L0003
        cmp     #$FF
;
; }
;
L0003:  rts

.endproc
```

Unused loads, stores and branches are already removed by the optimizer. The reason why the code above is left in place seems to be that there is no optimizer step that removes compare instructions just before an "rts" or jump out of the function. So the compares prevent all the other stuff from getting removed. 

Removing compares before "rts" should be a safe optimization since no C function returns something in the CPU flags. It is also of low complexity and probably easy to add.

kugelfuhr:
Since nothing is passed back in the flags, a compare before RTS has no effect. Since most functions will not evaluate  processor flags, but they get usually destroyed by a function call, it is also safe to remove compares before a function call.

Note: The first version of this PR removed only compares before RTS. The usefulness of this optimization is very limited, so I decided to extend it.

Fixes #2025.


kugelfuhr:
@mrdudz Is there a reason why sorted_codeopt.sh requires the `DOptXXX` variables to be sorted? They are just variables and there is no binary search, so this restriction seems unnecessary to me.

kugelfuhr:
@mrduz And while we are at it: sorted_codeopt.sh does also search backup files (*.c~) if they happen to lie around. I can add a small change that limits it to C files.

mrdudz:
Well, i just find it natural to have both sorted the same way - in particular because the "sort key" is the "name" string in those variable declarations. And based on that, it made the script easier :)

kugelfuhr:
That is one way to look at it. The other way is that the `DOptXXX` variables can be sorted for humans.

Correct for bsearch is:
```
OptCmp1
OptCmp10
OptCmp11
OptCmp2
OptCmp20
OptCmp21
OptCmp22
OptCmp3
...
```
While humans would prefer:
```
OptCmp1
OptCmp2
OptCmp3
OptCmp10
OptCmp11
OptCmp20
OptCmp21
OptCmp22
...
```
The current implementation of sorted_codeopt.sh prevents this more natural order.

mrdudz:
we could always use two digits... :)

kugelfuhr:
Which fails once we get numbers > 99. Anyway, I was just wondering about the unnecessary limitation. I've fixed the code.

How about the other change? Is it ok to only scan C files?

mrdudz:
perhaps limit to *.c|*.h ? (although we shouldnt need it in .h files... but who knows)

mrdudz:
thanks for cleaning up those scripts!

PR #2742, opened 2025-06-26
Allow "sp" as an alias for "c_sp" to avoid breaking old code

kugelfuhr:
This patch does several things:

* It allows usage of `sp` instead of `c_sp` to avoid breaking old user code.
* It adds a linker warning. When using `sp` the linker emits a warning `Symbol 'sp' is deprecated - please use 'c_sp' instead`.
* It adds a test that this warning is actually generated.
* To prevent cc65 library modules from using `sp` unintentionally, a GitHub check was added. The check runs on Linux only but I think this is ok, since the generated library modules are identical so it's not really useful to have it on windows, too.


mrdudz:
Very complete, thanks!

PR #2743, opened 2025-06-26
Do not read a file twice if it has an include guard

kugelfuhr:
This patch adds an optimization for (header) files containing include guards:

If such an include guard exists, the file is not read and parsed multiple times (as before) but duplicate inclusion is detected before opening the file and the additional overhead for reading and parsing is avoided.

The patch doesn't add anything new, it's just a performance optimization. But all "big" compilers have it, so I thought it would be cool to also add it to cc65. The nice thing is that it simply falls back to the old behavior for problematic cases. For example if a Windows user includes "t.h" and later "T.H" the optimization doesn't kick in and the file is read twice - but nothing bad happens.

To see the impact, I have generated a C source that includes `stdio.h` 1000 times. This is the result for cc65 without the patch:
```
$ time cc65 test.c

real    0m0,112s
user    0m0,104s
sys     0m0,008s
```
And this is the result with the patch:
```
$ time cc65 test.c

real    0m0,004s
user    0m0,000s
sys     0m0,004s
```
The test is not very meaningful since it depends on CPU and IO speed but it clearly shows the effect.

### DIFF
--- a/src/cc65/input.c
+++ b/src/cc65/input.c
@@ -40,6 +40,7 @@
 /* common */
 #include "check.h"
 #include "coll.h"
+#include "debugflag.h"
 #include "filestat.h"
 #include "fname.h"
 #include "print.h"
@@ -76,17 +77,6 @@ char NextC = '\0';
 
 /* Maximum count of nested includes */
 #define MAX_INC_NESTING         16
-
-/* Struct that describes an input file */
-typedef struct IFile IFile;
-struct IFile {
-    unsigned        Index;      /* File index */
-    unsigned        Usage;      /* Usage counter */
-    unsigned long   Size;       /* File size */
-    unsigned long   MTime;      /* Time of last modification */
-    InputType       Type;       /* Type of input file */
-    char            Name[1];    /* Name of file (dynamically allocated) */
-};
 
 /* Struct that describes an active input file */
 typedef struct AFile AFile;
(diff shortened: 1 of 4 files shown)


groessler:
What about files which have an unguarded area?

```
#define UNGUARDED_STUFF ...
#ifndef GUARD
#define GUARD
...
#endif
#define UNGUARDED_STUFF2 ...

```

I don't think that happens much though, but if UNGUARDED_STUFF/UNGUARDED_STUFF2 isn't defined with this change, it's not obvious to the user why it isn't defined.

kugelfuhr:
The optimization isn't applied for such files. This wouldn't work, since anything not covered by a guard macro *must* be read each time the file is included.

The patch does only cover the cases where the whole file is protected by a guard. Since that's the only situation where the file is not required to be opened/read/parsed a second time. Fortunately, this is true for most header files.

mrdudz:
I wonder how many minutes the 6502 community will save :)

guess i can merge soon, no complaints

GorillaSapiens:
it seems like this is a single small step away from implementing "#pragma once".

kugelfuhr:
> it seems like this is a single small step away from implementing "#pragma once".

Actually not. The PR relies on the internal file table that is generated for debugging info to check if a file is "equal" to some other file. If this fails, as in the example with "t.h" and "T.H", nothing bad happens, since the include guard is still in place and protects the contents of the file. This is not true for `#pragma once`. If duplicate file detection fails for `#pragma once` the compiler may spit out tons of errors. So for `#pragma once` it is an absolute requirement that duplicate files are detected 100% which needs OS specific magic.

GorillaSapiens:
> it is an absolute requirement that duplicate files are detected 100% which needs OS specific magic.

linux/macos
<pre>
#include &lt;sys/stat.h>
struct stat s;
stat("file.h", &s);
printf("inode: %lu, device: %lu\n", s.st_ino, s.st_dev);
</pre>
This gives you a unique (st_dev, st_ino) pair per file across the system.

windows
<pre>
#include &lt;windows.h>
BY_HANDLE_FILE_INFORMATION info;
HANDLE h = CreateFile("file.h", GENERIC_READ, FILE_SHARE_READ, NULL, OPEN_EXISTING, 0, NULL);
GetFileInformationByHandle(h, &info);
CloseHandle(h);
</pre>
Use (info.nFileIndexHigh, info.nFileIndexLow, info.dwVolumeSerialNumber)

kugelfuhr:
There's already a PR that implements `#pragma once`. It would suggest to discuss `#pragma once` related stuff in #2353.

mrdudz:
> There's already a PR that implements `#pragma once`. It would suggest to discuss `#pragma once` related stuff in #2353.

Seems a generally good advice for some people to not drag their pet peeves into every other ticket :)

Merging this one, thanks!

Issue #1399, opened 2021-02-11
Link drivers statically


sblendorio:
I read this documentation about statically link drivers using **co65**:
https://cc65.github.io/doc/co65.html

...but I didn't realize how to link statically, for example, **plus4-stdser.ser** for Plus4's serial port. For instance, there is a reference to **tgi_install** function, which is specific to "TGI", so I don't know which is the corresponding for plus4-stdser.

Referring to @mrdudz's comment here: https://github.com/cc65/cc65/issues/1398#issuecomment-776711735

How could I do it?

Thanks in advance,
Francesco

oliverschmidt:
https://github.com/cc65/cc65/blob/master/include/serial.h#L139

https://github.com/cc65/cc65/blob/master/include/plus4.h#L59

groessler:
You don't need to fiddle with co65. Just link the driver.

Here's an example (which doesn't work anymore since cc65 lost the `systime()` function, but it should show how to include the serial driver.

[serdump.zip](https://github.com/cc65/cc65/files/5965912/serdump.zip)

polluks:
@groessler BTW should co65.sgml be tagged "obsolete"?

groessler:
Hmm. Reading the documentation it appears that it's main purpose was to convert dynamic drivers to static ones. But some years ago already the cc65 build system was changed to create static driver object files alongside with the dynamic drivers.

I there's no other use for co65 I tend to tag it 'obsolete'.

@oliverschmidt, what do you think?

oliverschmidt:
I'd be absolutely fine with - and would in fact advocate for - removing co65 altogether right now.

oliverschmidt:
@sblendorio: Sorry if my https://github.com/cc65/cc65/issues/1399#issuecomment-777472958 above should have been too brief. Just forget about anything else and simply use the two links I mentioned. That's all to it. Your driver is loaded statically with just placing that call in your program.

polluks:
@oliverschmidt co65 should be kept for old projects...

mrdudz:
> co65 should be kept for old projects...

What would be the usecase?

polluks:
Old makefiles are still working.

mrdudz:
Fixing that one line in the makefile will be the least of your problems with making any non trivial old project work with recent cc65

polluks:
Indeed, however we are pretty backward compatible since 20 years :-)

oliverschmidt:
[...]

PR #2744, opened 2025-06-26
Added hint for co65.sgml #1399

polluks:


kugelfuhr:
Just wondering: If it is not recommended to load the drivers from disk, what's the reason for having a rather complex driver spec at all? Wouldn't it be better to split the drivers into separate linkable modules using a common API, effectively reducing overhead and memory footprint?

mrdudz:
Who said its not recommended anymore anyway? Being able to do it is the whole point of this entire setup

mrdudz:
Seriously though, it should instead tell that you can actually link a static driver - and how to do it.

The co65 tool is no more needed for this for that matter - i actually thought we had removed this thing long ago? @oliverschmidt ?

mrdudz:
I've reread #1399 now. The entire co65 section needs to be cleaned up. Or rather - removed.

> My opinion is "keep it", it might be useful in some cases. But remove all references to it when talking about static drivers.

So what isn't recommended is using co65. Not using dynamic drivers.

oliverschmidt:
> So what isn't recommended is using co65. Not using dynamic drivers.

+1

...and co65 can be removed by now from my POV.

mrdudz:
Indeed, closing here for this matter. And i'm totally willing to merge a PR that removes all traces of co65 in the sourcetree and documentation

PR #2745, opened 2025-06-26
Apple2: Fix permanently disabled IRQ

colinleroy:
Regression introduced in 990d65e:
Pushing status, initializing IRQ handler (which enables IRQ) then pulling status re-disables IRQ.

### DIFF
--- a/libsrc/apple2/crt0.s
+++ b/libsrc/apple2/crt0.s
@@ -56,9 +56,10 @@ init:   ldx     #zpspace-1
         bpl     :-
 
         ; Check for ProDOS.
-        ldy     $BF00           ; MLI call entry point
-        cpy     #$4C            ; Is MLI present? (JMP opcode)
-        php                     ; Remember whether we're running ProDOS
+        lda     $BF00           ; MLI call entry point
+        sec
+        sbc     #$4C            ; Is MLI present? (JMP opcode)
+        pha                     ; Backup the result for later
         bne     basic
 
         ; Check the ProDOS system bit map.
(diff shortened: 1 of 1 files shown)


libsrc/apple2/crt0.s
```diff
@@ -58,9 +58,11 @@ init:   ldx     #zpspace-1
         ; Check for ProDOS.
         ldy     $BF00           ; MLI call entry point
         cpy     #$4C            ; Is MLI present? (JMP opcode)
-        php                     ; Remember whether we're running ProDOS
         bne     basic
 
+        ; We are running ProDOS, disable the IRQ/RST vector patching
+        sty     patch_vectors
```
oliverschmidt, line 64: This is certainly "cool" as the $4C is already loaded but after all ONCE is declared as R/O: https://github.com/cc65/cc65/blob/master/cfg/apple2.cfg#L31

cleroymira-unabiz: Ha damn, I noticed `done` was patched and thought it was ok there :)
Do you have other ideas? using pla/and instead of plp was my first but it's hard to understand what it does at first sight, I find?

oliverschmidt: `done` is deliberately placed in `.data`.

colinleroy: I did forget to double-check that :grin:

colinleroy: That's good, yes! Updated.

oliverschmidt:
Thanks 😀

Issue #2119, opened 2023-05-18
Test Makefiles should QUIET=1 still print errors?


bbbradsmith:
I noticed that [the test actions don't report test errors](https://github.com/cc65/cc65/pull/2113#issuecomment-1553285622), which is apparently requested by ```QUIET=1``` in the make command line. However, I am wondering if this is really the behaviour we should want?

We can of course run the test locally without QUIET=1 as a workaround, but the build machine has already done the work.

Example: [test/val/Makefile](https://github.com/cc65/cc65/blob/master/test/val/Makefile)

We have a `$(NULLOUT)` and `$(NULLERR)` defined, to redirect stdout and stderr, in general. In most cases we're just using NULLERR.

My general suggestion:
1. Positive tests should not NULLERR, though we might want to add a compiler flag to suppress warnings in the QUIET=1 case?
2. Negative/NOT tests should use NULLERR since we expect the error.

We also use QUIET=1 on [libsrc/Makefile](https://github.com/cc65/cc65/blob/master/libsrc/Makefile) but it seems only about suppressing an echo listing each file built. However, it's inconsistent and should probably be reviewed as well.

mrdudz:
IMHO "QUIET" should mean "no messages except all errors"

PR #2746, opened 2025-06-26
Fix -s vs QUIET in the makefiles

mrdudz:
Inspired by #2119

The idea is that:

```make``` gives most "clean", but "all" output. in particular no stdout or stderr messages are suppressed, unless they are absolutely useless
```make QUIET=1``` gives a much less verbose output, it will suppress most output by the various tools, BUT when a tool errors out unexpectedly (!) and/or a test failed, then the output should be dumped to the console (*). Usually one status message per test is printed, to give some indicator of progress. This is the mode the CI/GHA uses
```make -s``` is really the "no messages unless errors" mode, it inherits "QUIET=1", and supresses some more messages. Usually one status message per test is printed, to give some indicator of progress. And of course on errors all related info should still be dumped to console.

(*) If some people could break random tests locally and see if that works as intended, that would be nice

PS: some of this may not work as advertised with cmd.exe - i have no idea. i tried to make the changes in a way that it would at least not break

done so far are "test" and "targettest" directories (most of "test" already worked that way), will check the rest as well and try to unify the behaviour a bit

### DIFF
--- a/targettest/Makefile
+++ b/targettest/Makefile
@@ -47,6 +47,16 @@ else
   LD := $(if $(wildcard ../bin/ld65*),../bin/ld65,ld65)
 endif
 
+ifeq ($(SILENT),s)
+    QUIET = 1
+endif
+
+ifdef QUIET
+    .SILENT:
+    PQ = "QUIET=1"
+    PD = --no-print-directory
+endif
+
 ifneq ($(filter disk testcode.%,$(MAKECMDGOALS)),)
   ifdef CC65_HOME
     TARGET_PATH = $(CC65_HOME)/target
@@ -120,10 +130,12 @@ DISK_atarixl   = testcode.atr
 %: %.s
 
 .c.o:
+	$(if $(QUIET),echo targettest/$*.c)
 	$(CC) $(CFLAGS) -Ors --codesize 500 -T -g -t $(SYS) $<
 	$(AS) $(<:.c=.s)
 
 .s.o:
+	$(if $(QUIET),echo targettest/$*.s)
 	$(AS) $(ASFLAGS) -t $(SYS) $<
 
 .PRECIOUS: %.o
@@ -721,7 +733,7 @@ endif
 
 define SUBDIR_recipe
 
-@+$(MAKE) -C $(dir) --no-print-directory $@
+@+$(MAKE) -C $(dir) $(PD) $@ $(PQ)
 
 endef # SUBDIR_recipe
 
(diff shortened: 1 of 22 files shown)


mrdudz:
but ofcourse windows blew up. gna

kugelfuhr:
Does this suppress all the useless clutter output to the screen when doing "make -s test"?

mrdudz:
> Does this suppress all the useless clutter output to the screen when doing "make -s test"?

thats the point :)

mrdudz:
so libsrc and src left... those will be most interesting :)

mrdudz:
Should be all now... look at the log of the runner, this is with QUIET=1, -s will remove a few more things. The output of the windows runner gives an idea of how it looks with just "make" now, because somehow things dont work with it (feel free to chime in and fix)

mrdudz:
PS: for those that are wondering WTH happened - try ```make -s``` :) Its different than before in src and libsrc now... but its consistent across all Makefiles (hopefully) and more "standard" (and less surprising IMHO)

Also: my argument against making it "very silent" by default, and then use V=1 to re-enable the output - and instead make it "verbose" by default and disable output on -s is:
* V=1 was not documented
* make -s is a standard option and does not have to be documented :)
* QUIET=1 was already used elsewhere, and it makes no sense when everything is quiet by default

Shoot (please propose a different scheme of how it should work - in the entire tree - if you dont agree with the OP. I am totally willing to change it to something different)

mrdudz:
> The output of the windows runner gives an idea of how it looks with just "make" now

nah not really :) just "make" gives you... all and everything

oliverschmidt:
I'm not 100% sure why I'm asked to review. So just a general remark:

At the time I maintained this project, I considered it very relevant that it can be built on Windows. That had - as far as I remember - two reasons:

1. To not lock out Windows users from contributing to cc65. I believe that with availability of WSL2, this reason isn't valid anymore.

2. To not keep a certain contributor who relied on Visual Studio (not Visual Studio Code) from further contributing. Since I stopped maintaining cc65, I have no overview whatsoever if this is considered valid anymore (or in fact if it was considered valid by anybody but me in the first place).

As a result I imagine getting rid of everything targeting cmd.exe as build environment is a totally viable option.

mrdudz:
> I'm not 100% sure why I'm asked to review.

Mostly because you wrote a lot of the Makefiles - and this PR changes the behaviour radically for at least libsrc and src (before it was non chatty by default, now you need to use -s)

(i'd like to keep cmd.exe support around for a while. if only for @acqn - should he ever come back :))

oliverschmidt:
Yes, I wrote those Makefiles - and I know that in several aspects you have a different perspective than I have (or rather had back then). Please go ahead and change them in any way you see fit. I'm a pure user of them - on WSL2 by now. As long as they still produce their files (be it with more or less screen output) I'm perfectly fine.

Yes, @acqn is the contributor I had in mind above.

mrdudz:
> Please go ahead and change them in any way you see fit.

Ah well :) right - merging. We can still change it to something else should that be desired

Issue #2605, opened 2025-02-14
Unable to install the cc65 binaries under macOS due to insufficient permission


wyatt-wong:
I have build the cc65 source into binary under macOS but when I run `make install`, I got the following permission error.

```
BUILD_ID: Git b75f872fe
install -d /bin
install: chmod 755 /bin: Read-only file system
install ../bin/* /bin
install: /bin/ar65: Read-only file system
make[1]: *** [Makefile:115: install] Error 71
make: *** [Makefile:6: install] Error 2
```

I suggest to revise the installation path to /usr/local/bin instead of /bin folder.

Furthermore, please also include an "uninstall" section in the Makefile.

mrdudz:
I'm surprised it tries to install to /bin (/usr/local/bin should be the default indeed)

but changing that wont fix the permission - you'd still need "sudo make install" in any case

rofl0r:
`make PREFIX=/usr/local && sudo make PREFIX=/usr/local install`

wyatt-wong:
> `make PREFIX=/usr/local && sudo make PREFIX=/usr/local install`

I think the Makefile need to be modified instead of running the `make` command with parameter. The general practice to build and install applications with the following commands:

```
make or make all
make install
make uninstall
make clean
```

Furthermore in macOS convention, running `make install` should NOT require `sudo`. The `sudo` command is generally used in Linux such as Ubuntu, Debian, CentOS, etc.

wyatt-wong:
> I'm surprised it tries to install to /bin (/usr/local/bin should be the default indeed)
> 
> but changing that wont fix the permission - you'd still need "sudo make install" in any case

Maybe you should test it in macOS by running the `make` command first and then `make install‘ if you have queries to my error messages.

rofl0r:
> Furthermore in macOS convention, running make install should NOT require sudo. 

Citation needed. `sudo` is part of OSX since the first days precisely to be able to get root rights in the terminal - e.g. to install stuff into a root-owned directory like /usr/local.

> I think the Makefile need to be modified instead of running the make command with parameter. 

The Makefile caters to a variety of people, and one use case it supports is to run the binaries from inside the source tree. Maybe you should read the docs instead of demanding that stuff be changed to meet your expectations.

mrdudz:
Nevertheless - /usr/local should be the default prefix when none was used. using a blank prefix is just broken

mrdudz:
[...]

Issue #2061, opened 2023-04-23
Error in running "make install"


wyatt-wong:
I run `make` to build the source but then when I run `make install`, I got the following errors:

wyattwong@WyattsMacBookPro cc65 % make install
BUILD_ID: Git e5f4ca6b8
install -d /bin
install: chmod 755 /bin: Operation not permitted
install ../bin/* /bin
install: /bin/ar65: Operation not permitted
make[1]: *** [Makefile:115: install] Error 71
make: *** [Makefile:6: install] Error 2
wyattwong@WyattsMacBookPro cc65 % 

What am I missing ?

wyatt-wong:
> I run `make` to build the source but then when I run `make install`, I got the following errors:
> 
> wyattwong@WyattsMacBookPro cc65 % make install BUILD_ID: Git [e5f4ca6](https://github.com/cc65/cc65/commit/e5f4ca6b898bf581751901fd5792764adfb7730a) install -d /bin install: chmod 755 /bin: Operation not permitted install ../bin/* /bin install: /bin/ar65: Operation not permitted make[1]: *** [Makefile:115: install] Error 71 make: *** [Makefile:6: install] Error 2 wyattwong@WyattsMacBookPro cc65 %
> 
> What am I missing ?

I found out the answer from [https://github.com/cc65/wiki/wiki/OS-X](url). However I found the Getting Started URL [http://cc65.github.io/cc65/getting-started.html](url) is incorrect. The new URL should be [https://cc65.github.io//getting-started.html](url). Unfortunately I have no idea how to create a pull request to update the link in [https://github.com/cc65/wiki/wiki/OS-X](url)

mrdudz:
I fixed the link... the way github wiki works is a bit annoying, you cant make it editable for the public.

Thanks!

rofl0r:
> install -d /bin

this looks wrong, i'm pretty sure MacOS uses another prefix. maybe the getting started for OSX should mention to run make with the correct prefix (whatever it is).
btw, the getting started for linux mentions to run `make avail` instead of `make install` which is probably not what the average user wants.

wyatt-wong:
> > install -d /bin
> 
> this looks wrong, i'm pretty sure MacOS uses another prefix. maybe the getting started for OSX should mention to run make with the correct prefix (whatever it is). btw, the getting started for linux mentions to run `make avail` instead of `make install` which is probably not what the average user wants.

[...]

PR #2747, opened 2025-06-26
see #2605, #2061, #1735, probably others

GorillaSapiens:
i obviously don't have all these systems to test with.
let the argument commence...


mrdudz:
this makes sense... would be nice if it:
* prints DESTDIR (if not windows detected)
* prints a warning if DESTDIR is empty (before setting the default, only if not windows)
* prints a warning if "unknown" was detected (just that might result in some bsd or amiga guy to wake up...)

GorillaSapiens:
> * prints a warning if "unknown" was detected (just that might result in some bsd or amiga guy to wake up...)

if it's not windows, "/usr/local" should be fine.  we could add AIX, Solaris, etc... but there's really no point in testing for a thing if it doesn't change what we do.

i suspect your hypothetical "amiga guy" will already know to be on the lookout for things to change.  without an amiga to test on, and with no subject matter experts on the team, i would just not bother, and wait for someone who cares and knows what they are doing tell us what they need changed.

mrdudz:
> i suspect your hypothetical "amiga guy" will already know to be on the lookout for things to change. without an amiga to test on, and with no subject matter experts on the team, i would just not bother, and wait for someone who cares and knows what they are doing tell us what they need changed.

There is @polluks ... and i have an amige right here as well :)

In any case i am a friend of not doing things blindly - a warning in that case is totally justified

(also this is another thing that assumes we are compiling on the host... oh well)

GorillaSapiens:
> > i suspect your hypothetical "amiga guy" will already know to be on the lookout for things to change. without an amiga to test on, and with no subject matter experts on the team, i would just not bother, and wait for someone who cares and knows what they are doing tell us what they need changed.
> 
> There is @polluks ... and i have an amige right here as well :)
> 
> In any case i am a friend of not doing things blindly - a warning in that case is totally justified
> 
> (also this is another thing that assumes we are compiling on the host... oh well)

if i'm wrong and we DO have amiga experts here, we should add the test and they should just tell me what to use for a default DESTDIR

mrdudz:
> if i'm wrong and we DO have amiga experts here, we should add the test and they should just tell me what to use for a default DESTDIR

The problem starts with "compiling on the host". Your test would just give "linux" or "windows"

Anyway, useless discussion. That warning just should be there

GorillaSapiens:
> The problem starts with "compiling on the host". Your test would just give "linux" or "windows"

sounds like you want a "canadian cross" compiler, which is a whole different set of issues.  i would even argue that this is better handled outside the scope of this project entirely.

> Anyway, useless discussion. That warning just should be there

NOT "useless".  if we have amiga experts here who can tell us what a good default DESTDIR is for that platform, let's hear it !!!
i would MUCH rather "get it right" than "issue a lame warning that we probably got it wrong".
even if the right answer is "leave DESTDIR blank for Amiga", we CAN do that, it's not a problem.
i just need to know, i am not an Amiga expert.

mrdudz:
> sounds like you want a "canadian cross" compiler, which is a whole different set of issues. i would even argue that this is better handled outside the scope of this project entirely.

no, just a regular cross compiler. and it is a pretty common thing (cc65 is one). Hardly anyone compiles natively on Amiga, not larger make based projects like this one anyway.

Plus: i don't want to see that Makefile becoming a huge mess of conditionals just to try to be smart and serve every OS automatically. I have written, and had to maintain, such crap before and don't want to go there again. (there will definitely be no AIX or Solaris in this Makefile. Or a dozen BSD. just no.)

So yes, a minimal solution that works on "the big three" is fine with me. But anything else gets a warning. Please.

GorillaSapiens:
> > sounds like you want a "canadian cross" compiler, which is a whole different set of issues. i would even argue that this is better handled outside the scope of this project entirely.
> 
> no, just a regular cross compiler. and it is a pretty common thing (cc65 is one). Hardly anyone compiles natively on Amiga, not larger make based projects like this one anyway.
> 
> Plus: i don't want to see that Makefile becoming a huge mess of conditionals just to try to be smart and serve every OS automatically. I have written, and had to maintain, such crap before and don't want to go there again. (there will definitely be no AIX or Solaris in this Makefile. Or a dozen BSD. just no.)
> 
> So yes, a minimal solution that works on "the big three" is fine with me. But anything else gets a warning. Please.

A Canadian Cross Compiler is a compiler built on machine A, that runs on machine B, and produces code for machine C. The classic use case is when you're building a cross-compiler on one system for a completely different target system, and the compiler itself is going to be run on a third system.

is that not what you're describing here?  it sounds like the Amiga compiler is created on a Linux machine.
i would like to understand better how the Amiga case is handled.  like i said, i'm not an expert here.

mrdudz:
I know what that stuff is. Amiga stuff is usually compiled with regular crosscompilers (like cc65 is). What matters is what platform will run "make". However, please lets stick to the topic. None of this matters.

GorillaSapiens:
> I know what that stuff is. Amiga stuff is usually compiled with regular crosscompilers (like cc65 is). What matters is what platform will run "make". However, please lets stick to the topic. None of this matters.

calm down.  it does matter.  i would like to do the right thing.  to do the right thing, i need to understand what's going on.

asking these questions should not annoy you.  even if you think they are irrelevant, i would like the answers so i can understand why you think they are irrelevant.

so please answer my questions.

the cc65 compiled for AmigaOS.  is that compile hosted on Amiga OS?  it sounds like the compile is hosted on a linux system, which would make this a canadian cross.  is that correct?  if it's not correct, can you explain the process to me?

i'm not just going to blindly do what you say because you say so.  i am not your trained monkey.  like i said, i'm asking because i want to understand and do the right thing.

mrdudz:
> calm down. it does matter. i would like to do the right thing. to do the right thing, i need to understand what's going on.

The right thing is: emit a warning. And the reason for why it doesn't matter was already given to you:

> Plus: i don't want to see that Makefile becoming a huge mess of conditionals just to try to be smart and serve every OS automatically.

GorillaSapiens:
> > calm down. it does matter. i would like to do the right thing. to do the right thing, i need to understand what's going on.
> 
> The right thing is: emit a warning. And the reason for why it doesn't matter was already given to you:
> 
> > Plus: i don't want to see that Makefile becoming a huge mess of conditionals just to try to be smart and serve every OS automatically.

nope.  not your trained monkey.  i still don't understand how the Amiga case is handled.  please explain it to me.

mrdudz:
> please explain it to me.

What is your intend? Add some voodoo to detect the Amiga case in the Makefile, so it won't get merged? I don't get it.

AND: none of this makes the warning obsolete. "unknown" means warning.

rofl0r:
this PR is ill-advised:

1) it uses DESTDIR in place of prefix
2) it uses uname which breaks cross-compilation (for example  linux user compiling for windows using mingw)

GorillaSapiens:
> > please explain it to me.
> 
> What is your intend? Add some voodoo to detect the Amiga case in the Makefile, so it won't get merged? I don't get it.
> 
> AND: none of this makes the warning obsolete. "unknown" means warning.

my "intend" is to understand what's going on.

your refusal to answer on the assumption that I might take an action you disapprove of presumes intent without evidence. not only is this logically unsound, it borders on petulance... punishing a hypothetical before it even exists.  declining to respond based on what you think I might do, not what I’ve done, is a textbook example of preemptive moralizing. it’s irrational, and frankly, juvenile.

i've asked the question at least twice now.  i shouldn't have to ask it again.

GorillaSapiens:
> this PR is ill-advised:
> 
> 1. it uses DESTDIR in place of prefix
> 2. it uses uname which breaks cross-compilation (for example  linux user compiling for windows using mingw)

can you explain why "prefix" is preferred over "DESTDIR" ?

the intent is to give a reasonable default to DESTDIR for linux and macOS systems.  an empty DESTDIR on these systems leads to an install in /bin, which everyone agrees is not right.  /usr/local/bin is preferred in these cases.

uname is only used on non windows systems.

what you are describing (linux user compiling for windows using mingw) is another canadian cross scenario.  is this a normal / expected use case?  people doing a canadian cross KNOW they're going to have additional hurdles.  is it worth breaking the more normal cases (linux user compiling for linux / macOS user compiling for macOS) for this?

more importantly, in all of these canadian cross scenarios, is the user really going to be doing a "make install" in a way where DESTDIR would even matter?

mrdudz:
> this PR is ill-advised:
> 
> 1. it uses DESTDIR in place of prefix
> 2. it uses uname which breaks cross-compilation (for example  linux user compiling for windows using mingw)

Yeah 1. should be fixed elsewhere too... and yes. What i said. That's why i don't really like those automagic things.

rofl0r:
there is no reasonable default for DESTDIR. it's a separate mechanism for packagers that's meant to be explicitly set on the command line. read the docs. https://www.gnu.org/prep/standards/html_node/DESTDIR.html . currently with your bogus PR someone who would set prefix to /usr/local in the agreed-upon manner, would end up with cc65 in /usr/local/usr/local/bin/cc65 .

GorillaSapiens:
> there is no reasonable default for DESTDIR. it's a separate mechanism for packagers that's meant to be explicitly set on the command line. read the docs. https://www.gnu.org/prep/standards/html_node/DESTDIR.html . currently with your bogus PR someone who would set prefix to /usr/local in the agreed-upon manner, would end up with cc65 in /usr/local/usr/local/bin/cc65 .

that's fair, and i understand now.  thank you.

in the future, please refrain from being a complete jackass by calling PRs "bogus".
it is intentionally inflammatory, and is not at all helpful.

GorillaSapiens:
> That's why i don't really like those automagic things.

if that is the case, and you don't want to do anything at all, you should close #2605 as RESOLVED / AS DESIGNED / WON'T FIX and just suffer the eternal customer support burden of explaining to newbs why it doesn't "just work straight out of the box".

or if you think #2605 is a legitimate bug, please explain how you think it should be fixed in a way that isn't one of "those automagic things".

*shrug*

rofl0r:
> in the future, please refrain from being a complete jackass by calling PRs "bogus".

"bogus" is way less inflammatory than "jackass".

you mention https://github.com/cc65/cc65/pull/1735 in your title, yet with the latest changes here you break that exact use case. it's perfectly valid to set PREFIX= (i.e. empty).

if there's anything to be done here it's basically:

```make
# default to /usr/local , which can be overridden via make PREFIX=whatever
PREFIX=/usr/local
if WINDOWS_DETECTED
PREFIX=
endif
```

or alternatively, without any automagic:
```make
PREFIX=/usr/local
```
and the documentation mentioning that for a "run from source-tree" build, such as desired by the windows folks, "make PREFIX=" should be used. that would make it so that the build system behaves like any other project's, but leaving the option open for that special use case.

GorillaSapiens:
> > in the future, please refrain from being a complete jackass by calling PRs "bogus".
> 
> "bogus" is way less inflammatory than "jackass".
> 
> you mention #1735 in your title, yet with the latest changes here you break that exact use case. it's perfectly valid to set PREFIX= (i.e. empty).
> 
> if there's anything to be done here it's basically:
> 
> ```makefile
> # default to /usr/local , which can be overridden via make PREFIX=whatever
> PREFIX=/usr/local
> if WINDOWS_DETECTED
> PREFIX=
> endif
> ```

to be fair, i was looking at DESTDIR when i started.

so it looks like you're lobbying for
<pre>
ifneq ($(OS),Windows_NT)
ifndef PREFIX
    PREFIX := /usr/local
endif
endif
</pre>
which should set the PREFIX to /usr/local if it's not set on any non windows machine.

that i could support.  really i just want installs to NOT default to /bin/ on linux and macOS.  maybe i was getting too fancy thinking i could give future people a HOST_OS variable.

kugelfuhr:
> in the future, please refrain from being a complete jackass by calling PRs "bogus". it is intentionally inflammatory, and is not at all helpful.

Thank you. I had a good laugh 😁

polluks:
By the way an Amiga should use the Amiga Installer. However cc65 is a special case, in the past I made a bin and a src archive.

GorillaSapiens:
> By the way an Amiga should use the Amiga Installer. However cc65 is a special case, in the past I made a bin and a src archive.

i still don't know how the amiga install works.  no one will tell me.  the secret seems more closely guarded than the Masonic Temple.

are you telling me the Amiga install doesn't care about PREFIX ?

mrdudz:
Amiga Systems are usually very customized, so there really is no good default for what goes where (which is why Amiga programs typically use the Installer). A binary package of cc65 would likely be compiled without PREFIX (or DESTDIR) set. And certainly it wont install "somewhere" automatically without asking - Amiga users would hate you for that :)

mrdudz:
As for the PR: this will still interfere with the (very legit) usecase that you actually want to pass an empty prefix.

Its probably the best idea to just warn if prefix is empty, and *suggest* a reasonable default

GorillaSapiens:
> As for the PR: this will still interfere with the (very legit) usecase that you actually want to pass an empty prefix.
> 
> Its probably the best idea to just warn if prefix is empty, and _suggest_ a reasonable default

nope, you can still pass in an empty PREFIX if you want:
<pre>
gorilla@guest:~/cc65 (multiple_os_makefile)$ make PREFIX="" install
Using Makefile: /home/gorilla/cc65/Makefile install
Using Makefile: /home/gorilla/cc65/src/Makefile install
BUILD_ID: Git 316a0104a
install -d /bin
install ../bin/* /bin
install: cannot create regular file '/bin/ar65': Permission denied
[ ... ]
</pre>

or

<pre>
gorilla@guest:~/cc65 (multiple_os_makefile)$ make PREFIX= install
Using Makefile: /home/gorilla/cc65/Makefile install
Using Makefile: /home/gorilla/cc65/src/Makefile install
BUILD_ID: Git 316a0104a
install -d /bin
install ../bin/* /bin
install: cannot create regular file '/bin/ar65': Permission denied
[...]
</pre>

GorillaSapiens:
> Amiga Systems are usually very customized, so there really is no good default for what goes where (which is why Amiga programs typically use the Installer). A binary package of cc65 would likely be compiled without PREFIX (or DESTDIR) set. And certainly it wont install "somewhere" automatically without asking - Amiga users would hate you for that :)

sounds like you're NOT running "make install" on Amiga, in which case the value of PREFIX is irrelevant.

but i don't know, because you STILL won't tell me how it is done.

PR #2749, opened 2025-06-27
hotfix for master

mrdudz:
sorry for breaking :(

@kugelfuhr any idea? i tried with .cpu and the bitfields... but that only produces problems (complains about non constant expressions...), and including the cpu macpac in an .inc file is also... quirky :(


kugelfuhr:
Sorry. These changes should have been part of my PR. Regarding the constants for evaluating .CPU - I have some thoughts about it and will open an issue.

PR #2751, opened 2025-06-27
Clean up supported CPUs/Instruction sets

mrdudz:
In a nutshell, this will:

- degrade the old "65C02" mode to be "CMOS with Rockwell extensions" (no wai/stp)
- add new CPU "W65C02" (has wai/stp). IE like "65C02" before, but with WDC Extensions
- add new CPU "65CE02" (CMOS with CSG Extensions, which includes 65SC02 and 65C02 and is the base for 4510)
- add/update documentation to say exactly what is what

After this, the instruction sets for the various CPUs can be defined in the ".CPU" bitfield more finegrained that before, and related glitches can be avoided (hopefully)


mrdudz:
OK well, some more tests (asm/opcodes...) are missing (will do that tomorrow), besides that it should be all

@kugelfuhr @spiro-trikaliotis @oliverschmidt @groessler et al (is there a way to just ping all... sigh)
please skim over it, it shouldnt be a complex change - but a small mistake or oversight here has potentially big impact

mrdudz:
Of course the tests revealed some oversights already - as they should :)

Not much more to do now other than braindump into docs :)

mrdudz:
so thats good to go from me.... (Sweet16 disassembler support is missing, but thats for another PR)

mrdudz:
A well i'm merging this now... @kugelfuhr you may go on with this "feature" pseudovariable, i am done in this area :)

PR #2755, opened 2025-06-29
Some more Makefile woes

mrdudz:
nothing spectacular...


GorillaSapiens:
is EVERY make now going to warn about "PREFIX not set"?  seems more noisy than necessary.
the check in "install" should be sufficient.

why does someone running "make zip" or "make" care whether or not the PREFIX is set?

mrdudz:
> is EVERY make now going to warn about "PREFIX not set"? seems more noisy than necessary.

But printing every makefile name isnt? are you kidding?

But right, zip shouldnt do it. Anything building the compiler/toolchain should though. (the prefix is baked into the binary)

GorillaSapiens:
> > is EVERY make now going to warn about "PREFIX not set"? seems more noisy than necessary.
> 
> But printing every makefile name isnt? are you kidding?
> 
> But right, zip shouldnt do it. Anything building the compiler/toolchain should though. (the prefix is baked into the binary)

i don't mind noisy makefiles. ;)

they get baked in?  this seems problematic then.  if i "make" w/o PREFIX then do a "make" WITH PREFIX, it should rebuild everything.  or at least anything that uses PREFIX.  current makefiles don't do this, and neither does this PR.

GorillaSapiens:
> > is EVERY make now going to warn about "PREFIX not set"? seems more noisy than necessary.
> 
> But printing every makefile name isnt? are you kidding?
> 
> But right, zip shouldnt do it. Anything building the compiler/toolchain should though. (the prefix is baked into the binary)

also, lol, it seems AMIGA and WINDOWS ignores all this altogether anyway, so defaulting PREFIX to "/usr/local" would be harmless...

```c
void FinishIncludePaths (void)
/* Finish creating the include path search lists. */
{
    /* Add specific paths from the environment */
    AddSearchPathFromEnv (SysIncSearchPath, "CC65_INC");
    AddSearchPathFromEnv (UsrIncSearchPath, "CC65_INC");

    /* Add paths relative to a main directory defined in an env. var. */
    AddSubSearchPathFromEnv (SysIncSearchPath, "CC65_HOME", "include");

    /* Add some compiled-in search paths if defined at compile time. */
#if defined(CC65_INC) && !defined(_WIN32) && !defined(_AMIGA)
    AddSearchPath (SysIncSearchPath, CC65_INC);
#endif

    /* Add paths relative to the parent directory of the Windows binary. */
    AddSubSearchPathFromBin (SysIncSearchPath, "include");
}
```

mrdudz:
> they get baked in? this seems problematic then.

srsly? This is standard for pretty much everything on *nix.

> if i "make" w/o PREFIX then do a "make" WITH PREFIX, it should rebuild everything. or at least anything that uses PREFIX.

I have yet to see a make based project that does that. Expected behaviour is to "make clean" when you use a different PREFIX.

> defaulting PREFIX to "/usr/local" would be harmless...

it makes the difference between "works by chance" and "works by design"

rofl0r:
this breaks a valid usecase i implemented in https://github.com/cc65/cc65/pull/1735

```
~/tmp/cc65[master]$ make PREFIX= DESTDIR=foo install
Using Makefile: /home/rofl/tmp/cc65/Makefile install
Makefile:28: *** Error: PREFIX must be set for install to work.  Stop.
```

mrdudz:
> this breaks a valid usecase i implemented in #1735
> 
> ```
> ~/tmp/cc65[master]$ make PREFIX= DESTDIR=foo install
> Using Makefile: /home/rofl/tmp/cc65/Makefile install
> Makefile:28: *** Error: PREFIX must be set for install to work.  Stop.
> ```

Gna... forgot DESTDIR about all the fuckery in that other ticket. Will check that too...

PR #2759, opened 2025-06-30
Fixed #1970 (Missing definition for ST on cbm610 with getdevice) (was #1972)

### LINKED ISSUE #1970: Missing definition for ST on cbm610 with getdevice

mrdudz:
@polluks @polluks2 @kugelfuhr @WartyMN so here it is again... please test a bit - will merge soon i guess, i don't see why not at least :)

### DIFF
--- a/asminc/cbm510.inc
+++ b/asminc/cbm510.inc
@@ -15,6 +15,8 @@ IndReg          := $01          ; Controls indirect indexed load-store bank
 
 TXTPTR          := $85          ; Far pointer into BASIC source code
 FNAM            := $90          ; Far pointer to LOAD/SAVE file-name
+FNAM_SEG        := $92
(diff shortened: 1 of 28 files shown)


mrdudz:
So, no complaints? :)

mrdudz:
Apperently not :)

polluks:
Maybe ONCE can be used for the init.

Issue #2753, opened 2025-06-29
Constants for evaluating .CPU need some rework


kugelfuhr:
There have some changes to `.cpu` and the constants to evaluate it lately. When looking at #2751 and #2749 I noticed that the constants have some problems.

The first one is that `.cpu` is internal to the assembler, but the constants to evaluate it are loaded via an include file. That's a bit strange, since `.cpu` cannot be used without these constants, and the include file is also tied closely to a specific assembler version. In my eyes, the constants should be assembler generated symbols instead. This would be cleaner and it would also allow to use these constants within assembler include files like zeropage.s resolving problems like in #2749. The symbols would have to be different from the current ones, since the existing ones are in the user name space. This could be achieved by using names using underscores or by moving them into a scope that has a name starting with an underscore. The existing include file should then map these names to the old ones and kept for some time for compatibility reasons.

The next thing is that `.cpu` describes existing CPU instruction sets. Unfortunately the instruction sets do not build nicely one upon the other. So it might be an idea to also include some information into the bits returned by `.cpu` that are not really a full instruction set. What I mean is for example that one bit might tell if the CPU supports `BRA` for an 8 bit distance. Currently this is supported by 65SC02, 65C02 and 65816 (I think) for which one is a superset of the other, so checking for 65SC02 is enough to find out if `BRA` is supported - for the three CPUs listed. But there's also the M740 which has `BRA` but is missing some other additional opcodes the 65SC02 has.

In most cases, `.cpu` is used to check for a specific instruction or addressing mode. So it might be an idea to either add a few bits to `.cpu` that tell if these instructions or addressing modes are available, or add another pseudo constant that contains this information. It might contain information about the following:

* `BRA` (short branch with 8 bit displacement)
* `PHX`/`PLX`/`PHY`/`PLY` (save and restore X/Y)
* `STA (zp)`/`LDA (zp)` (zero page indirect addressing for loads and stores) 
* `DEA`/`INA` (decrement and increment of the accumulator)
* Maybe others

mrdudz:
Having those constants indicate individual instructions isnt terribly useful IMHO. In reality you'll end up writing specific code for specific CPUs anyway, and not ifdef every other instruction. I'd say this is a completely different problem from the first issue, and i wouldn't bother much about it. But if we really want such kind of mechanism, i'd vote for a seperate one from .cpu

kugelfuhr:
Then, please look at the library. You'll find tons of stuff like:

```
[...]

PR #2760, opened 2025-07-01
Add ".CAPABILITY" to ca65, remove ".MACPACK cpu"

kugelfuhr:
This PR contains the changes discussed in #2753. It adds a new pseudo function `.capability` to ca65 and allows to query some capabilities. Others may be added later.

The PR will also make the constants from "cpu.mac" internal constants generated by the assembler and replace "cpu.mac" by a one liner that just emits a warning when used.

I have also updated most of the library to use `.cap` instead of the old `.cpu`. One exception is the apple library where things are a bit complex sometimes, and as far as I know the apple was built with just two cpus, so I expect the old code to still work (but it might have to be checked because of the incompatible changes to `.cpu` from #2751).

The docs have also been updated with one exception: Explaining the new behavior of `.cpu` and the `CPU_ISET_xxx` constants is so cringey  that someone else will have to do that. I have copied the old text from `.macpack cpu` to `.cpu` and changed part of it but the part that explains the `CPU_ISET_xxx` stuff is commented out.

@mrdudz Please check the per cpu capabilities in common/capability.c for correctness.


mrdudz:
> Please check the per cpu capabilities in common/capability.c for correctness.

Will do ... i find this hard to do properly to be honest - since i also don't know all instruction sets by heart. And certainly not the details, like what added what address modes to which instructions...

The problem i see is - if we are not very careful now with how this is implemented and defined, then we'll end up in a similar situation as with the .cpu bits (and we absolutely want to avoid that, right?)

If we have a bit for an address mode there, what exactly does this mean? Sometimes CPUs only added new address modes to a few instructions, then later even more address modes were added. How do we deal with this?

Or in other words: if two CPUs have the bit for an address mode set, does that imply both CPUs (ALL CPUs with that bit set) support the exact same instructions with that address mode?

(i have not checked if this is really a problem, right now, with the CPUs we already support - but we should still be able to handle such things, or someone WILL add another CPU that breaks it :))

I think we need a test for this verification.... ideally one we can somehow generate so we don't have to review it the same way and not win anything.

kugelfuhr:
> If we have a bit for an address mode there, what exactly does this mean? Sometimes CPUs only added new address modes to a few instructions, then later even more address modes were added. How do we deal with this?

You mean `CPU_HAS_ZPIND`? Currently this means it supports all instructions with "zp indirect" addressing mode that the 65SC02 has. If there is another CPU that adds this addressing mode for more instructions (or removes it from some), we have to exactly list the instructions in the docs and add a new capability that covers the new CPU. Fortunately the number of capabilities is not limited. The difficult part is to come up with reasonable sounding identifiers. But for now I wouldn't worry about it.

mrdudz:
> You mean CPU_HAS_ZPIND?

Sure - I have deliberately tried to word it some more generic though, not specifically referring to this particular one. As said, i don't want to run into the same situation as with the .cpu stuff at some point. Ie ideally what is defined now, never needs to be changed/fixed-

> Currently this means it supports all instructions with "zp indirect" addressing mode that the 65SC02 has.

The docs should then state exactly that, imho. This way its 100% clear what it means. (maybe put a link to the list of sc02 instructions on the cpu page as well?)

kugelfuhr:
Ok. I will add that to the docs.

mrdudz:
Ok merging, will have a look at the .cpu docs. Thanks!

PR #2763, opened 2025-07-01
Make the scrcode macros handle identifiers

### LINKED ISSUE #534: scrcode macro does not handle labels
At least using the `asminc/apple2.mac` version, this works:
```asm
        scrcode "hello", $D, "world", $7
```
But this does not:
```asm
        CR := $D
        BELL := $7
        scrcode "hello", CR, "world", BELL
```
Producing a **scrcode: invalid argument type** error.

I believe you can address this by adding a clause to the if/elseif for labels, e.g.:
```asm
        .local XXX

        ; Check for a label
        .elseif .match (.left (1, {arg1}), XXX)
                _scrcode        arg1
```

kugelfuhr:
This fixes #534.

Tested with this program:

```asm
.ifdef __ATARI__
        .macpack        atari
.endif
.ifdef __CBM__
        .macpack        cbm
.endif
.ifdef __APPLE2__
        .macpack        apple2
.endif

CR = $D
BELL = $7
scrcode "hello", CR, "world", BELL
```
No idea how to add this to the tests so it's assembled and checked for Atari, cbm,  and Apple2.

### DIFF
--- a/asminc/apple2.mac
+++ b/asminc/apple2.mac
@@ -36,6 +36,12 @@
                 ; Just output the character
                 _scrcode        arg1
 
+        ; Check for an identifier
+        .elseif .match (.left (1, {arg1}), identifier)
+
+                ; Just output the identifier
+                _scrcode        arg1
+
         ; Anything else is an error
         .else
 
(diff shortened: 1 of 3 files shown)


mrdudz:
OK this is simple enough to grasp in this heat (will check the others later, i am melting...)

PR #2767, opened 2025-07-03
Remove invalid characters from labels in the VICE label file

### LINKED ISSUE #836: Labelfile causing ERR:syntax error
After some hunting I seem to have eventually found the cause for VICE returning `ERR:syntax error` to me. Seems that when I feed it with my labelfile, the line:
```
    al 000E66 .LOCAL-MACRO_SYMBOL-0000
```
is what makes it unhappy. Looks like "minus" signs are understandably causing the trouble. Is this a known thing? If not - what should we use there instead of the un-kosher character?

kugelfuhr:
## Summary

When creating a VICE label file, replace characters in labels that VICE doesn't understand with '_'. Fixes #836.

## Checklist

Test file used: 

```
.macro  mac
        .local  Lab
Lab:    lda     #$00
.endmacro
mac
mac
mac
```

Translate using `bin/cl65 -g -t none -Ln test.lbl test.s` and check that the VICE label file contains '_' instead of '-' in the identifiers created by the macro. No idea how to automate this.


mrdudz:
I had a quick peek at the VICE code (*) - the labels are actually quite limited:

1) strictly speaking, valid characters are a-z, A-Z, 0-9 only (this is when you assign a label with a value in the monitor). Why this is different from the add_label command... no idea. Weird.
2) there are a bunch of reserved names (all register names), which may or may not be a problem in some expressions (that unfortunately includes all register names of all CPUs in VICE - eg z80 and 6809)

This seems very.... halfbaked actually (in particular #1). :(

Not sure if we should bother in cc65. I am leaning to blame it on VICE (and i already hate the idea that i will have to fix lex/yuck stuff). The labels that contain any non alphanumeric characters can't be assigned (not a big problem i guess) and the labels starting with @ will not work in breakpoint expressions (its used as a dereferencing operator).

(I am surprised now that we didnt get bug reports about this for VICE - ppl must only use the very basic features of the monitor/debugger)

*) https://sourceforge.net/p/vice-emu/code/HEAD/tree/trunk/vice/src/monitor/mon_lex.l

kugelfuhr:
I don't think that this is a VICE problem. Yes, VICE should be as liberal as possible, and it might be a good idea to open the VICE symbols up to more characters. But VICE simply cannot include all characters that may be used by every tool out in the wild, so there will always be some that won't work.

So actually VICE defines what is legal input and ld65, when creating a "VICE label file", has to adhere to that definition. This is what this PR is for.

Regarding cheap labels in breakpoint expressions: They are almost useless for this purpose, since cheap label names tend to be reused. So there are usually many labels named `@L1` or whatever, and VICE cannot cope with that anyway. It uses the last one (or maybe the first one) in the label file and all others are ignored. That makes them useless for breakpoints.

When we do "not bother in cc65", label files generated by ld65 cannot be used in VICE whenever there are macros with local labels or any other internally generated labels. This is definitely a problem.

mrdudz:
Good point about reused locals, so not much of an issue then with @

I should inspect VICE a bit more - that apparently  the "al" command and assignments match the labels differently is really strange 

> When we do "not bother in cc65", label files generated by ld65 cannot be used in VICE whenever there are macros with local labels or any other internally generated labels. This is definitely a problem.

Oh by "don't bother" i was referring to those weird problems, not your fix. That should be in, of course.

Will look closer later...some real work is waiting :(

mrdudz:
```
(C:$05b6) al 13 .@foo_bar
(C:$05b6) cond 1 if @io:.@foo_bar == 5
Setting checkpoint 1 condition to: @io:$0013 == $05
```

It actually works fine, sorry for the fuzz :)

Merging...

PR #2768, opened 2025-07-03
disable the basic interrupt before main(), fixes #2764

### LINKED ISSUE #2764: C128 Sprite initialization requires disabling BASIC part of the IRQ handler
When setting up and turning on a sprite targeting native Commodore 128 mode **one of** these statements needs to be executed first:

asm("sei"); 

OR

POKE(0x0a04, PEEK(0x0a04)&254);

Otherwise the sprite will appear for a fraction of a second and then disappear.

Discussion from Lemon forums: [https://www.lemon64.com/forum/viewtopic.php?p=1055429](https://www.lemon64.com/forum/viewtopic.php?p=1055429)

Quote from Lemon64 user Rosettif:
```
"Disable using the shadow registers by storing $ff to $d8 (or POKE 216,255).

And/or disable the Basic part of the IRQ handler by clearing bit $01 at $0a04, e.g. by POKE 2564,PEEK(2564) AND 254.
That generally contains $c1 by default, so you may just POKE 2564,192.
(The latter is also valid for the C65/MEGA65, but the corresponding address is $1104 there instead.)"
```

Note that POKE 216,255 did **not** resolve this issue for me however 'asm("sei");' does and so does 'POKE(0x0a04, PEEK(0x0a04)&254);'

Could one of these two fixes be added to the startup code for the C128?

mrdudz:
## Summary

The C128 BASIC does a bunch of things at vblank: it copies shadow registers to VIC registers, it moves sprites and plays music. All of this is not needed (and gets in the way) when our code runs, so it is disabled before main() and restored after.

It looks like the C65 and Mega65 libs will need the same fix, will look at that next and sneak into this PR before merge :)

## Checklist

### DIFF
--- a/libsrc/c128/crt0.s
+++ b/libsrc/c128/crt0.s
@@ -65,6 +65,15 @@ L1:     lda     c_sp,x
 
         jsr     initlib
 
+; Disable the BASIC part of the IRQ handler. It would usually (once per frame)
+; copy the VIC shadow register, move sprites, play music. This would only get
+; in the way, so we turn it off.
+
+        lda     INIT_STATUS
+        sta     initsave
+        and     #$fe
+        sta     INIT_STATUS
+
 ; Set the bank for the file name to our execution bank. We must do this
 ; *after* calling the constructors because some of them might depend on
 ; the original value of this register.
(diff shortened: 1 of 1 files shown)


mrdudz:
> It looks like the C65 and Mega65 libs will need the same fix, will look at that next and sneak into this PR before merge :)

Couldnt get that to work... in Mega65 mode the sprites appear without that fix (weird). In C65 mode i cant make stuff work at all, i am probably using xemu the wrong way, no idea.

So this is ready

PR #2769, opened 2025-07-03
Add strndup

colinleroy:
## Summary

This adds a (naive) strndup implementation. I have favored size over speed, and have decided not to factorize strdup() as it was making it much bigger even for people not using strndup.

As a bonus, I added str(n)len tests as they're used in str(n)dup tests.
## Checklist

### DIFF
--- a/doc/funcref.sgml
+++ b/doc/funcref.sgml
@@ -785,6 +785,7 @@ communication, see also <tt>testcode/lib/ser-test.c</tt>.
 <item><ref id="strcpy" name="strcpy">
 <item><ref id="strcspn" name="strcspn">
 <item><ref id="strdup" name="strdup">
+<item><ref id="strndup" name="strndup">
 <item><ref id="strerror" name="strerror">
 <item><ref id="stricmp" name="stricmp">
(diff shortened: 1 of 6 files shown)


kugelfuhr:
The behavior of `strndup` does not seem to match that from the `strndup` man page which describes the POSIX version. Implementing a POSIX function with differing semantics might not be a good idea.

 Maybe your function should be named `strndupa` which is the GNU name for this function.

colinleroy:
Hi @kugelfuhr, 
What's different from the posix manpage in the end result? Or did you mean that because of the full copy/cut/realloc?

kugelfuhr:
The POSIX function `strndup` does not allocate a new string on the heap, it just copies. There's another function in glibc (not POSIX) named `strndupa` that copies into an allocated memory area and returns a pointer to it.

See here: https://linux.die.net/man/3/strndup

kugelfuhr:
OOPS - sorry, got that wrong. The function is correct, ignore my comments 😀

mrdudz:
No complaints, thumbs up for the added tests. Thanks!

PR #2770, opened 2025-07-04
Add comments about cputs() relying on cputc() not to clobber ptr1

### LINKED ISSUE #1557: cputs depends on cputc behavior not guaranteed by calling convention
In order to bootstrap a library for my SBC, I wrote some of the conio functions in C.

This led to failure because the implementation of `cputs` in libsrc/conio depends on `cputs` preserving the zero page variables `ptr1` and `tmp1`, which C functions aren't guaranteed to do. As the documentation states:

"Many of the internal pseudo-registers used by cc65 are available for free use by any function called by C, and do not need to be preserved. Note that if another C function is called from your assembly function, it may clobber any of these itself."

In my case, my `cputs` implementation performs a multiplication, which clobbers both `ptr1` and `tmp1`.

I don't know what the right solution is, because I don't know if `cputs` and `cputc` were deliberately designed this way. If so, probably documenting the expectations for `cputc` are probably sufficient (i.e, it has to preserve `ptr1` and `tmp1` and therefore can't be written in C).

kugelfuhr:
## Summary

The implementation of `cputs()` calls `cputc()` repeatedly and relies on it not to clobber `ptr1`. This results in shorter code but might be problematic if someone changes the code or creates a `cputc()` function for a new target.

As discusses in #1557 this PR adds comments to the relevant sources describing the dependency.

Fixes #1557.

## Checklist

### DIFF
--- a/libsrc/agat/cputc.s
+++ b/libsrc/agat/cputc.s
@@ -5,6 +5,8 @@
 ; void __fastcall__ cputcxy (unsigned char x, unsigned char y, char c);
 ; void __fastcall__ cputc (char c);
 ;
+; Important note: The implementation of cputs() relies on the cputc() function
+; not clobbering ptr1. Beware when rewriting or changing this function!
 
     .import     COUT
     .export     _cputcxy, _cputc, newline, putchar,putchardirect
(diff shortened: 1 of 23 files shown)


mrdudz:
Thanks!

PR #2771, opened 2025-07-05
Make strndup smaller, safer, faster

colinleroy:
First implementation was doing:
- strdup (without checking result!)
- strlen
- terminate
- realloc

New one does:
- malloc to shortest +1
- strncpy
- terminate

In other words, the initial implementation was dangerous as it didn't check that strdup() succeeded; and pulled realloc. It pulled 1306 bytes into an otherwise empty program (from strdup, malloc, strlen, realloc), while the new one pulls in 537 bytes (from malloc, strncpy).

## Summary

Provide description of the bug (if an issue exists, link to it) and the fix.

## Checklist

- [ ] New unit tests have been added to prevent future regressions (not applicable)
- [ ] The documentation has been updated if necessary (not applicable)

### DIFF
--- a/libsrc/common/strndup.s
+++ b/libsrc/common/strndup.s
@@ -4,53 +4,50 @@
 ; char* __fastcall__ strndup (const char* S, size_t maxlen);
 ;
 
-        .importzp       ptr4, c_sp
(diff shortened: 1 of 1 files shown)


mrdudz:
looks fine to me, thanks!

PR #2776, opened 2025-07-06
Further optimize inc/decsp

### LINKED ISSUE #2748: jsr incsp2 jmp incsp2 optimization
i see some generated code do this
<pre>
	jsr     incsp2
	jmp     incsp2
</pre>

isn't that the same as
<pre>
	jmp     incsp4
</pre>

not sure if this hints at a broader class of optimizations

colinleroy:
## Summary

The optimizer can now merge incsp+decsp together, and merge jsr incsp/decsp + jmp incsp/decsp together. Drop the code altogether if both instructions cancel each other out.

Fixes #2748

Passes the test suite and all my projects still work :)

## Checklist

- [ ] New unit tests have been added to prevent future regressions (N/A)
- [ ] The documentation has been updated if necessary (N/A)

### DIFF
--- a/src/cc65/codeopt.c
+++ b/src/cc65/codeopt.c
@@ -907,6 +907,11 @@ static unsigned RunOptGroup7 (CodeSeg* S)
        between branches */
     C += RunOptFunc (S, &DOptBranchDist, 3);
 
+    /* Re-optimize inc/decsp that may now be grouped */
+    C += RunOptFunc (S, &DOptStackPtrOps, 5);
+    /* Re-optimize JSR/RTS that may now be grouped */
+    C += RunOptFunc (S, &DOptRTS, 1);
+
     Changes += C;
     /* If we had changes, we must run dead code elimination again,
     ** since the changes may have introduced dead code.
--- a/src/cc65/coptmisc.c
+++ b/src/cc65/coptmisc.c
@@ -380,28 +380,37 @@ unsigned OptIndLoads2 (CodeSeg* S)
 
 
 
-static unsigned IsDecSP (const CodeEntry* E)
-/* Check if this is an insn that decrements the stack pointer. If so, return
-** the decrement. If not, return zero.
-** The function expects E to be a subroutine call.
+static signed IsSPShift (const CodeEntry* E)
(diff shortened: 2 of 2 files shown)


mrdudz:
Nice one!

It would be cool to have a test for this, something with a couple functions which would typically produce code that is to be optimized like this... the simple case from #2748 would be a start - something more complex (that is more likely to fail) would be great :)

colinleroy:
@mrdudz I've never written tests that check the compiler output, can you point me to an example? Or did you mean to make a runtime test?

mrdudz:
A runtime test.... checking the compiler output isn't practical i think, that would break all the time

colinleroy:
Done :)

mrdudz:
A runtime test - making one that checks the compiler output (as in the generated assembler file) is not very practical i think.

I was thinking about some code (a few variation of it if possible) that generates those inc/decsp pairs without optimization, with some usage of local variables (so stack is used)

colinleroy:
> A runtime test - making one that checks the compiler output (as in the generated assembler file) is not very practical i think.
> 
> I was thinking about some code (a few variation of it if possible) that generates those inc/decsp pairs without optimization, with some usage of local variables (so stack is used)

I'm not sure if you missed the test I added in https://github.com/cc65/cc65/pull/2776/files#diff-b80bb4d4254531090a71bba8a45a0588e97ed279e3240b14d83cc9610edb009c . If you didn't, can you tell me what you expected? 

The advantage of the test suite being run with different -O options is that this change is tested by a lot of the unit tests, as you can see from the attached diff of the testwrk/*/*s files (generated from make test on master and this branch, keeping the .s files) - there are a lot of .dbg diffs in there but search for 'incsp' and you'll see the actual effect.
[incsp.diff.txt](https://github.com/user-attachments/files/21136678/incsp.diff.txt)

(these are the ones I love the most :blush: :
```
-	jsr     decsp2
-	jmp     incsp2
+	rts
```
)

mrdudz:
Ah, yeah didn't see that one, exactly what i meant. cheers :)

mrdudz:
merging, cheers!

PR #2777, opened 2025-07-06
Improve error handling for ca65

### LINKED ISSUE #1656: ca65: Strange error message for invalid code
This code:

```
        .repeat         1, Arg
        Arg:
        .endrepeat
```

Gives:

```
test.s:3: Error: Unexpected trailing garbage characters
test.s:2: Note: Expanded from here
```

But there is no problem in line 3. The wrong error line is even worse if there is more stuff within the repeat loop. The error message does always point to the `.endrepeat` while the real problem is the label in line 2.

kugelfuhr:
## Summary

This PR contains several improvements for ca65 error handling.

### 1.
Instead of `Syntax error` or similar, the assembler will now in most cases tell, what was expected and which token was seen instead.

As an example, for the following input
```
        lda     $00 foo
```
 instead of `Unexpected trailing garbage characters` you will now get `Expected 'end-of-line' but found 'foo'`.

As a consequence, `Syntax error` and `Unexpected trailing garbage characters` are completely gone.

### 2.
Error recovery has been improved in several places to avoid followup errors caused by some earlier problem.

### 3.
The PR introduces the capability to output notifications in addition to error messages and uses this to output helpful additional diagnostics in case of errors.

As an example, for the following input
```
.macro  foo     a1, a2
        .byte   a1, a2
.endmacro
foo     1, 2, 3
```
instead of
```
test.s:4: Error: Too many macro parameters
```
you will now get
```
test.s:4: Error: Too many parameters for macro 'foo'
test.s:1: Note: See definition of macro 'foo' which was here
```

Please see the added test programs for more examples.

### 4.
When skipping a macro, it applies the same handling for `.endmacro` as when parsing it: A `.endmacro` terminates the macro only if it is the first token on the line.

### 5.
It improves error handling in macros. If there was an error in a macro definition, you will now get something like
```
test.s:4: Error: Macro 'xyz' contains errors and cannot be expanded
test.s:1: Note: Definition of macro 'xyz' was here
```
This change might fix #1656, #392 and maybe others and might make #2068 obsolete.

## Checklist


kugelfuhr:
@mrdudz Sorry I had to redo your change because of git problems.

mrdudz:
great stuff, thanks!

PR #2778, opened 2025-07-06
Optimise ldaxi

### LINKED ISSUE #2055: cc65: Regression in optimized code after v2.19
I noticed a regression in the quality of optimized code sometime after version 2.19 of cc65.  I created a sample of the issue on godbolt:
https://godbolt.org/z/bqoKP4PWx

Sample code:
        struct Object
        {
            int a;
            int data[10];
        };
        
        int get_elem(struct Object* o) {
            return o->data[2];
        }

On "trunk" the following code is generated when passing -O:
.proc   _get_elem: near

        jsr     pushax
        jsr     ldax0sp
        jsr     incax2
        jsr     incax4
        jsr     ldaxi
        jmp     incsp2

On older versions of cc65 (e.g. 2.19) smaller/faster code is generated:
.proc   _get_elem: near

        jsr     pushax
        jsr     ldax0sp
        ldy     #$07
        jsr     ldaxidx
        jmp     incsp2

I did a "git bisect" to find the commit that introduced this regression, and found that 99c7fe0ada96dd7283e60cc16cc63b23ab919014 is what regressed the optimization:
commit 99c7fe0ada96dd7283e60cc16cc63b23ab919014
Author: acqn <acqn163@outlook.com>
Date:   Mon Feb 8 09:03:23 2021 +0800

    Reusing code from parseadd() for ArrayRef().
    Now index[ptr] works in constant expressions.
    Fixed codegen for cast type in constant expressions.
    Calls on swapstk in 'i[ptr]' is avoided when possible.

I thought I would report this issue here in the hope that the optimizer can be improved to avoid this regression.

colinleroy:
## Summary

    Optimize multiple incax* and incax*/ldaxi
    
    - Group multiple calls to incax* and decax* into a single one
    - Replace incaxN/jsr ldaxi with ldy #N+1/jsr ldaxidx
    
    Fixes #2055

Pushing this one now, but it depends on #2776 as it generalizes the incsp/decsp optimisation.

## Checklist

- [ ] The documentation has been updated if necessary (N/A)

### DIFF
--- a/src/cc65/codeopt.c
+++ b/src/cc65/codeopt.c
@@ -113,6 +113,8 @@ static OptFunc DOptAdd3         = { OptAdd3,         "OptAdd3",          65, 0, 
 static OptFunc DOptAdd4         = { OptAdd4,         "OptAdd4",          90, 0, 0, 0, 0, 0 };
 static OptFunc DOptAdd5         = { OptAdd5,         "OptAdd5",         100, 0, 0, 0, 0, 0 };
 static OptFunc DOptAdd6         = { OptAdd6,         "OptAdd6",          40, 0, 0, 0, 0, 0 };
+static OptFunc DOptAXLoad       = { OptAXLoad,       "OptAXLoad",        50, 0, 0, 0, 0, 0 };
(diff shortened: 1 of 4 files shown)


mrdudz:
thanks!

PR #2782, opened 2025-07-07
Use "static inline" instead of macros in the source

kugelfuhr:
## Summary

This PR removes the inline.h module and changes all places where separate inline functions and macros where supplied to always use inline functions.

It seems that the reason for using macros was that there were compilers not supporting `static inline` but this is no longer the case.

The change makes the code smaller (by about 1500 lines) with no loss of functionality and it improves readability and type safety.

## Checklist


mrdudz:
Yay, i was thinking of doing exactly this the other day :) Will skim over it and check for accidents later tonight.

mrdudz:
Looks fine to me, thanks!

PR #2783, opened 2025-07-07
Add info about cheap local names in macro local symbols.

### LINKED ISSUE #2656: local macro labels break cheap scopes
Here are my code snippets:

```a65
@wl:
    dey
    beq @end
    lda (r0),y
    ; check if ascii:
    cmp #$7b
    bcs @wl
    cmp #$61
    bcc @wl
    sbc #$20
    sta (r0), y
    bra @wl

@fail:
    print failtext
@end:
```
(`@fail` label is referenced earlier in the code)

```a65
.macro print src
    .local end
    .local leup
    ldx #0
leup:
    lda src, X
    beq end
    jsr cout
    inx
    bne leup
end:
.endmacro
```

The local labels in the macro end the cheap scope and because of that `@end` is not visible to the code above. Local labels by definition are supposed not to disrupt the flow of the code where the macro is inserted. 

I guess an easy solution to that problem could be to simply define the generated random replacement labels as a cheap locals. Though I'm not a compiler dev, so I may not be aware of cases where doing that could be undesirable.

Probably related to #938 

EDIT: It seems like if you define the local macro labels as cheap (call them `@leup` and `@end` instead of the names without the `@`), their "cheapness" is then inherited by the random unique labels generated by the assembler. Therefore the issue may be actually solved, but I guess documenting that behaviour would be good (and also making sure that no regression happens regarding this feature)

kugelfuhr:
## Summary

Add a paragraph about using cheap local names in `.local`. This fixes #2656.

## Checklist

### DIFF
--- a/doc/ca65.sgml
+++ b/doc/ca65.sgml
@@ -918,6 +918,7 @@ See the description of the <tt><ref id=".PROC" name=".PROC"></tt>
 directive for more information.
 
 
+<label id=cheap-locals>
 <sect1>Cheap local labels<p>
 
 Cheap local labels are defined like standard labels, but the name of the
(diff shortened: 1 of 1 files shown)


mrdudz:
thanks!

PR #2785, opened 2025-07-08
Modernize ca65 diagnostics

kugelfuhr:
## Summary

This PR tries to brush up ca65 diagnostics and gives them a fresh and modern look.

It consists of two parts, which are tightly coupled.

### Colored diagnostics

When a color terminal is detected, ca65 will now colorize diagnostics similar to what some compilers do.

Color output can be switched from the command line. The default is automatic detection which means that colors are enabled if the output goes to a console that supports color. This works for Windows (with some more effort as usual) and
Linux. It should also work for MacOS but I wasn't able to test that.

### UTF-8 output

When an UTF-8 capable console is detected, ca65 uses UTF-8 characters for its output which will allow typographically correct quoting and some other improvements in the future. UTF-8 output can be disabled by a command line option.

### Examples

This is an example for an error with a notification:
colortest5

Warnings look like this:
colortest2

Fatal errors are also colored:
colortest4

And even user errors and warnings might use coloring for quoted text:
colortest3

### Consequences

For typographically correct quotes to work developers will have to embed quoted text in \`single quotes'. To be compatible with the old behavior, the opening quote is replaced by ' if UTF-8 is unavailable. For the tests, it might also be advisable to use the `--no-utf8` option when piping output to a file to avoid random test failures when non UTF-8 terminals are used.

### Caveats

This PR is based on #2777 and must be merged after it.

## Checklist


mrdudz:
Mmmh it must be git acting up again with those conflicts.... @kugelfuhr please have a look (merging master locally and pushing again should do the trick, i guess)

mrdudz:
Nice, i have to try this (will you do this for the compiler as well? :))

Thanks!

kugelfuhr:
> Nice, i have to try this (will you do this for the compiler as well? :))

I have prepared similar changes for the linker. And no plans currently beyond that.

PR #2794, opened 2025-07-12
Fixgrc

### LINKED ISSUE #2227: Zero length samples-geos.d64
What's wrong?
```
$ make samples-geos
c1541 -format "samples-geos,01" d64 samples-geos.d64
formatting in unit 8 ...
c1541 -attach samples-geos.d64 -geoswrite bitmap-demo.cvt; c1541 -attach samples-geos.d64 -geoswrite filesel.cvt; c1541 -attach samples-geos.d64 -geoswrite geosver.cvt; c1541 -attach samples-geos.d64 -geoswrite getid.cvt; c1541 -attach samples-geos.d64 -geoswrite hello1.cvt; c1541 -attach samples-geos.d64 -geoswrite hello2.cvt; c1541 -attach samples-geos.d64 -geoswrite overlay-demo.cvt; c1541 -attach samples-geos.d64 -geoswrite vector-demo.cvt; c1541 -attach samples-geos.d64 -geoswrite yesno.cvt;
$ c1541 samples-geos.d64
c1541 (VICE 3.7.1)
Copyright 1995-2022 The VICE Development Team.
C1541 is free software, covered by the GNU General Public License, and you are
welcome to change it and/or distribute copies of it under certain conditions.
c1541 #8> dir
0 "samples-geos    " 01 2a
0    "bITMAP dEMO"      usr
0    "FILESEL"          usr
0    "GEOSVER"          usr
0    "GETID"            usr
0    "HELLO1"           usr
0    "HELLO2"           usr
0    "oVERLAY dEMO"     usr
0    "VECTORDEMO"       usr
0    "YESNO"            usr
628 blocks free.
c1541 #8>

### PR DESCRIPTION
**Pull request by mrdudz**
## Summary

Fixes #2227 (hopefully)

- the block size in the directory entry is now calculated from the VLIR size. i do NOT (yet) know if this is actually correct as it is, need to dig up some more docs (or some other way to verify this).
- if a barebones geos template .d64 exists as geos-template.d64, the samples will be copied to that one instead of creating an empty file (that makes testing them less hassle)

Did some quick checks and the samples worked, you can grab a geos-template.d64 from [here](https://github.com/mrdudz/portris/raw/refs/heads/master/geos-template.d64)

## Checklist

### CODE REVIEW

**Comment by mrdudz** [comment]
ok so what i did isn't correct. after "validate" (disk menu on the desktop) the dir shows values that are larger than before.

**Comment by mrdudz** [comment]
Mmmh now i tried
```
"\t.word (((__VLIR0_LAST__ - __VLIR0_START__ - __BSS_SIZE__) + 253) / 254) + (((__VLIR1_LAST__ - __VLIR1_START__) + 253) / 254) + (((__VLIR2_LAST__ - __VLIR2_START__) + 253) / 254) + (((__VLIR3_LAST__ - __VLIR3_START__) + 253) / 254) + (((__VLIR4_LAST__ - __VLIR4_START__) + 253) / 254) + (((__VLIR5_LAST__ - __VLIR5_START__) + 253) / 254) + (((__VLIR6_LAST__ - __VLIR6_START__) + 253) / 254) + (((__VLIR7_LAST__ - __VLIR7_START__) + 253) / 254) + (((__VLIR8_LAST__ - __VLIR8_START__) + 253) / 254) + (((__VLIR9_LAST__ - __VLIR9_START__) + 253) / 254)\n"   /* length in blocks */
```
.... but its STILL not right. most files are one block off (that i could accept, a rounding error somewhere) - but the overlay example is still TWO blocks off, which makes no sense *shrug*

File: src/grc65/main.c [hunk 1]
```diff
@@ -607,15 +610,14 @@ static void DoHeader (void)
             "\t.byte %i\n"
             "\t.byte %i\n"
             "\t.byte %i, %i, %i, %i, %i\n\n"
-            "\t.word 0\n"
+            "\t.word ((__VLIR0_LAST__ - __VLIR0_START__ - __BSS_SIZE__) + 253) / 254\n"   /* length in blocks */
```
polluks, line 613: 253 and 254 as a result of the two link bytes? Please use geoConvert converted files as reference.
By the way next file(1) will support CVT :-)

mrdudz: There is no way to use geoConvert files as a reference - unless you can provide a pair of those and the corresponding ca65 source to produce them.

What i am doing instead is writing them to disk, and then "validate" in the GEOS menu - if *that* doesn't produce the correct block numbers, then everything is broken beyond repair (in GEOS)

And yes of course, it is rounding up to the next full block. But i must be missing something in the calculation

polluks2: Indeed, we have no source of Desktop to see validate's implementation.

ytmytm:
> Mmmh now i tried
> 
> ```
> "\t.word (((__VLIR0_LAST__ - __VLIR0_START__ - __BSS_SIZE__) + 253) / 254) + (((__VLIR1_LAST__ - __VLIR1_START__) + 253) / 254) + (((__VLIR2_LAST__ - __VLIR2_START__) + 253) / 254) + (((__VLIR3_LAST__ - __VLIR3_START__) + 253) / 254) + (((__VLIR4_LAST__ - __VLIR4_START__) + 253) / 254) + (((__VLIR5_LAST__ - __VLIR5_START__) + 253) / 254) + (((__VLIR6_LAST__ - __VLIR6_START__) + 253) / 254) + (((__VLIR7_LAST__ - __VLIR7_START__) + 253) / 254) + (((__VLIR8_LAST__ - __VLIR8_START__) + 253) / 254) + (((__VLIR9_LAST__ - __VLIR9_START__) + 253) / 254)\n"   /* length in blocks */
> ```
> 
> .... but its STILL not right. most files are one block off (that i could accept, a rounding error somewhere) - but the overlay example is still TWO blocks off, which makes no sense _shrug_

Two blocks off makes sense to me :). Assuming that this correctly calculates the length of segments (records) in blocks, you need to add two more blocks to the file length: one for info block (icon) and another one for the VLIR record table.
For a SEQ structure file it would be always one block more - the info block.

mrdudz:
> Two blocks off makes sense to me :). Assuming that this correctly calculates the length of segments (records) in blocks, you need to add two more blocks to the file length: one for info block (icon) and another one for the VLIR record table.
For a SEQ structure file it would be always one block more - the info block.

I think i found the problem: the records table is not there when only one VLIR segment exists. Is that even legal? (Why does it still work in GEOS?)

mrdudz:
The way grc65 is implemented makes it a bit hard to fix this easily.... args

mrdudz:
Ha. Was easy after all. The problem was that files with only one VLIR record are created as "sequential" GEOS files, but they were not correctly marked as such.

It should work correctly now. The only thing that bothers me a bit is the ugly long line that does the calculation now.... not sure if/how this can be made a bit nicer

polluks:
How about a macro with token-pasting?

mrdudz:
Example? I only very rarely use macros, and never use any of those advanced things :)

mrdudz:
OK this is slightly less ugly.

If noone objects i'll merge this

polluks:
Some brackets are unnecessary...

mrdudz:
ok merging :)

PR #2795, opened 2025-07-13
Fix an error in the sort check scripts.

kugelfuhr:
## Summary

Fix an error that was introduced by me in a4a24280f26ddb3e319678db7a531320df7125fe: 

Using a pipe causes a subshell to be generated so the "exit" statements will just leave the subshell. As a consequence, the sortedxxx.sh scripts produced output but no error exit code. Which in turn caused an error in #2778 to slip through undetected.

## Checklist

### DIFF
--- a/.github/checks/sorted.sh
+++ b/.github/checks/sorted.sh
@@ -33,7 +33,8 @@ function checkarray_quoted_name
 }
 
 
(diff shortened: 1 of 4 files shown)


mrdudz:
Thanks!

Issue #2796, opened 2025-07-13
Tests emit data into BSS segment


kugelfuhr:
I'm currently trying to have assembler and linker abort on any warnings. One of them is that the cpudetect tests emit all data into BSS. This causes a linker warning which is now an error. I'm not sure why this is done, maybe someone can clarify. The comment is:

>    ; just emit to bss for now, this way the output wont get checked, which is
>    ; fine - we only want to see if the assembler accepts all the opcodes it
>    ; should accept for the current CPU mode 

I have to admit that I don't fully understand that. Wouldn't it be possible to simply use `.code` here?

Anyway, I think this should be fixed. There shouldn't be any warnings in the tests.

mrdudz:
I knew you would complain about this one when i read you want to kill all linker warnings in the tests :)

> Wouldn't it be possible to simply use .code here?

That test also produces other output, which IS compared against a binary reference. So either that tests need to be splitted into two individual tests, or (probably the better idea) the extra output must be added to the reference file(s).

kugelfuhr:
I had a closer look in the meantime and it seems using BSS is just a way to dump all the output to /dev/null. Which brings up the question why the data is not checked? It seems to me that this won't cause any problems. It is not supposed to change and it gives an additional guarantee that all the opcodes generate correct output.

mrdudz:
closed via #2797

PR #2797, opened 2025-07-13
For the cpudetect tests, check the binary results for all instructions

kugelfuhr:
## Summary

See #2796. Previously the cpudetect tests trashed part of the generated data meaning that most of the instructions weren't actually checked. The change in this PR causes all data to be output to the binaries and replaces the reference binaries with ones that contain the data for all instructions.

Note: I haven't checked the binaries for correctness and assumed that the output generated by the assembler is correct and used it as new reference data. 

## Checklist


mrdudz:
> I haven't checked the binaries for correctness and assumed that the output generated by the assembler is correct and used it as new reference data

They should be correct - since there is another (very similar) test in test/asm/opcodes which checks just that

Thanks!

PR #2798, opened 2025-07-14
Make the INIT segment for creativision r/w

### LINKED ISSUE #2787: Creativision samples generates linker warnings
When compiling one of the samples (ascii.c, hello.c) for creativision with current master I get `Segment 'INIT' with type 'bss' contains initialized data`.

The problem is probably somewhere in the library so I assume most programs will have the warning.

kugelfuhr:
## Summary

As described in #2787 data is placed into this segment so it cannot be of type `bss`.

Fixes #2787.

## Checklist


groessler:
Why didn't you change the contructors to not use INIT like the other configs? This would to me be the obvious fix.

kugelfuhr:
> Why didn't you change the contructors to not use INIT like the other configs? This would to me be the obvious fix.

Well, maybe because it wasn't so obvious for me 😉 

So is changing the segment for the constructors the desired solution?

groessler:
Yes. Segment "ONCE".

kugelfuhr:
Ok. So be it then.

groessler:
Thanks!

PR #2799, opened 2025-07-14
Atari 7800: Make the EXTZP segment optional

kugelfuhr:
## Summary

The `EXTZP` segment is not required but not being marked as `optional`. That causes warnings when compiling the samples (and probably user programs).

## Checklist


mrdudz:
thanks!

PR #2802, opened 2025-07-14
Modernize ld65 diagnostics

kugelfuhr:
## Summary

This PR adds colors and UTF-8 output to ld65 similar to what #2785 did for ca65. It does also add notifications to diagnostic output and in a few cases uses these notifications to add helpful information in case of warnings or errors.

In addition to all that, it adds the `--warnings-as-errors` option to cl65 and cc65. cc65 does already have `-W error` but `--warnings-as-errors` was added for compatibility with the other tools. cl65 passes `--warnings-as-errors` to ca65, cc65 and ld65.

## Checklist


mrdudz:
Looks good to me, thanks!

PR #2803, opened 2025-07-14
Add a description for --memory-model to the docs

### LINKED ISSUE #1162: [cc65] Address size qualifiers?
There are "address size" qualifiers for code/data pointers in cc65, but for now only `__near__` and `__far__` are supported, one of which will be automatically added to pointers at definitions according to the memory model if neither is explicitly specified. The question is, what is the consequence if a pointer does have no such qualifiers at all? This can easily happen as an array is converted to a pointer, which could also expose some hidden type comparison issues .

Example:
```
void f(int*);   /* the parameter will have '__near__' by default */
void f(int []); /* the parameter will have no qualifiers at all when converted to a 'int *' pointer */
```
The two function prototypes are considered mismatched even if the array parameter is converted to pointer due to the first having an `__near__` qualifier on the parameter while the second having none after conversion.

kugelfuhr:
## Summary

Implement the `-mm` option for cc65 as described in the usage information and add a description for `-mm`/`--memory-model` to the docs.

Fixes #1162.

## Checklist

### DIFF
--- a/doc/cc65.sgml
+++ b/doc/cc65.sgml
@@ -514,6 +514,28 @@ Here is a description of all the command line options:
   name="#pragma&nbsp;local-strings"></tt> for fine grained control.
 
 
+  <tag><tt>-mm model, --memory-model model</tt></tag>
+
+  This option sets the code and data models for the compiler to use. Please
+  note that memory models are an unfinished feature and using this option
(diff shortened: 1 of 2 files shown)


mrdudz:
Thanks!

PR #2805, opened 2025-07-16
Remove .s files generated from .grc files

kugelfuhr:
## Summary

Fixes a problem introduced with #2804: The .s file generated from the added .grc file wasn't deleted by "make clean".

## Checklist

### DIFF
--- a/samples/Makefile
+++ b/samples/Makefile
@@ -572,8 +572,7 @@ endif
 # GEOS programs need a resource file linked to them
 ifeq ($(SYS),geos-cbm)
 asciires.o: asciires.grc
-	$(GRC) -t $(SYS) $<
-	$(AS) $(<:.grc=.s)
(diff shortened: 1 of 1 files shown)


kugelfuhr:
In the meantime I realized that this change is ok, but there are more problems. Doing

```
make -sj clean
make -sj
make -sj check
```

fails after #2804 since `asciires.s` (generated from `asciires.grc`) contains tabs. The check script ignores just the "wrk" directories, but `asciires.s` is generated in `samples/`.

Maybe there's a fix that solves this, too?

mrdudz:
compile the .grc into an object file directly using cl65 perhaps?

kugelfuhr:
Ok. Using cl65 now. That should fix both problems.

mrdudz:
Thanks!

PR #2808, opened 2025-07-17
Simple flow analysis to find unreachable code

kugelfuhr:
## Summary

This PR adds a simple control flow analysis to the C compiler. It tracks all statements with the exception of `switch` (a really strange beast in C) to find occurrences of unreachable code. It has some shortcomings (see below) but works quite well.

Here is an example:

```
 1:    do {
 2:        do {
 3:            if (a == 2) {
 4:                return a;
 5:            } else {
 6:                continue;
 7:            }
 8:        } while (0);
 9:    } while (0);
10:    a = 2;
11:    return a;
```

The compiler warns correctly that the code in line 10 is unreachable. For more examples please see the added test programs.

There are a few cases where the compiler does *not* warn when it should, but I haven't found a case where it warns without a reason (but see "Shortcomings" below). It is probably possible to fool the compiler but so far I haven't found a way to do that.

## Shortcomings

The nature of the compiler leads to a few limitations. Since there is no intermediate code, the compiler cannot easily distinguish between statements that generate code and others that do not. That has a few consequences:

### Double semicolon

A double semicolon after a statement that makes the following code unreachable leads to a `unreachable code` warning as in the following example:

```
    return;;
```

### Compound statement

If a condition is followed by a compound statement, the compiler would point to the start of the compound statement (the opening curly bracket) for the warning. I have added code to check for that and point to the next token in case of a curly bracket, but this fails if yet another one follows. So in this example:

```
 1: if (0) {
 2:     {
 3:         a = 3;
 4:     }
 5: } else {
```

the compiler will warn at line 2 instead of line 3 (which generates code).

## Extras

If the end of a function is unreachable as in this example:

```
    int f(void)
    {
        ...
        while (1) { }
    }
```

the warning about a missing return is suppressed.

An `unreachable code` warning can be suppressed by an empty compound statement.

## Outlook

Based on this work it should be possible to

* add flow control analysis for switch statements, and
* add tracking for return statements, and
* track variable usage (needs much more work).

And someone will now have to fix all the `unreachable code` warnings in the samples.

## Checklist

### DIFF
--- a/libsrc/common/getopt.c
+++ b/libsrc/common/getopt.c
@@ -29,7 +29,7 @@ int optopt;                     /* character checked for validity */
 char *optarg;                   /* argument associated with option */
 
 #define tell(s) fputs(*argv,stderr);fputs(s,stderr); \
-                fputc(optopt,stderr);fputc('\n',stderr);return(BADCH);
+                fputc(optopt,stderr);fputc('\n',stderr);return(BADCH)
 
 int __fastcall__ getopt (int argc, char* const* argv, const char* optstring)
 /* Get option letter from argument vector */
--- a/src/cc65/error.c
+++ b/src/cc65/error.c
@@ -425,31 +425,31 @@ void UnreachableCodeWarning (void)
 {
     if (IS_Get (&WarnUnreachableCode)) {
 
-	LineInfo* LI;		  
+        LineInfo* LI;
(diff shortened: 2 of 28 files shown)


mrdudz:
Nice one! Thanks!

Issue #2806, opened 2025-07-16
Lost ability to create optimized code for w65c02 cpu


JimmyDansbo:
In commit c72126e68fc879229b677baf6443ed242b0262f7 .cap is being used instead of .cpu to check for the cpu type that is being build for.

I am not sure how many places this has broken correct detection of the w65c02 cpu, but at least in popa.s for the Commander X16 target, it no longer knows that the cpu can do direct zeropage addressing which makes the function use the Y register initialized to 0.

Before above commit, the code in popa.s was:
`.if (.cpu .bitand ::CPU_ISET_65SC02)`
which correctly detects that the cx16 targt platform should compile for w65c02
After the commit, the code is:
`.if .cap(CPU_HAS_ZPIND)`
This does not execute the code for the w65c02, but instead executes the code for the 6502.

mrdudz:
This mess was doomed to trigger problems like this :/

kugelfuhr:
> This mess was doomed to trigger problems like this :/

Which is why I strongly opposed this change. It breaks old code and in addition makes `.cpu` almost useless.

Anyway, looking at the problem with `.cap` it seems that for the w65c02 the `CPU_HAS_ZPIND` capability is not active. If the w65c02 CPU definitely has the zeropage indirect addressing mode, this has to be added and will solve the problem.

While we're at it, anyone is invited to look at the table in `common/cpu.c` and check the capabilities for the different cpus. There may be more problems like this.

mrdudz:
closed by #2809

PR #2809, opened 2025-07-17
Added correct capabilities to W65C02 and 65CE02 CPUs

JimmyDansbo:
In response to https://github.com/cc65/cc65/issues/2806
Added capabilities to the W65C02 and 65CE02 CPUs. To the best of my knowledge the W65C02 CPU is now complete, but the 65CE02 have plenty of capabilities that are not listed in common/capability.c


mrdudz:
Thanks!

PR #2811, opened 2025-07-20
Optimize jsr ldax0sp/incsp2

colinleroy:
## Summary

A lot of functions end with jsr ldax0sp / jmp incsp2, returning AX from top of stack and popping it. jmp popax is shorter and quicker.

The attached diff of testwrk show that it will be nice to add a merging of jsr pushax/jmp popax in a second time.
[OptTosLoadPop.diff.txt](https://github.com/user-attachments/files/21334183/OptTosLoadPop.diff.txt)

Second commit adds it, the new diff shows more useless code removal
[OptPushPop.diff.txt](https://github.com/user-attachments/files/21334430/OptPushPop.diff.txt)

## Checklist

### DIFF
--- a/src/cc65/coptmisc.c
+++ b/src/cc65/coptmisc.c
@@ -486,14 +486,14 @@ static unsigned OptIncDecOps (CodeSeg* S, const char* dec, const char* inc, cons
 
 unsigned OptStackPtrOps (CodeSeg* S)
 {
-    return OptIncDecOps(S, "decsp", "incsp", "subysp", "addysp");
+    return OptIncDecOps (S, "decsp", "incsp", "subysp", "addysp");
 }
 
 
 
 unsigned OptAXOps (CodeSeg* S)
 {
-    return OptIncDecOps(S, "decax", "incax", "decaxy", "incaxy");
+    return OptIncDecOps (S, "decax", "incax", "decaxy", "incaxy");
 }
 
 
@@ -573,11 +573,9 @@ unsigned OptTosLoadPop (CodeSeg* S)
         const CodeEntry* E = CS_GetEntry (S, I);
 
         /* Check for the sequence */
-        if (E->OPC == OP65_JSR                            &&
-            strncmp (E->Arg, "ldax0sp", 5) == 0           &&
-            (N = CS_GetNextEntry (S, I)) != 0             &&
-            (N->OPC == OP65_JSR || N->OPC == OP65_JMP)    &&
-            strcmp (N->Arg, "incsp2") == 0                &&
+        if (CE_IsCallTo (E, "ldax0sp")                               &&
+            (N = CS_GetNextEntry (S, I)) != 0                        &&
+            (CE_IsCallTo (N, "incsp2") || CE_IsJumpTo (N, "incsp2")) &&
             !CE_HasLabel (N)) {
 
             CodeEntry* X;
@@ -587,9 +585,6 @@ unsigned OptTosLoadPop (CodeSeg* S)
 
             /* Delete the old code */
             CS_DelEntries (S, I, 2);
-
-            /* Regenerate register info */
-            CS_GenRegInfo (S);
 
             /* Remember we had changes */
             ++Changes;
@@ -1203,7 +1198,7 @@ unsigned OptBinOps2 (CodeSeg* S)
 }
 
 
-unsigned OptTosPushPop(CodeSeg *S)
+unsigned OptTosPushPop (CodeSeg* S)
 /* Merge jsr pushax/j?? popax */
 {
     unsigned Changes = 0;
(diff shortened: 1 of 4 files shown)


colinleroy:
(I don't understand the Windows failure, could someone with Windows investigate it?)

mrdudz:
It is failing one of the regression tests (they only run in the windows build)
```
val/zx02.O.6502.prg
../../testwrk/val/cq84.O.6502.s:751: Error: Local proc ‘_glork’ was not closed
../../testwrk/val/cq84.O.6502.s:32: Error: Exported symbol ‘_main’ was never defined
../../testwrk/val/cq84.O.6502.s:31: Error: Exported symbol ‘_section’ was never defined
```
Try "make check" locally, that should give you the same

colinleroy:
> It is failing one of the regression tests (they only run in the windows build)
> 
> ```
> val/zx02.O.6502.prg
> ../../testwrk/val/cq84.O.6502.s:751: Error: Local proc ‘_glork’ was not closed
> ../../testwrk/val/cq84.O.6502.s:32: Error: Exported symbol ‘_main’ was never defined
> ../../testwrk/val/cq84.O.6502.s:31: Error: Exported symbol ‘_section’ was never defined
> ```
> 
> Try "make check" locally, that should give you the same

That's what I don't understand, it works on Linux (both make test and make check). With that optimisation pass, testwrk/val/cq84.O.6502.s' _glork proc consists of a single rts:

```
--- testwrk-ref/val/cq84.O.65c02.s	2025-07-20 11:29:44.162366713 +0200
+++ testwrk-OptPushPop/val/cq84.O.65c02.s	2025-07-20 12:51:13.298578633 +0200
@@ -753,14 +753,9 @@
 .segment	"CODE"
 
 ;
-; int glork(int x){
-;
-	jsr     pushax
-;
 ; return x;}
 ;
-	jsr     ldax0sp
-	jmp     incsp2
+	rts
 
 .endproc
 ```

But I don't understand why the result would be different between Windows and Linux.

**Comment by mrdudz** [comment]
>  don't understand why the result would be different between Windows and Linux.

Line endings involved somehow? But indeed, strange if it works on Linux and not on Windows

Another difference iirc is that on Linux "long" is 64bit, but on Windows it is 32bit.

**Comment by colinleroy** [comment]
I think it was a memory issue, removing code before adding the new one, with the new one referencing the old one's lineinfo (E->LI).

File: src/cc65/coptmisc.c [hunk 1]
```diff
@@ -1091,3 +1142,54 @@ unsigned OptBinOps2 (CodeSeg* S)
...
+    while (I < CS_GetEntryCount (S)) {
+
+        const CodeEntry* N;
+
+        /* Get the next entry */
+        const CodeEntry* E = CS_GetEntry (S, I);
+
+        /* Check for decspn, incspn, subysp or addysp */
+        if (E->OPC == OP65_JSR                            &&
+            strcmp(E->Arg, "pushax") == 0                 &&
```
kugelfuhr, line 1164:
```suggestion
        if (CS_IsCallTo (E, "pushax")                     &&
```

src/cc65/coptmisc.c
```diff
@@ -555,6 +555,57 @@ unsigned OptAXLoad (CodeSeg* S)
...
+    while (I < CS_GetEntryCount (S)) {
+
+        const CodeEntry* N;
+
+        /* Get the next entry */
+        const CodeEntry* E = CS_GetEntry (S, I);
+
+        /* Check for the sequence */
+        if (E->OPC == OP65_JSR                            &&
+            strncmp (E->Arg, "ldax0sp", 5) == 0           &&
```
kugelfuhr, line 575: Why do you check only the first 5 characters? There is currently no other runtime routine that starts with `ldax0` but once one is added the code will probably fail.

src/cc65/coptmisc.c
```diff
@@ -555,6 +555,57 @@ unsigned OptAXLoad (CodeSeg* S)
...
+            CodeEntry* X;
+
+            X = NewCodeEntry (N->OPC, AM65_ABS, "popax", 0, N->LI);
+            CS_InsertEntry (S, X, I+2);
+
+            /* Delete the old code */
+            CS_DelEntries (S, I, 2);
+
+            /* Regenerate register info */
+            CS_GenRegInfo (S);
```
kugelfuhr, line 590: The call to `CS_GenRegInfo` is only necessary if you add code that might generate a match in a later round which is not the case for `jsr popax`. It's just a performance issue so you can leave it in place, but it's not required.

src/cc65/coptmisc.c
```diff
@@ -1091,3 +1142,54 @@ unsigned OptBinOps2 (CodeSeg* S)
...
+
+        /* Get the next entry */
+        const CodeEntry* E = CS_GetEntry (S, I);
+
+        /* Check for decspn, incspn, subysp or addysp */
+        if (E->OPC == OP65_JSR                            &&
+            strcmp(E->Arg, "pushax") == 0                 &&
+            (N = CS_GetNextEntry (S, I)) != 0             &&
+            (N->OPC == OP65_JSR || N->OPC == OP65_JMP)    &&
+            strcmp(N->Arg, "popax") == 0                  &&
```
kugelfuhr, line 1167:
```suggestion
            strcmp (N->Arg, "popax") == 0                 &&
```
There are more places like this.

kugelfuhr (commented):
One additional problem: The optimization works only when `-O` is given, which limits its usefulness. With `-Oi` you will still get the old code.

src/cc65/coptmisc.c
kugelfuhr: Sorry, that must read `CE_IsCallTo`.

src/cc65/coptmisc.c
colinleroy: That's a copy/paste error, thanks!

colinleroy:
> One additional problem: The optimization works only when `-O` is given, which limits its usefulness. With `-Oi` you will still get the old code.

Indeed. That's going to be either also matching inlined functions, or maybe inlining after optimizing? Harder, anyway, maybe out of scope of this one?

kugelfuhr:
> Harder, anyway, maybe out of scope of this one?

It's a pity since the optimization is good but there's much code translated with `-Oi`. But you're right, out of scope.

mrdudz:
Would you fix the conflicts please? :)

colinleroy:
> Would you fix the conflicts please? :)

Sorry, done :)

mrdudz:
thanks!

PR #2813, opened 2025-07-20
Fix bug in PR #2778

colinleroy:
Don't replace incaxy like incax[1-8].

## Summary

Introduced in PR #2778. Fixed by special-casing ldy/jsr incaxy/jsr ldaxi.
Updated the test to make sure we test that code path too.

## Checklist

### DIFF
--- a/src/cc65/coptmisc.c
+++ b/src/cc65/coptmisc.c
@@ -517,7 +517,8 @@ unsigned OptAXLoad (CodeSeg* S)
         /* Check for incax[1-8] followed by jsr/jmp ldaxi */
         if (E->OPC == OP65_JSR                            &&
             strncmp (E->Arg, "incax", 5) == 0             &&
-            strcmp (E->Arg, "incaxy") != 0                &&
+            E->Arg[5] >= '1' && E->Arg[5] <= '8'          &&
+            E->Arg[6]  == '\0'                            &&
             (N = CS_GetNextEntry (S, I)) != 0             &&
             (N->OPC == OP65_JSR || N->OPC == OP65_JMP)    &&
             strcmp (N->Arg, "ldaxi") == 0                 &&


src/cc65/coptmisc.c
```diff
@@ -555,6 +556,65 @@ unsigned OptAXLoad (CodeSeg* S)
...
+        CodeEntry* E[3];
+        CodeEntry *X;
+        char *End;
+
+        /* Get the next entry */
+        E[0] = CS_GetEntry (S, I);
+
+        /* Check for ldy followed by incaxy followed by jsr/jmp ldaxi */
+        if (E[0]->OPC == OP65_LDY                            &&
+            E[0]->AM == AM65_IMM                             &&
```
kugelfuhr, line 579:
```suggestion
        if (E[0]->OPC == OP65_LDY                            &&
            CE_IsConstImm (E[0])                             &&
```
The reason why this is required is that the argument to "ldy immediate" is not necessarily numeric. The following translates to the immediate load of a symbolic constant into Y:

```
void f(void)
{
    asm ("ldy #<_f");
    asm ("sty _a");
}
```
Just look at the generated assembler code. It would match your test but converting it to a numerical value will silently fail.

src/cc65/coptmisc.c
```diff
@@ -555,6 +556,65 @@ unsigned OptAXLoad (CodeSeg* S)
...
+            CS_GetEntries (S, E+1, I+1, 2)                   &&
+            E[1]->OPC == OP65_JSR                            &&
+            strcmp (E[1]->Arg, "incaxy") == 0                &&
+            (E[2]->OPC == OP65_JSR || E[2]->OPC == OP65_JMP) &&
+            strcmp (E[2]->Arg, "ldaxi") == 0                 &&
+            !CS_RangeHasLabel (S, I, 3)) {
+
+            /* Replace with ldy (y+1) / jsr ldaxidx */
+            Val = strtoul(E[0]->Arg + 1, &End, 16);
+            Val++;
```
kugelfuhr, line 589:
```suggestion
            Val = E[0]->Num + 1;
```
If you use `CE_IsKnownImm()` above, you can just use the - already known - numeric value here.

src/cc65/coptmisc.c
colinleroy: Thanks! I wasn't aware of that :)

src/cc65/coptmisc.c
```diff
@@ -514,9 +514,10 @@ unsigned OptAXLoad (CodeSeg* S)
         /* Get the next entry */
         const CodeEntry* E = CS_GetEntry (S, I);
 
-        /* Check for incax followed by jsr/jmp ldaxi */
+        /* Check for incax[1-8] followed by jsr/jmp ldaxi */
         if (E->OPC == OP65_JSR                            &&
             strncmp (E->Arg, "incax", 5) == 0             &&
+            strcmp (E->Arg, "incaxy") != 0                &&
```
kugelfuhr, line 520:
```suggestion
            strncmp (E->Arg, "incax", 5) == 0             &&
            E->Arg[5] >= '1' && E->Arg[5] <= '8'          &&
            E->Arg[6]  == '\0'                            &&
```
It is better to explicitly check for incax[1-8]. Otherwise the code will fail if someone adds an `incaxa` or similar.

colinleroy:
If everything is OK with this one, it could be nice to merge it, it fixes a regression :)

mrdudz:
Sorry, i have an active (work) project again, so things will be a bit slower... (ping me if needed)

Looks good - so merged :)

colinleroy:
> Sorry, i have an active (work) project again, so things will be a bit slower... (ping me if needed)

No problem! I'm not in a hurry for the two other ones :)

PR #2824, opened 2025-08-26
ca65 Fixing segfault when using `--expend-macros` without `--listing`

GrosChien:
## Summary

With ca65, using `-x` / `--expand-macros` without `-l` / `--listing` will cause a segmentation fault when expending a macro:

```
;test.s
.macro test_mac

.endmacro

test_mac
```
Running
```
$ ca65 -x test.s
```
will result in a segmentation fault.
This is also the case with `.define`.

This fix will make `-x` have no effect if `-l` is omitted.

## Checklist

### DIFF
--- a/src/ca65/macro.c
+++ b/src/ca65/macro.c
@@ -728,7 +728,7 @@ ExpandParam:
 
         /* Use next macro token */
         TokSet (Mac->Exp);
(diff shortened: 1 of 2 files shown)


mrdudz:
Thanks!

PR #2829, opened 2025-09-08
Fix hardcoded upper limit of ld65 input files

### LINKED ISSUE #2826: ld65: Increase MAX_INPUTFILES or Make Runtime Configurable
**Is your feature request related to a problem? Please describe.**
Currently a project I am working on is running up against an arbitrary 256 file limit in ld65.

**Describe the solution you'd like**
Either simply the MAX_INPUTFILES definition in ld65's main.c file needs to be increased or, ideally, this can be converted into reading an environment variable or field in the linker config and only fall back to this restrictive 256 limit when other directives are not given. That is of course if an arbitrary limit must be provided at all.

For the record, I pulled cc65 down, increased this limit from 256 to 512, and my project once again had zero problems linking. For this reason I believe another solution would be to attempt the necessary RAM allocation to handle whatever number of files are needed, rejecting the job if the *machine* cannot give enough RAM, rather than limiting every single user to a maximum of 256 assembly units.

Unfortunately even with a fix this makes me very hesitant to continue working on the current project given a fix wouldn't result in end users having that toolkit revision available immediately, kneecapping anyone using my code with an older copy of ld65.

Happy to take alternative suggestions to link more than 256 files without modifying ld65 outright....

kugelfuhr:
## Summary

Fixes #2826. 

## Checklist

### DIFF
--- a/src/ld65/main.c
+++ b/src/ld65/main.c
@@ -88,16 +88,46 @@ static unsigned         LibFiles   = 0; /* Count of library files linked */
 #define INPUT_FILES_SGROUP     3        /* Entry is 'StartGroup' */
 #define INPUT_FILES_EGROUP     4        /* Entry is 'EndGroup' */
 
-#define MAX_INPUTFILES         256
-
 /* Array of inputs (libraries and object files) */
-static struct InputFile {
-    const char *FileName;
-    unsigned Type;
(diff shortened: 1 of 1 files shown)


mrdudz:
Thanks!

PR #2834, opened 2025-09-17
Fixed overflow

polluks:
## Summary

A C string like "0123456789ABCDEF" is 17 bytes long, including the null terminator '\0'.
```
common/gentype.c: In function 'GT_AsString':
common/gentype.c:105:36: warning: initializer-string for array of 'char' truncates NUL terminator but destination lacks 'nonstring' attribute (17 chars into 16 available) [-Wunterminated-string-initialization]
  105 |     static const char HexTab[16] = "0123456789ABCDEF";
      |                                    ^~~~~~~~~~~~~~~~~~
```
## Checklist

### DIFF
--- a/src/common/gentype.c
+++ b/src/common/gentype.c
@@ -102,10 +102,7 @@ const char* GT_AsString (const StrBuf* Type, StrBuf* String)
 ** will be zero terminated and a pointer to the contents are returned.
 */
 {
-    static const char HexTab[16] = {
-        '0','1','2','3','4','5','6','7',
-        '8','9','A','B','C','D','E','F'
-    };
+    static const char HexTab[] = "0123456789ABCDEF";
     unsigned I;
 
     /* Convert Type into readable hex. String will have twice then length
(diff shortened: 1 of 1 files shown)


kugelfuhr:
If it ain't broke, don't fix it. The original code works, is compliant and has no overflow. See ISO/IEC 9899:1999 6.7.8.15:

> An array of character type may be initialized by a character string literal, optionally enclosed in braces. Successive characters of the character string literal (including the terminating null character if there is room or if the array is of unknown size) initialize the elements of the array.

polluks:
@kugelfuhr gcc gives a warning, however a simple [17] may be more readable.

kugelfuhr:
Ok. If gcc warns then that's a reason to fix it. But you might want to change the title to "Fix gcc warning" or something like that because there is no overflow.
The change as in this PR is better than using `[17]` in my eyes since the latter adds an unused element for no good reason.

polluks:
Well, this won't affect code generation. One byte for better readability :-)

mrdudz:
What GCC is it that warns about this? Interesting...

> The change as in this PR is better than using [17] in my eyes since the latter adds an unused element for no good reason.

yeah, [17] would be weird here

Thanks!

PR #2839, opened 2025-09-30
Fix uncatalogued name clash with libsrc/runtime/leave.s

danielg4:
Even though these labels are not exported, they can show up in linker maps and thus break o65 modules.

## Summary

Provide description of the bug (if an issue exists, link to it) and the fix.

## Checklist

### DIFF
--- a/libsrc/common/memset.s
+++ b/libsrc/common/memset.s
@@ -81,13 +81,13 @@ L1:     .repeat 2               ; Unroll this a bit to make it faster
 ; Set the remaining bytes if any
 
 L2:     ldy     ptr3            ; Get the low byte of n
-        beq     leave           ; something to set? No -> leave
+        beq     @leave          ; something to set? No -> leave
 
-L3:     dey
+@L3:    dey
         sta     (ptr1),y                ; set bytes in low
         sta     (ptr2),y                ; and high section
-        bne     L3              ; flags still up to date from dey!
-leave:
+        bne     @L3             ; flags still up to date from dey!
+@leave:
         jmp     popax           ; Pop ptr and return as result
 
 
--- a/libsrc/common/strcspn.s
+++ b/libsrc/common/strcspn.s
@@ -27,21 +27,21 @@ _strcspn:
 loadChar:
         ldy #0
         lda (ptr1),y        ; get next char from s1
-        beq leave           ; handly byte of s1
-advance:
+        beq @leave          ; handly byte of s1
+@advance:
         inc ptr1            ; advance string position to test
-        bne check
+        bne @check
         inc ptr1+1
         dey                 ; correct next iny (faster/shorter than bne...)
 
-checkNext:
+@checkNext:
         iny
-check:  cpy tmp1            ; compare with length of test character string
+@check: cpy tmp1            ; compare with length of test character string
         beq endOfTestChars
         cmp (ptr4),y        ; found matching char?
-        bne checkNext
+        bne @checkNext
 
-leave:  txa                 ; restore position of finding
+@leave: txa                 ; restore position of finding
         ldx tmp2            ; and return
         rts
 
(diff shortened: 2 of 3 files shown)


mrdudz:
Interesting problem.... doesn't this mean that basically the library should only ever use local labels (except when they are exported)? mmmh

anyway, no problem with this change, so merging. thanks!

danielg4:
It would potentially mean that, yes. Take it up with the person who designed the o65 scheme.
### How I ran into this:
Twenty years ago, I downloaded Contiki, intrigued by the idea that it could be built for even a vic20 target. However, at that time, the Contiki 1.x -> 2.x transition was pretty recent, and I was puzzled by the lack of vic20 support in 2.x and thought I'd look into adapting 2.x's c64 support for it, but I didn't get far. Well, the other day, while waiting for Android to build inside a slow VM, I thought I'd take another look, and I saw that the 2.x/3.x codebase was abandoned in favor of Contiki-NG, but that's aimed only at hardware from the present century, so I cloned the 2.x/3.x repo and gave it a build with current cc65. Trying that made me remember why I gave up back then: Even the most basic build of any revision at all of that repo ends up being too big to fit on a c64, let alone a vic20. However, this time, the curiosity about vic20 support led me to do something with which I didn't bother back then: Grab the 1.x sources and try to build that. However, those effectively haven't been touched since 2007, so building with **current** cc65 would take a bit of doing. I ended up modifying 78 source files in order for the build not to fail for vic20, c64, and c128, but it still required THIS fix to cc65 to keep it from erroring out due to duplicate symbols.
### Epilogue
The vic20 build completes but doesn't appear to run, and I haven't really looked into why. The c64 build runs, but trying to launch any program from within it crashes the whole thing, and again, I haven't looked into why. The c128 build runs and launches programs fine, but the GUI can display only simple text on the screen, unlike the c64 build, which has a very impressive GUI display. Maybe someday, if I'm still alive, I might look into this more. Thanks for your work.

mrdudz:
I see. That GUI might actually be the reason for why the c64 one crashes - it might just run out of memory :) IIRC you can build contiki without it, ie use the text based UI - perhaps try that. That was also the reason for why that GUI was removed completely in 2.x back then (and at least in the C64 port changed to be single-program only)

kugelfuhr:
@mrdudz @danielg4 I know this has already been merged, but could someone please explain the problem to me? I don't understand why symbols being added to a text file (the linker map) should break o65 modules.

danielg4:
> @mrdudz @danielg4 I know this has already been merged, but could someone please explain the problem to me? I don't understand why symbols being added to a text file (the linker map) should break o65 modules.

Binding o65 modules to their API's requires being able to resolve their symbols, and this is done by grabbing them from the linker map. Example:
https://github.com/adamdunkels/contiki-1.x/blob/master/contiki-cc65/make-labels

kugelfuhr:
So this is definitely not a cc65 problem and the change should never have been merged. At least not without a thorough scrutiny.

What happened was this: The contiki make process uses the VICE label file (not the map file!) and parses it for some symbols that it wants to export. The VICE label file is a hack, and it was never thought for things like that. Making library code changes just to keep this sort of abuse working is a quite a bad idea in my eyes.

If you want this ugly hack to work, you will have to disallow any duplicate names in the library, even for local, non exported symbols. Meaning, two completely unrelated library modules written by different authors must have a completely distinct set of local, non exported symbols like "loop" or whatever. Such symbols must be unique within the complete library now and in the future. Otherwise all but one of them get dropped in the VICE label file leading to problems for people abusing the label file. Good luck with that!

It's spilled milk now, but my suggestion would be to understand a problem before trying to fix it.

mrdudz:
> and the change should never have been merged.

To me this is merely a cosmetical thing. The change is just fine, it creates no potential problem, so why not.

kugelfuhr:
"Fixing" the problem in cc65 prevents the real problem to be fixed - which is in Contiki. That is bad.

mrdudz:
> "Fixing" the problem in cc65 prevents the real problem to be fixed - which is in Contiki. That is bad.

That is true indeed. However, contiki moved on long ago - it wont get fixed :)

PR #2843, opened 2025-10-19
Emit warnings for wasted bytes when aligning

colinleroy:
## Summary

This patch makes it easier to see where and how many bytes are "wasted" when aligning, which can help the user fill in the holes with variables or code that don't require alignment.

## Checklist

### DIFF
--- a/doc/ca65.sgml
+++ b/doc/ca65.sgml
@@ -133,6 +133,7 @@ Long options:
   --target sys                  Set the target system
   --verbose                     Increase verbosity
   --version                     Print the assembler version
+  --warn-align-waste            Print bytes "wasted" for alignment
   --warnings-as-errors          Treat warnings as errors
 ---------------------------------------------------------------------------
 </verb></tscreen>
@@ -405,6 +406,13 @@ Here is a description of all the command line options:
   even warn about such things like unused imported symbols. The default
   warning level is 1, and it would probably be silly to set it to
   something lower.
+
+
+  <label id="option--warn-align-waste">
+  <tag><tt>--warn-align-waste</tt></tag>
+
+  Warnings will be generated when alignment requirements cause emission of
+  fill bytes.
 
 
   <label id="option--warnings-as-errors">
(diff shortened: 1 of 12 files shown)


colinleroy:
(this breaks tests because it adds warnings, if you're interested in that PR I'll fix them)

kugelfuhr:
While I can see the usefulness of this warning in some situations, I don't like the idea to have a warning for a completely legal pseudo op. With this change most code using `.align` will output warnings.

My suggestion is to make it switchable and maybe disable it by default.

mrdudz:
Yeah please add an option (and default off) - that will also fix the problem with the failing tests :)

polluks:
I was thinking of another level "note" which could be just below warning.

colinleroy:
I've been looking at it and it's a bit tiresome to add an option (it has to be passed from cc65 to ld65 too), so I'm marshalling motivation to do it :smile:

kugelfuhr:
Notes are already available but they are used as additional information in case of warnings or errors, so I would not use them here.

I would also say that passing the option from cl65 is not required, it's simply not important enough to justify the additional code. There's the `-Wa assembler-option` for cl65 for exactly those options that aren't important enough to be added to cl65. So adding something like `--warn-align-space` to the assembler would be sufficient - if you manage to build up enough motivation for that 😄

colinleroy:
OK, done :)

> I would also say that passing the option from cl65 is not required, it's simply not important enough to justify the additional code. There's the `-Wa assembler-option` for cl65 for exactly those options that aren't important enough to be added to cl65. So adding something like `--warn-align-space` to the assembler would be sufficient - if you manage to build up enough motivation for that 😄

In the end I added it also to cl65 as it was very easy, and that'll be shorter for users than passing `--asm-args` and `--ld-args`.

Documentation and `--help`s updated (`--warning-as-errors` was missing for ld65).

colinleroy:
ping @mrdudz :)

mrdudz:
I'm here!

looks good now, thanks!

PR #2846, opened 2025-10-25
Apple2 small size optimisations

colinleroy:
## Summary

A few factorizations and reworks to spare some bytes around Apple II ProDOS I/O. 
The total saving is about 64 bytes. 
Everything tested on my file manager project that links it all.
Review is easier commit-by-commit.

## Checklist

### DIFF
--- a/libsrc/apple2/mli.inc
+++ b/libsrc/apple2/mli.inc
@@ -53,6 +53,13 @@ EOF_COUNT       = 2
 
         .struct MLI
                 .union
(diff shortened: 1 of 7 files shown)


colinleroy:
ping @oliverschmidt :)

oliverschmidt:
I must admit that I don't fully understand all details, but the general concept of makes sense to me and I don't find any style issue to complain about ;-)
Thanks for pointing out that you tested it with your file manager :-)

PR #2848, opened 2025-10-27
Apple2: Don't consider STARTUP when computing available iobufs

colinleroy:
## Summary

When using --start-addr 0xBF4, we get a nicely aligned LOWCODE at $C00:
```
Name                   Start     End    Size  Align
----------------------------------------------------
NULL                  000000  000000  000000  00001
ZEROPAGE              000080  000099  00001A  00001
EXEHDR                000BBA  000BF3  00003A  00001
STARTUP               000BF4  000BFF  00000C  00001
LOWCODE               000C00  001FD9  0013DA  00100
```

But if using `iobuf-0800`, the computation considers the STARTUP as __MAIN_START__ = start-addr - STARTUP size, and refuses to allocate a buffer.

At the time one can actually use iobuf-0800 for IO, the STARTUP segment has become useless.

## Checklist

### DIFF
--- a/libsrc/apple2/crt0.s
+++ b/libsrc/apple2/crt0.s
@@ -7,6 +7,7 @@
         .export         done, return
         .export         zpsave, rvsave, reset
         .export         __STARTUP__ : absolute = 1      ; Mark as startup
+        .export         __STARTUP_SEGMENT_START__
 
         .import         initlib, _exit
         .import         zerobss, callmain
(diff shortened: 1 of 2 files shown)


mrdudz:
one for @oliverschmidt :)

oliverschmidt:
* I'm wondering if it is really worth the additional complexity to save 0xC bytes.
* I can't exactly point out a good example, but it seems "somehow" that the PR makes assumptions that might easily be wrong with other linker configs.

colinleroy:
> * I'm wondering if it is really worth the additional complexity to save 0xC bytes.

Understandably. It actually saves much more when I want to start LOWCODE with an aligned lookup-table, but it may be a corner-case. 

>     * I can't exactly point out a good example, but it seems "somehow" that the PR makes assumptions that might easily be wrong with other linker configs.

That's also possible. If you're unsure it's worth it, we can drop that PR, no worry :)

oliverschmidt:
I must admit that I'd rather like to not have this PR. Thank you very much for making that decision easy!

Issue #2861, opened 2025-11-18
C65 target: cputc / cpeekcolor don't know how to read the bottom half of color RAM in 80 column mode


dansanderson:
**Describe the bug**
As reported in https://github.com/cc65/cc65/issues/2857 , the Commodore 65 target does not know how to access color memory for the bottom half of the screen. cputc and cpeekcolor access $D800 for color RAM, but this only represents the first 1,000 positions for 40x25 text mode, as on a C64. To access the rest in 80x25 text mode, a program must flip the CRAM2K register D030 bit 1, which shadows I/O registers to expose more color memory.

I'm filing this bug to track my PR that has a fix.

**To Reproduce**
```
#include <conio.h>

void main(void)
{
    textcolor(1);
    clrscr();
    textcolor(0); cprintf("one\r\n");
    textcolor(1); cprintf("two\r\n");
    textcolor(2); cprintf("three\r\n");
    textcolor(3); cprintf("four\r\n");
    textcolor(4); cprintf("five\r\n");
    textcolor(5); cprintf("six\r\n");
    textcolor(6); cprintf("seven\r\n");
    textcolor(7); cprintf("eight\r\n");
    textcolor(8); cprintf("nine\r\n");
    textcolor(9); cprintf("ten\r\n");
    textcolor(10); cprintf("eleven\r\n");
    textcolor(11); cprintf("twelve\r\n");
    textcolor(12); cprintf("thirteen\r\n");
    textcolor(13); cprintf("fourteen\r\n");
    textcolor(14); cprintf("fifteen\r\n");
    textcolor(15); cprintf("sixteen\r\n");
    textcolor(16); cprintf("seventeen\r\n");
    textcolor(17); cprintf("eighteen\r\n");
    textcolor(18); cprintf("nineteen\r\n");
    textcolor(19); cprintf("twenty\r\n");
    textcolor(20); cprintf("twenty-one\r\n");
    textcolor(21); cprintf("twenty-two\r\n");
    textcolor(22); cprintf("twenty-three\r\n");
    textcolor(23); cprintf("twenty-four\r\n");
    textcolor(24); cprintf("twenty-five\r\n");

    while (1) {}
}
```

Compiled with: `cl65 -t c65 -o mytest.prg mytest.c`

Tested with the xmega65 emulator. Prior to the fix, the bottommost lines appear white. After the fix, they appear in the requested colors.

[...]

PR #2862, opened 2025-11-18
c65 target: conio accesses all color RAM in 80x25 text mode

dansanderson:
## Summary

For the Commodore 65 (c65) target, conio's cputc and cpeekcolor can now access color RAM for the bottom half of the 80x25 text screen. Previously, it could only access the top half, causing the bottom half to appear in the foreground color set prior to running the program. The C65 puts the first 1,000 bytes of color RAM at $D800 similar to the C64, and uses a bank switching register CRAM2K to expose the next 1,000 bytes, shadowing the I/O registers.

For this fix I decided to always flip the Interrupts flag and CRAM2K when setting or reading color RAM, instead of trying to decide whether or not the bottom half of the screen was being accessed. There might be cycles that can be shaved with a branching path, and it might be worth considering for cputc given how frequently it is called. With this version it looks to be about +20 cycles per character on a 3.5 MHz CPU.

I did not implement this fix for the mega65 target because there's a better option available for that platform. It'll require a bit more work to implement, but the result will be able to support the MEGA65's 80x50 text mode as well. I'll consider that separately.

Fixes https://github.com/cc65/cc65/issues/2861

## Checklist

### DIFF
--- a/libsrc/c65/cpeekcolor.s
+++ b/libsrc/c65/cpeekcolor.s
@@ -18,6 +18,7 @@ _cpeekcolor:
         lda     SCREEN_PTR
         sta     ptr1
 
+        php
         sei
         lda     $D030
         ora     #$01
@@ -28,8 +29,8 @@ _cpeekcolor:
         lda     $D030
         and     #$FE
         sta     $D030
-        cli
         tya
+        plp
 
         ldx     #>$0000
         rts
--- a/libsrc/c65/cputc.s
+++ b/libsrc/c65/cputc.s
@@ -116,6 +116,7 @@ putchar:
         adc     #>$D000
         sta     ptr4 + 1
 
+        php
         sei
         lda     $D030
         ora     #$01
@@ -125,6 +126,6 @@ putchar:
         lda     $D030
         and     #$FE
         sta     $D030
-        cli
+        plp
 
         rts


dansanderson:
(I have a MEGA65 version also but will submit it separately.)

mrdudz:
> php, sei, plp is better than sei, cli?

Just sei/cli should be good enough

dansanderson:
php sei and plp preserve the previous state of the interrupts flag. If conio or cc65 in general want to own the interrupt flag and there's no guarantee that conio will preserve the flag, I'm happy to change it to sei/cli.

Same goes for CRAM2K, which I'm also trying to preserve the previous setting of. Your call.

dansanderson:
Updated so that conio owns the Interrupt flag and CRAM2K register, don't stash and restore.

mrdudz:
> php sei and plp preserve the previous state of the interrupts flag. If conio or cc65 in general want to own the interrupt flag and there's no guarantee that conio will preserve the flag, I'm happy to change it to sei/cli.

Now that i read this again.... i am unsure if it wouldnt be better to preserve the state of the interrupt flag - although we don't really support calling C code in the interrupt ... mmh

(and sorry for late reply - in december i am busy with VICE release....)

polluks:
Indeed, the I flag, like every CPU flag, is just undefined behavior.

mrdudz:
> Indeed, the I flag, like every CPU flag, is just undefined behavior.

yes and no. There might be cases where the main program disables interrupts, and then this conio function would enable them again.

@dansanderson please revert that again (sorry) and use php/sei/plp instead

For $d030 its fine to not preserve it, i think

dansanderson:
Done, re-tested.

libsrc/c65/cpeekcolor.s
```diff
@@ -18,8 +18,19 @@ _cpeekcolor:
...
+        ora     #$01
+        sta     $D030
         ldy     #0
         lda     (ptr1),y
+        tay
+        lda     $D030
+        and     #$FE
+        sta     $D030
+        tya
+        plp
```
mrdudz: I wonder if it wouldn't make sense to also preserve the D030 bit here.... ie
lda $d030
pha
ora #1 
sta $d030
.....
pla
sta $d030

?

dansanderson: Are you asking me to roll back to my original version that preserves the previous CRAM2K setting? (6fb5934)

D030 consists of eight 1-bit registers that do different things, including other VIC RAM banking behaviors and PAL mode. Stashing and restoring the full eight bits with interrupts disabled should be safe.

mrdudz:
OK well, i guess it is fine as it is now - thanks!

PR #2863, opened 2025-11-18
Fix broken boolean not operator in ca65

### LINKED ISSUE #2859: Operator precedence is broken(?)
Someone came up with this in this thread: https://www.lemon64.com/forum/viewtopic.php?t=88106

The "boolean not" operator - quite unexpectedly - has the lowest precedence currently. The table in the documentation actually reflects this correctly: https://cc65.github.io/doc/ca65.html#ss5.5

However, this leads to quite unexpected behaviour:

these two work:
```
.if !.defined(__DOCART__) && .defined(__C64__)
.byte 1
.endif

.if .defined(__C64__) && (!.defined(__DOCART__))
.byte 2
.endif
```
but this produces an error:
```
.if .defined(__C64__) && !.defined(__DOCART__)
.byte 3
.endif
```
due to current operator precedence, it behaves as if it was written like this:
```
.if (.defined(__C64__) && !) .defined(__DOCART__)
.byte 4
.endif
```
... which is certainly not what i'd expect.

IMHO the precedence should be changed, so "boolean not" has the same precedence as unary operators, eg "bitwise not"

Unfortunately this stuff seems to be hardcoded in the assembler (expr.c) and cant be easily changed

kugelfuhr:
## Summary

This change fixes the incorrect handling of the boolean not operator (`.not`/`!`). It worked only when it was used at the start of a full expression. Fixes #2859.

Note: Operator precedence isn't changed, so `.not` still has the lowest precedence as described in the docs.

## Checklist

### DIFF
--- a/test/asm/val/bug2859.s
+++ b/test/asm/val/bug2859.s
@@ -1,0 +1,33 @@
+; Tests for the bug reported in #2859. Boolean not had the correct precedence
+; (as specified in the docs) but worked only correctly at the start of a full
+; expression.
+
+; This one assembles ok since ! is at the start
+.if !.defined(__DOCART__) && .defined(__C64__)
+.byte 1
+.endif
+
+; This one assembles too since a parenthesis restarts a full expression
+.if .defined(__C64__) && (!.defined(__DOCART__))
+.byte 2
+.endif
+
+; This one doesn't work since ! is somewhere in between
+.if .defined(__C64__) && !.defined(__DOCART__)
+.byte 3
+.endif
+
+
+        .import _exit
+        .export _main
+
+                 
+; The following code is an indirect test for the precedence of .not.
+; .not has the lowest precedence, so the expression that is loaded into A
+; evaluates to zero. This has of course to be changed when the precedence
+; of .not is changed.
+_main:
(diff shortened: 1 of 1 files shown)


mrdudz:
Thanks. i am really curious what the 3rd expression evaluates to now (i cant even imagine how it can evaluated when .not comes last)

kugelfuhr:
The third expression evaluates as you would expect it since no other operators follow. Boolean not having the lowest precedence means that things like

```.if !.defined(__C64__) && .defined(__DOCART__)```

come out differently from C since it is equal to

```.if !(.defined(__C64__) && .defined(__DOCART__))```

(`&&` binds tighter than `!`).

mrdudz:
> Boolean not having the lowest precedence means that things like [...] come out differently from C 

Its how it currently works, yes ("expected" as in "as documented")- but very different from every other assembler i used so far, and not what i'd expect.

IMHO it should be changed to match what C does, because that is basically what people would think it works like. (and writing !defined isnt exactly an uncommon thing to do in such expressions)

(But we should discuss it in a seperate issue - and call for comments from others)

PR #2911, opened 2025-12-06
Time zone support for target RP6502

rumbledethumps:
Take two. A perfect storm kept blowing me off the path. It's clear now how the original time author wanted things to progress.

Nothing changes in common/localtime. I originally thought it was unfinished, but it's actually correct for how C did things before we could justify the bytes for daylight savings time. The change is only adding a prototype for the standard tzset() and the target replacing localtime() in its entirety. Trying to keep the common/localtime logic for dst is most of what was leading me astray.

The entire time library appears to be working on RP6502 target exactly like any other C platform. Because the RTC on the RP6502-RIA-W sets itself with NTP, the time library is now truly state-of-the-art!

### DIFF
--- a/libsrc/rp6502/_localtime.c
+++ b/libsrc/rp6502/_localtime.c
@@ -8,9 +8,9 @@ struct tm* __fastcall__ _time_t_to_tm (const time_t t);
 
 struct tm* __fastcall__ _localtime (const time_t* timep)
 {
-    time_t time = *timep;
     struct tm* tm;
-    ria_set_axsreg (*timep);
+    time_t time = *timep;
+    ria_set_axsreg (time);
     time += ria_call_long (RIA_OP_TZQUERY);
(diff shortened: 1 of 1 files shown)


libsrc/rp6502/_localtime.c
```diff
@@ -0,0 +1,31 @@
...
+
+static bool tzset_set = false;
+
+struct tm* __fastcall__ _time_t_to_tm (const time_t t);
+
+struct tm* __fastcall__ _localtime (const time_t* timep)
+{
+    time_t time = *timep;
+    struct tm* tm;
+    ria_set_axsreg (*timep);
```
kugelfuhr: The following change saves about 10 bytes:
```suggestion
    struct tm* tm;
    time_t time = *timep;
    ria_set_axsreg (time);
```

mrdudz:
thanks!

Issue #2920, opened 2026-01-24
Missing NUL after __func__


willisblackburn:
The special `__func__` identifier has stopped working sometime in last few years. It doesn't add a NUL after the function name, so it's not a proper C string.

```
$ cc65 --version
cc65 V2.19 - Git 034f73a9e

```

func.c:
```
#include <stdio.h>
int main(void) {
  printf("%s\n", __func__);
  return 0;
}
```

Compile func.c with `cc65 func.c`. In func.s we see:

```
.segment	"RODATA"

S0002:
	.byte	$6D,$61,$69,$6E
S0001:
	.byte	$25,$73,$0A,$00
```

`S0002` is "main" and the `S0001` is the format string `%s\n`. There should be a NUL ($00) after "main." If I run func as-is, the two strings run together:

```
$ cl65 -t sim6502 func.c -o func
$ sim65 func
main%s
```

kugelfuhr:
This was caused by 6a17aedd812059e3a71e2d3e2e655cc2c643270c (#2610). I will prepare a PR.

willisblackburn:
> This was caused by [6a17aed](https://github.com/cc65/cc65/commit/6a17aedd812059e3a71e2d3e2e655cc2c643270c) ([#2610](https://github.com/cc65/cc65/pull/2610)). I will prepare a PR.

I sent my own PR before I saw your reply, but not sure if it's the fix you had in mind.

kugelfuhr:
The PR is ok👍

mrdudz:
fixed via #2921

PR #2921, opened 2026-01-24
Add terminating NUL after __func__ string literal #2920

willisblackburn:
## Summary

Fixes the issue of `__func__` not being NUL-terminated by adding the NUL in `AddLiteral`. This function only seems to be used when the compiler wants to add a literal *defined by the compiler*. Literals defined in the program being compiled go through `StringConst`, which adds the NUL just after `ExitPoint`.

## Checklist

- [X] New unit tests have been added to prevent future regressions
- [ ] ~The documentation has been updated if necessary~ (N/A)

### DIFF
--- a/src/cc65/litpool.c
+++ b/src/cc65/litpool.c
@@ -470,6 +470,7 @@ Literal* AddLiteral (const char* S)
 {
     StrBuf SB;
     SB_InitFromString(&SB, S);
+    SB_AppendChar(&SB, '\0');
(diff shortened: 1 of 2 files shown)


kugelfuhr:
This fix is correct, simple to understand, and it also adds a test. So why not merge it?

mrdudz:
Because as always my time time is limited and i am busy with other stuff - just ping me if there is something pending that i forgot about...

PR #2928, opened 2026-02-08
Debug file format documentation

willisblackburn:
## Summary

Documentation for the debug file format.

Includes unit tests that exercise capabilities of debug file format (val_asm.s, val_func.c, val_types.c), then generate debug info for each file and compare with reference files. "make update" to update the reference files to reflect any (intended) changes.

## Checklist

- [X] New unit tests have been added to prevent future regressions
- [X] The documentation has been updated if necessary

### DIFF
--- a/test/dbginfo/Makefile
+++ b/test/dbginfo/Makefile
@@ -17,6 +17,7 @@ endif
 CL65 := $(if $(wildcard ../../bin/cl65*),..$S..$Sbin$Scl65,cl65)
 CC65 := $(if $(wildcard ../../bin/cc65*),..$S..$Sbin$Scc65,cc65)
 CA65 := $(if $(wildcard ../../bin/ca65*),..$S..$Sbin$Sca65,ca65)
+LD65 := $(if $(wildcard ../../bin/ld65*),..$S..$Sbin$Sld65,ld65)
 
 WORKDIR = ../../testwrk/dbginfo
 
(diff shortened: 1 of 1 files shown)


doc/dbginfo.sgml
```diff
@@ -0,0 +1,276 @@
+<!doctype linuxdoc system>
+
+<article>
+<title>cc65 debug info file format
+<author>
+<url url="mailto:uz@cc65.org" name="Ullrich von Bassewitz">
```
mrdudz: please change this to your name/email :)

willisblackburn:
Looks like some build failures... Will look at this later tonight.

mrdudz:
> Looks like some build failures... Will look at this later tonight.

Actually the test(s) you added fail, building works OK :)

willisblackburn:
I had to change the makefile to specify all intermediate file names, else the temporary files generated by the compiler will always produce diffs. The mtime of the compiler-generated files is also a problem, so I changed the process to zero out all the mtime attributes, which should avoid future diffs if other people's operating systems or Git clients don't handle file modification timestamps the same way that mine do.

Also updated the author and added the comment about the API module.

willisblackburn:
Looks like paths are also a problem.

spiro-trikaliotis:
> I had to change the makefile to specify all intermediate file names, else the temporary files generated by the compiler will always produce diffs. The mtime of the compiler-generated files is also a problem, so I changed the process to zero out all the mtime attributes, which should avoid future diffs if other people's operating systems or Git clients don't handle file modification timestamps the same way that mine do.
> 
> Also updated the author and added the comment about the API module.

Have you had a look into test/asm/listing/Makefile? It had the same problems that parts of the output depend on file names and paths output of the assembler, and so on. For this, I introduced the $(ISEQUAL) function (test/isequal.c) that allows me to add a wildcard <<<#PATH#>>> for a path name, and <<<#INTEGER#>>> for an integer value if the option --wildcard is given.

Would that help you in your case?

Issue #2925, opened 2026-02-06
Compiler input printed in diagnostics is wrong


kugelfuhr:
**Compiler and OS**

```
$ cc65 --version
cc65 V2.19 - Git 034f73a9e
```

**Describe the bug**
Using the `-v` switch, the compiler will print what it thinks is the current input when a diagnostic is given. Unfortunately this is wrong or misleading in several cases.

For one, the input is preprocessed, so one doesn't see the "real" input. For 

```C
int main() { a = __LINE__; return 0; }
```

it will output
```
test.c:1: Error: Undeclared identifier 'a'
Input: int main() { a = 1; return 0; }
```

In some other cases, the input is garbled because of preprocessing. For

```C
#include "foo.h"
```

the compiler will print
```
test.c:1: Error: Include file 'foo.h' not found
Input: "foo.h"
```

And in even other cases, the given input is simply wrong and doesn't match the error. If test.c contains

```C
#include "test.h"
int main() { return 0; }
```

and test.h contains
```C
foo();
```

the compiler will output:

```
test.h:1: Warning: Implicit 'int' type specifier is an obsolete feature
Input: int main() { return 0; }
```

The given input does clearly not match the line that caused the error.

**To Reproduce**
See samples above.

PR #2929, opened 2026-02-11
Modernize cc65 diagnostics

kugelfuhr:
## Summary

This PR changes diagnostics output by the compiler in a similar way as #2785 and #2802 did for the assembler and linker. Diagnostics will be colored and will use UTF-8 for typographically correct quotes. Similar as for the other tools, there are `--no-utf8` and `--color` command line switches to control usage in diagnostics. cl65 has been changed to pass these options also to the compiler.

Some other changes were also required:

* There was no good way to integrate printing the input line when coloring is enabled. Since the feature didn't work correctly in some cases, I have removed it alltogether instead of fixing it. So this change makes #2925 obsolete.
* The old code used `fprintf` while the new code requires more effort and uses formatted printing to string buffers instead. With this change a problem with the `xsnprintf` implementation popped up: It doesn't support floating point format specs. Since there is no longer a reason for a custom `snprintf` anyway (even Microsoft managed to deliver a conforming implementation in 2015), I simply removed it. `xsnprintf` is now a wrapper for `snprintf`. New code should call snprintf directly. Note: `xsprintf` is still useful since it does some checks.
* The `DiagnosticStrBuf` class has been renamed to `MngdStrBuf` and moved into its own module. The string buffers weren't really related to diagnostics and cluttered the error module.

## Checklist

### DIFF
--- a/src/common/xsprintf.h
+++ b/src/common/xsprintf.h
@@ -34,7 +34,7 @@
 
 
 /* The snprintf style functions in this module are obsolete. They call just
-** snprintf under the hood. New code should call snprintf directly. 
+** snprintf under the hood. New code should call snprintf directly.
 ** The sprintf style functions are still useful. They do some checks and will
 ** FAIL in case of errors (buffer too small or similar).
 */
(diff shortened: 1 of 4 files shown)


mrdudz:
?

Issue #1664, opened 2022-02-12
ca65 feature request: macros with variable number of parameters


WesleyAC:
I'd love to be able to have a macro that takes an arbitrary number of parameters. Most of the infrastructure to do what I need exists already — `.PARAMCOUNT` and `.REPEAT`. In fact, you can sorta hack it together if you need:

```
.macro push_constants args
	.repeat .paramcount + 1, I
		lda .mid(I*2, 2, args)
		pha
	.endrepeat
.endmacro

push_constants {#$69 #$42}
```

But that comes with the drawback that arguments must all be the same number of tokens, and you'll get very strange errors if any params are the wrong number of tokens.

I would love to see a `.param(n)` function that gets the nth parameter to a macro, and some way of indicating that a macro is allowed to have an indefinite number of parameters. That would be enough to make quite usable macros with variable numbers of parameters.

I also totally understand if this is a direction you don't want to go in, it's just something I found myself wishing was available :)

WesleyAC:
So, uh, I played with this more and you can actually totally do the parsing of a comma separated list yourself:

```
.macro push_args args
	.repeat .tcount({args}), RSTART
		START .set RSTART
		.if (RSTART = 0 || .xmatch({.mid(RSTART, 1, {args})}, {,}))
			.if (.xmatch({.mid(RSTART, 1, {args})}, {,}))
				START .set RSTART + 1
			.endif
			FOUND_END .set 0
			.repeat .tcount({args}) - START, REND
				END .set REND
				.if (FOUND_END = 0 && (REND = .tcount({args}) - START - 1 || .xmatch({.mid(START + REND + 1, 1, {args})}, {,})))
					.if (REND = .tcount({args}) - START - 1)
						END .set REND + 1
					.endif
					FOUND_END .set 1
					lda .mid(START, END + 1, {args})
					pha
				.endif
			.endrepeat
		.endif
	.endrepeat
.endmacro

push_args { #$69, 47, #$42 }
```

I definitely don't condone actually doing this in practice, and it has some limitations. The main one is that you can't use commas in the list items. You can replace `{,}` with `{.}` or `{::}` or some other separator of your choosing instead, but the ideal solution of escaping via `{` and `}` doesn't work, because ca65 (quite reasonably) does not allow those to be nested, afaict.

Anyway, the point is, I'd be very happy if it was possible to have macros with indefinite numbers of args without gross hacks like this :)

Movax12:
[...]

PR #2933, opened 2026-02-18
add .BANKS pseudo instruction

zlago:
## Summary

adds the .BANKS pseudo instruction to complete the .LOBYTES / .HIBYTES family, on targets which do not have a 24 bit address space, requiring the use of bankswitching instead

## Checklist

### DIFF
--- a/doc/ca65.sgml
+++ b/doc/ca65.sgml
@@ -2540,7 +2540,7 @@ See: <tt><ref id=".BYTE" name=".BYTE"></tt>,<tt><ref id=".CHARMAP" name=".CHARMA
 
   Please note that because <tt><ref id=".BANK" name=".BANK"></tt> is evaluated at link time,
   this command requires passing the --relax-checks switch on the command line.
-  
+
   Example:
 
   <tscreen><verb>


mrdudz:
Please describe what this does even do and why it is needed :)

And please add documentation for the added commands (and ideally i test as well)

pinobatch:
If I had to guess based on how this feature was pitched in another space (NESdev Discord server), it's for retrieving the `bank` value associated with the `MEMORY` area into which `ld65` places each of several symbols. Each of the following pairs would be equivalent:
```
.lobytes ww, xx, yy, zz
.byte <ww, <xx, <yy, <zz
.byte .lobyte(ww), .lobyte(xx), .lobyte(yy), .lobyte(zz)

.hibytes ww, xx, yy, zz
.byte >ww, >xx, >yy, >zz
.byte .hibyte(ww), .hibyte(xx), .hibyte(yy), .hibyte(zz)

.bankbytes ww, xx, yy, zz
.byte ^ww, ^xx, ^yy, ^zz
.byte .bankbyte(ww), .bankbyte(xx), .bankbyte(yy), .bankbyte(zz)

; This is the new one
.banks ww, xx, yy, zz
.byte <.bank(ww), <.bank(xx), <.bank(yy), <.bank(zz)
```
Its documentation might then look like this:
> Define byte sized data by extracting the bank associated with the memory area of each expression, usually a label. This is equivalent to `.BYTE` with `<.BANK(expr)` around each expression in its list.

This behavior cannot be implemented in terms of `.macro` because of #1664: ca65 does not support variadic macros, only macros with a fixed maximum argument count.

@zlago please confirm if I guessed right.

zlago:
> @zlago please confirm if I guessed right.

yes, thats the intended purpose

> > This is equivalent to `.BYTE` with `<.BANK(expr)` around each expression in its list.

this is not right, its intentionally implemented like `.BYTE .BANK(expr)` because that causes the linker to check for truncation (but requires passing `--relax-checks` to the assembler, or it wont assemble at all)

zlago:
proofreading/suggestions appreciated, and/or where i can read about adding unit tests (if theyre even necessary)

kugelfuhr:
The new control command seems a bit strange. What it does is defining bytes, with the `.bank` function applied to each expression However, the result of the `.bank` function is a 32 bit integer, not an 8 bit one. I cannot see an immediate justification to map it into an 8 bit value. If you do that, you might need `.banks16` or similar commands later.

Wouldn't it be possible to use a macro instead of a new control command?

zlago:
> The new control command seems a bit strange. What it does is defining bytes, with the `.bank` function applied to each expression However, the result of the `.bank` function is a 32 bit integer, not an 8 bit one. I cannot see an immediate justification to map it into an 8 bit value. If you do that, you might need `.banks16` or similar commands later.

- the target audience of `.banks` is the NESDev scene, where having less than 256 banks is the norm (and people using plain batch files with prebuilt ca65 isnt unusual)
- for the few projects which are big enough to have more than 256 banks, i expect them to either
  * already have some preprocessor to generate the `.byte <foo, >foo, <.bank(foo), >.bank(foo)` statements,
  * have bigger issues to worry about than lack of a `.hibanks` directive
  * and/or be able to "just" use their own ca65 fork with it.
- as for the name...
  * calling it [`.bankbytes`](https://cc65.github.io/doc/ca65.html#.BANKBYTES) is of out of the question,
  * `.lobanks` is not very indicative either since ca65 has both [`.lobyte`](https://cc65.github.io/doc/ca65.html#.LOBYTE) and [`.loword`](https://cc65.github.io/doc/ca65.html#.LOWORD),
  * `.lobankbytes`, `.lobytebanks`, etc. are probably not good names either
  * requiring `--relax-checks`[^checks] has the side effect of giving potential users an incentive to check the documentation for how to properly use the directive
  * ...that being said, `.bytebanks` might actually be a better name

> Wouldn't it be possible to use a macro instead of a new control command?

ca65 does not have variadic macros, so.. Depends, but there probably isnt a "normal" way to do this

[^checks]: i could not figure out a way to have range checking on the bank without requiring `--relax-checks`, and promptly settled on requiring it after figuring out this was exactly my solution the one time i mis-used `--auto-import` and had to deal with it complaining about by-default-absolute symbols being used in a `.byte` directive

kugelfuhr:
I see. It's a pity that the term "bank" is used for two different things within ca65: For the third byte of a 24 bit address and as a name for an attribute assigned to the memory area. Otherwise, `.bankbytes` could be used here making it clear that it's the byte value of the `bank` attribute.

So while I'm still not happy, I can understand the request.

Regarding macros: ca65 doesn't have variadic macros per se, but it could be emulated with an upper limit for the argument count as

```
.macro  bb a0, a1, a2, a3, a4, a5, a6, a7, a8, a9
        .ifnblank a0
                .byte   .bank(a0)
                bb      a1,a2,a3,a4,a5,a6,a7,a8,a9
        .endif
.endmacro
```

This allows usage as

```
.import foo, bar, baz
bb foo
bb foo, bar
bb foo, bar, baz
```

which is almost identical to what is implemented in this PR.

PR #2940, opened 2026-03-10
Compare operator test suite ahead of cc65/coptstop.c refactoring (refer #2935)

avolkov76:
## Summary

Compare operator test suite: forward and reverse compares; edge and corner cases; signed char, unsigned char, signed int, unsigned int; literals, autos, statics, simple expressions; some special cases. This is ahead of cc65/coptstop.c refactoring proposed in #2935.
There is overlap with existing `compare[1..10].c` and several other regression tests. It might be overkill, but it ensures just about every edge case is covered with forward and reverse forms. It may be more difficult to diagnose the failed cases due to the extensive use of preprocessor macros.

**Note:** Some of the compiled 6502 binaries are starting to push the limit. The largest is ~55KB.

## Checklist

### DIFF
--- a/test/val/compare-rev.h
+++ b/test/val/compare-rev.h
@@ -1,0 +1,92 @@
+/* Exercise compare operator cases in forward and reverse forms.
+** Common test framework and macros.
+** 
+** License: Public Domain
+**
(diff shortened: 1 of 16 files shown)


mrdudz:
great, thanks!

PR #2956, opened 2026-04-11
Refactor: Split `OptStackOps` into 12 top-level optimizers

avolkov76:
## Summary

This splits the monolithic `OptStackOps` optimizer into 12 top-level optimizers with operations grouped by operand size + operator kind. The change immediately enables optimization cases previously left behind through: (a) better ordering of optimizer steps (arithmetic and bitwise before compares); (b) automatic fallback from reg A-only optimizers to full AX ones.
The code size factors assigned to the top-level optimizers is currently unchanged from `OptStackOps` (all 100), and may need to be reviewed in the future.

Benefits:
- Finer control over optimizer step ordering and enable/disable state.
- Improved optimizer step statistics.
- Easier optimizer troubleshooting.

The drawback is more passes to be made over the code segments. But that does not appear to have a significant effect on the compilation speed.

c65-Chess compiled binaries are ~20 bytes smaller.

Observed compiled binary size reduction in the testbench (`diff` over `ls -lUR testwrk`):
```diff
--- cc65-git-head-test-sizes.lst	2026-04-11 09:15:44.870373100 -0400
+++ cc65-refactor-2-test-sizes.lst	2026-04-11 09:16:57.716798700 -0400

-pointer2.O.6502.prg 2923
+pointer2.O.6502.prg 2902
-pointer2.O.65c02.prg 2905
+pointer2.O.65c02.prg 2885

-bitfield-1.O.6502.prg 2941
+bitfield-1.O.6502.prg 2901
-bitfield-1.O.65c02.prg 2909
+bitfield-1.O.65c02.prg 2876

-bitfield-union.O.6502.prg 2807
+bitfield-union.O.6502.prg 2753
-bitfield-union.O.65c02.prg 2767
+bitfield-union.O.65c02.prg 2726

-bug1462-biefield-assign-1.O.6502.prg 3374
+bug1462-biefield-assign-1.O.6502.prg 3350
-bug1462-biefield-assign-1.O.65c02.prg 3350
+bug1462-biefield-assign-1.O.65c02.prg 3327

-bug1462-biefield-assign-2.O.6502.prg 2694
+bug1462-biefield-assign-2.O.6502.prg 2672
-bug1462-biefield-assign-2.O.65c02.prg 2676
+bug1462-biefield-assign-2.O.65c02.prg 2655

-bug1462-biefield-assign-3.O.6502.prg 3646
+bug1462-biefield-assign-3.O.6502.prg 3596
-bug1462-biefield-assign-3.O.65c02.prg 3621
+bug1462-biefield-assign-3.O.65c02.prg 3572

-bug2395.O.6502.prg 4441
+bug2395.O.6502.prg 4335
-bug2395.O.65c02.prg 4421
+bug2395.O.65c02.prg 4318

-bug2947.O.6502.prg 3862
+bug2947.O.6502.prg 3786
-bug2947.O.65c02.prg 3837
+bug2947.O.65c02.prg 3759

-cq85.O.6502.prg 3323
+cq85.O.6502.prg 3293
-cq85.O.65c02.prg 3311
+cq85.O.65c02.prg 3282
```

## Checklist

### DIFF
--- a/src/cc65/codeopt.c
+++ b/src/cc65/codeopt.c
@@ -122,6 +122,7 @@ static OptFunc DOptBNegAX1      = { OptBNegAX1,      "OptBNegAX1",      100, 0, 
 static OptFunc DOptBNegAX2      = { OptBNegAX2,      "OptBNegAX2",      100, 0, 0, 0, 0, 0 };
 static OptFunc DOptBNegAX3      = { OptBNegAX3,      "OptBNegAX3",      100, 0, 0, 0, 0, 0 };
 static OptFunc DOptBNegAX4      = { OptBNegAX4,      "OptBNegAX4",      100, 0, 0, 0, 0, 0 };
+static OptFunc DOptBZero        = { OptBZero,        "OptBZero",        100, 0, 0, 0, 0, 0 };
 static OptFunc DOptBinOps1      = { OptBinOps1,      "OptBinOps1",        0, 0, 0, 0, 0, 0 };
 static OptFunc DOptBinOps2      = { OptBinOps2,      "OptBinOps2",        0, 0, 0, 0, 0, 0 };
 static OptFunc DOptBoolCmp      = { OptBoolCmp,      "OptBoolCmp",      100, 0, 0, 0, 0, 0 };
@@ -187,6 +188,7 @@ static OptFunc DOptPtrLoad7     = { OptPtrLoad7,     "OptPtrLoad7",     140, 0, 
 static OptFunc DOptPtrStore1    = { OptPtrStore1,    "OptPtrStore1",     65, 0, 0, 0, 0, 0 };
 static OptFunc DOptPtrStore2    = { OptPtrStore2,    "OptPtrStore2",     65, 0, 0, 0, 0, 0 };
 static OptFunc DOptPtrStore3    = { OptPtrStore3,    "OptPtrStore3",    100, 0, 0, 0, 0, 0 };
+static OptFunc DOptPtrStore4    = { OptPtrStore4,    "OptPtrStore4",    100, 0, 0, 0, 0, 0 };
 static OptFunc DOptPush1        = { OptPush1,        "OptPush1",         65, 0, 0, 0, 0, 0 };
 static OptFunc DOptPush2        = { OptPush2,        "OptPush2",         50, 0, 0, 0, 0, 0 };
 static OptFunc DOptPushPop1     = { OptPushPop1,     "OptPushPop1",       0, 0, 0, 0, 0, 0 };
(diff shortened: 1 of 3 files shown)


src/cc65/codeopt.c
```diff
@@ -205,8 +207,17 @@ static OptFunc DOptShiftBack    = { OptShiftBack,    "OptShiftBack",      0, 0,
 static OptFunc DOptSignExtended = { OptSignExtended, "OptSignExtended",   0, 0, 0, 0, 0, 0 };
 static OptFunc DOptSize1        = { OptSize1,        "OptSize1",        100, 0, 0, 0, 0, 0 };
 static OptFunc DOptSize2        = { OptSize2,        "OptSize2",        100, 0, 0, 0, 0, 0 };
-static OptFunc DOptStackOps     = { OptStackOps,     "OptStackOps",     100, 0, 0, 0, 0, 0 };
 static OptFunc DOptStackPtrOps  = { OptStackPtrOps,  "OptStackPtrOps",   50, 0, 0, 0, 0, 0 };
```
kugelfuhr, line 210: Just a cosmetic issue, nothing serious: There are now names using "Stack" and other that abbreviate "Stack" to "Stk". I would suggest to always use either "Stk" or "Stack".

kugelfuhr (commented):
Good work!

src/cc65/codeopt.c
avolkov76: Let me see if I can fit the "OptStack" names into the existing tables. I originally started off with "OptStack..." but some of the names did not fit and I abbreviated to "OptStk...". However, I later shortened the names by using numbers.
"Bitwise" is the problem child. I can shorten it to "BinOps", which fits with "CmpOps" and follows your prior pattern. That makes it fit nicely. Does that work?

kugelfuhr: Reformatting should be easy. You can remove all of the counters if you want to shorten the lines. They are zero anyway.

Or, use a macro:

```
#define OPTFUNC(name, codesize)         \
    static OptFunc D##name = { name, #name, codesize, 0, 0, 0, 0, 0 }
```

The entries can then be defined using
```
OPTFUNC(Opt65C02Ind, 100);
```

However, the sort check might need an adjustment when using the macro.

avolkov76: Hmm, so let's open that can of worms, why not. 😉 Reformatting the optimizer structures should probably be done separately and _before_ this PR. Or at least in a separate commit before this one. I am willing to do the work if there is consensus.

Since you mentioned it, there is one thing missing in my mind: I was unable to track down the rationale for having the `static OptFunc DOpt...` definitions sorted. That is not necessary _functionally_. It appears to be either code discipline or cosmetic.

But, if we are reformatting with macros et al, we do not have to stop there. The `static OptFunc DOpt...` definitions can be merged with `static OptFunc* OptFuncs[]` into a single sorted table, removing some maintenance overhead. And the optimizers then invoked by name:
```C
RunOptFunc (S, "OptPtrLoad2", 1);

/* Replacing the ugly-ish loop I wrote: */
RunOptFuncGroup (S, 2, "OptStkArith1", "OptStkArith2", "OptStkBitwise1", "OptStkBitwise2");

/* Tidy-up RunOptGroup7(): */
RunOptFuncFollowUpGroup (S, "OptSize1",   "OptUnusedLoads", "OptAXLoad", "OptAXLoad2", ... );

```
There are ~130 optimizers after this PR, and I plan to add a few more. The additional `bsearch()` over the `OptFuncs[]` table for every invocation should be almost nothing on modern CPUs. And while cc65 is not a bootstrapping compiler, I am not sure about any retro platforms that might be trying to build and use cc65 though.

Of course, the helper functions offhandedly proposed above can be used with `DOpt...` definitions as well as string names. Variadics don't care.

Thoughts?

kugelfuhr: I have to admit that I'm not a fan of adding additional overhead. Maybe this is caused by my small device background😊 I think we had a discussion about this some time ago. So I would prefer to leave the `DOptXXX` entries separately. Variadics can still be used to simplify the functions.

Having the `DOptXXX` variables sorted is a requirement for the `sorted_codeopt.sh` script to check if both tables are equal. Meaning that each `DOptXXX` variable is also in the `OptFuncs` table and vice versa. It happened in the past that variables were added without an entry in the table.

avolkov76: How much of C99 are we allowing in the compiler codebase? I believe C99 designated initializers are a _prime_ feature to facilitate the optimizer tables while reducing maintenance overhead. Then only one list needs to be explicitly sorted, and there is no need to cross-check the lists.

Neither MSVC nor GCC have issues with this:
(Some `OPT()` macro could be used to also produce the `D##name` identifiers to take it further, but I rather like it the way it is.)
```C
#define OPTFUNCDEF(name, codesize) \
    [D##name] = { name, #name, codesize, 0, 0, 0, 0, 0 }

/* Declarations become indexes: */
typedef enum {
/* BEGIN SORTED_CODEOPT.SH */
    DOptCmp1,
    DOptCmp2,
    DOptCmp3,
/* END SORTED_CODEOPT.SH */

    OPTFUNC_COUNT_, /* XXX: Do not clash with OPTFUNC_COUNT for now */
} OptIndex;

/* Definitions are merged into one table with explicit size so that
 * missing entries will be automatically disabled. */
static OptFunc OptFunc_DesignatedInit[OPTFUNC_COUNT_] =
{
    /* For illustration of designated initializer: */
    [DOptCmp1] = { OptCmp1,  "OptCmp1",    42,    0, 0, 0, 0, 0 },
    /* Definitions can go in any order and still be in correct slots */
    OPTFUNCDEF (OptCmp3, 75),
    OPTFUNCDEF (OptCmp2, 85),
};

static unsigned RunOptFunc_ (CodeSeg* S, OptIndex Idx, unsigned Max)
{
    OptFunc* F = &OptFunc_DesignatedInit[Idx];

    CHECK (Idx >= 0 && Idx < OPTFUNC_COUNT_);

    // . . .

    return 0;
}

static unsigned RunOptGroupX (CodeSeg* S)
{
    unsigned Changes = 0;

    /* Optimizers invoked by index; no additional overhead */
    Changes += RunOptFunc_ (S, DOptCmp1, 1);
    Changes += RunOptFunc_ (S, DOptCmp2, 5);
    Changes += RunOptFunc_ (S, DOptCmp3, 5);

    return Changes;
}
```

kugelfuhr: The idea of using enums and array designators is quite clever and I really like it! It allows to merge anything into one table without additional overhead. However, the developers are usually shy when it comes to C99 features, so there has to be some official blessing.

avolkov76: >  the developers are usually shy when it comes to C99 features, so there has to be some official blessing.

So very possibly futile at the current response level. I could prepare an RFC PR regardless and just leave it there for a while. It shouldn't be more than a couple hours of work, plus 1 night of test scripts.

@mrdudz What are your thoughts on C99 array designators here? This is not a show-stopper, but more of a "while I am down there..." I could always claim that it is a _single_ line of C99 code, because it is -- the macro. 😉

mrdudz: > @mrdudz What are your thoughts on C99 array designators here? This is not a show-stopper, but more of a "while I am down there..." I could always claim that it is a single line of C99 code, because it is -- the macro. 😉

I'm totally fine with code like shown above (add some reasonable similar comments for us old farts please :))

(That said, i somehow remember C99 style array designators being used elsewhere...mmh)

avolkov76: Unfortunately, I have to scrap the enum+array designators idea. It just moves the problem from one end to the other. If someone were to remove an optimizer step definition from the table and leave the enum index, that would create a null entry in the table. And `bsearch()` cannot operate on sparse arrays. So the two lists still have to be cross-checked, making the change rather gratuitous.

But if you really want it for some reason, I can still make the change.

mrdudz: > But if you really want it for some reason, I can still make the change.

I'm usually in the "don't make changes unless they solve a problem" camp for that matter :)

avolkov76:
Unfortunately, GH will not allow me to reopen this PR after some branch manipulations. 🤷

PR #2957, opened 2026-04-13
Fix reverse lowercase on Apple][

### LINKED ISSUE #2831: revers() is broken for the apple2 platform
**Compiler and OS**
cc65 V2.19 - Git 29f7ab380

**Describe the bug**
using the revers() function produces yields reverse text garbage instead of the expected output

**To Reproduce**
This test program:

```
#include <apple2.h>
#include <conio.h>

int main(void) {
   revers(1);
   cputs("this should be reverse");
   revers(0);
   while(1);
   return 0;
}

```

compiled using
`cl65 -t apple2 -o test.apple2 -C ../src/apple2/apple2-system.cfg a2test.c `

generates the following output on an Apple //e:

Image

Normal text is printed ok, only reversed text is affected.

colinleroy:
Fixes #2831.
Thanks to Alex Volkov for the patch

### DIFF
--- a/libsrc/apple2/cputc.s
+++ b/libsrc/apple2/cputc.s
@@ -109,16 +109,14 @@ newline:
 :       jmp     VTABZ
 
 putchar:
-        .ifdef  __APPLE2ENH__
-        ldy     INVFLG
-        cpy     #$FF            ; Normal character display mode?
-        beq     putchardirect
         cmp     #$E0            ; Lowercase?
-        bcc     mask
-        and     #$7F            ; Inverse lowercase
-        bra     putchardirect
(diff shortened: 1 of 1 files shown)


avolkov76:
For what it's worth, I endorse this patch. 😉

oliverschmidt:
Thanks to @avolkov76 and @colinleroy - and sorry for the delay on my part!

avolkov76:
> sorry for the delay on my part!

No worries at all! I am sure you were busy with a2pico. 😉

oliverschmidt:
> I am sure you were busy with a2pico. 😉

Bullseye :-)

PR #2960, opened 2026-05-08
Fix sp65 PCX reader: 1bpp monochrome path wrote past bitmap allocation

polluks:
ReadPlane() already expands packed bits to one byte per pixel for case 1 (1bpp), but the loop in ReadPCXFile treated the buffer as still-packed bits using a mask-and-shift pattern. The for-loop also failed to increment X, causing an infinite loop that wrote past the allocated bitmap and triggered SIGBUS.

## Summary

Fix: use the same direct L[X] indexing as the 8bpp path, matching ReadPlane's pre-expanded output.

## Checklist


mrdudz:
thanks!

PR #2963, opened 2026-05-14
grc65: Added warning for string truncation

polluks:
## Summary

Don't truncate silently anymore.

## Checklist


mrdudz:
Mmmh. Would it be possible to also print the line number? Just a warning without any reference to where the problem is is a bit useless IMHO

mrdudz:
thanks!

PR #2966, opened 2026-06-22
RP6502 0.28 support, add argv, replace time

rumbledethumps:
Strictly an RP6502 SDK update. Interesting bits:

The argc/argv support on the RP6502 necessitated a hook for the app to provide the memory or reject too large of a request. I had to change the init priority so the app could use malloc if it wanted - which could be released after it's consumed.

The time system was replaced entirely with RP6502-OS calls. Nothing wrong with cc65 lib - it's just unavoidably large.


mrdudz:
Thanks!

PR #2969, opened 2026-07-17
Make noexec check portable across BSD and GNU find

hindermath:
## Summary

- avoid GNU-specific `find -executable` in the noexec style check
- keep the existing Makefile targets unchanged
- keep the check usable on both BSD/macOS and GNU/Linux find implementations

## Why

The existing check uses `find -executable`, which is not supported by BSD `find` on macOS. This makes `make checkstyle` emit a portability error on macOS even though the overall check can still complete.

## Validation

- `make checkstyle` on macOS

## Scope

This is a build/style portability fix only. No runtime code or user-facing behavior is changed.

### DIFF
--- a/.github/checks/noexec.sh
+++ b/.github/checks/noexec.sh
@@ -1,19 +1,22 @@
 #! /bin/bash
+
 OLDCWD=`pwd`
-SCRIPT_PATH=`dirname $0`
+SCRIPT_PATH=`dirname "$0"`
 CHECK_PATH=.
+FOUND=0
 
-cd $SCRIPT_PATH/../../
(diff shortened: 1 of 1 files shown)


mrdudz:
Is there really no simpler way to fix this? seems like a huge diff for fixing such a simple thing

hindermath:
Thanks, agreed. I simplified the PR: it now only changes `noexec.sh` and removes the extra test/Makefile changes. The fix no longer uses GNU-specific `find -executable`; it keeps the existing file selection and checks executability in the shell instead.

mrdudz:
thanks!

PR #2970, opened 2026-07-17
Fix RP6502 RIA readline error return

hindermath:
## Summary

- return immediately from `ria_rln_peek()` when the RIA call reports an error
- keep successful zero-length reads stack-balanced by still popping the returned position
- remove the `ria_execl()` / `ria_execv()` changes from this PR

## Why

This follows the RP6502 maintainer feedback in the discussion. The accepted part is the `ria_rln_peek()` error-path cleanup; the exec argument changes were dropped because XSTACK and OS-level validation already cover those cases.

## Validation

- `make checkstyle` passed
- `make bin` passed
- `make lib` passed
- `make test` was started after the reduced patch and progressed into `sim65c02` validation without a visible failure, but was stopped locally because the full matrix runtime was excessive
- GitHub PR CI is running the full Linux and Windows validation for the pushed commit

### DIFF
--- a/libsrc/rp6502/ria_rln_peek.c
+++ b/libsrc/rp6502/ria_rln_peek.c
@@ -4,13 +4,12 @@ int __fastcall__ ria_rln_peek (char* peek, unsigned char* pos)
 {
     int i, ax;
     ax = ria_call_int (RIA_OP_RLN_PEEK);
-    if (ax > 0) {
-        *pos = ria_pop_char ();
-        for (i = 0; i < ax; i++) {
-            peek[i] = ria_pop_char ();
-        }
-    } else {
-        *pos = 0;
+    if (ax < 0) {
+        return ax;
+    }
+    *pos = ria_pop_char ();
+    for (i = 0; i < ax; i++) {
+        peek[i] = ria_pop_char ();
     }
     peek[ax] = 0;
     return ax;


mrdudz:
I am not sure something like this makes sense in the library - does this solve an actual problem in practise?

hindermath:
Fair question. I do not want to overstate this as a major security fix. The intent is a small boundary-validation cleanup: reject invalid readline indexes and length accumulation edge cases before using them.

If the project preference is that this validation should stay with callers / the RP6502 environment rather than the cc65 library layer, I am fine with adjusting or closing the PR. Guidance on that boundary would be useful.

mrdudz:
I'd leave the final decision to those who use or program for the RP6502 target.... If this makes sense in every day use, then why not. Generally however, the cc65 library is always a compromise between code size (this often being the major issue), speed, and "completeness"

hindermath:
Hi @rumbledethumps and @GorillaSapiens, could I ask for your RP6502 perspective on this PR?

The PR adds small boundary checks in `ria_rln_peek.c`, `ria_execl.c`, and `ria_execv.c`: reject invalid readline indexes before use and guard exec argument length accumulation.

My intent is not to overstate this as a major security fix, but as a robustness / boundary-validation cleanup.

@rumbledethumps, since you authored and maintain most of the RP6502 target code, your view would be especially useful: do these checks make sense in the cc65 RP6502 library layer, or should this remain the responsibility of callers / the RP6502 environment because of code size and library tradeoffs?

@GorillaSapiens, from a practical RP6502 user perspective, would this kind of validation be useful in everyday use, or is the extra library code not worth the code-size tradeoff?

rumbledethumps:
The rln peek early return should be adopted. That fixes a stray pointer on error, which doesn't happen in practice, but should be fixed on principle. However, the fix is broken, because a success with len=0 still has the pos on the stack. So the next call that uses the stack with see this leftover and throw an error. A failed OS call is not required to make the return values safe, so just do this: ```if (ax < 0) {return ax;}```

The exec stuff doesn't add any value. XSTACK has over/underrun protection. The truncated stack won't validate by the OS and you get the EINVAL that way. Even if you created pathological args, you're not bypassing any auth.

I will look at that clock test next time I'm working on cc65:
https://github.com/picocomputer/rp6502/issues/189

hindermath:
Thanks, that makes sense. I reduced the PR accordingly:

- kept only the `ria_rln_peek()` error return
- changed it to `if (ax < 0) { return ax; }` so `len == 0` still pops the returned position and leaves the stack balanced
- removed the `ria_execl()` / `ria_execv()` changes from this PR

The branch is rebased onto current `master` and force-pushed with lease. Local validation so far:

- `make checkstyle` passed
- `make bin` passed
- `make lib` passed
- `make test` is running locally after the reduced patch; no failure has appeared so far

hindermath:
Small validation update: I stopped the local `make test` run after it had progressed into the `sim65c02` validation matrix without a visible failure. The full local matrix runtime was excessive for this small one-file change.

The completed local checks remain:

- `make checkstyle`
- `make bin`
- `make lib`

The pushed commit is `2dede6fda`, and the PR CI is running the full Linux and Windows validation for that commit.

mrdudz:
That patch looks reasonable, thanks!

PR #2971, opened 2026-07-17
Fix scanning of /* when CComments is disabled

tlgkccampbell:
## Summary

An oversight in the ca65 scanner causes an error to be produced when a `*` token immediately follows a `/` token, even when the `c_comments` feature is disabled. Specifically:

```c
case '/':
    NextChar ();
    if (C != '*') {
        CurTok.Tok = TOK_DIV;
    } else if (CComments) {
         /* ... skip the comment ... */
         goto Again;
    }
    return;
```

In the case where `C` is `*`, but `CComments` is *disabled*, `CurTok.Tok` is never assigned and the previous token is simply repeated. This PR clarifies the conditional expression and ensures that the token is correctly assigned everywhere that is should be.

## Checklist

### DIFF
--- a/src/ca65/scanner.c
+++ b/src/ca65/scanner.c
@@ -1389,15 +1389,13 @@ CharAgain:
 
         case '/':
             NextChar ();
-            if (C != '*') {
-                CurTok.Tok = TOK_DIV;
-            } else if (CComments) {
+            if (C == '*' && CComments) {
                 /* Remember the position, then skip the '*' */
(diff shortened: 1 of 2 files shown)


mrdudz:
good catch - it would be nice if that regression test would also check the actual value produced by that division though (should be easy enough)

tlgkccampbell:
> good catch - it would be nice if that regression test would also check the actual value produced by that division though (should be easy enough)

OK, I've updated the test to assert against an absolute PC.

mrdudz:
thanks!

PR #2972, opened 2026-07-18
Fix parsing of octal values in escaped strings

tlgkccampbell:
## Summary

A bug in the ca65 scanner prevents escaped octal values from being interpreted correctly when C-style string escapes are enabled. Consider the following code:

```
.feature string_escapes
.byte "\101"
```

The escaped value here *should* be octal 101, i.e. decimal 65. Instead, the scanner produces a `Newline in string constant` error. The `Final` variable created in `ReadStringConst` is never actually used anywhere, and the loop that consumes octal digits runs past the end of the escaped value and consumes the `"` terminator, causing the entire string to be scanned incorrectly.

This PR modifies `ReadStringConst` to keep track of the lookahead character produced by the loop, restoring it when needed. New tests ensure both that an octal escape is processed correctly, and that a value higher than 255 produces the correct range error.

## Checklist


mrdudz:
looks good, thanks!

PR #2975, opened 2026-07-29
ca65: Add --merge-scopes

codemeow:
## Summary

Adds the `--merge-scopes` option to ca65.

At this memont, a named `.scope` can only be defined once in its parent scope. This makes large namespaces hard to organize: for example, a chip description containing `CHIPNAME::REGISTERS`, `CHIPNAME::MEMORY`, and other groups must live in one arge source file. Splitting these groups into separate include files produces a duplicate-scope error.

With `--merge-scopes`, separate source or include files may reopen the same named `.scope` and contribute different nested namespaces to it. This makes large scope definitions easier to organize while preserving the current behavior by default.

Duplicate symbol definitions inside a merged scope still produce an error. `.sizeof()` accumulates the sizes of all occurrences, and debug spans are kept for every occurrence.

Regression tests cover:
- merging scope fragments in one file;
- merging scope fragments from include files;
- nested scopes;
- a project-style namespace split between `HARDWARE` and `SOFTWARE` include files;
- duplicate symbols in a merged scope;
- unchanged default behavior without `--merge-scopes`.

### Example

Imagine defining a huge project settings or (mentioned) chip description. In my case it was the global scope for my project's symbols. Without this patch I *must* list all the project-related constants either:
- In one huge file, wrapped with my project's scopename:
```
.scope MYPRJ
    .scope MEMTABLE
        ...
    .endscope
    .scope HARDWARE
        ...
    .endscope
    x N different definitions
.endscope
```
- Or without the global scope at all if I want to split it between smaller files to organize my workspace:
_myprj/memtable.inc_:
```
.scope MYPRJ_MEMTABLE
    ...
.endscope
```
_myprj/hardware.inc_:
```
.scope MYPRJ_HARDWARE
    ...
.endscope
```

With this patch the subscope data could be split between files to help organize the workspace/isolate the edits:
_myprj/memtable.inc_:
```
.scope MYPRJ
    .scope MEMTABLE
        ...
    .endscope
.endscope
```
_myprj/hardware.inc_:
```
.scope MYPRJ
    .scope HARDWARE
       ...
    .endscope
.endscope
```
And thus be used as if it was defined in one huge file:
```
    .include "myprj/memtable.inc"
    .include "myprj/hardware.inc"
    
    .segment "RODATA"
    
    testdef:
        .byte MYPRJ::MEMTABLE::VAL1
        .byte MYPRJ::HARDWARE::CHIP1
```

### How it works

Technically, the option reopens the existing symbol table when ca65 encounters a named `.scope` with the same parent, instead of reporting a duplicate scope. Symbols from every occurrence share one namespace, while duplicate symbol definitions are still diagnosed by the existing symbol-definition checks. Each occurrence tracks its own debug spans before they are appended to the scope’s complete span list, and the scope size accumulates the sizes of all occurrences.

## Checklist

- [x] The documentation has been updated

Tests:
- `make -C test/asm all`
- `make -C .github/checks checkstyle`


codemeow:
WIP: fixing the windows build and adding corresponding `.feature` for this one
UPD:
Done.
- Windows build successful
- `.feature` option `merge_scopes` added

kugelfuhr:
The code seems to work as described and probably has its uses. I'm not sure if I like the idea of scopes changing between two uses (for example, `.sizeof` may yield different results). But on the other side, the feature must be explicitly enabled, so it's probably ok.

mrdudz:
I agree, seems a useful feature to me as well (and doesnt break anything when not enabled)

thanks!

PR #2977, opened 2026-08-05
Revert "Serial: Optimize SER_GET"

### LINKED ISSUE #2974: Serial driver needs a way to flush the transmit buffer
I'm messing with the terminal sample and VICE. I'm currently focusing only on the C64, but a lot of the serial drivers use the same code.
Currently there's no way to flush the transmit buffer cleanly, and it seams to be a design issue with the driver.

The problem:
- When there's a lot of data to receive the NmiHnadler will terminate the transmit in TryToSend. This leaves data in the transmit buffer.
- The documentation is explicit that transmitting is not interrupt driven, and the documentation is correct in that regard.
- The only way to flush the transmit buffer is to send more bytes.

The solution?
- Allow calling the TryToSend procedure in the serial driver.
- It would be nice to get back the number of free and queued bytes.

To have some tangibles I made a sample to showcase the problem. Please see [this repo](https://github.com/bananabeard/cc65-ser-try-to-send).
- I've modified the C64 serial driver so ioctl calls TryToSend.
- There's a modified terminal that calls TryToSend.
- There's a server that send whole-screen updates on every keypress.

colinleroy:
This reverts commit 734541ee05000bad9e9dd2f903a2db754e8c36f3. 
Fixes #2974.
The ACIA can't send while RTS is off, so this commit was wrong.

### DIFF
--- a/libsrc/apple2/ser/a2.ssc.s
+++ b/libsrc/apple2/ser/a2.ssc.s
@@ -354,8 +354,13 @@ Out:
 
 SER_GET:
         ldx     Index
+        ldy     SendFreeCnt     ; Send data if necessary
+        iny                     ; Y == $FF?
+        beq     :+
+        lda     #$00            ; TryHard = false
+        jsr     TryToSend
 
-        lda     RecvFreeCnt     ; Check for buffer empty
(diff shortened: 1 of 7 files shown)


mrdudz:
thanks!

PR #2981, opened 2026-09-01
Add target for the Apple3

robjustice:
Add target for the Apple ///. The default cfg and header support creates a SOS Interpreter that can be run from Apple /// SOS.

It is mostly the Apple2 lib code ported to run on SOS (instead of ProDOS). And supporting the Standard Apple /// console driver for character input/output. This has its challenges, but i have tried to keep the original intent of how SOS and its drivers were designed to be used.

The code has some room for size optimising, its been more of get things working correctly first.

This has ended up quite a bit bigger than i was expecting to commit in one hit, its sort of grew to get what I wanted to run working.


polluks:
finally update README.md

PR #2987, opened 2026-09-12
Fix indirect Goto

### LINKED ISSUE #1209: backwards indirect goto does not compile

ikorb:
Add tracking of labels used for indirect jumps to fix two bugs related to indirect gotos.

Disclaimer: Patch and commit message are AI generated. I glanced over the changes and they look plausible to me, but I'm not an expert in cc65 internals.

## Summary

Fixes #1209 and #1211

## Checklist


avolkov76:
The first overall impression is good. I see a couple places where it could use some refinements. This also adds a couple fixes for previously undiscovered optimizer issues (see `CS_IsBasicBlock()` and `CS_GenRegInfo()`) related __and__ possibly unrelated to the computed goto labels. These will need a deeper look before giving thumbs-up. I do not know that code well enough yet.

src/cc65/codelab.c
```diff
@@ -108,8 +109,13 @@ void CL_MoveRefs (CodeLabel* OldLabel, CodeLabel* NewLabel)
 ** more references on return.
 */
 {
+    unsigned Count;
+
+    /* The new label inherits the properties of the old one */
+    NewLabel->Flags |= OldLabel->Flags;
```
avolkov76: This is a hidden invariant (no-op) because we are not actually able to move indirect jump references (yet).
```suggestion
    /* References to indirect jump target labels for computed goto are not
    ** tracked and cannot be moved.
    */
    CHECK (!CL_IsIndJumpTarget (OldLabel));
```

src/cc65/codelab.h
```diff
@@ -59,12 +59,16 @@ struct CodeEntry;
 
 
 
+/* Flags used */
+#define LF_INDJUMP      0x0001U         /* Label may be reached by an indirect jump */
```
avolkov76, line 63:
```suggestion
#define CLF_INDJUMP     0x0001U         /* Label may be reached by an indirect jump */
```

src/cc65/codeseg.c
```diff
@@ -887,6 +856,27 @@ CodeLabel* CS_GenLabel (CodeSeg* S, struct CodeEntry* E)
 
 
 
+void CS_MarkIndJumpTarget (CodeSeg* S, const char* Name)
```
avolkov76, line 859:
```suggestion
void CS_RegIndJumpLabel (CodeSeg* S, const char* Name)
```

src/cc65/codeseg.c
```diff
@@ -983,10 +973,10 @@ void CS_MergeLabels (CodeSeg* S)
             continue;
         }
 
-        /* Pick a label to keep, all references will be moved to this one, and the other labels
-        ** will be deleted.  PickRefLab moves this to index 0.
+        /* Keep the first label, all references will be moved to this one, and
+        ** the other labels will be deleted.
         */
-        RefLab = PickRefLab (E);
```
avolkov76: I think we should retain the `PickRefLab()` heuristic (in modified form) to increase the chances of labels getting merged.

src/cc65/codeseg.c
```diff
@@ -1551,7 +1569,20 @@ void CS_GenRegInfo (CodeSeg* S)
...
+                    /* We don't know all entry points of this insn: it may be
+                    ** the target of an indirect jump, it may have more than
+                    ** one label of which we look at just the first one, or
+                    ** the preceeding insn was an unconditional branch and no
+                    ** insn jumps here. Assume unknown register contents and
+                    ** skip the loop over the entry points below.
+                    */
+                    RC_Invalidate (&Regs);
+                    RC_InvalidatePS (&Regs);
+                    Entry = CL_GetRefCount (Label);
```
avolkov76: Could use a comment: "Skip the rest of the entries."

src/cc65/goto.c
```diff
@@ -180,11 +162,7 @@ void DoLabel (void)
     SymEntry* Entry = AddLabelSym (CurTok.Ident, SC_DEF);
 
     /* Emit the jump label */
-    CodeLabel* L = CS_AddLabel (CS->Code, LocalLabelName (Entry->V.L.Label));
-
-    if (Entry->V.L.IndJumpFrom) {
-        CollAppend (&L->JumpFrom, Entry->V.L.IndJumpFrom);
-    }
+    CS_AddLabel (CS->Code, LocalLabelName (Entry->V.L.Label));
```
avolkov76, line 165: Revert to original; use the available top-level codegen functions:
```suggestion
    g_defcodelabel (Entry->V.L.Label);
```

src/cc65/codeseg.c
```diff
@@ -1551,7 +1569,20 @@ void CS_GenRegInfo (CodeSeg* S)
                 */
                 CodeLabel* Label = CE_GetLabel (E, 0);
                 unsigned Entry;
-                if (WasJump) {
+                if (CE_GetLabelCount (E) > 1                    ||
+                    CL_IsIndJumpTarget (Label)                  ||
+                    (WasJump && CL_GetRefCount (Label) == 0)) {
```
avolkov76, line 1574: This is a test for a condition that would have existed without computed goto: a label with no known references following a jump instruction. Before, the `if (WasJump) { }` code simply assumed the label had at least one reference -- a potential silent bug. While I am not able to trace this condition to a use case known to me, it makes the optimizer safer.

src/cc65/codeseg.h
```diff
@@ -195,6 +195,12 @@ CodeLabel* CS_GenLabel (CodeSeg* S, struct CodeEntry* E);
 ** create a new label, attach it to E and return it.
 */
 
+void CS_MarkIndJumpTarget (CodeSeg* S, const char* Name);
```
avolkov76, line 198:
```suggestion
void CS_RegIndJumpLabel (CodeSeg* S, const char* Name);
```

src/cc65/codeseg.c
```diff
@@ -183,37 +183,6 @@ static void CS_RemoveLabelFromHash (CodeSeg* S, CodeLabel* L)
```
avolkov76: Retain `PickRefLab()` in modified form:
```suggestion
static CodeLabel* PickRefLab (CodeEntry* E)
/* Pick a reference label and move it to index 0 in E. */
{
    unsigned I;
    unsigned LabelCount = CE_GetLabelCount (E);
    CodeLabel* L0;

    CHECK (LabelCount > 0);

    /* Use either the first one as reference label, or an indirect jump target label.
    ** This is a work around for computed goto labels referenced from data segments.
    ** These references are not tracked. We pick an indirect jump target label
    ** because it must remain, and this increases the chances of merging the labels.
    */
    L0 = CE_GetLabel (E, 0);
    if (CL_IsIndJumpTarget (L0)) {
        /* The first label is already an indirect jump target; done. */
        return L0;
    }

    for (I = 1; I < LabelCount; ++I) {
        CodeLabel* L = CE_GetLabel (E, I);

        if (CL_IsIndJumpTarget (L)) {
            /* Move it to the beginning since it's simpler to handle the removal this way. */
            CE_ReplaceLabel (E, L, 0);
            CE_ReplaceLabel (E, L0, I);
            return L;
        }
    }

    return L0;
}
```

src/cc65/goto.c
avolkov76, line 1: After all updates with suggestions, the following `#include's` are not needed in goto.c and should be removed:
```diff
-#include "asmlabel.h"
-#include "codeent.h"
-#include "codeseg.h"
-#include "function.h"
```

src/cc65/expr.c
```diff
@@ -1201,6 +1203,14 @@ static void Primary (ExprDesc* E)
                 SymEntry* Entry;
                 NextToken ();
                 Entry = AddLabelSym (CurTok.Ident, SC_REF | SC_GOTO_IND);
+                /* Taking the address of a label means that it may be the
+                ** target of an indirect jump, and that it is referenced from
+                ** a data segment. Tell the code segment about it, so the
+                ** optimizer keeps the label and doesn't assume that it knows
+                ** all entry points of the code that follows it.
+                */
+                CS_MarkIndJumpTarget (CS->Code,
```
avolkov76: This is a novel operation which I think deserves a new top-level codegen function `g_regindcodelabel()` even if only used once. Without it, expr.c becomes needlessly aware of `CodeSeg`, increasing coupling.

```suggestion
                g_regindcodelabel (Entry->V.L.Label);
```

Move the line to codegen.c :
```C
void g_regindcodelabel (unsigned label)
/* Register an indirect jump target local code label */
{
    CS_RegIndJumpLabel (CS->Code, LocalLabelName (label));
}
```

src/cc65/expr.c
```diff
@@ -20,6 +20,7 @@
 #include "asmstmt.h"
 #include "assignment.h"
 #include "codegen.h"
+#include "codeseg.h"
```
avolkov76: Not needed with `g_regindcodelabel()`

src/cc65/expr.c
```diff
@@ -31,6 +32,7 @@
 #include "macrotab.h"
 #include "preproc.h"
 #include "scanner.h"
+#include "segments.h"
```
avolkov76: Not needed with `g_regindcodelabel()`

src/cc65/expr.c
```diff
@@ -1201,6 +1203,14 @@ static void Primary (ExprDesc* E)
...
                 NextToken ();
                 Entry = AddLabelSym (CurTok.Ident, SC_REF | SC_GOTO_IND);
+                /* Taking the address of a label means that it may be the
+                ** target of an indirect jump, and that it is referenced from
+                ** a data segment. Tell the code segment about it, so the
+                ** optimizer keeps the label and doesn't assume that it knows
+                ** all entry points of the code that follows it.
+                */
+                CS_MarkIndJumpTarget (CS->Code,
+                                      LocalLabelName (Entry->V.L.Label));
```
avolkov76, line 1213: This is a novel operation which I think deserves a new top-level codegen function `g_regindcodelabel()` even if only used once. Without it, expr.c becomes needlessly aware of `CodeSeg`, increasing coupling.

```suggestion
                g_regindcodelabel (Entry->V.L.Label);
```

Move the line to codegen.c :
```C
void g_regindcodelabel (unsigned label)
/* Register an indirect jump target local code label */
{
    CS_RegIndJumpLabel (CS->Code, LocalLabelName (label));
}
```

src/cc65/codeseg.c
```diff
@@ -183,37 +183,6 @@ static void CS_RemoveLabelFromHash (CodeSeg* S, CodeLabel* L)
...
-            if (EJ->JumpTo == NULL) {
-                /* Move it to the beginning since it's simpler to handle the removal this way. */
-                CE_ReplaceLabel (E, L, 0);
-                CE_ReplaceLabel (E, L0, I);
-                return L;
-            }
-        }
-    }
-    return L0;
-}
```
avolkov76, line 213: Retain `PickRefLab()` in modified form:
```suggestion
static CodeLabel* PickRefLab (CodeEntry* E)
/* Pick a reference label and move it to index 0 in E. */
{
    unsigned I;
    unsigned LabelCount = CE_GetLabelCount (E);
    CodeLabel* L0;

    CHECK (LabelCount > 0);

    /* Use either the first one as reference label, or an indirect jump target label.
    ** This is a work around for computed goto labels referenced from data segments.
    ** These references are not tracked. We pick an indirect jump target label
    ** because it must remain, and this increases the chances of merging the labels.
    */
    L0 = CE_GetLabel (E, 0);
    if (CL_IsIndJumpTarget (L0)) {
        /* The first label is already an indirect jump target; done. */
        return L0;
    }

    for (I = 1; I < LabelCount; ++I) {
        CodeLabel* L = CE_GetLabel (E, I);

        if (CL_IsIndJumpTarget (L)) {
            /* Move it to the beginning since it's simpler to handle the removal this way. */
            CE_ReplaceLabel (E, L, 0);
            CE_ReplaceLabel (E, L0, I);
            return L;
        }
    }

    return L0;
}
```

src/cc65/codeseg.c
```diff
@@ -983,10 +973,10 @@ void CS_MergeLabels (CodeSeg* S)
             continue;
         }
 
-        /* Pick a label to keep, all references will be moved to this one, and the other labels
-        ** will be deleted.  PickRefLab moves this to index 0.
+        /* Keep the first label, all references will be moved to this one, and
+        ** the other labels will be deleted.
         */
-        RefLab = PickRefLab (E);
+        RefLab = CE_GetLabel (E, 0);
```
avolkov76, line 979: I think we should retain the `PickRefLab()` heuristic (in modified form) to increase the chances of labels getting merged.

src/cc65/codeseg.c
avolkov76, line 1584: Could use a comment:
```suggestion
                    /* Skip the rest of the entries. */
                    Entry = CL_GetRefCount (Label);
```

src/cc65/expr.c
avolkov76, line 23: Not needed with `g_regindcodelabel()`

```suggestion
```

src/cc65/expr.c
avolkov76, line 35: Not needed with `g_regindcodelabel()`

```suggestion
```

src/cc65/codelab.c
avolkov76, line 115: This is a hidden invariant (no-op) because we are not actually able to move indirect jump references (yet).
```suggestion
    /* References to indirect jump target labels for computed goto are not
    ** tracked and cannot be moved.
    */
    CHECK (!CL_IsIndJumpTarget (OldLabel));
```

avolkov76 (commented):
With refinements, I would stand behind this fix. I can generate a clean, refined branch if there is interest.

### Pros:
- Removes some hacks related to computed goto.
- Makes the optimizer safer. Previously, the optimizer could subtly bug in computed goto scenarios on a 6502 because the entry points came from the wrong instruction, and the determined register contents could be incorrect (e.g. `RegContents` tracks `ptr1` while `callax` runtime clobbers it undetected.)
- No evident effect on code that is not using computed goto.
  - No diff in unrelated tests in the testbench or samples.
  - No diff in checked projects.

### Cons:
- Does not improve label reference tracking.
- All code using label references must also be aware of untracked indirect jump labels from now on (not a big deal; but some common `CL_xxx()` predicates are probably warranted.)

### Observations:
- Abandons indirect jump label reference tracking entirely and simply makes the optimizer reset upon encountering such labels. IMO, a safe approach overall, considering how many ways of using computed goto in user code exist.
- We do not have any tests that exercise the optimizer in computed goto scenarios.

printwithbrackets:
looked at the code, im not a professional on cc65 and it looks decent but naming wise the flag should be CLF_INDJUMP, not LF_INDJUMP because codeent uses CEF for it flags so codelab should use CLF

mrdudz:
Just for the records: i prefer this one over #2986 - but that is just a gut feeling and i cant really judge the code at all.

I'd like more tests exercising the various problems that might occur for that matter. And someone like @kugelfuhr could perhaps look at it :)

(And to be this isn't a very pressing issue so i am not merging this early)

Issue #2989, opened 2026-09-13
[NES] Compiled programs fail to boot if the PPU is reset


Merlin1247:
**Compiler version**
CC65 V2.19 (but I don't think it matters)

**Describe the bug**
NES ROMs compiled with cc65 don't boot on console when it just powered on or got RESET

**To Reproduce**
From what I can tell, the issue occurs because the NES's library fails to wait until the PPU has "warmed up" before it initializes PPU Registers.

Most emulators don't seem to emulate this behaviour, and so the ROM "erroneously" boots without issues.
On the latest version of Mesen, enable the emulation setting "Clear PPU $2000/1/5/6/7 throughout first frame after reset/power on".
I reproduced the issue on my flashcart(Everdrive N8 pro) by enabling "Reset delay", launching the ROM and pressing RESET.

[Video of me reproducing the issue](https://github.com/user-attachments/assets/fd0ed237-87f6-429d-8794-0dd386491189)

For the sake of completeness, here is the code:
```
#include <stdlib.h>
#include <stdint.h>
#include <conio.h>

int main(void)
{
  clrscr();
  gotoxy(2,2);
  cputs("Hello from C");

  while(1) { }
}
```

**To Fix**
If my hypothesis is correct, call `_waitvsync` twice in the library's init code before interacting with any other PPU registers

mrdudz:
Would be great if you could try that locally and submit a PR in case it works :)

Merlin1247:
For sure. I also saw a couple other things in the lib that I should be able to optimize so I might do that while i'm at it

PR #2991, opened 2026-09-20
NES Library update

Merlin1247:
## Summary

- Provided an implementation for `textcolor(char color)` in `conio.h`
- Updated the header to NES2.0 
- Updated the `.gitattributes` to prefer LF line endings
- Stored CHR-ROM in a binary file (instead of `.byte` commands)
- Freed up some memory, increased the C stack size from `$300` to `$400`
- Fixed issues #2989, #1570
- Fixed the first frame of startup displaying an uninitialized nametable
- Fixed the `conio` peek functions causing graphical glitches
- Made the `waitvsync()` function more robust
- Various other code optimizations


libsrc/nes/crt0.s
```diff
@@ -30,44 +32,22 @@
...
-;    | ..-EOF |      | CHR-ROM pages (in ascending order).      |
-;    +--------+------+------------------------------------------+
-
-        .byte   $4e,$45,$53,$1a ; "NES"^Z
-        .byte   2               ; ines prg  - Specifies the number of 16k prg banks.
-        .byte   1               ; ines chr  - Specifies the number of 8k chr banks.
-        .byte   %00000011       ; ines mir  - Specifies VRAM mirroring of the banks.
-        .byte   %00000000       ; ines map  - Specifies the NES mapper used.
-        .byte   0,0,0,0,0,0,0,0 ; 8 zeroes
+; For complete documentation, visit https://nesdev.org/wiki/NES_2.0
```
mrdudz: I don't think removing documentation is a good idea - please fix the documentation instead (and you can still add the link of course)

libsrc/nes/setcursor.s
```diff
@@ -1,37 +1,39 @@
...
-        .endrepeat
+        sty      SCREEN_PTR      ; Target val: --1000xx xxxyyyyy
+        inx
+        txa
+        sec
+        ror     a
+        ror     a
+        ror     a
+        sta     SCREEN_PTR+1    ; Contains garbage in the most significant 2 bits
+        and     #%11000000
```
mrdudz: cant you use LSR instead of ROR and then you dont need the masking?

.gitattributes
mrdudz: This shouldn't be part of this PR (not sure what implications it would even have)

cfg/nes.cfg
```diff
@@ -1,13 +1,13 @@
 SYMBOLS {
-    __STACKSIZE__: type = weak, value = $0300; # 3 pages stack
+    __STACKSIZE__: type = weak, value = $0400; # 4 pages stack
```
mrdudz: Is that really necessary? I have written some huge programs and just used one page of stack with them :)

libsrc/nes/crt0.s
Merlin1247: Fair enough!

libsrc/nes/setcursor.s
Merlin1247: LSR wouldn't work there, at least not without making the code larger and slower. I'm using the ROR to shift the low 3 bits of the Y position into the carry and most significant 2 accumulator bits. The point of this is that, instead of loading the Y position twice for both address bytes, while shifting the hi 2 Y bits to the least significant bits of the MSB anyway, I shift the low bits into C and "unused" bits of A (the NES's PPU address bus is only 14 bits, so the most significant 2 bits don't do anything). The AND is there to isolate those bits so that they can be used for the LSB. Also it saves a byte since I can use an SEC instead of ORA.

I should probably add better documentation here, that code is pretty messy.

.gitattributes
Merlin1247: Oops. Was having issues with my IDE saving files as CRLF so I changed that in an attempt to fix the issue. Didn't mean to push this

cfg/nes.cfg
Merlin1247: Objectively no, but I couldn't think of anything better to use it for

